PR #20, opened 2017-09-01
[WIP] Add cryptographic code to client.py

tuxxy:
This fills out the contents of the API commented in `client.py`.

This is a work-in-progress and all code is subject to change before final merge.

### DIFF
--- a/nkms/client.py
+++ b/nkms/client.py
@@ -61,13 +61,7 @@ class Client(object):
         :rtype: list of bytes
         """
         dirs = path.split(b'/')
-        subpaths = []
-        path = b'/'
-        for subdir in dirs:
-            path += subdir
-            subpaths.append(path)
-        subpaths[-1].rstrip(b'/')
-        return subpaths
+        return [b'/'.join(dirs[:i + 1]) for i in range(len(dirs))]
 
     def encrypt_key(self, key, pubkey=None, path=None, algorithm=None):
         """
--- a/nkms/client.py
+++ b/nkms/client.py
@@ -1,6 +1,6 @@
 from nkms.network import dummy
 from nkms.crypto import (default_algorithm, pre_from_algorithm,
-    symmetric_from_algorithm)
+    symmetric_from_algorithm, kmac)
 from nacl.utils import random
 
 
@@ -30,13 +30,24 @@ class Client(object):
             or created
         """
         self._nclient = Client.network_client_factory()
-
         self._pre = pre_from_algorithm(default_algorithm)
 
         # TODO: Check for existing keypair before generation
         # TODO: Save newly generated keypair
         self._priv_key = self._pre.gen_priv(dtype='bytes')
         self._pub_key = self._pre.priv2pub(self._priv_key)
+
+    def _derive_path_key(self, path):
+        """
+        Derives a public key for the specific path.
+
+        :param str path: Path to generate key for.
+
+        :return: Derived key
+        :rtype: bytes
+        """
+        priv = kmac.KMAC_256().digest(self._priv_key, path.encode())
+        return self._pre.priv2pub(priv)
 
     def encrypt_key(self, key, pubkey=None, path=None, algorithm=None):
         """
@@ -65,13 +76,19 @@ class Client(object):
         :return: Encrypted key(s)
         :rtype: bytes
         """
-        # TODO Encryption by path
         if not pubkey:
             pubkey = self._pub_key
 
-        # Encrypt symmetric key
-        enc_key = self._pre.encrypt(pubkey, key)
-        return enc_key
+        if len(path) > 1 and not type(path) is str:
+            enc_keys = []
+            for subpath in path:
+                path_pubkey = self._derive_path_key(subpath)
+                enc_keys.append(self.encrypt_key(key, pubkey=path_pubkey))
+            return enc_keys
+        elif len(path) == 1 or type(path) is str:
+            return self.encrypt_key(key, pubkey=self._derive_path_key(path))
+        elif not path:
+            return self._pre.encrypt(pubkey, key)
 
     def decrypt_key(self, enc_key, pubkey=None, path=None, owner=None):
         """
--- a/nkms/client.py
+++ b/nkms/client.py
@@ -37,17 +37,18 @@ class Client(object):
         self._priv_key = self._pre.gen_priv(dtype='bytes')
         self._pub_key = self._pre.priv2pub(self._priv_key)
 
-    def _derive_path_key(self, path):
+    def _derive_path_key(self, path, is_pub=True):
         """
         Derives a public key for the specific path.
 
         :param str path: Path to generate key for.
+        :param bool is_pub: Is the derived key a public key?
 
         :return: Derived key
         :rtype: bytes
         """
         priv = kmac.KMAC_256().digest(self._priv_key, path.encode())
-        return self._pre.priv2pub(priv)
+        return self._pre.priv2pub(priv) if is_pub else priv
 
     def encrypt_key(self, key, pubkey=None, path=None, algorithm=None):
         """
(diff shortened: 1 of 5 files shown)


nkms/client.py
```diff
@@ -55,13 +77,35 @@ def encrypt_key(self, key, pubkey=None, path=None, algorithm=None):
...
 
-    def decrypt_key(self, key, pubkey=None, path=None, owner=None):
+        if len(path) > 1 and not type(path) is str:
+            enc_keys = []
+            for subpath in path:
+                path_pubkey = self._derive_path_key(subpath)
+                enc_keys.append(self.encrypt_key(key, pubkey=path_pubkey))
+            return enc_keys
+        elif type(path) is str:
+            return self.encrypt_key(key, pubkey=self._derive_path_key(path))
```
michwill: Ideally, I'd like to split the path into subpaths here. Or, better, call a function which does it

tuxxy: So, as we discussed earlier, you'd like it so that we pass `path` as a string and then we split it into the paths ourselves?

In pseudocode:

```
path = '/foo/bar'
subpaths = path_split(path)
```
Where `subpaths` is a tuple or a list that has each path?

nkms/client.py
```diff
@@ -80,7 +81,7 @@ def encrypt_key(self, key, pubkey=None, path=None, algorithm=None):
         if not pubkey:
             pubkey = self._pub_key
 
-        if len(path) > 1 and not type(path) is str:
+        if type(path) is tuple and len(path) > 1:
```
tuxxy: Instead of checking length first, I check if it's a tuple and then check the length.
This avoided an issue where a `NoneType` `path` was being passed.

However, this can all change with @michwill's propose dchanges where `path` will simply be a string.

nkms/client.py
michwill: Yeah, I wanted this to be a "high-level" function where splitting into subpaths happens inside

nkms/client.py
michwill: I think you're right swapping them, yes

nkms/client.py
tuxxy: Understood, will do.

nkms/client.py
```diff
@@ -55,13 +78,35 @@ def encrypt_key(self, key, pubkey=None, path=None, algorithm=None):
         :return: Encrypted key(s)
         :rtype: bytes
         """
-        pass
+        if not pubkey:
+            pubkey = self._pub_key
 
-    def decrypt_key(self, key, pubkey=None, path=None, owner=None):
+        if type(path) is tuple and len(path) > 1:
+            enc_keys = []
```
jMyles: Does this make more sense as a `set` to enforce uniqueness?

nkms/client.py
```diff
@@ -27,6 +30,26 @@ def __init__(self, conf=None):
...
+        # TODO: Check for existing keypair before generation
+        # TODO: Save newly generated keypair
+        self._priv_key = self._pre.gen_priv(dtype='bytes')
+        self._pub_key = self._pre.priv2pub(self._priv_key)
+
+    def _derive_path_key(self, path, is_pub=True):
+        """
+        Derives a public key for the specific path.
+
+        :param str path: Path to generate key for.
```
jMyles: Does it make more sense for path to be `bytes`?  Some users might have path names that are not utf-8.

michwill: I think it does indeed!

nkms/client.py
michwill: Probably not - the order is important

michwill: And uniqueness here is more like "if I used the same path, it means it's the same file"

nkms/crypto/kmac.py
```diff
@@ -0,0 +1,34 @@
+import hashlib
+
+
+class KMAC_256(object):
+    # TODO If performance is needed, this could be optimized a bit...
+    # TODO Would be preferable to follow NIST and use cSHAKE.
+    def __init__(self):
+        self.LENGTH_BYTES = 32
```
jMyles: Why are these instance attrs rather than class attrs?

tests/test_client.py
```diff
@@ -0,0 +1,94 @@
...
+
+class TestClient(unittest.TestCase):
+    def setUp(self):
+        self.pre = pre_from_algorithm(default_algorithm)
+        self.symm = symmetric_from_algorithm(default_algorithm)
+        self.priv_key = self.pre.gen_priv(dtype='bytes')
+        self.pub_key = self.pre.priv2pub(self.priv_key)
+
+        self.client = Client()
+
```
jMyles: Wonderfully named test methods; great job @tuxxy.

nkms/client.py
jMyles: Right - of course.  

Another interesting option - does it make sense for it to be a dict, where the keys are paths?

michwill: Well, that is a very good point. Not necessarily we want to show what the path is to everyone, but maybe if it's an (ordered) dict {hash(subpath) -> key}, that could be valuable!

jMyles: Is there a security implication to have the path in the clear as a key?  They'll have access to the path anyway at run time by dint of this function.

jMyles: ...and it might be useful to be able to easily get a iterator of paths with enc_keys.keys().

tuxxy: The security professional in me wants to think that this isn't a very good idea. I'll look into better structuring this, but in a secure manner.

nkms/client.py
tuxxy: Very good point. Thanks!

nkms/crypto/kmac.py
tuxxy: Just moving quickly, but I had the idea of ripping this code out and making another lib for it since I couldn't find anything online for a KMAC implementation in Python.

The idea is that you would set the digest `LENGTH_BYTES` arbitrarily based on how you need to use it and `BLOCK_SIZE_BYTES` would be set based on the length you choose for the most secure option.

tuxxy: Though, in this case, you are right. Those should be class attributes. 👍

nkms/client.py
jMyles: Can you help me understand the security implications here?  If it's a problem to use the path as a dict key, then it's a problem to use the path at all, isn't it?  IE, why do we even have the path crossing the wire in the clear in the first place?

tuxxy:
It's not the part crossing the wire that's an issue, this hashing would be
how the recipient figures out which key is which in the encrypted header.

The problem there is that a simple hash of the file path isn't very random
and becomes very easy to bruteforce.

We're not worried about the KMAC because it combines the path with the
(random) key material which has much more entropy and becomes harder to
bruteforce.

jMyles:
But the path, in clear bytes, is available in the context of this function.  How does using it as a key in a dict change the security condition?

michwill:
The cleartext path isn't passed to the remote function. Instead, it's hashes of pubkey + subpath essentially, which are used as a 'key' (not encryption key) in key-value store on the server. So, no clear text, but not hard to brute force

I actually have some idea how to fix that if needed (not always we want it)

jMyles:
But here in `Client`, which people are welcome to subclass for their purposes, the path *is* available in the clear.  So the security rubicon is already crossed, is it not?

Therefore, we may as well use it as a key for a dict, providing easy access to an iterable which maps it to its encryption key.

It's completely possible that there's a security implication I'm not consider here - if that's the case, can one of you clarify it?

michwill:
The path is available in clear on the client side, no problem here I guess.

However, the dict (for which it is the key) is server-side. So, we somewhat obfuscate it for the server. I think it is important to leave it possible to have different ways of hash-from-path derivation: we can make it so that it's private I think (I have some idea for that).

As for the map - in one case, you show the hash(path) to the reencryption nodes only, on the other - you also show these hashes to the storage layer.

michwill:
The most trivial way is to have some sort of "path-secret key" for deriving the path hash, available to all the clients. In the worst key, if it leaks, someone else can brute force paths (not ideal but still ok).

But if we want something really secure there, that would require some attribute-based proxy re-encryption.. Need to talk to scientists, but perhaps doesn't matter much for now :-)

PR #21, opened 2017-09-02
[WIP] Seed-only DHT server implementation. 

jMyles:


entry_points/run_local_server.py
```diff
@@ -0,0 +1,17 @@
+from nkms.network.server import NuCypherDHTServer
+import asyncio
+
+
+if __name__ == "__main__":
+    loop = asyncio.get_event_loop()
+    loop.set_debug(True)
+    loop.set_debug(True)
```
michwill: loop.set_debug() twice?

PR #34, opened 2017-09-12
Nodes remember each other's capabilities

jMyles:
Nodes need to be able to remember which Nodes have which capabilities.  

If a Node is "seed-only" (Sam?) , then other Nodes need to remember not to try to send payloads to it.


nkms/network/protocols.py
```diff
@@ -1,13 +1,43 @@
...
+
+        self.log.info("never seen %s before, adding to router and setting nearby " % node)
+        for key, value in self.storage.items():
+            keynode = Node(digest(key))
+            neighbors = self.router.findNeighbors(keynode)
+            if len(neighbors) > 0:
+                newNodeClose = node.distanceTo(keynode) < neighbors[-1].distanceTo(keynode)
+                thisNodeClosest = self.sourceNode.distanceTo(keynode) < neighbors[0].distanceTo(keynode)
+            if len(neighbors) == 0 or (newNodeClose and thisNodeClosest):
+                asyncio.ensure_future(self.callStore(node, key, value))
```
michwill: Is it where we'll insert "symlinks"?

nkms/network/protocols.py
```diff
@@ -1,13 +1,43 @@
...
+        source = Node(nodeid, sender[0], sender[1])
+        self.welcomeIfNewNode(source, node_properties)
+        return self.sourceNode.id
+
+    def welcomeIfNewNode(self, node, node_properties):
+        if not self.router.isNewNode(node):
+            return
+
+        self.log.info("never seen %s before, adding to router and setting nearby " % node)
+        for key, value in self.storage.items():
```
michwill: I didn't realize we have to iterate over all storage. Although, perhaps, it is possible to do some sort of index here (e.g. each node stores rekeys ordered by distanceTo(keynode) because it's always the same for this node, and iterate over saved elements which have the largest distance (because they're candidates to go away)

PR #35, opened 2017-09-12
Handle high level encryption/decryption

tuxxy:
## What this does
1. Some minor bug fixes
2. Adds methods for creating headers to attach to files for storage.
2a. Defines what a basic header should/will look like.
3. Adds high level Encryption and Decryption methods (`encrypt` and `decrypt`) that should "just work".
3a. This does not contain tests yet. Tests will use the `tempfile` module.
4. Adds a high level `open` method that opens and decrypts an encrypted file.
4a. Test not built yet, will also use the `tempfile` module.
5. Adds a minimal version numbering system for the prealpha API.

Tests will be added in a future PR/commit in `tests/test_client_crypto.py`.

### DIFF
--- a/nkms/client.py
+++ b/nkms/client.py
@@ -244,27 +244,10 @@ class Client(object):
         file_path = fd or path
         try:
             with open(file_path, mode=mode) as f:
-                enc_file = BytesIO(f.read())
+                enc_data = f.read()
         except Exception as E:
             raise E
-
-        header_length_bytes = enc_file.read(4)
-        header_length = int.from_bytes(header_length_bytes, byteorder='big')
-
-        header = enc_file.read(header_length)
-        ciphertext = enc_file.read()
-
-        version, enc_keys = self._read_header(header)
-        if version < 1000:
-            valid_key = None
-            for enc_key in enc_keys:
-                dec_key = self.decrypt_key(enc_key, path=path)
-                if dec_key != b'':
-                    valid_key = dec_key
-                    break
-
-            plaintext = self.decrypt_bulk(ciphertext, valid_key)
-        return plaintext
+        return self.decrypt(enc_data, path=path)
 
     def remove(self, pubkey=None, path=None):
         """
@@ -294,7 +277,7 @@ class Client(object):
 
         # Derive keys and encrypt them
         # TODO: https://github.com/nucypher/nucypher-kms/issues/33
-        if path not None:
+        if path is not None:
             enc_keys = self.encrypt_key(data_key, path=path)
         else:
             enc_keys = [self.encrypt_key(data_key, path=path)]
@@ -321,5 +304,20 @@ class Client(object):
         :return: Unencrypted data
         :rtype: bytes
         """
-        # Not needed if open() is there?
-        pass
+        enc_file = BytesIO(edata)
+
+        header_length = int.from_bytes(enc_file.read(4), byteorder='big')
+        header = enc_file.read(header_length)
+        version, enc_keys = self._read_header(header)
+
+        ciphertext = msgpack.loads(enc_file.read())
+
+        if version < 1000:
+            valid_key = None
+            for enc_key in enc_keys:
+                dec_key = self.decrypt_key(enc_key, path=path)
+                if dec_key != b'':
+                    valid_key = dec_key
+                    break
+            plaintext = self.decrypt_bulk(ciphertext, valid_key)
+        return plaintext
--- a/nkms/client.py
+++ b/nkms/client.py
@@ -77,14 +77,15 @@ class Client(object):
         :param enc_keys: List of encrypted keys in bytes
         :param version: Version number of Cryptographic API (default: 0.1.0.0)
 
-        :return: Complete header for the encrypted file msgpack encoded.
-        :rtype: bytes
+        :return: Complete header msgpack encoded and length of raw header
+        :rtype: Tuple of the header and the header length e.g: (<header>, 1200)
         """
         if version < 1000:
             vers_bytes = version.to_bytes(4, byteorder='big')
             num_keys_bytes = len(enc_keys).to_bytes(4, byteorder='big')
             keys = b''.join(enc_keys)
-            return msgpack.dumps(vers_bytes + num_keys_bytes + keys)
+            header = vers_bytes + num_keys_bytes + keys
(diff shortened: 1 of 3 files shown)


nkms/client.py
```diff
@@ -67,6 +71,43 @@ def _split_path(self, path):
...
+
+        :param enc_keys: List of encrypted keys in bytes
+        :param version: Version number of Cryptographic API (default: 0.1.0.0)
+
+        :return: Complete header msgpack encoded and length of raw header
+        :rtype: Tuple of the header and the header length e.g: (<header>, 1200)
+        """
+        if version < 1000:
+            vers_bytes = version.to_bytes(4, byteorder='big')
+            num_keys_bytes = len(enc_keys).to_bytes(4, byteorder='big')
```
michwill: Are you sure we will ever have 4 bytes of keys for the path? I'd say, it will probably be always less than 255 subpaths

nkms/client.py
```diff
@@ -67,6 +71,43 @@ def _split_path(self, path):
...
+            keys = b''.join(enc_keys)
+            header = msgpack.dumps(vers_bytes + num_keys_bytes + keys)
```
michwill: I think, the point of msgpack is to avoid these sorts of things. So, you can just say msgpack([version] + enc_keys), and it will care about everything.
Your way works also, but then there is no need to use msgpack

nkms/client.py
```diff
@@ -67,6 +71,43 @@ def _split_path(self, path):
...
+    def _read_header(self, header):
+        """
+        Reads a NuCypher header.
+
+        :param header: Msgpack encoded header to read
+
+        :return: Version number, and list of encrypted keys
+        :rtype: Tuple of an int and a list e.g: (100, [...])
+        """
+        header = BytesIO(msgpack.loads(header))
```
michwill: And if you use msgpack as intended, you'll have something like:
```python
header = msgpack.loads(header)
version = header[0]
enc_keys = header[1:]
```

nkms/client.py
```diff
@@ -200,7 +241,13 @@ def open(self, pubkey=None, path=None, mode='r', fd=None, algorithm=None):
 
         If pubkey is not set, we're working on our own files.
         """
-        pass
+        file_path = fd or path
+        try:
+            with open(file_path, mode=mode) as f:
+                enc_data = f.read()
```
michwill: With open, it's better to decrypt file as a stream (gradually), so that we're working with the resulting file descriptor as if we opened an unencrypted file. E.g.:
```python
f = client.open(pub, '/hello/world.txt')
x = f.read(10)
...
f.close
```
or even:
```python
with client.open(...) as f:
    do_something_with(f)
```

tuxxy: So, should I implement a file-like interface for this method and not have it return the unencrypted data?

My first thought was to have it just decrypt the entire file at once for testing, then when we start focusing more on the cryptographic API to place the NaCl nonce in the NuCypher header so we can decrypt it as a stream.

michwill: Yes, exactly, I wanted the file-like interface. The file could potentially be bulky. I think, nacl can decrypt streams, right?

nkms/client.py
tuxxy: I thought it would just be more standard to use a four byte integer.

We can make it a short int with two bytes. I just think that adding a second byte to allow for more than 255 subpaths seems better so that someone could do that if they wanted to. It's only one byte more and it doesn't have such a small ceiling.

michwill: Why don't we just msgpack it and let the msgpack decide?

nkms/client.py
tuxxy: NaCl can, I think. You have to pass in the `nonce` param when performing decryption otherwise I think it tries to pull it out of the ciphertext.

michwill: So, why don't we put nonce at the front (or with the header)?

tuxxy: That's what my plan was after I get this interface working and tested.

michwill: I even think: why not even allow to open the file for writing (when it's for our own pubkey)? In this case, we don't have to keep all the ciphertext in memory

nkms/client.py
tuxxy: Sure, I just figured out how msg pack works like this. :)

nkms/client.py
michwill: I think, we might leave both options here as different methods (file-like interface open() and full decrypted data decrypt()). The full decrypted data will most likely be when the path is not the path to the real file but just, say, "path" to the chat room

tuxxy: Why would we need to open a file that we're reading for decryption in a write mode?

michwill: For write mode: it's if either we create a new one, or we want to add new data to it (as the owner)

tuxxy: Ah, okay.

However, if we're adding new data to it, we need to decrypt it and reencrypt it again with a different random nonce.

michwill: When you're not over-writing but appending data, cannot you use the same nonce for the whole thing in stream ciphers? As if you continued adding blocks to the same file

tuxxy: No, it causes the keystream to repeat which can allow someone to break the cipher.

michwill: I mean, if you add data to an encrypted file - wouldn't it be the same as if you encrypted a larger file to start with (with that nonce)?

tuxxy: Yes, but any mistake made when writing the data could allow the cipher to be broken.

You're essentially adding a *huge* attack surface on the schema and that's always a horrible idea. It would be better to reuse the key, generate a new nonce and reencrypt the file again to avoid any major detrimental pit falls.

michwill: (of course, only if you continue the keystream, not restart it!)

michwill: Anyway. I guess, let's make it so that you encrypt it all at once at first.

I think ideally we'd like to have it so that each update has its own symmetric key (encrypted by a public key) so that updates to the file will be unreadable by Bob after we revoked the access, even if he saved the symmetric key. And that requires us to not update the old ciphertext.

tuxxy:
I'm requesting that this PR be merged. I'm working on the changes recommended above, but they are being implemented in a much different interface. The commits and changes in the new version don't really fit within this PR, but still works with the code referenced in this one.

PR #37, opened 2017-09-14
Add EncryptedFile object for local encryption/decryption functions

tuxxy:
## What this does
1. Adds an `EncryptedFile` object for high level encrypt/decrypt functions.

## What's going to happen next?
1. Refactor of the `EncryptedFile` object. This will include a new `Header` object that allows users to attack headers to other storage objects if they choose.
2. Refactor of the `Client` object to allow for the use of the `EncryptedFile` and `Header` objects. This will also make `Client` code a bit cleaner.

## What I'd like:
1. Please comment on changes that you would like to see during the refactor. Could something be done differently and more efficiently? Can anything be done more cleanly? :)


nkms/storage/encrypted_file.py
```diff
@@ -0,0 +1,200 @@
...
+        self.header.update(header)
+        try:
+            encoded_header = msgpack.dumps(self.header)
+        except ValueError as e:
+            raise e
+        self.header_length = len(encoded_header)
+        return (self.header_length, encoded_header)
+
+    def _read_header(self):
+        """
```
jMyles: Maybe it makes sense to turn 4 into a named constant, `HEADER_LENGTH`, for readability?

nkms/storage/encrypted_file.py
```diff
@@ -0,0 +1,200 @@
...
+        :param int chunk_size: Size of chunk to read from self.file_obj
+        :param bytes nonce: Nonce to use during decryption
+
+        :return: Decrypted/Authenticated chunk
+        :rtype: Bytes
+        """
+        ciphertext = self.file_obj.read(chunk_size)
+        return self.cipher.decrypt(ciphertext, nonce=nonce)
+
+    def open_new(self, keys, chunk_size=1000000, nonce=None):
```
jMyles: Why don't we init a completely new `EncryptedFile` here?

nkms/storage/encrypted_file.py
```diff
@@ -0,0 +1,200 @@
...
+            chunk_num += 1
+        self._update_header({'num_chunks': chunk_num})
+
+    def close(self):
+        """
+        Writes the header to the file_obj and closes it. Called after the user
+        is finished writing data to the file_obj.
+        """
+        header, header_length = self._encode_header()
+        self.file_obj.write(header + header_length)
```
jMyles: Did you mean to actually close `self.file_obj` here?

nkms/storage/encrypted_file.py
```diff
@@ -0,0 +1,200 @@
...
+        """
+        Reads num_chunks of encrypted ciphertext and decrypt/authenticate it.
+
+        :param int num_chunks: Number of chunks to read. Default is all chunks
+
+        :return: List of decrypted/authenticated ciphertext chunks
+        :rtype: List
+        """
+        if num_chunks == 0:
+            num_chunks = self.header['chunks']
```
jMyles: Meaning, read *all the chunks*?

nkms/storage/encrypted_file.py
```diff
@@ -0,0 +1,200 @@
...
+
+        chunks = []
+        for chunk_num in range(num_chunks):
+            nonce = (self.header['nonce']
```
jMyles: I think that the next two lines can reasonably be turned into a method of their own.

PR #42, opened 2017-09-14
[WIP] Add Header object and refactors EncryptedFile and Client

tuxxy:
## What this does:
1. Work in progress.

## What I'd like:
1. Please comment on this code. Mind that it isn't complete yet.


tuxxy:
@jMyles It's pretty common to return the number of written chunks after writing like this. The user doesn't actually have to receive it or use it in any way. It also helps us validate for testing.

The `header` is used as the actual header dictionary that contains the metadata about the `EncryptedFile`.

The `header_obj` is the actual `Header` object. I need to access that to write to it. I'm not offering a `update_header` method in `EncryptedFile` because it can cause corruption and make the file unreadable. Instead, I'm relying on the fact that the User will properly form their header before creating the object and pass it in with the initializer.

I'm basically making it difficult for the user to even access the header once it has been created so they don't corrupt it.

jMyles:
Sounds good!

tuxxy:
See #52 for the discussion.

PR #43, opened 2017-09-18
[WIP] Add KeyRing object with sign/verify/encrypt/decrypt functionalities

tuxxy:
### What this does:
1. Adds two objects -- `KeyRing` and `SigningKeypair`.
1a. Users should use the `KeyRing` class to generate the keys, not `SigningKeypair` directly.
2. Implements `sign` and `verify` functions to `KeyRing` and `SigningKeypair`.
2a. Users should use `KeyRing.sign` and `KeyRing.verify` instead of the `SigningKeypair` methods.

### What remains left:
1. Adding an `EncryptionKeypair` object for encryption.

## Notice:
1. This verification doesn't use a constant time comparison, there may be security implications here. This will be worked on in the future.
2. Key generation happens such that a random number is chosen that satisfies `priv = random([1, N - 1])` where N is a component of the elliptic curve used (`secp256k1`). This is correct as far as I know, however, I'll look at how other people are generating keys.
2a. We use the `random.SystemRandom` object to generate the private key. This uses `/dev/urandom` which is considered a secure source of entropy.


nkms/crypto/keyring/keyring.py
```diff
@@ -0,0 +1,63 @@
+import sha3
+from nkms.crypto.keyring.keys import SigningKeypair, EncryptingKeypair
+
+
+class KeyRing(object):
+    def __init__(self, sig_privkey=None, enc_privkey=None):
+        """
+        Initializes a KeyRing object. Uses the private keys to initialize
```
jMyles: Groovy.  I like these kinds of interfaces.

nkms/crypto/keyring/keyring.py
```diff
@@ -0,0 +1,63 @@
...
+        return self.sig_keypair.verify(msg_digest, signature)
+
+    def encrypt(self, plaintext):
+        """
+        Encrypts the plaintext provided.
+
+        :param bytes plaintext: Plaintext to encrypt w/ EncryptingKeypair
+
+        :rtype: bytes
+        :return: Ciphertext of plaintext
```
jMyles: What type is this?  bytes right?  Maybe we do a type hint for the return?

nkms/crypto/keyring/keys/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .signing_keypair import SigningKeypair
```
jMyles: I'm not a huge fan of this pattern.  Why not put both of these classes in `nkms/crypto/keypairs.py`?  Cleaner namespacing; no awkward import in `__init__` - don't get me wrong, plenty of packages do this, but I think it's just confusing.

tests/crypto/test_keyring.py
```diff
@@ -0,0 +1,34 @@
...
+    def setUp(self):
+        self.keyring = KeyRing()
+        self.msg = b'this is a test'
+
+    def test_signing(self):
+        signature = self.keyring.sign(self.msg)
+
+        try:
+            sig = msgpack.loads(signature)
+            self.assertTrue(1, len(sig[0]))     # Check v
```
jMyles: For my part, I'd be happy to have some more verbosity here.  What is this test *actually* doing?  As a reader, I don't have any reason to understand *why* 1, 32, and 32 are appropriate lengths for these three elements of `sig`.

tests/crypto/test_signing_keypair.py
```diff
@@ -0,0 +1,37 @@
...
+        msg_digest = sha3.keccak_256(self.msg).digest()
+        signature = self.keypair.sign(msg_digest)
+
+        try:
+            sig = msgpack.loads(signature)
+            self.assertTrue(1, len(sig[0]))     # Check v
+            self.assertTrue(32, len(sig[1]))    # Check r
+            self.assertTrue(32, len(sig[2]))    # Check s
+        except Exception as e:
+            self.fail("Failed to msgpack.load signature: {}".format(e))
```
jMyles: This strikes me as odd - what Exception do we expect here and why?

tuxxy: We don't expect any `Exception`. We encapsulate it in a try to check if the `msgpack.loads` fails for any reason. If it does, we can capture the Exception it throws, whatever it may be, and fail the test.

jMyles: Unless there some special reason to do this in the context of crypto, best practice here is to just let the exception raise.

nkms/crypto/keyring/keys/signing_keypair.py
```diff
@@ -0,0 +1,58 @@
+import msgpack
+from random import SystemRandom
+from py_ecc.secp256k1 import N, privtopub, ecdsa_raw_sign, ecdsa_raw_recover
+
+
+class SigningKeypair(object):
+    def __init__(self, privkey_bytes=None):
+        self.secure_rand = SystemRandom()
```
michwill: Oh wow, didn't know that random provides this also :-)

nkms/crypto/keyring/keys/signing_keypair.py
```diff
@@ -0,0 +1,58 @@
...
+            priv_number = self.secure_rand.randrange(1, N)
+            self.priv_key = priv_number.to_bytes(32, byteorder='big')
+        # Get the public component
+        self.pub_key = privtopub(self.priv_key)
+
+    def _vrs_msgpack_dump(self, v, r, s):
+        v_bytes = v.to_bytes(1, byteorder='big')
+        r_bytes = r.to_bytes(32, byteorder='big')
+        s_bytes = s.to_bytes(32, byteorder='big')
+        return msgpack.dumps((v_bytes, r_bytes, s_bytes))
```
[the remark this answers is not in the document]
michwill: Or, I guess, let's see how serialized signatures look like in Ethereum?

tests/crypto/test_keyring.py
michwill: Btw that's a good point. Curves different from secp256k1 could have size different from 32 bytes

nkms/crypto/keyring/keys/signing_keypair.py
tuxxy: Yeah, `SystemRandom` is great. It saves a lot of headache from trying to implement similar functions with raw output from `os.urandom`.

nkms/crypto/keyring/keys/signing_keypair.py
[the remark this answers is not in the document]
tuxxy: Definitely. I couldn't find a quick reference to the signatures, so I just went with what I thought was best for the moment.

I'm using msgpack here because it answers the byte reading problem we had before. If I didn't do this, I'd have to do something bytestring substring stuff for it, and this just seemed much simpler.

In this case, I can pack it as a tuple and unpack it as a list keeping the elements in order. It "just works" haha.

I completely agree that we should see how Ethereum serializes signatures.

tests/crypto/test_keyring.py
tuxxy: Yep, I have this on my TODO for this code. I need to find a way to determine the size of the curve explicitly. I think I can do this with one of the components of the curve and some math.

nkms/crypto/keyring/keys/signing_keypair.py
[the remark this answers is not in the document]
michwill: Bytestring substring (I guess that's what you need for Ethereum-way) should be also simple though. E.g. for parsing:
```
assert (len(sig) - 1) == bitsize * 2
v_bytes = sig[0]
r_bytes = sig[1: len(sig) // 2 + 1]
s_bytes = sig[len(sig) // 2 + 1:]
```
Although, not as expressive as with msgpack of course

PR #44, opened 2017-09-19
Policy and PolicyGroup.  Some methods are faked with random data.

jMyles:


nkms/policy/models.py
```diff
@@ -0,0 +1,107 @@
...
+        self.ursula = networky_stuff.find_ursula(self.id, self.hashed_part)
+        return self.ursula.encrypt_for((self.kfrag, self.challenge_pack, self.treasure_map))
+
+    def generate_challenge_pack(self):
+        if self.kfrag == UNKNOWN_KFRAG:
+            raise TypeError("Can't generate a challenge pack unless we know the kfrag.  Are you Alice?")
+
+        # TODO: make this work instead of being random.
+        import random
+        self.challenge_pack = [(random.getrandbits(32), random.getrandbits(32)) for x in range(self.challenge_size)]
```
michwill: Btw it can be encrypt of really random data

nkms/policy/models.py
```diff
@@ -0,0 +1,107 @@
...
+class PolicyGroup(object):
+    """
+    The terms and conditions by which Alice shares with Bob.
+    """
+
+    def __init__(self, policies=None):
+        self.policies = policies or []
+
+    @staticmethod
+    def craft(keychain_alice: "KeyChain",
```
michwill: Two questions here:

* Why `craft` and not `make`, for example? (or some commonly used word which you can guess without introspecting)
* Why staticmethod rather than, say, classmethod?

jMyles: I did some thinking about this class last night that I think obsoletes both of these questions: I think what we actually want here is a `PolicyManager` class to generate and manager `PolicyGroup` and `Policy` instances (ie, for "table-level" logic).

But to answer your questions:

1) I liked `craft` because this method has to be called by Alice, who is the one who originally crafts the policy.  
2) Yeah, a `@classmethod` is probably better long-term here, but I sometimes use `@staticmethod` at first before I'm 100% sure that the method will remain part of the class.

nkms/policy/models.py
```diff
@@ -0,0 +1,107 @@
...
+        policy = Policy(kfrag)
+        policy.generate_challenge_pack()
+        policy.hash(pubkey_sig_alice=pubkey_sig_alice, hash_input=(pubkey_enc_bob, uri, kfrag_id))
+
+        return policy
+
+    def hash(self, pubkey_sig_alice, hash_input):
+        hash_input = str(hash_input).encode()
+        self.hashed_part = content_hash(hash_input)
+        hash_input_for_id = str(pubkey_sig_alice).encode() + str(self.hashed_part).encode()
```
tuxxy: Instead of encoding this, you can hash it and, instead of returning the raw digest, you can call `hexdigest` to get the stringified version of it in hexidecimal.

tuxxy:
```>>> sha3.keccak_256(b'this is a test').hexdigest()
'9fd09c38c2a5ae0a0bcd617872b735e37909ccc05c956460be7d3d03d881a0dc'
```

jMyles: Oh yeah, I agree that this is a bit of a shitshow.  :-)

But I'm not clear on what you're suggesting here.  Your sample shows a bytes object being passed to `keccak_256()`, but how to proceed for an arbitrary object?  For example, in this case, `hash_input` will often be a tuple.  And then, on line 84, how to properly do this concatenation?

tuxxy: The `keccak_256` function will always need a bytes object to hash it, so I recommend taking each element of `hash_input`, encoding it, then hash it to get a bytes object for that element. Do this for each element, then hash them all together and get a `hexdigest` version of it to avoid weird nuances in the string when converting them.

For example:
```
keccak(keccak(pubkey_enc_bob).digest() + keccak(uri).digest() + keccak(kfrag_id).digest).hexdigest()
```

tuxxy: This allows two things:
1) You avoid the string conversion which can add weird elements to the hashed string.
2) Allows the hashed results to be more replicatable.

tuxxy: Personally, I would have the hashing functions accept a list of things to hash. Check if the type is of `bytes`. If it is not, encode it. Then update the hash via `update(..)`.

I would also include a boolean param like `hex_format` which will cause the hash output to be `hexdigest()` instead of `digest()`.

michwill: Btw should the hash_input_for_id be necessarily string (hexdigest) not bytes (digest)?

PR #54, opened 2017-09-25
Travis and Pipenv setup.

jMyles:


setup.py
```diff
@@ -23,7 +23,7 @@
 LINKS = [
         'https://github.com/nucypher/kademlia/archive/kms-dependency.tar.gz#egg=kademlia-1.0',
```
michwill: Should we do the same (`kms-dependency` branch) for kademlia and rpcudp?

tests/conftest.py
```diff
@@ -1,16 +0,0 @@
-import nkms.db
```
michwill: Didn't quite get about the removal of this file. It sets up the debug db which we're using to store kFrags (if not doing that, it'll reuse the per-user db which is bad).

PR #62, opened 2017-09-28
Character classes pursuant to our call today.

jMyles:


jMyles:
I don't immediately see a need to split them into multiple files (and namespaces), but if you think we need to, I don't really see a huge problem with it.  I do like `from nkms.characters import <CharacterName>` more than `from nkms.characters.<character_name> import <CharacterName>`.

nkms/characters.py
```diff
@@ -0,0 +1,37 @@
...
+    A base-class for any character in our cryptography protocol narrative.
+    """
+    _server = None
+    _server_class = Server
+
+    def __init__(self, attach_server=True):
+        if attach_server:
+            self.attach_server()
+
+    def attach_server(self, ksize: int = 20, alpha: int = 3, id: None = None, storage: None = None,
```
KPrasch: Since you've got defaults set, I'd say the type hints are a bit wet.

jMyles: Yeah so what's the skinny here? Never type hints with defaults unless you expect additional types?

jMyles: I thought it was interesting that the type-hinted default variant wasn't demonstrated in the docs.  Seems like an oversight (or a less-than-thorough reader?)

KPrasch: I don't know that there is a best practice established on this.  Looking through [PEP 484](https://www.python.org/dev/peps/pep-0484/#arbitrary-argument-lists-and-default-argument-values)

jMyles: I read it as thoroughly as ever (which is to say, I heavily skimmed it) while writing this, and I didn't see anything that I was able to interpret as guidance.

tuxxy: I'm just gonna chime in here. I don't know the PEP standard for this, but using a plain default looks much cleaner (and more pythonic) than type hinting + defaults.

jMyles: Well I'm glad we can all agree that it's an affront to deceny. That the first step to full-on dissent.

I'll yank it out like it's 2.6.

KPrasch: Agreed, I'd omit. The typing system does not inference any types, so technically it is correct. from the docs: "A function parameter without an annotation is assumed to be annotated with Any."

PR #67, opened 2017-10-03
Allow iterables to be digested and signed

tuxxy:
### What this does:
1. Allows iterable objects to be passed instead of just expecting plain `bytes` objects.

2. Adds a `digest` method to `SigningKeypair` that accepts an `Iterable` as `data`.
    1. If `data` is of `bytes` type, then it just hashes like normal.
    2. If `data` is not `bytes` but is an `Iterable`, then each item within it is updated into the digest.

### DIFF
--- a/nkms/crypto/keypairs.py
+++ b/nkms/crypto/keypairs.py
@@ -1,7 +1,9 @@
 import msgpack
+import sha3
 from random import SystemRandom
 from py_ecc.secp256k1 import N, privtopub, ecdsa_raw_sign, ecdsa_raw_recover
 from npre import umbral
+from typing import Iterable
 
 
 class EncryptingKeypair(object):
@@ -124,6 +126,23 @@ class SigningKeypair(object):
         r = int.from_bytes(sig[1], byteorder='big')
         s = int.from_bytes(sig[2], byteorder='big')
         return (v, r, s)
+
+    def digest(self, data: Iterable) -> bytes:
+        """
+        Accepts an iterable containing bytes and digests it.
+
+        :param Iterable data: Data to digest
+
+        :rtype: bytes
+        :return: bytestring of digested data
+        """
+        if type(data) == bytes:
+            return sha3.keccak_256(data).digest()
+
+        hash = sha3.keccak_256()
+        for item in data:
+            hash.update(item)
+        return hash.digest()
 
     def sign(self, msghash):
         """
--- a/nkms/crypto/keypairs.py
+++ b/nkms/crypto/keypairs.py
@@ -127,21 +127,18 @@ class SigningKeypair(object):
         s = int.from_bytes(sig[2], byteorder='big')
         return (v, r, s)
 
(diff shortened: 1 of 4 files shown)


nkms/crypto/keypairs.py
```diff
@@ -125,6 +127,23 @@ def _vrs_msgpack_load(self, msgpack_vrs):
...
+        :param Iterable data: Data to digest
+
+        :rtype: bytes
+        :return: bytestring of digested data
+        """
+        if type(data) == bytes:
+            return sha3.keccak_256(data).digest()
+
+        hash = sha3.keccak_256()
+        for item in data:
```
jMyles: This line will do some things that are perhaps unexpected:

* For a non-iterable, it will raise "TypeError: '{whatever}'' object is not iterable"
* For a string - instead of casting to bytes, it will iterate through the string and update the hash.  Is that what we want?  Maybe it is.

nkms/crypto/keyring.py
```diff
@@ -109,12 +109,12 @@ def sign(self, message):
         """
         Signs a message and returns a signature with the keccak hash.
 
-        :param bytes message: Message to sign in bytes
+        :param Iterable message: Message to sign in an iterable of bytes
```
jMyles: Wait, can't it still be `bytes`?

nkms/crypto/keypairs.py
tuxxy: A `String` can't be used in the hash digest, it needs to be of type `bytes`. Most APIs hand this off to the user to handle their own encoding.

If it raises a `TypeError`, I would completely consider that expected behavior.

nkms/crypto/keyring.py
tuxxy: It can, but I just thought to make it clear that it can be an iterable of `bytes` objects.

nkms/crypto/keypairs.py
jMyles: Well, as I said, it won't with this implementation.  It will iterate through the string.

jMyles: Oh, right, and then I guess it will raise TypeError in the next line.

jMyles: So I guess that's fine.  But I think that the "TypeError: '{whatever}'' object is not iterable" is a little strange - better to have "TypeError: message must be bytes or an collection of bytes."

tuxxy: It has almost exactly that error message:

```
>>> sha3.keccak_256('hash a string').digest()
  File "<stdin>", line 1, in <module>
TypeError: Unicode-objects must be encoded before hashing
```

jMyles: Sure - string is fine.  But what if someone passes `int`?

tuxxy: Similar `TypeError`, but just informs the user that it requires an object that supports the buffer API.

PR #72, opened 2017-10-04
TreasureMap is stored on a 'free tier' k:v DHT and is available via deterministic ID.  Policy IDs are now random-ish.

jMyles:


nkms/crypto/keypairs.py
```diff
@@ -114,6 +114,9 @@ def __init__(self, privkey_bytes=None):
         # Get the public component
         self.pub_key = privtopub(self.priv_key)
 
+    def pubkey_bytes(self):
+        return b''.join(i.to_bytes(32, 'big') for i in self.pub_key)
+
```
tuxxy: Why are we doing this instead of using the `elliptic_curve.serialize` method?

michwill: Because it's a signing key which uses plain pythonbig integers?

tuxxy:
Ah, you're right. Lapse in judgement there. :P

PR #73, opened 2017-10-05
[WIP] CryptoPower, CryptoPowerUps, and implementation of the new Crypto API.

jMyles:


nkms/characters.py
```diff
@@ -9,10 +11,26 @@ class Character(object):
...
+        """raised when we try to interact with an actor of whom we haven't learned yet."""
+
+    def __init__(self, attach_server=True, crypto_power: CryptoPower=None):
         if attach_server:
             self.attach_server()
+        self._crypto_power = crypto_power or CryptoPower(power_ups=[SigningKeypair])
+
+        class Seal(object):
+            """
+            Can be used
```
jMyles: Hell of a docstring, if I may say so myself.

(obviously this was meant to be longer and clearer.  Next commit.)

nkms/crypto/powers.py
```diff
@@ -55,7 +55,8 @@ def sign(self, *messages):
             sig_keypair = self._power_ups[SigningKeypair]
         except KeyError:
             raise NoSigningPower
-        msg_digest = sig_keypair.digest(*messages)
+        # msg_digest = sig_keypair.digest(*messages)  #  This didn't work.  I guess we don't hash things before signing / verifying?
```
jMyles: @tuxxy, @michwill: So it looks like we *don't* expect messages to have been hashed before signing / verifying (contrary to the docstrings in `sign` and `verify`).

What's the best practice here?  Shall we actually hash the message first?  Or no?

nkms/crypto/powers.py
```diff
@@ -0,0 +1,122 @@
...
+        :param Iterable messages: Messages to sign in an iterable of bytes
+
+        :rtype: bytestring
+        :return: Signature of message
+        """
+        try:
+            sig_keypair = self._power_ups[SigningKeypair]
+        except KeyError:
+            raise NoSigningPower
+        msg_digest = b"".join(signature_hash(m) for m in messages)
```
michwill: I am surprised that sign() can actually handle it (when you join multiple hashes). Or can it?
Anyway, I think it should probably be:
```
signature_hash(b"".join(signature_hash(m) for m in messages))
```

nkms/crypto/keyring.py
```diff
@@ -1,13 +1,25 @@
...
 
 class KeyRing(object):
-    def __init__(self, sig_privkey=None, enc_privkey=None):
+    """
+    KeyRing is an accessory that a Character carries in order to give her high-level, easy-to-read access to certain specific powers, such as key generation, encryption, signing, and re-encryption.
+
+    A Character's KeyRing is the canonical representation of that Character's capacity to relate to other Characters using lower-level cryptographic functions.
+    """
+
+    class ActorNotFound(KeyError):
```
michwill: Why inside the class?

PR #74, opened 2017-10-05
[WIP] Implement a low-level, frozen ECIES PRE Cryptographic API w/ serialization

### LINKED ISSUE #29: Freeze the v1 Cryptographic API
The output/input of the v1 cryptographic API needs to be frozen to prevent breaking changes in the future.

tuxxy:
### What this does:
1. Adds an `api.py` file in `nkms.crypto`.
2. Implements `nacl.SecretBox` `encrypt` and `decrypt`.
    1. Key agnostic
3. Implements `npre.umbral.PRE` (ECIES) as standalone static methods.
    1. Performs serialization `bytes <--> elliptic_curve.Element`.
    2. Handles `encapsulate`, `decapsulate`, `rekey`, `split_rekey`, `combine`, `reencrypt`.
    3. Adds two utility methods -- `priv_bytes2ec` and `pub_bytes2ec`. This allows for easy serialization between `bytes` and `elliptic_curve.Element` objects.
    4. Completely key agnostic, a higher level module will implement the keys to prevent frequent changes.
    5. This can be frozen, which means no more changes to the API layer here. (Fixes #29)
4. Implements ECDSA signing and verification.
    1. Adds utility functions `ecdsa_pub2bytes` and `ecdsa_bytes2pub`.
        1. Converts the public key `Tuple(int, int)` to `bytes` via `int.to_bytes`.
    2. Adds ECDSA private key generation via `ecdsa_gen_priv`.
    3. Signatures are generated via `msgpack.dumps`.
5. Implements a secure random interface via `random.SystemRandom` as `SYSTEM_RAND`.
    1. `secure_random` -- Uses `getrandbits(bytes*8)` and converts to `bytes`.
        1. Avoids importing both `os` and uses just plain `SystemRandom`. (Would appreciate comments on this.)
    2. `secure_random_range` -- Uses `randrange` to generate a number between `min` and `max-1`.
6. Implements Keccak 256 hashing via `keccak_digest`. Accepts any number of parameters.
7. Adds tests to `tests.crypto.test_api`
8. Removes `nkms.crypto.hash` -- `nkms.crypto.api.keccak_digest` can be used instead.
    1. All uses of `hash.content_hash` or `hash.signature_hash` have been replaced with `api.keccak_digest`.


nkms/crypto/crypto.py
```diff
@@ -0,0 +1,219 @@
+from npre import umbral
+from npre import elliptic_curve
+from nacl.secret import SecretBox
+from typing import Tuple, Union, List
+
+
+class Crypto(object):
```
jMyles: So what made the decision to do a class instead of just a module?

PR #76, opened 2017-10-11
[WIP] Moving EncryptingKeypair over to powers.

jMyles:


nkms/crypto/powers.py
```diff
@@ -119,3 +123,103 @@ def sign(self, msghash):
...
     def public_key(self):
         return self.pub_key
+
+
+class EncryptingKeypair(CryptoPowerUp):
+    KEYSIZE = 32
+    confers_public_key = True
+
+    def __init__(self, privkey=None):
+        self.pre = umbral.PRE()
```
michwill: Just thought - how often do we create this? In principle, `pre` can be a class attribute

tuxxy: In `nkms.crypto.api` we already have it created. It can be used as `from nkms.crypto.api import PRE`.

tuxxy:
This has been merged into my branch https://github.com/tuxxy/nucypher-kms/tree/encrypting-power

PR #78, opened 2017-10-13
Character-level encryption, decryption, signing, and verification.  Alice, Bob, and Ursula classes working.

jMyles:
Notice that `Character.encrypt_for` and `Character.verify_from` are "cheatable" via kwargs - this will be removed once we have `EncryptingPower`.


jMyles:
I agree that `decrypt(verify=True)` makes sense, but there also occasions when we want to verify without decrypting, so it seems that we still need to keep at least a semblance of this method.

Also, I think that since it matches the flow of `encrypt_for` that users will find it useful - the tests read as nicely as they do in part because of lines like `bob.verify_from(alice, ...)`.

PR #80, opened 2017-10-17
Implement SigningPower

tuxxy:
## What this does:
1. Implements `SigningPower` in `nkms.crypto.powers`.
    1. Adds tests in `tests.keystore.test_keypairs` and `tests.crypto.test_signing_power`.
    2. Fixes tests that use `SigningKeypair` and replaces it with `SigningPower`.
    3. Fixes character code that use `SigningKeypair` and uses `SigningPower`.

2. Implements a `sign` and `verify` method on `SigningKeypair`.
    1. `verify` verifies with the keypair's own public key.

### DIFF
--- a/nkms/crypto/powers.py
+++ b/nkms/crypto/powers.py
@@ -101,13 +101,13 @@ class CryptoPowerUp(object):
 class SigningPower(CryptoPowerUp):
     confers_public_key = True
 
-    def __init__(self, keypair: keypairs.keypairs.SigningKeypair = None):
+    def __init__(self, keypair: keypairs.SigningKeypair = None):
         """
         Initializes a SigningPower object for CryptoPower.
         """
         self.keypair = keypair or keypairs.SigningKeypair()
         self.priv_key = self.keypair.privkey
-        self.pub_key = self.real_keypair.pubkey
+        self.pub_key = self.keypair.pubkey
 
     def sign(self, msghash):
         """
--- a/nkms/crypto/powers.py
+++ b/nkms/crypto/powers.py
@@ -101,12 +101,13 @@ class CryptoPowerUp(object):
 class SigningPower(CryptoPowerUp):
     confers_public_key = True
 
-    def __init__(self, keypair=None):  # TODO: Pretty much move this __init__ to SigningPower
-
-        self.real_keypair = keypair or RealSigingKeypair()  # Total throwaway line - this will not be "real_keypair" because this will be in SigningPower
-        self.real_keypair.gen_privkey()
-
-        self.priv_key, self.pub_key = self.real_keypair.privkey, self.real_keypair.pubkey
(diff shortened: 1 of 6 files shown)


nkms/crypto/powers.py
```diff
@@ -94,29 +98,26 @@ class CryptoPowerUp(object):
...
 
     def sign(self, msghash):
         """
-        TODO: Use crypto API sign()
-
-        Signs a hashed message and returns a msgpack'ed v, r, and s.
+        Signs a hashed message and returns a signature.
 
-        :param bytes msghash: Hash of the message
+        :param msghash: The hashed message to sign
```
jMyles: Nice fix.

nkms/keystore/keypairs.py
```diff
@@ -100,4 +99,30 @@ def gen_privkey(self, create_pubkey: bool = True):
...
+        Signs a hashed message and returns a signature.
+
+        :param msghash: The hashed message to sign
+
+        :return: Signature in bytes
+        """
+        v, r, s = API.ecdsa_sign(msghash, self.privkey)
+        return API.ecdsa_gen_sig(v, r, s)
+
+    def verify(self, msghash: bytes, signature: bytes) -> bool:
```
jMyles: Looks good.

tests/crypto/test_signing_power.py
```diff
@@ -0,0 +1,20 @@
...
+    def setUp(self):
+        self.keypair = keypairs.SigningKeypair()
+        self.power = SigningPower(self.keypair)
+
+    def test_signing(self):
+        msghash = API.keccak_digest(b'hello world!')
+
+        sig = self.power.sign(msghash)
+        self.assertEqual(bytes, type(sig))
+        self.assertEqual(65, len(sig))
```
jMyles: Why 65?  Does it make more sense to make this a constant?

tuxxy: `v` is a version byte (usually 1 in Ethereum, iirc). `r` and `s` are each 32 bytes.
Thus, 65 bytes.

Could be constant something like `SIGNATURE_SIZE = 65`

tuxxy: We have 32 hardcoded in many places, I didn't really think this would be much different.

jMyles: I think let's have nicename constants for 32, 64, and 65 for these three scenarios - it will make the tests much easier to skim.

michwill: I think, it makes sense to replace `32` or whatever we have in the code with constants also, but that's in a totally separate PR. This one is good to merge!

tuxxy: I agree, as well. We can just use `nkms.crypto.constants` and `nkms.keystore.constants`. But, yes, that would be for another PR.

jMyles: OK then, let's merge.

PR #83, opened 2017-10-19
Rekey api

tuxxy:
1. Adds an `ecies_ephemeral_split_rekey` function to `nkms.crypto.api`.
    1. Creates an `_internal.py` file to house internal crypto api functions.
        1. Adds a `_ecies_gen_ephemeral_key` function to the `_internal` module.
2. Adds tests for the above changes.

### DIFF
--- a/nkms/characters.py
+++ b/nkms/characters.py
@@ -165,9 +165,9 @@ class Alice(Character):
         # TODO: Make this actually work.
         pubkey_enc_bob = bob.seal  # ???  We need Bob's enc key, not sig.
         kfrags = [
-            'sfasdfsd9',
-            'dfasd09fi',
-            'sdfksd3f9',
+            b'sfasdfsd9',
+            b'dfasd09fi',
+            b'sdfksd3f9',
         ]
 
         return kfrags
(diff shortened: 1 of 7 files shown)


nkms/characters.py
```diff
@@ -158,19 +158,21 @@ def find_best_ursula(self):
...
-        # TODO: Make this actually work.
-        pubkey_enc_bob = bob.seal  # ???  We need Bob's enc key, not sig.
-        kfrags = [
-            'sfasdfsd9',
-            'dfasd09fi',
-            'sdfksd3f9',
-        ]
-
-        return kfrags
+    def generate_rekey_frags(self, alice_privkey, bob_pubkey, m, n):
```
jMyles: I like the proper name usage here - if we *know* that something is only the job of an alice or bob...

include their names in the blob?

(I'll see myself out).

nkms/crypto/_internal.py
```diff
@@ -0,0 +1,21 @@
+from nkms.crypto import api as API
+from typing import Tuple, Union
+from npre import elliptic_curve
+
+
+def _ecies_gen_ephemeral_key(
```
jMyles: Very sensible.  I like `_internal`.

PR #84, opened 2017-10-20
Implement lmdb in KeyStore

tuxxy:
### What this does:
1. Implements lmdb in `KeyStore`.
2. Fills out the `add_key`, `get_key`, and `del_key` methods.
    1. `add_key` returns the fingerprint (keccak 256) of the key that was added.
    2. `get_key` accepts a key fingerprint (as a hexdigest) and returns the keypair.
    3. `del_key` removes the key from lmdb.
3. Adds tests for lmdb as `test_lmdb_keystore` in `tests.keystore.test_keystore`.


nkms/keystore/keystore.py
```diff
@@ -1,19 +1,38 @@
...
+        Initalizes a KeyStore object.
 
-        TODO: Actually store keys.
-        TODO: Load keys from system.
+        :param lmdb_path: LMDB path to open for reading
         """
-        pass
+        self.lmdb_env = lmdb.open(lmdb_path)
+
+    def __del__(self):
```
tuxxy: @jMyles I would love your input here. I think this will do what I want to (when program ends, close the database), but not 100%.

jMyles: This is an interesting approach, but not the one you're looking for.  This method will only be called when this object is garbage-collected, which is not guaranteed in python.  Also, the time of execution of `__del__` varies across implementations.

Generally, for logic that *needs* to execute (finalizing I/O is usually first on lists of examples of this sort of thing), Python provides two patterns, each with a reasonably obvious use case:

* A context manager (will *always* be evaluated at some point after the context collapses)
* A `finally` clause (will *always* be evaluated after the cursor completes it traversal through the appropriate `try`, `except`, or `else` clauses of the same block).

In this case, it looks like you're already covered by the context object returned by `begin()` - reading the docs, it doesn't look like you're expected to call `close` explicitly - it's just a convenience method for breaking open iterators and such.  (somewhat disturbingly, [the docs](http://lmdb.readthedocs.io/en/release/#lmdb.Environment.close) link to a 404 that apparently used to describe this).

If you really want to be sure that `.close()` gets called, I think it's probably a good idea to run the entire program in a `try-finally` block.  While perhaps not perfectly pythonic, this pattern is also not unprecedented - for example, it's how Django closes connections to Postgres (or at least it used to be; I haven't looked at it in a wihle).

tuxxy: Thanks. I saw that `.begin()` handled it, but I threw that in here for this very reason. :)

Should I just remove that?

jMyles: Although I think that the answer is 'yes', I'm not comfortable giving a complete endorsement because, as I said above, the relevant part of the documentation is a link to a 404.

Also: I notice that the top of the README on github reads:

> I simply don't have time for this project right now, and still the issues keep piling in. Are you a heavy py-lmdb user and understand most bits of the API? Got some spare time to give a binding you use a little love? Dab hand at C and CFFI? Access to a Visual Studio build machine? Please drop me an e-mail: dw at botanicus dot net. TLC and hand-holding will be provided as necessary, I just have no bandwidth left to write new code.

I'm wondering: are we willing to possibly need to maintain this project?  Can't we just use sqlite3?

PR #87, opened 2017-10-23
Implement KeyNotFound Exception

tuxxy:
### What this does:
1. Adds a `KeyNotFound` Exception class.
    1. This gets raised in `KeyStore.get_key` when the fingerprint requested isn't found.

2. Implements `KeyNotFound` in the `tests.keystore.test_keystore` tests.

### DIFF
--- a/nkms/keystore/keystore.py
+++ b/nkms/keystore/keystore.py
@@ -5,6 +5,13 @@ from nkms.keystore import constants
 from typing import Union
 
 
+class KeyNotFound(KeyError):
+    """
+    Exception class for KeyStore get_key calls for keys that don't exist.
+    """
+    pass
+
+
 class KeyStore(object):
     """
     A storage class of cryptographic keys.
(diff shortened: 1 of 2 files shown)


nkms/keystore/keystore.py
```diff
@@ -78,7 +85,7 @@ def get_key(self, fingerprint: bytes) -> Union[keypairs.EncryptingKeypair,
             key = txn.get(fingerprint)
 
         if not key:
-            return None
+            raise KeyNotFound("No key found with the provided fingerprint.")
```
jMyles: A better message here is, "No key with fingerprint {fingerprint}" - don't you think?

tests/keystore/test_keystore.py
```diff
@@ -57,10 +57,10 @@ def test_lmdb_keystore(self):
 
         # Test del_key pubkey
         self.ks.del_key(fingerprint_pub)
-        key = self.ks.get_key(fingerprint_pub)
-        self.assertIsNone(key)
+        with self.assertRaises(keystore.KeyNotFound):
```
jMyles: beautiful.

nkms/keystore/keystore.py
tuxxy: Yeah, I'll throw that in.

Issue #99, opened 2017-11-06
More convenient verify method


jMyles:
Currently, we still call verify in multiple calls to API, ie:

```
msg_digest = API.keccak_digest(message)
sig = API.ecdsa_load_sig(signature)
API.ecdsa_verify(*sig, msg_digest, signature_pub_key)
```

Better is to have one function:

`API.verify(signature, message, public_key)`

PR #100, opened 2017-11-06
Verify method described in #99.

jMyles:


jMyles:
`_internal`?  I definitely don't think that this belongs there.  This is supposed to be a higher-level public function to replace the three-line idiom that we currently have for verification.  I thought we were conspiring around `utils`?  Is that what you meant?

tuxxy:
Until we figure that out more concretely, I'm more comfortable placing it in `_internal.py`. 

If we're going to add a `utils` file, let's talk about what all should be added to it and put it all in one PR.

jMyles:
Well, there's no way this belongs in `_internal`.  If you want to go that direction, it makes more sense to shift `ecdsa_verify` and `ecdsa_load_sig` to `_internal` and put this in `api`.

tuxxy:
I think of `_internal.py` as a staging area for the crypto API. Whether that be some addition to `api` or something that we want to add but don't have a place for yet, see the current function in there.

jMyles:
Subsumed by #101

PR #120, opened 2017-11-16
Alice finds Ursula, opens REST connection, and transmits Policy, the kFrag of which is properly saved by Ursula.

jMyles:


nkms/characters.py
```diff
@@ -254,6 +259,15 @@ class Ursula(Character):
...
     interface_ttl = 0
 
+    def __init__(self, urulsas_keystore=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if urulsas_keystore:
+            self.keystore = urulsas_keystore
+        else:
+            engine = create_engine('sqlite:///:memory:')
+            Base.metadata.create_all(engine)
+            self.keystore = keystore.KeyStore(engine)
```
michwill: Maybe some warning that we're not persistent here?

jMyles: OK, just pushed a commit that fixes this.

nkms/characters.py
```diff
@@ -254,9 +268,29 @@ class Ursula(Character):
     interface = None
     interface_ttl = 0
 
+    def __init__(self, urulsas_keystore=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        if urulsas_keystore:
+            self.keystore = urulsas_keystore
+        else:
```
tuxxy: For an Ursula, this might be very dangerous to start a non persistent storage system. Maybe we should not let her store kFrags if this is the case, otherwise she's doomed to lose them. Maybe fail instead of warn for Ursula?

nkms/characters.py
```diff
@@ -298,6 +338,24 @@ def publish_interface_information(self):
         loop = asyncio.get_event_loop()
         loop.run_until_complete(setter)
 
+    def set_policy(self, hrac_as_hex, request: http.Request):
```
tuxxy: Looks great, love this!

nkms/crypto/api.py
```diff
@@ -27,13 +28,13 @@ def secure_random(
...
-    return SYSTEM_RAND.getrandbits(num_bytes*8).to_bytes(num_bytes,
-                                                         byteorder='big')
+    return SYSTEM_RAND.getrandbits(num_bytes * 8).to_bytes(num_bytes,
+                                                           byteorder='big')
 
 
 def secure_random_range(
-    min: int,
-    max: int
+        min: int,
```
tuxxy: Mind explaining why this is more appropriate here? PyLint doesn't warn me about anything when it's indented as it was, is this not the case?

nkms/keystore/keypairs.py
```diff
@@ -198,3 +200,11 @@ def serialize_privkey(self) -> bytes:
                           constants.PRIV_KEY_BYTE +
                           self.privkey)
         return serialized_key
+
+
+class PublicKey(bytes):
```
tuxxy: Neat!

nkms/network/protocols.py
```diff
@@ -4,14 +4,15 @@
...
+from nkms.crypto.constants import HASH_DIGEST_LENGTH
 from nkms.crypto.signature import Signature
 from nkms.crypto.utils import BytestringSplitter
+from nkms.keystore.keypairs import PublicKey
 from nkms.network.constants import NODE_HAS_NO_STORAGE
 from nkms.network.node import NuCypherNode
 from nkms.network.routing import NuCypherRoutingTable
 
-dht_value_splitter = BytestringSplitter(Signature, (bytes, PUBKEY_SIG_LENGTH), (bytes, HASH_DIGEST_LENGTH))
+dht_value_splitter = BytestringSplitter(Signature, PublicKey, (bytes, HASH_DIGEST_LENGTH))
```
tuxxy: Yup, using PublicKey here is definitely better.

nkms/network/protocols.py
```diff
@@ -46,13 +47,7 @@ def rpc_ping(self, sender, nodeid, node_capabilities=[]):
...
         proper_key = digest(keccak_digest(bytes(sender_pubkey_sig) + bytes(hrac)))
 
-        # TODO: This try block is not the right approach - a Ciphertext class can resolve this instead.
-        try:
-            # Ursula uaddr scenario
-            verified = signature.verify(hrac, sender_pubkey_sig)
-        except Exception as e:
-            # trmap scenario
-            verified = signature.verify(msgpack.dumps(message), sender_pubkey_sig)
+        verified = signature.verify(hrac, sender_pubkey_sig)
```
tuxxy: Love these negative lines, much more readable! 💯

nkms/rest/app.py
```diff
@@ -0,0 +1,25 @@
...
+    Route('/kFrag/{kmac}', 'POST', create_kfrag),
+    Include('/docs', docs_urls),
+    Include('/static', static_urls)
+]
+
+app = App(routes=routes)
+
+
+if __name__ == '__main__':
+    app.main(('run',))
```
tuxxy: Still gotta wrap this with TLS, but I think we've figured that out.

nkms/characters.py
jMyles: I mostly agree.  For now, though, this makes it so much easier to test.  Maybe have an affirmative kwarg like, `use_memory_db` or something?

michwill: Or what if we make it impossible to use a memory db unless we replace it by mocking or in a conftest?

Idk if it's a good practice

nkms/rest/app.py
jMyles: Yeah, exactly - at least now we know *where* the transport gets wrapped, so that's a pretty big relief for me.

nkms/crypto/api.py
tuxxy: Hmm, strange. I actually am questioning whether what I'm doing is PEP8 then and whether PyLint will tell me about it.

There's nothing in there about the structure of:

```
def func(
    self,
    foo,
    bar
) -> bytes:
```

The reasoning in the PEP8 is because it's not distinguishable, but the definition is... more C-like?? because the colon is on a separate line entirely and the params are distinguishable. Can't find anything on that in the PEP8, though.

This is fine, just curious about that specific rule now.

PR #124, opened 2017-11-18
Metabytes for key generation, PublicKey class, and safety for public key de/serialization.

jMyles:
Public Keys are now generated with metabytes.

**This PR does not include metabytes for Private Keys - only public.**

A `PublicKey` class includes metadata for total length and metabyte length as well as a facility for stripping metabytes.

### DIFF
--- a/nkms/keystore/keypairs.py
+++ b/nkms/keystore/keypairs.py
@@ -56,6 +56,8 @@ class Keypair(object):
 
             elif key_type_byte == constants.PRIV_KEY_BYTE:
                 return SigningKeypair(privkey=key)
+        else:
+            raise ValueError("Unable to determine which type of keypair this is - keypair_byte was {}".format(keypair_byte))
 
 
 class EncryptingKeypair(Keypair):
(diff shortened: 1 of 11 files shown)


nkms/keystore/keypairs.py
```diff
@@ -200,3 +200,11 @@ def serialize_privkey(self) -> bytes:
...
                           self.privkey)
         return serialized_key
+
+
+class PublicKey(bytes):
+    _EXPECTED_LENGTH = 66
+    _METABYTES_LENGTH = 2
+
+    def without_metabytes(self):
+        return self[self._METABYTES_LENGTH::]
```
michwill: Doesn't matter, but I think `[self._METABYTES_LENGTH:]` will be sufficient

jMyles: Somehow, I got on a kick of using two, thinking it's more explicit, but perhaps it's just merely less terse.

michwill: Interesting. I'm totally fine with it though

PR #131, opened 2017-11-28
Changing actor ID to something more realistic.  This may become the b…

jMyles:
…asis for looking up an actor if we ever represent them with REST.

### DIFF
--- a/nkms/characters.py
+++ b/nkms/characters.py
@@ -1,5 +1,6 @@
 import asyncio
 import binascii
+from binascii import hexlify
 from logging import getLogger
 
 import msgpack
@@ -171,7 +172,7 @@ class Character(object):
             raise self.NotFound("We haven't learned of an actor with ID {}".format(actor.id()))
 
     def id(self):
-        return "whatever actor id ends up being - {}".format(id(self))
+        return hexlify(bytes(self.seal))
 
     def public_key(self, key_class):
         try:


tuxxy:
I'm wondering if this might be better represented and handled as a Keccak hex digest of the key instead, any thoughts @jMyles?

jMyles:
Yeah, I had that thought as well.  So, the difference of course is that in the hexlify scenario, the actor's public signing key can be derived from the URL.  In the digest scenario, it will be sent separately and verified against the URL.

So which is better?

I think that it's better to be able to get their public key from the URL.  

Is there a compelling argument against this?

tuxxy:
I'll add in here that anything in the URL is basically sent in the clear, so is there anything there that we may not want to be in the clear, in logs, etc?

It's just a public key, but logging is a thing.

jMyles:
Right, absolutely.  But my sense is that we *want* the public key to be in the clear. 

It's not unusual - and is in fact RESTful - for the object identifier in the URL to have semantic value.

tuxxy:
Since we're not focusing on anonymity quite yet, this is fine for now. Otherwise it opens it up to timing attacks and whatnot to identify users from logs.

Yeah, this is fine. Just wanted to make sure we're thinking it through here.

jMyles:
Using a hash of the public key won't change that, unless we include something unique in the hash per host (ie, the actor doing the learning - typically Ursula).

Obviously we can do that.  Is that preferable?

jMyles:
Also: I think that concerns about a timing attack are better handled by fuzzing than obfuscation.

jMyles:
But I'm totally flexible - if you think that a hash is better, let's just do that.

michwill:
Right now we want to identify at least Ursula. Just to check that she's staking enough.
In future, making it anonymous will require to do some tricks which make any kind of "salting" obsolete also, so probably no need

tuxxy:
No, this is good.

PR #133, opened 2017-12-01
REST relationship for Bob and Ursula, with complete PRE using all three Fragment classes.

jMyles:


nkms/policy/models.py
```diff
@@ -288,3 +288,36 @@ def __eq__(self, other):
 
     def __iter__(self):
         return iter(self.ids)
+
+
+class WorkOrder(object):
+    def __init__(self, kfrag_hrac, pfrags, receipt_bytes, receipt_signature, bob_pubkey_sig, ursula_id=None):
+        self.kfrag_hrac = kfrag_hrac
+        self.pfrags = pfrags
```
michwill: Why do we have many `pfrags`? Did you mean to be able to process many ciphertexts in one order?
The input (which we call `pfrag`) is actually the same for all Ursulas, so it's not really correct to call it `frag`

tests/keystore/test_keystore.py
```diff
@@ -72,19 +74,24 @@ def test_key_sqlite_keystore(self):
             key = self.ks.get_key(fingerprint_priv)
 
     def test_keyfrag_sqlite(self):
+        kfrag_component_length = 32
         rand_sig = API.secure_random(65)
-        rand_id = b'\x00' + API.secure_random(32)
-        rand_key = b'\x00' + API.secure_random(32)
+        rand_id = b'\x00' + API.secure_random(kfrag_component_length)
+        rand_key = b'\x00' + API.secure_random(kfrag_component_length)
```
tuxxy: This won't be a valid key, fyi. I guess it has a chance of being valid as long as it's within the parameters of the curve, though..

tuxxy:
Also, it would be so nice if GitHub could not use red/green as their damn addition/remove colors. As a color blind person, it's hard af to see and not inclusive at all.

tests/network/test_network_upgrade.py
```diff
@@ -15,3 +24,41 @@ def test_bob_can_follow_treasure_map(enacted_policy_group, ursulas):
...
+    # We'll test against just a single Ursula - here, we made a WorkOrder for just one.
+    work_orders = bob.generate_work_orders(enacted_policy_group, the_pfrag, num_ursulas=1)
+    assert len(work_orders) == 1
+
+    networky_stuff = MockNetworkyStuff(ursulas)
+
+    ursula_dht_key, work_order = list(work_orders.items())[0]
+    cfrags = bob.get_reencrypted_c_frag(networky_stuff, work_order)
+
+    the_cfrag = cfrags[0]  # We only gave one pFrag, so we only got one cFrag.
```
michwill: You maybe meant that we have only one Ursula, that's why we've got only one cFrag? I think, number of cfrags is equal to the number of Ursulas?
Or it combines also?

tests/keystore/test_keystore.py
michwill: Eh, not sure. I think from Ursula's perspective it will be valid.
It has a rekey (big number) and id (which Ursula just attaches). And with any big number, Ursula can even re-encrypt (she'll just get some garbage out)

jMyles: @tuxxy: If you think that there's lacking truth in this test, I'm definitely interested in reforming it.  My efforts were to keep it as close to your pattern as possible while also forcing it to actually test that the de/serialization was happening (not just the storage).

At the moment, this test does shine crucial light on how kFrags are represented as bytes.

jMyles:
@tuxxy: On the matter of the API returning Fragment objects: I'm with ya, kid.

We need to have a documented and maintained membrane, below which everything is bytes, and above which most things are more sculpted abstractions.  If `api` ain't that membrane, that's fine, but we do need it to be somewhere.

I agree that this feels wrong - we're intermixing the world of byte interplay with the world of human narrative, and I prefer that they only talk to each other through enumerated, approved, maintained, agreed-upon channels.

PR #143, opened 2017-12-10
grant - convenience method for creating and enacting a policy_group by Alice.

jMyles:


nkms/characters.py
```diff
@@ -226,6 +231,25 @@ def publish_treasure_map(self, policy_group):
...
+        if not m:
+            # TODO: get m from config
+            raise NotImplementedError
+        if not n:
+            # TODO: get n from config
+            raise NotImplementedError
+        if not expiration:
+            # TODO: check default duration in config
+            raise NotImplementedError
+
```
michwill: I'd also take deposit proportional to time from config (e.g. mETH/day/n or something) if deposit is not set. Wdyt?

jMyles:
I suppose it depends what exactly `deposit` ends up being.  I imagined it as an object that needs to be specially sculpted and signed by Alice, and that if no such object is provided, `grant` lacks the *ability* to spend her money.

michwill:
I imagined that grand _does_ spend Alice's money and deploys a paid policy. How it spends money (e.g. what the wallet is etc) is set in config.

jMyles:
OK, yeah that makes sense.

jMyles:
I'll deal with it tomorrow; I think I'm about wrapped for today.

nkms/characters.py
tuxxy: I would research the costs of Amazon KMS and Google KMS and build a competitive algorithm to determine cost/day.

michwill: Eh, I suspect amazon kms might be way too expensive. We might (?) have some price discovery. But that's not an easy problem.
Still, good to know what they offer

tuxxy: I meant to say that we build a pricing algorithm to beat their pricing model. They charge per key, IIRC.

tuxxy:
This looks great!

michwill:
That's true indeed, looks good!

jMyles:
@tux: Yeah I thought you'd like this PR.  I like this general pattern:

`grant(...)` - fairly widely understood concept throughout security models

And then, for folks who care to dive in, it calls four functions which represent "shop talk" for us internally (ie, uses *our* terminology, but in an expressive, poetic way):

`create_policy_group`
`craft_offer`
`find_n_ursulas`
`enact_policies`

...and then returns a `PolicyGroup`, telling the reader were to look next for the logic that drives the project.

PR #149, opened 2017-12-20
Think of it like the final dungeon, but with a bunch of side-quests still left to do.

jMyles:


nkms/crypto/encrypting_keypair.py
```diff
@@ -71,6 +71,19 @@ def encrypt(self,
         return ((ec.serialize(ekey.ekey), None),
                 cipher.encrypt(data))
 
+    def _decrypt_refactored_for_posterity(self,
+                                          combined_critical_mass, encrypted_message,
```
michwill: Ah, I see™. You pretty much replaced ugly indexing [0] and [1] with human-readable names. Great

nkms/crypto/powers.py
```diff
@@ -94,6 +94,59 @@ def encrypt_for(self, pubkey, cleartext):
...
+        Decrypt data which was re-encrypted for our public key
+        Format of edata is the same as the output of reencrypt() method:
+        data encrypted with an ephemeral key
+        and the ephemeral private key encrypted for recepient (Bob)
+        """
+        edata_by_eph, encrypted_eph = edata
+        priv_eph = self._decrypt(encrypted_eph)
+        return self._decrypt(edata_by_eph, privkey=priv_eph)
+
+    def _decrypt(self,
```
michwill: Btw this doesn't have to be underscored. At the end of the day, Alice uses this to read her own data

michwill: Also am I right that this will go to `keypair`?

nkms/characters.py
```diff
@@ -373,13 +377,10 @@ def alice(self, alice_object):
...
-            getter = self.server.get(ursula_interface_id)
-            loop = asyncio.get_event_loop()
-            value = loop.run_until_complete(getter)
-            
+            value = self.server.get_now(ursula_interface_id)
+
             # TODO: Make this much prettier
-            header, signature, ursula_pubkey_sig, _hrac, (port, interface, ttl) =\
-            dht_value_splitter(value, msgpack_remainder=True)
+            header, signature, ursula_pubkey_sig, _hrac, (port, interface, ttl) = dht_value_splitter(value, msgpack_remainder=True)
```
michwill: Why not 80 chars? ;-)

PR #160, opened 2018-02-13
A very large change; this introduces Umbral and many of the abstractions we created to go with it.

jMyles:
### What this does:
1. Introduces pyumbral, our proxy re-encryption library.
2. Deprecates previous proxy re-encryption tooling.
3. Introduces some important abstractions to accommodate pyumbral metaphor:
- a. `nkms.crypto.kits.CryptoKit` - a tool for carrying common crypto entities together
- b. `Capsule` - an umbral concept, this is the object which `Bob` must bring with him on his journey.  It completely replaces pfrag (though is a different abstraction entirely)
4. `Characters` now use the same class of key (`UmbralPublicKey` and `UmbralPrivateKey`) for both signing and encrypting (though different instances of course).


nkms/crypto/signature.py
```diff
@@ -1,57 +1,59 @@
...
+from cryptography.hazmat.primitives.asymmetric.utils import decode_dss_signature, encode_dss_signature
 
 
-class Signature(bytes):
+class Signature(object):
     """
     The Signature object allows signatures to be made and verified.
     """
-    _EXPECTED_LENGTH = 65
+    _EXPECTED_LENGTH = 64  # With secp256k1 and BLAKE2b(64).
```
cygnusv: The hash function doesn't affect this, only the order of the curve

nkms/crypto/signature.py
```diff
@@ -1,57 +1,59 @@
...
-            if not len(sig_as_bytes) == self._EXPECTED_LENGTH:
-                raise ValueError("{} must be {} bytes.".format(self.__class__.__name__, self._EXPECTED_LENGTH))
-            v, r, s = API.ecdsa_load_sig(sig_as_bytes)
-
-        self._v = v
-        self._r = r
-        self._s = s
+    def __init__(self, r: int, s: int):
+        self.r = r
+        self.s = s
```
cygnusv: Shouldn't this check for invalid `r` and `s` inputs? For example, ints greater than the order of the curve.

nkms/crypto/signature.py
```diff
@@ -1,57 +1,59 @@
...
 
     def __repr__(self):
-        return "{} v{}: {} - {}".format(__class__.__name__, self._v, self._r, self._s)
+        return "ECDSA Signature: {}".format(bytes(self).hex()[:15])
```
cygnusv: Why just the first 15 bytes?

nkms/crypto/signature.py
```diff
@@ -1,57 +1,59 @@
...
+
+    def __add__(self, other):
+        return bytes(self) + other
+
+    def __radd__(self, other):
+        return other + bytes(self)
+
+    def __eq__(self, other):
+        # TODO: Consider constant time
+        return bytes(self) == bytes(other) or self._der_encoded_bytes() == other
```
cygnusv: https://cryptography.io/en/latest/hazmat/primitives/constant-time/#cryptography.hazmat.primitives.constant_time.bytes_eq

nkms/crypto/api.py
```diff
@@ -67,378 +57,43 @@ def keccak_digest(
...
-        rekey = umbral.RekeyFrag(None, priv_bytes2ec(rekey))
-    if type(enc_key) == bytes:
-        enc_key = umbral.EncryptedKey(priv_bytes2ec(enc_key), None)
-    reencrypted_data = PRE.reencrypt(rekey, enc_key)
-    return CFrag(reencrypted_data=reencrypted_data)
+    try:
+        cryptography_pub_key.verify(
+            signature,
+            message,
+            ec.ECDSA(hashes.BLAKE2b(64))
```
cygnusv: I wonder if we should have `hashes.BLAKE2b(64)` like a constant somewhere

nkms/crypto/utils.py
```diff
@@ -60,3 +126,11 @@ def __call__(self, splittable):
...
             splittable = remainder
         return messages
+
+
+def fingerprint_from_key(public_key: Any):
+    """
+    Hashes a key using keccak-256 and returns the hexdigest in bytes.
+    :return: Hexdigest fingerprint of key (keccak-256) in bytes
+    """
+    return keccak_digest(bytes(public_key)).hex().encode()
```
cygnusv: Why is Keccak here? Weren't we using Blake2b?

tuxxy: If we need to store a fingerprint or something in a smart contract, it's easier for us to just use what Ethereum uses -- Keccak-256.

nkms/crypto/signature.py
tuxxy: This would be a great use of this cryptography method.

nkms/crypto/signature.py
tuxxy: Yeah, it probably should. I'm not sure if we have figured out an interface to store umbral/curve params yet. We need to talk about this.

nkms/crypto/signature.py
michwill: I guess, to just identify the signature in a readable way rather than to print the full signature

nkms/crypto/signature.py
jMyles: The digest length of the hash affects it, no?  Or do I have that wrong?

nkms/crypto/signature.py
jMyles: Yeah, I agree with @tuxxy.  Can we perhaps make a TODO and address it after a merge?

nkms/crypto/signature.py
jMyles: I like to keep `__repr__` reasonably short so that it can be skimmed in logs.  If you think there's a need to have the whole sig, I don't have a huge problem with that.

nkms/crypto/signature.py
jMyles: Yeah, absolutely.  Since I already pointed to it as a TODO, do you think it needs to be handled in this merge?  Or can we break it out in a future issue / PR?

nkms/crypto/api.py
jMyles: 👍

nkms/crypto/utils.py
jMyles: Yeah, @tuxxy and I went around and around on this and decided that, since it's quite plausible that we'll need to do this in a smart contract, it makes sense to stick to keccak.

nkms/keystore/keystore.py
```diff
@@ -80,89 +58,90 @@ def get_key(self, fingerprint: bytes) -> Union[keypairs.EncryptingKeypair,
...
+        return policy_contract
 
-        kfrag = KeyFrag(hrac, kfrag_data)
-        self.session.add(kfrag)
+    def del_policy_contract(self, hrac: bytes):
+        """
+        Deletes a PolicyContract from the Keystore.
+        """
+        self.session.query(PolicyContract).filter_by(hrac=hrac).delete()
         self.session.commit()
```
KPrasch: Perhaps for a future PR - but we probably want to return (at minimum) the number of deleted records.

nkms/policy/models.py
```diff
@@ -1,19 +1,20 @@
...
 from nkms.crypto.signature import Signature
+from nkms.crypto.splitters import key_splitter
 from nkms.crypto.utils import BytestringSplitter
-from nkms.keystore.keypairs import PublicKey
+from nkms.network.constants import BYTESTRING_IS_TREASURE_MAP
+from umbral.keys import UmbralPublicKey
+from umbral.umbral import Capsule
 
 
 class Contract(object):
```
KPrasch: Not to be confused with "smart contract"...

nkms/policy/models.py
```diff
@@ -86,9 +93,9 @@ class Policy(object):
...
     """
     _ursula = None
-    hashed_part = None
 
-    def __init__(self, alice, bob=None, kfrags=(UNKNOWN_KFRAG,), pfrag=None, uri=None, alices_signature=NOT_SIGNED):
+    def __init__(self, alice, bob=None, kfrags=(UNKNOWN_KFRAG,), uri=None,
+                 alices_signature=NOT_SIGNED):
         """
         :param kfrags:  A list of KFrags to distribute per this Policy.
         :param pfrag: The input ciphertext which Bob will give to Ursula to re-encrypt.
```
KPrasch: Stale docstring here

nkms/policy/models.py
```diff
@@ -262,47 +267,52 @@ def __len__(self):
...
+        other.receipt_bytes, other.receipt_signature)
 
     def __len__(self):
-        return len(self.pfrags)
+        return len(self.capsules)
 
     @classmethod
-    def constructed_by_bob(cls, kfrag_hrac, pfrags, ursula_dht_key, bob):
-        receipt_bytes = b"wo:" + ursula_dht_key  # TODO: represent the pfrags as bytes and hash them as part of the receipt, ie  + keccak_digest(b"".join(pfrags))  - See #137
+    def constructed_by_bob(cls, kfrag_hrac, capsules, ursula_dht_key, bob):
```
KPrasch: Wrong tense..?

tests/characters/test_alice_can_grant_and_revoke.py
```diff
@@ -1,36 +1,56 @@
...
     uri = b"this_is_the_path_to_which_access_is_being_granted"
-    policy = alice.grant(bob, uri, networky_stuff, m=3, n=n, expiration=policy_end_datetime)
+    policy = alice.grant(bob, uri, networky_stuff, m=3, n=n,
+                         expiration=policy_end_datetime)
 
     # The number of policies is equal to the number of Ursulas we're using (n)
     assert len(policy._accepted_contracts) == n
 
     # Let's look at the first Ursula.
     ursula = list(policy._accepted_contracts.values())[0].ursula
```
KPrasch: A call to `next()` may suffice here instead of casting to a list.

tests/characters/test_bob_handles_frags.py
```diff
@@ -22,7 +24,7 @@ def test_bob_can_follow_treasure_map(enacted_policy, ursulas, alice, bob):
     assert len(bob._ursulas) == len(treasure_map)
 
 
-def test_bob_can_issue_a_work_order_to_a_specific_ursula(enacted_policy, alice, bob, ursulas):
+def test_bob_can_issue_a_work_order_to_a_specific_ursula(enacted_policy, alice, bob, ursulas, alicebob_side_channel):
```
KPrasch: +1

nkms/crypto/signature.py
cygnusv: No, both r and s are integers modulo the order of the curve group. The hash is irrelevant to that matter.
BTW, the choice of a 512-bits hash is somewhat weird since the curve order is 256 bits. @tuxxy do you know what's the performance of blake2b with 32 bytes vs 64 bytes?

tests/network/test_network_actors.py
```diff
@@ -91,8 +96,10 @@ def test_alice_finds_ursula(alice, ursulas):
     getter = alice.server.get(all_ursulas[ursula_index])
     loop = asyncio.get_event_loop()
     value = loop.run_until_complete(getter)
-    _signature, _ursula_pubkey_sig, _hrac, interface_info = dht_value_splitter(value.lstrip(b"uaddr-"),
```
KPrasch: +1; Awesome work finding this bug.

tests/network/test_network_upgrade.py
```diff
@@ -4,5 +4,5 @@ def test_alice_enacts_policies_in_policy_group_via_rest(enacted_policy):
     and transmitting them via REST.
     """
     ursula = list(enacted_policy._accepted_contracts.values())[0].ursula
-    kfrag_that_was_set = ursula.keystore.get_kfrag(enacted_policy.hrac())
-    assert bool(kfrag_that_was_set)  # TODO: This can be a more poignant assertion.
+    policy_contract = ursula.keystore.get_policy_contract(enacted_policy.hrac().hex().encode())
+    assert bool(policy_contract)  # TODO: This can be a more poignant assertion.
```
KPrasch: Yes, agreed.

nkms/crypto/signature.py
jMyles: Oh right - that makes sense.  

As for the digest: it's sadly a moot(-ish) point, as cryptography.io only allows `sign()` with 64 bytes.

nkms/keystore/keystore.py
jMyles: 👍

nkms/policy/models.py
jMyles: Hold my beer.

nkms/policy/models.py
jMyles: Stale?  More like putrefied.

tests/characters/test_alice_can_grant_and_revoke.py
jMyles: 👍

nkms/crypto/signature.py
cygnusv: Sounds good :-) just wanted to point this out

nkms/crypto/signature.py
cygnusv: https://cryptography.io/en/latest/hazmat/primitives/asymmetric/ec/#cryptography.hazmat.primitives.asymmetric.ec.ECDSA

The example in here uses a 32-bytes hash function

nkms/crypto/signature.py
cygnusv: No no, just wanted to know the reason. Plus, 15 seems weird since it encodes 7.5 bytes

nkms/crypto/signature.py
cygnusv: Not urgent at all, since this doesn't involve secret data or secret keys

nkms/crypto/utils.py
cygnusv: Ok, great. Perhaps a comment?

nkms/crypto/signature.py
jMyles: Perhaps there's something (in the internals of cryptography.io) that is specific to BLAKE2b - trying to use a 32-bit digest raises an exception.

nkms/crypto/utils.py
jMyles: OK, good idea - coming up.

nkms/policy/models.py
jMyles: Yep, fixing.  Good call.

nkms/crypto/signature.py
cygnusv: https://cryptography.io/en/latest/hazmat/primitives/cryptographic-hashes/#cryptography.hazmat.primitives.hashes.BLAKE2b

You're right, man. This is a cryptography.io issue. Nevermind then. 

The thing is that the message is first hashed to 64 bytes, but afterwards it's reduced to the order of the curve group (32 bytes), which makes it somewhat pointless to hash to 64 bytes initially. In any case this doesn't affect security.

tests/characters/test_alice_can_grant_and_revoke.py
jMyles: Actually, we (sadly and frustratingly) can't do that here, because this is a `dict_values` and not an iterator.

nkms/crypto/signature.py
jMyles: Sure, but it might affect performance, especially for a high-throughput Alice or Bob.

nkms/crypto/signature.py
jMyles: Heh... well, shall I change it?

michwill: Fine for me.. It's clear that repr isn't for serialization, and that is more clear with 7.5 bytes!

nkms/policy/models.py
michwill: "Not smart contract" :-D

cygnusv: DumbContract

nkms/crypto/signature.py
jMyles: ✔️

nkms/crypto/signature.py
jMyles: ✔️

nkms/crypto/utils.py
jMyles: ✔️

nkms/keystore/keystore.py
jMyles: ✔️

nkms/policy/models.py
jMyles: ✔️

tests/characters/test_alice_can_grant_and_revoke.py
jMyles: ✔️

tests/characters/test_bob_handles_frags.py
jMyles: ✔️

tests/network/test_network_actors.py
jMyles: ✔️

tests/network/test_network_upgrade.py
jMyles: ✔️

nkms/crypto/signature.py
```diff
@@ -7,9 +7,10 @@ class Signature(object):
     """
     The Signature object allows signatures to be made and verified.
     """
-    _EXPECTED_LENGTH = 64  # With secp256k1 and BLAKE2b(64).
+    _EXPECTED_LENGTH = 64  # With secp256k1
 
     def __init__(self, r: int, s: int):
+        #  TODO: Sanity check for proper r and s.
```
tuxxy: 👍

jMyles: ✔️

nkms/crypto/utils.py
```diff
@@ -45,9 +85,35 @@ def __call__(self, splittable, return_remainder=False, msgpack_remainder=False):
     def __len__(self):
         return sum(self.get_message_meta(m)[1] for m in self.message_types)
 
+
     @staticmethod
     def get_message_meta(message_type):
-        return message_type if isinstance(message_type, tuple) else (message_type, message_type._EXPECTED_LENGTH)
+        try:
```
michwill: Probably ok for now, but looks like quite a bit of try/except magic is going on. In future, I'd consider looking at this place again to reduce possibilities of passing the wrong message_type which produces a non-error output

Pipfile
```diff
@@ -8,29 +8,29 @@ name = "pypi"
...
+SQLAlchemy = "*"
+Twisted = "*"
+pyOpenSSL = "*"
+service_identity = "*"
 apistar = "*"
 mypy = "*"
-"pytest-mypy" = "*"
+pytest-mypy = "*"
 maya = "*"
+kms-depend = {git = "https://github.com/nucypher/pyumbral.git"}
```
tuxxy: Let's get this on PyPi, as well.

nkms/crypto/api.py
```diff
@@ -49,13 +38,14 @@ def secure_random_range(
     return SYSTEM_RAND.randrange(min, max)
 
 
-def keccak_digest(
-        *messages: bytes
-) -> bytes:
+def keccak_digest(*messages: bytes) -> bytes:
```
tuxxy: For reasonable compatibility, shall we include a `blake2b_digest` function that performs similarly?

nkms/crypto/powers.py
```diff
@@ -239,49 +218,11 @@ def gen_path_keys(
...
-        :return: Decrypted data
-        """
-        privkey = privkey or self.priv_key
-        ciphertext, enc_key = enc_data
-
-        enc_key = API.elliptic_curve.deserialize(API.PRE.ecgroup, enc_key)
-        enc_key = API.umbral.EncryptedKey(ekey=enc_key, re_id=None)
-
-        dec_key = API.ecies_decapsulate(privkey, enc_key)
+    def decrypt(self, message_kit: MessageKit) -> bytes:
```
tuxxy: Absolutely beautiful!

nkms/crypto/powers.py
```diff
@@ -21,80 +22,62 @@ class NoEncryptingPower(PowerUpError):
...
             raise NoEncryptingPower
 
-    def encrypt_for(self, pubkey, cleartext):
-        try:
-            encrypting_power = self._power_ups[EncryptingPower]
-            ciphertext = encrypting_power.encrypt(cleartext, bytes(pubkey))
-            return ciphertext
-        except KeyError:
-            raise NoEncryptingPower
+    def encrypt_for(self, recipient_pubkey_enc, plaintext):
```
tuxxy: Love these red lines, this file looks so much better now! Great job, everyone!

nkms/crypto/constants.py
```diff
@@ -1,5 +1,15 @@
-# TODO: Turn these into classes?
-HASH_DIGEST_LENGTH = 32
+from cryptography.hazmat.primitives import hashes
+BLAKE2B = hashes.BLAKE2b(64)
```
tuxxy: I don't think this should be in this file, but it can stay for now. Let's think of a better place for it.

nkms/crypto/signature.py
```diff
@@ -1,57 +1,60 @@
...
-        msg_digest = API.keccak_digest(message)
-        return API.ecdsa_verify(self._v, self._r, self._s, msg_digest, pubkey.without_metabytes())
+        return API.ecdsa_verify(message, self._der_encoded_bytes(), pubkey)
+
+    @classmethod
+    def from_bytes(cls, signature_as_bytes, der_encoded=False):
+        if der_encoded:
+            r, s = decode_dss_signature(signature_as_bytes)
+        else:
+            if not len(signature_as_bytes) == 64:
```
tuxxy: We should probably add a TODO here to get this number as two times the order of the curve being used instead of hardcoding this `64` in here.

We can probably do this through our Umbral params.

nkms/crypto/signature.py
```diff
@@ -1,57 +1,60 @@
...
+                raise ValueError("Looking for exactly 64 bytes if you call from_bytes with der_encoded=False.")
+            else:
+                r = int.from_bytes(signature_as_bytes[:32], "big")
```
tuxxy: We definitely need to be clear here that this `32` is the size of the order of the curve. This is also in Umbral params, but we can probably make constants for this somewhere, too.

nkms/crypto/utils.py
tuxxy: Agreed, we should probably be more clear with what's going on here. It's not inherently clear to me what each `try`  block is trying to do first.

nkms/keystore/keystore.py
tuxxy: This will only delete a single `PolicyContract`. See the call to `.first()`.

nkms/network/protocols.py
```diff
@@ -59,14 +68,19 @@ def rpc_store(self, sender, nodeid, key, value):
         self.welcomeIfNewNode(source)
         self.log.debug("got a store request from %s" % str(sender))
 
-        if value.startswith(b"uaddr") or value.startswith(b"trmap"):
-            signature, sender_pubkey_sig, hrac, message = dht_value_splitter(value[5::], return_remainder=True)
+        if value.startswith(BYTESTRING_IS_URSULA_IFACE_INFO) or value.startswith(
+                BYTESTRING_IS_TREASURE_MAP):
+            header, signature, sender_pubkey_sig, hrac, message = dht_value_splitter(
```
tuxxy: In the wise words of @jMyles, "Should this be a class?"

:)

nkms/policy/models.py
```diff
@@ -176,37 +178,39 @@ def treasure_map_dht_key(self):
...
-            encrypted_treasure_map)  # TODO: Ideally, this is a Ciphertext object instead of msgpack (see #112)
+        dht_value = signature_for_ursula + self.alice.seal + self.hrac() + tmap_message_kit.to_bytes()
         dht_key = self.treasure_map_dht_key()
 
-        setter = self.alice.server.set(dht_key, b"trmap" + dht_value)
+        setter = self.alice.server.set(dht_key, BYTESTRING_IS_TREASURE_MAP + dht_value)
         event_loop = asyncio.get_event_loop()
         event_loop.run_until_complete(setter)
-        return encrypted_treasure_map, dht_value, signature_for_bob, signature_for_ursula
+        return tmap_message_kit, dht_value, signature_for_bob, signature_for_ursula
```
tuxxy: @jMyles What do you think? Would this make a good `Kit` or just a different class?

I think it would be ideal to avoid returning so many values like this.

tests/crypto/test_bytestring_types.py
```diff
@@ -20,30 +24,17 @@ def test_split_two_signatures():
 
 def test_split_signature_from_arbitrary_bytes():
     how_many_bytes = 10
-    signature = Signature(secure_random(65))
+    signature = Signature.from_bytes(secure_random(64))
```
tuxxy: This will fail in the future when we verify that `r` and `s` are within the order of the curve.

tests/crypto/test_bytestring_types.py
```diff
@@ -20,30 +24,17 @@ def test_split_two_signatures():
...
-    splitter = BytestringSplitter(KFrag, (bytes, how_many_bytes))
-    rebuilt_kfrag, rebuilt_bytes = splitter(kfrag + some_bytes)
-    assert kfrag == rebuilt_kfrag
-
-
 def test_trying_to_extract_too_many_bytes_raises_typeerror():
     how_many_bytes = 10
     too_many_bytes = 11
-    signature = Signature(secure_random(65))
+    signature = Signature.from_bytes(secure_random(64))
```
tuxxy: Same failure here, too.

tests/keystore/test_keystore.py
```diff
@@ -1,99 +1,64 @@
...
+    new_workorder1 = test_keystore.add_workorder(bob_keypair_sig1.pubkey, b'test0', hrac)
+    new_workorder2 = test_keystore.add_workorder(bob_keypair_sig2.pubkey, b'test1', hrac)
+
+    # Test get workorder
+    query_workorders = test_keystore.get_workorders(hrac)
+    assert {new_workorder1, new_workorder2}.issubset(query_workorders)
+
+    # Test del workorder
+    deleted = test_keystore.del_workorders(hrac)
+    assert deleted > 0
```
tuxxy: 😅 Thanks, guys! 🥈

tests/network/test_network_actors.py
tuxxy: Also, +1. This was a tricky find. Good job!

tests/network/test_network_upgrade.py
tuxxy: Also, agreed.

tests/crypto/test_bytestring_types.py
cygnusv: Only with probability 3.7e-39, but yes.

nkms/crypto/utils.py
KPrasch: Perhaps another in-line comment or two on the subscriptions, but I actually think this is a fairly elegant and idiomatic way of determining what category of byte-string you have, and gives the user info they need to use ByteStingSplitter correctly.

nkms/policy/models.py
KPrasch: ...AdventureKit

tests/characters/test_crypto_characters_and_their_powers.py
```diff
@@ -75,51 +76,79 @@ def test_anybody_can_verify():
...
-    cleartext = b"This is Officer Rod Farva. Come in, Ursula!  Come in Ursula!"
+"""
+What follows are various combinations of signing and encrypting, to match real-world scenarios.
+"""
+
+def test_sign_cleartext_and_encrypt(alice, bob):
+    """
+    Exhibit One: Alice signs the cleartext and encrypts her signature inside the ciphertext.
+    """
+    message = b"Have you accepted my answer on StackOverflow yet?"
```
KPrasch: No.

nkms/keystore/keystore.py
KPrasch: @tuxxy - Hmmm. not seeing a call to `.first()` in `del_policy_contract`. Looks like it bulk deletes all matching records from the `filter_by`.

nkms/network/protocols.py
```diff
@@ -1,16 +1,20 @@
...
-from nkms.keystore.keypairs import PublicKey
-from nkms.network.constants import NODE_HAS_NO_STORAGE
+from nkms.network.constants import NODE_HAS_NO_STORAGE, BYTESTRING_IS_URSULA_IFACE_INFO, \
+    BYTESTRING_IS_TREASURE_MAP, DHT_VALUE_HEADER_LENGTH
 from nkms.network.node import NuCypherNode
 from nkms.network.routing import NuCypherRoutingTable
+from umbral.keys import UmbralPublicKey
 
-dht_value_splitter = BytestringSplitter(Signature, PublicKey, (bytes, HASH_DIGEST_LENGTH))
+dht_value_splitter = BytestringSplitter((bytes, DHT_VALUE_HEADER_LENGTH), Signature,
```
KPrasch: :+1: BytestringSplitter!

nkms/keystore/keystore.py
jMyles: So where are we on this?  `hrac` is already unique on `PolicyContract`, so we know that we're deleting at most one here.  However, I also don't see a call to `.first()`.  

What do we need to do to make it ready for merge?

nkms/crypto/utils.py
jMyles: I think I'm with @KPrasch on this one - it's not like it's weird nesting or anything.  It's three try blocks: one for each expected element in the tuple.  

Ask yourself: will you be more comfortable if this is three `ifs`?  If so, I assert that this is more elegant than that.

nkms/crypto/api.py
jMyles: Do you just mean a version that takes an iterable?  Do you feel like that's a good thing for this PR?  If so, sure I'll through it together.

nkms/crypto/powers.py
jMyles: ✔️

nkms/crypto/powers.py
jMyles: ✔️

tests/crypto/test_bytestring_types.py
jMyles: Yeah, I mean, there are many reasons why I don't like this literal here.  So you needn't sell me on that.  However, I'm really just trying (to the degree this is possible) to refrain from adding enhancements here and just keep doing what we were doing, but with pyumbral.  Maybe make a new issue for this?

tests/crypto/test_bytestring_types.py
jMyles: True story.  And same comment above.

tests/characters/test_crypto_characters_and_their_powers.py
jMyles: 🔴

tests/keystore/test_keystore.py
jMyles: ✔️

nkms/network/protocols.py
jMyles: ✔️

nkms/crypto/signature.py
jMyles: Added such a TODO.

✔️

nkms/crypto/signature.py
jMyles: Added such a TODO.

✔️

tests/network/test_network_actors.py
jMyles: ✔️

tests/network/test_network_upgrade.py
jMyles: ✔️

Pipfile
jMyles: ✔️

tests/characters/test_crypto_characters_and_their_powers.py
jMyles: ✔️

nkms/network/protocols.py
jMyles: ✔️

tests/characters/test_crypto_characters_and_their_powers.py
tuxxy: I didn't see this. 😂

PR #162, opened 2018-02-13
WIde variety of fixes, mostly dealing with byte length.

jMyles:
This PR is somewhat hard to summarize; reviewing is probably easier done by looking at commit messages (or just the entire diff).


nkms/characters.py
```diff
@@ -701,8 +703,8 @@ def __radd__(self, other):
     def __len__(self):
         return len(bytes(self))
 
-    def without_metabytes(self):
-        return self.character._crypto_power.pubkey_sig_bytes().without_metabytes()
+    def as_umbral_pubkey(self):
+        return self.character.public_key(SigningPower)
```
KPrasch: very nice.

nkms/crypto/utils.py
```diff
@@ -1,12 +1,46 @@
+from contextlib import suppress
```
KPrasch: An actual use case?

PR #170, opened 2018-02-24
Reflecting refactor of umbral.umbral to umbral.pre, some touched-up docstrings and code org.

jMyles:


nkms/network/protocols.py
```diff
@@ -55,6 +55,7 @@ def determine_legality_of_dht_key(self, signature, sender_pubkey_sig,
         verified = signature.verify(hrac, sender_pubkey_sig)
 
         if not verified or not proper_key == dht_key:
+            # Hachidan Kiritsu, it's illegal!
```
tuxxy: .. I... hmm... ?

nkms/network/server.py
```diff
@@ -62,14 +62,14 @@ def serialize_capabilities(self):
 
     def get_now(self, key):
```
tuxxy: Do you think we could rename this to something more self documenting?

jMyles: Hmm - sure!  I just took `get` and made it happen synchronous / blocking (ie, "get it *now*").  

What do you suggest instead?

nkms/network/protocols.py
KPrasch: Anbo-jyutsu

nkms/network/server.py
tuxxy: Ah, I guess that makes sense then. It just feels weird to simply see `self.get` or `self.get_now`.

I think I prefer `get_<thing>_now` so it's more clear, but this is fine too.

jMyles: Yeah... I'm just not sure what to make <thing>.  :-)

`get_value_now`?  I'm not sure if `value` adds anything to me or not, but I guess it does reduce the initial cryptic phrasing.

PR #171, opened 2018-02-24
Lots of good redlines.  Cleaning up Character and Powers.

jMyles:


nkms/characters.py
```diff
@@ -173,31 +162,32 @@ def encrypt_for(self,
...
+                signature = self.stamp(plaintext)
+                ciphertext, capsule = pre.encrypt(recipient_pubkey_enc, signature + plaintext)
             else:
-                message_kit = self._crypto_power.encrypt_for(
-                        actor.public_key(EncryptingPower), plaintext)
-                signature = self.seal(message_kit.ciphertext)
-            message_kit.alice_pubkey = self.public_key(SigningPower)
+                # Encrypt first, sign second.
+                ciphertext, capsule = pre.encrypt(recipient_pubkey_enc, plaintext)
+                signature = self.stamp(ciphertext)
```
tuxxy: This is great

nkms/characters.py
```diff
@@ -253,44 +243,41 @@ def verify_from(self,
...
+        power_up = self._crypto_power.power_ups(power_up_class)
+        return power_up.public_key()
 
 
 class Alice(Character):
     _server_class = NuCypherSeedOnlyDHTServer
     _default_crypto_powerups = [SigningPower, EncryptingPower]
 
-    def generate_kfrags(self, bob, m, n):
+    def generate_kfrags(self, bob, m, n) -> List:
```
tuxxy: I feel like we can give this a better name than `generate_kfrags` as that doesn't really sound like it's doing anything like `split_rekey`. However, it's too late for me to offer a solution at this time. I'll think about this and come back to you.

nkms/characters.py
```diff
@@ -367,19 +357,15 @@ def alice(self):
...
-            getter = self.server.get(ursula_interface_id)
-            loop = asyncio.get_event_loop()
-            value = loop.run_until_complete(getter)
-            
+            value = self.server.get_now(ursula_interface_id)
+
             # TODO: Make this much prettier
-            header, signature, ursula_pubkey_sig, _hrac, (port, interface, ttl) =\
-            dht_value_splitter(value, msgpack_remainder=True)
+            header, signature, ursula_pubkey_sig, _hrac, (port, interface, ttl) = dht_value_splitter(value, msgpack_remainder=True)
```
tuxxy: I still think we need to make this a `CryptoKit` of some sort.

nkms/characters.py
```diff
@@ -515,20 +500,19 @@ def from_rest_url(cls, url):
...
+                                          return_remainder=True)
         stranger_ursula_from_public_keys = cls.from_public_keys(
-                    signing=signing_key_bytes, encrypting=encrypting_key_bytes)
+            signing=signing_key_bytes, encrypting=encrypting_key_bytes)
         return stranger_ursula_from_public_keys
 
     def attach_server(self, ksize=20, alpha=3, id=None,
                       storage=None, *args, **kwargs):
-
         # TODO: Network-wide deterministic ID generation (ie, auction or
```
tuxxy: Is this a blocking issue for testnet?

Let's have a conversation about how to achieve this for testnet sooner than later.

nkms/policy/models.py
```diff
@@ -45,14 +46,14 @@ def __init__(self, alice, hrac, expiration, deposit=None, ursula=None,
...
-        return bytes(self.alice.seal) + bytes(
+        return bytes(self.alice.stamp) + bytes(
             self.hrac) + self.expiration.isoformat().encode() + bytes(
             self.deposit)
 
     @classmethod
     def from_bytes(cls, contract_as_bytes):
         contract_splitter = key_splitter + BytestringSplitter((bytes, KECCAK_DIGEST_LENGTH),
-            (bytes, 26))
+                                                              (bytes, 26))
```
tuxxy: `26` should probably be a constant somewhere, imho.

nkms/characters.py
tuxxy: Still would prefer to perform this as a `CryptoKit` though.

nkms/characters.py
```diff
@@ -147,21 +143,14 @@ def server(self) -> Server:
     def name(self):
         return self.__class__.__name__
 
-    def hash(self, message):
```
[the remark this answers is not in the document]
jMyles: :-)

nkms/characters.py
jMyles: haha yeah.  No question.  That's what I think every time I run into that TODO.

nkms/characters.py
jMyles: No it's not.  In fact it's obsolete as far as I can tell at this moment.

nkms/policy/models.py
jMyles: `TWENTY_SIX = 26` ?

I can do that.

nkms/characters.py
jMyles: It is - just a few lines down I make a `MessageKit` from it.  Or do you mean that you prefer for `pre.encrypt` to return the `MessageKit`?

tuxxy: I meant I prefer the `stamp` to be a kit instead of a `stamp`

jMyles: Oh right, I had forgotten that you had that notion.  I still don't quite get it - what goes in this kit?

nkms/characters.py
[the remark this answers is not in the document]
KPrasch: Excellent.

nkms/keystore/keypairs.py
```diff
@@ -71,7 +72,24 @@ def decrypt(self, message_kit: MessageKit) -> bytes:
 
         :return: bytes
         """
-        return message_kit.decypt(self.privkey)
+        cleartext = pre.decrypt(message_kit.capsule, self._privkey,
```
KPrasch: Why not go all the way here and just let `pre.decrypt` just take a `MessageKit` as an arg.  It may even be more intuitive to users who are already familiar with PKE to just pass a private key and a message... `pre.decrypt(message_kit, private_key)`

nkms/network/server.py
```diff
@@ -80,3 +90,111 @@ class NuCypherSeedOnlyDHTServer(NuCypherDHTServer):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.storage = SeedOnlyStorage()
+
+
+class ProxyRESTServer(object):
```
KPrasch: :+1:  I imagine this with have some implementation similarity with the `Miner` class in the kms-eth codebase.

nkms/network/server.py
```diff
@@ -80,3 +90,111 @@ class NuCypherSeedOnlyDHTServer(NuCypherDHTServer):
...
+        """
+        return Response(
+            content=bytes(self.stamp) + bytes(self.public_key(EncryptingPower)),
+            content_type="application/octet-stream")
+
+    def consider_contract(self, hrac_as_hex, request: http.Request):
+        # TODO: This actually needs to be a REST endpoint, with the payload
+        # carrying the kfrag hash separately.
+        from nkms.policy.models import Contract
+        contract, deposit_as_bytes = \
```
KPrasch: Ugly line break!

nkms/policy/models.py
```diff
@@ -276,30 +280,30 @@ def __init__(self, bob, kfrag_hrac, capsules, receipt_bytes,
         self.ursula_id = ursula_id  # TODO: We may still need a more elegant system for ID'ing Ursula.  See #136.
 
     def __repr__(self):
```
KPrasch: This is practically a logging method.

nkms/policy/models.py
```diff
@@ -311,9 +315,42 @@ def payload(self):
...
+    def __len__(self):
+        return sum(len(work_orders) for work_orders in self.by_ursula.values())
+
+    @property
+    def ursulas(self):
+        return self.by_ursula.keys()
+
+    def by_capsule(self, capsule):
+        ursulas_by_capsules = {}
+        for ursula, pfrags in self.by_ursula.items():
```
KPrasch: "pfrags" are used here?, also, any change to speed this up from O(n2)?

nkms/keystore/keypairs.py
jMyles: I mean you know I'm :+1: on that.

PR #173, opened 2018-02-27
Working nodes: Ursula and AliceBob.

jMyles:
This includes a number of compelling upgrades:

1) Alice and Bob can now get metadata from a REST endpoint instead of the DHT if they already know Ursula's interface info.
2) Alice can push a TreasureMap via REST instead of the DHT.
3) Major cleanup of Characters (still more to do, but it's a big step)
4) Ursula's datastore is now multithreaded and includes eager loading of KFrags.
5) Entry-point script for non-mining Ursula is working (ie, runs a real node, as shown on Twitch earlier today).
6) AliceBob entry script properly engages real Ursula node.


jMyles:
Here's the copy of Finnegan's Wake I was using in my Twitch demo:

[finnegans-wake.txt](https://github.com/nucypher/nucypher-kms/files/1765576/finnegans-wake.txt)

tests/characters/test_bob_handles_frags.py
```diff
@@ -61,6 +61,9 @@ def test_bob_can_issue_a_work_order_to_a_specific_ursula(enacted_policy, alice,
     networky_stuff = MockNetworkyStuff(ursulas)
     ursula_dht_key, work_order = list(work_orders.items())[0]
 
+    # In the real world, we'll have a full Ursula node here.  But in this case, we need to fake it.
+    work_order.ursula = ursulas[0]
```
michwill: ursulas[0] isn't necessarily the one corresponding to ursula_dht_key
So at the end, bob._saved_work_orders has both ursula_dht_key and ursulas[0].interface_dht_key() if they are different

michwill: Fix:
```
    for ursula in ursulas:
        if ursula.interface_dht_key() == ursula_dht_key:
            work_order.ursula = ursula
            break
```

jMyles: Fixed instead by implementing an `OrderedDict` both going into and coming out of the `TreasureMap`.

PR #178, opened 2018-03-07
Ursula now uses TLS.

jMyles:


nkms/keystore/keypairs.py
```diff
@@ -109,3 +110,8 @@ def sign(self, message: bytes) -> bytes:
         """
         signature_der_bytes = API.ecdsa_sign(message, self._privkey)
         return Signature.from_bytes(signature_der_bytes, der_encoded=True)
+
+    def generate_self_signed_cert(self, common_name):
+        # TODO: Let's have a shortcut method for getting the cryptography key(s).
+        cryptography_key = self._privkey.bn_key.to_cryptography_priv_key()
```
cygnusv: Oh yes, definitely, I remember thinking about this at some moment, but it seems I forgot

entry_points/run_ursula_with_rest_and_dht_but_no_mining.py
```diff
@@ -3,18 +3,33 @@
...
 _URSULA.listen()
 
-from hendrix.deploy.base import HendrixDeploy
+CURVE = ec.SECP256R1
+cert, private_key = generate_self_signed_certificate(_URSULA.stamp.fingerprint().decode(), CURVE)
 
-deployer = HendrixDeploy("start", {"wsgi":_URSULA.rest_app, "http_port": 3500})
+deployer = HendrixDeployTLS("start",
+                            {"wsgi":_URSULA.rest_app, "https_port": 3550},
+                            key=private_key,
```
tuxxy: How does hendrix handle this? Do you specify a key, then the curve for that key?

jMyles: Checkout the context factory class that's used here:

PR #189, opened 2018-04-05
[WIP] Add config and begin integration of KMS-Ethereum

tuxxy:
TODO: Fill this out


nkms/config/utils.py
```diff
@@ -0,0 +1,148 @@
...
+    )
+
+    for rule, failure_message in rules:
+        if not rule:
+            raise KMSConfig.KMSConfigurationError(failure_message)
+    else:
+        return passphrase
+
+
+def _derive_master_key_from_passphrase(salt: bytes, passphrase: str) -> bytes:
```
jMyles: passphrase - `bytes`?  Or do we really want unicode here?

tuxxy: It's user provided input and it gets encoded to bytes for derivation.

PR #190, opened 2018-04-05
Better constants and a new test.

jMyles:


jMyles:
Aye, good lookin' out.  Fixed now.

PR #193, opened 2018-04-07
Tests Passing with some kms-eth integration and config.

jMyles:


nkms/config/interface.py
```diff
@@ -0,0 +1,49 @@
...
+    return user_passphrase
+
+
+def start():
+    print(title, welcome, description, newlines, sep='\n')
+    input(press_any)
+    gather_passphrase()
+
+
+start()
```
michwill: What do you guys like using for cli tools? `click`? Something else?

PR #195, opened 2018-04-10
Absorb KMS-ETH codebase.

jMyles:


nkms/blockchain/eth/nkms_eth/blockchain.py
```diff
@@ -0,0 +1,83 @@
...
+
+    @classmethod
+    def get(cls):
+        if cls.__instance is None:
+            class_name = cls.__name__
+            raise Exception('{} has not been created.'.format(class_name))
+        return cls.__instance
+
+    def disconnect(self):
+        self._chain.__exit__(None, None, None)
```
michwill: Perhaps, makes sense to specify keyword arguments. Not super-important here, but I'm much more comfortable when I know what these Nones actually stand for

nkms/blockchain/eth/nkms_eth/config.py
```diff
@@ -0,0 +1,79 @@
...
+    __max_allowed_locked = 10 ** 7 * NuCypherTokenConfig._M
+
+    _null_addr = '0x' + '0' * 40
+    __reward = NuCypherTokenConfig._reward
+
+    __mining_coeff = [
+        _hours_per_period,
+        2 * 10 ** 7,
+        __max_awarded_periods,
+        __max_awarded_periods,
```
michwill: I don't think it's self-evident what these parameters are, esp. when we have two `__max_awarded_periods`. Should we add some explaining comments?

jMyles:
@michwill: No, this is not where we wants tests to live.  Once we achieve consensus on this merge strategy, let's:

1) Make a few subsequent commits to promote clean organization and namespacing
2) Import Issues and PRs from the other repo.

PR #217, opened 2018-04-11
Separating Delegation from Encapsulation

### LINKED ISSUE #182: Enrique: the forgotten hero of PRE
I have been increasingly bothered lately by a nagging thought that our cryptology has given short-shrift to a Character whose power in the context of the PRE narrative is potentially incredible.

Until now, all of our logic assumes that Alice does two things:

1) Generates KFrags and puts in motion the mechanics of a `Policy`
2) Encrypts The Secret for Bob, leaving a `Capsule` for him to find.

...but why do we presume that these two things are done by the same person?

Now, of course, in an orthodox PKE cryptological narrative, Alice is chiefly identified by her performance of the latter function - she is the one with the message for Bob.

On that basis, it may be tempting to continue to tie her to this idea, carving out the former and naming a different Character.  But, as @tuxxy pointed out to me in the pool just a little while ago, this actually doesn't make sense: the consequence of Alice's presence on the network follows her private key (ie, that's what makes her who she is, and reminds us of the attack surfaces which we must defend on her behalf).  And it is the "delegator" who needs a private key, not the "encapsulator."

So, on that basis, I propose that we begin a tradition of regarding the PRE narrative as involving *four* actors by introducing **Enrique the Encapsulator** (or, if we're feeling cheeky, "the pharmacist" - ie, he makes `Capsules`).

This opens up the ability to reason about a number of compelling use-cases, such as the following:

A user wants to build a distributed feed of operations data for a swarm of nodes (NuCypher or otherwise).  This user wants to delegate in advance who may consume such a feed, but wants to disappear from the network afterward, leaving the authority structure in place.  Here's how that works:

* Alice meets and selects 12 Bobs who are authorized to consume the data feeds.
* She `grants` access to each of those 12 Bobs, and then deletes her private key (or perhaps keeps it on a hardware wallet or similar).  Let's refer to this scenario as "DisappearingAlice".
* Each app which wants to publish data for the 12 Bobs encrypts for Alice's public key.
* Each Bob embarks on The Quest per usual.

Of course it is step 3 here that is typically absent from our use case discussions, but notice the turn of events here: Now, *anybody* can encrypt for all 12 Bobs, simply by knowing Alice's public key. 

My sense - and this is just preliminary, based on a few days of deep thoughts about it - is that all of the most mind-blowing uses cases for NuCypher will be of the four character narrative.

I also propose that if indeed Alice is also Enrique, we call her BigAlice.  

Among the consequences of this change for the codebase:

* We create a `DelegatingPower` which is part of the `grant` flow; with it, a user can perform `generate_kfrags(...)`.
* We establish `Enrique` and move `encrypt()` there.
* We establish the necessary facilities for DisappearingAlice.

jMyles:
Alice can now `grant` and then disappear from the network, never to be seen again.

Subsequently, any number of `DataSources` can supply messages for `Bob`.

This is comparable to our discussions about a four-character narrative involving a `Character` named `Enrique` the Encapsulator (See #182).

However, once `Enrique` was written, it was clear that he *wasn't* a `Character`.  He didn't require any special powers (remember, his only notable role was to encrypt, which any `Character` can do).

Additionally, unlike the other characters in our cryptological narrative, `Enrique`'s exact interests are very hard to pin down: he doesn't pay, nor does he stake, nor does he rely on messages from the other actors.  In fact, his role is arguably optional: `Alice` *can* do everything he does.

So, `DataSource` now does what we imagined `Enrique` doing: signing messages and making `Capsules` our of them.

This PR closes #182 and gives finality to #218.

See what y'all think.

### DIFF
--- a/nkms/characters.py
+++ b/nkms/characters.py
@@ -19,7 +19,7 @@ from nkms.crypto.api import secure_random, keccak_digest, encrypt_and_sign
 from nkms.crypto.constants import PUBLIC_KEY_LENGTH
 from nkms.crypto.kits import UmbralMessageKit
 from nkms.crypto.powers import CryptoPower, SigningPower, EncryptingPower, DelegatingPower
-from nkms.crypto.signature import Signature, signature_splitter, SignatureStamp, StrangerStamp
+from nkms.crypto.signature import Signature, signature_splitter, SignatureStamp
 from nkms.network import blockchain_client
 from nkms.network.protocols import dht_value_splitter
 from nkms.network.server import NuCypherDHTServer, NuCypherSeedOnlyDHTServer, ProxyRESTServer
@@ -70,7 +70,7 @@ class Character(object):
             crypto_power_ups = []
 
         if is_me:
-            self._stamp = SignatureStamp(self)
+            self._stamp = CharacterStamp(self)
 
             if attach_server:
                 self.attach_server()
@@ -595,3 +595,28 @@ class Ursula(Character, ProxyRESTServer):
                 if work_order.bob == bob:
                     work_orders_from_bob.append(work_order)
             return work_orders_from_bob
+
+
+class CharacterStamp(SignatureStamp):
+    """
+    A stamp meant to be used by a character.
+    """
+    def __init__(self, character):
+        self.character = character
+        self._sign = character.sign
+
+    def __bytes__(self):
+        return bytes(self.character.public_key(SigningPower))
+
+    def as_umbral_pubkey(self):
+        return self.character.public_key(SigningPower)
+
+
+class StrangerStamp(CharacterStamp):
+    """
+    SignatureStamp of a stranger (ie, can only be used to glean public key, not to sign)
+    """
+
+    def __call__(self, *args, **kwargs):
+        message = "This isn't your SignatureStamp; it belongs to {} (a Stranger).  You can't sign with it."
+        raise TypeError(message.format(self.character))
--- a/nkms/characters.py
+++ b/nkms/characters.py
@@ -19,7 +19,7 @@ from nkms.config.configs import KMSConfig
 from nkms.crypto.api import secure_random, keccak_digest
 from nkms.crypto.constants import PUBLIC_KEY_LENGTH
 from nkms.crypto.kits import UmbralMessageKit
-from nkms.crypto.powers import CryptoPower, SigningPower, EncryptingPower
+from nkms.crypto.powers import CryptoPower, SigningPower, EncryptingPower, DelegatingPower
 from nkms.crypto.signature import Signature
 from nkms.crypto.splitters import signature_splitter
 from nkms.network import blockchain_client
@@ -303,7 +303,7 @@ class FakePolicyAgent:  # TODO: #192
 
 class Alice(Character, PolicyAuthor):
     _server_class = NuCypherSeedOnlyDHTServer
-    _default_crypto_powerups = [SigningPower, EncryptingPower]
+    _default_crypto_powerups = [SigningPower, EncryptingPower, DelegatingPower]
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -320,7 +320,7 @@ class Alice(Character, PolicyAuthor):
         :param n: Total number of kfrags to generate
         """
         bob_pubkey_enc = bob.public_key(EncryptingPower)
-        return self._crypto_power.power_ups(EncryptingPower).generate_kfrags(bob_pubkey_enc, m, n)
+        return self._crypto_power.power_ups(DelegatingPower).generate_kfrags(bob_pubkey_enc, m, n)
 
     def create_policy(self, bob: "Bob", uri: bytes, m: int, n: int):
         """
--- a/nkms/characters.py
+++ b/nkms/characters.py
@@ -171,26 +171,13 @@ class Character(object):
         :return: A tuple, (ciphertext, signature).  If sign==False,
             then signature will be NOT_SIGNED.
         """
-        recipient_pubkey_enc = recipient.public_key(EncryptingPower)
-        if sign:
-            if sign_plaintext:
-                # Sign first, encrypt second.
-                sig_header = constants.SIGNATURE_TO_FOLLOW
-                signature = self.stamp(plaintext)
-                ciphertext, capsule = pre.encrypt(recipient_pubkey_enc, sig_header + signature + plaintext)
-            else:
-                # Encrypt first, sign second.
-                sig_header = constants.SIGNATURE_IS_ON_CIPHERTEXT
-                ciphertext, capsule = pre.encrypt(recipient_pubkey_enc, sig_header + plaintext)
-                signature = self.stamp(ciphertext)
-            alice_pubkey = self.public_key(SigningPower)
-        else:
-            # Don't sign.
-            signature = sig_header = constants.NOT_SIGNED
-            alice_pubkey = None
-            ciphertext, capsule = pre.encrypt(recipient_pubkey_enc, sig_header + plaintext)
-        message_kit = UmbralMessageKit(ciphertext=ciphertext, capsule=capsule, alice_pubkey=alice_pubkey)
+        signer = self.stamp if sign else constants.DO_NOT_SIGN.bool_value(False)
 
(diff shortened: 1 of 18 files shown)


nkms/characters.py
```diff
@@ -171,26 +169,13 @@ def encrypt_for(self,
...
-                ciphertext, capsule = pre.encrypt(recipient_pubkey_enc, sig_header + plaintext)
-                signature = self.stamp(ciphertext)
-            alice_pubkey = self.public_key(SigningPower)
-        else:
-            # Don't sign.
-            signature = sig_header = constants.NOT_SIGNED
-            alice_pubkey = None
-            ciphertext, capsule = pre.encrypt(recipient_pubkey_enc, sig_header + plaintext)
-        message_kit = UmbralMessageKit(ciphertext=ciphertext, capsule=capsule, alice_pubkey=alice_pubkey)
+        signer = self.stamp if sign else constants.DO_NOT_SIGN.bool_value(False)
```
tuxxy: I find `constants.DO_NOT_SIGN.bool_value(False)` to be very confusing in comparison to `constants.DO_NOT_SIGN(False)`.

nkms/characters.py
```diff
@@ -213,13 +198,13 @@ def verify_from(self,
         :return: Whether or not the signature is valid, the decrypted plaintext
             or NO_DECRYPTION_PERFORMED
         """
+        sender_pubkey_sig = actor_whom_sender_claims_to_be.stamp.as_umbral_pubkey()
         with suppress(AttributeError):
-            if message_kit.alice_pubkey:
-                if not message_kit.alice_pubkey == actor_whom_sender_claims_to_be.public_key(SigningPower):
+            if message_kit.sender_pubkey_sig:
+                if not message_kit.sender_pubkey_sig == sender_pubkey_sig:
```
tuxxy: This is a nice improvement. :)

nkms/characters.py
```diff
@@ -595,28 +599,3 @@ def work_orders(self, bob=None):
                 if work_order.bob == bob:
                     work_orders_from_bob.append(work_order)
             return work_orders_from_bob
-
-
-class CharacterStamp(SignatureStamp):
```
tuxxy: Cool, yeah this makes more sense too.

jMyles: Yeah, good looking out on your part.

nkms/crypto/signature.py
```diff
@@ -59,3 +61,63 @@ def __radd__(self, other):
...
+        return bytes(self) + other
+
+    def __radd__(self, other):
+        return other + bytes(self)
+
+    def __len__(self):
+        return len(bytes(self))
+
+    def __bool__(self):
+        return True
```
KPrasch: I cannot tell whether or not this is True

nkms/crypto/signature.py
```diff
@@ -59,3 +61,63 @@ def __radd__(self, other):
...
+    def fingerprint(self):
+        """
+        Hashes the key using keccak-256 and returns the hexdigest in bytes.
+
+        :return: Hexdigest fingerprint of key (keccak-256) in bytes
+        """
+        return keccak_digest(bytes(self)).hex().encode()
+
+
+class StrangerStamp(SignatureStamp):
```
KPrasch: Its a very feature-rich class

nkms/data_sources.py
```diff
@@ -0,0 +1,15 @@
+from nkms.crypto.api import encrypt_and_sign
+
+
+class DataSource:
```
KPrasch: Needs docstring, but :+1:

tests/network/test_network_actors.py
```diff
@@ -5,7 +5,6 @@
 from kademlia.utils import digest
 
 from constant_sorrow import constants
-from nkms.blockchain.eth.utilities import spawn_miners
```
KPrasch: No miners for you?

nkms/crypto/signature.py
jMyles: `return True if not KPrasch.can_tell_whether_this_is_True else True`

nkms/crypto/signature.py
jMyles: Not only that - and I don't blame you if you didn't notice this on first read - but it's also transient, in-memory, local, and private.

nkms/data_sources.py
jMyles: Mmm yeah.  It does.

nkms/crypto/signature.py
jMyles: No money down.

nkms/data_sources.py
```diff
@@ -0,0 +1,15 @@
...
+
+    def __init__(self, policy_pubkey_enc, signer):
+        self._policy_pubkey_enc = policy_pubkey_enc
```
cygnusv: Is `policy_pubkey_enc` the `UmbralPublicKey` of Alice? If that's the case, I think the name is not very good, since the same `UmbralPublicKey` can be used in more than one policies.

nkms/keystore/keypairs.py
```diff
@@ -76,7 +76,7 @@ def decrypt(self, message_kit: MessageKit) -> bytes:
         cleartext = pre.decrypt(ciphertext=message_kit.ciphertext,
                                 capsule=message_kit.capsule,
                                 priv_key=self._privkey,
-                                alice_pub_key=message_kit.alice_pubkey)
+                                alice_pub_key=message_kit.policy_pubkey)
```
cygnusv: Ah yes, `policy_pubkey` is the former `alice_pubkey`. I don't like that name. What's wrong with `alice_pubkey`?

tests/characters/test_bob_handles_frags.py
```diff
@@ -93,22 +93,22 @@ def test_bob_can_issue_a_work_order_to_a_specific_ursula(enacted_policy, alice,
 
 
 def test_bob_remembers_that_he_has_cfrags_for_a_particular_capsule(enacted_policy, alice, bob,
-                                                                   ursulas, alicebob_side_channel):
+                                                                   ursulas, capsule_side_channel):
     # In our last episode, Bob made a WorkOrder for the capsule...
```
cygnusv: Great narrative :)

tests/network/test_network_actors.py
cygnusv: NEXT!

cygnusv: NEXT!

jMyles: COME BACK, ONE YEAR!

nkms/characters.py
```diff
@@ -320,7 +309,7 @@ def generate_kfrags(self, bob, m, n) -> List:
         :param n: Total number of kfrags to generate
         """
         bob_pubkey_enc = bob.public_key(EncryptingPower)
-        return self._crypto_power.power_ups(EncryptingPower).generate_kfrags(bob_pubkey_enc, m, n)
+        return self._crypto_power.power_ups(DelegatingPower).generate_kfrags(bob_pubkey_enc, m, n)
```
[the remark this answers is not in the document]
jMyles: Ahh, right.  Well, we'll pass Alice's stamp in there.

nkms/crypto/signature.py
```diff
@@ -59,3 +61,63 @@ def __radd__(self, other):
...
+class SignatureStamp(object):
+    """
+    Can be called to sign something or used to express the signing public
+    key as bytes.
+    """
+
+    def __init__(self, signing_keypair):
+        self._sign = signing_keypair.sign
+        self._as_bytes = bytes(signing_keypair.pubkey)
+        self._as_umbral_pubkey = signing_keypair.pubkey
```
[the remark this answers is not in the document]
jMyles: > And also UmbralPublicKeys are used for encryption only

Well of course currently, with our ECDSA scheme, this isn't the case.  If/when we switch to Curve25519 / EdDSA, we'll obviously stop representing the key this way.

nkms/data_sources.py
jMyles: Hmm... does anything about `policy_pubkey_enc` make it sound like it is exclusive to this `Policy`?  It's just the public key of this (and perhaps other) policies.

nkms/keystore/keypairs.py
jMyles: Well, is it really alice's pubkey if she disappears from the internet, memory of her lost to the sands of time?  Does anybody at that point care who Alice is or was?  Isn't the only important thing that the key is the right key for that Policy?

PR #219, opened 2018-04-11
Spice up config

tuxxy:
### What this does:
1. Changes some names (ie: `master_key` to `key_material`).
2. Creates public keyfiles in `Keyring.generate`.
3. Uses BLAKE2b as the hash in the HKDF.

### DIFF
--- a/nkms/config/keys.py
+++ b/nkms/config/keys.py
@@ -145,8 +145,10 @@ class KMSKeyring:
 
     __default_key_filepaths = {
         'root': os.path.join(__default_private_key_dir, 'root_key.priv'),
+        'root_pub': os.path.join(__default_public_key_dir, 'root_key.pub'),
         'signing': os.path.join(__default_private_key_dir, 'signing_key.priv'),
-        'transacting': os.path.join(__default_private_key_dir, 'wallet.json')
+        'signing_pub': os.path.join(__default_public_key_dir, 'signing_key.pub'),
+        'transacting': os.path.join(__default_private_key_dir, 'wallet.json'),
     }
 
     class KeyringError(Exception):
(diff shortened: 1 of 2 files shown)


nkms/config/keys.py
```diff
@@ -37,13 +37,13 @@ def validate_passphrase(passphrase) -> bool:
...
 
-def _derive_master_key_from_passphrase(salt: bytes, passphrase: str) -> bytes:
+def _derive_key_material_from_passphrase(salt: bytes, passphrase: str) -> bytes:
     """
     Uses Scrypt derivation to derive a  key for encrypting key material.
     See RFC 7914 for n, r, and p value selections.
     This takes around ~5 seconds to perform.
     """
-    master_key = Scrypt(
+    key_material = Scrypt(
```
KPrasch: :+1:

nkms/config/utils.py
```diff
@@ -5,7 +5,7 @@
 from .configs import _DEFAULT_CONFIGURATION_DIR, KMSConfigurationError
```
KPrasch: Hmm - perhaps this is a constant via `ConstantSorrow`?

nkms/config/keys.py
```diff
@@ -253,18 +263,21 @@ def generate(cls, passphrase: str, encryption: bool=True, transacting: bool=True
...
         os.mkdir(cls.__default_public_key_dir, mode=0o744)  # public
-        os.mkdir(_private_key_dir, mode=0o744)              # private
+        os.mkdir(_private_key_dir, mode=0o700)              # private
 
         # Generate keys
         keyring_args = dict()
         if encryption is True:
-            enc_key, _ = _generate_encryption_keys()
-            sig_key, _ = _generate_signing_keys()
+            enc_privkey, enc_pubkey = _generate_encryption_keys()
```
KPrasch: Let's move this logic into a private method and call it here.

tuxxy: :-1: Until this gets cleaned up more, I don't really feel like moving the code to another function will help things. :/

KPrasch:
:+1:

PR #224, opened 2018-04-13
[WIP] Add Registrar and Provider

tuxxy:
### What this does:
1. Adds a `Registrar` to handle smart contract bytecode ABI using the filesystem and JSON.
2. Adds tests for the `Registrar`.
3. Adds a `Provider` to interact with the `Registrar`. This allows the `Provider` to compile, deploy, and cache contracts as needed.
4. Assists in removing Populus from the codebase.


tests/eth_fixtures.py
```diff
@@ -1,16 +1,33 @@
...
+from nkms.blockchain.eth.interfaces import Registrar, Provider
 
 
 @pytest.fixture(scope='session')
-def chain():
+def tester_registrar():
+    _, filepath = tempfile.mkstemp()
+    tester_registrar = Registrar(chain_name='tester', registrar_filepath=filepath)
+    yield tester_registrar
+    os.remove(filepath)
```
michwill: Checked `py.test` docs - looks like indeed, the line after `yield` will execute even if an exception happened :+1:

tuxxy:
Too big to review on its own. Closing and merging into branches on the side. This is the culmination of lots of work.

PR #225, opened 2018-04-13
Updates miner config constants

KPrasch:


nkms/blockchain/eth/constants.py
```diff
@@ -1,31 +1,31 @@
...
         return self.__saturation
 
 
 class NuCypherMinerConfig:
-    _hours_per_period = 24       # Hours
-    _min_release_periods = 30    # 720 Hours
-    __max_awarded_periods = 365
+    _hours_per_period = 24       # Hours in single period
+    _min_release_periods = 30    # 720 Hours minimum
+    __max_minting_periods = 365  # Maximum number of periods
```
szotov: It's not really maximum number of periods. It's maximum number of  periods that will be rewarded additionally. Ursula can lock longer but will have same reward for locking in case `__max_minting_periods` or 2 * `__max_minting_periods`

nkms/blockchain/eth/constants.py
```diff
@@ -1,31 +1,31 @@
...
-    __reward = NuCypherTokenConfig._reward
+    __remaining_supply = NuCypherTokenConfig._remaining_supply
 
-    __mining_coeff = [
+    __mining_coeff = [           # TODO
         _hours_per_period,
         2 * 10 ** 7,
-        __max_awarded_periods,
-        __max_awarded_periods,
+        __max_minting_periods,
```
szotov: Maybe this will help: this line is coefficient that used to calculate reward for long locking. For example, when this coefficient is equivalent to `__max_minting_periods`, then max reward for long locking will be 2.

nkms/blockchain/eth/constants.py
jMyles: `__increasing_reward_periods`?

michwill: Maybe we can include a formula which explains what it is additionally?
E.g.:
```
reward_rate ~ (1 + min(lock_time, __max_minting_periods) / __max_minting_periods) / 2
```
Would it make it more understandable?

PR #229, opened 2018-04-14
Actors can now discover all known nodes from a given node.  Fixes #175.

### LINKED ISSUE #175: Character.learn_about_nodes() needs to learn about nodes about which target node knows, not just target node.
Currently, `learn_about_nodes()` only informs the caller about the target node.  Instead, it needs to inform about all knows about which target node is aware. 

Also, it might make sense (for example, for Bob) to be able to only ask about certain nodes by passing... a list or something.

jMyles:


tests/utilities.py
```diff
@@ -87,3 +87,25 @@ def get_treasure_map_from_node(self, node, map_id):
...
+            if ursula.rest_port == port:
+                rest_app = ursula.rest_app
+                break
+        else:
+            raise RuntimeError("Can't find that one - did you spin up the right test ursulas?")
+        mock_client = TestClient(ursula.rest_app)
+        response = mock_client.get("http://localhost/public_keys")
+        return response
+
+    def get_nodes_via_rest(self, address, port):
```
michwill: Is this defined outside `MockNetworkyStuff`, or the real version of this method still has to be implemented?

jMyles: Yeah; I'll implement it in `NetworkyStuff` (maybe to be renamed `NetworkMiddleware`?) when I refresh the Finnegan's Wake demo (ie, soon).

jMyles:
Subsumed after a rebase by #242.

Issue #241, opened 2018-04-17
How shall we generate a signing key for a DataSource?


jMyles:

tuxxy:
The same way that any one else does, in the config when the node first comes online.

michwill:
The minimal version is when DataSource provides a signing key from their config and communicates to everyone by whatever means (how to communicate can be outside the scope of the testnet release).

Do we want to allow unsigned data, or do we want to have signing key to be short-lived instead, for a whistleblower application or alike?

tuxxy:
We should always sign data for security purposes. If a whistleblower wants to use the application, they need to use ephemeral keys to do so.

tuxxy:
B U M P
U
M
P

cygnusv:
I think this belongs in pyUmbral.

Issue #231, opened 2018-04-15
Single, authoritative piece of logic for deriving policy HRAC


jMyles:
Currently, we do this sort of thing on the fly all over the place:

`keccak_digest(bytes(alice.stamp) + bytes(bob.stamp) + labeli)`

Let's have one place where this occurs.

PR #242, opened 2018-04-17
Plausible draft developer-facing API.

### LINKED ISSUE #228: Stop including hash of interface info in node interface details.
Actors can learn about nodes (for now, Ursulas) by consuming the node interface details (available on the DHT or via REST).

We include a hash of the interface info, which is signed by the node.  But... we also include the interface info itself, rendering the hash superfluous.

### LINKED ISSUE #232: Clarify two Ursula ports - one for DHT, one for REST.
There are a few places where we fudge this.  No good.

### LINKED ISSUE #239: known_nodes: tuples, Ursula instances, or something else?
Currently, there are places in the codebases where we add Ursula instances to a Character's `known_nodes`, and other places where we add only tuples of interface info.

For now, we probably want to just use Ursulas, because we may want to be forward-compatible with "non-Ursula helper nodes" - ie, nodes that help find other nodes, but which don't stake.

jMyles:
# What this does
* Alice and Bob no longer need to use the DHT.
* Alice automatically pushes `TreasureMap` during `grant()`.
* "network middleware" (new working name for `NetworkyStuff`) is directly composed on each `Character`.
* `TreasureMap` now includes the value of `m` and, for each Ursula, both a REST and DHT port.  Fixes #232.
* `known_nodes` are Ursula instances, not tuples.  Fixes #239.
* Alice now hashes Ursula's interface info, eliminating the need for Ursula to include it - fixes #228 
* High-level interfaces for Bob:
** `join_policy()` connects Bob to the network (via a known REST endpoint), gets the appropriate `TreasureMap`, and learns about all available nodes mentioned in it.
** `retrieve()` takes a `MessageKit` from a `DataSource` and asks for re-encryption from `m` Ursulas, attaches the CFrags, verifies the message, and returns the cleartext.
* Finnegan's Wake demo is updated with the above.

This has been a pretty crazy solo sprint for me - do take some time to review it.

## Other notes

Thanks to @arjunhassard for jamming with me in PDX on the API design, which in turn kicked my ass into finally getting Bob off the DHT.

Arj, here's the Finnegan's Wake demo with the new toys - `DataSource`, `Bob.join_policy(...)`, `Bob.retrieve(...)`, and so on:

https://github.com/jMyles/nucypher-kms/blob/50915fc1f2338371bf486141a7b8a8a0373b0921/examples/alicebob-grant-and-line-by-line-PRE.py

We'll need to clean it up and get some better comments in there.


examples/sandbox_resources.py
```diff
@@ -0,0 +1,32 @@
+import requests
+from nkms.characters import Ursula
+from nkms.network.node import NetworkyStuff
+from nkms.crypto.powers import SigningPower, EncryptingPower
+
+
+class SandboxNetworkyStuff(NetworkyStuff):
```
tuxxy: I thought you said you changed the name to `NetworkMiddleware`?

Did this get overwritten or something?

nkms/network/node.py
```diff
@@ -38,25 +38,36 @@ def find_ursula(self, id, offer=None):
 
     def reencrypt(self, work_order):
         ursula_rest_response = self.send_work_order_payload_to_ursula(work_order)
-        cfrags = RepeatingBytestringSplitter((CapsuleFrag, CFRAG_LENGTH))(ursula_rest_response.content)
+        cfrags = BytestringSplitter((CapsuleFrag, CFRAG_LENGTH)).repeat(ursula_rest_response.content)
```
tuxxy: Do we have public-facing documentation on exactly what `.repeat()` does? Perhaps it should be in the BytestringSplitter repo?

nkms/data_sources.py
```diff
@@ -1,10 +1,14 @@
...
+from nkms.keystore.keypairs import SigningKeypair
+from constant_sorrow.constants import NO_SIGNING_POWER
 
 class DataSource:
 
-    def __init__(self, policy_pubkey_enc, signer):
+    def __init__(self, policy_pubkey_enc, signer=NO_SIGNING_POWER):
         self._policy_pubkey_enc = policy_pubkey_enc
+        if signer is NO_SIGNING_POWER:
+            signer = SignatureStamp(SigningKeypair())  # TODO: Generate signing key properly.  #241
```
tuxxy: I don't think I like generating the signing key here. I feel like maybe an error should be raised like `NoSigningKey`.

Key generation shouldn't really be an implicit experience, imho. You either have a key available to perform this action, or you don't.

nkms/characters.py
```diff
@@ -281,9 +285,32 @@ def learn_about_nodes(self, address, port):
...
+                    address, dht_port, rest_port = msgpack.loads(interface_info)
+                    new_nodes[pubkey] = \
+                        Ursula.as_discovered_on_network(
+                            rest_port=rest_port,
+                            dht_port=dht_port,
+                            ip_address=address.decode("utf-8"),
+                            powers_and_keys=({SigningPower: pubkey})
+                        )
+                else:
+                    self.log.warn("Discovered node with bad signature: {}".format(node_meta))
```
tuxxy: This is kinda a big deal. We only want to warn about this?

nkms/characters.py
```diff
@@ -372,33 +400,64 @@ def __init__(self, *args, **kwargs):
...
-
-    def get_treasure_map(self, policy, networky_stuff, using_dht=False):
-        map_id = policy.treasure_map_dht_key()
+    def follow_treasure_map(self, hrac, using_dht=False):
+
+        treasure_map = self.treasure_maps[hrac]
+        number_of_known_treasure_ursulas = 0
+        if not using_dht:
+            for ursula_interface_id in treasure_map:
+                pubkey = UmbralPublicKey.from_bytes(ursula_interface_id)
```
tuxxy: `ursula_interface_id` doesn't seem like a good name for something that is a public key.

nkms/characters.py
```diff
@@ -372,33 +400,64 @@ def __init__(self, *args, **kwargs):
...
+                if pubkey in self.known_nodes:
+                    number_of_known_treasure_ursulas += 1
+
+            newly_discovered_nodes = {}
+            nodes_to_check = iter(self.known_nodes.values())
+
+            while number_of_known_treasure_ursulas < treasure_map.m:
```
tuxxy: This might allow Alice to collude with Ursula nodes to have Bob traverse the map in an order designated by her. Whether or not this can be done maliciously is something I haven't thought about, but certainly could have some implications somewhere.

For example, long ago in the revolutionary war when the first US spy rings were made, knowing the order that a spy would travel to deliver/receive/drop information would be enough to correlate who the spy was and would have horrible consequences for the entire spy ring. If Alice can influence the order that Bob traverses the nodes, this might have bad consequences. In practice, the Tor network is vulnerable to similar attacks that involve timing.

Perhaps we randomize the map before we traverse it? I know that dictionaries are already randomized to an extent, but might it be enough to thwart something here.

nkms/characters.py
```diff
@@ -372,33 +400,64 @@ def __init__(self, *args, **kwargs):
...
+                try:
+                    node_to_check = next(nodes_to_check)
```
tuxxy: Lots of Python magic happening in this function. Can we document the exact algorithm how we traverse this in plain English somewhere? Perhaps in the wiki?

nkms/network/server.py
```diff
@@ -248,25 +253,31 @@ def reencrypt_via_rest(self, hrac_as_hex, request: http.Request):
...
         # TODO: This next line is possibly the worst in the entire codebase at the moment.  #172.
         # Also TODO: TTL?
         do_store = self.server.protocol.determine_legality_of_dht_key(
                     signature_for_ursula, pubkey_sig_alice, tmap_message_kit,
                     hrac, digest(treasure_map_id), request.body)
         if do_store:
-            # TODO: Stop storing things in the protocol storage.  Do this better.
-            # TODO: Propagate to other nodes.
+            # TODO: Stop storing things in the protocol storage.  Do this better.  #227
+            # TODO: Propagate to other nodes.  #235
```
tuxxy: Great work in enumerating all of these issues. Many thanks. :pray:

tests/network/test_network_actors.py
```diff
@@ -48,21 +48,34 @@ def test_vladimir_illegal_interface_key_does_not_propagate(ursulas):
...
                                                                                return_remainder=True)
 
     assert header == constants.BYTESTRING_IS_URSULA_IFACE_INFO
-    port = msgpack.loads(interface_info)[0]
+    port = msgpack.loads(interface_info)[1]
     assert port == URSULA_PORT + ursula_index
 
 
+def test_alice_finds_ursula_via_rest(alice, ursulas):
+    networky_stuff = MockNetworkyStuff(ursulas)
```
tuxxy: I'm confused why we are still calling it `networky_stuff` when it's supposed to be `network_middleware`?

nkms/data_sources.py
jMyles: Let's talk about this on #241.

nkms/characters.py
jMyles: Yeah, I think so.  Although it's a big deal in the sense that we've seemingly found an Ursula who is doing the Wrong Thing(TM), it's not a big deal to this individual Alice/Bob.  They will almost certainly be able to continue to transmit, so I think they'd be pretty annoyed if this raised instead.

Hopefully we get some valuable reports in this scenario.

examples/sandbox_resources.py
jMyles: No, I haven't changed the class yet; just composed it on each Character as `network_middleware` - we can do the full refactor if we all like it, which it sounds like we do.

examples/sandbox_resources.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def __init__(self, ursulas):
+        self.ursulas = ursulas
+
+    def find_ursula(self, contract=None):
```
michwill: Would we have this fort of simplified API not only in demos?

nkms/characters.py
```diff
@@ -372,33 +400,64 @@ def __init__(self, *args, **kwargs):
...
+
+                # TODO: If we're going to implement TTL, it will be here.
+                self.known_nodes[ursula_interface_id] = \
+                    Ursula.as_discovered_on_network(
+                        dht_port=port,
+                        dht_interface=interface,
+                        powers_and_keys=({SigningPower: ursula_pubkey_sig})
+                    )
+
+    def get_treasure_map(self, alice, hrac, using_dht=False):
```
michwill: Ah, I see, we normally use `get_treasure_map` and `follow_treasure_map`

nkms/network/node.py
jMyles: Yeah, true story.

jMyles: See https://github.com/nucypher/byteStringSplitter/issues/1

nkms/characters.py
jMyles: As I said earlier in person, I'm not convinced there's any possible attack based on this, but we might as well do it.  It also has the possible benefit of relieving the "first in line" Ursulas in the case of a high-throughput `DataSource`-=>`Bob` pair.

jMyles: See #245

nkms/characters.py
jMyles: Hmm - so this isn't traversal; this is Bob discovering new nodes until he has >= `m`.  I don't think there's anything resembling "magic" here, although I agree that it can use at least a docstring.

nkms/network/server.py
jMyles: Still many left to do!  :-)

I encourage you and @KPrasch to do the same as you're putting together #237.

jMyles:
Yeah @michwill; agreed - I actually opened an issue for that a couple of days ago - see #231.

nkms/characters.py
arjunhassard: I agree – if I've understood correctly, this would be issue for the network to deal with (possibly through a challenge), not the customer. In the event of a bad signature, would that node be removed from the group that is performing the re-encryption, and replaced by another?

examples/sandbox_resources.py
arjunhassard: The stimulus for this sprint was to work out the simplest, highest-level set of commands that an external developer could write into their codebase, and then walk through these in alicebob-grant-and-line-by-line-PRE.py. This file can then serve as a 'beach head' for customers, who can trial relatively basic functions like ALICE.grant, BOB.join_policy and BOB.retrieve before moving on to greater customization, attempts to economize their usage, more complex sharing workflows, etc. 

So, I think the answer is yes – this simplified API is helpful beyond a demo.

examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -78,16 +56,10 @@ def enact_policy(self, ursula, hrac, payload):
         print("PREs per second: {}".format(counter / seconds))
         print("********************************")
 
-    ciphertext, capsule = pre.encrypt(ALICE.public_key(EncryptingPower), plaintext)
-
-    message_kit = MessageKit(ciphertext=ciphertext, capsule=capsule,
-                      alice_pubkey=ALICE.public_key(EncryptingPower))
+    data_source = DataSource(policy_pubkey_enc=policy.public_key())
+    message_kit, _signature = data_source.encapsulate_single_message(plaintext)
```
arjunhassard: This change manages to convey the greater flexibility afforded to developers in using a DataSource character to handle encryption on behalf of Alice, and yet is easier to understand than the code it replaces. I tip my hat to you sir. 

I will still write some instructions (JN style) plus an in-line note explaining this in more detail here, since it would be cool if they totally get everything in this file while reading it – then they can go dig through all the lower-level stuff.

nkms/characters.py
jMyles: Well, in this case, the user is just learning about any available nodes on the network.  These aren't necessarily the nodes that will be performing re-encryption later.  If the node being learned about doesn't have a valid signature for its ID, then we don't even consider it real - we just toss it aside and emit a warning.

So, @arjunhassard, no, the node isn't remove per se.  Hopefully, a different Ursula provides us with a valid signature for the node's ID.

At the end of the day, the node *from whom we're learning* has done the wrong thing here - they were supposed to refrain from propagating this information after they themselves checked the Signature.  

I have modified the warning message to reflect this in #246, the PR that follows this one.

examples/sandbox_resources.py
jMyles: :+1:

examples/alicebob-grant-and-line-by-line-PRE.py
jMyles: Sure.  And then, in #246, I have evolved it a state further.

PR #243, opened 2018-04-18
Pytest + Geth Fixture

KPrasch:


tests/eth_fixtures.py
```diff
@@ -16,51 +17,83 @@
...
-    test_provider = EthereumTesterProvider(ethereum_tester=eth_tester)  # , api_endpoints=None)
+    #
+    # spin-up geth
+    #
+    testing_dir = tempfile.mkdtemp()
+    chain_name = 'nkms_tester'
+    geth = DevGethProcess(chain_name=chain_name, base_dir=testing_dir)
+    geth.start()
+
+    geth.wait_for_ipc(timeout=2)
```
michwill: 30?

tests/eth_fixtures.py
```diff
@@ -16,51 +17,83 @@
...
     """
     Provider backend
     https: // github.com / ethereum / eth - tester     # available-backends
     """
-    eth_tester = EthereumTester(backend=PyEVMBackend())
-    test_provider = EthereumTesterProvider(ethereum_tester=eth_tester)  # , api_endpoints=None)
+    #
+    # spin-up geth
+    #
+    testing_dir = tempfile.mkdtemp()
```
michwill: Context manager? https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

KPrasch: We decided against this for better multi-platform support.

tests/eth_fixtures.py
KPrasch: Seems long! 10? (this is a module level fixture)

PR #246, opened 2018-04-18
Evolving the Finnegan's Wake demo and its interfaces.

jMyles:
# What this does:

* Further simplifies the Finnegan's Wake demo and adds quite a bit of commentary.
* Uses Policy.public_key (as @michwill points out, probably soon to be a "label key" of some variety) as the basis for the `hrac` instead of Alice's signing key.
* Allows `DataSource` to be reconstructed by Bob for the purpose of `retrieving` content.


examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -1,70 +1,77 @@
...
-# Bob learns about Ursula, gets the TreasureMap, and follows it.
-BOB.learn_about_nodes(address="https://localhost", port="3550")
-networky_stuff = NetworkyStuff()
-BOB.get_treasure_map(policy, networky_stuff)
-BOB.follow_treasure_map(hrac)
-
-# Now, Alice and Bob are ready for some throughput.
 
+# Alice puts her public key somewhere for Bob to find later...
+alices_pubkey_saved_for_posterity = bytes(ALICE.stamp)
```
michwill: When we have a per-label derived key, do we have a label-specific pubkey (stamp)?

examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -1,70 +1,77 @@
...
 
-# Alice grants to Bob.
+# Alice gets on the network and, knowing about at least one Ursula,
+# Is able to discover all Ursulas.
+ALICE.network_bootstrap([("localhost", 3601)])
 
-policy = ALICE.grant(BOB, uri, networky_stuff, m=1, n=n,
+# Alice grants to Bob.
+BOB = Bob()
+policy = ALICE.grant(BOB, label, m=m, n=n,
```
michwill: I think, we still should allow label=None. Maybe in this implementation - still with a _chimney_

michwill: (it would be `label=''` in such case)

jMyles:
Obseleted by #248

PR #248, opened 2018-04-19
Using new label derivation and evolving Finnegan's Wake demo.

jMyles:
# What this does:

* Further simplifies the Finnegan's Wake demo and adds quite a bit of commentary.
* Allows `DataSource` to be reconstructed by Bob for the purpose of `retrieving` content.
* Rewrite of `DelegatingPower` - it now derives its keypair from its `label`.

### DIFF
--- a/examples/alicebob-grant-and-line-by-line-PRE.py
+++ b/examples/alicebob-grant-and-line-by-line-PRE.py
@@ -1,24 +1,27 @@
 # This is an example of Alice setting a Policy on the NuCypher network.
 # In this example, Alice uses n=1, which is almost always a bad idea.  Don't do it.
 
-# WIP w/ hendrix@8227c4abcb37ee6d27528a13ec22d55ee106107f
+# WIP w/ hendrix@3.0.0
 
 import datetime
 import sys
 
 from examples.sandbox_resources import SandboxNetworkyStuff
 from nkms.characters import Alice, Bob, Ursula
-from nkms.crypto.api import keccak_digest
 from nkms.data_sources import DataSource
 from nkms.network.node import NetworkyStuff
 import maya
 
-# Some basic setup.
-
-ALICE = Alice()
-BOB = Bob()
+# This is already running in another process.
 URSULA = Ursula.from_rest_url(NetworkyStuff(), address="localhost", port=3601)
 network_middleware = SandboxNetworkyStuff([URSULA])
+
+
+#########
+# Alice #
+#########
+
+ALICE = Alice(network_middleware=network_middleware)
 
 # Here are our Policy details.
 policy_end_datetime = maya.now() + datetime.timedelta(days=5)
@@ -32,16 +35,42 @@ label = b"secret/files/and/stuff"
 ALICE.network_bootstrap([("localhost", 3601)])
 
 # Alice grants to Bob.
-policy = ALICE.grant(BOB, label, network_middleware, m=m, n=n,
+BOB = Bob()
+policy = ALICE.grant(BOB, label, m=m, n=n,
                      expiration=policy_end_datetime)
-hrac, treasure_map = policy.hrac(), policy.treasure_map
 
-# Bob can re-assemble the hrac himself with knowledge he already has.
-hrac = keccak_digest(bytes(ALICE.stamp) + bytes(BOB.stamp) + label)
-BOB.join_policy(ALICE, hrac, node_list=[("localhost", 3601)])
+# Alice puts her public key somewhere for Bob to find later...
+alices_pubkey_saved_for_posterity = bytes(ALICE.stamp)
 
-# Now, Alice and Bob are ready for some throughput.
+# ...and then disappears from the internet.
+del ALICE
+# (this is optional of course - she may wish to remain in order to create
+# new policies in the future.  The point is - she is no longer obligated.
+
+#####################
+# some time passes. #
+# ...               #
+# And now for Bob.  #
+#####################
+
+# Bob wants to join the policy so that he can receive any future
+# data shared on it.
+# He needs a few piece of knowledge to do that.
+BOB.join_policy(label,  # The label - he needs to know what data he's after.
+                alices_pubkey_saved_for_posterity,  # To verify the signature, he'll need Alice's public key.
+                verify_sig=True,  # And yes, he usually wants to verify that signature.
+                # He can also bootstrap himself onto the network more quickly
+                # by providing a list of known nodes at this time.
+                node_list=[("localhost", 3601)]
+                )
+
+# Now that Bob has joined the Policy, let's show how DataSources
+# can share data with the members of this Policy and then how Bob retrieves it.
 finnegans_wake = open(sys.argv[1], 'rb')
+
+# We'll also keep track of some metadata to gauge performance.
+# You can safely ignore from here until...
+################################################################################
 
 start_time = datetime.datetime.now()
 
@@ -56,10 +85,58 @@ for counter, plaintext in enumerate(finnegans_wake):
         print("PREs per second: {}".format(counter / seconds))
         print("********************************")
 
-    data_source = DataSource(policy_pubkey_enc=policy.public_key())
+
+################################################################################
+# ...here.  OK, pay attention again.
+# Now it's time for...
+
+    #####################
+    # Using DataSources #
+    #####################
+
+    # Now Alice has set a Policy and Bob has joined it.
+    # You're ready to make some DataSources and encrypt for Bob.
+
+    # It may also be helpful to imagine that you have multiple Bobs,
+    # multiple Labels, or both.
+
+    # First we make a DataSource for this policy.
+    data_source = DataSource(policy_pubkey_enc=policy.public_key)
+
+    # Here's how we generate a MessageKit for the Policy.  We also get a signature
+    # here, which can be passed via a side-channel (or posted somewhere public as
+    # testimony) and verified if desired.  In this case, the plaintext is a
+    # single passage from James Joyce's Finnegan's Wake.
+    # The matter of whether encryption makes the passage more or less readable
+    # is left to the reader to determine.
     message_kit, _signature = data_source.encapsulate_single_message(plaintext)
 
-    delivered_cleartext = BOB.retrieve(hrac=hrac, message_kit=message_kit, data_source=data_source)
+    # The DataSource will want to be able to be verified by Bob, so it leaves
+    # its Public Key somewhere.
+    data_source_public_key = bytes(data_source.stamp)
 
+    # It can save the MessageKit somewhere (IPFS, etc) and then it too can
+    # choose to disappear (although it may also opt to continue transmitting
+    # as many messages as may be appropriate).
+    del data_source
+
+    ###############
+    # Back to Bob #
+    ###############
+
+    # Bob needs to reconstruct the DataSource.
+    datasource_as_understood_by_bob = DataSource.from_public_keys(
+        policy_public_key=policy.public_key,
+        datasource_public_key=data_source_public_key,
+        label=label
+    )
+
+    # Now Bob can retrieve the original message.  He just needs the MessageKit
+    # and the DataSource which produced it.
+    delivered_cleartext = BOB.retrieve(message_kit=message_kit,
+                                       data_source=datasource_as_understood_by_bob,
+                                       alice_pubkey_sig=alices_pubkey_saved_for_posterity)
+
+    # We show that indeed this is the passage originally encrypted by the DataSource.
     assert plaintext == delivered_cleartext
     print("Retrieved: {}".format(delivered_cleartext))
--- a/nkms/characters.py
+++ b/nkms/characters.py
@@ -73,11 +73,9 @@ class Character(object):
         if crypto_power:
             self._crypto_power = crypto_power
         elif crypto_power_ups:
-            self._crypto_power = CryptoPower(power_ups=crypto_power_ups,
-                                             generate_keys_if_needed=is_me)
+            self._crypto_power = CryptoPower(power_ups=crypto_power_ups)
         else:
-            self._crypto_power = CryptoPower(self._default_crypto_powerups,
-                                             generate_keys_if_needed=is_me)
+            self._crypto_power = CryptoPower(self._default_crypto_powerups)
         if is_me:
             self.network_middleware = network_middleware or NetworkyStuff()
             try:
@@ -326,7 +324,7 @@ class Alice(Character, PolicyAuthor):
         super().__init__(*args, **kwargs)
         PolicyAuthor.__init__(self, self.address, policy_agent=FakePolicyAgent())
 
-    def generate_kfrags(self, bob, m, n) -> List:
+    def generate_kfrags(self, bob, label, m, n) -> List:
         """
         Generates re-encryption key frags ("KFrags") and returns them.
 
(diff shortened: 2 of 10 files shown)


nkms/data_sources.py
```diff
@@ -2,18 +2,29 @@
...
 from nkms.keystore.keypairs import SigningKeypair
 from constant_sorrow.constants import NO_SIGNING_POWER
+from umbral.keys import UmbralPublicKey
+
 
 class DataSource:
 
-    def __init__(self, policy_pubkey_enc, signer=NO_SIGNING_POWER):
-        self._policy_pubkey_enc = policy_pubkey_enc
+    def __init__(self, policy_pubkey_enc, signer=NO_SIGNING_POWER, label=None):
```
cygnusv: See comment above re: label=None

nkms/policy/models.py
```diff
@@ -114,17 +114,18 @@ class Policy(object):
...
     _ursula = None
 
-    def __init__(self, alice, bob=None, kfrags=(constants.UNKNOWN_KFRAG,),
-                 label=None, m=None, alices_signature=constants.NOT_SIGNED):
+    def __init__(self, alice, label, bob=None, kfrags=(constants.UNKNOWN_KFRAG,),
+                 public_key=None, m=None, alices_signature=constants.NOT_SIGNED):
 
         """
         :param kfrags:  A list of KFrags to distribute per this Policy.
         :param label: The identity of the resource to which Bob is granted access.
```
cygnusv: Well, the label is not necessarily the "identity of the resource", it's more a "category" or similar. Of course, if the category is unique to the resource, then it can be considered an identity, but there can be many use cases where Alice defines category labels (e.g., "healthcare_information") that may have many associated resources.

nkms/characters.py
```diff
@@ -336,27 +335,27 @@ def generate_kfrags(self, bob, m, n) -> List:
...
-            uri=uri,
+            public_key=public_key,
             m=m,
         )
 
         return policy
 
-    def grant(self, bob, uri, networky_stuff,
-              m=None, n=None, expiration=None, deposit=None):
+    def grant(self, bob, uri, m=None, n=None, expiration=None, deposit=None):
```
cygnusv: uri --> label

tests/characters/test_alice_can_grant_and_revoke.py
```diff
@@ -7,19 +7,17 @@
...
 from umbral.keys import UmbralPublicKey
 import maya
 
 
-def test_grant(alice, bob, ursulas):
-    networky_stuff = MockNetworkyStuff(ursulas)
+def test_grant(alice, bob):
     policy_end_datetime = maya.now() + datetime.timedelta(days=5)
     n = 5
     uri = b"this_is_the_path_to_which_access_is_being_granted"
```
cygnusv: uri --> label

nkms/characters.py
```diff
@@ -336,27 +335,27 @@ def generate_kfrags(self, bob, m, n) -> List:
...
+        return self._crypto_power.power_ups(DelegatingPower).generate_kfrags(bob_pubkey_enc, label, m, n)
 
-    def create_policy(self, bob: "Bob", uri: bytes, m: int, n: int):
+    def create_policy(self, bob: "Bob", label: bytes, m: int, n: int):
         """
         Create a Policy to share uri with bob.
         Generates KFrags and attaches them.
         """
-        kfrags = self.generate_kfrags(bob, m, n)
+        public_key, kfrags = self.generate_kfrags(bob, label, m, n)
```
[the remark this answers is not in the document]
tuxxy: A careless user who doesn't use labels is losing their forward secrecy to an extent.

examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -56,10 +85,58 @@
...
+    # here, which can be passed via a side-channel (or posted somewhere public as
+    # testimony) and verified if desired.  In this case, the plaintext is a
+    # single passage from James Joyce's Finnegan's Wake.
+    # The matter of whether encryption makes the passage more or less readable
+    # is left to the reader to determine.
+    message_kit, _signature = data_source.encapsulate_single_message(plaintext)
 
+    # The DataSource will want to be able to be verified by Bob, so it leaves
+    # its Public Key somewhere.
+    data_source_public_key = bytes(data_source.stamp)
```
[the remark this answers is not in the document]
tuxxy: That's probably a good idea.

examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -1,25 +1,28 @@
...
+
+#########
+# Alice #
+#########
+
+ALICE = Alice(network_middleware=network_middleware)
+
 # Here are our Policy details.
 policy_end_datetime = maya.now() + datetime.timedelta(days=5)
 m = 1
```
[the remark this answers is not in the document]
tuxxy: Also, recall that I had a strong distaste for setting `m=1` in the demo, if possible.

nkms/characters.py
```diff
@@ -379,12 +378,12 @@ def grant(self, bob, uri, networky_stuff,
...
         # deposits and expirations on a limited number of Ursulas.
         # Users may decide to inject some market strategies here.
-        found_ursulas = policy.find_ursulas(networky_stuff, deposit,
+        found_ursulas = policy.find_ursulas(self.network_middleware, deposit,
                                             expiration, num_ursulas=n)
         policy.match_kfrags_to_found_ursulas(found_ursulas)
         # REST call happens here, as does population of TreasureMap.
-        policy.enact(networky_stuff)
-        policy.publish_treasure_map(networky_stuff)
+        policy.enact(self.network_middleware)
```
tuxxy: 👍

examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -1,25 +1,28 @@
...
+# This is already running in another process.
 URSULA = Ursula.from_rest_url(NetworkyStuff(), address="localhost", port=3601)
 network_middleware = SandboxNetworkyStuff([URSULA])
 
+
+#########
+# Alice #
+#########
+
+ALICE = Alice(network_middleware=network_middleware)
```
tuxxy: All-purpose Lightweight Individual Carrying Equipment? 😁

examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -32,17 +35,43 @@
...
                      expiration=policy_end_datetime)
-hrac, treasure_map = policy.hrac(), policy.treasure_map
 
-# Bob can re-assemble the hrac himself with knowledge he already has.
-hrac = keccak_digest(bytes(ALICE.stamp) + bytes(BOB.stamp) + label)
-BOB.join_policy(ALICE, hrac, node_list=[("localhost", 3601)])
-
-# Now, Alice and Bob are ready for some throughput.
+# Alice puts her public key somewhere for Bob to find later...
+alices_pubkey_saved_for_posterity = bytes(ALICE.stamp)
```
tuxxy: Is stamp starting to break down here?
We're admitting this is a public key, but not the type of public key. Stamp doesn't help make this more clear, imho.

examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -56,10 +85,58 @@
...
+    # First we make a DataSource for this policy.
+    data_source = DataSource(policy_pubkey_enc=policy.public_key)
+
+    # Here's how we generate a MessageKit for the Policy.  We also get a signature
+    # here, which can be passed via a side-channel (or posted somewhere public as
+    # testimony) and verified if desired.  In this case, the plaintext is a
+    # single passage from James Joyce's Finnegan's Wake.
+    # The matter of whether encryption makes the passage more or less readable
+    # is left to the reader to determine.
+    message_kit, _signature = data_source.encapsulate_single_message(plaintext)
```
tuxxy: Why can't we add the signature to the kit?

nkms/crypto/powers.py
```diff
@@ -75,10 +75,16 @@ def __init__(self,
...
+            # UmbralPublicKey if they provided such a thing.
             if pubkey:
-                key_to_pass_to_keypair = pubkey
+                try:
+                    key_to_pass_to_keypair = pubkey.as_umbral_pubkey()
+                except AttributeError:
+                    try:
+                        key_to_pass_to_keypair = UmbralPublicKey.from_bytes(pubkey)
+                    except TypeError:
+                        key_to_pass_to_keypair = pubkey
```
KPrasch: Phew! ... so what type do we expect they passed on this line?

nkms/characters.py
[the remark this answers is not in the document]
cygnusv: Labels, apart from being a great way to enforce per-category policies for data sharing, can also be seen as a mechanism to mitigate the impact of collusion attacks. In the event of a collusion of Bob and the Ursulas involved in a policy, only the labels in that policy would be affected.

examples/alicebob-grant-and-line-by-line-PRE.py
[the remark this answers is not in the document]
jMyles: Agreed - see #250.

examples/alicebob-grant-and-line-by-line-PRE.py
jMyles: I'm not sure I understand what you mean by "not the type of public key".  `SignatureStamp` readily casts to bytes when concatenated.  In this case I just wanted to show that it is passed as arbitrary bytes hither and yon.

I mean, what might we do instead?  `bytes(ALICE.public_keys(SigningPower))`?  That's worse, no?

examples/alicebob-grant-and-line-by-line-PRE.py
```diff
@@ -32,17 +35,43 @@
...
-# Bob can re-assemble the hrac himself with knowledge he already has.
-hrac = keccak_digest(bytes(ALICE.stamp) + bytes(BOB.stamp) + label)
-BOB.join_policy(ALICE, hrac, node_list=[("localhost", 3601)])
-
-# Now, Alice and Bob are ready for some throughput.
+# Alice puts her public key somewhere for Bob to find later...
+alices_pubkey_saved_for_posterity = bytes(ALICE.stamp)
+
+# ...and then disappears from the internet.
+del ALICE
```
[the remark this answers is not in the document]
jMyles: Well, really "-1", but who's counting?

examples/alicebob-grant-and-line-by-line-PRE.py
jMyles: We do - this is just showing that you have a place to grab it if you want to do something else with it (such as use it for public verifiability or something).  That's why I began it with an underscore.

examples/alicebob-grant-and-line-by-line-PRE.py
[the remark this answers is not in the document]
jMyles: ...I'm not sure I get this.  The `DataSource` doesn't have an `EncryptingKeypair`.  In what scenario do you imagine encrypting *for* the `DataSource`?  

It has only one public key.

In scenarios where more than one public key is implied with the phrase "public key", I agree with you.  But this isn't one of those.

nkms/characters.py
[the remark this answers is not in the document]
jMyles: I think @michwill's notion is to have a "chimney-less" paradigm (for simplicity's sake) for those AliceBobs who don't have a need for a label.

But I'm not sure it really simplifies anything anyway.

examples/alicebob-grant-and-line-by-line-PRE.py
tuxxy: Is it not feasible to just have the user do `message_kit.signature` or something?

examples/alicebob-grant-and-line-by-line-PRE.py
tuxxy: It's more verbose, but ten-fold more explicit on which key it is. I don't think this is an either/or situation here. There should probably be a more explicit, yet simple method of doing this.

examples/alicebob-grant-and-line-by-line-PRE.py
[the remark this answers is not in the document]
cygnusv: OK, _we_ know that `DataSource` doesn't have an `EncryptingKeypair`, but this may not be immediate for implementers that will rely on this kind of examples to understand the whole thing. For example, not so long ago, even _we_ only thought of Alice and there were no `DataSources`

examples/alicebob-grant-and-line-by-line-PRE.py
michwill: I would agree that it took me a few seconds to understand that `stamp` is actually a public key. Although it's nice and short (and a verb!), it's not in line with common intuition

examples/alicebob-grant-and-line-by-line-PRE.py
jMyles: Yeah, good question.  Currently, we don't have a facility for setting `_signature` on a `MessageKit`, and it's actually not trivial to think of a solid, cognizable workflow for doing that.  

I can simply set it in `encapsulate_single_message` if you think that makes sense (instead of passing it back).

Be reminded that, short of that, the signature is encrypted as part of the ciphertext and is not otherwise available.  

I think this kinda needs to be part of our overall discussion re: kits.  But yeah, for now, if you want me to just set the attr and move on, I think that makes good enough sense.

examples/alicebob-grant-and-line-by-line-PRE.py
jMyles: Yeah, in isolation, I can definitely see how `bytes(ALICE.public_keys(SigningPower))` adds clarity.  But don't we *want* the user to pass Alice's stamp for these sorts of purposes?  It only looks strange here because it's a one-off in a demo.  But if they had a function:

```
def sign_something_and_return_signature_and_public_key(something, signer):
...
```

...then obviously we want to train them that Alice's `stamp` can be used as both a signer and a public key, right?

Maybe this isn't the right moment to teach that, in which case I'm fine switching to `bytes(ALICE.public_keys(SigningPower))`.  However, we might also be able to make it a good teachable moment with a quick additional comment.

nkms/characters.py
jMyles: Yeah, rolls nicely right?  I think we're probably go to adopt this everywhere.

nkms/policy/models.py
jMyles: Oh yeah, absolutely.  That language is a holdover from `path` (and then `uri`).  

There's no question that the docstrings through `policy` need a refresher at this point.

examples/alicebob-grant-and-line-by-line-PRE.py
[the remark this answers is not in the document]
jMyles: OK, true.

So, to what shall I change this name?

nkms/crypto/powers.py
```diff
@@ -113,8 +126,23 @@ class EncryptingPower(KeyPairBasedPower):
...
+        """
+        Generates re-encryption key frags ("KFrags") and returns them.
+
+        These KFrags can be used by Ursula to re-encrypt a Capsule for Bob so
+        that he can activate the Capsule.
+        :param bob_pubkey_enc: Bob's public key
+        :param m: Minimum number of KFrags needed to rebuild ciphertext
+        :param n: Total number of rekey shares to generate
+        """
+        # TODO: salt?
```
michwill: Btw if we have application-specific salts, empty labels may start making sense.. But perhaps we can think of it after the testnet release

examples/alicebob-grant-and-line-by-line-PRE.py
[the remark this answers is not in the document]
cygnusv: This seems related to the `bytes(Alice.stamp)` before. I'm all in for educating the user into our conventions, but if has to be done carefully. We can leave it just as `data_source_public_key` if we make it clear that it is a signature public key.

examples/alicebob-grant-and-line-by-line-PRE.py
arjunhassard: I'd lean more towards `bytes(ALICE.public_keys(SigningPower))` because it is not only avoids confusion with regard to the type of public key is at play here, it also introduces newbies to the general concept of powers (SigningPower, CrypoPower, etc) with a relatively straightforward example.

Also, we can always use a comment to explain what @jMyles is keen on teaching here – that Alice's `stamp` can be used as both a signer and a public key.

nkms/characters.py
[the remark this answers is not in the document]
arjunhassard: ^ Thanks, useful discussion – whether we allow empty labels or not (there may be an edge case where a developer wants to do that), I'll add a note of the consequences (forward-secrecy loss, greater chance of collusion) to the Jupyter Notebook.

PR #251, opened 2018-04-19
Extract solidity dependencies + Tweak compiler and Registrar

KPrasch:
Helps prepare to relocate these dependencies to separate repositories if we so choose.

- Update compiler class to solc 4.22
- Fix default path generation
- Extract `proxy` and `zeppelin` and remap imports at compile time
- Updates solidity imports
- Update pytest fixture for new path generation


KPrasch:
Yes, this works for me.

KPrasch:
Rebased over @szotov's changes.

PR #255, opened 2018-04-21
Fixes for solidity and web3 tests

szotov:
All contracts tests pass under one condition:
Max gas limit should be increased up to 4626271 of gas if gas estimate buffer is 1.5.
Eth tester has no api for that (or I missed it), so for quick fix may change GENESIS_GAS_LIMIT or GAS_ESTIMATE_BUFFER in eth_tester\backends\pyevm\main.py


KPrasch:
:mountain_cableway: - Good updates!

PR #258, opened 2018-04-23
Upgrade web3 contract method usage

KPrasch:
Depends on #257 

Resolves deprecation warnings described here in the *contract* test suite: http://web3py.readthedocs.io/en/stable/contracts.html#web3.contract.Contract.eventFilter

- deprecates `contract.eventFilter` in favor of `contract.events.Event.createFilter`
- deprecates `contract.transact` in favor of `contract.functions.method.transact`
- deprecates `contract.call` in favor of `contract.functions.method.call`

PR #260, opened 2018-04-24
Custom private chain genesis parameters 

KPrasch:
Depends on PR #258 and PR #257 

- `PyEVMBackend` testing subclass, allowing a dictionary of overrides to be passed;
- Increases the genesis block's gas limit from eth_tester's default `3141592` to `4626271`
- Sets `pyevm` as the default testing backed


nkms/blockchain/eth/chains.py
```diff
@@ -119,17 +120,16 @@ def spawn_miners(self, miner_agent, addresses: list, locktime: int, random_amoun
...
+        tx_hashes = list()
         for address in addresses:
-            tx = {'to': address,
-                  'from': coinbase,
-                  'value': amount}
-            _txhash = self.provider.w3.eth.sendTransaction(tx)
-
-
+            tx = {'to': address, 'from': coinbase, 'value': amount}
+            txhash = self.provider.w3.eth.sendTransaction(tx)
```
michwill: Do we ever wait for receipt? Sounds like a sensible thing to do

PR #261, opened 2018-04-26
Step-wise unlock + policy first period

szotov:
* Optimization for MinersEscrow by reducing stored confirmed periods
* Payment for the first/current period when creating a policy
* Step-wise unlock


tuxxy:
🚀 This is awesome

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
```diff
@@ -63,27 +69,27 @@ contract MinersEscrow is Issuer {
...
+        uint256 confirmedPeriod2;
         // downtime
         uint256 lastActivePeriod;
         Downtime[] downtime;
         bytes32[] minerIds;
     }
 
-    uint256 constant MAX_PERIODS = 3;
+    uint256 constant EMPTY_CONFIRMED_PERIOD = 0;
     uint256 constant MAX_OWNERS = 50000;
```
michwill: Why do we limit `MAX_OWNERS`? In principle, minimal amount locked is limiting max_owners anyway

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
```diff
@@ -305,48 +228,64 @@ contract MinersEscrow is Issuer {
...
-        LockSwitched(msg.sender, info.release);
+        for (uint256 index = 0; index < info.stakes.length; index++) {
+            StakeInfo storage stake = info.stakes[index];
+            if (stake.lockedValue == _oldValue &&
+                stake.lastPeriod == _lastPeriod) {
+                break;
+            }
+        }
+        // TODO lastPeriod can be equal current period but need to call confirmActivity
+        require(index < info.stakes.length && stake.lastPeriod >= getCurrentPeriod().add(uint256(1)));
```
michwill: What is `index` equal to if we go through all the cycle without ever satisfying the `if` condition? Can we accidentally get a situation where wrong inputs is given for `_oldValue`, but still the stake which is selected has an index `info.stakes.length - 1`, instead of erroring out

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
```diff
@@ -356,46 +295,50 @@ contract MinersEscrow is Issuer {
     function withdraw(uint256 _value) public onlyTokenOwner {
         MinerInfo storage info = minerInfo[msg.sender];
         // TODO optimize
-        uint256 lockedTokens = Math.max256(calculateLockedTokens(msg.sender, 1),
-            getLockedTokens(msg.sender));
+        uint256 lockedTokens = Math.max256(getLockedTokens(msg.sender, 1),
```
michwill: Eh.. why so?

michwill: Is it in case we call after or before confirmActivity?

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
```diff
@@ -423,53 +366,76 @@ contract MinersEscrow is Issuer {
...
 
-        uint256 currentLockedValue = getLockedTokens(msg.sender);
-        ConfirmedPeriodInfo storage last = info.confirmedPeriods[numberPeriodsForMinting - 1];
-        uint256 allLockedPeriods = last.lockedValue
-            .divCeil(info.releaseRate)
-            .sub(uint(1))
-            .add(numberPeriodsForMinting);
+        uint256 first;
+        uint256 last;
+        if (info.confirmedPeriod1 > info.confirmedPeriod2) {
```
michwill: This optimization (with `confirmedPeriod1` / `confirmedPeriod2`) needs a comment. I am not 100% sure I understand how it works!

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
```diff
@@ -63,27 +69,27 @@ contract MinersEscrow is Issuer {
...
+        uint256 confirmedPeriod1;
+        uint256 confirmedPeriod2;
         // downtime
         uint256 lastActivePeriod;
         Downtime[] downtime;
         bytes32[] minerIds;
     }
 
-    uint256 constant MAX_PERIODS = 3;
+    uint256 constant EMPTY_CONFIRMED_PERIOD = 0;
```
michwill: `EMPTY_CONFIRMED_PERIOD` also requires some comment. What is that doing exactly?

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
```diff
@@ -570,20 +519,18 @@ contract MinersEscrow is Issuer {
...
+        } else if (_field == MinerInfoField.StakeFirstPeriod) {
+            return bytes32(info.stakes[_index].firstPeriod);
+        } else if (_field == MinerInfoField.StakeLastPeriod) {
+            return bytes32(info.stakes[_index].lastPeriod);
+        } else if (_field == MinerInfoField.StakeLockedValue) {
+            return bytes32(info.stakes[_index].lockedValue);
+        } else if (_field == MinerInfoField.ConfirmedPeriod1) {
+            return bytes32(info.confirmedPeriod1);
+        } else if (_field == MinerInfoField.ConfirmedPeriod2) {
+            return bytes32(info.confirmedPeriod2);
```
michwill: Btw could it be possible to return the whole thing with new version of solidity?

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
michwill: Ok, `require(index < info.stakes.length)` will raise if the cycle has ended, great

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
szotov: In most cases in current period will have more locked tokens than all future periods. But if Ursula will increase stake (deposit or lock methods) than in next period will have maximum locked tokens

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
szotov: About small stakes - old and new values are checked with minAllowableLockedTokens

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
szotov: Right, I'll remove it

nkms/blockchain/eth/sol/source/contracts/MinersEscrow.sol
szotov: I am trying to simplify it with changes from Byzantium fork

PR #268, opened 2018-05-08
Renaming NuCypher KMS to NuCypher

### LINKED ISSUE #263: Does using the word "KMS" confer any advantage to this project?
Since the beginning of this project, we have called it `nucypher-kms`.  But does the "KMS" part of the name add anything useful?  Are we in fact building a "KMS"?  Is it even possible, given the current use of this word, to determine with clarity exactly what a "KMS" is?

----

As this project has progressed, I think it has become clear that what we're building isn't primarily a "Key Management System."

First, let's consider exactly what a "Key Management System" is, and whether this phrase is unambiguous enough for us to use as part of the name of this project (hint: I think it isn't).  Let's also consider whether this phrase is sufficiently widespread to give this project any name-recognition advantage by using it.

### What is a KMS?

The [Wikipedia disambiguation page for "KMS"](https://en.wikipedia.org/wiki/KMS) lists, under the "Technology" heading:
```
    Kernel mode-setting, setting up screen mode and resolution in operating system kernels
        KMS driver, Linux kernel device driver for display controllers
    KMS (hypertext), the Knowledge Management System hypertext system
    Kaspersky Mobile Security
    Key Management Service, a Microsoft technology
    Kerio MailServer
    Kwangmyŏngsŏng (disambiguation), a class of experimental satellites developed by North Korea
```
None of these have anything to do with cryptographic tooling.  Even the MS "Key Management Service" is not about cryptographic keys - it's about Microsoft license keys. 

*In other words, there is literally nothing on the Wikipedia page for the word "KMS" that has anything to do with this project.*  I think that says something.  I think this alone is a strong indication that using the word "KMS" will create confusion rather than relieving it.

But let's keep reading to see if "KMS" makes sense when examining its actual uses.  Hopping over to the [Wikipedia page for "Key Management"](https://en.wikipedia.org/wiki/Key_management).  Here, we find, under a subheading, "Key management system":

> A key management system (KMS), also known as a cryptographic key management system (CKMS), is an integrated approach for generating, distributing and managing cryptographic keys for devices and applications. They may cover all aspects of security - from the secure generation of keys over the secure exchange of keys up to secure key handling and storage on the client. Thus, a KMS includes the backend functionality for key generation, distribution, and replacement as well as the client functionality for injecting keys, storing and managing keys on devices.

So this description is closer obviously.  It does cover some of the basic mechanics that drive this project.  But does it actually describe anything user-facing?

This project does indeed feature "integrated approach for generating, distributing and managing cryptographic keys."  But is that its primary purpose?  Is that what makes it different or useful?  Neither Alice, Bob, nor Ursula ever even become "aware" that keys exist.  They never need to handle keys to *do* anything with them.  

The user here has far more interesting interfaces to work with than "Keys" - they get fully built nodes, Policies, DataSources, and MessageKits.  Rather than a crude "KMS", what we have here is a feature-rich network for doing types of communication that aren't even possible in the context of a "KMS" alone.

Nor does this description describe features like the payment / medallion model, DisappearingAlice, or the capsule / label system.  These are the things that make this project different and interesting, and none of them are captured by this (or any) description of a "KMS".

OK, but maybe WikiPedia just isn't a good source of textual definition for this concept.  Let's instead look at how this phrase is used in the wild.

### Is this thing similar to AWS KMS?

Googling "Key management system" brings, as the first result, AWS KMS.  Now, first, note something important here: Amazon uses the phrase "Key Management Service (KMS)," not "Key Management System."  So it seems immediately that there's a disparity about what "KMS" even means: is it a *system* or a *service*?

Amazon's description of their Key Management Serivce: 

> AWS Key Management Service (KMS) is a managed service that makes it easy for you to create and control the encryption keys used to encrypt your data, and uses FIPS 140-2 validated hardware security modules to protect the security of your keys.

Is that what you'd say about this project?  I don't think so.  If I had to write a description of this project in exactly the same tembre as this one, I'd say something like:

> NuCypher is a decentralized service that makes is possible for you to define recipients for private data and then to build (or encourage others to build) applications that share data with them.  To do this, it uses a distributed swarm of nodes which perform proxy re-encryption, and which never have access to your private data.

### What about other KMS's?

Back to Google - the [second result is the Microsoft Technet article on "Understanding KMS."](https://technet.microsoft.com/en-us/library/ff793434.aspx#E1B)  It reads:

> KMS activates computers on a local network, eliminating the need for individual computers to connect to Microsoft. To do this, KMS uses a client–server topology. 

In other words, the *second result on Google for "KMS"* is about a *completely* unrelated concept.  I'm left with the feeling that using the phrase "KMS" *adds* confusion to our project rather than subtracting it.

### What do we say instead?

So, in a nutshell, what I'm trying to say is:

**If we give the impression that NuCypher is a "KMS", it might not be clear what that even means.  In the best-case scenario, we hope that it evokes an image of a drop-in replacement for things like Amazon KMS, but even in this case, we're both over- and underselling it.  This comparison misses the *main things* that make NuCypher different and compelling, and fails to encourage people to imagine the best use cases.**

Furthermore, "KMS" is not a phrase that is of such household-name status and disambiguity as to gain us anything interesting by trying to leverage it.  

**So, I propose that we call this thing simply "NuCypher."**

Now, I understand that the word "NuCypher" is currently a name given to the organization, and that under its umbrella lie multiple products with different use-cases.  But now might be time to change that, just as we have done with our website.  The website certainly gives the impression that this project is the entire kit and caboodle.  In fact, our website features the word "KMS" only once: 

> NuCypher KMS helps dApp developers secure their data on the blockchain with decentralized proxy re-encryption as a service.

And of course, this reads equally well if the word "KMS" is simply removed:

> NuCypher helps dApp developers secure their data on the blockchain with decentralized proxy re-encryption as a service.

jMyles:
For the purposes of this refactor, I used (and propose that we use, going forward) the following spelling and case conventions:

* NuCypher - the organization, brand, and network.  (uppercase "C")
* `Nucypher` - a prefix for classes of things in the NuCypher domain.  (lowercase "c")
* nucypher - the codebase.
* `nucypher` - the python app inside the codebase.
* NU - the symbol of the ERC20 token described in the nucypher codebase.
* NuCypherToken - the name of the ERC20 token with the symbol "NU"
* `NucypherToken` - a class and a prefix for classes of things pertaining to the NuCypherToken

Fixes #263.

If this PR is merged, I will also endeavor to change "NuCypher KMS" to "NuCypher" in places where I am able to do that, such as the name of this repository and our Discord server.


jMyles:
> Let's ask the question; why was this product ever named NuCypher KMS, and not just NuCypher? Well, it's a classic 'anchor-and-twist' approach to explaining our  [...] Later on in #263, a list of open source Non-KMIP compliant key management projects came up. It's true, none of them use 'KMS'.  But notice that a lot of them have named their systems/products in such a way that you immediately get a good idea of their purpose/relevance

All but two on this list use the word "key" as their "good idea of their purpose" - is that any more of a touchstone than "..Cypher"?  The other two use the words "vault" (good) or "privacy" (vague).  

I'm not sure that I can see a discernible difference in anchorable material between any of these names and "NuCypher".  I mean... KeyAppliance?

Instead of being a tractor beam in an anchor-and-twist scheme, I want to suggest that "KMS" is a repulsor beam.  It makes it harder to become conceptually anchored instead of easier.

> A separate issue with using 'NuCypher' is that it's hierarchically confusing – we have 'Umbral', which is a NuCypher scheme, and then just 'NuCypher', which is also a NuCypher scheme

"Umbral" is in no sense a NuCypher scheme.  It is instead a proxy reencryption scheme, replete with a reference implementation and whitepaper, developed at NuCypher for the purposes of advancing the NuCypher network.

Nor is "NuCypher" a "NuCypher scheme."  NuCypher is an organization and network focused on specific features which empower privacy in decentralized systems.  Our codebase (which we call "nucypher") reflects those values and is an implementation of that vision.  

> Plus, this problem will be compounded when we roll out more schemes – say we decide to build an even simpler, higher-level touching point than the current KMS – what would we call that? 

This has the appearance of a good argument, and it's what kept me from opening this issue for so long.  But the truth is: we're *already* using the word "NuCypher" to mean what it will mean if this PR is merged (see comments from both @tuxxy and @mswilkison, describing their habit of doing this to completely different audiences), so that ship has already sailed.  Look at our website - it already uses the word "NuCypher" this way in all but one instance.  

Additionally, I actually think that, in the medium- and long-term, this is a source of strength rather than weakness.  We can either choose to leverage the NuCypher name (with things like `nucypher-health` or `nucypher-fileshare`) or we can choose to deliver new codebases using only the organization's name, like nucypher/searchSafe.  (Obviously I've just made up these three names).

Organizations of all sizes already do this.  A huge, corporate example: Google is a search engine, but all of their other products are simply names "Google <blank>".  A small, nonprofit example: requests is a hugely popular HTTPS framework, but in addition to requests/requests, its github also has many other "branded" projects like requests/oathlib and requests/toolbelt.

>  It may be inaccurate/inappropriate/boring, but we need to replace it with something that is also well-known.

Even if this were true (and again, I think it's not, for the reasons I've just outlined), as a practical matter it's almost certainly too late to effectively do this.  I think we're running out of time to even remove "KMS", but trying to re-educate the public on a totally different term threatens to be a sink of our internal energy and of our audience's attention span.

I don't think "NuCypher" is a perfect name; I'll wager a guess that nobody does.  But it's a pretty good name, and used with discipline, has the power to play all the roles that I've outlined in the description of this PR.  And even if you want to later change it to something else, I still think it behooves us to move forward with a merge, as this is a step in a positive direction, even if not quite a big enough one to satisfy all of your concerns.

arjunhassard:
My suggested course of action is that we quickly brainstorm a replacement to 'KMS'. Here are some initial ideas: 

- **NuCypher Access Management** 
- **NuCypher Permission Management** -> **NPM**
- **NuCypher Identity & Access Management** -> **NC-IAM**
- **NuCypher Secure Sharing** 
- **NuCypher Private Data Management** 
- **NuCypher Access Control** 
- **NuCypher Privacy Layer** 

Btw, I think 'identity' is valid since we enable signing, sender verification etc. You could also add 'Decentralized' to any of the above – e.g. 'NuCypher Decentralized IAM' 

I think the assertion that it's too late to effectively replace KMS with another word is unfounded. @cygnusv mentioned that any change will come across as erratic – I think there's truth to that – plus we'll need to change things like our white paper – but from my perspective, talking to users, it would actually be more difficult to handle switching from NuCypher KMS -> NuCypher than it would be from NuCypher KMS -> NuCypher IAM/PM/SS/etc, because announcing the latter is more concrete, clean and quickly understandable. 

Some responses to your response: 

> All but two on this list use the word "key" as their "good idea of their purpose" - is that any more of a touchstone than "..Cypher"? 

For some reason, you've ignored the descriptive, unambiguous **product** names in my list: PKIAppliance, EC2Box, SSH Key Base, Public Key Share. You've also ignored the AWS products – which are relevant, since we often use AWS's security services as a reference point.  

I agree that 'Key' is vaguer than 'PKI', but it's far less vague than a portmanteau containing the less common spelling of cipher!

> Instead of being a tractor beam in an anchor-and-twist scheme, I want to suggest that "KMS" is a repulsor beam. It makes it harder to become conceptually anchored instead of easier.

Totally agree with this. I began my comment by with some additional reasons for ditching 'KMS'. 

> "Umbral" is in no sense a NuCypher scheme. It is instead a proxy reencryption scheme, replete with a reference implementation and whitepaper, developed at NuCypher for the purposes of advancing the NuCypher network.

"in no sense"? - what about in the sense that it was developed by NuCypher, for NuCypher, as you say right after? You're choosing one semantic interpretation of 'NuCypher + scheme' out of many - even if it's technically the most accurate - that doesn't mean the outside world will interpret it this way. 

> Nor is "NuCypher" a "NuCypher scheme." NuCypher is an organization and network focused on specific features which empower privacy in decentralized systems. Our codebase (which we call "nucypher") reflects those values and is an implementation of that vision.

This is exactly what's confusing about using one word to describe multiple, distinctive things. Uncapitalizing the 'n' isn't enough of a difference between names for a codebase and a company. 

>  Look at our website - it already uses the word "NuCypher" this way in all but one instance.

I don't know what you mean by this. Every example of NuCypher on the nucypher.com/blockchain page is followed by 'KMS'.  

> A huge, corporate example: Google is a search engine, but all of their other products are simply names "Google "

It's official name is Google Search. Inside the company, you work for the 'Search' team. Of course normal people say 'I googled it', but that's a totally different audience, not developers looking for specificity. 

> requests is a hugely popular HTTPS framework, but in addition to requests/requests, its github also has many other "branded" projects like requests/oathlib and requests/toolbelt.

Again, this example seems to undermine your argument. The fact they added the second 'requests', i.e. it's requests/**requests**, not just 'requests' on its own, shows that they delineated their core service from secondary offerings. If they had followed what you're proposing, their product suite would be something like: product 1 = requests, product 2 = requests/oathlib. We should copy them and go with nucypher/iam, nucypher/fileshare, etc. 

Also, they used 'requests' twice, instead of a new word, because 'requests' describes what they do far better than 'NuCypher' describes decentralized private data management and access control. This is their one-line description: "Requests allows you to send organic, grass-fed HTTP/1.1 requests". Simple!

tuxxy:
NuCypher does not do identity anything _yet_. It can be made to do it and it can also be leveraged to build identity and access management apps.

But Identity is not a usecase for NuCypher at this moment. Identity and what we do are very different. Identity management solves very different problems and it has its own set of pitfalls associated to it that we are not solving. Though identity has keys and key management, this is a whole different set of cryptographic narratives and problems.

I will agree that Umbral is not a NuCypher scheme. Its research and development is being funded by NuCypher to be _used_ by NuCypher, but it is not a _NuCypher scheme_.

I have grown on simply calling "the KMS" NuCypher. NuCypher can be leveraged as a KMS, but we empower privacy for all decentralized applications.

jMyles:
Holy shit @arjunhassard, you are a good debater.  I'll try to address every single element of what you're saying here; let me know if I miss anything.

> My suggested course of action is that we quickly brainstorm a replacement to 'KMS'. Here are some initial ideas:

    NuCypher Access Management
    NuCypher Permission Management -> NPM
    NuCypher Identity & Access Management -> NC-IAM
    NuCypher Secure Sharing
    NuCypher Private Data Management
    NuCypher Access Control
    NuCypher Privacy Layer

These are all reasonably good names (some more than others - probably the first one is best, and the "identity" one is a little shaky).  But ask yourself:  What does *it mean* that you can, within a fair playing field of reasonableness, apply any of these names (and probably others)?

Here's what I think it says:

**We only know *some* of the things NuCypher will be used for, and there's a good chance that the best things are of a form that we haven't even considered and that isn't on this list**.

> Btw, I think 'identity' is valid since we enable signing, sender verification etc.

Oooch, I disagree vehemently on this point.  Identity is a political and social function, and a crucially important one.  Signing can be done by things of every level of sovereignty, and mistaking the power to sign for the power to command identity stems, at least in my estimation, from one of the components of society that I hope we combat: that power (in this case, to sign) is somehow equivalent to righteousness (in this case, the righteousness of preserving and being proud of one's identity).  

Identity is *hard* and while, as @tuxxy points out, we have a bright future for empowering identity, it's not something we do yet.

> I think the assertion that it's too late to effectively replace KMS with another word is unfounded. @cygnusv mentioned that any change will come across as erratic – I think there's truth to that – plus we'll need to change things like our white paper – but from my perspective, talking to users, it would actually be more difficult to handle switching from NuCypher KMS -> NuCypher than it would be from NuCypher KMS -> NuCypher IAM/PM/SS/etc, because announcing the latter is more concrete, clean and quickly understandable.

Interesting.  

> For some reason, you've ignored the descriptive, unambiguous product names in my list: PKIAppliance, EC2Box, SSH Key Base, Public Key Share.

So I think that all four of those names are *awful*.  I have no idea what they do, and I doubt I'd remember any of them for long, even if I was using them.  

> You've also ignored the AWS products – which are relevant, since we often use AWS's security services as a reference point.

michwill:
I think, the Google point is a pretty strong one. People will tend to call the thing by the brand name, no matter what's attached to the name.

My thought is this. If there is a class of products, and a company tries to borrow credibility and descriptional power from this class, one can attach the class name. But if you want your brand to spread everywhere, it's probably better to keep it just as a brand name.

Also I'd say that Amazon names such as "IAM", "PM", "SS" etc are even less descriptive than KMS (to me).

Interestingly enough, no one in community seems to be supporting Bluzelle's SwarmDB renaming and everyone still calls the product Bluzelle. More than that, there appeared to be a clash with SwarmDB made by Wolk (where they refer to Ethereum Swarm as something they are building upon!).

So I think, `nucypher` is a good name, and it at least tells that "it's about encryption". So I am all for the renaming in this PR.

jMyles:
> The reason I dug my heels in a bit here (i.e. made a case against this PR) is because it's clearly a lot of work editing 113 files.

It's true that this has been a lot of work, but the work has been conceiving the plan, doing the research, convincing the team, and coming to understand and enumerate the various ways we use our brand in our project (ie, the list in the description).

The actual edits to the codebase represent less than 10 minutes of work.  So if we find a perfect name and want to make a similar change, it is by no means unthinkable on the basis of its gravity as a clerical matter.

PR #269, opened 2018-05-09
Deprecates size constants for cfrags and kfrags - uses .get_size() instead.

jMyles:


tuxxy:
I can't review this due to the large KMS rename diff.

jMyles:
Eh?  Its just the one commit.

PR #276, opened 2018-05-21
Character + TokenActor integration and configuration

KPrasch:
Characters and Arrangements can successfully be created with an optionally injectable `NuCypherConfig` class. Blockchain "connection" is established  through configured provider backends on character creation (leading to slower tests)

What this does: 
- Working `NucypherTokenActor` ->  `Character` inheritance scheme 
- Abstract `ContractProviders` into `ContractInterface` and `DeployerInterface`
- `BlockchainConfig` defaulting (get&set) and singleton functionality 
- Cleanup names and constants

*Note:* Slow local test run times: 8s (Quick suite) and 9min (Slow suite)


KPrasch:
I still have a failing test: `test_miner_collects_staking_reward_tokens` with `TransactionFailed`.  @szotov - does this sound familiar?

tests/blockchain/eth/entities/test_actors.py
```diff
@@ -2,98 +2,95 @@
...
+        initial_balance = miner.token_balance()
+        assert mock_token_agent.get_balance(miner.address) == miner.token_balance()
 
-    # ...wait more...
-    chain.time_travel(periods=2)
-    miner.mint()
-    miner.collect_staking_reward()
+        # Have other address lock tokens
+        _origin, ursula, *everybody_else = chain.interface.w3.eth.accounts
+        mock_miner_agent.spawn_random_miners(addresses=everybody_else)
```
szotov: `ursula` (`Miner`) should deposit tokens but `spawn_random_miners(addresses=everybody_else)`  deposited only for `everybody_else`

tests/blockchain/eth/entities/test_actors.py
```diff
@@ -2,98 +2,95 @@
...
-    assert mock_token_agent.get_balance(miner.address) == miner.token_balance()
+        # Wait for it...
+        chain.time_travel(periods=1)
+        assert mock_miner_agent.contract.functions.getLockedTokens(miner.address).call() == mock_miner_agent.min_allowed_locked
 
-    # Have other address lock tokens
-    _origin, *everybody_else = chain.provider.w3.eth.accounts
-    mock_miner_agent.spawn_random_miners(addresses=everybody_else)
+    @pytest.mark.slow()
+    def test_miner_collects_staking_reward_tokens(self, chain, miner, mock_token_agent, mock_miner_agent):
```
szotov: `mock_policy_agent` in parameters is used for the policy manager initialization (`setPolicyManager`)
`mint` do not work without configured policy manager in the `MinersEscrow`

nucypher/config/configs.py
```diff
@@ -64,34 +43,113 @@ def db_path(self):
         return self.__db_path
 
 
+class BlockchainConfig:
+
+    __default_providers = (IPCProvider(ipc_path=os.path.join('/tmp/geth.ipc')),
```
michwill: Is `os.path.join` really needed on a single value? I think, this could be just `/tmp/geth.ipc`. Or, maybe more cross-platform: `os.path,join(tempfile.gettempdir(), 'geth.ipc')` (although UNIX socket wouldn't work in Windows anyway)

nucypher/config/configs.py
```diff
@@ -64,34 +43,113 @@ def db_path(self):
...
+            interface_class = ContractInterface
+        else:
+            from nucypher.blockchain.eth.interfaces import DeployerInterface
+            interface_class = DeployerInterface
+
+        interface = interface_class(blockchain_config=self, sol_compiler=compiler, registrar=registrar)
+
+        if geth is True:
+            w3 = interface.w3
+            w3.middleware_stack.inject(geth_poa_middleware, layer=0)
```
michwill: Well, it's POA until it's not. I can certainly see this class being used with real blockchain (even on testnet!) where it's not POA anymore

tests/eth_fixtures.py
```diff
@@ -123,54 +124,59 @@ def solidity_compiler():
...
 
 
 @pytest.fixture(scope='module')
-def registrar():
-    _, filepath = tempfile.mkstemp()
-    registrar = Registrar(chain_name='tester', registrar_filepath=filepath)
-    yield registrar
-    os.remove(filepath)
+def web3(deployer_interface):
+    """Compadibility fixture"""
```
michwill: Uh. Typo

nucypher/config/configs.py
KPrasch: Agreed, this either needs to move or be renamed.

tests/eth_fixtures.py
KPrasch: Whoops!

tests/blockchain/eth/entities/test_actors.py
KPrasch: :heavy_check_mark: Fixed

nucypher/blockchain/eth/constants.py
```diff
@@ -7,19 +7,15 @@ class TokenConfigError(ValueError):
         pass
 
     __subdigits = 18
-    _M = 10 ** __subdigits                                 # Unit designation
-    __initial_supply = int(1e9) * _M                       # Initial token supply
-    __saturation = int(3.89e9) * _M                        # Token supply cap
+    M = 10 ** __subdigits                                 # Unit designation
+    __initial_supply = int(1e9) * M                       # Initial token supply
+    __saturation = int(3.89e9) * M                        # Token supply cap
```
michwill: Eh.. do these parameters also have to be with underscores?

KPrasch: No; however, they are only used internally for constant calculations.

michwill: Ah, makes sense then.
How do we _set_ those values though? I can expect them to be different on testnet vs mainnet

KPrasch: Seems like these values won't ever be set during runtime..right? 
I've always felt that this module is at least somewhat configuration (in fact this class had the name `NuCypherMinerConfig` for some time).

Perhaps default values can be set and optionally used in the constant classes's `__init__` like we do with most injectable configs.  

Maybe we make subclasses if there are a few unchanging sets of constants we will use.

KPrasch: This is categorically different then every other config class because Alice, Ursula, and Bob don't set/change the values for typical (and most atypical) scenarios I can imagine.  Once a network is deployed with one set of values, those values are then *required* to be compatible with the rest of the network.  Still, if we intend to change these values programmatically, its best to follow our existing dependency injection scheme.

michwill: I am just saying that these values will be different on mainnet and testnet, possibly. Maybe it can be hardcoded for two of them, but what you select in configuration is either one preset or the other

KPrasch: Got it.  Yeah - I think both subclassing *and* using default values will work well and remain extensible. :-)

.travis.yml
```diff
@@ -3,6 +3,7 @@ cache: pip
 python:
 - '3.5'
 - '3.6'
+- '3.7-dev'
```
tuxxy: 👍 I can start using 3.7, if needed.

nucypher/blockchain/eth/actors.py
```diff
@@ -1,17 +1,19 @@
...
 from collections import OrderedDict
 from datetime import datetime
 from typing import Tuple, List, Union
 
-from nucypher.blockchain.eth.agents import NucypherTokenAgent
+from nucypher.blockchain.eth.agents import NucypherTokenAgent, MinerAgent, PolicyAgent
 
 
-class TokenActor(ABC):
+class NucypherTokenActor:
```
tuxxy: 👍

nucypher/blockchain/eth/chains.py
```diff
@@ -1,88 +1,74 @@
...
-    """
-    http://populus.readthedocs.io/en/latest/config.html#chains
-
-    mainnet: Connects to the public ethereum mainnet via geth.
-    ropsten: Connects to the public ethereum ropsten testnet via geth.
-    tester: Uses an ephemeral in-memory chain backed by pyethereum.
-    testrpc: Uses an ephemeral in-memory chain backed by pyethereum.
-    temp: Local private chain whos data directory is removed when the chain is shutdown. Runs via geth.
-    """
+class Blockchain:
```
tuxxy: I like this name much more.

nucypher/blockchain/eth/chains.py
```diff
@@ -1,88 +1,74 @@
...
+    """A view of a blockchain through a provided interface"""
```
tuxxy: Do we have docs/docstrings on the previous types of testnets we can connect to and how to connect to them? Since this was removed, I want to make sure it's really obvious about _which_ network we're connecting to, especially by default.

nucypher/blockchain/eth/constants.py
tuxxy: Re: the variable names, perhaps add some comments explaining why they will never be changed during runtime?

.travis.yml
KPrasch: it's release candidate will be published on the 30th.  We need a 3.7 wheel, for starters.

nucypher/blockchain/eth/chains.py
KPrasch: Yes - this may need a bigger docstring;  `network_name` was moved over to `BlockchainConfig`.

nucypher/blockchain/eth/chains.py
michwill: Currently - I doubt it. In future - perhaps other blockchains? Anyway, too early to think that.

I cannot imagine testnet and mainnet being used at the same time

jMyles: Right.  So, it's clearly singular in this runtime.  And since, unlike most singletons, there exists an authoritative, universal mechanism for describing its contents, I actually think that the word "The" gives meaningful insight to the user in the phrase, `TheBlockchain`.  It's not "A Blockchain" - it's a representation of *The* Blockchain.  A separate process will still need to regard the same details as canonical, given connection to the same blockchain.  There is no way for this object to conceptualize "A blockchain" without also being "the blockchain."

nucypher/blockchain/eth/actors.py
```diff
@@ -219,12 +216,19 @@ def fetch_data(self) -> tuple:
...
     """Alice"""
 
-    def __init__(self, address: bytes, policy_agent):
-        super().__init__(token_agent=policy_agent._token, address=address)
+    def __init__(self, policy_agent: PolicyAgent=None, *args, **kwargs):
+
+        if policy_agent is None:
+            # all defaults
+            token_agent = NucypherTokenAgent()
+            miner_agent = MinerAgent(token_agent=token_agent)
```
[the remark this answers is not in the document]
KPrasch: This ensures actor classes can be instantiated even without manually providing an agent.

PR #280, opened 2018-06-01
[WIP] Ursula checks KFrag validity before enacting Arrangement.

jMyles:


jMyles:
Going a slightly different direction for the moment.

PR #285, opened 2018-06-04
Reversing inheritance of Arrangement -=> BlockchainArrangement; nixin…

jMyles:
…g deposit from the former and keeping it only in the latter.


KPrasch:
Changes included upstream: See PR #288

PR #291, opened 2018-06-06
Make dev installation easier

michwill:
* Keep cryptography.io wheels here for now
* Remove dependencies from setup.py (they were wrong anyway)
* Create a shell scrypt to install everything in pipenv environment (including pipenv itself locally, if needed)


michwill:
Still an open question, I think: what do we do with deps in `setup.py`?
My thought here was that better empty than wrong. But perhaps, we can pull them from requirements.txt (like pyethereum does) which, in turn, can be generated by pipenv?

michwill:
Then we need a correct method to fetch those wheels from `pyUmbral` git (the one which was there before wasn't working)

KPrasch:
This isn't the final form of the installation process, but it does ease the process for now. :-)

PR #293, opened 2018-06-06
Changes storage of 33 bytes

szotov:
I tried other several options: `byte[33]`, `bytes32` + `bool`, `bytes` as input and `bytes32` + `byte` as storage.
`bytes32` + `bool` and `bytes32` + `byte` has the same gas consumption and contract size.
`byte[33]` looks like `bytes32[33]`, so it's not an option.
`bytes` as input has the same gas consumption as `bytes32` + `byte` but bigger contract size.

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -195,7 +195,8 @@ class Miner(NucypherTokenActor):
     def publish_data(self, data) -> str:
         """Store new data"""
 
-        txhash = self.miner_agent.contract.functions.setMinerId(data).transact({'from': self.address})
+        txhash = self.miner_agent.contract.functions.setMinerId(
+            self.blockchain.interface.w3.toBytes(data[0]), data[1:]).transact({'from': self.address})
         self.blockchain.wait_for_receipt(txhash)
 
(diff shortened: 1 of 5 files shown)


jMyles:
Yeah, this is good.  I also want to point out, though, that we can change things up so that we need to only record the pubkey fingerprint instead of the entire material.  It will be non-trivial, but definitely possible.  Is that something we need to sketch out?

szotov:
How often we will save new key for one miner?
Now it costs ~80k-100k
So if this occurs less than once per a day/period then it's not a problem, because `confirmActivity` costs more than 120k and `createPolicy` is also expensive (170k + 100k per node)

KPrasch:
Very Infrequently: most of the time it's just once per miner; However it is possible that Ursula looses her signing key, and she will need to set it again. This is one of the first operations that a Miner/Ursula performs, when they are "open for business".

szotov:
I think we have two good options: 
1) current `bytes` array - easy to use and can hold anything additional information for miner (100-120k of gas to save, 1kb additional size of contract) 
2) `bytes32` + `byte` - optimized by gas and size of contract (80k-100k)

KPrasch:
@szotov - Can you re-poen this PR against `coffee`?

PR #298, opened 2018-06-11
[WIP] Divide stake client method

KPrasch:
Depends on PR #300


KPrasch:
Included upstream in #306

PR #308, opened 2018-06-15
TokenActor client optimization

KPrasch:
- Adds decorator to reduce repeated logic of ownership enforcement
- Deprecates client methods for MinersEscrow datastore


nucypher/blockchain/eth/actors.py
```diff
@@ -11,6 +11,14 @@
 from nucypher.blockchain.eth.constants import calculate_period_duration, datetime_to_period, validate_stake_amount
 
 
+def only_me(func):
+    def wrapped(actor=None, *args, **kwargs):
+        if not actor.is_me:
+            raise actor.MinerError("Cannot execute contract staking functions with a non-self Miner instance.")
```
jMyles: Some stuff got merged.

KPrasch: Yep - Good catch

Issue #316, opened 2018-06-17
Pass proper arguments into policy_agent.contract.functions.createPolicy


jMyles:
`BlockchainArrangement.publish` passes arguments into `PolicyAgent.contract.functions.createPolicy` that are consistent with a previous contract version and inconsistent with the current one.

jMyles:
@KPrasch, @szotov: Still need to pow-wow on this one.

KPrasch:
Since we cannot reproduce this Issue any longer - Im closing the ticket.

PR #318, opened 2018-06-17
Learning Loop Part II: Rise of the Federated Policies

### LINKED ISSUE #289: Make connectivity logic more swappable
Currently, some specialized logic for the decentralized scenario is defined in unusual places, while some federated logic is not swappable.  Instead, let's have fully swappable logic for all client methods that need to interact with REST.

- [x] Make sampling logic swappable
- [x] Don't spin up blockchain instances unless `is_me`

### LINKED ISSUE #326: The right way to cast ether address to bytes.
Currently, we cast the string of an ether address to bytes:

`bytes(self.ether_address, encoding="ascii")`

This results in a 42-byte object because we're simply ascii-encoding the address.  We need instead to treat it like hex.  However, casting to bytes as hex obviously causes loss of the case of the string variant, and so casting it back to a string violates the ethereum checksum rules.

See: https://github.com/ethereum/eth-utils#to_checksum_addressvalue---text

jMyles:
*Warning: this PR breaks `Bob` badly.  He'll be fixed in Learning Loop Part III.*

## Major Changes
* Introducing `FederatedPolicy`
* Alice has learning loop logic to select Ursulas from `_known_nodes` for federated `grant()`
* Federated Characters have a `public_address` made from their signing key, even if they don't have an `ether_address`
* `token_airdrop` is now a utility function
* `deployed_testerchain` is morphed into `three_agents`, a fixture which provides access to the test agents.
* Blockchain Ursulas can be found by `grant` (although we're still blocked by #316)

## Smaller Changes
* Splitting logic for Arrangements now includes public address.
* Alice can `publish()` in federated mode; this just publishes the Treasure Map.

Fixes #289 
Fixes #326


nucypher/blockchain/eth/policies.py
```diff
@@ -23,9 +24,14 @@ def __init__(self, author: PolicyAuthor,
                  miner: Miner,
                  value: int,
                  lock_periods: int,
+                 expiration: maya.MayaDT,
                  *args, **kwargs):
         super().__init__(alice=author, ursula=miner, *args, **kwargs)
 
+        delta = expiration - maya.now()
```
szotov: There is problem with this subtraction in windows: result is seconds but treated as days

nucypher/blockchain/eth/policies.py
```diff
@@ -112,54 +119,30 @@ def __find_ursulas(self, ether_addresses: List[str], target_quantity: int, timeo
...
 
             try:
                 # Check if this is a known node.
-                selected_ursula = self.alice.known_nodes[ether_address]
+                selected_ursula = self.alice._known_nodes[ether_address]
 
             except KeyError:
                 # Unknown Node
-                self.alice.nodes_to_seek.add(ether_address)  # enter address in learning loop
+                self.alice.learn_about_specific_node(ether_address)  # enter address in learning loop
```
michwill: Does this have a time delay?
The loop won't consume 100% CPU, right?

jMyles: Correct - it's called only every 10 seconds (a subsequent change will make this tuneable).  Also, it's mostly network I/O; it doesn't use much CPU.

nucypher/blockchain/eth/policies.py
michwill: Hmm, should be of timedelta type, I guess

nucypher/policy/models.py
```diff
@@ -216,11 +210,17 @@ def __assign_kfrags(self) -> Generator[Arrangement, None, None]:
 
         for kfrag in self.kfrags:
             for arrangement in self._accepted_arrangements:
-                if kfrag not in self._enacted_arrangements:
+                if not arrangement in self._enacted_arrangements.values():
                     arrangement.kfrag = kfrag
-
                     self._enacted_arrangements[kfrag] = arrangement
+                    print(kfrag, arrangement, arrangement.ursula)
```
michwill: Why `print` though? Is it a debug message?

jMyles: Ah ha... yeah that's not supposed to be there.

nucypher/blockchain/eth/policies.py
KPrasch: The datetime/period conversion code here is also included in a utility function in `constants.py` with fixes for windows.

nucypher/blockchain/eth/policies.py
```diff
@@ -112,54 +119,30 @@ def __find_ursulas(self, ether_addresses: List[str], target_quantity: int, timeo
 
             # Select an ether_address: Prefer the selection pool, then unknowns queue
             if ether_addresses:
-                ether_address = ether_addresses.pop()
+                ether_address = bytes(ether_addresses.pop(), encoding="ascii")
```
KPrasch: This is 42 bytes? The `eth_utils` module includes functions to cast to/from bytes while preserving a valid checksum address.

nucypher/blockchain/eth/policies.py
```diff
@@ -175,23 +158,27 @@ def make_arrangements(self, network_middleware: RestMiddleware,
         selected_addresses = set()
         try:  # Sample by reading from the Blockchain
             actual_sample_quantity = math.ceil(target_sample_quantity * ADDITIONAL_URSULAS)
-            sampled_addresses = self.alice.recruit(quantity=actual_sample_quantity)
+            duration = int(calculate_period_duration(expiration))
```
KPrasch: Here is an example of the time/period conversion.

nucypher/blockchain/eth/policies.py
```diff
@@ -175,23 +158,27 @@ def make_arrangements(self, network_middleware: RestMiddleware,
         selected_addresses = set()
         try:  # Sample by reading from the Blockchain
             actual_sample_quantity = math.ceil(target_sample_quantity * ADDITIONAL_URSULAS)
```
KPrasch: Sampling `1.5 * len(kfrags)` Ursulas here in case some reject. Additionally sampling from the blockchain has another buffer of 1.7.  This makes the minimum number of nodes the network `(1.5 * N) * 1.7`

nucypher/blockchain/eth/policies.py
```diff
@@ -175,23 +158,27 @@ def make_arrangements(self, network_middleware: RestMiddleware,
...
             raise self.NotEnoughBlockchainUrsulas(error.format(self.n))
         else:
             selected_addresses.update(sampled_addresses)
 
-        found_ursulas, spare_addresses = self.__find_ursulas(sampled_addresses, target_sample_quantity)
+        found_ursulas = self.__find_ursulas(sampled_addresses, target_sample_quantity)
 
-        candidates = handpicked_ursulas + found_ursulas
+        candidates = handpicked_ursulas
+        candidates.update(found_ursulas)
```
KPrasch: Here we combine handpicked with sampled Ursulas for consideration, where there is no guarantee that handpicked Ursulas will be included in the policy if some of them reject, or are not assigned kfrags when the candidate pool is not exhausted.  IMO, this is good behaviour, but perhaps we can further prioritize the handpicked ursulas consideration.

nucypher/blockchain/eth/policies.py
```diff
@@ -210,5 +199,14 @@ def make_arrangements(self, network_middleware: RestMiddleware,
             if len(accepted) < self.n:
                 raise Exception("Selected Ursulas rejected too many arrangements")  # TODO: Better exception
 
-        self._accepted_arrangements.update(accepted)
-        self._rejected_arrangements.append(rejected)
+    def publish(self, network_middleware) -> None:
```
KPrasch: This is a public interface separate from enact?  This means that policies can be enacted over REST, but not published to the blockchain. Can duplicate arrangements possibly be generated? When will users want enact but not publish?

nucypher/characters.py
```diff
@@ -55,7 +55,7 @@ def __init__(self, is_me=True,
                  network_middleware=None,
                  crypto_power: CryptoPower = None,
                  crypto_power_ups=None,
-                 federated=False,
+                 federated_only=False,
```
KPrasch: Hmph - I liked `federated` and `is_federated` better :-p

nucypher/characters.py
```diff
@@ -207,9 +207,13 @@ def start_learning(self):
...
         self.log.warning("Unhandled error during node learning: {}".format(args))
 
-    def select_teacher_nodes(self):
+    def shuffled_known_nodes(self):
         nodes_we_know_about = list(self._known_nodes.values())
         random.shuffle(nodes_we_know_about)
+        return nodes_we_know_about
+
+    def select_teacher_nodes(self):
+        nodes_we_know_about = self.shuffled_known_nodes()
```
KPrasch: Some ascii art will go a long way sound here somewhere.

nucypher/characters.py
```diff
@@ -207,9 +207,13 @@ def start_learning(self):
     def handle_learning_errors(self, *args, **kwargs):
         self.log.warning("Unhandled error during node learning: {}".format(args))
 
-    def select_teacher_nodes(self):
+    def shuffled_known_nodes(self):
         nodes_we_know_about = list(self._known_nodes.values())
```
KPrasch: +1 Derives a new list for use as teacher nodes.

nucypher/characters.py
```diff
@@ -434,15 +442,34 @@ def public_key(self, power_up_class: ClassVar):
         power_up = self._crypto_power.power_ups(power_up_class)
         return power_up.public_key()
 
+    @property
+    def public_address(self):
+        # TODO: Figure out the *real* way to cast addresses to bytes (ie, hex or whatever without violating checksum).
```
KPrasch: https://github.com/ethereum/eth-utils#to_canonical_addressvalue---bytes

nucypher/characters.py
```diff
@@ -219,7 +223,10 @@ def select_teacher_nodes(self):
     def cycle_teacher_node(self):
         if not self.teacher_nodes:
             self.select_teacher_nodes()
-        self._current_teacher_node = self.teacher_nodes.pop()
+        try:
+            self._current_teacher_node = self.teacher_nodes.pop()
+        except IndexError:
+            raise self.NotEnoughUrsulas("Don't have enough nodes to select a good teacher.  This is nearly an impossible situation - do you have seed nodes defined?  Is your network connection OK?")
```
KPrasch: What are the conditions of escaping the learning loop process? We need to build in a timeout, I think.

nucypher/characters.py
```diff
@@ -558,12 +586,12 @@ def follow_treasure_map(self, hrac, using_dht=False):
...
                 newly_discovered_nodes.update(new_nodes)
 
-            self.known_nodes.update(newly_discovered_nodes)
+            self._known_nodes.update(newly_discovered_nodes)
             return newly_discovered_nodes, number_of_known_treasure_ursulas
         else:
             for ursula_interface_id in self.treasure_maps[hrac]:
                 pubkey = UmbralPublicKey.from_bytes(ursula_interface_id)
-                if ursula_interface_id in self.known_nodes:
+                if ursula_interface_id in self._known_nodes:
```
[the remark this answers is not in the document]
KPrasch: Since blockchain sampling returns ethereum addresses, We need a way to quickly lookup known nodes by their addresses, while still providing compatibility for a federated version where known nodes is keyed differently.

nucypher/crypto/constants.py
```diff
@@ -9,4 +9,4 @@
 # These lengths are specific to secp256k1
 CAPSULE_LENGTH(98)
 PUBLIC_KEY_LENGTH(33)
-PUBLIC_ADDRESS_LENGTH(42)
+PUBLIC_ADDRESS_LENGTH(20)
```
KPrasch: err..? Yes ok good.

nucypher/characters.py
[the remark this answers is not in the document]
cygnusv: So we do mix both types, right?

nucypher/blockchain/eth/policies.py
jMyles: Yeah, well, as you can see, it unhexlify'd it elsewhere.

But more to the point: "preserving"?  As far as I can tell, it just checksums it again.  Is there such a thing as "preserving"?  Obviously it can't be with a hex encoding, if so.  The [document in question](https://github.com/ethereum/eth-utils#to_checksum_addressvalue---text) appears to re-checksum them, with a deterministic result per EIP-55.  What is this "preserving"?  Can you link me to the document you're talking about?

jMyles: To be clear: the unhexlify routine, which I have done below, is the obvious way forward (ie, `public_address`).  This change is essentially temporary to get this PR in.  Consider it similarly transitory to BrokenBob.  :-)

nucypher/blockchain/eth/policies.py
jMyles: Hmmm... Right.  OK, so are you asking anything here?  Or just commenting on the existing code?

nucypher/blockchain/eth/policies.py
jMyles: Yeah, I mean I just moved this from one place to another obviously.  I see the convenience method that y'all have pointed out; we'll use that in the future.

nucypher/blockchain/eth/policies.py
jMyles: Aye.  :+1:

nucypher/blockchain/eth/policies.py
jMyles: :+1:

nucypher/characters.py
jMyles: OK, now I understand - I gave `to_bytes` a spin and didn't like it, but I didn't realize that I had inadvertently rebuilt `to_canonical_address` :-)

nucypher/characters.py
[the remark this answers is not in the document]
jMyles: Yeah, we've gone around and around about this, but I think the current solution is ideal.  We talked about having two dicts (one for each), but now instead we provide a way (`public_address`) to get an ethereum-address like construct from an actor who doesn't have a blockchain presence, using their `UmbralPublicKey`.

nucypher/crypto/constants.py
jMyles: Yeah, I'm sayin'.  I think you understand now.  :-)

tests/characters/test_alice_can_grant_and_revoke.py
```diff
@@ -3,62 +3,80 @@
...
-from umbral.keys import UmbralPublicKey
 
 from nucypher.characters import Ursula
 from nucypher.crypto.api import keccak_digest
 from nucypher.crypto.powers import SigningPower, EncryptingPower
+from umbral.fragments import KFrag
+
+
+class MockPolicyCreation:
+    """
```
jMyles: I actually pulled this out of use at the last minute, but it will be used in LL3.

nucypher/characters.py
```diff
@@ -473,17 +500,19 @@ def create_policy(self, bob: "Bob", label: bytes, m: int, n: int):
                        public_key=public_key,
                        m=m)
 
-        from nucypher.policy.models import Policy
-
         if self.federated_only is True:
-            policy = Policy(alice=self, **payload)
+            from nucypher.policy.models import FederatedPolicy
+            # We can't sample; we can only use known nodes.
+            known_nodes = list(self._known_nodes.values())
```
jMyles: This is supposed to use `shuffled_known_nodes`.

nucypher/blockchain/eth/policies.py
jMyles: :heavy_check_mark:

nucypher/characters.py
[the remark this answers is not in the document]
jMyles: @cygnusv: No, we don't mix both types - we use a 20 byte "address-style" representation in either case.

nucypher/blockchain/eth/policies.py
jMyles: :heavy_check_mark:

nucypher/blockchain/eth/policies.py
jMyles: :heavy_check_mark:

nucypher/characters.py
jMyles: :heavy_check_mark:

nucypher/characters.py
jMyles: :heavy_check_mark:

nucypher/crypto/constants.py
jMyles: :heavy_check_mark:

tests/characters/test_alice_can_grant_and_revoke.py
jMyles: :heavy_check_mark:

jMyles:
@KPrasch, I'm needing to see the traceback (or at least, some runtime details), and CircleCI doesn't appear to be providing them.  The configuration says that pytest was invoked with `-v`, but the typical verbose output appears to be missing.  What am I missing?

KPrasch:
@jMyles Looks like the Issue is with an old version of bytestringSplitter being cached.

PR #325, opened 2018-06-22
High level ecomomic parameters

michwill:
Converting parameters from the token economics paper into what smart contract can consume

### DIFF
--- a/nucypher/blockchain/economics.py
+++ b/nucypher/blockchain/economics.py
@@ -1,0 +1,57 @@
+from math import log
+
+
+class Economics:
+    def __init__(
+            self,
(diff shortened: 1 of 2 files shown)


michwill:
I am imagining this being used when deploying the MinersEscrow. Hence, the question: would you call using it + MinersEscrow, and testing that after `N` periods we have the right number of coins (according to formulas), an integration test?
E.g. having semi-realistic distributions of miners?

Another question: does it seem to be the right place for this logic?

szotov:
If we will have this logic in the MinersEscrow deploying then I prefer to leave this in main package. If not - move to scripts or another dev stuff

PR #327, opened 2018-06-24
Create a BlockchainPower with signing and verifying methods

tuxxy:
### What this does:
1. Adds `BlockchainPower` with `sign_message` and `verify_message` methods.
2. Adds `call_backend_sign` and `call_backend_verify` methods on `DeployerCircumflex`.
3. `call_backend_sign` uses some `eth_tester` stuff to sign messages if the `EthereumTester` provider is being used, otherwise, it defaults to the (deprecated) `w3.eth.sign` which might be problematic.

### DIFF
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -1,3 +1,4 @@
+import binascii
 import json
 import os
 import warnings
@@ -5,6 +6,9 @@ from pathlib import Path
 from typing import Tuple, List
 
 from constant_sorrow import constants
+from eth_utils import to_canonical_address
+from eth_keys.datatypes import PublicKey, Signature
+from web3.providers.eth_tester.main import EthereumTesterProvider
 from web3 import Web3, WebsocketProvider, HTTPProvider, IPCProvider
 from web3.contract import Contract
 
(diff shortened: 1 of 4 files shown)


nucypher/blockchain/eth/interfaces.py
```diff
@@ -390,3 +394,28 @@ def deploy_contract(self, contract_name: str, *args, **kwargs) -> Tuple[Contract
...
+        eth-tester signing interface to do so.
+        """
+        provider = self._providers[0]
+        if isinstance(provider, EthereumTesterProvider):
+            address = to_canonical_address(account)
+            sig_key = provider.ethereum_tester.backend._key_lookup[address]
+            signed_message = sig_key.sign_msg(message)
+            return signed_message.to_hex()
+        else:
+            return self.w3.eth.sign(account, data=message) # Technically deprecated...
```
michwill: ...

tuxxy: Yeah, this isn't great. Unless we're going with a full wallet implementation before testnet, this is what were stuck with for the current ecosystem.

I don't think handling wallet keys is something we're ready to tackle _right now_.

jMyles: :eyes:

cygnusv: > ...

Uff, that's tough...

nucypher/crypto/powers.py
```diff
@@ -61,6 +65,65 @@ class CryptoPowerUp(object):
...
 
 
+class BlockchainPower(CryptoPowerUp):
+    """
+    Allows for transacting on a Blockchain via web3 backend.
+    """
+
+    def __init__(self, blockchain: 'Blockchain', account: str):
+        """
+        Instantiates a BlockchainPower for the given account id.
```
KPrasch: ...ether address?

tuxxy:
I am willing to move this code to `NuCypherTokenActor` and reserve a proper `BlockchainPower` for a much better solution.

PR #336, opened 2018-06-27
Remove version specifications from Pipfile

KPrasch:


Pipfile
```diff
@@ -3,30 +3,27 @@ url = "https://pypi.python.org/simple"
...
-appdirs = "*"
-web3 = "==4.3.0"
+sqlalchemy = "*"
+apistar = "*"
+rpcudp = {git = "https://github.com/nucypher/rpcudp", ref = "nucypher-depend"}
+kademlia = {git = "https://github.com/nucypher/kademlia", ref = "nucypher-depend"}
+pynacl = "*"
+pysha3 = "*"
+py_ecc = "*"
+pyumbral = {git = "https://github.com/nucypher/pyumbral.git", ref = "nucypher-depend"}
```
michwill: Are we still doing the `nucypher-depend` thing for `pyUmbral`? I thought, we wanted to switch back to `master`?

jMyles: Hmm... I don't (at least immediately) recall agreeing to switch back to `master`, but I do think we're ready (or maybe we wait another 1-2 weeks for this final chips to fall) to move to a depend-on-release format.

PR #329, opened 2018-06-25
Learning Loop Part III

jMyles:
### Major changes
* Bob is working again.
* Ursula is now fully serializable and splittable.
* `TreasureMap` serialization, splitting, and verification is all handled internally.
* Bob now follows TreasureMaps asynchronously
* Ursula now propagates TreasureMaps received via REST

### Minor changes
* Can't specify a teacher node as method params anymore.  Just set it as the next teacher.
* Removed some deprecated fixtures

Fixes #172 
Fixes #227
Fixes #235
Fixes #238
Fixes #326 

Depends on bytestringSplitter 9: https://github.com/nucypher/bytestringSplitter/pull/9

PR #301, opened 2018-06-12
Updates web3 usage to version 4.3.0

szotov:

PR #344, opened 2018-06-30
Store test Results; Upgrade Web3

KPrasch:
- Store XML Test result summary for circleci API
- Updates web3.py to 4.4.1 (Causes secondary Issue requiring a downgrade of `py-evm` + `eth-tester`). 
- Fixes `Pipfile` versioning for PR #329 
- Removes unused dependencies
- Organize and label `Pipfile`

Prior Issue reference with working versions: PR #301

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -23,7 +23,7 @@ pynacl = "*"
 pysha3 = "*"
 py_ecc = "*"
 pyumbral = {git = "https://github.com/nucypher/pyumbral.git", ref = "nucypher-depend"}
-web3 = "*"
+web3 = "==4.4.1"
 
 [dev-packages]
 pytest = "*"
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -10,31 +10,33 @@ cryptography35_linux = { path='./wheelhouse/cryptography-2.3.dev1-cp35-cp35m-lin
...
 pytest-mypy = "*"
-codecov = "*"
-coverage = "*"
 pytest-cov = "*"
+
+# Ethereum Dev Packages #
 py-solc = "*"
-py-evm = "*"
-eth-tester = "*"
+py-evm = "==0.2.0a18"
```
michwill: Wait.. are you sure?

KPrasch: We are still being effected by a `evm.exceptions.ValidationError: mix hash mismatch; 0x...` when running `py-evm<=0.2.0a18`.  Similar to the Issue seen here:  https://github.com/ethereum/eth-tester/issues/99 - except that `eth-tester` seemingly does support newer versions of `py-evm`: https://github.com/ethereum/eth-tester/blob/master/setup.py#L32.

I'm unsure what the fix is at this time.

Issue #180, opened 2018-03-13
Better Arrangement ID.


jMyles:
Presently, we use an hrac as a bare identifying instrument in a few places.  For example, the URL which Ursula exposes is `/kFrag/{hrac_as_hex}`.  Let's make is `/kFrag/{arrangement.id}` per above.  Both Alice and Bob will be able to calculate the arrangement.id, but it will prevent Sybil / Eve from poking various Ursulas in an effort to either see which Ursulas have a given KFrag or determine the value of `n`.

michwill:
Actually, to go to the extreme, the `arrangement.id` can also be purely random and stored in the treasure map

jMyles:
Ahh right, and then the TreasureMap is attainable by `hrac`.  

Yeah, that makes more sense.  It occurs to me that the `hrac` is public anyway (as part of the DHT, although replacing the DHT might (but might not) afford us the removal of it from public view).

michwill:
Yeah, but I think eventually we may use anything which interfaces as a public key-value store

jMyles:
New way:

* Alice generates a random stub, puts it in `TreasureMap` for Bob to find.
* Alice attempts to set `Arrangement` with ID hash(alice_pubkey_sig + ursula_pubkey_sig + stub).  
* Request must include signature of Arrangement ID by Alice.
* Ursula checks to see that ID is properly signed.
* If all is well, she sets the policy as normal.
* If signature doesn't check out, she 403's.
* If arrangement ID already exists, she 409's.

jMyles:
Update (this was from weeks ago, but I neglected to add it until now) - @michwill points out that it actually makes more sense for the Arrangement ID to be:

`hash(alice_pubkey_sig + hash(ursula_pubkey_sig + nonce))` - that way Ursula doesn't need to know the nonce to verify the string; she can simply be handed `hash(ursula_pubkey_sig + nonce))`, while Bob can still construct it himself.

jMyles:
8:57 AM] jMyles: @kprasch, @dnunez (and everyone): 

First things first: It's totally safe and fine to go back to deterministic IDs, but doing it right probably requires us to do some thoughtful engineering around it.

[...]

PR #348, opened 2018-07-04
All tests passing.

jMyles:
Again, not a heck of a lot of change here.

* We now pass `checksum_address` into the actor classes from `Character` (after a vigorous discussion on the matter last night with @KPrasch).
* We lookup Ursulas "manually" in a couple of tests now.
* Federated-only nodes which learn about other nodes build them as federated-only strangers (ie, their canonical address much match their signing key)
* Test cleanups and fixups to account for all the recent changes.


jMyles:
Presently, a `federated_only` Alice will discover decentralized nodes but save them by their signing key (probably undesirable behavior).

Some of these questions remain to be worked out obviously.

Presently, a federated Alice has no way to substantiate a decentralized Ursula's claim to have signed a particular signing key (which, as of #180, will become a crucial undertaking) because she can't verify without `BlockchainPower`.

It's possible that our best move is to leave this stuff until either EIP-712 is implemented or we're facing real demand on mainnet, whichever comes first.

michwill:
Right, doing anything about it is probably beyond this PR. Also just throwing an idea that, perhaps, we should have a version string when discovering nodes, and the behavior can depend on it.
Also, in a "federated" network, we'll probably have a list of node IDs which we trust in a config file (which both Alices and Ursulas have), while in the decentralized network we discover it from blockchain

PR #286, opened 2018-06-04
Coffee

jMyles:
#### The initial goals for this PR were as follows:

- The highest-level goal of this PR is to resolve #295.  Doing that also requires:
-- Ursula becoming "fully available" for Alice to find, as described in #294.
-- Complete adjustment of Arrangement ID as described in #180 
-- Versioned Arrangements #281 
-- Better node caching on Alice, perhaps sharing logic with #227  

We have pushed #180 and #281 down the road a bit, but the rest are closed, along with many others along the way (see the various included PRs for details: #307, #308, #318, #324, #327, #330, #336, #344, #345, #348, and #351).

#### What this PR represents
Closing the gap between Alice having access to the blockchain and actually being able to create real Policies - the core of the NuCypher offering - is one of the crucial details to work out to realize this project's goals.  This PR specifically addresses the portions of that gap involving discovering, remembering, and verifying nodes (especially Ursulas).

As you read this PR and consider it for merge, consider these as the main points of change:

#### Character
- All actors now learn about nodes now learn about each other in a uniform fashion.
- Learning happens off the main thread in a "learning loop."
- All Characters now have a canonical address.
- DHT legality logic has blossomed into a full-blown verification scheme for Ursulas, and is used both by other Ursulas when deciding whether to propagate and also by other actors when deciding when to interact with Ursula.

#### Policy
- Creates Arrangements and Policies from sampled addresses
- Deprecate mocked blockchain client
- Create Arrangements from sampled ursulas
- Enact and publish policies via rest and blockchain network
- Invert `Arrangement` and `BlockchainArrangement` inheritance order

#### Network
- Deprecates `NuCypherDHTNode` and `Capabilities`
- Renames `NetworkyStuff` -> `NetworkMiddleware`
- Eliminates DHT usage entirely for Alice and Bob (although Alice still has some entirely optional DHT capability)

#### Blockchain
- New stake logic; stake dividing is possible, unlocking is stepwise.
- Embed default web3.py providers in Interface
- Use Interface-style API for Blockchain connection
- Reduces complexity of `ContractAgent` API with default values
- Implements miner's datastore publish and fetch agent methods
- Uses Ursula's ethereum keypair to sign and verify her other metadata
- Improves solidity compile-time exception handling
- Optimzes contract source code
- Introduce blockchain actors to existing tests, fixtures, and utilities

#### Configuration
- Compose `NucypherConfiguration` and `BlockchainConfiguration`
- Decouple configuration: Keep it optional
[...]

PR #351, opened 2018-07-05
The stuff to get us off coffee.

### LINKED ISSUE #294: After recruiting, how does Alice connect to reliably connect to an authentic Ursula?
When Alice meets Ursula for the first time, Ursula is a mysterious figure, solidified in existence only by the presence of an otherwise forgettable entry in the blockchain that she owns, vests, and locks a medallion-sized stake with a particular ethereum address.

But let's go one step further back in history, to `Alice.recruit` / `sample(...)`.  Here, Alice, using only knowledge publicly available on the blockchain, receives a representation of Ursulas, roughly selected in proportion to the amount of their stake.  This representation currently is simply the ethereum address I referenced in paragraph 1.

How then does Alice reliably take a subsequent step?

Here are some possibilities that we have pondered since we went down this path of architecture:

#### A) Alice is able to look up Ursulas, by ethereum address, on the DHT.

This option is now precluded, at least as the exclusive possibility, for two reasons:

1) We have never established a great way for Ursula, with her ethereum address alone, to establish certainty over her signing key (more on this later - this is important).

2) We now assume that Alice uses the DHT only optionally.

#### B) Alice continually tracks Ursulas as `known_nodes`, IDing them by ethereum address.

This suffers from two similar problems to the DHT option:

1) Same as above - no great way to match.

2) If this is the only way, we lose the future ability to deploy a federated variant.

#### C) Alice performs `n` additional reads to the blockchain, getting Ursula's ID, which we uses to perform `n` *more* lookups using the REST endpoint of an already-known node

This has been our presumed approach so far, and it suffers from three problems:

1) This requires Ursula to spend the gas to publish her complete 33-byte signing key to the blockchain, so that her interface info can be reliably verified.  (this can be modified somewhat so that she only includes a fingerprint

2) There is no straightforward way for Ursula to ever cycle her signing key without ending all policies (and her stake) and starting fresh.  See #169 

3) If the Ursula in question is down when Alice tries to authenticate her, Alice will need to run `recruit` again, which may become a pain point.

#### D) (This is the one I'm currently preferring) Ursula, using her transacting key, pre-signs her own signing public key and caches it

In this scenario, Ursula uses the signature to "prove herself" when propagating her interface info (either via the DHT or in whatever other way we invent).  Then, her verifying key is always provided along with her interface info.

This allows us to remove `minerId` from smart contracts completely (potentially saving quite a lot of money in gas over the life of the network) and also produces a cognizable path for cycling her signing keys.

The only disadvantage I can think of is that it slightly complicates the ability of an independent oracle to gain certainty over the correct public signing key to use when verifying a message from Ursula for the purposes of adjudicating a dispute or whatever.

- [ ] Ursula uses ethereum keypair to sign public key
- [ ] Ursula adds public key to interface info on DHT, which is verified with currently legality rules with her ethereum keypair

 (more checklist items to follow)

### LINKED ISSUE #295: Close the gap between recruit and make_arrangements.
Currently, Alice is able to `.recruit()`, using `sample()`, `n` Ursulas, weighed by stake, and indicated by ethereum address.

However, there is no authoritative way to get the REST details from this information, which Alice needs to move on to `make_arrangements()`.

jMyles:
* Ursula now signs her verifying key with her ethereum key.
* Nodes in decentralized mode verify Ursula's wallet signature.
* Ursula includes her public address in her signed interface info.
* All learning nodes now track other nodes by address.
* Several new tests again plausible attacks, such as #353 

Fixes #294
Fixes #295


nucypher/network/protocols.py
```diff
@@ -58,7 +58,7 @@ def rpc_store(self, sender, nodeid, key, value):
         if header == constants.BYTESTRING_IS_URSULA_IFACE_INFO:
             from nucypher.characters import Ursula
             stranger_ursula = Ursula.from_bytes(payload,
-                                                federated_only=True)  # TODO: Is federated_only the right thing here?
+                                                federated_only=self.sourceNode.federated_only)  # TODO: Is federated_only the right thing here?
```
tuxxy: :+1: On having this discussion at some point. Good TODO.

tests/utilities.py
```diff
@@ -196,8 +190,7 @@ def get_nodes_via_rest(self, address, port, node_ids):
         return response
 
     def put_treasure_map_on_node(self, node, map_id, map_payload):
-        running_node = _ALL_URSULAS[node.rest_interface.port]
-        mock_client = TestClient(running_node.rest_app)
+        mock_client = self.__get_mock_client_by_port(node.rest_interface.port)
```
tuxxy: Thoughts on making this a fixture?

.circleci/config.yml
```diff
@@ -5,9 +5,9 @@ workflows:
   test:
     jobs:
       - bundle_dependencies-36
-      - mypy_type_check:
-          requires:
-            - bundle_dependencies-36
+#      - mypy_type_check:
```
tuxxy: :+1:

nucypher/crypto/powers.py
```diff
@@ -99,7 +104,7 @@ def sign_message(self, message: bytes):
         signature = self.blockchain.interface.call_backend_sign(self.account, message)
         return signature
 
-    def verify_message(self, address: str, pubkey: bytes, message: bytes, signature: str):
+    def verify_message(self, address: str, pubkey: bytes, message: bytes, signature: Signature):
```
tuxxy: This is a step in the wrong direction.

This is meant to be an interface to what the blockchain sees. Providing the `eth_keys` `Signature` object is only going to allow users to make mistakes with that.

This is meant to be an interface for exactly what a smart contract would most likely interact with. I'm not very keen on providing an object that we don't have control over.

nucypher/blockchain/eth/interfaces.py
```diff
@@ -418,7 +418,7 @@ def call_backend_sign(self, account: str, message: bytes) -> str:
             address = to_canonical_address(account)
             sig_key = provider.ethereum_tester.backend._key_lookup[address]
             signed_message = sig_key.sign_msg(message)
-            return signed_message.to_hex()
+            return signed_message
```
tuxxy: This breaks parity with the rest of the web3 API.

So, if we're running a test, this API is called, which returns a string. In most cases, this is what we should expect when calling some node to sign something for us, not the object.

tuxxy: This is why I hate this code and want a better signing solution from web3py here.

nucypher/crypto/powers.py
jMyles: > Providing the eth_keys Signature object is only going to allow users to make mistakes with that.

...why?  Why doesn't it also allow them to do the right thing with that?

> I'm not very keen on providing an object that we don't have control over.

Wait, what?  Are you saying you are just wholesale against returning objects from dependency libraries?

I don't see how it makes *any* sense to return the hex.  We never use it that way.

nucypher/blockchain/eth/interfaces.py
jMyles: I... don't see the problem.

nucypher/crypto/powers.py
tuxxy: Essentially, I'm saying that `eth_keys` `Signature` class is too low-level to be exposed here. If we can restrict what can be done on this `Signature` class, then we could return that. Like we discussed in PDX, I would prefer this, but the ecosystem has yet to catch up here.

nucypher/blockchain/eth/interfaces.py
tuxxy: In one case (running tests), we return a `Signature`. In another, we return a string.

This is not good API design for signature, imho. They should both return an object (which is problematic), or they should both return a string.

jMyles: Ahh, I see.  So `w3.eth.sign` returns a string?

In any case, remember this is a [WIP] PR - I haven't even looked at the `else` clause in this function yet.

nucypher/crypto/powers.py
jMyles: I see. 

Well, I think that the proper pattern here is:

* Have address to verify
* Have message purporting to come from address
* Have signature against this message
* Validate signature
* Get public key from signature
* Derive address from signature
* Confirm that addresses match

That appears to be the intention of this object, and it performs this task very well.  So, do you want to cast to bytes and then cast back to `Signature`?  I guess that's fine.

tuxxy: Yes, that would be ideal (for now). In the `verify_message` method, at least.

jMyles:
I propose that, upon merging this, we freeze #286 for review.

nucypher/crypto/powers.py
```diff
@@ -97,22 +101,22 @@ def sign_message(self, message: bytes):
             raise PowerUpError("Account is not unlocked.")
 
         signature = self.blockchain.interface.call_backend_sign(self.account, message)
-        return signature
+        return signature.to_bytes()
 
-    def verify_message(self, address: str, pubkey: bytes, message: bytes, signature: str):
+    def verify_message(self, address: str, pubkey: bytes, message: bytes, signature_bytes: bytes):
```
michwill: Not that it matters much, but is it necessary to have the type indicated in the argument name as `signature_bytes` rather than just naming `signature`?

jMyles: Yeah, this is a perennial wonder of mine.  Since we make a `Signature` object in the next line, doesn't it make sense to clarify that this object is a bytes representation?  I dunno.

michwill: I'm not too strong on one or the other way though. It's not super-important here. So, I'm ok with `_bytes`

nucypher/blockchain/eth/actors.py
```diff
@@ -27,7 +25,7 @@ class NucypherTokenActor:
     class ActorError(Exception):
         pass
 
-    def __init__(self, checksum_address: str=None, token_agent: NucypherTokenAgent=None):
+    def __init__(self, checksum_address: str = None, token_agent: NucypherTokenAgent = None):
         """
         :param checksum_address:  If not passed, we assume this is an unknown actor
```
cygnusv: I don't understand what's the deal with `self.checksum_public_address` attribute below (L37). Isn't it always going to raise an AttributeError since it doesn't exist yet?

jMyles: It's a compromise for the case (which occurs in practice, almost without exception) where this class is mixed in with Character.

.circleci/config.yml
KPrasch: Perhaps we just allow failure here

PR #352, opened 2018-07-05
Minor improvements for Windows compatibility

szotov:
Helps with developing smart contracts (not full compatibility)

### DIFF
--- a/nucypher/config/utils.py
+++ b/nucypher/config/utils.py
@@ -125,7 +125,7 @@ def check_config_tree(configuration_dir: str=None) -> bool:
 
 def check_config_runtime() -> bool:
     rules = (
-        (os.getuid() != 0, 'Cannot run as root user.'),
+        (os.name == 'nt' or os.getuid() != 0, 'Cannot run as root user.'),
     )
 
     for rule, failure_reason in rules:
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -40,8 +40,9 @@ def tempfile_path():
     """
     User is responsible for closing the file given at the path.
     """
-    _, path = tempfile.mkstemp()
+    fd, path = tempfile.mkstemp()
     yield path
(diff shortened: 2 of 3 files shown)


nucypher/config/utils.py
```diff
@@ -125,7 +125,7 @@ def check_config_tree(configuration_dir: str=None) -> bool:
 
 def check_config_runtime() -> bool:
     rules = (
-        (os.getuid() != 0, 'Cannot run as root user.'),
+        (os.name == 'nt' or os.getuid() != 0, 'Cannot run as root user.'),
```
michwill: Presumably, ok if `Administrator`?

tests/fixtures.py
```diff
@@ -118,6 +118,7 @@ def tempfile_path():
     """
     User is responsible for closing the file given at the path.
     """
-    _, path = tempfile.mkstemp()
+    fd, path = tempfile.mkstemp()
     yield path
+    os.close(fd)
```
michwill: So, Windows doesn't complain that this file is opened elsewhere?

nucypher/config/utils.py
szotov: For now it's not for real usage on Windows, so it's ok

tests/fixtures.py
szotov: File opened in the same process using `with` statement, I'm not sure why it still open

KPrasch:
This will need a rebase against master after merging #286

tuxxy:
What problem is this exactly solving for Windows? I don't have anything to run it on.

szotov:
These changes allow to run blockchain tests on Windows with a predefined environment. There are no tests for those lines, main goal is not to break existing code/tests

michwill:
I guess, we should rebase it with recent master and merge?

PR #356, opened 2018-07-12
New Arrangement ID

### LINKED ISSUE #136: Ursula's Node ID on the DHT
It may not be important anymore, but we're still determining Ursula's DHT ID with random bytes.  Do we want a better way to do this?  Or is it no longer important?

### LINKED ISSUE #180: Better Arrangement ID.

jMyles:
Arrangement IDs no longer rely on HRAC; they are random and different for each Ursula.
Small optimizations on Learning Loop.

Fixes #136
Fixes #180

### DIFF
--- a/nucypher/characters.py
+++ b/nucypher/characters.py
@@ -307,13 +307,16 @@ class Character:
         self._node_ids_to_learn_about_immediately.update(canonical_addresses)  # hmmmm
         self.learn_about_nodes_now()
 
-    def block_until_nodes_are_known(self, canonical_addresses: Set, timeout=10, allow_missing=0):
+    def block_until_nodes_are_known(self, canonical_addresses: Set, timeout=10, allow_missing=0,
+                                    learn_on_this_thread=False):
         start = maya.now()
         starting_round = self._learning_round
 
         while True:
             if not self._learning_task.running:
                 self.log.warning("Blocking to learn about nodes, but learning loop isn't running.")
+            if learn_on_this_thread:
+                self.learn_from_teacher_node(eager=True)
             rounds_undertaken = self._learning_round - starting_round
             if (maya.now() - start).seconds < timeout:
                 if canonical_addresses.issubset(self._known_nodes):
@@ -363,7 +366,7 @@ class Character:
 
             try:
                 if eager:
-                    node.verify(self.network_middleware, accept_federated_only=self.federated_only)
+                    node.verify_node(self.network_middleware, accept_federated_only=self.federated_only)
                 else:
                     node.validate_metadata(accept_federated_only=self.federated_only)
             except node.SuspiciousActivity:
--- a/nucypher/characters.py
+++ b/nucypher/characters.py
@@ -813,7 +813,7 @@ class Bob(Character):
 
             if capsules_to_include:
                 work_order = WorkOrder.construct_by_bob(
-                    hrac, capsules_to_include, ursula, self)
+                    arrangement_id, capsules_to_include, ursula, self)
                 generated_work_orders[node_id] = work_order
                 self._saved_work_orders[node_id][capsule] = work_order
 
--- a/nucypher/characters.py
+++ b/nucypher/characters.py
@@ -725,7 +725,10 @@ class Bob(Character):
                                              timeout=timeout,
                                              allow_missing=allow_missing)
             else:
-                self.block_until_nodes_are_known(unknown_ursulas, timeout=timeout, allow_missing=allow_missing)
+                self.block_until_nodes_are_known(unknown_ursulas,
+                                                 timeout=timeout,
+                                                 allow_missing=allow_missing,
+                                                 learn_on_this_thread=True)
 
         return unknown_ursulas, known_ursulas
 
--- a/nucypher/characters.py
+++ b/nucypher/characters.py
@@ -686,8 +686,11 @@ class Bob(Character):
         """
         treasure_map = self.treasure_maps[map_id]
 
-        known_treasure_ursulas = treasure_map.node_ids.intersection(self._known_nodes)
-        unknown_treasure_ursulas = treasure_map.node_ids.difference(self._known_nodes)
+        # The intersection of the map and our known nodes will be the known Ursulas...
+        known_treasure_ursulas = treasure_map.destinations.keys() & self._known_nodes.keys()
+
+        # while the difference will be the unknown Ursulas.
+        unknown_treasure_ursulas = treasure_map.destinations.keys() - self._known_nodes.keys()
 
         return unknown_treasure_ursulas, known_treasure_ursulas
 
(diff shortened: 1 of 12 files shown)


nucypher/policy/models.py
```diff
@@ -52,31 +53,16 @@ def __init__(self, alice, hrac, expiration, ursula=None,
...
-        return bytes(self.alice.stamp) + bytes(
-            self.hrac()) + self.expiration.iso8601().encode()
-
-    def id(self):
-        if not self.ursula:
-            raise TypeError("Can't make an ID for this arrangement yet as we don't know the Ursula.")
-        id_nugget = keccak_digest(self.ursula + self.uuid)
-        full_id = keccak_digest(self.alice.stamp + id_nugget)
-        return id_nugget, full_id
+        return bytes(self.alice.stamp) + self.id + self.expiration.iso8601().encode()
```
tuxxy: :+1:

nucypher/policy/models.py
```diff
@@ -52,31 +53,16 @@ def __init__(self, alice, hrac, expiration, ursula=None,
...
-        #
-        # super().__init__(author=self.alice, miner=ursula,
-        #                  value=self.deposit, lock_periods=lock_periods,
-        #                  arrangement_id=self.id())
-
     def __bytes__(self):
-        return bytes(self.alice.stamp) + bytes(
-            self.hrac()) + self.expiration.iso8601().encode()
-
-    def id(self):
```
tuxxy: Neat :+1:

nucypher/policy/models.py
```diff
@@ -226,14 +212,14 @@ def enact(self, network_middleware, publish=True) -> None:
...
-                                                       self.hrac(),
+                                                       arrangement.id,
                                                        policy_message_kit.to_bytes())
 
             if not response:
                 pass  # TODO: Parse response for confirmation.
 
             # Assuming response is what we hope for.
-            self.treasure_map.add_ursula(arrangement.ursula)
+            self.treasure_map.add_arrangement(arrangement)
```
tuxxy: +1 for `add_arrangement`. :D

nucypher/policy/models.py
```diff
@@ -322,14 +308,14 @@ class TreasureMap:
                                   (bytes, KECCAK_DIGEST_LENGTH),  # hrac
                                   (UmbralMessageKit, VariableLengthBytestring)
                                   )
-    node_id_splitter = BytestringSplitter(PUBLIC_ADDRESS_LENGTH)
+    node_id_splitter = BytestringSplitter(int(PUBLIC_ADDRESS_LENGTH), Arrangement.ID_LENGTH)
 
-    class InvalidPublicSignature(Exception):
+    class InvalidSignature(Exception):
```
tuxxy: :+1:

jMyles: Hmmm, thanks.  :-)

In the back of my head, I thought you might object to this change.  IE, if the publicly verifiable signature is invalid, do we want a *different* exception for that?

It turns out that, currently, our logic doesn't demand it, so I think it's best to go with this simpler version unless and until it does.

nucypher/policy/models.py
jMyles: Yeah, those are the best redlines in the PR right there IMO.  Good eye.

nucypher/policy/models.py
```diff
@@ -26,11 +26,12 @@ class Arrangement:
     A Policy must be implemented by arrangements with n Ursulas.  This class tracks the status of that implementation.
     """
     federated = True
+    ID_LENGTH = 32
```
cygnusv: Is there a particular reason to have arrangement IDs 32 bytes long?

nucypher/policy/models.py
tuxxy: No, I think a general `InvalidSignature` is great. I can see it needing to be moved somewhere else in the future, maybe, but it looks great!

jMyles: Yeah, awesome.  My thoughts exactly.  

At some point, it might make sense for this exception (whatever we end up calling it) and `VerifiableNode.InvalidNode` to inherit from the same base (currently, the later inherits `SuspiciousActivity`).

tuxxy: Hmmm.... I could see that. I think I like that, but an `InvalidSignature` may not necessarily be suspicious activity _all the time_.

nucypher/policy/models.py
cygnusv: I guess this depends only in the UX treatment of arrangement IDs. If in the end they're only used internally in our protocols, then 32 bytes it's fine; if, otherwise, users need to interact with them, then 32-bytes arrangement IDs can be too big.

jMyles: Can you point me to a simple, quick document that guides the selection of choosing minimum viable sizes of bytestrings via different means so as to preclude collision?

cygnusv: If the only concern is avoiding collisions, and given IDs are uniformly sampled, then we can reuse the probability tables from [birthday attacks](https://en.wikipedia.org/wiki/Birthday_attack#Mathematics).

So, for example, with 64-bit IDs, then we need ~190M of IDs to find a collision with probability 0.1%. 

What's an acceptable number? I think that 128 bits are more than enough (in comparison, secp256k1 provides ~128 bits of security), but I guess something in the range between 64 and 128 is good.

jMyles: Yeah, I typically glance at the wikipedia article whenever I'm in this situation (actually, I find [this very similar one](https://en.wikipedia.org/wiki/Birthday_problem#Probability_table) to be a bit more instructive), but what I don't have (and can't immediately find with a bit of googling) is a practical, situational guide about applying birthday problem math to situations like this.

You're saying between 64 and 128 is good (surely closer to 64, right?), but what makes you say that this is in the 64-128 range rather than, say, 32 bits?  When you are playing out a scenario, how do you know how to think about values for p and the collision likelihood?

michwill: I don't think, human readability is really a requirement.
If one _really_ wants human readability, a better way would be to combine with something like ENS

jMyles: I think human readability is important; this is one of the most important pieces of data appearing in logs, and one of the only low-sensitivity bytestrings appearing in log emissions for all three Characters.

nucypher/characters.py
```diff
@@ -736,11 +745,12 @@ def get_treasure_map(self, alice_verifying_key, label):
 
         alice = Alice.from_public_keys({SigningPower: alice_verifying_key})
         compass = self.make_compass_for_alice(alice)
-        treasure_map.orient(compass)
         try:
-            self.treasure_maps[map_id] = treasure_map
-        except treasure_map.InvalidPublicSignature:
+            treasure_map.orient(compass)
```
KPrasch: amazing

cygnusv: image

Issue #363, opened 2018-07-15
Pause learning when all nodes are known.


jMyles:
We need a way to know that a node already has a comprehensive list of all nodes.  When it does, it makes sense to stop learning for a while.

jMyles:
As of #366, this is good enough.  Node Learning slows after no new nodes are learned for a while, and speeds back up when a new node is discovered.

Issue #347, opened 2018-07-03
Cleanup on aisle make_ursulas


jMyles:
`make_ursulas` has gotten very messy.  `bare` ? `miners`?  With the event loop at the beginning?  What a cluster.

PR #366, opened 2018-07-17
Learning Loop Part IV and Working n==3 Federated Demo.

### LINKED ISSUE #311: Provide proper eagerness for different learning scenarios.
Currently, whenever a `Character` learns about a node, that character proceeds to hit the `public_keys` endpoint of that node immediately, instead of waiting until it's actually necessary.

### LINKED ISSUE #332: New legality check for Ursula public info using public address
Now that we're tracking nodes by `canonical_public_address`, we need a legality check that uses that instead of their stamp.  This is necessary to close #180.

I'm temporarily hard-coding the legality check (for nodes only, not TreasureMaps) to True to get around this while we work on it.

### LINKED ISSUE #339: Include Alice's canonical_public_address with TreasureMap?
Currently, we include Alice's `canonical_public_address` (typically, her ethereum address) with the `TreasureMap` payload.  

Is this a good idea?

Pro: We can use it to authenticate her somehow, to ensure that she's "allowed" to publish a TreasureMap.

Con: It further de-anonymizes Alice.

Generally, I think it makes sense to remove it.

jMyles:
Freshens up federated logic to be totally up to speed with all learning loop changes and n>1 states.

Federated demo is now `n==3`, `m==2`.  Thus, to run it, you need to run `run_ursula_with_rest_and_dht_but_no_mining.py` three times.

Also:

* Introducing `TLSHostingPower`, which does exactly what it sounds like.
* Nodes now announce themselves when learning
* Nodes continue to learn even if something goes wrong with one teacher. 

Fixes #311
Fixes #332
Fixes #339

Partial #347 
Partial #363

### DIFF
--- a/examples/finnegans-wake-federated.py
+++ b/examples/finnegans-wake-federated.py
@@ -3,11 +3,12 @@
 
 # WIP w/ hendrix@3.0.0
 
+import binascii
 import datetime
+import logging
 import sys
 
 import maya
-from sandbox_resources import SandboxRestMiddleware
 
 from nucypher.characters import Alice, Bob, Ursula
 from nucypher.data_sources import DataSource
@@ -15,29 +16,43 @@ from nucypher.data_sources import DataSource
 from nucypher.network.middleware import RestMiddleware
 from umbral.keys import UmbralPublicKey
 
-URSULA = Ursula.from_rest_url(network_middleware=RestMiddleware(),
-                              host="localhost",
-                              port=3601,
-                              federated_only=True)
+root = logging.getLogger()
+root.setLevel(logging.DEBUG)
 
-network_middleware = SandboxRestMiddleware([URSULA])
+ch = logging.StreamHandler(sys.stdout)
+ch.setLevel(logging.INFO)
+formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+ch.setFormatter(formatter)
+root.addHandler(ch)
+
+teacher_dht_port = 3552
+teacher_rest_port = 3652
+with open("examples-runtime-cruft/node-metadata-{}".format(teacher_rest_port), "r") as f:
+    f.seek(0)
+    teacher_bytes = binascii.unhexlify(f.read())
+URSULA = Ursula.from_bytes(teacher_bytes, federated_only=True)
+print("Will learn from {}".format(URSULA))
+
+# network_middleware = SandboxRestMiddleware([URSULA])
 
 #########
 # Alice #
 #########
 
-ALICE = Alice(network_middleware=network_middleware,
+ALICE = Alice(network_middleware=RestMiddleware(),
               known_nodes=(URSULA,),  # in lieu of seed nodes
-              federated_only=True)  # TODO: 289
+              federated_only=True,
+              always_be_learning=True)  # TODO: 289
 
 # Here are our Policy details.
 policy_end_datetime = maya.now() + datetime.timedelta(days=5)
-m = 1
-n = 1
+m = 2
+n = 3
 label = b"secret/files/and/stuff"
 
 # Alice grants to Bob.
-BOB = Bob(known_nodes=(URSULA,), federated_only=True)
+BOB = Bob(known_nodes=(URSULA,), federated_only=True, always_be_learning=True)
+ALICE.start_learning_loop(now=True)
 policy = ALICE.grant(BOB, label, m=m, n=n,
                      expiration=policy_end_datetime)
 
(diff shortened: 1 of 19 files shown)


examples/finnegans-wake-federated.py
```diff
@@ -3,41 +3,56 @@
...
+ch.setLevel(logging.INFO)
+formatter = logging.Formatter('%(asctime)s - %(name)s - %(levelname)s - %(message)s')
+ch.setFormatter(formatter)
+root.addHandler(ch)
+
+teacher_dht_port = 3552
+teacher_rest_port = 3652
+with open("examples-runtime-cruft/node-metadata-{}".format(teacher_rest_port), "r") as f:
+    f.seek(0)
+    teacher_bytes = binascii.unhexlify(f.read())
```
tuxxy: You can actually just do `bytes.fromhex(f.read())`, there is no need for using binascii here.

examples/finnegans-wake-federated.py
```diff
@@ -3,41 +3,56 @@
...
-              federated_only=True)  # TODO: 289
+              federated_only=True,
+              always_be_learning=True)  # TODO: 289
 
 # Here are our Policy details.
 policy_end_datetime = maya.now() + datetime.timedelta(days=5)
-m = 1
-n = 1
+m = 2
+n = 3
```
tuxxy: 👍

nucypher/keystore/keypairs.py
```diff
@@ -17,26 +22,29 @@ class Keypair(object):
...
+
     def __init__(self,
-                 umbral_key: Union[UmbralPrivateKey, UmbralPublicKey] = None,
+                 private_key: Union[UmbralPrivateKey, UmbralPublicKey] = None,
                  generate_keys_if_needed=True):
         """
         Initalizes a Keypair object with an Umbral key object.
 
-        :param umbral_key: An UmbralPrivateKey or UmbralPublicKey
+        :param private_key: An UmbralPrivateKey or UmbralPublicKey
```
cygnusv: Why call this `private_key` if it can be an `UmbralPublicKey`?

nucypher/keystore/keypairs.py
```diff
@@ -17,26 +22,29 @@ class Keypair(object):
...
         try:
-            self.pubkey = umbral_key.get_pubkey()
-            self._privkey = umbral_key
+            self.pubkey = getattr(private_key, self._public_key_method)()
+            self._privkey = private_key
         except NotImplementedError:
-            self.pubkey = umbral_key
-            self._privkey = PUBLIC_ONLY
+            self.pubkey = private_key
+            self._privkey = constants.PUBLIC_ONLY
```
cygnusv: OK, I guess this is the Pythonic way to do it, but I find it very obscure. Why can you just do the same with a couple of `isinstance`? What if, for some strange/stupid reason, we decide later in pyUmbral that `UmbralPublicKey.get_pubkey()` returns `self` (which doesn't sound that stupid after all)?

nucypher/keystore/keypairs.py
```diff
@@ -103,14 +111,49 @@ def sign(self, message: bytes) -> bytes:
...
 
+
+class HostingKeypair(Keypair):
+    """
+    A keypair for TLS'ing.
+    """
+    _private_key_source = ec.generate_private_key
+    _public_key_method = "public_key"
+
+    _DEFAULT_CURVE = ec.SECP384R1
```
cygnusv: :+1:

nucypher/network/server.py
```diff
@@ -282,6 +282,7 @@ def reencrypt_via_rest(self, id_as_hex, request: http.Request):
         from nucypher.policy.models import WorkOrder  # Avoid circular import
         id = binascii.unhexlify(id_as_hex)
         work_order = WorkOrder.from_rest_payload(id, request.body)
+        self.log.info("Work Order from {}, signed {}".format(work_order.bob, work_order.receipt_signature))
         with ThreadedSession(self.db_engine) as session:
             kfrag_bytes = self.datastore.get_policy_arrangement(id.hex().encode(),
                                                                 session=session).k_frag  # Careful!  :-)
```
cygnusv: k_frag --> kfrag?

nucypher/characters.py
```diff
@@ -658,10 +734,26 @@ def grant(self, bob, uri, m=None, n=None, expiration=None, deposit=None, ursulas
         # Users may decide to inject some market strategies here.
         #
         # TODO: 289
+
+        # If we're federated only, we need to block to make sure we have enough nodes.
+        if self.federated_only and len(self._known_nodes) < n:
+            good_to_go = self.block_until_number_of_known_nodes_is(n, learn_on_this_thread=True)
+            if not good_to_go:
+                raise ValueError(
+                    "To make a Policy in federated mode, you need to know about\
```
michwill: I suspect, this will produce lots of spaces which you use for alignment, in the string every time you make a new line (or, at least, it does for me when I repeat this pattern in terminal).
Maybe better:
```
    raise ValueError(
        "Starting the message,"
        "continuing the message")
```

PR #372, opened 2018-07-27
[WIP] Verifying certificate when talking to node.

### LINKED ISSUE #60: Establish certificate-like authentication for public keys.
Is there currently a smart contract template for publishing and signing public key fingerprints?  Is there an implementation of TLS using this?

jMyles:
We've always used `verify=False` when talking to nodes via TLS.  But that's really not necessary. 

Fixes #60.


jMyles:
I like these tools for examining the TLS offering:

http://www.taddong.com/en/lab.html#TLSSLED

tuxxy:
@jMyles What are you using from the list?

jMyles:
Well TLSSLed (to which the link directly points) is a favorite of mine.  Easy reliable way to look at the actual TLS details for a HTTPS server.

jMyles:
Subsumed by #424.

PR #374, opened 2018-08-03
Multi-signature contract with off-chain signing

szotov:
Multi-signature contract for the upgrading management:
* Contract checks pre-signed transactions
* Allows to remove and add owners and change required number of signed transactions

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/MultiSig.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/MultiSig.sol
@@ -10,21 +10,35 @@ import "zeppelin/math/SafeMath.sol";
 contract MultiSig {
     using SafeMath for uint256;
 
+    event Executed(address indexed sender, uint256 indexed nonce, address indexed destination, uint256 value);
+    event OwnerAdded(address indexed owner);
+    event OwnerRemoved(address indexed owner);
+    event RequirementChanged(uint16 required);
+
+    uint256 constant public MAX_OWNER_COUNT = 50;
+
     uint256 public nonce;
-    uint16 public threshold;
+    uint8 public required;
     mapping (address => bool) public isOwner;
     address[] public owners;
+
+    // @notice Only this contract can call method
+    modifier onlyThisContract() {
+        require(msg.sender == address(this));
+        _;
+    }
 
     function () public payable {}
 
     /**
-    * @param _threshold Number of required signings
+    * @param _required Number of required signings
     * @param _owners List of initial owners.
     **/
-    constructor (uint16 _threshold, address[] _owners) public {
+    constructor (uint8 _required, address[] _owners) public {
         require(_owners.length > 0 &&
-        _threshold <= _owners.length &&
-        _threshold > 0);
+            _owners.length <= MAX_OWNER_COUNT &&
+            _required <= _owners.length &&
+            _required > 0);
 
         for (uint256 i = 0; i < _owners.length; i++) {
             address owner = _owners[i];
@@ -32,7 +46,7 @@ contract MultiSig {
             isOwner[owner] = true;
         }
         owners = _owners;
-        threshold = _threshold;
+        required = _required;
     }
 
     /**
@@ -74,13 +88,13 @@ contract MultiSig {
     )
         external
     {
-        require(_sigR.length == threshold &&
+        require(_sigR.length == required &&
             _sigR.length == _sigS.length &&
             _sigR.length == _sigV.length);
 
         bytes32 txHash = getUnsignedTransactionHash(_destination, _value, _data, nonce);
         address lastAdd = 0x0;
-        for (uint256 i = 0; i < threshold; i++) {
+        for (uint256 i = 0; i < required; i++) {
             address recovered = ecrecover(txHash, _sigV[i], _sigR[i], _sigS[i]);
             require(recovered > lastAdd && isOwner[recovered]);
             lastAdd = recovered;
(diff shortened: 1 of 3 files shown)


nucypher/blockchain/eth/sol/source/contracts/MultiSig.sol
```diff
@@ -0,0 +1,160 @@
...
+        bytes32[] _sigS,
+        address _destination,
+        uint256 _value,
+        bytes _data
+    )
+        external
+    {
+        require(_sigR.length == required &&
+            _sigR.length == _sigS.length &&
+            _sigR.length == _sigV.length);
```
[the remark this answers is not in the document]
cygnusv: It's related to the public key recovery process, but yeah, it's one byte.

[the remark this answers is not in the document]
szotov: Here length is number of owners that create signatures (required field).  And type of V is uint8 - 1 byte

szotov:
Several prerequisites for use multisig:
We have three contracts with upgrading and I think there will be more with time;
Somebody can steal keys, so need a voting or something;
Such management contrtact should be upgreadable or should be able to transfer ownership;
Also there are some functions that we can use in management contract - initialiazation of the MinerEscrow contract, for example.
All these thoughts lead to a multisig contract.
I found two multisig contracts  - with on-chain voting and with off-chain signing.
If we use blockchain for voting - than our upgrading secret will be visible a little bit more than with off-chain signing. And also off-chain signing contract has less methods and it looks like that there is less attack surface.

michwill:
Yeah, we totally should do whatever makes the attack surface smallest. But what I meant is something like:

```
function addVote() {
    ...
    if (msg.sender in allowed_list) {
       vote += 1;
    }
}

function startUprade {
    ...
    vote = 0;
}

function upgrade() {
    if (vote > threshold) {
        ... do the upgrade...
    }
    vote = 0;
}
```

In this case signatures are checked explicitly (when Ethereum checks ownerships of the contracts). But idk, maybe I am missing something here

szotov:
Voting example should be bigger, because we have multiple contracts to upgrade and number will change, plus upgrade of voting contract or transfer ownership, plus owners management.
Multisig example is another example with on-chain voting, but there should be methods for owners management and contract become not so simple:
https://github.com/gnosis/MultiSigWallet/blob/master/contracts/MultiSigWallet.sol

michwill:
Well, yes, the choice is between:
* having each "vote" in its own transaction (the plus here is that we don't have to touch raw signatures, `ecrecover` and whatever);
* off-chain signing for the upgrade

nucypher/blockchain/eth/constants.py
```diff
@@ -50,14 +62,22 @@ class MinerConfigError(ValueError):
...
-    MAX_ALLOWED_LOCKED
+# Mining formula for one stake in one period
+# (totalSupply - currentSupply) * (lockedValue / totalLockedValue) * (k1 + allLockedPeriods) / k2
+# totalSupply - Token supply cap
+# currentSupply - Current supply cap
+# lockedValue - Amount of tokens in one miner's stake in one period
+# totalLockedValue - Amount of tokens in all stakes in one period
+# allLockedPeriods - Duration of the current miner's stake
+# if allLockedPeriods > rewardedPeriods then allLockedPeriods = rewardedPeriods
+__mining_coeff = (
```
KPrasch: Phew! Finally

nucypher/blockchain/eth/sol/source/contracts/MinersEscrow.sol
```diff
@@ -92,13 +92,13 @@ contract MinersEscrow is Issuer {
     * @param _token Token contract
     * @param _hoursPerPeriod Size of period in hours
     * @param _miningCoefficient Mining coefficient
-    * @param _minLockedPeriods Min amount of periods during which tokens will be locked
+    * @param _minLockedPeriods Min amount of periods during which tokens can be locked
```
KPrasch: Good cactch

nucypher/blockchain/eth/sol/source/contracts/MultiSig.sol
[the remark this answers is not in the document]
KPrasch: @cygnusv / @tuxxy bump - Can you take a second look here before we move forward?

nucypher/blockchain/eth/sol/source/contracts/proxy/README.MD
```diff
@@ -30,21 +30,9 @@ New versions of contract can expand values, but must contain all old values (fir
...
-
-# Interaction order
-* Admin creates contract_v1 and creates dispatcher using contract_v1 address
-* Client instantiates contract_v1 using dispatcher address and sends requests
-* Admin or any client creates contract_v2 and creates voting in the government contract
-* All clients vote for or against upgrade during the voting time
-* After successful voting admin or any client commits upgrade in the government contract
-* Client uses old instance without changing (if the methods are not changed)
 
 # Development
```
KPrasch: Where do we want this documentation to be published?  It seems isolated from the rest of the codebase

nucypher/blockchain/eth/sol/source/contracts/proxy/Upgradeable.sol
```diff
@@ -19,11 +19,17 @@ contract Upgradeable is Ownable {
...
+
     /**
     * @dev Previous contract address (if available). Used for rollback
     **/
     address public previousTarget;
 
+    /**
+    * @dev Secret hash to proof that user owns previous version of a contract
+    **/
+    bytes32 public secretHash;
```
KPrasch: Heh - This reads funny.  @szotov - Can you clarify what is being exposed here?

nucypher/blockchain/eth/sol/source/contracts/proxy/README.MD
szotov: It's really old docs, maybe move it to wiki or remove it?

nucypher/blockchain/eth/sol/source/contracts/proxy/Upgradeable.sol
szotov: Maybe previous version of contract is not the best name
Who owns contract that already deployed also should know secret (proof of owning) to use it in upgrade

michwill:
So, when do we merge this?

EvaLovlac3:
Like others have said, needs auditing before merge.

nucypher/blockchain/eth/sol/source/contracts/MultiSig.sol
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+    * @param _required Number of required signings
+    * @param _owners List of initial owners.
+    **/
+    constructor (uint8 _required, address[] memory _owners) public {
+        require(_owners.length > 0 &&
+            _owners.length <= MAX_OWNER_COUNT &&
+            _required <= _owners.length &&
+            _required > 0);
```
incertia: minor detail, but there is `_owners.length > 0` when it already requires `0 < _required <= _owners.length <= MAX_OWNER_COUNT`

nucypher/blockchain/eth/sol/source/contracts/MultiSig.sol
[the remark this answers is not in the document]
incertia: this should be fine, especially since it's technically an array of bytes and the value of `v` depends on the parity of some number.

nucypher/blockchain/eth/sol/source/contracts/MultiSig.sol
szotov: 👍

KPrasch:
@szotov - Any remaining changes before final review and merge?

szotov:
No, everything done and rebased

Issue #200, opened 2018-04-10
[CLOSED] setup.sh causes `cytoolz` version conflict (populus)


jMyles:
<a href="https://github.com/KPrasch"><img src="https://avatars1.githubusercontent.com/u/679404?v=4" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [KPrasch](https://github.com/KPrasch)**
_Friday Jan 26, 2018 at 00:32 GMT_
_Originally opened as https://github.com/nucypher/nucypher-kms-ethereum/issues/4_

----

Similar ticket #416 @populus: https://github.com/ethereum/populus/issues/416

```
 py.test
...
...
_pytest.vendored_packages.pluggy.PluginValidationError: Plugin 'ethereum' could not be loaded:
 (cytoolz 0.8.2 (/home/kieran/Git/nucypher-kms-ethereum/.venv/lib/python3.6/site-packages),
Requirement.parse('cytoolz<1.0.0,>=0.9.0'), {'eth-keyfile', 'eth-keys', 'eth-tester'})!

```

jMyles:
<a href="https://github.com/KPrasch"><img src="https://avatars1.githubusercontent.com/u/679404?v=4" align="left" width="48" height="48" hspace="10"></img></a> **Comment by [KPrasch](https://github.com/KPrasch)**
_Friday Jan 26, 2018 at 00:47 GMT_

----

Test suite runs and passes after upgrading cytoolz to version 0.9.0.

```bash
source .venv/bin/activate
pip install cytoolz --upgrade
...
py.test
```

Issue #23, opened 2017-09-05
Seed-only protocol saves the data sometimes in tests


michwill:
How to reproduce: run py.test multiple times
```
(py35) michwill ~/Projects/nucypher-kms $ py.test                           
===================================================================== test session starts =====================================================================
platform linux -- Python 3.5.2, pytest-3.2.1, py-1.4.34, pluggy-0.4.0
rootdir: /home/michwill/Projects/nucypher-kms, inifile: tox.ini
plugins: cov-2.5.1
collected 13 items                                                                                                                                             

tests/test_client.py ........
tests/test_default_crypto.py ..
tests/test_kmac.py ..
tests/test_network.py F

========================================================================== FAILURES ===========================================================================
_____________________________________________________________________ test_seed_only_node _____________________________________________________________________

    def test_seed_only_node():
        """
        Shows that when we set up two nodes, a "full" node and a "seed-only" node,
        that the "seed-only" node can set key-value pairs that the "full" node will store,
        but not vice-versa.
        """
        loop = asyncio.get_event_loop()
    
        # First, let's set up two servers:
        # A full node...
        full_server = Server()
        full_server.listen(8468)
        loop.run_until_complete(full_server.bootstrap([("127.0.0.1", 8468)]))
    
        # ...and a seed-only node.
        seed_only_server = NuCypherSeedOnlyDHTServer()
        seed_only_server.listen(8471)
        loop.run_until_complete(seed_only_server.bootstrap([("127.0.0.1", 8468)]))
    
        # The seed-only node is able to set a key and retrieve it again.
        key_to_store = "llamas"
        value_to_store = "tons_of_things_keyed_llamas"
        setter = seed_only_server.set(key_to_store, value_to_store)
        loop.run_until_complete(setter)
    
        # Now, the item is stored on the full server.
        full_server_stored_items = list(full_server.storage.items())
        assert len(full_server_stored_items) == 1
        assert full_server_stored_items[0][1] == value_to_store
    
        # ...but nothing is stored on the seed-only server.
        seed_only_server_stored_items = list(seed_only_server.storage.items())
>       assert len(seed_only_server_stored_items) == 0
E       AssertionError: assert 1 == 0
E        +  where 1 = len([(b'\xf2\xe2\xd8D\xb3\xe0Ma\x10\x9cM\xea\xd6\xe1!\xbf\xbd\x98\xb0\xa3', 'tons_of_things_keyed_llamas')])

[...]

PR #377, opened 2018-08-11
Nucypher Configuration Parsers

KPrasch:
Changes the way file-based configuration is handled for initialising Ursula:

- Deprecates old style config classes
- Introduces .ini config and parsers

This PR is also the beginning of a chain of PRs that has a similar history to PR #370 that aims to builds a command line interface for "spinning-up" Ursula, setting-up node configurations, managing stakes, gleaning network status and executing network simulations.

#377 : Config file and parsers
#378 nucypher-cli prototype and Ursula.stake
#379 CLI + Configuration + Parser Integration
#380 CLI + Blockchain Agency API Integration
#381 Follow-up
#382 Follow-up again

Depends on nucypher #366 (Merged)
Depends on coincurve #23 ofek/coincurve#23 (Merged)
Depends on pyUmbral #200 nucypher/pyUmbral#200 (Merged)


nucypher/config/keys.py
```diff
@@ -15,7 +16,7 @@
 from web3.auto import w3
 
 from nucypher.config import utils
-from nucypher.config.configs import _DEFAULT_CONFIGURATION_DIR
+from nucypher.config.configs import _DEFAULT_CONFIGURATION_DIR, NucypherConfiguration
```
cygnusv: Is `nucypher.config.configs` coming in a later PR? It seems it doesn't exist right now

KPrasch: Hmm, ultimately its been deprecated.  Looks like this is cruft from the refactoring.

nucypher/characters.py
```diff
@@ -68,7 +67,7 @@ def __init__(self, is_me=True,
                  crypto_power: CryptoPower = None,
                  crypto_power_ups=None,
                  federated_only=False,
-                 config: CharacterConfiguration = None,
+                 config = None,
```
tuxxy: 👍

nucypher/config/utils.py
```diff
@@ -134,3 +55,165 @@ def check_config_runtime() -> bool:
...
+
+    if config is None:
+        config = configparser.ConfigParser()
+
+    try:
+        config.read(filepath)
+    except:
+        raise  # FIXME
+
+    required_sections = ("provider", "nucypher")
```
tuxxy: Neat! 👍

nucypher/config/utils.py
```diff
@@ -134,3 +55,165 @@ def check_config_runtime() -> bool:
...
+    else:
+        raise NotImplementedError
+
+    poa = config.getboolean(section='provider', option='poa', fallback=True)
+    tester = config.getboolean(section='blockchain', option='tester', fallback=False)
+    test_accounts = config.getint(section='blockchain', option='test_accounts', fallback=0)
+    deploy = config.getboolean(section='blockchain', option='deploy', fallback=False)
+    compile = config.getboolean(section='blockchain', option='compile', fallback=False)
+    timeout = config.getint(section='blockchain', option='timeout', fallback=10)
+    tmp_registry = config.getboolean(section='blockchain', option='temporary_registry', fallback=False)
```
tuxxy: Also cool 👍

nucypher/config/keys.py
cygnusv: I checked the top of #382 and it seems that `nucypher/config/configs.py` is still gone there, so it seems `nucypher/config/keys.py` it's outdated?

KPrasch: Yes - as mentioned, `configs.py` is ultimately deprecated; however config/keys.py does exist.

KPrasch: It looks like the `config.keys.py` module has is not being executed in automated testing! 

Good catch

nucypher/config/utils.py
```diff
@@ -1,104 +1,25 @@
...
-    """Parses a keyfile and returns key metadata as a dict."""
-
-    with open(keypath, 'r') as keyfile:
-        try:
-            key_metadata = json.loads(keyfile)
-        except json.JSONDecodeError:
-            raise NucypherConfiguration.NucypherConfigurationError("Invalid data in keyfile {}".format(keypath))
-        else:
-            return key_metadata
+DEFAULT_CONFIG_DIR = "~"
```
michwill: If we don't do it already somewhere else, we probably would want to leave a `# TODO` to leave it somewhere else than `~` (`.local` or whatever)

nucypher/config/utils.py
```diff
@@ -134,3 +55,165 @@ def check_config_runtime() -> bool:
...
+            provider = IPCProvider(config['provider']['ipc_path'])
+        except KeyError:
+            message = "ipc_path must be provided when using an IPC provider"
+            raise Exception(message)  # FIXME
+        else:
+            providers.append(provider)
+    else:
+        raise NotImplementedError
+
+    poa = config.getboolean(section='provider', option='poa', fallback=True)
```
michwill: It feels to me that we're writing a schema with datatypes by doing the `getboolean` etc, which kinda suggests that we need some facility for that (hopefully, not XML lol). I think, could be not a bad idea to write the schema in Python.

But I'd leave this to the future: masering configurability of configs is not the focus of what we need right now. Can leave a little `# TODO` note though.

For now, this is great stuff

PR #420, opened 2018-09-01
Enforce MyPy Checks

### LINKED ISSUE #350: Tone-down mypy checking
When we merge the current outstanding set of PRs, with a presumed resolution order of #329, #345, #338,  #348, #333, #286, #275, we'll have passing tests but a failing build by dint of the failing mypy check.

Since we are aiming for a merge of coffee into master very soon (perhaps before the end of the week), let's comment-out the mypy check so that this merge does not appear in the record as a regression.

Then, when the time is right to actually fix all of the issues about which mypy complains (and to be fair, it's possible that that time is now), let's open a *separate* PR which both fixes these things and uncomments the mypy check.

KPrasch:
Fixes #350 . Opens #421 

Annotates all `__init__` functions with None... see: https://github.com/python/mypy/issues/604

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -171,7 +171,7 @@ jobs:
           command: |
               mkdir ./mypy_reports ./mypy_results
               export MYPYPATH=./nucypher
-              pipenv run mypy ./nucypher --config-file=mypy.ini --xslt-html-report ./mypy_reports/html/ --linecount-report ./mypy_reports/linecount --junit-xml ./mypy_results/results.xml || true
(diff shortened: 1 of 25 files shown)


jMyles:
@michwill: There is.  Just leave it the helloot.  :-)

KPrasch:
er... is there a way to turn off that warning? I was unable to find a way; although I will be glad if there is such an option.  Perhaps we use a different tool altogether, but in the meantime mypy is picking up plenty of legitimate errors.

KPrasch:
Moved to #439

PR #424, opened 2018-09-06
Certificates; "Kirby" pattern for RESTServer, other changes.

jMyles:
**Note**: This branch recently underwent a substantial cherry-pick.  Currently, the test suite fails with `AttributeError: type object 'PyEVMBackend' has no attribute 'from_genesis_overrides'` and it's not obvious to me how to identify the interplay between the current state and @KPrasch's work in this area.

----
This PR:

* Establishes certificates for nodes.
* Verifies certificates for nodes.
* Signs certificates using the `SigningPower`, and also signs the `stamp` with the blockchain keypair.

* Removes `ProxyRESTServer` as a mixin; instead, this is "eaten" by `TLSHostingPower`, which is then consumed as a regular power up, turning Ursula into a rest server.

Small changes:
* Some test failures now happen more quickly with the introduction of a graceful crashing facility. 
* Ursula's rest information is now provided via a standard public method, `rest_information`.


nucypher/network/server.py
```diff
@@ -101,39 +102,52 @@ def __init__(self, *args, **kwargs):
...
-                 rest_port=None,
-                 db_name=None,
-                 tls_private_key=None,
-                 tls_curve=None,
-                 certificate=None,
-                 *args, ** kwargs) -> None:
+                 rest_host,
+                 rest_port,
+                 routes=None,
+                 ) -> None:
```
michwill: oh mypy...

PR #428, opened 2018-09-10
Kokonusswasser

jMyles:
Remove the DHT, and Run a node from the NuCypher CLI


cli/main.py
```diff
@@ -58,15 +58,18 @@
...
+        # click.echo("Using configuration filepath {}".format(self.config_filepath))
+
+        self.operating_mode = operating_mode
+        self.simulation_mode = simulation_mode
 
         # Set operating and run modes
-        operating_modes = parse_running_modes(filepath=self.config_filepath)
-        self.operating_mode = operating_modes.get('mode', 'decentralized')
-        self.simulation_mode = operating_modes.get('simulation', False)
+        if use_config is True:
```
cygnusv: Just nitpicking, but this expression seems redundant. Perhaps just `if use_config:` ?

cygnusv: The same patterns seems to be repeating below.

cli/demos/finnegans-wake-demo.py
```diff
@@ -0,0 +1,165 @@
+# This is an example of Alice setting a Policy on the NuCypher network.
```
cygnusv: So cool to see this in the CLI !! 👏

cli/main.py
```diff
@@ -69,8 +84,9 @@ def __init__(self):
         sim_mode = 'simulation' if self.simulation_mode else 'live'
         click.echo("Running in {} {} mode".format(sim_mode, self.operating_mode))
 
-        self.payload = parse_nucypher_ini_config(filepath=self.config_filepath)
-        click.echo("Successfully parsed configuration file")
+        if use_config is True:
```
tuxxy: Can this not be moved under the condition on L70?

KPrasch: This is deprecated upstream, it was just a holdover for the hackathon.

tuxxy: :+1:

cli/main.py
KPrasch: Deprecated upstream, but I generally adopted this pattern in case the user unintentionally passes a non-boolean type.

PR #431, opened 2018-09-13
In-memory cli configuration options; Split `make_ursulas`

### LINKED ISSUE #347: Cleanup on aisle make_ursulas

KPrasch:
Fixes #347


nucypher/config/constants.py
```diff
@@ -1,3 +1,23 @@
...
+
+import nucypher
+
+# Base Filepaths
+BASE_DIR = abspath(dirname(dirname(nucypher.__file__)))
+PROJECT_ROOT = abspath(dirname(nucypher.__file__))
+APP_DIR = AppDirs("nucypher", "NuCypher")
+
+# Configuration File
+TEMPLATE_CONFIG_FILE_LOCATION = os.path.join(BASE_DIR, 'cli', '.nucypher.ini')
```
michwill: We should keep in mind that the `.ini` file will unlikely be alongside the code if `nucypher` is installed with `setup.py` into system directories. Ok for now, but we'll come across this once we start testing with properly installed `nucypher` package

michwill: Perhaps, easy enough to fix even in this PR: just need to include `.nucypher.ini` in `package_data` in `setup.py`. We already have Solidity files there

nucypher/config/node.py
```diff
@@ -1,106 +1,100 @@
...
-        self.known_metadata_dir = known_metedata_dir
+
+        if temp is True:
+            self.temp = True
+            self.temp_dir = TemporaryDirectory()
+            self.config_root = self.temp_dir.name
+        else:
+            self.config_root = config_root
+
+        self.operating_mode = 'federated'
```
michwill: You probably mean `self.operating_mode = operating_mode`?

nucypher/utilities/sandbox/blockchain.py
```diff
@@ -0,0 +1,131 @@
...
+    _test_account_cache = list()
+
+    def __init__(self, test_accounts=None, poa=True, airdrop=True, *args, **kwargs):
+
+        # Depends on circumflex
+        super().__init__(*args, **kwargs)
+
+        # For use with Proof-Of-Authority test-blockchains
+        if poa is True:
+            w3 = self.interface.w3
```
michwill: In recent web3, one can do `w3 = web3.auto.gethdev.w3` (although, manually injecting into `middleware_stack` should also continue to work... Until high velocity breaks it again!)

nucypher/utilities/sandbox/ursula.py
```diff
@@ -0,0 +1,166 @@
...
+    else:
+        starting_port = max(TEST_KNOWN_URSULAS_CACHE.keys()) + 1
+
+    federated_ursulas = set()
+    for port in range(starting_port, starting_port+quantity):
+
+        ursula = Ursula(is_me=True,
+                        db_name="test-{}".format(port),
+                        rest_host="localhost",
+                        rest_port=port + 100,
```
michwill: If we don't have a DHT anymore, why `+ 100`?

nucypher/config/constants.py
KPrasch: I'm using this hidden file as a 'template' to write new configuration file via the CLI (perhaps interactively at some point) and a method: `NodeConfiguration.write_default_config_file(filepath=/my/config/file/path/)`.  If I understand correctly, if we ship a required default config file then we'll need to install it with `setup.py`, but for now it's optional.

nucypher/config/node.py
KPrasch: Yes - Fixed upstream.

nucypher/utilities/sandbox/blockchain.py
KPrasch: This is only for supporting test runs with geth... at the moment - Ill look at the `web3.auto.gethdev.w3` API.
```
        # For use with Proof-Of-Authority test-blockchains
        if poa is True:
            w3 = self.interface.w3
            w3.middleware_stack.inject(geth_poa_middleware, layer=0)
```

nucypher/utilities/sandbox/ursula.py
KPrasch: Hmm, good point. I'll change it.

nucypher/config/node.py
```diff
@@ -1,106 +1,100 @@
...
+            except FileExistsError:
+                raise
+            except FileNotFoundError:
+                raise
+
+        #
+        # Create Config Subdirectories
+        #
 
+        os.mkdir(self.keyring_dir, mode=0o700)               # keyring
```
tuxxy: :+1:

nucypher/utilities/sandbox/middleware.py
```diff
@@ -0,0 +1,96 @@
...
+    def put_treasure_map_on_node(self, node, map_id, map_payload):
+        mock_client = self._get_mock_client_by_ursula(node)
+        response = mock_client.post("http://localhost/treasure_map/{}".format(map_id),
+                                    data=map_payload, verify=False)
+        return response
+
+
+class EvilMiddleWare(MockRestMiddleware):
+    """
+    Middleware for assholes.
```
tuxxy: :joy:

nucypher/utilities/sandbox/ursula.py
KPrasch: Changed upstream (to avoid additional rebases).

PR #436, opened 2018-09-14
Add timestamp on Ursula

tuxxy:
WIP

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -368,7 +368,8 @@ class Bob(Character):
 
 
 class Ursula(Character, VerifiableNode, Miner):
-    _internal_splitter = BytestringSplitter(Signature,
+    _internal_splitter = BytestringSplitter((int, 4, {'byteorder': 'big'}),
+                                            Signature,
                                             VariableLengthBytestring,
                                             (UmbralPublicKey, PUBLIC_KEY_LENGTH),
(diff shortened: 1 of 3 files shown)


nucypher/characters/lawful.py
```diff
@@ -544,7 +545,10 @@ def __bytes__(self):
         certificate = self.rest_server_certificate()
         cert_vbytes = VariableLengthBytestring(certificate.public_bytes(Encoding.PEM))
 
-        as_bytes = bytes().join((bytes(self._interface_signature),
+        timestamp = maya.now().epoch.to_bytes(4, 'big')
```
KPrasch: Cool!  Looks like we're using epoch as to version ursula *each* time she is serialized?

tuxxy: Yes, that's what @jMyles said to do. I think this was more of a thing when we hadn't removed the DHT, though.

KPrasch: Yeah - its good. :+1:

KPrasch:
Looks like there is a bug in the CircleCI configuration causing the `eth_contracts` job to run the entire test suite.  In any case, the test run produced another `ValueError`: 
https://circleci.com/gh/nucypher/nucypher/2465#queue-placeholder/containers/0

nucypher/characters/lawful.py
jMyles: I gotta think more about this.

PR #442, opened 2018-09-20
--dev flag for `run_ursula` 

KPrasch:
Uses the new UrsulaConfiguration API to integrate the temporary storage area and in-memory database with the `run_ursula` command, switched on /off with the `--dev`.

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -135,17 +135,14 @@ def configure(config, action, config_file, config_root, temp):
     """Manage the nucypher .ini configuration file"""
 
     if config_root:
-        node_configuration = NodeConfiguration(config_root=config_root)
+        node_configuration = NodeConfiguration(config_root=config_root, auto_initialize=False)
     elif temp:
-        node_configuration = NodeConfiguration(temp=temp, auto_initialize=True)
+        node_configuration = NodeConfiguration(temp=temp, auto_initialize=False)
     elif config_file:
         click.echo("Using configuration file at: {}".format(config_file))
         node_configuration = NodeConfiguration.from_configuration_file(filepath=config_file)
     else:
-        node_configuration = NodeConfiguration()
-
-    filepaths = node_configuration._generate_runtime_filepaths(commit=False)
-    click.echo("Running in directory: {}".format(config_root or filepaths['config_root']))
+        node_configuration = NodeConfiguration()  # Fully Default
 
     def __destroy():
         click.confirm("Permanently destroy all nucypher configurations, known nodes, certificates and keys?", abort=True)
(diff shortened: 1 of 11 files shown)


cli/main.py
```diff
@@ -697,13 +697,16 @@ def run_ursula(rest_port,
     but can be overridden (mostly for testing purposes) with inline cli options.
 
     """
+    if not dev:
+        click.echo("WARNING: Development mode is disabled")
```
tuxxy: :+1:

nucypher/config/characters.py
```diff
@@ -121,7 +130,7 @@ def produce(self, **overrides):
         ursula = Ursula(**merged_parameters)
 
         if self.temp:
-            class MockDatastoreThreadPool(object):
+            class MockDatastoreThreadPool(object):  # TODO: Does this belong here..?
```
tuxxy: no heh

nucypher/config/node.py
```diff
@@ -8,11 +8,14 @@
 from itertools import islice
 
 from nucypher.config.constants import DEFAULT_CONFIG_ROOT, DEFAULT_CONFIG_FILE_LOCATION, TEMPLATE_CONFIG_FILE_LOCATION
+from nucypher.network.middleware import RestMiddleware
 
 
 class NodeConfiguration:
 
+    DEFAULT_OPERATING_MODE = 'decentralized'
```
tuxxy: "It's that easy!"

nucypher/config/node.py
```diff
@@ -28,43 +31,63 @@ def __init__(self,
 
                  checksum_address: str = None,
                  is_me: bool = True,
-                 federated_only: bool = False,
-                 network_middleware=None,
+                 federated_only: bool = None,
```
tuxxy: This is weird, why not make it `False`? `bool(None) == False`

nucypher/config/node.py
```diff
@@ -116,18 +137,36 @@ def payload(self):
...
+    def _generate_runtime_filepaths(self, commit=True) -> dict:
+        """Dynamically generate paths based on configuration root directory"""
+        if self.temp and commit and self.config_root is constants.UNINITIALIZED_CONFIGURATION:
+            raise self.ConfigurationError("Cannot pre-generate filepaths for temporary node configurations.")
+        filepaths = dict(config_root=self.config_root,
+                         keyring_dir=os.path.join(self.config_root, 'keyring'),
+                         known_nodes_dir=os.path.join(self.config_root, 'known_nodes'),
+                         known_certificates_dir=os.path.join(self.config_root, 'certificates'),
+                         known_metadata_dir=os.path.join(self.config_root, 'metadata'))
+        if commit:
```
tuxxy: I keep seeing this, not really apparent what `commit` is a boolean for. Just that it sets some attrs?

KPrasch: Yeah - I think I'm going to walk back this Idea a bit, but its so that you can peek at config filepaths before setting (and writing) them.

nucypher/config/node.py
KPrasch: it's the expected type, and It can be set to None by a subclass method... this "override passthrough" can be tricky stuff.

tuxxy: :/

PR #446, opened 2018-09-21
Cleaned up cert handling; demo runs again.

jMyles:


nucypher/policy/models.py
```diff
@@ -104,26 +106,32 @@ class Policy:
...
         """
         :param kfrags:  A list of KFrags to distribute per this Policy.
         :param label: The identity of the resource to which Bob is granted access.
         """
-        self.alice = alice
-        self.label = label
-        self.bob = bob
-        self.kfrags = kfrags
+        self.alice = alice                     # type: Alice
+        self.label = label                     # type: bytes
```
michwill: Should these comments be type hints, or we want to keep them flexible?

tests/cli/test_config.py
```diff
@@ -9,28 +11,50 @@
...
-    assert result.exit_code == 0
+    try:
+        os.mkdir(custom_filepath)  # TODO: Create a top-level installation directory, instead of writing only subdirs
 
-    args = ['configure', 'destroy', '--config-root', temp_config_root]
-    result = runner.invoke(cli, args, input='Y', catch_exceptions=False)
-    assert temp_config_root in result.output
-    assert result.exit_code == 0
+        args = ['configure', 'init', '--config-root', custom_filepath]
+        result = runner.invoke(cli, args, input='Y', catch_exceptions=False)
```
michwill: Great, finally testing the CLI!

jMyles: Total side note: I think it's a good idea to have debuggable entry points that are built with this syntax.  It's been great for me to poke around and learn.

nucypher/policy/models.py
jMyles: Something strange is going on - I don't even show this material on the diff.  If I click the filename above (`nucypher/policy/models.py`), I am taken to the diff page, but this file doesn't appear.  Am I the only one?

jMyles: ...OK, so it looks like it's *not* actually part of this diff, as it was merged on Friday as 4c02e5cd9e6f41621ea3af696c9a4e31794f15f3 with #444.  

And no, a base-swap doesn't fix the issue.

Github, you're looking increasingly unstable.

michwill: If I remember correctly, it was not shown as a part of the diff for me, but I still had an opportunity to comment

PR #450, opened 2018-09-24
[WIP] Return 400 if arrangements aren't signed

tuxxy:
WIP


nucypher/network/server.py
```diff
@@ -191,8 +191,8 @@ def set_policy(self, id_as_hex, request: http.Request):
         try:
             cleartext = self._verifier(alice, policy_message_kit, decrypt=True)
         except self.InvalidSignature:
-            # TODO: What do we do if the Policy isn't signed properly?
-            pass
+            # TODO: Perhaps we log this?
+            return Response(status_code=400)
```
KPrasch: Wiki on HTTP 400: 
> 400 Bad Request
The server cannot or will not process the request due to an apparent client error (e.g., malformed request syntax, size too large, invalid request message framing, or deceptive request routing).[33]

PR #457, opened 2018-09-28
[WIP] Simulate CLI cleanup

KPrasch:


cli/main.py
```diff
@@ -500,22 +504,23 @@ def simulate(config, action, nodes, federated_only):
             python_exec = 'python'
 
             proc_params = '''
-            python3 {} run_ursula --host {} --rest-port {} --db-name {}
-            '''.format(python_exec, cli_exec, localhost, rest_port, db_name).split()
+            python3 {} run_ursula --rest-port {} --db-name {}
+            '''.format(python_exec, cli_exec, rest_port, db_name).split()
```
tuxxy: As long as these are never specified or accessible to remote operators, this is okay. Otherwise, this string is vulnerable. Technically, this could be considered a local arbitrary code execution vulnerability.

KPrasch: Yes... if local filesystem access is in the threat model (and the attacker is able to replace the user's system python binary).

KPrasch: Also - this is in the context of a multi-proc local simulation.

PR #467, opened 2018-10-02
Protection against node metadata replay attacks.

jMyles:
Previously, it was possible for a chaotic actor to replay old node metadata, causing out-of-date information, including possibly public keys for compromised keypairs, to propagate across the network.

This PR address this in the following way:
* Ursula now creates a new timestamp whenever she signs her interface info (building on @tuxxy's previous work on timestamp)
* When nodes are remembered, their timestamp is kept as an attribute.
* When a node is passed to `remember_node` it is only remembered if it is the newest timestamp among those nodes.

Other things:

* Learners now fully verify nodes at "remember time" (recall that, if the learning loop is not run in eager mode, they only validate the metadata, rather than fully verifying the node during learning).  (This essentially nullifies eagerness for the moment, so it probably makes sense to move this logic further down the line during testnet).
* Vladimir now steals the timestamp in addition to the interface signature.

### DIFF
--- a/nucypher/characters/unlawful.py
+++ b/nucypher/characters/unlawful.py
@@ -1,3 +1,4 @@
+import maya
 from eth_tester.exceptions import ValidationError
 
 from nucypher.characters.lawful import Ursula
--- a/nucypher/characters/unlawful.py
+++ b/nucypher/characters/unlawful.py
@@ -35,7 +35,9 @@ class Vladimir(Ursula):
                        certificate=target_ursula.rest_server_certificate(),
                        is_me=False)
 
-        vladimir._interface_signature_object = target_ursula._interface_signature_object  # Asshole.
+        # Asshole.
+        vladimir._interface_signature_object = target_ursula._interface_signature_object
+        vladimir._timestamp = target_ursula._timestamp
 
         cls.attach_transacting_key(blockchain=target_ursula.blockchain)
 
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -79,7 +79,8 @@ class VerifiableNode:
         """
         Checks that the interface info is valid for this node's canonical address.
         """
-        message = self._signable_interface_info_message()  # Contains canonical address.
+        interface_info_message = self._signable_interface_info_message()  # Contains canonical address.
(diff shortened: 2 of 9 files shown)


nucypher/network/nodes.py
```diff
@@ -142,19 +149,32 @@ def _signable_interface_info_message(self):
...
+    @property
+    def timestamp(self):
+        if not self._timestamp:
+            try:
+                self._sign_and_date_interface_info()
+            except NoSigningPower:
+                raise NoSigningPower("This Node is a Stranger; you didn't init with a timestamp, so you can't verify.")
+        return self._timestamp
+
+    def timestamp_bytes(self):
```
tuxxy: Does this really make sense as a method like this? I see this more as a helper/utility type function and not necessarily in this class.

nucypher/characters/unlawful.py
```diff
@@ -34,7 +35,9 @@ def from_target_ursula(cls, target_ursula, claim_signing_key=False):
                        certificate=target_ursula.rest_server_certificate(),
                        is_me=False)
 
-        vladimir._interface_signature_object = target_ursula._interface_signature_object  # Asshole.
+        # Asshole.
```
tuxxy: :joy:

tests/network/test_network_actors.py
```diff
@@ -176,3 +177,28 @@ class FakeArrangement:
...
+
+    # Indeed, her metadata is not the same now.
+    assert bytes(ursula) != old_metadata
+
+    old_ursula = Ursula.from_bytes(old_metadata, federated_only=True)
+
+    # Once Alice learns about Ursula's updated info...
+    federated_alice.remember_node(ursula)
+
+    # ...she can't learn about old ursula anymore.
```
tuxxy: Way cool. :+1:

nucypher/network/nodes.py
jMyles: Yeah, I wondered the same thing.  But it gets called three times (four if you include the test), so I figure it's at least a good thing to have somewhere.  But yeah, if we use it anywhere else, I think a utility function makes more sense.

PR #473, opened 2018-10-07
Remove Kademlia digest function 

### LINKED ISSUE #472: Remove dependency on kademlia
The DHT has been deprecated from the codebase, but the dependency on kademlia remains for some of it's utilities.

tuxxy:
### What this does:
1. Removes the Kademlia `digest` function and uses `nucypher.crypto.api.keccak_digest` instead.
2. Calls `unhexlify` on the `treasure_map_id` to hash the raw bytes instead of the digest.
3. Removes `rpcudp` dependency
4. Resolves #472 #474

### DIFF
--- a/nucypher/network/protocols.py
+++ b/nucypher/network/protocols.py
@@ -1,4 +1,10 @@
 from bytestring_splitter import VariableLengthBytestring
+from constant_sorrow import default_constant_splitter, constants
+from kademlia.node import Node
+from kademlia.protocol import KademliaProtocol
+
+from nucypher.crypto.api import keccak_digest
+from nucypher.network.routing import NucypherRoutingTable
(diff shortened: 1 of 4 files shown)


nucypher/network/protocols.py
```diff
@@ -1,4 +1,10 @@
 from bytestring_splitter import VariableLengthBytestring
+from constant_sorrow import default_constant_splitter, constants
+from kademlia.node import Node
+from kademlia.protocol import KademliaProtocol
+
+from nucypher.crypto.api import keccak_digest
+from nucypher.network.routing import NucypherRoutingTable
```
KPrasch: Remove this? Also the Pipenv dependency?

PR #476, opened 2018-10-09
Configurable Node Storages

KPrasch:
Introduces swappable `NodeStorage` backends with the following subclasses:
- `NodeStorage(ABC)`
- `InMemoryNodeStorage(NodeStorage)`
- `LocalFileBaseNodeStorage(NodeStorage)`
- `TemporaryFileBasedNodeStorage(LocalFileBasedNodeStorage)`
- `S3NodeStorage(NodeStorage)`

Also includes integration with the AWS S3 mocking library `moto` for testing.


cli/main.py
```diff
@@ -81,6 +74,12 @@ def echo_version(ctx, param, value):
 # CLI Configuration
 #
 
+# CLI Constants
+DEBUG = True
+KEYRING_PASSPHRASE_ENVVAR = 'NUCYPHER_KEYRING_PASSPHRASE'
```
tuxxy: Hmm, maybe name this a bit different or set it to `os.environ['NUCYPHER_KEYRING_PASSPHRASE']` to make it more clear. I found it a bit confusing until I scrolled down a bit more.

nucypher/config/storages.py
```diff
@@ -0,0 +1,307 @@
...
+    @abstractmethod
+    def from_payload(self, data: str, *args, **kwargs) -> 'NodeStorage':
+        raise NotImplementedError
+
+    @abstractmethod
+    def initialize(self):
+        raise NotImplementedError
+
+
+class InMemoryNodeStorage(NodeStorage):
```
tuxxy: Hmm, these are neat.

nucypher/config/storages.py
```diff
@@ -0,0 +1,307 @@
...
+    def get(self, checksum_address: str, federated_only: bool):
+        metadata_path = self.__generate_filepath(checksum_address=checksum_address)
+        node = self.__read(filepath=metadata_path, federated_only=federated_only)   # TODO: 466
+        return node
+
+    def save(self, node):
+        try:
+            filepath = self.__generate_filepath(checksum_address=node.checksum_public_address)
+        except AttributeError:
+            raise AttributeError("{} does not have a rest_interface attached".format(self))  # TODO.. eh?
```
tuxxy: Wanna clarify here?

nucypher/config/storages.py
```diff
@@ -0,0 +1,307 @@
...
+        if not self.__temp_dir is constants.NO_STORAGE_AVAILIBLE:
+            shutil.rmtree(self.__temp_dir, ignore_errors=True)
+
+    def initialize(self):
+        self.__temp_dir = tempfile.mkdtemp(prefix="nucypher-tmp-nodes-")
+        self.known_metadata_dir = self.__temp_dir
+
+
+class S3NodeStorage(NodeStorage):
+    S3_ACL = 'private'  # Canned S3 Permissions
```
tuxxy: 👍

cli/main.py
KPrasch: Yeah - Do you think its better to read the environment variable here?

nucypher/config/storages.py
KPrasch: There are times when we have a character without a checksum address associated, although, this seems out of place, and was derived from the original file-only storages.  Id like to remove it, but will need some tests to show that this condition is not possible.

cli/main.py
tuxxy: Potentially, it's fine as is, but I just thought it was a little weird to read like this. You could probably rename the value so it doesn't sound like you're setting a passphrase there.
I'll leave it up to you to decide.

KPrasch: At the moment, this environment variable is also used when creating new keyrings.  Is that good & expected behaviour, in your opinion?

nucypher/config/storages.py
KPrasch: Yes - They help speed up tests and is used when running a node in development mode as the default node storage.

nucypher/config/storages.py
KPrasch: Simple enough, as a default.

cli/main.py
tuxxy: I think it's _okay_. I just have a hard time trusting Python to do it right (or the user). I would feel more comfortable to see a param passed into the CLI app to tell it to use the env var (or, potentially, _the_ env var) rather than just use it if it is set.

PR #480, opened 2018-10-11
Keyring Derived TLSHostingPower

KPrasch:
Restore TLS certificates saved to the disk, deriving them from a `NucypherKeyring` via the `CryptoPower`a API; Removes the handling of private keys within the scope of a `Character`

### DIFF
--- a/cli/main.py
+++ b/cli/main.py
@@ -1170,7 +1170,7 @@ def ursula(config,
     click.secho("Running in {} mode".format(config.operating_mode), fg='blue')
 
     # Bootnodes, Seeds, Known Nodes
-    ursula_config.get_bootnodes()
+    ursula_config.load_bootnodes()
     quantity_known_nodes = len(ursula_config.known_nodes)
     if quantity_known_nodes > 0:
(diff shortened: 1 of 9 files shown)


cli/main.py
```diff
@@ -941,12 +943,10 @@ def __deploy_contract(deployer_class: ClassVar,
...
             if upgradeable:
-                def __collect_secret_hash():
-                    secret = click.prompt("Enter secret hash for {}".format(__contract_name), hide_input=True, confirmation_prompt=True)
-                    secret_hash = hashlib.sha256(secret)
-                    __deployer_init_args.update({'secret_hash': secret_hash})
-                    return secret
-                __collect_secret_hash()
+                secret = click.prompt("Enter deployment secret for {}".format(__contract_name),
+                                      hide_input=True, confirmation_prompt=True)
+                secret_hash = hashlib.sha256(secret)
```
michwill: What are the security requirements for the deployment secret (if at all)?

KPrasch: Yeah - nothing is enforced here, we can pipe this through a validator first.

PR #490, opened 2018-10-24
[WIP] Cedar

jMyles:


KPrasch:
There is still a bit more work to do here: #533 #502 #535 #518 #520 #517 #522 #534 at minimum.  Then several additional commits for publication.

setup.py
```diff
@@ -1,27 +1,167 @@
...
+    'pytest-cov',
+    'mypy',
+    'codecov',
+    'coverage',
+    'moto'
+]
+
+DEPLOY_REQUIRES = [
+    'bumpversion',
+    'boto3',
```
michwill: When is `boto3` used, for Ursulas or for Alices also?

Currently, if we `pip install` nucypher, `boto3` is not installed unless it's `deployment`, and anything from `storages` module is unusable (becase `boto3` is imported there)

KPrasch: Good catch.  We use `boto3` for ansible provisioning, as well as the S3 Storage backend adaptor.  I can move it over to `INSTALL_REQUIRES`.

michwill: :+1:

KPrasch:
Still waiting on #540 #531

PR #498, opened 2018-10-27
Nicknames

jMyles:
Nodes now get nicknames generated based on their checksum address in the format:

"{color}ish color {western zodiac sign} {eatern zodiac sign}"

For example:

* Aquaish DarkTurquoise Leo Dragon
* SeaGreenish CadetBlue Pisces Goat

These names are not generated in a cryptographically-secure way; they are meant for use only in monitoring dashboards or other places where human-readability is paramount and verification can occur against the actual checksum address.  The total space is only 10819200, so we can expect an honest collision every 5000 or so nodes.

In addition to the `nickname`, nodes are also given `nickname_metadata`, which contains information useful for making visual icons based on the nickname: hex values for the two colors and unicode characters for the two signs.

### DIFF
--- a/nucypher/characters/base.py
+++ b/nucypher/characters/base.py
@@ -354,8 +354,9 @@ class Learner:
                 elif not self._learning_task.running:
                     raise self.NotEnoughTeachers("The learning loop is not running.  Start it with start_learning().")
                 else:
-                    raise self.NotEnoughTeachers("After {} seconds and {} rounds, didn't find these {} nodes: {}".format(
-                        timeout, rounds_undertaken, len(still_unknown), still_unknown))
+                    raise self.NotEnoughTeachers(
+                        "After {} seconds and {} rounds, didn't find these {} nodes: {}".format(
+                            timeout, rounds_undertaken, len(still_unknown), still_unknown))
 
             else:
                 time.sleep(.1)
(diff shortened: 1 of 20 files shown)


nucypher/network/nicknames/__init__.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def nickname_from_seed(seed):
+    random.seed(seed)
+    color1 = random.choice(colors)
+    color2 = random.choice(colors)
+    western_symbol = random.choice(western_list)
+    eastern_symbol = random.choice(eastern_list)
+    nickname = "{}ish {} {} {}".format(color1['color'], color2['color'], western_symbol, eastern_symbol)
+    nickname_metadata = (color1, color2, western[western_symbol], eastern[eastern_symbol])
```
cygnusv: just my 2 cents: I find the format `color symbol color symbol` to be more appealing. Kind of a "_crouching tiger hidden dragon_" sort of thing.

nucypher/characters/base.py
```diff
@@ -620,8 +622,8 @@ def __hash__(self):
 
     def __repr__(self):
         class_name = self.__class__.__name__
-        r = "{} {}"
-        r = r.format(class_name, self.canonical_public_address)
+        r = "⇀{}↽ ({})"
```
cygnusv: Why the arrows?

PR #499, opened 2018-10-27
Implement revoke on Alice

tuxxy:


tests/characters/test_alice_can_grant_and_revoke.py
```diff
@@ -1,6 +1,7 @@
 import datetime
 import maya
 import pytest
+from apistar.test import TestClient
```
cygnusv: you're not using this.
```suggestion
```

jMyles:
...wait a minute: am I reading this correctly that anybody can hit this endpoint and cause revocation (ie, there's no check at all that this is actually Alice)?  Or am I missing something right in front of me?

tuxxy:
> ...wait a minute: am I reading this correctly that anybody can hit this endpoint and cause revocation (ie, there's no check at all that this is actually Alice)? Or am I missing something right in front of me?

Yes, that's correct. How we want to authenticate and who can revoke is a needed discussion.

michwill:
Oh, true. I think, at least Alice should be able to revoke, but maybe Bob also should be able to..

michwill:
:joy:

nucypher/characters/lawful.py
```diff
@@ -141,6 +141,23 @@ def grant(self, bob, uri, m=None, n=None, expiration=None, deposit=None, handpic
         policy.enact(network_middleware=self.network_middleware)
         return policy  # Now with TreasureMap affixed!
 
+    def revoke(self, policy):
+        """
+        Goes through the RevocationKit in Policy and revokes re-encryption
```
KPrasch: Goes through?

nucypher/characters/lawful.py
```diff
@@ -141,6 +141,23 @@ def grant(self, bob, uri, m=None, n=None, expiration=None, deposit=None, handpic
...
+        access to Bob from each arrangement.
+        """
+        # Sign the RevocationNotices in the RevocationKit in preparation for
+        # sending to Ursula.
+        policy.revocation_kit.sign(self.stamp)
+
+        failed_revocations = list()
+        for node_id, notice in policy.revocation_kit:
+            ursula = self.known_nodes[node_id]
```
KPrasch: Hmm - this may not be robust against `KeyError`, for example, if there are no known nodes.

nucypher/characters/lawful.py
```diff
@@ -141,6 +141,23 @@ def grant(self, bob, uri, m=None, n=None, expiration=None, deposit=None, handpic
...
+            response = self.network_middleware.revoke_arrangement(ursula, notice)
+            if response.status_code != 200:
```
KPrasch: Let's also handle a non-200 response code

nucypher/crypto/kits.py
```diff
@@ -60,3 +60,38 @@ def __init__(self, *args, **kwargs) -> None:
     def from_bytes(cls, some_bytes):
         capsule, sender_pubkey_sig, ciphertext = cls.split_bytes(some_bytes)
         return cls(capsule=capsule, sender_pubkey_sig=sender_pubkey_sig, ciphertext=ciphertext)
+
+
+class RevocationKit:
+
+    def __init__(self, revocation_notices):
```
KPrasch: Type hints pretty-please :cake:

nucypher/crypto/kits.py
```diff
@@ -60,3 +60,38 @@ def __init__(self, *args, **kwargs) -> None:
...
+        self.revocation_notices = revocation_notices
+        self.revocations = dict()
+
+    def __iter__(self):
```
KPrasch: Interesting, Iteration returns tuples here - is this intentional?

nucypher/crypto/kits.py
```diff
@@ -60,3 +60,38 @@ def __init__(self, *args, **kwargs) -> None:
...
+        return iter(self.revocations.items())
+
+    def __len__(self):
+        return len(self.revocation_notices)
+
+    @classmethod
+    def from_bytes(cls, some_bytes):
+        # TODO: Implement serialization
```
KPrasch: Thoughts on implementation here?  Do we need to use BSS?

nucypher/crypto/kits.py
```diff
@@ -60,3 +60,38 @@ def __init__(self, *args, **kwargs) -> None:
...
+        pass
+
+    def sign(self, signer):
+        """
+        Signs all the revocation notices provided in the kit and adds them to
+        the revocation dict ready for use.
+        """
+        for notice in self.revocation_notices:
+            notice.sign(signer)
```
KPrasch: Can this return a bool, or does it raise an exception?

nucypher/crypto/kits.py
```diff
@@ -60,3 +60,38 @@ def __init__(self, *args, **kwargs) -> None:
...
+            self.revocations[notice.node_id] = notice
+
+    def add_receipt(self, node_id, signed_receipt):
+        """
+        Adds a signed receipt of Ursula's ability to revoke the arrangement.
+        """
+        # TODO: Verify Ursula's signature
+        # TODO: Implement receipts
+        raise NotImplementedError
```
KPrasch: Yeah - This is a codebase-wide project.

nucypher/crypto/signing.py
```diff
@@ -6,7 +6,7 @@
 signature_splitter = BytestringSplitter(Signature)
 
 
-class SignatureStamp(object):
+class SignatureStamp:
```
KPrasch: :hammer:

nucypher/keystore/keystore.py
```diff
@@ -107,18 +107,18 @@ def add_policy_arrangement(self, expiration, id, kfrag=None,
 
         return new_policy_arrangement
 
-    def get_policy_arrangement(self, arrangement_id: bytes, session=None) -> PolicyArrangement:
+    def get_policy_arrangement(self, arrangement_id_as_hex: bytes, session=None) -> PolicyArrangement:
```
KPrasch: :-1: ... what?

nucypher/network/middleware.py
```diff
@@ -102,6 +102,17 @@ def reencrypt(self, work_order):
             cfrags_and_signatures)
         return cfrags
 
+    def revoke_arrangement(self, ursula, revocation_notice):
+        # TODO: Implement revocation receipts
+        response = requests.delete("https://{}/kFrag/{}".format(ursula.rest_interface,
```
KPrasch: :+1: Nice usage of `DELETE` here

nucypher/network/middleware.py
```diff
@@ -102,6 +102,17 @@ def reencrypt(self, work_order):
...
+                                                              revocation_notice.arrangement_id.hex()),
+                                 bytes(revocation_notice), verify=ursula.certificate_filepath)
+        if not response.status_code == 200:
+            if response.status_code == 404:
+                raise RuntimeError("404 - KFrag doesn't exist to revoke with id {}".format(notice.arrangement_id))
```
KPrasch: So - this is essentially an endpoint that will reveal weather or not an Ursula has a particular k_frag?

nucypher/network/server.py
```diff
@@ -222,13 +225,40 @@ def set_policy(self, id_as_hex, request: Request):
...
+            with ThreadedSession(self.db_engine) as session:
+                # Verify the Notice was signed by Alice
+                policy_arrangement = self.datastore.get_policy_arrangement(
+                    revocation_notice.arrangement_id.hex().encode(),
+                    session=session)
+                alice_pubkey = UmbralPublicKey.from_bytes(
+                    policy_arrangement.alice_pubkey_sig.key_data)
+                is_valid = revocation_notice.verify(alice_pubkey)
+
+                if is_valid:
```
KPrasch: Do you need to handle the `else` clause here?

nucypher/network/server.py
```diff
@@ -222,13 +225,40 @@ def set_policy(self, id_as_hex, request: Request):
...
+        return 200
+
     def reencrypt_via_rest(self, id_as_hex, request: Request):
         from nucypher.policy.models import WorkOrder  # Avoid circular import
         arrangement_id = binascii.unhexlify(id_as_hex)
         work_order = WorkOrder.from_rest_payload(arrangement_id, request.body)
         self.log.info("Work Order from {}, signed {}".format(work_order.bob, work_order.receipt_signature))
         with ThreadedSession(self.db_engine) as session:
-            policy_arrangement = self.datastore.get_policy_arrangement(arrangement_id=id_as_hex.encode(),
+            policy_arrangement = self.datastore.get_policy_arrangement(id_as_hex.encode(),
```
KPrasch: :-1:  - Once again confused about types here, and we lost some semantic value, too

nucypher/policy/models.py
```diff
@@ -468,7 +467,55 @@ def __len__(self):
         return len(self.destinations)
 
 
-class WorkOrder(object):
+class RevocationNotice:
+    splitter = BytestringSplitter((bytes, 7),
+                                  (bytes, constants.NOTICE_LENGTH(32)),
```
KPrasch: Let's use a non-sorrowful constant here.

nucypher/policy/models.py
```diff
@@ -468,7 +467,55 @@ def __len__(self):
...
+        self.notice = b"REVOKE-" + arrangement_id
+        self.signature = signature or constants.NOT_SIGNED
+
+    def __bytes__(self):
+        if not self.signature is constants.NOT_SIGNED:
+            return self.notice + bytes(self.signature)
+        return constants.NOT_SIGNED
+
+    def __len__(self):
+        return len(b'REVOKE-') + len(self.arrangement_id)
```
KPrasch: Hmmm - not quite understanding what this means

nucypher/policy/models.py
```diff
@@ -468,7 +467,55 @@ def __len__(self):
...
+
+    @classmethod
+    def from_bytes(cls, some_bytes):
+        # TODO: Implement BytestringSplitter
+        _, arrangement_id, signature = cls.splitter(some_bytes)
+        return cls(None, arrangement_id, signature=signature)
```
KPrasch: `None`?

nucypher/policy/models.py
```diff
@@ -468,7 +467,55 @@ def __len__(self):
...
+
+    def sign(self, signer):
+        """
+        Signs the notice so Ursula can validate its authenticity.
+        """
+        self.signature = signer(self.notice)
```
KPrasch: Let's cache a boolean attribute indicating the current singed status.

nucypher/utilities/sandbox/middleware.py
```diff
@@ -91,6 +91,18 @@ def put_treasure_map_on_node(self, node, map_id, map_payload):
                                     data=map_payload, verify=certificate_filepath)
         return response
 
+    def revoke_arrangement(self, ursula, revocation_notice):
+        mock_client = self._get_mock_client_by_ursula(ursula)
+        response = mock_client.delete('http://localhost/kFrag/{}'.format(
```
KPrasch: Hmm case-sensitivity in the URL here?

nucypher/characters/lawful.py
michwill: Yeah right. If node isn't found, it's also a failed revocation

nucypher/network/middleware.py
michwill: There will also be a 404 if someone doesn't have a right to revoke though. And Alice certainly has a right to know if Ursula has this k_frag

nucypher/network/server.py
```diff
@@ -222,13 +225,40 @@ def set_policy(self, id_as_hex, request: Request):
...
+                    session=session)
+                alice_pubkey = UmbralPublicKey.from_bytes(
+                    policy_arrangement.alice_pubkey_sig.key_data)
+                is_valid = revocation_notice.verify(alice_pubkey)
+
+                if is_valid:
+                    self.datastore.del_policy_arrangement(
+                        revocation_notice.arrangement_id.hex().encode(),
+                        session=session)
+        except:
```
michwill: I think, we should specify the exceptions so that we don't eat up any unexpected ones

nucypher/network/server.py
```diff
@@ -222,13 +225,40 @@ def set_policy(self, id_as_hex, request: Request):
...
+        TODO: Have Ursula return a signed receipt.
+        """
+        from nucypher.policy.models import RevocationNotice
+
+        revocation_notice = RevocationNotice.from_bytes(request.body)
+
+        try:
+            with ThreadedSession(self.db_engine) as session:
+                # Verify the Notice was signed by Alice
+                policy_arrangement = self.datastore.get_policy_arrangement(
```
michwill: I think, better to verify before getting this from the DB.
Checking signature is slow, getting the arrangement from the DB is faster, so an attacker (who is not Alice) can figure out, who has a particular kfrag and who doesn't

michwill: Ahh, we must get Alice's public key from the policy arrangement...
A different way would be to get it from the signature and compare.

Anyway, my point was about constant-timeness of this :-)

nucypher/policy/models.py
michwill: Yeah, it's never a good idea to do this. If this is really desired, I'd do `cls(node_id=None, ...)`... But should the `node_id` be there at all then?

nucypher/characters/lawful.py
```diff
@@ -141,6 +141,23 @@ def grant(self, bob, uri, m=None, n=None, expiration=None, deposit=None, handpic
...
+                failed_revocations.append(notice)
+
+        return failed_revocations
```
michwill: Still an open question... What does Alice do with failed revocations (if there are some)?
Of course, an immediate option is to do nothing. Another option is to use them later. Or, she can hand them to someone to submit on her behalf later...

Anyway, that's not an immediate thing to solve in this PR

nucypher/network/server.py
jMyles: I mean, sure, we can check an arbitrary signature next to it to give it some semblance of time parity.

But I think (and I don't want to speak for @tuxxy, but I think he somewhat agrees here based on our conversation last night) that we might be wandering into something like security theatre here.

What's the attack here?  A timing attack across a huge swath of the network to determine whether Ursula has an `Arrangement` matching a particular ID?

I'm not sure this is even particularly sensitive information (though I don't advocate shouting it from the mountaintops).

If an attacker has stolen a `TreasureMap` (and I'm not sure where else she'd glean the arrangement ID), then she already knows which Ursulas are in the `Policy`.

But yes, fine, let's check a blank sig in the negative case and call it a night.

nucypher/network/server.py
jMyles: ....no?

I'm decreasingly in favor of this "true or raise" pattern, so I kinda wish we did have to.

nucypher/network/server.py
jMyles: :+1:

nucypher/network/server.py
jMyles: It doesn't bother me that much, especially since we're `binascci`'ing above.

nucypher/policy/models.py
jMyles: :+1:

nucypher/policy/models.py
```diff
@@ -468,7 +467,55 @@ def __len__(self):
...
 
-class WorkOrder(object):
+class RevocationNotice:
+    splitter = BytestringSplitter((bytes, 7),
+                                  (bytes, constants.NOTICE_LENGTH(32)),
+                                  Signature)
+    def __init__(self, node_id, arrangement_id, signature=None):
+        self.node_id = node_id
+        self.arrangement_id = arrangement_id
+        self.notice = b"REVOKE-" + arrangement_id
```
jMyles: This pattern really gets my :goat:.  If the whole class is the `RevocationNotice`, then what the heck is this?

Which thing is *the* notice?

nucypher/policy/models.py
jMyles: https://getyarn.io/yarn-clip/ba600782-d4a9-46be-8432-78f34cb9fc0c#SkHjKzkq3Q.copy

nucypher/policy/models.py
jMyles: Do we need that?  Doesn't
```
if rfrag.signature:
    print("yep, it's signed.")
```
...accomplish that?

nucypher/network/server.py
michwill: actually, _right now_ it's not sensitive at all because we do shout from the mountaintops when writing onto Ethereum blockchain...

nucypher/network/server.py
michwill: Why do we have it as hex (rather than the pure bytestring) here btw?

nucypher/network/server.py
tuxxy: Rather than optimize for this (it's kinda like security theater), we should be using exponential backoff on the API endpoints anyway.

nucypher/characters/lawful.py
tuxxy: The idea is that you serialize failed revocations and read them at startup to attempt revocation the next time you run the CLI.

nucypher/crypto/kits.py
tuxxy: It's not verification, it's just signing.

PR #507, opened 2018-11-01
VODKA: Slashing for incorrect re-encryption

szotov:
The merge commits of the original PRs that constitute vodka was lost during some rebase.
To facilitate reviewing, this is the order of merged PRs:

- #591: Slashing in the MinersEscrow
- #509: On-chain ZKP verification for re-encryption correctness 
- #606: Improvement of the MiningAdjudicator contract
- #653: Check KFrag signature as part of on-chain ZKP verification
- #770: Updates version of solidity in vodka


cygnusv:
Rebased `vodka` on top of current `master` (17c2a5749dcd46fa5bf6324d436d8f3675a20d05). 

Proposed ramp order to merge:
- [ ] #592 Adapt NuCypher to umbral==0.1.2a0
- [ ] #591 Slashing in the MinersEscrow
- [ ] #509 On-chain ZKP verification (WIP)

actually, #591 and #592 are independent from each other so the merge order doesn't matter.

tuxxy:
Re-running CI tests to check against latest master.

cygnusv:
@michwill yes, that's a good idea. I re-opened issue #749 to cover this.

Pipfile
```diff
@@ -21,8 +21,9 @@ sqlalchemy = "*"
 maya = "*"
 flask = "*"
 # Third-Party Ethereum
-py-evm = "*"
+py-evm = "==0.2.0a39"
```
KPrasch: :+1: This a very recent version.

dev-requirements.txt
```diff
@@ -1,58 +1,109 @@
 -i https://pypi.python.org/simple
+-e .
```
KPrasch: I think this line is not needed.

dev-requirements.txt
```diff
@@ -1,58 +1,109 @@
...
+eth-account==0.3.0
+eth-bloom==1.0.3
+eth-hash[pycryptodome,pysha3]==0.2.0
+eth-keyfile==0.5.1
+eth-keys==0.2.1
+eth-rlp==0.1.2
+eth-tester==0.1.0b37
+eth-typing==2.0.0
+eth-utils==1.4.1
+ethpm==0.1.4a12
```
KPrasch: This package depends on `ethereum/py-solc` from PYPI; which conflicts with `nucypher/py-solc` (which I also do not see here in `dev-requirementst.txt` - see `master` for an example) - so merging this PR into master may result in build failure.

nucypher/blockchain/eth/constants.py
```diff
@@ -17,7 +17,7 @@
 """Nucypher Token and Miner constants."""
 
 ONE_YEAR_IN_SECONDS = 31540000
-NUCYPHER_GAS_LIMIT = 6000000  # TODO: move elsewhere?
+NUCYPHER_GAS_LIMIT = 6500000  # TODO: move elsewhere?
```
KPrasch: Still growing!  @szotov - In your opinion are we still within "safe" limits here?  Do you think we need to take caution when deploying to transact across multiple blocks instead of all at once?

Pipfile
szotov: I had some problems with `py-evm==0.2.0a41`

nucypher/blockchain/eth/constants.py
szotov: Right now we are in safe limit, because real gas usage around 4 millions of gas, but buffer correction and gas adjustment in tests in py-evm leads to growing this limit.

dev-requirements.txt
szotov: It was prepared using `pipenv`
How properly to fix this?
Remove `ethpm`?
Or take file from `master`?

nucypher/blockchain/eth/constants.py
michwill: Does it change when there is a preallocation list of addresses?

dev-requirements.txt
michwill: Perhaps, for now, we may fork py-solc (ugh!) and put it into a different namespace.
If we're uncomfortable with putting it in pypi, we can put that right into nucypher repository (I know, that is not nice)

michwill: @KPrasch and I actually had a conversation about it in Denver. It's a bit trickier if we need both.
A little bit easier if we need only `nucypher/py-solc`

dev-requirements.txt
cygnusv: Isn't this a direct consequence of having `nucypher = {editable = true,path = "."}` listed as a dev dependency? The requirements files were generated automatically using the `pipenv lock --requirements` command.

dev-requirements.txt
KPrasch: I have needed to manually change requirements files generated by pipenv.  I can successfully run the project by removing `eth-pm`, then installing `nucypher/py-solc` from GH.  installing `web3` will result in the installation of `eth-pm` (and thus `ethereum/py-solc`) so it must be done in this order.

KPrasch: Perhaps, just take the requirements files from `master` and update them manually.

Pipfile
KPrasch: Not a blocker, but what Issues did you get?

dev-requirements.txt
KPrasch: Indeed, this is a result of having the editable nucypher package listed in the `Pipfile` - however, we do not want *all* pip installations to be editable (and they do not need to be).  In short `Pipenv` is helpful, but not 100% accurate for our needs.  I suggest always keeping a close eye on machine-generated requirements files, as they almost are never exactly right.

nucypher/blockchain/eth/constants.py
KPrasch: > Do you think we need to take caution when deploying to transact across multiple blocks instead of all at once?

Perhaps this is a good Idea to build into deployment?

requirements.txt
```diff
@@ -30,6 +31,7 @@ eth-rlp==0.1.2
 eth-tester==0.1.0b37
 eth-typing==2.0.0
 eth-utils==1.4.1
+ethpm==0.1.4a12
```
KPrasch: This package is OK here - it does not interfere. (only dev installation is affected by `py-solc` issues.)

nucypher/blockchain/eth/constants.py
szotov: Sorry, missed question before
I think we already have this in some way - wait for transaction after deployment plus local gas limit if I understand correctly

Pipfile
szotov: ...
```
eth_tester/backends/pyevm/main.py", line 63, in <module>
    from eth.vm.spoof import (
ModuleNotFoundError: No module named 'eth.vm.spoof'
```

nucypher/blockchain/eth/sol/source/contracts/MiningAdjudicator.sol
```diff
@@ -0,0 +1,485 @@
...
+        bytes memory _preComputedData
+    )
+        public
+    {
+        require(_minerPublicKey.length == 65 && _requesterPublicKey.length == 65);
+
+        // Check that CFrag is not evaluated yet
+        bytes32 evaluationHash = SignatureVerifier.hash(
+            abi.encodePacked(_capsuleBytes, _cFragBytes), hashAlgorithm);
+        require(!evaluatedCFrags[evaluationHash]);
```
jMyles: Forgive my ignorance here.  So, the `evaluatedCFrags` mapping will change any time a block includes a completed transaction against this contract, right?

So what happens if the same CFrags is challenged many times in the same block time, before the adjudication is finalized in a block?  Will some of the adjudications end up in Uncles or thrown away entirely?

Sub-question:  If an Uncle contains a proper adjudication but the block itself didn't include it, is there a way to use the adjudication result on the Uncle?

tests/metrics/estimate_gas.py
```diff
@@ -537,10 +631,96 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
...
+    testerchain.wait_for_receipt(tx)
+
+    sug_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
+    slashing_args = generate_args_for_slashing(testerchain, ursula1)
+    log.info("Second slashing one sub stake and saving old one (" + sug_stakes_length + " sub stakes) = " +
+             str(adjudicator_functions.evaluateCFrag(*slashing_args).estimateGas({'from': alice1})))
+    tx = adjudicator_functions.evaluateCFrag(*slashing_args).transact({'from': alice1})
+    testerchain.wait_for_receipt(tx)
+
+    sug_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
```
mswilkison:
```suggestion
    sub_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
```

tests/metrics/estimate_gas.py
```diff
@@ -537,10 +631,96 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
...
+    slashing_args = generate_args_for_slashing(testerchain, ursula1)
+    log.info("Third slashing one sub stake and saving old one (" + sug_stakes_length + " sub stakes) = " +
```
mswilkison:
```suggestion
    log.info("Third slashing one sub stake and saving old one (" + sub_stakes_length + " sub stakes) = " +
```

tests/metrics/estimate_gas.py
```diff
@@ -537,10 +631,96 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
...
+             str(adjudicator_functions.evaluateCFrag(*slashing_args).estimateGas({'from': alice1})))
+    tx = adjudicator_functions.evaluateCFrag(*slashing_args).transact({'from': alice1})
+    testerchain.wait_for_receipt(tx)
+
+    sug_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
```
mswilkison:
```suggestion
    sub_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
```

tests/metrics/estimate_gas.py
```diff
@@ -537,10 +631,96 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
...
+    slashing_args = generate_args_for_slashing(testerchain, ursula1)
+    log.info("Slashing two sub stakes and saving old one (" + sug_stakes_length + " sub stakes) = " +
```
mswilkison:
```suggestion
    log.info("Slashing two sub stakes and saving old one (" + sub_stakes_length + " sub stakes) = " +
```

tests/metrics/estimate_gas.py
```diff
@@ -537,10 +631,96 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
...
+        testerchain.time_travel(periods=1)
+
+    tx = miner_functions.lock(MIN_ALLOWED_LOCKED, MIN_LOCKED_PERIODS).transact({'from': ursula1})
+    testerchain.wait_for_receipt(tx)
+    deposit = miner_functions.minerInfo(ursula1).call()[0]
+    unlocked = deposit - miner_functions.getLockedTokens(ursula1, 1).call()
+    tx = miner_functions.withdraw(unlocked).transact({'from': ursula1})
+    testerchain.wait_for_receipt(tx)
+
+    sug_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
```
mswilkison:
```suggestion
    sub_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
```

tests/metrics/estimate_gas.py
```diff
@@ -537,10 +631,96 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
...
+    slashing_args = generate_args_for_slashing(testerchain, ursula1)
+    log.info("Slashing two sub stakes, shortest and new one (" + sug_stakes_length + " sub stakes) = " +
```
mswilkison:
```suggestion
    log.info("Slashing two sub stakes, shortest and new one (" + sub_stakes_length + " sub stakes) = " +
```

tests/metrics/estimate_gas.py
```diff
@@ -537,10 +631,96 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
...
+             str(adjudicator_functions.evaluateCFrag(*slashing_args).estimateGas({'from': alice1})))
+    tx = adjudicator_functions.evaluateCFrag(*slashing_args).transact({'from': alice1})
+    testerchain.wait_for_receipt(tx)
+
+    sug_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
```
mswilkison:
```suggestion
    sub_stakes_length = str(miner_functions.getSubStakesLength(ursula1).call())
```

tests/metrics/estimate_gas.py
```diff
@@ -537,10 +631,96 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
...
+    slashing_args = generate_args_for_slashing(testerchain, ursula1)
+    log.info("Slashing three sub stakes, two shortest and new one (" + sug_stakes_length + " sub stakes) = " +
```
mswilkison:
```suggestion
    log.info("Slashing three sub stakes, two shortest and new one (" + sub_stakes_length + " sub stakes) = " +
```

nucypher/blockchain/eth/sol/source/contracts/lib/SignatureVerifier.sol
```diff
@@ -0,0 +1,89 @@
+pragma solidity ^0.5.3;
```
jMyles: It looks like we're ultimately passing the important stuff into `ecrecover`.  It's not immediately clear to me why this wrapper is necessary; maybe I'm just getting review fatigue.  :-)

michwill: Reads more natural to me (e.g. `verify` returns a boolean instead of an Ethereum address which logically makes no sense). Plus, hashing and parsing signatures is embedded.
IMO would be less readable when not wrapped and when r/s/v are parsed in Python

michwill: But I agree that it's a really, really dense read...

cygnusv: As @michwill points out, `ecrecover` expects the individual `r`, `s` and `v` components the signature, the message already prehashed, and outputs an ETH address corresponding to the public key. The `SignatureVerifier` wrapper is makes things easier by allowing to pass the original message, the signature as `bytes` (which is how we'll have it most of the time) and the verifying public key.

nucypher/blockchain/eth/sol/source/contracts/MiningAdjudicator.sol
```diff
@@ -0,0 +1,485 @@
...
+
+    event CFragEvaluated(
+        bytes32 indexed evaluationHash,
+        address indexed miner,
+        address indexed investigator,
+        bool correctness
+    );
+
+    uint8 public constant UMBRAL_PARAMETER_U_SIGN = 0x02;
+    uint256 public constant UMBRAL_PARAMETER_U_XCOORD = 0x03c98795773ff1c241fc0b1cced85e80f8366581dda5c9452175ebd41385fa1f;
```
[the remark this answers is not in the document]
cygnusv: This constant refer to [this point in `UmbralParameters`](https://github.com/nucypher/pyUmbral/blob/17c1990c1e2e4ccaab41b4d1dc8dc16b489d6327/umbral/params.py#L33), hence the name. If that's not self-explanatory I guess we can add a comment.

nucypher/blockchain/eth/sol/source/contracts/MiningAdjudicator.sol
szotov: First transaction in block will be completed and others will fail because of `require(!evaluatedCFrags[evaluationHash]);` (state changes after each transaction in a block). So others can't make any changes on state. The same as we put them in different blocks.

I dont know if miner could use precalculated result from uncle, but for client it's useless. Because uncle does not change anything for client, and result from this transaction just could show is CFrag correct. But client can check it locally without transaction.

nucypher/blockchain/eth/sol/source/contracts/lib/SignatureVerifier.sol
jMyles: ...but then aren't we paying gas for this small operations when we can just as easily do them in python?

nucypher/blockchain/eth/sol/source/contracts/MiningAdjudicator.sol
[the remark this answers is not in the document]
jMyles: Yeah, that's exactly what I thought it was.

Nevertheless, I think a comment is a good idea; this hits the eyes quite strangely at first blush.

nucypher/blockchain/eth/sol/source/contracts/lib/SignatureVerifier.sol
szotov: we must have signatures verification in contract and of course better to just get `r`, `s`, `v`. But if we try to put it in method `evaluateCFrag` - then will be not enough variables in stack

jMyles: Ahh, so this is more for scope hygiene than readability?

szotov: at first place yes
but also could be only 16 local variables at once, so we can't put everything in separate variables

cygnusv: Solidity (or more correctly, the EVM) sucks on this regard. The "Stack too deep" forces us to do weird and ugly things to fit everything in the stack space.

tests/blockchain/eth/contracts/integration/test_intercontract_integration.py
```diff
@@ -118,12 +241,13 @@ def user_escrow_proxy(testerchain, token, escrow, policy_manager):
...
 
 @pytest.mark.slow
-def test_all(testerchain, token, escrow, policy_manager, user_escrow_proxy):
+def test_all(testerchain, token, escrow, policy_manager, adjudicator, user_escrow_proxy):
     # Travel to the start of the next period to prevent problems with unexpected overflow first period
     testerchain.time_travel(hours=1)
 
     escrow, escrow_dispatcher = escrow
     policy_manager, policy_manager_dispatcher = policy_manager
+    adjudicator, adjudicator_dispatcher = adjudicator
```
jMyles: Lines like this drive me crazy.  What was `adjudicator` if the first member is `adjudicator`?

szotov: Agreed, not the best naming
first `adjudicator`, `escrow`, `policy_manager` are tuples: (contract_abi, dispatcher_abi)

nucypher/blockchain/eth/sol/source/contracts/MiningAdjudicator.sol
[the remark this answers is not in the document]
cygnusv: ✔️

nucypher/blockchain/eth/sol/source/contracts/MiningAdjudicator.sol
michwill: Yes, the client can check locally. There could be several competing transactions in the block though. If there are, the first will collect the reward, and others will pat small fee for checking this condition before fail.
That's not ideal, but probably ok for now

KPrasch:
I have similar thoughts to @michwill  and @jMyles - and not much more to add.

Let's continue to question our approach to building upgradable contracts as we move into the next phases of deployment and testing.

PR #512, opened 2018-11-04
Versioning, Packaging, Logging

KPrasch:
Prepares nucypher for alpha release and publication to PyPI

* `setup.py` script
* Package metadata, Trove Classification, Authorship, etc.
* Single source versioning
* bumpversion configuration
* Sentry Log collection
* Demo Polishing


nucypher/__about__.py
```diff
@@ -0,0 +1,21 @@
...
+
+__summary__ = 'A proxy re-encryption network to empower privacy in decentralized systems.',
+
+__version__ = "0.1.0-alpha.0"
+
+__author__ = "NuCypher"
+
+__email__ = "dev@nucypher.com"
+
+__license__ = "GNU General Public License, Version 3"
```
tuxxy: Looks good.

.bumpversion.cfg
```diff
@@ -0,0 +1,23 @@
...
+	alpha
+	beta
+	rc
+	stable
+
+[bumpversion:part:devnum]
+
+[bumpversion:file:README.md]
+
+[bumpversion:file:umbral/__about__.py]
```
cygnusv:
```suggestion
[bumpversion:file:nucypher/__about__.py]
```

setup.py
```diff
@@ -1,19 +1,134 @@
...
+    'bytestringSplitter',
+    'hendrix>=3.1.0',
+
+    # Third Party (General)
+    'cryptography>=2.3',
+    'pysha3',
+    'requests',
+    'sqlalchemy',
+    'apistar==0.5.42',
+    'tzlocal==2.0.0b1',
```
michwill: This doesn't work for me. As soon as `tzlocal==2.0.0b1` got into `setup.py`, `nucypher` command fails for me. It shouldn't be pinned to this version neither in `setup.py`, nor in `Pipfile`

tuxxy: @KPrasch and I discussed this, taking care of this now is going to needlessly break some dev environments. We're going to push this for a later PR.

cli/main.py
```diff
@@ -1,22 +1,22 @@
 #!/usr/bin/env python3
```
michwill: Wait... I guess, cli should be a module (so it should have `__init__.py`). Also, it should probably live in nucypher namespace.

Currently, I am installing in virtualenv (so no Pipenv to help us, like in real world), and:
```
  File "/home/michwill/Projects/nucypher-kms/_test/bin/nucypher", line 7, in <module>
    from cli.main import cli
ModuleNotFoundError: No module named 'cli'
```

setup.py
michwill: Sure. I guess, we should have an issue for this, or add a comment in an open PR where this is to be done, in order to not forget.
Although, it wouldn't even install for me without me fixing it every time, so I won't forget :joy:

cli/main.py
jMyles: Yeah, I think so too.  But not today.

PR #514, opened 2018-11-04
License NuCypher as GPLv3

tuxxy:
This software is free as in _freedom_.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -20,4 +20,4 @@ values =
 
 [bumpversion:file:README.md]
 
-[bumpversion:file:umbral/__about__.py]
+[bumpversion:file:nucypher/__about__.py]
--- a/tests/characters/test_bob_handles_frags.py
+++ b/tests/characters/test_bob_handles_frags.py
@@ -16,11 +16,11 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
 
-from tempfile import TemporaryDirectory
-
 import pytest
 import pytest_twisted
+from tempfile import TemporaryDirectory
 from twisted.internet import threads
+
 from umbral import pre
 from umbral.fragments import KFrag, CapsuleFrag
 


cli/main.py
```diff
@@ -1,3 +1,19 @@
+"""
```
michwill: I respect the communism, but the comment should be after `#!/usr/bin/env python3`. Vim with flake8 should have told you that!

tuxxy: I didn't use vim, I used a script to prepend the license header. If you see anymore, lmk.

examples/finnegans_wake_demo/finnegans-wake-concise-demo.py
```diff
@@ -1,3 +1,19 @@
+"""
```
michwill: Is it really necessary to have this (huge!) disclaimer even in the demo script?

I know, RMS does this, but still...

jMyles: I'm absolutely in favor of removing it.

mswilkison:
Under GPL, if someone forks Ursula into Forkula, and connects to the network they are not required to open source the code. With AGPL, they would be. Is my understanding correct?

Since AGPL is GPL for code that runs primarily over a network should we be using that?

michwill:
Well, that is right. But seriously: imagine someone connecting to the network using a proprietary and perhaps malicious Ursula breaking all possible licenses. But you don't even know who that is.
How would AGPL help you in this case?

In the very worst case, we are able to hard fork and ban some Ursula if everyone agrees that this Ursula is malicious (and thus will follow the new fork).

So, I think, AGPL is a bit of an overkill here. GPL is helpful mainly dealing with largeco clients who wouldn't use the public network (so we force them to not ride for free), and I think just GPL seems enough for that.

Besides, as another data point, go-ethereum uses LGPL3, parity is GPL3

examples/finnegans_wake_demo/finnegans-wake-concise-demo.py
KPrasch: I like the formality in some cases.. but not this one.  Im in favor of removing it.

cli/main.py
KPrasch: Just to nit-pick, I think this looks better / more convention with two new lines after the license text.

PR #516, opened 2018-11-05
Don't report to Sentry if pytest is running

tuxxy:
As the title states


nucypher/__init__.py
```diff
@@ -37,23 +39,24 @@
...
-        integrations=[sentry_logging],
-        release='{}@{}'.format(PACKAGE_NAME, __version__)
-    )
+if not hasattr(sys, '_pytest_is_running'):
+    try:
+        import sentry_sdk
+        from sentry_sdk.integrations.logging import LoggingIntegration
+    except ImportError:
+        if REPORT_TO_SENTRY is True:
+            raise ImportError("Sentry")
```
michwill: I think, it's enough to just `raise` here, to reraise the original error (which is, imo, slightly more informative than just "Sentry")

michwill: Moreover, do we even want to try to import `sentry_sdk` if `REPORT_TO_SENTRY` is `False`?

KPrasch: Haha - This looks like an over-zealous cut!

jMyles: I think simply "Sentry" is an incredible and gripping error message.  It sounds like a Stephen King title.  I know I'll think twice about doing anything that might risk raising this.

KPrasch: > Moreover, do we even want to try to import sentry_sdk if REPORT_TO_SENTRY is False?

For testnet - I'm leaning into it!  Only slight differences in flow control in either case.

tuxxy: > Haha - This looks like an over-zealous cut!

I can't see evidence of this. I think the PR is failing CI due to the `pytest_configure` and `pytest_unconfigure` methods.

KPrasch: @tux sounds like you are talking about two sperate Issues.  The Error message above is too terse.

nucypher/config/constants.py
```diff
@@ -5,14 +5,22 @@
...
 
-#
-# seednodes
-#
+# Static Seednodes
 SeednodeMetadata = namedtuple('seednode', ['checksum_address', 'rest_host', 'rest_port'])
 SEEDNODES = tuple()
+
+# Sentry (Set to False to disable sending Errors and Logs to NuCypher's Sentry.)
+REPORT_TO_SENTRY = True
```
cygnusv: Not necessarily related to this PR: Say I'm working with the CLI/demo script/etc. (i.e., anything that is not running with pytest) in my local environment, and I don't want to report to Sentry. Is this the only way to turn it off?

cygnusv: Exactly. An issue here is that `nucypher` currently is the main application for Ursulas, but also  the library for all the remaining actors in NuCypher's narrative. Both will have different defaults.

KPrasch: PR #526 is a reasonable fix for this - scoping sentry reports *only* to the CLI

PR #519, opened 2018-11-05
Inline Teacher URI Option; CLI Touch-ups

KPrasch:
##### Touch Ups
* Deprecate CLI Simulate Command
* Accept Load-Balanced Teacher URI's 
* Cleanup Simple Observer Output

##### Temporary Fixes
* Skip S3 Storage Tests (Recent Version Incompatibility with `boto`, `botocore`, `moto`) #527
* Include 'cli' as a top-level module.  #502 
* Pin `tzlocal==1.5.1` (Version Conflicts unresolved via Pipenv)

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -19,7 +19,7 @@ pysha3="*"
 requests = "*"
 sqlalchemy = "*"
 apistar = "==0.5.42"
-tzlocal = "==2.0.0b1"
+tzlocal = "==1.5.1"
 maya = "*"
 #
 # Third-Party Ethereum
@@ -38,7 +38,6 @@ appdirs = "*"
 click = "*"
 colorama = "*"
 moto = "*"
-nucypher = {editable = true, path = "."}
 
 [dev-packages]
 #
@@ -54,6 +53,7 @@ coverage = "*"
 boto3 = "*"
 sentry-sdk = "==0.5.2"
 ansible = "*"
+nucypher = {editable = true, path = "."}
 
 [pipenv]
 allow_prereleases = true
--- a/Pipfile
+++ b/Pipfile
@@ -37,7 +37,7 @@ web3 = "*"
 appdirs = "*"
 click = "*"
 colorama = "*"
-moto = "*"
+boto3 = "*"
 
 [dev-packages]
 #
@@ -50,10 +50,10 @@ pytest-cov = "*"
 mypy = "*"
 codecov = "*"
 coverage = "*"
-boto3 = "*"
 sentry-sdk = "==0.5.2"
 ansible = "*"
 nucypher = {editable = true, path = "."}
+moto = "*"
 
 [pipenv]
 allow_prereleases = true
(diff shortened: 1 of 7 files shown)


Pipfile
```diff
@@ -19,7 +19,7 @@ pysha3="*"
 requests = "*"
 sqlalchemy = "*"
 apistar = "==0.5.42"
-tzlocal = "==2.0.0b1"
+tzlocal = "==1.5.1"
```
michwill: Later I'd certainly try `"<2.0.0"`. But it works for now

KPrasch: :+1: Agreed

cli/main.py
```diff
@@ -678,217 +681,6 @@ def stake(config,
...
-            
-            Simulation processes .............. {}
-            
-            """.format(ursula_processes)
-        click.echo(status_message)
-
-    else:
-        raise click.BadArgumentUsage
-
-
```
jMyles: Phew!  That's a breath of fresh air right there.  A goddam strawberry patch of redlines in here.

cli/main.py
```diff
@@ -933,14 +933,12 @@ def status(config,
 
 @cli.command()
 @click.option('--debug', is_flag=True)
-@click.option('--trusted-teacher-uri', type=click.STRING)
+@click.option('--teacher-uri', type=click.STRING)
+@click.option('--min-stake', type=click.INT, default=0)
```
jMyles: It's weird and wild and I think I like it.

cli/main.py
```diff
@@ -1018,14 +1009,31 @@ def __make_ursula():
...
-        seed_metadata = [SeednodeMetadata(address, host, port)]  # TODO: Ensure seed injection works
-    else:
-        seed_metadata = ()
+    #
+    # Seed
+    #
+    seed_metadata, teacher_nodes = list(), list()
+    if teacher_uri:
+
+        if all(('@' in teacher_uri, ':' in teacher_uri)):
```
jMyles: This seems like odd logic to me.  If they don't specify a port (but do specify a checksum address) then we shove them into `from_seed_and_stake_info`?

Pipfile
jMyles:
```suggestion
tzlocal = "<2.0.0"
```

I mean... I'll make it.  :-)

cli/main.py
michwill: Validation regex?

Pipfile
michwill: Also in `setup.py`. And, if it doesn't break stuff (@KPrasch was saying that it does)

cli/main.py
KPrasch: Yeah - this needs a bit of work, fixing #522 will start to open this logic a bit

Pipfile
KPrasch: Leaving this until later... worried about keeping a good set a dependencies pre-cedar

PR #521, opened 2018-11-05
Implement Revoke 

### LINKED ISSUE #346: Alice needs to hang on to a copy of the TreasureMap to use later for revoke

tuxxy:
Moved here from #499 

### What this does:
1. Adds a `RevocationKit` to `nucypher.crypto.kits`.
2. Adds a RESTful endpoint utilizing the `DELETE` verb to revoke arrangements with the `revoke_arrangement` method.

Fixes #346

### DIFF
--- a/nucypher/network/middleware.py
+++ b/nucypher/network/middleware.py
@@ -86,7 +86,7 @@ class RestMiddleware:
         if response.status_code == 200:
             return response
         elif response.status_code == 404:
-            raise RuntimeError("404 - KFrag doesn't exist to revoke with id {}".format(revocation.arrangement_id))
+            raise RuntimeError("KFrag doesn't exist to revoke with id {}".format(revocation.arrangement_id), response.status_code)
         else:
             self.log.debug("Bad response during revocation: {}".format(response))
             raise RuntimeError("Bad response: {}".format(response.content), response.status_code)


nucypher/crypto/kits.py
```diff
@@ -76,3 +80,78 @@ def __init__(self, *args, **kwargs) -> None:
     def from_bytes(cls, some_bytes):
         capsule, sender_pubkey_sig, ciphertext = cls.split_bytes(some_bytes)
         return cls(capsule=capsule, sender_pubkey_sig=sender_pubkey_sig, ciphertext=ciphertext)
+
+
+Revocation = namedtuple('Revocation', ['prefix', 'arrangement_id', 'signature'])
```
tuxxy: What do y'all think about this? I think it makes way more sense like this, but feels weird putting it outside of the `RevocationKit`.

I can either make those static methods class-methods, or leave this here, or something else entirely different.

nucypher/characters/lawful.py
```diff
@@ -163,6 +165,38 @@ def get_policy_pubkey_from_label(self, label: bytes) -> UmbralPublicKey:
...
+        # Sign the revocations in the RevocationKit in preparation for
+        # sending to Ursula.
+        policy.revocation_kit.sign_revocations(self.stamp)
+
+        try:
+            # Wait for a revocation threshold of nodes to be known ((n - m) + 1)
+            revocation_threshold = ((policy.n - policy.treasure_map.m) + 1)
+            self.block_until_specific_nodes_are_known(
+                    policy.revocation_kit.revokable_addresses,
+                    allow_missing=revocation_threshold)
```
tuxxy: Should this not be `n - revocation_threshold`?

KPrasch:
Looks like you are being effected by the `pytest` update,  - Im working on a fix for this

KPrasch:
needs rebase over #561

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,78 @@ def __init__(self, *args, **kwargs) -> None:
...
+    def revocation_to_bytes(revocation):
+        prefix, arrangement_id, signature = revocation
+        return b''.join((prefix, arrangement_id, bytes(signature)))
+
+    @staticmethod
+    def revocation_from_bytes(revocation_as_bytes: bytes):
+        from nucypher.crypto.signing import Signature
+        revocation_splitter = BytestringSplitter((bytes, 7),
+                                                 (bytes, 32),
+                                                 Signature)
```
cygnusv: This fails when revocations haven't been signed, as `Signature` is expected to be 64 bytes, and `constants.NOT_SIGNED` is just 4 bytes.

tests/characters/test_alice_can_grant_and_revoke.py
```diff
@@ -99,7 +99,40 @@ def test_federated_grant(federated_alice, federated_bob):
...
+    # Test RevocationKit signing and verification
+    policy.revocation_kit.sign_revocations(federated_alice.stamp)
+    for revocation in policy.revocation_kit:
+        assert RevocationKit.verify_revocation(revocation, federated_alice.stamp.as_umbral_pubkey())
+
+    # Test Revocation deserialization
+    revocation = policy.revocation_kit[node_id]
+    revocation_bytes = RevocationKit.revocation_to_bytes(revocation)
+    deserialized_revocation = RevocationKit.revocation_from_bytes(revocation_bytes)
+    assert deserialized_revocation == revocation
```
cygnusv: You can reproduce the problem with serialization by running this block before the `Test RevocationKit signing and verification`

nucypher/network/server.py
```diff
@@ -106,8 +106,8 @@ def __init__(self,
             Route('/kFrag/{id_as_hex}/reencrypt',
                   'POST',
                   self.reencrypt_via_rest),
-            Route('/kFrag/revoke',
-                  'POST',
+            Route('/kFrag/{id_as_hex}',
+                  'DELETE',
```
cygnusv: Classy!

nucypher/crypto/kits.py
tuxxy: This I know, the problem is that serializing unsigned revocations makes little to no sense. It hardly seems worth the complexity to build such a thing. You can construct it with a copy of the treasure map with no effort.

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,78 @@ def __init__(self, *args, **kwargs) -> None:
...
+        return self.revocations == other.revocations
+
+    @staticmethod
+    def revocation_to_bytes(revocation):
+        prefix, arrangement_id, signature = revocation
+        return b''.join((prefix, arrangement_id, bytes(signature)))
+
+    @staticmethod
+    def revocation_from_bytes(revocation_as_bytes: bytes):
+        from nucypher.crypto.signing import Signature
```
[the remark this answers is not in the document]
tuxxy: I dunno, I feel like `nucypher.crypto.signing.InvalidSignature` makes more sense here than the Umbral library one.

nucypher/network/middleware.py
```diff
@@ -77,6 +77,18 @@ def reencrypt(self, work_order):
...
+        # TODO: Implement revocation receipts
+        response = requests.delete("https://{}/kFrag/{}".format(ursula.rest_interface,
+                                                                revocation.arrangement_id.hex()),
+                                                                b''.join(revocation),
+                                                                verify=ursula.certificate_filepath)
+        if not response.status_code == 200:
+            if response.status_code == 404:
+                raise RuntimeError("404 - KFrag doesn't exist to revoke with id {}".format(revocation.arrangement_id))
+            raise RuntimeError("Bad response: {}".format(response.content), response.status_code)
+        return response
```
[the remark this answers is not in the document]
tuxxy: I'll defer on this to someone else. I find what I've written perfectly clear, but I guess it may not be. Another opinion is appreciated.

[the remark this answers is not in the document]
KPrasch: It's a small change but I think this a bit more readable at a glance:

```python
if response.status_code == 200:
   return response
elif response.status_code == 404:
   raise UnknownKFrag("404 ...")
else:
   log.debug(...)
   raise RuntimeError("Bad response:...")
```

nucypher/network/server.py
KPrasch: :+1:

KPrasch: Why the capital `F`?

tuxxy: Dunno, that's how it's always been. :P

nucypher/crypto/kits.py
KPrasch: Perhaps a conditional to handle the case @cygnusv described?  or am I misunderstanding something?

tuxxy: I think you're misunderstanding it a bit. Serializing unsigned revocations is _absolutely useless_. The treasure map is used to construct the revocation requests. If the revocation is unsigned, there is no point in actually serializing it at all. It's better to serialize a treasure map, then generate the revocation.

tuxxy: My point is that I would rather not even include such an interface. The proper way to do this is to use the Treasure Map to generate those revocations.

KPrasch: OK - Got it.  Nonetheless - a crash is possible here if `signature` is a `_Constant`, right?

tuxxy: Yes, how should it be handled? I'm presuming in a similar fashion to how we handle malformed input to other BSS calls?

KPrasch: Yeah - I'd regard this as a sensitive area not - from our experience with BSS - its important to have a very informative error message here.  What about preventing malformed transmissions altogether? Do you think its bad to look before you leap here (or in other similar situations) @jMyles?

nucypher/crypto/kits.py
KPrasch: Yeah - this is feeling a bit awkward and informal out here...

nucypher/network/server.py
KPrasch: Hm - OK!

tests/characters/test_alice_can_grant_and_revoke.py
KPrasch: Yeah, this needs to be handled, IMO.

nucypher/crypto/kits.py
tuxxy: It can be moved inside the class, but I think I'd have to use a `classmethod` to use the methods I want.

nucypher/characters/lawful.py
```diff
@@ -163,6 +165,38 @@ def get_policy_pubkey_from_label(self, label: bytes) -> UmbralPublicKey:
...
+        except self.NotEnoughTeachers as e:
+            raise e
+        else:
+            failed_revocations = list()
+            for node_id in policy.revocation_kit.revokable_addresses:
+                ursula = self.known_nodes[node_id]
+                revocation = policy.revocation_kit[node_id]
+                response = self.network_middleware.revoke_arrangement(ursula, revocation)
+                if response.status_code != 200:
+                    if response.status_code == 404:
```
KPrasch: Phew - nested! :bird:

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,78 @@ def __init__(self, *args, **kwargs) -> None:
...
+            self.revocations[node_id] = Revocation(revocation.prefix,
+                                                   revocation.arrangement_id,
+                                                   signature)
+
+    def add_receipt(self, node_id, signed_receipt):
+        """
+        Adds a signed receipt of Ursula's ability to revoke the arrangement.
+        """
+        # TODO: Verify Ursula's signature
+        # TODO: Implement receipts
```
KPrasch: Hmm - this concept name-shadows blockchain transaction receipts.

tuxxy: Ah, good point. I'm not sure if it will interfere, what do you think?

KPrasch: I think it needs it's own name.

tuxxy: Any ideas?

nucypher/crypto/kits.py
KPrasch: Help me understand - is there a useful function for unsigned revocations?

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,78 @@ def __init__(self, *args, **kwargs) -> None:
...
+            signature = signer(revocation.prefix + revocation.arrangement_id)
+            self.revocations[node_id] = Revocation(revocation.prefix,
+                                                   revocation.arrangement_id,
+                                                   signature)
+
+    def add_receipt(self, node_id, signed_receipt):
+        """
+        Adds a signed receipt of Ursula's ability to revoke the arrangement.
+        """
+        # TODO: Verify Ursula's signature
```
KPrasch: Any details on this `TODO`?

nucypher/crypto/kits.py
tuxxy: They exist purely as a in-between stage for revocations that need to be signed. Once they're signed, you could serialize them and distribute them to someone, if needed.

Other than that, they don't have a function. Considering that signed revocations can be used to cancel a policy, I think of them as a not-so-toxic-yet-toxic material that shouldn't really be held unless you intend to store them or distribute them.

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,78 @@ def __init__(self, *args, **kwargs) -> None:
...
+    def __len__(self):
+        return len(self.revocations)
+
+    def __eq__(self, other):
+        return self.revocations == other.revocations
+
+    @staticmethod
+    def revocation_to_bytes(revocation):
+        prefix, arrangement_id, signature = revocation
+        return b''.join((prefix, arrangement_id, bytes(signature)))
```
KPrasch: ... well then why is this not `bytes(revocation)`?

nucypher/crypto/kits.py
tuxxy: Sure, the overarching idea of a revocation receipt is that Ursula would sign the _signed revocation_ after performing the revocation. She would then return this signed receipt to Alice as output of the `DELETE` call. This is her agreement that she performed the revocation.

This `add_receipt` method gives Alice, or whoever performing the revocations, the ability to verify Ursula's signature as a means to verify that she agreed to remove the KFrag. What she does with this receipt is sorta an open question.

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,78 @@ def __init__(self, *args, **kwargs) -> None:
...
+        return True
+
+    @property
+    def revokable_addresses(self):
+        """
+        Returns a Set of revokable addresses in the checksum address formatting
+        """
+        return set(self.revocations.keys())
+
+    def sign_revocations(self, signer):
```
KPrasch: I think do this off-the-bat and require revocations to be signed all the time; Since, as we discussed above there is no real advantage to letting them exist unsigned.

nucypher/crypto/kits.py
tuxxy: Can you implement `__bytes__` on a `namedtuple`?

KPrasch: Heh - when you ask yourself this kind of question - you need a class.

tuxxy: It was a class before and I didn't really like it.

nucypher/crypto/kits.py
KPrasch: perhaps just `confirmation`

nucypher/crypto/kits.py
KPrasch: You also said that you dont like it as a namedtuple.

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,78 @@ def __init__(self, *args, **kwargs) -> None:
...
+
+Revocation = namedtuple('Revocation', ['prefix', 'arrangement_id', 'signature'])
+
+
+class RevocationKit:
+
+    def __init__(self, policy_treasure_map):
+        self.revocations = dict()
+        for node_id, arrangement_id in policy_treasure_map:
+            self.revocations[node_id] = Revocation(b'REVOKE-', arrangement_id,
```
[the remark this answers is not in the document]
tuxxy: When a policy is enacted, a revocation kit is made on the policy.

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,39 @@ def __init__(self, *args, **kwargs) -> None:
...
+        """
+        return set(self.revocations.keys())
+
+    def add_confirmation(self, node_id, signed_receipt):
+        """
+        Adds a signed confirmation of Ursula's ability to revoke the arrangement.
+        """
+        # TODO: Verify Ursula's signature
+        # TODO: Implement receipts
+        raise NotImplementedError
```
KPrasch: So - Ideas here: How about a `dict` of arrangements to ... .  Also - do we want to persist these to disk?

nucypher/network/middleware.py
```diff
@@ -77,6 +77,21 @@ def reencrypt(self, work_order):
             cfrags_and_signatures)
         return cfrags
 
+    def revoke_arrangement(self, ursula, revocation):
+        # TODO: Implement revocation confirmations
+        response = requests.delete("https://{}/kFrag/{}".format(ursula.rest_interface,
+                                                                revocation.arrangement_id.hex()),
+                                                                b''.join(revocation),
+                                                                verify=ursula.certificate_filepath)
+        if response.status_code == 200:
```
KPrasch: :+1: Looks pretty

nucypher/network/middleware.py
```diff
@@ -77,6 +77,21 @@ def reencrypt(self, work_order):
...
+            return response
+        elif response.status_code == 404:
+            raise RuntimeError("404 - KFrag doesn't exist to revoke with id {}".format(revocation.arrangement_id))
```
KPrasch: Why the `RuntimeError` though?

nucypher/crypto/kits.py
```diff
@@ -76,3 +80,39 @@ def __init__(self, *args, **kwargs) -> None:
     def from_bytes(cls, some_bytes):
         capsule, sender_pubkey_sig, ciphertext = cls.split_bytes(some_bytes)
         return cls(capsule=capsule, sender_pubkey_sig=sender_pubkey_sig, ciphertext=ciphertext)
+
+
+class RevocationKit:
+
+    def __init__(self, policy_treasure_map, signer: 'SignatureStamp'):
```
KPrasch: Perhaps put a `self.__confirmations` dict here?

nucypher/characters/lawful.py
```diff
@@ -163,6 +165,33 @@ def get_policy_pubkey_from_label(self, label: bytes) -> UmbralPublicKey:
...
+        except self.NotEnoughTeachers as e:
+            raise e
+        else:
+            failed_revocations = list()
+            for node_id in policy.revocation_kit.revokable_addresses:
+                ursula = self.known_nodes[node_id]
+                revocation = policy.revocation_kit[node_id]
+                response = self.network_middleware.revoke_arrangement(ursula, revocation)
+                if response.status_code == 404:
+                    failed_revocations.append((revocation, None))
```
KPrasch: Can this `None` instead be a constant sorrow?

nucypher/policy/models.py
```diff
@@ -596,3 +600,46 @@ def by_capsule(self, capsule):
                 if saved_capsule == capsule:
                     ursulas_by_capsules[ursula] = work_order
         return ursulas_by_capsules
+
+
+class Revocation:
```
KPrasch: Mmm - This is tasty as a `class`. (Still dont love the name, though).  

It... almost... feels to me like this class needs a "verb" method, but maybe it's a more static object.  Can it execute itself somehow (as well as through the Kit)?

tests/characters/test_alice_can_grant_and_revoke.py
```diff
@@ -96,6 +102,37 @@ def test_federated_grant(federated_alice, federated_bob):
         assert kfrag == retrieved_kfrag
 
 
+@pytest.mark.usefixtures('federated_ursulas')
+def test_revocation(federated_alice, federated_bob):
```
KPrasch: Do you want this test cover blockchain I/O or is this aimed at federated only?

tuxxy: This PR is aimed at Federated only. Definitely need to discuss more on blockchain revocation.

nucypher/network/server.py
```diff
@@ -250,6 +257,35 @@ def set_policy(self, id_as_hex, request: Request):
         # TODO: Sign the arrangement here.  #495
         return  # TODO: Return A 200, with whatever policy metadata.
 
+    def revoke_arrangement(self, id_as_hex, request: Request):
+        """
+        REST endpoint for revoking/deleting a KFrag from a node.
+        TODO: How do we want to verify that this request comes from Alice?
```
KPrasch: Thoughts on this / do we have a related issue?

nucypher/policy/models.py
```diff
@@ -596,3 +600,46 @@ def by_capsule(self, capsule):
...
+        self.arrangement_id = arrangement_id
+
+        if not (bool(signer) ^ bool(signature)):
+            raise ValueError("Either pass a signer or a signature; not both.")
+        elif signer:
+            self.signature = signer(self.prefix + self.arrangement_id)
+        elif signature:
+            self.signature = signature
+
+    def __bytes__(self):
```
KPrasch: :+1:

nucypher/policy/models.py
tuxxy: Hmm, I don't really like the idea of an executable `Revocation`.

The word "Revocation" is a noun, btw.

nucypher/characters/lawful.py
tuxxy: I'm not sure I understand where y'all think things are better as constant sorrow constants. It seems absolutely arbitrary. What is the criteria here?

nucypher/crypto/kits.py
tuxxy: I'd like to leave confirmations for another PR. Probably the same one that will deal with failed revocations.

nucypher/policy/models.py
KPrasch: Ah - it is a noun!

nucypher/network/middleware.py
tuxxy: Sorta a placeholder, I guess. Not really sure what to give it, but it made sense as a `RuntimeError` at the time.

nucypher/policy/models.py
KPrasch: So then - what does one instance cancel? ( "the official cancellation of a decree, decision, or promise.)

nucypher/network/server.py
tuxxy: Ah. That's an old TODO. :grimacing:

nucypher/crypto/kits.py
KPrasch: Sounds like good scoping. In that case I think it deserves an Issue.

nucypher/policy/models.py
tuxxy: A `Revocation` is used to revoke an `Arrangement`.

KPrasch: So to be clear 1 `revocation` can cancel 1 `arrangement`?

nucypher/network/server.py
```diff
@@ -250,6 +257,33 @@ def set_policy(self, id_as_hex, request: Request):
...
+                alice_pubkey = UmbralPublicKey.from_bytes(
+                    policy_arrangement.alice_pubkey_sig.key_data)
+
+                # Check that the request is the same for the provided revocation
+                if id_as_hex != revocation.arrangement_id.hex():
+                    return Response(status_code=400)
+                elif revocation.verify_signature(alice_pubkey):
+                    self.datastore.del_policy_arrangement(
+                        id_as_hex.encode(), session=session)
+        except (NotFound, InvalidSignature):
```
KPrasch: Is it OK that this exception targets the entire group?

nucypher/crypto/kits.py
tuxxy: See #569

nucypher/characters/lawful.py
KPrasch: I just have no idea what this `None` indicates when I look at it; You are indicating something specific here, what is it?

nucypher/crypto/kits.py
KPrasch: :+1:

KPrasch: (adding this number to the TODOs is a nice touch)

nucypher/policy/models.py
tuxxy: Yep, it includes the `arrangement_id`.

KPrasch: So then why not give it a method to do so?

nucypher/characters/lawful.py
tuxxy: Ah, I see your point. I have a better idea..

tests/characters/test_alice_can_grant_and_revoke.py
KPrasch: Needs an Issue.

nucypher/characters/lawful.py
tuxxy: Alright, see the latest commit.

nucypher/network/server.py
```diff
@@ -250,6 +257,33 @@ def set_policy(self, id_as_hex, request: Request):
...
+            return Response(content='KFrag not found or revocation signature is invalid.', status_code=404)
```
KPrasch: ... in which case this singular language? (this is effectively a semantic squash, unsure if this is intentional)

nucypher/network/middleware.py
KPrasch: I think it deserves it's own exception subclass :-)

tests/characters/test_alice_can_grant_and_revoke.py
tuxxy: See #570

nucypher/policy/models.py
tuxxy: I guess you could, but I'm not sure if that's the best place to do it. We probably need to talk about if that's something we want to support there.

KPrasch: Well - without it - the only way to interface with revoke is with the kit, right?

KPrasch:
Be sure to add related / fixed Issues, too :cowboy_hat_face:

nucypher/policy/models.py
tuxxy: As of now, this only supports Alice's ability to revoke. What you're suggesting is a departure from that and might require some further work to allow _anyone_ to revoke as long as they have the signed revocation.

KPrasch: Oh - I understood that to be the concept.  What additional work is required?

tuxxy: To answer your question, the correct way to revoke is to have Alice call `revoke` with the specific `Policy`.

KPrasch: Yeah - I see - interesting.   Well - I think it's worth the extra leg work :-)

tuxxy: Is that something we want to do _now_, though? I'm okay with doing it, it's not much extra. But it does represent a large feature addition to allow Alice have _anyone_ revoke when she simply gives away this signed revocation.

KPrasch: Need some team input here.

nucypher/network/server.py
tuxxy: Completely intentional. It prevents someone from knowing whether the policy is revoked or if the node actually has the KFrag requested if they get an old revocation notice.

nucypher/policy/models.py
```diff
@@ -596,3 +600,46 @@ def by_capsule(self, capsule):
...
         return ursulas_by_capsules
+
+
+class Revocation:
+    revocation_splitter = BytestringSplitter((bytes, 7), (bytes, 32), Signature)
+
+    def __init__(self, arrangement_id: bytes,
+                       signer: 'SignatureStamp' = None,
+                       signature: Signature = None):
+        self.prefix = b'REVOKE-'
```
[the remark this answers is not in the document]
KPrasch: > so that this specific string that gets signed is only for revoking 

Is this enforced somewhere?

> a plain signed `arrangement_id` may be of more use to a potential attacker somewhere

In what way?

tests/characters/test_alice_can_grant_and_revoke.py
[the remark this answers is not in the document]
KPrasch: Indeed.

nucypher/network/server.py
tuxxy: What exact situation are you talking about? That is why we have the possibility of `failed_revocations` which we can attempt to revoke later.

There is only ever one revocation to one arrangement. Are you saying "some" as in the whole of the revocation process, or am I missing something?

KPrasch: Ah - got it.  Yeah - so this is the wrong line for this comment.

>  Are you saying "some" as in the whole of the revocation process

Yes

tests/characters/test_alice_can_grant_and_revoke.py
```diff
@@ -96,6 +102,37 @@ def test_federated_grant(federated_alice, federated_bob):
...
+
+    # Test Revocation deserialization
+    revocation = policy.revocation_kit[node_id]
+    revocation_bytes = bytes(revocation)
+    deserialized_revocation = Revocation.from_bytes(revocation_bytes)
+    assert deserialized_revocation == revocation
+
+    # Attempt to revoke the new policy
+    failed_revocations = federated_alice.revoke(policy)
+    assert len(failed_revocations) == 0
```
KPrasch: It would be helpful to see a test showing failures occurring, IMO.

nucypher/characters/lawful.py
```diff
@@ -181,15 +181,13 @@ def revoke(self, policy):
         except self.NotEnoughTeachers as e:
             raise e
         else:
-            failed_revocations = list()
+            failed_revocations = dict()
```
jMyles: Yeah, seems better.

tests/characters/test_alice_can_grant_and_revoke.py
tuxxy: Read right below that. :)

KPrasch: Yes - I mean I situation where some revocations fail

tuxxy: ... They did fail in that test, though. :confused:

nucypher/network/middleware.py
jMyles: And status code inside the exception message?

tests/characters/test_alice_can_grant_and_revoke.py
KPrasch: I guess I am misunderstanding something.  This line reads `assert len(failed_revocations) == 0` and you're telling me they failed?

nucypher/crypto/kits.py
jMyles: ...I think "receipt" was better.  What is being "confirmed" here?  All Ursula can really do is ack receipt of the rfra... notice.

tests/characters/test_alice_can_grant_and_revoke.py
tuxxy: Read the statements right below that:

```python
# Try to revoke the already revoked policy
already_revoked = federated_alice.revoke(policy)
assert len(already_revoked) == 3
```

nucypher/policy/models.py
jMyles: Tasty or not, can we get a docstring?

tests/characters/test_alice_can_grant_and_revoke.py
KPrasch: Yes - but that tests something different.

tuxxy: No, it tests the exact case you're asking for -- a failed revocation. If there is no KFrag to revoke, it's a failed revocation. You can see what the specific fail case was by checking the status code, whether it be a 404 or non-200.

KPrasch: but this line reads `assert len(failed_revocations) == 0` .. it literally says there are 0 failed revocations. I dont understand how this constitutes the narrative I am describing.

nucypher/crypto/kits.py
KPrasch: @jMyles - Were worried about nameshadowing here.   I suppose confirmation isnt much better.

tests/characters/test_alice_can_grant_and_revoke.py
tuxxy: Then `already_revoked` could be changed to `failed_revocations` and it would satisfy you?

KPrasch: Well - I was suggesting an alternate narrative - the policy has already been revoked in the existing test.  Another possibly common pathway is that the revocations fail the first time around, no?

nucypher/characters/lawful.py
```diff
@@ -163,6 +165,31 @@ def get_policy_pubkey_from_label(self, label: bytes) -> UmbralPublicKey:
...
+            raise e
+        else:
+            failed_revocations = dict()
+            for node_id in policy.revocation_kit.revokable_addresses:
+                ursula = self.known_nodes[node_id]
+                revocation = policy.revocation_kit[node_id]
+                response = self.network_middleware.revoke_arrangement(ursula, revocation)
+                if response.status_code != 200:
+                    failed_revocations[node_id] = (revocation, response.status_code)
+        return failed_revocations
```
cygnusv: what if `len(failed_revocations) < n-m+1`? Shouldn't we raise an exception in that case (e.g., `NotEnoughRevocations`)?

nucypher/crypto/kits.py
```diff
@@ -14,6 +14,10 @@
 You should have received a copy of the GNU General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
+
+
+from bytestring_splitter import BytestringSplitter
+from collections import namedtuple
```
cygnusv: You don't need this anymore.

nucypher/network/middleware.py
cygnusv: The 404 status code should be included in the Exception, like you did below.

nucypher/network/server.py
```diff
@@ -250,6 +257,37 @@ def set_policy(self, id_as_hex, request: Request):
...
+                    return Response(status_code=400)
+                elif revocation.verify_signature(alice_pubkey):
+                    self.datastore.del_policy_arrangement(
+                        id_as_hex.encode(), session=session)
+        except (NotFound, InvalidSignature) as e:
+            self.log.debug("Exception attempting to revoke: {}".format(e))
+            return Response(content='KFrag not found or revocation signature is invalid.', status_code=404)
+        else:
+            self.log.info("KFrag successfully removed.")
+            return Response(content='KFrag deleted!', status_code=200)
```
cygnusv: Yay!

nucypher/utilities/sandbox/middleware.py
```diff
@@ -107,6 +110,17 @@ def put_treasure_map_on_node(self, node, map_id, map_payload):
...
         return response
 
+    def revoke_arrangement(self, ursula, revocation):
+        mock_client = self._get_mock_client_by_ursula(ursula)
+        response = mock_client.delete('http://localhost/kFrag/{}'.format(
+                                      revocation.arrangement_id.hex()),
+                                      data=bytes(revocation))
+        
+        if response.status_code != 200:
+            if response.status_code != 404:
```
cygnusv: What about replacing these two lines by `if response.status_code not in (200, 404):`

Another options is `if response.status_code != 200 and response.status_code != 404:`, although it's more verbose.

tests/characters/test_alice_can_grant_and_revoke.py
```diff
@@ -96,6 +102,37 @@ def test_federated_grant(federated_alice, federated_bob):
...
+
+    # Try to revoke the already revoked policy
+    already_revoked = federated_alice.revoke(policy)
+    assert len(already_revoked) == 3
+
+
```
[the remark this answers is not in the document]
cygnusv: +1 on that. As a suggestion, you can add the following to the end of `test_bob_joins_policy_and_retrieves` in `test_bob_joins_policy_and_retrieves.py`:

```python
    # Now let's try to retrieve again, but Alice revoked the policy.
    failed_revocations = federated_alice.revoke(policy)
    assert len(failed_revocations) == 0

    with pytest.raises(Ursula.NotEnoughUrsulas):
        _cleartexts = bob.retrieve(message_kit=message_kit,
                                   data_source=data_source,
                                   alice_verifying_key=alices_verifying_key)
```

Perhaps there should be a better place for integration tests, but I'd be more than OK with this.

KPrasch:
@tux - Looks like test tests are passing on CI; However, If I recall correctly, there is still an outstanding Issue?  What can we do so that the test suite fails, instead of passes until it's fixed?

tuxxy:
You're remembering wrong, those issues were fixed. Unless you have another issue in mind?

nucypher/characters/lawful.py
michwill: Yes, I think _that_ is the right approach

tests/characters/test_bob_joins_policy_and_retrieves.py
```diff
@@ -100,3 +101,12 @@ def test_bob_joins_policy_and_retrieves(federated_alice,
...
     assert plaintext == delivered_cleartexts[0]
+
+    # Let's try retrieve again, but Alice revoked the policy.
+    failed_revocations = federated_alice.revoke(policy)
+    assert len(failed_revocations) == 0
+
+    with pytest.raises(Ursula.NotEnoughUrsulas):
+        _cleartexts = bob.retrieve(message_kit=message_kit,
+                                   data_source=data_source,
+                                   alice_verifying_key=alices_verifying_key)
```
KPrasch: missing newline at EOF?

nucypher/characters/lawful.py
jMyles: Yeah, seems to me that it needs to be `n - revocation_threshold`.

nucypher/policy/models.py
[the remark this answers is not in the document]
jMyles: Yeah, I can't quickly think of a reason that it's useful.

But also: doesn't it make sense to have it be different than the `arrangement_id`?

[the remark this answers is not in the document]
tuxxy: Not really, imho. What else would you use to identify it? Could be something policy specific and not arrangement specific, but that's what I see `Revocations` being -- arrangement specific.

nucypher/characters/lawful.py
tuxxy: https://github.com/nucypher/nucypher/issues/595

PR #524, opened 2018-11-05
Simple Status page.

jMyles:
image


nucypher/network/server.py
```diff
@@ -17,10 +33,14 @@
 from nucypher.keystore.keypairs import HostingKeypair
 from nucypher.keystore.threading import ThreadedSession
 from nucypher.network.protocols import InterfaceInfo
+from jinja2 import Template
 
 
+HERE = BASE_DIR = os.path.abspath(os.path.dirname(__file__))
+TEMPLATES_DIR = os.path.join(HERE, "templates")
```
michwill: Whoops. I think, you forgot to commit the actual directory with templates?

michwill: Great!

I think, it should also be included in `package_data` in `setup.py` (otherwise it won't be available unless with `-e`). Idk if it should be a concern of this PR or some other

nucypher/network/server.py
```diff
@@ -318,6 +330,11 @@ def receive_treasure_map(self, treasure_map_id, request: Request):
             self.log.info("Bad TreasureMap ID; not storing {}".format(treasure_map_id))
             assert False
 
+    def status(self, request: Request):
+        headers = {"Content-Type": "text/html", "charset":"utf-8"}
+        content = self._status_template.render(known_nodes=self._node_tracker.values())
```
michwill: Why don't I see the name of my node there? (while `status` in console shows it)

michwill: Although, could be a problem on my side (e.g. not necessarily the port is truly open)

michwill: Upd: actually no, status doesn't show it, but the teacher seems to know me

jMyles: I'm not sure I understand the question - are you just wondering how / why the node names are shown in the template without being passed individually?

The template iterates through the nodes.

michwill: The question is: when you iterate, is your own node in the list?

KPrasch: No - This iterates over `known_nodes`

michwill: For me, looks like it's not, but I am not sure if it's a bug or feature. The learningloop tells me that other nodes seemingly see my node (judging by the number of nodes).

michwill:
Yes, great concept indeed!

nucypher/network/server.py
KPrasch: Noted - See #502

nucypher/network/server.py
KPrasch: Heh - sounds like you have an unexpected number of nodes?

michwill: Ah, I see. I think, makes sense to include the "self" node in the list also (maybe on the top). But yeah, that's a feature than

KPrasch: Yeah - I agree that it's useful to have the self-Ursula on the page somewhere conspicuous

KPrasch: https://github.com/nucypher/nucypher/issues/528

Issue #536, opened 2018-11-08
Deserialization of port number from raw bytes fails in some cases


michwill:
In Ursula's responses, port numbers are serialized as raw bytes. We split host and serialized port by `:` which sometimes causes the bug:
```
nucypher/utilities/sandbox/ursula.py:70: in make_federated_ursulas
    ursula_to_teach.remember_node(ursula_to_learn_about)
nucypher/characters/base.py:197: in remember_node
    certificate_filepath=certificate_filepath)
nucypher/network/nodes.py:158: in verify_node
    public_address, certificate_vbytes, rest_info = self._internal_splitter(response.content)
../../.local/share/virtualenvs/nucypher-kms-5SSQ9ZHE/lib/python3.6/site-packages/bytestring_splitter/__init__.py:68: in __call__
    message = message_class.from_bytes(bytes_for_this_object, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

cls = <class 'nucypher.network.protocols.InterfaceInfo'>, url_string = b'127.0.0.1:\x00\x00$:'

    @classmethod
    def from_bytes(cls, url_string):
>       host_bytes, port_bytes = url_string.split(b":")
E       ValueError: too many values to unpack (expected 2)

nucypher/network/protocols.py:36: ValueError
```
For this particular case, `url_string = b'127.0.0.1:\x00\x00$:'`.

Proposed fix - use only the fist `:` position:
```python
    def from_bytes(cls, url_string):
        pos = url_string.find(b':')
        host_bytes = url_string[:pos]
        port_bytes = url_string[pos + 1:]
        ...
```

michwill:
I see this issue every time I run _all_ the tests, including the slow ones

PR #537, opened 2018-11-08
Solves issue #536 - Deserialization of port number from raw bytes

michwill:
Full description of the problem here https://github.com/nucypher/nucypher/issues/536

### DIFF
--- a/nucypher/network/protocols.py
+++ b/nucypher/network/protocols.py
@@ -33,7 +33,9 @@ class InterfaceInfo:
 
     @classmethod
     def from_bytes(cls, url_string):
-        host_bytes, port_bytes = url_string.split(b":")
+        pos = url_string.find(b":")
+        host_bytes = url_string[:pos]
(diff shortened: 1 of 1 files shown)


nucypher/network/protocols.py
```diff
@@ -33,7 +31,9 @@ def __init__(self, host, port) -> None:
 
     @classmethod
     def from_bytes(cls, url_string):
-        host_bytes, port_bytes = url_string.split(b":")
+        pos = url_string.find(b":")
```
derekpierre: Question: would `url_string.split(":", 1)` achieve the same functionality? (asked as a python novice 😄 )

KPrasch: Yes - I think `url_string.split(b":", 1)` will also work.

michwill: Actually true, let's do that. It's more readable

Issue #529, opened 2018-11-06
Remove License Text from Demo code


KPrasch:
Based on Discussion in PR #514 - I'm in favor of removing the licence text from the top of demo code, and perhaps also elsewhere.

Issue #520, opened 2018-11-05
Catch CryptoError in CLI when user inputs wrong password


cygnusv:
Traceback:
```
Traceback (most recent call last):
  File "/home/david/.local/share/virtualenvs/nucypher-rz58eSz8/bin/nucypher", line 11, in <module>
    load_entry_point('nucypher', 'console_scripts', 'nucypher')()
  File "/home/david/.local/share/virtualenvs/nucypher-rz58eSz8/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/home/david/.local/share/virtualenvs/nucypher-rz58eSz8/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/home/david/.local/share/virtualenvs/nucypher-rz58eSz8/lib/python3.6/site-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/david/.local/share/virtualenvs/nucypher-rz58eSz8/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/david/.local/share/virtualenvs/nucypher-rz58eSz8/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/david/.local/share/virtualenvs/nucypher-rz58eSz8/lib/python3.6/site-packages/click/decorators.py", line 64, in new_func
    return ctx.invoke(f, obj, *args, **kwargs)
  File "/home/david/.local/share/virtualenvs/nucypher-rz58eSz8/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/home/david/NuCypher/dev/nucypher/cli/main.py", line 1225, in ursula
    URSULA = ursula_config.produce(passphrase=password, seed_nodes=seed_metadata, **overrides)  # 2
  File "/home/david/NuCypher/dev/nucypher/nucypher/config/characters.py", line 122, in produce
    merged_parameters = {**self.static_payload, **self.dynamic_payload, **overrides}
  File "/home/david/NuCypher/dev/nucypher/nucypher/config/characters.py", line 113, in dynamic_payload
    return {**super().dynamic_payload, **payload}
[...]

Issue #518, opened 2018-11-05
Make CLI command `nucypher ursula forget` to forget known nodes and certs


cygnusv:

KPrasch:
Perhaps `nucypher configure forget`

Issue #517, opened 2018-11-05
`nucypher status` fails since 'UrsulaConfiguration' object has no attribute 'read_known_nodes'


cygnusv:
```
File "/home/david/NuCypher/dev/nucypher/cli/main.py", line 1091, in status
    known_nodes = config.node_configuration.read_known_nodes()
AttributeError: 'UrsulaConfiguration' object has no attribute 'read_known_nodes'
```

I've fixed this locally by setting that line to `known_nodes = config.node_configuration.known_nodes`

cygnusv:
The status command needs to be updated to the latest learning API.

Sentry Issue: [NUCYPHER-ED](https://sentry.io/nucypher-41/nucypher/issues/752724099/)

PR #539, opened 2018-11-09
Preversioning; Packaging and CLI bug fixes

### LINKED ISSUE #533: Summary of unresolved packaging issues
Right now, packaging doesn't actually work: if one tries to `pip install git+https://github.com/nucypher/nucypher.git`, it will result in unworkable nucypher installation, and only "editable" installation works. Here's the summary of things to fix that:
* Install nucypher submodules. Currently `setup.py` has only `packages=[PACKAGE_NAME, 'cli']`. In fact, there are many more which currently aren't installed (unless the installation is editable, so they are already there). Usually `find_modules()` function from `setuptools` is used to get the full package list for this keyword argument.
* `tests` directory is a module, so it will appear in `find_modules()`. We don't want `tests` to be a module. If `__init__.py` is removed from `tests`, and `from .fixtures import *` is replaced with `from fixtures import *`, it works (yes, even in Pyhon3).
* `cli` is a global module. That's a very generic name which can clash with something. Actually,  `cli/main.py` can be moved to `nucypher/cli.py`. Alternatively, it can live in a namespace separate from `nucypher`, but I don't see a point of it. Entry point in `setup.py` should be changed accordingly.
* Finally, CircleCI config should be changed so that it installs `nucypher` as non-editable. This will make a more realistic test which will break down if our packaging somehow stops working.

KPrasch:
This PR makes nucypher installable via pip with the proper package metadata, data files, up-to-date dependency specifications, and CLI entry points.

* Introduces `nucypher configure forget`
* More informative exception handing
* Various bug fixes

Fixes #533 #502  #520  #522 #517 #518 #529, #481, #454  and #497


setup.py
```diff
@@ -137,31 +138,38 @@ def run(self):
...
+          'blockchain/eth/sol/source/contracts/*',
+          'blockchain/eth/sol/source/contracts/lib/*',
+          'blockchain/eth/sol/source/contracts/proxy/*',
+          'blockchain/eth/sol/source/zepellin/math/*',
+          'blockchain/eth/sol/source/zepellin/ownership/*',
+          'blockchain/eth/sol/source/zepellin/token/*']},
       include_package_data=True,
-      entry_points='[console_scripts]\n{}=cli.main:cli'.format(PACKAGE_NAME),
+
+      entry_points={'console_scripts': ['{0}={0}.cli:cli'.format(PACKAGE_NAME)]},
```
michwill: Well, again. Would that ever work with non-editable installation?

michwill: (since it's not a module)

KPrasch: To clarify - I think this is acceptable for a developer installation - I went back-and-forth for a while on where to put this file, but ultimately inside `nucypher` is the easiest place to manage it. 

Are you able to pip install the project directory? `pip install ./nucypher`

michwill: Ah wait, you moved it to nucypher/cli.py. Perfect!

michwill: Can test like:
```
virtualenv .test -p python3
source .test/bin/activate
pip3 install git+https://github.com/KPrasch/nucypher.git@preversioning
nucypher
```

PR #550, opened 2018-11-12
Touch-ups to CLI and status page

cygnusv:
- Fixes CLI errors when `nucypher stake` and `nucypher accounts` are called with no arguments.
- Adds docs to CLI commands `nucypher configure` and `nucypher stake`
- Shallow changes to the node status page (favicon, sans-serif font, removes text shadow which reads badly, etc)

### DIFF
--- a/nucypher/cli.py
+++ b/nucypher/cli.py
@@ -511,6 +511,10 @@ def accounts(config,
         keyring._export_wallet_to_node(blockchain=config.blockchain, passphrase=passphrase)
 
     elif action == 'list':
+        if config.accounts == NO_BLOCKCHAIN_CONNECTION:
+            click.echo('No account found.')
+            return
+
         for index, checksum_address in enumerate(config.accounts):
             token_balance = config.token_agent.get_balance(address=checksum_address)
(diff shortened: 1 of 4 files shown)


nucypher/cli.py
```diff
@@ -600,6 +604,10 @@ def stake(config,
 
     if not checksum_address:
 
+        if config.accounts == NO_BLOCKCHAIN_CONNECTION:
+            click.echo('No account found.')
+            return
```
KPrasch: why not `raise click.Abort()`?

nucypher/cli.py
```diff
@@ -567,26 +582,15 @@ def stake(config,
...
-
-    periods - The duration (in periods) of the stake.
-
-    Options
-    ========
-
-    --wallet-address - A valid ethereum checksum address to use instead of the default
-    --stake-index - The zero-based stake index, or stake tag for this wallet-address
+\b
+list              List all stakes for this node
```
KPrasch: Thank You! :+1:

nucypher/cli.py
cygnusv: ✅

PR #574, opened 2018-11-27
[WIP] Versioning, domains, and decentralized metadata validation

jMyles:


nucypher/characters/lawful.py
```diff
@@ -12,38 +12,38 @@
 GNU General Public License for more details.
 
 You should have received a copy of the GNU General Public License
-along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+along with nucypher.  If not, see <https://ww"Unable to glean address from node that purported to be version {}.  We're only version {}"w.gnu.org/licenses/>.
```
michwill: Wait.. what happened here?

jMyles: Bad paste; my fault.  I'll push up fixes for some of the cruft soon.

nucypher/characters/lawful.py
```diff
@@ -458,12 +453,13 @@ def __init__(self,
...
                            abort_on_learning_error=abort_on_learning_error,
                            known_nodes=known_nodes,
+                           domains=domains,
                            **character_kwargs)
 
         #
         # Self-Ursula
         #
-        if is_me is True:           # TODO: 340
+        if is_me is True:  # TODO: 340
```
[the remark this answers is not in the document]
tuxxy: "Depends on what the meaning of the word _is_ is." :joy: 

Depends on what you're going for. If you're familiar with `===` in JavaScript, the `is` is used sorta kinda like this in this case.

In this case, `if is_me is True` will only be `True`, if and only if, `is_me` is actually a boolean `True`. If it's a truthy object, it won't be True.

I'd say it's more Pythonic to have `if is_me` and call it done here, but it does make sense here if we _only_ want `is_me` to be a boolean value. Python is weird.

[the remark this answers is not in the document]
tuxxy: For example:
```python
a = "test"
bool(a) == True # True indeed
if a:
    print("This executes")
if a is True:
    print("This doesn't execute")
if bool(a) is True:
    print("This does execute")
```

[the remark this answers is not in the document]
jMyles: It depends.  They do different things.

Here, the equality variant (`if is_me:`) is correct and the code is wrong.  (At least that's how I see it).

The identity variant (`if thing is bool(...)`) is useful because you might have an object whose `__bool__` returns the proper bool type (and thus will evaluate `True` for equality), but which isn't actually the object you're testing for.

In this case, since we aren't checking for that, I assert that we're doing the wrong thing.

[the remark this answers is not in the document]
tuxxy: The second answer to this SO question I think answers this succinctly. TL;DR: `==` is value equality and `is` is reference equality. The former is for checking the value of two objects, the latter is for checking if the two _refer_ to the same object.

https://stackoverflow.com/questions/132988/is-there-a-difference-between-and-is-in-python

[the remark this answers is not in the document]
jMyles: At the end of the day: we've almost certainly got it wrong here.

nucypher/cli/painting.py
```diff
@@ -0,0 +1,184 @@
...
+def build_fleet_state_status(ursula) -> str:
+    # Build FleetState status line
+    if ursula.known_nodes.checksum is not NO_KNOWN_NODES:
+        fleet_state_checksum = ursula.known_nodes.checksum[:7]
+        fleet_state_nickname = ursula.known_nodes.nickname
+        fleet_state_icon = ursula.known_nodes.icon
+        fleet_state = '{checksum} ⇀{nickname}↽ {icon}'.format(icon=fleet_state_icon,
+                                                              nickname=fleet_state_nickname,
+                                                              checksum=fleet_state_checksum)
+    elif ursula.known_nodes.checksum is not NO_KNOWN_NODES:
```
derekpierre: Did you mean
```suggestion
    elif ursula.known_nodes.checksum is NO_KNOWN_NODES:
```

`is not` is already checked in the if above.

examples/finnegans_wake_demo/run_federated_ursula.py
```diff
@@ -32,16 +32,16 @@
 # TODO: Use real path tooling here.
 SHARED_CRUFTSPACE = "{}/examples-runtime-cruft".format(os.path.dirname(os.path.abspath(__file__)))
 CRUFTSPACE = "{}/{}".format(SHARED_CRUFTSPACE, MY_REST_PORT)
-DB_NAME = "{}/database".format(CRUFTSPACE)
+db_filepath = "{}/database".format(CRUFTSPACE)
```
EvaLovlac3: I was wondering why the demo hung here.

nucypher/characters/lawful.py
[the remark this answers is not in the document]
EvaLovlac3: ++ to `if is_me:`

EvaLovlac3:
Fix CI issues.

nucypher/characters/base.py
```diff
@@ -67,6 +68,7 @@ class Character(Learner):
     from nucypher.crypto.signing import InvalidSignature
 
     def __init__(self,
+                 domains: Set = (constants.GLOBAL_DOMAIN,),
```
KPrasch: Well - which is it, a set or a tuple? :-)

nucypher/network/__init__.py
```diff
@@ -14,3 +14,4 @@
 You should have received a copy of the GNU General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
+LEARNING_LOOP_VERSION = 1
```
KPrasch: missing newline

tests/learning/test_fault_tolerance.py
```diff
@@ -0,0 +1,191 @@
...
+                                  quantity=2,
+                                  know_each_other=True)
+    learner = lonely_ursula_maker().pop()
+    teacher, new_node = lonely_ursula_maker()
+
+    new_node.TEACHER_VERSION = learner.LEARNER_VERSION + 1
+
+    learner._current_teacher_node = teacher
+
+    warnings = []
```
KPrasch:
```suggestion
    warnings = list()
```

tests/learning/test_fault_tolerance.py
```diff
@@ -0,0 +1,191 @@
...
+
+    lonely_blockchain_learner.learn_from_teacher_node()
+
+    globalLogPublisher.removeObserver(warning_trapper)
+
+    # We received one warning during learning, and it was about this very matter.
+    assert len(warnings) == 1
+    assert warnings[0]['log_format'] == unsigned.invalid_metadata_message.format(unsigned)
+
+    assert len(lonely_blockchain_learner.known_nodes) == len(
```
KPrasch: I know we dont have a length convention established, but why break on `(`?

nucypher/characters/base.py
jMyles: `NamedEmptySet`

tests/learning/test_fault_tolerance.py
jMyles: https://giphy.com/gifs/DeoY3iC6VLBHG/html5

jMyles: https://gph.is/XIBFzF

jMyles: <iframe src="https://giphy.com/embed/DeoY3iC6VLBHG" width="480" height="213" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/come-at-me-bro-rejected-basketball-DeoY3iC6VLBHG">via GIPHY</a></p>

KPrasch: :cactus:

jMyles: https://www.google.com/url?sa=i&rct=j&q=&esrc=s&source=images&cd=&ved=2ahUKEwi79b7TwpHfAhWBTt8KHfbRB5wQjRx6BAgBEAU&url=https%3A%2F%2Ftenor.com%2Fview%2Fblock-gumby-basketball-gif-5549764&psig=AOvVaw1ksMlAsjDigOuYzbEwxR6Q&ust=1544402730783155

jMyles: tenor

jMyles: ![https://giphy.com/gifs/tPrHcw5cz9hp6/html5]

jMyles: gumby

tests/learning/test_fault_tolerance.py
jMyles: Yeeeahhh, that'd be the old CTRL-ALT-L.  I'll fix it.

nucypher/cli/painting.py
jMyles: Correct - good eye.

We'll start to need some end-to-end tests (ideally, even browser-automation if indeed we're going to do web admin type stuff).

nucypher/network/nodes.py
```diff
@@ -152,24 +156,26 @@ def snapshot(self):
         fleet_state_updated_bytes = self.updated.epoch.to_bytes(4, byteorder="big")
         return fleet_state_checksum_bytes + fleet_state_updated_bytes
 
-    def update_fleet_state(self):
-        checksum = keccak_digest(b"".join(bytes(n) for n in self.sorted())).hex()
-        if checksum != self.checksum:
+    def record_fleet_state(self, additional_nodes_to_track=None):
```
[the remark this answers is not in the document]
jMyles: huh?  :-)

jMyles:
> Fix CI issues.

@EvaLovlac3 Not sure what you mean here either.

nucypher/characters/base.py
jMyles: Actually, kidding aside, what do you think of this?  It is OK to presume that a Set - and only a Set - is passed, but that a tuple is default?

I was just going to change it to a set when I realized that it actually makes more sense for the default to be immutable.

nucypher/characters/lawful.py
```diff
@@ -16,41 +16,44 @@
...
-from typing import Iterable
-from typing import List
+from umbral.keys import UmbralPublicKey
+from umbral.signing import Signature
 
+from bytestring_splitter import BytestringKwargifier, BytestringSplittingError
 from bytestring_splitter import BytestringSplitter, VariableLengthBytestring
 from constant_sorrow import constants
-from constant_sorrow.constants import PUBLIC_ONLY
+from constant_sorrow.constants import INCLUDED_IN_BYTESTRING, constant_or_bytes
```
KPrasch: This results in a `ValueError`, I believe, with the latest version of constant sorrow:

```
E   ValueError: Use ALL_CAPS names for constants.  See https://www.python.org/dev/peps/pep-0008/#constants.
```

nucypher/characters/lawful.py
```diff
@@ -516,23 +514,24 @@ def __init__(self,
             # Ephemeral Self-Ursula
             #
             if is_me:
-                self.suspicious_activities_witnessed = {'vladimirs': list(), 'bad_treasure_maps': list()}
+                self.suspicious_activities_witnessed = {'vladimirs': [], 'bad_treasure_maps': []}
```
KPrasch: :1st_place_medal:

jMyles: That's what we call a "merge commit" with "conflict resolution" ;-)

nucypher/characters/lawful.py
jMyles: Umm yeah, it's probably defined in the wrong place.  It doesn't belong in `constants`.

But indeed, it's a utility function, and a pretty interesting one if I do say so myself.  :-)

KPrasch: In any case I cannot run the tests with the latest version of constant sorrow, due to this.

nucypher/characters/base.py
KPrasch: I'm inclined to remove this from the signature entirely and use `None`.

examples/finnegans_wake_demo/run_federated_ursula.py
KPrasch: Hmm, I dont understand what you are indicating here. :question:

jMyles: Nor I.  

We probably actually need some rudimentary tests for the demo (did we have an Issue this once upon a time?  I can't find it now.).

nucypher/network/nodes.py
```diff
@@ -252,7 +275,8 @@ def __init__(self,
         self.unresponsive_startup_nodes = list()  # TODO: Attempt to use these again later
         for node in known_nodes:
             try:
-                self.remember_node(node)  # TODO: Need to test this better - do we ever init an Ursula-Learner with Node Storage?
+                self.remember_node(
+                    node)  # TODO: Need to test this better - do we ever init an Ursula-Learner with Node Storage?
```
cygnusv: huh? why the new line?

nucypher/cli/painting.py
```diff
@@ -56,7 +56,7 @@ def build_fleet_state_status(ursula) -> str:
         fleet_state = '{checksum} ⇀{nickname}↽ {icon}'.format(icon=fleet_state_icon,
                                                               nickname=fleet_state_nickname,
                                                               checksum=fleet_state_checksum)
-    elif ursula.known_nodes.checksum is not NO_KNOWN_NODES:
+    elif ursula.known_nodes.checksum is NO_KNOWN_NODES:
```
cygnusv: I don't get this pattern I've seen multiple times in the codebase:
```python
if foo is bar:
    do_something()
elif foo is not bar:
    do_another_thing()
```
Sometimes with `is`, but sometimes with `==`; it doesn't matter. The thing is that the `elif` could be safely substituted by an `else`, since you're testing the logical opposite. Is this pattern a Python thing?

In this case, there's even an `else` clause after the `elif`, which doesn't make any sense: it's unreachable code.

derekpierre: It may be because it is possible for the value to be `None` or some equivalent - I assumed that was why there was an `elif` and an `else`. If not, then the `elif` can be replaced with just `else`

cygnusv: If `ursula.known_nodes.checksum` is `None`, then `ursula.known_nodes.checksum is foo` will only be true when `foo` is None, too. And conversely, if `ursula.known_nodes.checksum` is `None`, `ursula.known_nodes.checksum is not bar` will always be true, unless `bar` is `None`.

derekpierre: Ah! - good point - yeah there is unreachable code.

nucypher/characters/lawful.py
```diff
@@ -689,7 +690,7 @@ def from_bytes(cls,
         node_info['timestamp'] = maya.MayaDT(node_info.pop("timestamp"))
         node_info['checksum_address'] = to_checksum_address(node_info.pop("public_address"))
 
-        domains_vbytes = VariableLengthBytestring.discharge(node_info['domains'])
+        domains_vbytes = VariableLengthBytestring.dispense(node_info['domains'])
```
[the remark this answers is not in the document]
KPrasch: oh thank goodness

nucypher/characters/lawful.py
```diff
@@ -748,82 +755,104 @@ def from_seed_and_stake_info(cls,
...
+        version_splitter = BytestringSplitter((int, 2, {"byteorder": "big"}))
+        versions_and_node_bytes = [version_splitter(n, return_remainder=True) for n in nodes_vbytes]
+
+        ursulas = []
+        for version, node_bytes in versions_and_node_bytes:
+            try:
+                ursula = cls.from_bytes(node_bytes, version, federated_only=federated_only)
+            except Ursula.IsFromTheFuture as e:
+                if fail_fast:
+                    raise
```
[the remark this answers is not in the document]
jMyles: I mean... I make the presumption here (and everywhere) that we're logging exceptions and that there's no need to log the exception text additionally.

nucypher/network/nodes.py
```diff
@@ -15,48 +15,54 @@
...
+from nucypher.network import LEARNING_LOOP_VERSION
 from nucypher.network.middleware import RestMiddleware
 from nucypher.network.nicknames import nickname_from_seed
-from nucypher.network.protocols import SuspiciousActivity, parse_node_uri
+from nucypher.network.protocols import SuspiciousActivity
 from nucypher.network.server import TLSHostingPower
-from nucypher.utilities.decorators import validate_checksum_address
 
 
+GLOBAL_DOMAIN.set_constant_documentation(
```
KPrasch: :+1:

nucypher/network/nodes.py
```diff
@@ -692,8 +725,7 @@ def learn_from_teacher_node(self, eager=True):
 
 
 class Teacher:
-    TEACHER_VERSION = 1
-    _evidence_of_decentralized_identity = constants.NOT_SIGNED
+    TEACHER_VERSION = LEARNING_LOOP_VERSION
```
KPrasch: Just to be sure this is intentional; This will create two separate references, and incrementing the constant will not effect a running teacher:

```
>>> LEARNING_LOOP_VERSION = 1
>>> class Teacher:
...     TEACHER_VERSION = LEARNING_LOOP_VERSION
... 
>>> Teacher.TEACHER_VERSION
1
>>> LEARNING_LOOP_VERSION = 2
>>> Teacher.TEACHER_VERSION
1
```

tests/learning/test_fault_tolerance.py
```diff
@@ -0,0 +1,191 @@
...
+
+    globalLogPublisher.addObserver(warning_trapper)
+
+    crazy_node = b"invalid-node"
+    middleware.get_nodes_via_rest(ursula.rest_url(),
+                                  certificate_filepath=ursula.certificate_filepath,
+                                  announce_nodes=(crazy_node,))
+    assert len(warnings) == 1
+    future_node = list(federated_ursulas)[1]
+    future_node.TEACHER_VERSION = future_node.TEACHER_VERSION + 10
```
KPrasch: Do you think we need some sort of method to increment versions on a running node?

nucypher/network/nodes.py
jMyles: I don't imagine ever changing this value during runtime... Maybe I'll use constantSorrow for it?

tests/network/test_node_storage.py
```diff
@@ -39,16 +31,8 @@ def test_one_node_stores_a_bunch_of_others(federated_ursulas, ursula_federated_t
...
-            if passed.seconds > 5:
-                pytest.fail("Didn't find the seed node.")
-
-    yield deferToThread(start_lonely_learning_loop)
+    newcomer.start_learning_loop(now=True)
 
-    # The known_nodes are all saved in storage (and no others have been saved)
-    assert list(newcomer.known_nodes) == list(newcomer.node_storage.all(True))
+    assert list(newcomer.known_nodes)
+    assert len(list(newcomer.known_nodes)) == len(list(newcomer.node_storage.all(True)))
```
KPrasch: Heh - I know I wrote this but, what is `True` here? Perhaps kwargs :-)

tests/learning/test_fault_tolerance.py
jMyles: Presently, I'm unable to conceive how a node can continue running (and not restart) when changing versions.  At the very least, it will always require re-evaluation of the modules, right?

tests/network/test_node_storage.py
jMyles: `True`.

jMyles: I see how it is.  You know there's a merge commit incoming so you include some garbage, complain about it later, and pin it on me.  `True` story. 

I thought we were friends.

KPrasch: `False`

KPrasch: Really, no need to make this change right now. :-)

tests/learning/test_fault_tolerance.py
KPrasch: At the moment its not clear to me how to best minimize or automate the updating process, but this seems potentially relevant.

nucypher/network/nodes.py
```diff
@@ -597,82 +623,89 @@ def learn_from_teacher_node(self, eager=True):
...
 
-        self.cycle_teacher_node()
-
         # TODO: This doesn't make sense - a decentralized node can still learn about a federated-only node.
         from nucypher.characters.lawful import Ursula
         if response.status_code == 204:
             return constants.FLEET_STATES_MATCH
 
-        node_list = Ursula.batch_from_bytes(nodes, federated_only=self.federated_only)  # TODO: 466
+        node_list = Ursula.batch_from_bytes(node_payload, federated_only=self.federated_only)  # TODO: 466
```
KPrasch: Man, this method is getting big

nucypher/cli/painting.py
jMyles: Haha, yeah I have no idea.  This has the clear signature of @KPrasch having written some code about an hour after he started to really need some sleep.  What's happening here, man?  Why are you even making a string in the case that there are no known nodes??  Can't you just return the constant or something?

tests/learning/test_fault_tolerance.py
jMyles: But on a *running* node?  Like, literally without stopping the process?  What kind of update can we possibly perform without stopping the process?

I mean, I know that in some really crazy zero-downtime environments people do crazy shit like dynamically re-eval modules and stuff.  But I don't think we're planning to do that.  And if we are, incrementing an integer is the least of our worries.

tests/network/test_node_storage.py
jMyles: Shit man, I really left myself open to that one.  You win this round.

EvaLovlac3:
At the time this failed CircleCI, so my comment was to fix the CircleCI issues and I would approve.

PR #575, opened 2018-11-27
Dependency version freeze; Pipenv lock, legacy pip requirements.

KPrasch:


Pipfile
```diff
@@ -27,10 +27,10 @@ maya = "*"
 #
 # Third-Party Ethereum
 #
-eth-utils="==1.2.2"
+eth-utils="==1.2.1"
```
cygnusv: Why is this? AFAIK, v1.2.2 was working.

KPrasch: It causes an unresolvable sub-dependency condition with `eth-utils`

Pipfile
```diff
@@ -10,7 +10,7 @@ python_version = "3"
 #
 # NuCypher
 #
-umbral = "==0.1.1a3"
+umbral = "*"
```
cygnusv: We need the current release of Umbral, since the next one will break stuff.

```suggestion
umbral = "==0.1.1a3"
```

EvaLovlac3: Is this a version wild card?

KPrasch: Just fixed in `92dc863`

KPrasch: Good Eye

setup.py
```diff
@@ -67,39 +69,9 @@ def run(self):
...
-    'py-solc',
-    'web3',
-
-    # Third Party (Configuration + CLI)
-    'appdirs',
-    'click>=7.0',
-    'colorama',
-    'sentry-sdk==0.5.2',
-    'boto3',
+with open(os.path.join(BASE_DIR, "requirements.txt")) as f:
```
tuxxy: Is this safe?

Feels kinda odd, are any other projects doing this?

KPrasch: Yes - many.  What is the concern?

tuxxy: Just the idea that a modified requirements.txt can cause it to install the depends. Just curious.

PR #577, opened 2018-11-28
Try on some Coveralls; Faster workflow; Update pinned deps

KPrasch:
- Configures GitHub, circleCI for usage of coveralls.
- Parallelizes more jobs in the circleCI workflow (globed by stored timing results).
- Properly pins sub-dependencies 

Dashboard: https://coveralls.io/github/nucypher
Badge: [Coverage Status](https://coveralls.io/github/nucypher/nucypher?branch=pull%2F577)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -135,10 +135,19 @@ workflows:
             branches:
               ignore: /.*/
 
+
 python_36_base: &python_36_base
     working_directory: ~/nucypher-depends
     docker:
       - image: circleci/python:3.6
+
+_run:
+  coveralls: &coveralls
+    name: Upload Coverage Report
+    command: |
+      echo -e "COVERALLS_REPO_TOKEN=$COVERALLS_REPO_TOKEN" >> .env
+      pipenv run coveralls
+
 
 jobs:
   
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -129,15 +159,60 @@ jobs:
...
+    steps:
+      - checkout
+      - run:
+          name: Install Python Dependencies with Pip
+          command: |
+            pip3 install --user .
+      - run:
+          name: Install Python Development Dependencies with Pip
+          command: |
+            pip3 install --user -e .[testing]
```
michwill: When we are testing, do we even need `-e`?

KPrasch: True, we don't need it.

KPrasch: :heavy_check_mark:

requirements.txt
```diff
@@ -56,6 +56,7 @@ pyasn1-modules==0.2.2
 pyasn1==0.4.4
 pychalk==2.0.0
 pycparser==2.19
+pycryptodome==3.7.2
```
michwill: I'm actually not super-positive on pinning every single dependency in `requirements.txt`. I can see why it might be needed, but it can break a lot of stuff.

That said, we can take discussion of this outside the PR (e.g. I can see why dependency pinning is good)

KPrasch: :heavy_check_mark: (removed `pycryptodome`)

Pipfile
```diff
@@ -10,7 +10,7 @@ python_version = "3"
 #
 # NuCypher
 #
-umbral = "*"
+umbral = "==0.1.1a3"
```
cygnusv: This is already pinned in `master`. Shouldn't this be detected by GH as a merge conflict? BTW, perhaps this PR should be rebased on top of `master`.

KPrasch: Dropped and rebased. :heavy_check_mark:

jMyles:
There is an incredible (and, with this PR, growing) amount of repetition in our circle YAML configuration.  Is this par for the course for circle?  Is there no better way?

KPrasch:
I added an additional anchor for the new coveralls yaml logic, which helps.  Let me know what else, if anything, in particular can be reduced.

Issue #489, opened 2018-10-23
Docker image with dependencies


szotov:
Prepare docker file and image with installed dependecies and pushed it

EvaLovlac3:
Can I get more information on this?

KPrasch:
Yeah - perhaps poorly named:  We'd like to ship an official docker container to the docker hub on each release.

Ideally this is done alongsideautomated PyPI publications by building tagged commit-releases on CircleCI (See: <https://circleci.com/blog/how-to-build-a-docker-image-on-circleci-2-0/>).  I'd say this is a highly requested item at this time, and a prerequisite for many near-future deployment tasks. 

Currently there is an existing `Dockerfile` and `docker-compose.yml` that need some TLC.

EvaLovlac3:
Let me look into how to do this with CircleCI, I know how to just ship a docker container that has all of the deps, but I don't know a whole lot about CircleCI.

KPrasch:
Addressed in #613

PR #581, opened 2018-11-28
[WIP] Vanilla nucypher docker image

EvaLovlac3:
Fixes the default vanilla docker image, and completes an environment with required dependencies.

### DIFF
--- a/deploy/docker/Dockerfile
+++ b/deploy/docker/Dockerfile
@@ -2,9 +2,10 @@ FROM python:3.7.0-slim-stretch
 
 # Update
 RUN apt update -y && apt upgrade -y
-RUN apt install gcc libffi-dev wget -y
+RUN apt install gcc libffi-dev wget -y git
 
 # Install pipenv
+RUN apt install python3-pip -y
 RUN pip3 install pipenv
 
 # Set the working directory to /app
@@ -13,8 +14,12 @@ WORKDIR /app
 # Copy the current directory contents into the container at /app
 COPY . /app
 
+# Fetch NuCypher
+RUN git clone https://github.com/nucypher/nucypher
+RUN cd nucypher
+
 # Run pipenv
 RUN pipenv install --dev --python python3.7 --skip-lock --verbose
-RUN pipenv run ./scripts/install_solc.sh
+RUN pipenv run ./nucypher/scripts/install_solc.sh
 
 CMD ["/bin/bash"]


deploy/docker/Dockerfile
```diff
@@ -13,8 +14,12 @@ WORKDIR /app
 # Copy the current directory contents into the container at /app
 COPY . /app
 
+# Fetch NuCypher
+RUN git clone https://github.com/nucypher/nucypher
+RUN cd nucypher
+
 # Run pipenv
 RUN pipenv install --dev --python python3.7 --skip-lock --verbose
```
KPrasch: Since we just merged the version freeze - Let's make this `pipenv sync` instead

EvaLovlac3:
`sudo docker build -t nucypher-container .`

deploy/docker/Dockerfile
```diff
@@ -13,8 +14,12 @@ WORKDIR /app
 # Copy the current directory contents into the container at /app
 COPY . /app
 
+# Fetch NuCypher
+RUN git clone https://github.com/nucypher/nucypher
```
KPrasch: Since we want to control which versions are out in the wild - can we specify a tagged version to install?  I know there is not one now, but there will be shortly.

EvaLovlac3: We can certainly do that, what would you recommend?

KPrasch:
OK - That's a bit terse - If I understand the steps correctly:

1. `sudo docker build -t nucypher-container .`
2. `nucypher --help`

Is that right?

EvaLovlac3:
That should be correct, you'll also need to access the container, give it a go.

deploy/docker/Dockerfile
KPrasch: I think there is a clone flag for this:

`git clone --branch 0.1.0a0 https://github.com/nucypher/nucypher`

EvaLovlac3:
https://phase2.github.io/devtools/common-tasks/ssh-into-a-container/

deploy/docker/Dockerfile
KPrasch: Then we can hookup to `bumpversion` to alter the string ~ Cool!

KPrasch:
Im getting a bit lost - can you describe the steps in explicit detail? Also consider the node needs to be exposed to the system network to function as a node.

KPrasch:
@RomanS - Thanks for the comment:

> `RUN git clone https://github.com/nucypher/nucypher` <-- would recommend the user does this outside of the Dockerfile (or a script for the user) and the Dockerfile itself focuses on taking the source and putting it into an image/environment where it can run.

> The technical "reasoning" has a bit to do with docker layer caching

KPrasch:
Related to #489

EvaLovlac3:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/), As far as I understand it's because docker layer caching uses COPY during multi-stage builds (see https://docs.docker.com/develop/develop-images/multistage-build/) - which we'll use with pyUmbral, for example.

FROM KPrasch

KPrasch:
Will reopen in another PR

PR #605, opened 2018-12-11
Ursula Development Configuration Test

KPrasch:
Includes a test that proves `UrsulaConfiguration` is temporary and ephemeral when called with `dev_mode=True`

Basically what we are trying to show here is that development configurations produce unique character instances, and that they use ephemeral resources, like the in-memory database and temporary disk storage

### DIFF
--- a/tests/config/test_ursula_config.py
+++ b/tests/config/test_ursula_config.py
@@ -1,0 +1,41 @@
+from constant_sorrow.constants import NO_KEYRING_ATTACHED, CERTIFICATE_NOT_SAVED
+from nucypher.characters.lawful import Ursula
+from nucypher.config.characters import UrsulaConfiguration
+
+
+from nucypher.config.storages import ForgetfulNodeStorage
+
+
+# @pytest.mark.parametrize("federated_only", [True, ])  TODO
+def test_ursula_development_configuration(federated_only=True):
+    config = UrsulaConfiguration(dev_mode=True, federated_only=federated_only)
+    assert config.is_me is True
+    assert config.dev_mode is True
+    assert config.keyring == NO_KEYRING_ATTACHED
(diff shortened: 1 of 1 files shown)


tests/config/test_ursula_config.py
```diff
@@ -0,0 +1,41 @@
...
+    ursula_one = config()
+
+    # Ensure we do in fact have an Ursula here
+    assert isinstance(ursula_one, Ursula)
+    assert len(ursula_one.checksum_public_address) == 42
+    assert ursula_one.federated_only is federated_only
+
+    # A Temporary Ursula
+    assert ursula_one.datastore.engine.url.database == ":memory:"
+    assert ursula_one.certificate_filepath is CERTIFICATE_NOT_SAVED
```
cygnusv: Just curious...shouldn't this be an equality comparison?
```suggestion
    assert ursula_one.certificate_filepath == CERTIFICATE_NOT_SAVED
```

tests/config/test_ursula_config.py
```diff
@@ -0,0 +1,41 @@
...
+    assert UrsulaConfiguration.TEMP_CONFIGURATION_DIR_PREFIX in ursula_one.keyring_dir
+    assert isinstance(ursula_one.node_storage, ForgetfulNodeStorage)
+    assert ursula_one.node_storage._name == ":memory:"
+
+    # Alternate way to produce a character with a direct call
+    ursula_two = config.produce()
+    assert isinstance(ursula_one, Ursula)
```
cygnusv: Perhaps you mean this?
```suggestion
    assert isinstance(ursula_two, Ursula)
```

tests/config/test_ursula_config.py
KPrasch: No - I really want to know that this value `is` the constant `CERTIFICATE_NOT_SAVED`

PR #617, opened 2018-12-15
High-level TesterBlockchain network bootstrap method

KPrasch:
We have a growing need to run test or scripts within the context of an already deployed nucypher network.

##### Proposed Solution

Overrides `TesterBlockchain.connect`; Disregarding arguments, and redirecting to the default PyEVM configuration with an `InMemoryContractRegistry` and deployer options set on `BlockchainDeployerInterface`.

##### Usage
```
testerchain, accounts = TesterBlockchain.bootstrap_network()
```

Once the nucypher network is deployed, you can use the `Agents` API as usual:
```
token_agent = NucypherTokenAgent(blockchain=testerchain)
miner_agent = MinerAgent(blockchain=testerchain)
policy_agent = PolicyAgent(blockchain=testerchain)
```

Included in automated testing in `estimate_gas.py`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,7 +32,4 @@ variables*.yml
 *ansible/*.retry
 *ansible/inventory/*
 .env
-examples/heartbeat_demo/*.json
-examples/heartbeat_demo/*.msgpack
-examples/heartbeat_demo/doctor-files/
-examples/heartbeat_demo/alicia-files/
+/tests/metrics/results/
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,10 @@ chains
 .ethash
 nucypher_cli/examples/examples-runtime-cruft/*
 nucypher_cli/examples/finnegans-wake.txt
+examples/heartbeat_demo/*.json
+examples/heartbeat_demo/*.msgpack
+examples/heartbeat_demo/doctor-files/
+examples/heartbeat_demo/alicia-files/
 mypy_reports/
 reports/
 test-*
--- a/nucypher/utilities/sandbox/blockchain.py
+++ b/nucypher/utilities/sandbox/blockchain.py
@@ -14,10 +14,20 @@ GNU General Public License for more details.
 You should have received a copy of the GNU General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
+import os
+
 from twisted.logger import Logger
 
 from constant_sorrow.constants import NO_BLOCKCHAIN_AVAILABLE
-from typing import List
+from typing import List, Tuple
+
+from nucypher.blockchain.eth.actors import Deployer
+from nucypher.blockchain.eth.agents import NucypherTokenAgent, MinerAgent, PolicyAgent
+from nucypher.blockchain.eth.constants import DISPATCHER_SECRET_LENGTH
+from nucypher.blockchain.eth.interfaces import BlockchainDeployerInterface
+from nucypher.blockchain.eth.registry import InMemoryEthereumContractRegistry
+from nucypher.blockchain.eth.sol.compile import SolidityCompiler
+from nucypher.config.constants import CONTRACT_ROOT
 from umbral.keys import UmbralPrivateKey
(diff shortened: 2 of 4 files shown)


nucypher/utilities/sandbox/blockchain.py
```diff
@@ -48,10 +58,12 @@ class TesterBlockchain(Blockchain):
     Blockchain subclass with additional test utility methods and options.
     """
 
+    _PROVIDER_URI = 'tester://pyevm'
     _instance = NO_BLOCKCHAIN_AVAILABLE
     _test_account_cache = list()
+    _default_test_accounts = 10
 
-    def __init__(self, test_accounts=None, poa=True, airdrop=True, *args, **kwargs):
+    def __init__(self, test_accounts=_default_test_accounts, poa=True, airdrop=True, *args, **kwargs):
```
jMyles: This is an odd pattern - what made you decide to do this?

KPrasch: To add the `10`?  Just because we always expect there to be => 10 accounts in tests.

jMyles: I meant the pattern of resolution here - you have `_default_test_accounts`, a class variable, but which is only ever used at class evaluation time to set the default of `__init__`... why not just either make it `10` in `__init__` or have `__init__` grab the class variable?

nucypher/utilities/sandbox/blockchain.py
```diff
@@ -143,3 +155,24 @@ def time_travel(self, hours: int=None, seconds: int=None, periods: int=None):
...
+        solidity_compiler = SolidityCompiler(test_contract_dir=CONTRACT_ROOT)
+        memory_registry = InMemoryEthereumContractRegistry()
+        interface = BlockchainDeployerInterface(provider_uri=cls._PROVIDER_URI, compiler=solidity_compiler, registry=memory_registry)
+        testerchain = TesterBlockchain(interface=interface, test_accounts=cls._default_test_accounts, airdrop=False)
+        return testerchain
+
+    @classmethod
+    def bootstrap_network(cls) -> Tuple['TesterBlockchain', List[str]]:
+
+        def __deploy_contracts(testerchain: TesterBlockchain) -> None:
```
jMyles: Why separate this logic into a function?  Why not just put these three lines in the place where this function is currently called?

KPrasch: Honestly - I think this was re-purposed code - Yeah this doesn't need to be enclosed.

:heavy_check_mark:

nucypher/utilities/sandbox/blockchain.py
KPrasch: Hmm.. I *think* I want this at the instance level ultimately.  :heavy_check_mark:

tests/metrics/estimate_gas.py
```diff
@@ -165,25 +128,29 @@ def estimate_gas(analyzer: AnalyzeGas = None) -> None:
     #
     # Setup
     #
+
     if AnalyzeGas is None:
```
cygnusv: Perhaps you mean this?
```suggestion
    if analyzer is None:
```
(This GH suggestion may be buggy, it seems it removes a white-space even if the number of spaces is correct. I'm not sure if the patch is wrong, or only the rendering here)

.gitignore
```diff
@@ -32,7 +32,4 @@ variables*.yml
 *ansible/*.retry
 *ansible/inventory/*
 .env
-examples/heartbeat_demo/*.json
-examples/heartbeat_demo/*.msgpack
-examples/heartbeat_demo/doctor-files/
-examples/heartbeat_demo/alicia-files/
```
cygnusv: Don't remove this!

KPrasch: :heavy_check_mark: (restored)

tests/metrics/estimate_gas.py
KPrasch: Whoops! :heavy_check_mark:

PR #618, opened 2018-12-16
Adapt Federated

KPrasch:


pyproject.toml
```diff
@@ -0,0 +1,54 @@
...
+authors = ["NuCypher"]
+license = "GPL-3.0"
+
+[tool.poetry.dependencies]
+python = "^3.6"
+
+#
+# NuCypher
+#
+umbral = "==0.1.2a2"
```
cygnusv: We should be using `==0.1.3a0`. I'm afraid this file may be inconsistent with current Pipfile on master.

cygnusv: Perhaps we should remove poetry config for the moment. Later on we can include these files in a dedicated PR, including changes to integrate it with CI

KPrasch: :checkered_flag: PR #622

cygnusv: You're fast, man!

PR #622, opened 2018-12-18
[WIP] Initial stab at poetry config

KPrasch:


pyproject.toml
```diff
@@ -0,0 +1,54 @@
...
+authors = ["NuCypher"]
+license = "GPL-3.0"
+
+[tool.poetry.dependencies]
+python = "^3.6"
+
+#
+# NuCypher
+#
+umbral = "==0.1.2a2"
```
cygnusv: This has to be updated to `0.1.3a0`.

pyproject.toml
```diff
@@ -0,0 +1,54 @@
...
+constant-sorrow = ">=0.1.0a7"
+bytestring-splitter = "1.0.0a4"
```
cygnusv: Is this the right syntax? Perhaps you mean `"==1.0.0a4"`?

PR #641, opened 2018-12-21
Separate automated docker build from upload steps

### LINKED ISSUE #628: Separate Publication Builds from Uploads
We want to ensure that a failing docker build on circleICI does not cause the entire workflow to fail, in the case of PyPI upload success.

A major improvement will be to separate the build and publication steps for tagged releases, placing the "request approval" between them; This will give us more opportunity to fix Issues without making extra releases.

KPrasch:
Fixes #628 (Docker) and multiple bug fixes in `.cirelce/config.yml`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -158,6 +158,15 @@ workflows:
           requires:
             - cli
             - ursula_command
+      - build_docker:
+          filters:
+            tags:
+              only: /v[0-9]+.*/
+            branches:
+              ignore: /.*/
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -198,13 +208,13 @@ workflows:
 
 python_36_base: &python_36_base
   parallelism: 1
-  working_directory: ~/nucypher-depends
```
tuxxy: :pray:

PR #652, opened 2019-01-07
Federated Ursula Provisioning API Updates

KPrasch:
- Provisions federated ursulas with ansible - using an inventory group named by nucypher network domain and nucypher node role.
- Manage Nucypher dependencies with pip; Deprecate provisioning via pipenv
- Update CLI usage to latest usage


KPrasch:
@tuxxy - Turns out this is correct - it creates dynamic groups ie `ropsten-ursulas`

PR #676, opened 2019-01-21
Make domain constants bytestrings

### LINKED ISSUE #671: Human-readable domain serialization without Constant Sorrow
At the moment domains are specified as constant sorrow constants, which introduces some complexity regarding node and configuration serialisation.  Instead of using constants, use bytes-serializable strings to specify domains.

tuxxy:
### What this does:
1. Deserializes domain constants from config files as UTF-8 bytestrings.
    - They get serialized as strings.
2. Moves `GLOBAL_DOMAIN` definition to `config.constants`
3. Removes an unused `logging.Logger` import that shadows a `twisted.Logger` import.
4. Resolves #671

### DIFF
--- a/nucypher/config/node.py
+++ b/nucypher/config/node.py
@@ -33,8 +33,6 @@ from cryptography.x509 import Certificate
 from twisted.logger import Logger
 from umbral.signing import Signature
 
-from constant_sorrow import constants
-from constant_sorrow.constants import GLOBAL_DOMAIN
 from constant_sorrow.constants import (
     UNINITIALIZED_CONFIGURATION,
     STRANGER_CONFIGURATION,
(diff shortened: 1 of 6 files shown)


nucypher/config/node.py
```diff
@@ -374,8 +372,8 @@ def from_configuration_file(cls, filepath: str = None, **overrides) -> 'NodeConf
                                                   serializer=cls.NODE_SERIALIZER,
                                                   deserializer=cls.NODE_DESERIALIZER)
 
-        # Deserialize domains to Constants vis Constant Sorrow
-        domains = set(getattr(constants, domain.upper()) for domain in payload['domains'])
+        # Deserialize domains to UTF-8 bytestrings
+        domains = list(domain.encode() for domain in payload['domains'])
```
KPrasch: Ah - here it is.

PR #677, opened 2019-01-21
Updates Finnegan's Wake Demo API

KPrasch:
Simplifies and updates Finnegan's Wake Demo


examples/finnegans_wake_demo/finnegans-wake--demo.py
```diff
@@ -19,32 +19,20 @@
...
-os.mkdir(TEMP_CERTIFICATE_DIR)
+globalLogPublisher.addObserver(SimpleObserver())
 
 
 #######################################
 # Finnegan's Wake on NuCypher Testnet #
 # (will fail with bad connection) #####
 #######################################
 
-TESTNET_LOAD_BALANCER = "eu-federated-balancer-40be4480ec380cd7.elb.eu-central-1.amazonaws.com"
```
tuxxy: :+1:

examples/finnegans_wake_demo/finnegans-wake--demo.py
```diff
@@ -19,32 +19,20 @@
...
-# Temporary file storage
-TEMP_FILES_DIR = "{}/examples-runtime-cruft".format(os.path.dirname(os.path.abspath(__file__)))
-TEMP_DEMO_DIR = "{}/finnegans-wake-demo".format(TEMP_FILES_DIR)
-TEMP_CERTIFICATE_DIR = "{}/certs".format(TEMP_DEMO_DIR)
-
-# Remove previous demo files and create new ones
-shutil.rmtree(TEMP_FILES_DIR, ignore_errors=True)
-os.mkdir(TEMP_FILES_DIR)
-os.mkdir(TEMP_DEMO_DIR)
-os.mkdir(TEMP_CERTIFICATE_DIR)
```
jMyles: Nice redlines here for sure.

examples/finnegans_wake_demo/finnegans-wake--demo.py
```diff
@@ -19,32 +19,20 @@
...
+globalLogPublisher.addObserver(SimpleObserver())
 
 
 #######################################
 # Finnegan's Wake on NuCypher Testnet #
 # (will fail with bad connection) #####
 #######################################
 
-TESTNET_LOAD_BALANCER = "eu-federated-balancer-40be4480ec380cd7.elb.eu-central-1.amazonaws.com"
+SEEDNODE_URI = "https://localhost:11501"
```
jMyles: Well, by hardcoding this, are we back to presuming that this will *always* be run against a local Ursula (and never a testnet)?

KPrasch: No - but its a better default, since it's fairly easy to change the demo code (even encouraged).  Also, this is still a decent way to connect to a live network, if your local node is running on that domain.

jMyles: So, if you're presenting this, you'll just have a URL sitting around to paste in during the demo?

Seems reasonable to me.

KPrasch: Yeah - that's more or less what I'm thinking. (perhaps there will be something easier wrt domains later)

PR #686, opened 2019-01-25
[Docs] Clean up README and fix logo

tuxxy:


README.md
```diff
@@ -1,24 +1,15 @@
...
 
-.. image:: https://circleci.com/gh/nucypher/nucypher/tree/master.svg?style=svg
-    :target: https://circleci.com/gh/nucypher/nucypher/tree/master
+----
 
+![](https://img.shields.io/pypi/wheel/nucypher.svg)
+![](https://img.shields.io/pypi/pyversions/nucypher.svg)
+![](https://coveralls.io/repos/github/nucypher/nucypher/badge.svg?branch=master)
+![](https://circleci.com/gh/nucypher/nucypher/tree/master.svg?style=svg)
+![](/docs/source/.static/img/GPLv3.png)
```
KPrasch: How about  ->

KPrasch:
Can you squash some of these commits to ~3?

README.md
tuxxy: Why does it say missing? I just use the GPLv3 logo. It's not like we need to show the status of our license lol

KPrasch: It says missing because the package it not yet on PyPI - No - but I prefer the uniformity pf the badges

KPrasch: Plus - that logo looks like a 90's power ranger

tuxxy: Let's do it incrementally, once we get the package on PyPI then let's switch to that badge.

tuxxy: I don't want it to say "license: missing" when it really shouldn't.

KPrasch: Well - we already do this in several places in the docs (the other badges for example), and I anticipate the period of time before publication to be small - I think a follow-up PR on this basis is overkill.

README.md
```diff
@@ -1,24 +1,21 @@
...
 
+![](https://img.shields.io/pypi/wheel/nucypher.svg)
+![](https://img.shields.io/pypi/pyversions/nucypher.svg)
+![](https://coveralls.io/repos/github/nucypher/nucypher/badge.svg?branch=master)
+![](https://circleci.com/gh/nucypher/nucypher/tree/master.svg?style=svg)
+![](/docs/source/.static/img/GPLv3.png)
```
KPrasch: These badges are included twice

PR #691, opened 2019-01-27
Real-time learning monitor

jMyles:
* Introducing `Moe`, an ultralight `Character` which reports results of `learn_from_teacher_node` and `remember_node` via websocket (using the hendrix `hey_joe` API).  `Moe` also uses a customized `FleetStateTracker` which similarly reports changes to the recorded fleet state.

* Provides a small Flask application to access a real-time status page.

* Updates real-time fleet information using handlebars templates.

### DIFF
--- a/deploy/learning_monitor.py
+++ b/deploy/learning_monitor.py
@@ -1,0 +1,63 @@
+import sys
+
+from flask import Flask, render_template
+from twisted.logger import globalLogPublisher
+
+from hendrix.deploy.base import HendrixDeploy
+from hendrix.experience import crosstown_traffic, hey_joe
+from nucypher.characters.base import Character
+from nucypher.characters.lawful import Ursula
+from nucypher.config.constants import GLOBAL_DOMAIN
+from nucypher.network.middleware import RestMiddleware
+from nucypher.utilities.logging import SimpleObserver
+
+websocket_service = hey_joe.WebSocketService("127.0.0.1", 9000)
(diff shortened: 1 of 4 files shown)


KPrasch:
needs rebase!

deploy/learning_monitor.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+@rest_app.route("/")
+def status():
+
+
+        # for node in monitor.known_nodes:
+        #     hey_joe.send(node.status_json(), topic="nodes")
+
+    return render_template('monitor.html')
```
michwill: Is it possible to do so that it doesn't require to be in `deploy` as a current directory?
Right now, running `python3 deploy/learning_monitor.py ...` will produce `TemplateNotFound`

KPrasch: IMO, let's make an Issue out of this rather than include it in this PR - but I do agree :100:

michwill: Well, the fix is very trivial though:
```
import os.path
...
rest_app = Flask("fleet-monitor", root_path=os.path.dirname(__file__))
```
... of course, if setup.py will save the template alongside the file

michwill: Probably easy enough to fix here https://github.com/jMyles/nucypher/pull/2/files

KPrasch:
I can imagine this getting it's own character CLI entry.

PR #699, opened 2019-01-29
[WIP] Touchup to Keyring

cygnusv:


cygnusv:
Closing this for the moment. In any case, merging this could make fail previous installations due to changes in the Keyring, so we should perhaps synchronize it with a new deployment stage of testnet (or mainnet?).

PR #704, opened 2019-01-30
[WIP] Documentation index/landing page: more compelling

arjunhassard:
Generally aiming to remedy the assumption that inbound developers have the time or energy to commit to reading through pages and pages of documentation in order to understand a complex system, without first being convinced that it is exciting and/or of use to them. 

Initial addition: example code blocks accompanied by a (slightly silly) narrative to give arrivals a non-intimidating taste of the project's context, capabilities and style


docs/source/index.rst
```diff
@@ -34,6 +40,28 @@ NuCypher
 - Website https://www.nucypher.com/
 
 
+A short and efficient love story 
+--------------------------------
+
+Alice composes and sends a romantic poem to her lover, Bob. Sadly, she then drops her device into a pond. A few hours later, Bob turns on his mobile. How will he receive the poem?
```
KPrasch: :heart_eyes:

docs/source/index.rst
```diff
@@ -34,6 +40,28 @@ NuCypher
...
 
 
+A short and efficient love story 
+--------------------------------
+
+Alice composes and sends a romantic poem to her lover, Bob. Sadly, she then drops her device into a pond. A few hours later, Bob turns on his mobile. How will he receive the poem? 
+
+.. code::
+
+    $ delivered_cleartexts = BOB.retrieve(message_kit=single_passage_ciphertext,
```
tuxxy: Why `$`?

docs/source/index.rst
```diff
@@ -34,6 +40,28 @@ NuCypher
...
+                                          data_source=enrico_as_understood_by_bob, 
+                                          alice_verifying_key=alice_pubkey_restored_from_ancient_scroll)
+
+
+Bob's .retrieve() power sets off a series of exciting events, leading to the involvement of Ursula, an untrusted but diligent *proxy*. Ursula takes the ciphertext associated with Alice's message, then uses a special *re-encryption key* to transform it into a form decryptable by Bob. 
+
+.. code::
```
KPrasch:
``` suggestion
.. code-block:: python
```

docs/source/index.rst
```diff
@@ -34,6 +40,28 @@ NuCypher
 - Website https://www.nucypher.com/
 
 
+A short and efficient love story 
+--------------------------------
+
+Alice composes and sends a romantic poem to her lover, Bob. Sadly, she then drops her device into a pond. A few hours later, Bob turns on his mobile. How will he receive the poem? 
+
+.. code::
```
KPrasch:
``` suggestion
.. code-block:: python
```

docs/source/index.rst
derekpierre: Interesting angle with the story @arjunhassard 👍.

Even though I know we are trying to keep the story short and efficient, I wonder if the example may need to a bit meatier. In the current scenario, Alice could just send the seemingly short poem encrypted using Bob's public key, and that's it - regardless of whether she loses her device or not.

I'm still trying to form a cohesive story in my mind as you did, but what about something along the lines of :
1. Alice has been performing a reconnaissance mission and collecting data for several days/weeks storing her findings in encrypted files using her public key on some remote filesystem. Through her persistence, she has uncovered some very important information which if it were to become public would be monumental
2. She notices some men in suits walking toward her, who will in all likelihood want to question her and confiscate her phone
3. She decides that it would be best to have Bob (a confidante) gain access to the data, for safe-keeping...just in case 😨 
3. Her encrypted data files are large and she has no time to decrypt the data and then encrypt it for Bob as is necessary for regular public-key cryptography
4. Instead of getting her phone confiscated, Alice wants to destroy it, but if she does her private key to decrypt her stored data goes away as well. 
5. Destroying her phone prevents the men in suits from using her private key to access her data, but renders the data inaccessible to anyone in the future including her
6. Her conundrum: How can she destroy her phone and still allow Bob to access her data without sending Bob her private key?

I'm sure the scenario can be changed/improved, but I think it helps showcase similar thoughts you may have had. We could even throw multiple people in there (eg. a Bob and Charlie) to showcase scalability.

I'm sure @jMyles would want to take a look at this.

derekpierre: Perhaps we can also add a code block of Alice granting access to Bob?

arjunhassard: @derekpierre This is great! Number of things to consider here:
- How to organise code blocks, examples, use cases, walkthroughs, demos etc. On the front page, I think the ideas conveyed should be very simple and short as the first priority, and hint at some unique capabilities as the second. Someone arriving there for the first time should see that they can quickly consume a section of ~100 words. I do think the story you've written is an excellent way to illustrate NC's censorship resistant qualities, and may deserve a whole section focusing on this. As for scaleability, I think we need more use cases plugging this angle – especially because if we want overall demand/fees to be non-negligible, we need adopting platforms that require billions of re-encryptions per month, not thousands.  
- NC vs. client-side basic PKI. Using the latter technique, if Alice sends a message to Bob while his device is off, and loses her device before he turns it on, then where would that message (/sym key) be stored in the interim? In most apps, access to the message is 'kept' on a trusted server. I think Signal has a way around this, but with some drawbacks. Basically, I've attempted to take a scenario that's common and easy to follow (breaking your phone), and marry to one NC utility: having an always-online third party managing access on your behalf, minus trusting them.

arjunhassard: I considered adding Alice.grant() - so open to it. My only hesitation was having to explain the concept of a NC sharing policy so early on

docs/source/index.rst
arjunhassard: Error, will remove it, thanks

docs/source/index.rst
derekpierre: I think in that case the messaging system context -- messaging server (or Signal's alternative), that Alice may have previously issued a re-encryption key i.e. some sort of sharing arrangement with Bob -- should be mentioned albeit briefly. 

Maybe the context should not be here but rather after you indicate Bob.retrieve(). Without it, IMHO, the scenario seems disjointed.

KPrasch:
@arjunhassard - Are we still WIP here?

PR #710, opened 2019-01-30
Don't require to change directory to where the script is

michwill:
+ some PEP8-worthy changes


cygnusv:
Just being picky: can you rebase instead of merging? We try to get merge commits (e.g., 54c49bb) to a minimum.

PR #714, opened 2019-01-31
[Hotfix] Fix leftover attr from APIStar -> Flask transition

tuxxy:


KPrasch:
Looks like someone didn't run the tests :-P

PR #719, opened 2019-02-02
Another small step with a few small fixes.

jMyles:
* Displays node icons on `Moe` monitor
* Displays better status icons on `Moe` monitor
* Fixes `test_emit_warning_upon_new_version`, which previously relied on an apistar `Response` object


KPrasch:
Included in #721

PR #721, opened 2019-02-03
Reworking middleware to share more logic.

jMyles:
* Creates new shared middleware `_Client`, which wraps all calls (production and test) to network library code (ie, requests and werkzeug). 
* "cleans" all returning responses, ensuring that `content` is an available attr in either case.
* Moved status-code checking up to the client so that it happens all in one place.
* Raises `NotFound` if an endpoint returns a 404, `UnexpectedResponse` for other > 300 responses.

### DIFF
--- a/nucypher/network/middleware.py
+++ b/nucypher/network/middleware.py
@@ -42,7 +42,7 @@ class RestMiddleware:
     class _Client:
 
         def __init__(self, library, response_cleaner=None):
-            if self.response_cleaner is not None:
+            if response_cleaner is not None:
                 self.response_cleaner = response_cleaner
             else:
                 self.response_cleaner = lambda r: r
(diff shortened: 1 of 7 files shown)


nucypher/network/middleware.py
```diff
@@ -28,18 +28,41 @@
...
+            self._library = library
+
+        def __getattr__(self, item):
+            def method_wrapper(*args, **kwargs):
+                method = getattr(self._library, item)
+                response = method(*args, **kwargs)
+                cleaned_response = self.response_cleaner(response)
+                if cleaned_response.status_code >= 300:
+                    if cleaned_response.status_code == 404:
+                        raise NotFound("Server claims not to have found it: {}".format(cleaned_response.content))
```
michwill: Maybe we can also say, what `item` it actually was?

michwill: Although, perhaps, not as important b/c it's just get/post usually

nucypher/network/middleware.py
```diff
@@ -28,18 +28,41 @@
...
+
 class RestMiddleware:
     log = Logger()
 
-    def consider_arrangement(self, arrangement):
-        node = arrangement.ursula
-        response = requests.post("https://{}/consider_arrangement".format(node.rest_interface),
-                                 bytes(arrangement),
-                                 verify=node.certificate_filepath, timeout=2)
+    class _Client:
```
KPrasch: Hmmm, curious

nucypher/network/middleware.py
```diff
@@ -28,18 +28,41 @@
...
-        return response
+        def __init__(self, library, response_cleaner=None):
+            if response_cleaner is not None:
+                self.response_cleaner = response_cleaner
+            else:
+                self.response_cleaner = lambda r: r
+
+            self._library = library
+
+        def __getattr__(self, item):
```
KPrasch: ...Curious indeed.

michwill: Yeah, I _love_ hacks like that

PR #740, opened 2019-02-09
Integrate py-solc, v0.5.0+ contract updates

KPrasch:
*note: `nucypher/py-solc` is incompatible with windows.*


nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -158,15 +158,15 @@ contract Issuer is Upgradeable {
...
+        require(address(uint160(delegateGet(_testTarget, "token()"))) == address(token));
+        require(delegateGet(_testTarget, "miningCoefficient()") == miningCoefficient);
+        require(delegateGet(_testTarget, "lockedPeriodsCoefficient()") == lockedPeriodsCoefficient);
         require(uint32(delegateGet(_testTarget, "secondsPerPeriod()")) == secondsPerPeriod);
         require(uint16(delegateGet(_testTarget, "rewardedPeriods()")) == rewardedPeriods);
         require(uint16(delegateGet(_testTarget, "lastMintedPeriod()")) == lastMintedPeriod);
-        require(uint256(delegateGet(_testTarget, "currentSupply1()")) == currentSupply1);
-        require(uint256(delegateGet(_testTarget, "currentSupply2()")) == currentSupply2);
-        require(uint256(delegateGet(_testTarget, "totalSupply()")) == totalSupply);
+        require(delegateGet(_testTarget, "currentSupply1()") == currentSupply1);
```
jMyles: Is this happening because you don't need to cast to `uint256` anymore in 0.5?

nucypher/blockchain/eth/sol/source/contracts/UserEscrow.sol
```diff
@@ -108,17 +108,17 @@ contract UserEscrow is Ownable {
...
-        emit ETHWithdrawn(owner, balance);
+        msg.sender.transfer(balance);
+        emit ETHWithdrawn(msg.sender, balance);
     }
 
     /**
     * @dev Fallback function send all requests to the target proxy contract
     **/
-    function () public payable onlyOwner {
+    function () external payable onlyOwner {
```
jMyles: Why the change from public to external here?

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
szotov: casting was restricted in 0.5, because of different padding for different types (bytes and uint). With this changings method `delegateGet` (in `Upgreadable`) have padding like `uint256` and not how `bytes32`, so changed return type and using of this method

nucypher/blockchain/eth/sol/source/contracts/proxy/Dispatcher.sol
```diff
@@ -81,9 +85,9 @@ contract Dispatcher is Upgradeable {
     * @dev Fallback function send all requests to the target contract.
     * If contract not exists then result will be unpredictable (see DELEGATECALL)
     **/
-    function () public payable {
-        assert(target != 0x0);
-        bool callSuccess = target.delegatecall(msg.data);
+    function () external payable {
```
jMyles: Here too - why public to external?

nucypher/blockchain/eth/sol/source/contracts/UserEscrow.sol
szotov: because of breaking changes in v0.5.0
https://solidity.readthedocs.io/en/v0.5.0/050-breaking-changes.html

nucypher/blockchain/eth/sol/source/contracts/proxy/Dispatcher.sol
```diff
@@ -81,9 +85,9 @@ contract Dispatcher is Upgradeable {
...
+        assert(target != address(0));
+        (bool callSuccess,) = target.delegatecall(msg.data);
```
jMyles: The addition of these parentheses and comma are foreign to me.  What do we have happening here?  `delegatecall` started returning a tuple?  When previously a returned a single object?

nucypher/blockchain/eth/sol/source/contracts/proxy/Dispatcher.sol
szotov: the same, now fallback function only external

nucypher/blockchain/eth/sol/source/contracts/proxy/Dispatcher.sol
szotov: yep, you're right
second value `bytes memory data`

jMyles: I don't recall every having seen a tuple with an anonymous member to the left of an assignment operator.  Is this typical in solidity 0.5+?

nucypher/blockchain/eth/sol/source/contracts/UserEscrow.sol
jMyles: I still don't get it though - doesn't this change the visibility of this function?

nucypher/blockchain/eth/sol/source/contracts/proxy/Dispatcher.sol
szotov: It was added in some 0.4.2? version

nucypher/blockchain/eth/sol/source/contracts/UserEscrow.sol
szotov: yes, `external` is more strict than `public`. `public` means that method also can be called from this contract, but `fallback` function can be called only outside of the contract. So for `fallback` function `public` works like `external` and in 0.5+ only `external` is allowed

PR #743, opened 2019-02-10
Character Control for Alice, Bob, and Enrico

tuxxy:
### What this does:
1. Adds three character control flask applications:
    - `make_alice_control`
    - `make_bob_control`
    - `make_enrico_control`
    - All of these take a "drone" character and a teacher `Ursula`, except `make_enrico_control` which only takes a `drone_enrico`.
2. Adds endpoints for `Alice`:
     - `/create_policy` -- Undocumented
     - `/grant`
3. Adds endpoints for `Bob`:
    - `/join_policy` -- Undocumented
    - `/retrieve`
4. Adds an endpoint for `Enrico`:
    - `/encrypt_message`
5. Adds documentation for all application-ready endpoints in `character_control_guide.rst`.
6. Adds a boolean param to `Teacher.seed_node_metadata` called `as_teacher_uri` which defaults to `False`.
    - When `True`, in lieu of a `SeednodeMetadata` object, it will return the seednode metadata as a formatted teacher URI string.
7. Makes the `m` and `destinations` attributes on `TreasureMap` properties.
    - Raises a `TypeError` if it hasn't been decrypted yet.
8. Adds tests for all character control endpoints
9. Adds `test_treasure_map_serialization` test.
10. Adds fixtures for `alice_control`, `bob_control`, and `enrico_control` which are used as Flask test clients.
11. Some tiny fixes to the documentation index page.

Note: all undocumented endpoints aren't ready to be used yet. They're included in the PR for posterity and have working tests. Anyone who finishes these methods should also include documentation in the character control guide.


nucypher/network/character_control/alice_control.py
```diff
@@ -24,20 +26,42 @@ def create_policy():
...
+            bob = Bob.from_public_keys({DecryptingPower: bob_pubkey,
+                                        SigningPower: None},
+                                       federated_only=True)
+        except KeyError as e:
+            return Response(str(e), status=500)
+
+        new_policy = drone_alice.grant(bob, label, m=m, n=n,
+                                       expiration=expiration_time)
+        # TODO: Serialize the policy
+        import pudb; pudb.set_trace()
```
jMyles: Even in these intermediate commits, try not to commit breakpoints.

nucypher/network/nodes.py
```diff
@@ -855,7 +855,10 @@ def from_tls_hosting_power(cls, tls_hosting_power: TLSHostingPower, *args, **kwa
     # Known Nodes
     #
 
-    def seed_node_metadata(self):
+    def seed_node_metadata(self, as_teacher_uri=False):
+        if as_teacher_uri:
```
jMyles: I think I'd prefer if we give these local names and then use them in each logic branch (to prevent refactor divergence later).

tests/network/test_character_control.py
```diff
@@ -99,3 +100,29 @@ def test_bob_character_control_retrieve(bob_control, enacted_federated_policy, c
...
+    assert 'message_kit' in response_data['result']
+    assert 'signature' in response_data['result']
+
+    # Check that it serializes correctly.
+    message_kit = UmbralMessageKit.from_bytes(
+                            b64decode(response_data['result']['message_kit']))
+
+    # Send bad data to assert error return
+    response = enrico_control.post('/encrypt_message', data='bad')
+    assert response.status_code == 400
```
jMyles: Not immediately evident to me why this is a 400 and not a 5xx - is this the problem you were telling me about today?

docs/source/guides/character_control_guide.rst
```diff
@@ -0,0 +1,159 @@
+================================
+NuCypher Character Control Guide
+================================
+
+The `Character Control` is a module that contains useful REST-like HTTP endpoints for working with NuCypher characters.
```
jMyles: "REST-inspired"?

nucypher/network/character_control/alice_control.py
```diff
@@ -0,0 +1,84 @@
...
+from nucypher.characters.lawful import Alice, Bob, Ursula
+from nucypher.crypto.powers import DecryptingPower, SigningPower
+
+
+def make_alice_control(drone_alice: Alice, teacher_node: Ursula):
+    alice_control = Flask("alice-control")
+
+    teacher_node.verify_node(drone_alice.network_middleware)
+    drone_alice.remember_node(teacher_node)
+    drone_alice.start_learning_loop(now=True)
```
jMyles: As we evolve, we won't want this to be `now=True` except in tests.

docs/source/guides/character_control_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+        'label': 'dGVzdA==',
+        'm': 2,
+        'n': 3,
+        'expiration_time': '2019-02-14T22:23:10.771093Z',
+    }
+
+Take a look at ``bob_encrypting_key``. Take note that it's a hex-encoded string.
+The character control API endpoints expect `all` keys to be encoded as hex.
+
+Now, look at ``label``. Notice that it's a base64-encoded string.
```
michwill: Interesting. So, we wouldn't support using ASCII labels like `"my_label"` in the API to avoid ambiguity and use `base64` only?

nucypher/network/character_control/alice_control.py
```diff
@@ -0,0 +1,84 @@
...
+        """
+        # TODO: Needs input cleansing and validation
+        # TODO: Provide more informative errors
+        try:
+            request_data = json.loads(request.data)
+
+            bob_pubkey = bytes.fromhex(request_data['bob_encrypting_key'])
+            label = b64decode(request_data['label'])
+            # TODO: Do we change this to something like "threshold"
+            m, n = request_data['m'], request_data['n']
```
michwill: Should we check `isinstance(m, int)`?

KPrasch: Hmm - in order to *force* an early failure?

docs/source/guides/character_control_guide.rst
mswilkison: I think we need a section (perhaps on index.rst or a separate page) that explains each of the different characters and their roles in the network in clear, plain language. Right now, the docs feel like they're assuming a familiarity with the characters. Maybe this is captures in @arjunhassard's PR.

docs/source/guides/character_control_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+=======================================
+
+Status Codes
+------------
+All documented API endpoints use JSON and are REST-like.
+
+Some common returned status codes you may encounter are:
+
+- ``200 OK`` -- The request has succeeded.
+- ``400 BAD REQUEST`` -- The server cannot or will not process the request do to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).
```
mswilkison:
```suggestion
- ``400 BAD REQUEST`` -- The server cannot or will not process the request due to something that is perceived to be a client error (e.g., malformed request syntax, invalid request message framing, or deceptive request routing).
```

docs/source/guides/character_control_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+- ``500 INTERNAL SERVER ERROR`` -- The server encountered an unexpected condition that prevented it from fulfilling the request.
+
+Typically, you will want to ensure that any given response from a character control endpoint results in a 200 status code.
+This tells you that the server successfully completed the call.
+
+If you are returned a 400, check the data that you're sending to the server.
```
mswilkison:
```suggestion
If you are returned a 400, check that you're not sending invalid data to the server.
```

docs/source/guides/character_control_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+Chances are you might be sending some invalid data to the server.
```
mswilkison:
```suggestion
```

docs/source/guides/character_control_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+This endpoint controls the ``Alice.grant`` method.
+
+- URL: ``/grant``
+- HTTP Method: ``PUT``
+- Required arguments:
+    - ``bob_encrypting_key`` -- encoded as hex
+    - ``label`` -- encoded as base64
+    - ``m`` -- an integer
+    - ``n`` -- an integer
+    - ``expiration_time`` -- an ISO-8601 formatted datetime string.
```
mswilkison:
```suggestion
    - ``expiration_time`` -- an ISO-8601 formatted datetime string
```

docs/source/guides/character_control_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+- Returns: a base64-encoded ``treasure_map``
+
+For more details on these arguments, see the nucypher documentation on the ``Alice.grant`` Python API method.
```
mswilkison: Where is this? Is it not written yet?

docs/source/guides/character_control_guide.rst
```diff
@@ -0,0 +1,159 @@
...
+
+This endpoint controls the ``DataSource.encrypt_message`` method.
+
+- URL: ``/encrypt_message``
+- HTTP Method: ``POST``
+- Required arguments:
+    - ``message`` -- encoded as base64
+- Returns: ``message_kit`` and ``signature`` encoded as base64
+
+For more details on these arguments, see the nucypher documentation on the ``DataSource.encrypt_message`` Python API method.
```
mswilkison: Where is this? Is it not written yet?

docs/source/guides/character_control_guide.rst
cygnusv: I think using base64 is wise, since there may be use cases where the label is a bytes sequence.

tuxxy: It's only base64 as a serialization layer over the wire since it can be any bytestring.

docs/source/guides/character_control_guide.rst
tuxxy: The character control guide is not, imho, meant for explanation of roles. This is strictly for developers who need access to an API. They should familiarize theirselves elsewhere in our documentation.

With that said, I would love to merge @arjunhassard's docs ASAP.

docs/source/guides/character_control_guide.rst
tuxxy: I'm.... not sure. I was just guessing we would add some at some point. It should probably direct folks to the code itself.

docs/source/guides/character_control_guide.rst
tuxxy: I'm.... not sure. I was just guessing we would add some at some point. It should probably direct folks to the code itself.

nucypher/network/character_control/alice_control.py
tuxxy: We can if we want to prevent a 500 for anything. I don't think we should do this now because why not check _all_ the byte strings if they're `bytes` first? We certainly can, but I think this is fine for now.

nucypher/network/control/bob.py
```diff
@@ -0,0 +1,79 @@
...
+
+        drone_bob.join_policy(label=label, alice_pubkey_sig=alice_pubkey_sig)
+
+        return Response('Policy joined!', status=200)
+
+
+    @bob_control.route('/retrieve', methods=['POST'])
+    def retrieve():
+        """
+        Character control endpoint for re-encrypting and decrypting policy
```
cygnusv: Perhaps it should be rephrased to avoid implying that Bob re-encrypts.

tests/network/test_character_control.py
```diff
@@ -0,0 +1,131 @@
...
+from nucypher.crypto.powers import DecryptingPower
+from nucypher.policy.models import TreasureMap
+
+
+def test_alice_character_control_create_policy(alice_control, federated_bob):
+    bob_pubkey_enc = federated_bob.public_keys(DecryptingPower)
+
+    request_data = {
+        'bob_encrypting_key': bytes(bob_pubkey_enc).hex(),
+        'label': b64encode(bytes(b'test')).decode(),
```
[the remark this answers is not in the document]
tuxxy: It needs to be transferred over the wire in a compatible format. Yes, we could probably encode it from `bytes` to UTF-8, but we wouldn't be able to use non-UTF-8 characters (like random bytes) in the label anymore.

PR #752, opened 2019-02-12
[Hotfix] Update and Test Moe Network Monitor

KPrasch:


deploy/services/moe_monitor.j2
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description="Run 'Moe', A NuCypher Network Monitor."
+
+[Service]
+User=ubuntu
```
tuxxy: Ubuntu specific?

KPrasch: Opened Issue #753

PR #762, opened 2019-02-14
Fix bob_pubkey_sig issue for the character control lifecycle

tuxxy:


tuxxy:
This PR is b u s t e d. Closing for #763 instead.

PR #770, opened 2019-02-15
Updates version of solidity in vodka

### LINKED ISSUE #647: Update to latest solc (0.5)
There's a few issues when running with the latest solidity compiler. One involves the `--combined-json` flag, I think. This needs investigation.

szotov:
Fixes #647


KPrasch:
Needs Rebase!

szotov:
> Am I right that `bytes32` implies `memory`?

You could say that, all variable that fit in one word (32 bytes) are placed in memory when using in functions.
Arrays contain more than 32 -> length (32 bytes) + data. So `bytes` and `bytes32` are completely different types

Pipfile
```diff
@@ -21,7 +21,7 @@ sqlalchemy = "*"
 maya = "*"
 flask = "*"
 # Third-Party Ethereum
-py-evm = "*"
+py-evm = "==0.2.0a39"
```
cygnusv: I guess pinning py-evm to this specific version is intended, right?
Also, apart from creating a Pipfile.lock, it is necessary to create the `requirements.txt` and `dev-requirements.txt` files: 
```
pipenv lock --three --clear --pre --requirements > requirements.txt
pipenv lock --three --dev --clear --pre --requirements > dev-requirements.txt
```

nucypher/blockchain/eth/sol/source/contracts/lib/SignatureVerifier.sol
```diff
@@ -42,16 +42,16 @@ library SignatureVerifier {
     * @notice Transform public key to address
     * @param _publicKey secp256k1 public key
     **/
-    function toAddress(bytes _publicKey) internal pure returns (address) {
-        return address(keccak256(_publicKey));
+    function toAddress(bytes memory _publicKey) internal pure returns (address) {
+        return address(uint160(uint256(keccak256(_publicKey))));
```
jMyles: What's happening here?  Is this related to a new explicitness requirement?

Do you think that this pattern will be (increasingly?) typical of solidity 0.5+?  It seems pretty unusual to have a branchless doulbe-cast like this.

szotov: Seems that it will be in all places with casting `address` to something. Before 0.5.0 this casting was automatic, but now because of padding only manual casting is allowed. It's looks like `bytes32` -> `uint256` -> `uint160` -> `address`. Can cast something with 20 bytes to `address` and `bytes??` type can cast only to uint/int with the same size

cygnusv: @jMyles Yes, it's a new explicitness requirement (although apparently it's been there since 0.4.24, but perhaps it was not an error before 0.5).

From the [Solidity docs](https://solidity.readthedocs.io/en/latest/types.html#address):

> If you convert a type that uses a larger byte size to an address, for example bytes32, then the address is truncated. To reduce conversion ambiguity version 0.4.24 and higher of the compiler force you make the truncation explicit in the conversion. Take for example the address 0x111122223333444455556666777788889999AAAABBBBCCCCDDDDEEEEFFFFCCCC.
> 
> You can use address(uint160(bytes20(b))), which results in 0x111122223333444455556666777788889999aAaa, or you can use address(uint160(uint256(b))), which results in 0x777788889999AaAAbBbbCcccddDdeeeEfFFfCcCc.

PR #781, opened 2019-02-17
[WIP] Heartbeat Demo UI with REST Endpoints

KPrasch:


examples/heartbeat_rest_ui/README.md
```diff
@@ -0,0 +1,56 @@
...
+  - Alicia, or any one knowing the policy public key (e.g., the Heart Monitor), can produce encrypted data that belongs 
+  to the policy. Again, this can happen before granting access to any consumer.
+  - As a consequence of the previous point, Data Sources, like the Heart Monitor, are completely unaware of the 
+  recipients. In their mind, they are producing data **for Alicia**.
+  - Alicia never interacts with the Doctor: she only needs the Doctor's public key.
+  - Alicia only interacts with the NuCypher network for granting access to the Doctor. After this, she can even 
+  disappear from the face of the Earth.
+  - The Doctor never interacts with Alicia or the Heart Monitor: he only needs the encrypted data and some policy metadata.
+
+### How to run the demo
```
derekpierre: We should add a TODO to update the documentation for running the demo. I can do that if you like.

examples/heartbeat_rest_ui/Pipfile
```diff
@@ -0,0 +1,23 @@
...
+[packages]
+dash = "==0.34.0"
+dash-html-components = "==0.13.4"
+dash-core-components = "==0.41.0"
+dash-table = "==3.1.11"
+pandas = "==0.23.4"
+plotly = "==3.5.0"
+requests = "*"
+maya = "*"
+msgpack = "*"
```
derekpierre: Depending on how we do the serialization, we may not need `msgpack` anymore.

examples/heartbeat_rest_ui/bob.py
```diff
@@ -0,0 +1,259 @@
...
+        df = pd.read_sql_query(f'SELECT Timestamp, EncryptedData '
+                               f'FROM {DB_NAME} '
+                               f'WHERE Timestamp > "{last_timestamp}" '
+                               f'ORDER BY Timestamp;',
+                               db_conn)
+
+        for index, row in df.iterrows():
+            message_kit_b64 = row['EncryptedData']
+
+            assert UmbralMessageKit.from_bytes(b64decode(message_kit_b64))
```
derekpierre: I guess we should remove these Umbral asserts before submitting.

KPrasch:
Related: <https://circleci.com/docs/2.0/browser-testing/>

.circleci/config.yml
```diff
@@ -12,6 +12,14 @@ workflows:
           filters:
             tags:
               only: /.*/
+      - pipenv_install_36_browser:
```
KPrasch: I think these can be omitted - Unless you are just trying to confirm that the deps can be installed with this version of docker?

.circleci/config.yml
```diff
@@ -142,6 +150,14 @@ workflows:
               only: /.*/
           requires:
           - cli
+      - web_examples:
+          context: "NuCypher Tests"
+          filters:
+            tags:
+              only: /.*/
+          requires:
```
KPrasch: I also think this is far too early in the workflow.  Let's move this closer to the other demo & example tests.

.circleci/config.yml
derekpierre: Ah ok. so I can remove `pipenv_install_36_browser` and `pip_install_36_browser` jobs.

I guess essentially they duplicate the `pipenv_install_36` and `pip_install_36` jobs anyway.

.circleci/config.yml
```diff
@@ -156,8 +148,7 @@ workflows:
             tags:
               only: /.*/
           requires:
-            - pip_install_36_browser
-            - pipenv_install_36_browser
+            - cli
```
KPrasch: :+1:

derekpierre: I just noticed that `test_build` requires the other demos, should I make it require `web_examples` as well?

KPrasch: Depends if you want `test_build` to require `web_examples` to pass before continuing.  Doing so will prevent publication from continuing to PyPI on failure, for example.

I think it's a good idea to require these to pass as a form of acceptance testing, but I defer to your judgement on this, one.

derekpierre: Agreed. I think this would be a good thing since we are essentially using the code similar to how an app would.

KPrasch:
Excellent - looks like you've got this narrowed down to dependencies.

KPrasch:
Rebase over master was too difficult to manage, so I went with a merge commit instead.

derekpierre:
The problem seems to be with an outdated pipfile.lock file that doesn't include some since `pipenv sync` is used. I'll look into getting the pipfile.lock updated.

derekpierre:
@KPrasch CI finally passed! 🎉

derekpierre:
@KPrasch I'm finished from my side now.

One question: What do we want to do about the local webdriver executable, for others who may want to run the tests locally on their machines? We previously chatted about setting up a script for setting up the webdriver...

derekpierre:
Some options: https://pytest-selenium.readthedocs.io/en/latest/user_guide.html#specifying-a-browser

KPrasch:
Lots of documented options - what do you think about using PhantomJS?

derekpierre:
I'm not tied to any particular browser for the tests.
Any worry here about PhantomJS maintenance? https://github.com/ariya/phantomjs/issues/15344

The fundamental problem is how to specify the path to the relevant webdriver. I looked at the possibility of doing something similar to install_solc.sh - but we have to contend with determining os version etc. I was looking at https://pypi.org/project/instapy-chromedriver/#description to try to do it via code, but haven't made any inroads just yet.

Still trying to figure out a reasonable/simple option.

KPrasch:
Wow - I did not realize PhantomJS is EOL. My main angle is compatibility - perhaps I'm a bit out-of-date here but Instead of trying to determines which specific binary to download per platform, I'd like to find a solution that uses no browser at all. (PhantomJS or otherwise) this is especially compelling on our CI server, where the default container does not have any browser anyways.  Also - sometimes I prefer running the tests without the browser appearing on-screen.

derekpierre:
I noticed the following in selenium code:
```python
        warnings.warn('Selenium support for PhantomJS has been deprecated, please use headless '
                      'versions of Chrome or Firefox instead')
```

so I stuck with chrome for now.

derekpierre:
@KPrasch ready to take this PR off of "draft" status? 🎉

KPrasch:
OK - It's off "WIP".  Let's be sure this is as easy to review as it can be: A PR Description, any follow-up Issues, last look at the docs. :-)

derekpierre:
> Instead of instructions in README files in each directory, should we put them in the docs?

I worry that people don't really go to the docs, but just go directly to github. If we merge the cli/ui/rest demos into one overarching heartbeat_demo folder, we can just have one README in there with the general overview and then just individual instructions in the folders.

cygnusv:
@derekpierre @KPrasch I've peeked at the commit history and I noticed sometimes it's difficult to follow. I guess this is the result of you guys being forced to accommodate changes to master via merges, but I'm wondering if it can be cleaned somehow. As an example, the following five commits in your history have the same message and they do the same core changes in essence (of course, there are a lot of differences):

8666b640dd1ac99587634d92734d2b615f7fbbc6
99fe7ec7d80e3c0728c91f7930af2d10f46898d2
d7992e7f545e009e82bf9d68713189310837a3de
91140be21653b69bb96d6189b00b42bcd0669d8c
50b2871f93d9aa7276c8240a961617772dd4429f

derekpierre:
@cygnusv there will be some duplicate commits since we moved folders from the `nucypher-demos` repos into `nucypher` using https://github.community/t5/How-to-use-Git-and-GitHub/Adding-a-folder-from-one-repo-to-another/td-p/5425. This enabled us to maintain the original commit history. 

It was the first time we used that command  - basically you filter commits for a specific subfolder and then move it into the new repos. We did one folder at a time for simplicity. However, some of the original commits modified multiple folders, and so the same commit can be included in each git filter command that was executed.

KPrasch:
Yeah it's been alot of overhead trying to maintain this PR against master for so long.  @cygnusv - Do you suggest we squash or drop some of those commits?  Also this PR has significantly behind master again, so the overhead continues to grow.

Issue #773, opened 2019-02-16
Document the undocumented Character Control Endpoints


KPrasch:

KPrasch:
@tuxxy - They just needs to be updated to the latest API scheme

KPrasch:
This Issue has grown a bit worse - were missing create policy, join policy is incomplete, and derive policy encrypting key is actually *incorrect*.

KPrasch:
OK - adding another endpoint to this Issue introduced in PR #795 : `revoke`

KPrasch:
Closing in favor of more scoped future Issues now that most endpoints have baseline docs.

Issue #569, opened 2018-11-21
Implement revocation confirmations from Ursula


tuxxy:
When Ursula accepts a `Revocation` from Alice, she should sign the Revocation (w/ signature) and send it back as confirmation that she agreed to revoke the arrangement.

KPrasch:
Duplicate of #504

KPrasch:
Dupe of #504

Issue #504, opened 2018-10-30
Return revocation receipt (confirmation) from Ursula to Alice


tuxxy:
When Ursula receives a `RevocationNotice` from Alice, she verifies that it was properly signed by her, then deletes the `KFrag`.

Ursula should generate a _receipt_ that is a signature of the received notice and the signature _in_ the `RevocationNotice`, then return it in the response body with a 200 status code.

cygnusv:
Off-chain revocation is out of the scope for the moment

PR #795, opened 2019-02-20
Add a revoke endpoint

tuxxy:


nucypher/characters/lawful.py
```diff
@@ -318,8 +316,17 @@ def grant():
 
         @alice_control.route('/revoke/<policy_encrypting_key>', methods=['DELETE'])
         def revoke(policy_encrypting_key):
+            policy = drone_alice.active_policies[policy_encrypting_key]
```
cygnusv: I've just realized that `active_policies` are not persisted anywhere, right? So if Alice tries to revoke at a latter time, this won't work.

tuxxy: It's what we've meant to do for the time being, actually. Enabling persistent policies will encompass a larger PR.

nucypher/policy/models.py
```diff
@@ -255,6 +255,7 @@ def enact(self, network_middleware, publish=True) -> dict:
         else:  # ...After *all* the policies are enacted
             # Create Alice's revocation kit
             self.revocation_kit = RevocationKit(self, self.alice.stamp)
+            self.alice.active_policies[bytes(self.public_key).hex()] = self
```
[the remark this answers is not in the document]
tuxxy: Mostly because we're accessing the policies from the character control API.

Obviously, this storage is problematic mostly because one can immediately imagine having multiple policies per pubkey. I made a point in your other comment that this will all, hopefully, be answered in another PR that builds a much better solution.

nucypher/characters/lawful.py
```diff
@@ -312,6 +314,25 @@ def grant():
 
             return Response(json.dumps(response_data), status=200)
 
+        @alice_control.route('/revoke/<policy_encrypting_key>', methods=['DELETE'])
+        def revoke(policy_encrypting_key):
+            policy = drone_alice.active_policies[policy_encrypting_key]
+
+            failed_revocations = drone_alice.revoke(policy)
```
michwill: What happens if we want to retry later using the `failed_revocations` from `result`? Can we do it?

tuxxy: As of right now, we only return the number of failed revocations just to inform the user. They can retry the revoke at a later point.

It just occurred to me though that it's pointless to return the number of failed revocations if the policy was successfully revoked, though.

michwill: Oh let say this. We revoke a policy, and only half of Ursulas revoke (fail). Then we revoke, and another half revokes. Will it not fail for trying to revoke what was already revoked?
I think, it's totally ok to return `0` for no failed revocations

tuxxy: Yeah, that's something I figured out today. Without persistent policies it makes it kinda difficult.

michwill: At the very least, we can make it so that it doesn't fail if it was already revoked... Or if it was "not found". Does Ursula return different errors for "not found" and "something else"?

tuxxy: Yeah, that's the solution I came up with as well.

If you notice, we return the error response. If a policy isn't found, it's a specific error code, so it requires some introspection. Nonetheless, it's still a failed revocation -- especially if an Alice was  not expecting it.

tuxxy: Pushed up a commit that makes these changes.

nucypher/policy/models.py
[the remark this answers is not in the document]
cygnusv: Fair enough! Do you have this controlled or would it be good to file an issue?

nucypher/characters/lawful.py
cygnusv: I see, that's fine. Same as above: does this deserve an issue?

tuxxy: Probably does deserve an issue, yeah. We're all well aware of it and it would be nice to talk about it in an issue.

nucypher/characters/control/specifications.py
```diff
@@ -47,9 +47,10 @@ class AliceSpecification(CharacterSpecification):
     __grant = (('bob_encrypting_key', 'bob_verifying_key', 'm', 'n', 'label', 'expiration'),  # In
                ('treasure_map', 'policy_encrypting_key', 'alice_verifying_key'))              # Out
 
-    # TODO: Implement Revoke Spec
-    __revoke = ((),  # In
-                ())  # Out
+    # A revoke specificatio isn't required because we don't use a serializer
+    # with it anyway.
+    __revoke = (('policy_encrypting_key'),  # In
```
KPrasch:
``` suggestion
    __revoke = (('policy_encrypting_key', ),  # In
```

nucypher/characters/control/specifications.py
```diff
@@ -47,9 +47,10 @@ class AliceSpecification(CharacterSpecification):
...
+                ('failed_revocations'))     # Out
```
KPrasch:
``` suggestion
                ('failed_revocations',))     # Out
```

nucypher/characters/control/specifications.py
KPrasch: These will not be tuples without the commas, just a `str`!

tuxxy: Ah, good catch.

KPrasch:
@tuxxy - Nice to see some progress on this control interface :+1: 

What about this `TODO`: <https://github.com/nucypher/nucypher/blob/master/nucypher/cli/characters/alice.py#L189>

tuxxy:
Ah, whoops. Didn't catch that addition in the CLI. Thanks! :D

KPrasch:
Some additional related Issues to consider:

https://github.com/nucypher/nucypher/issues/504, #569

tuxxy:
Yeah, definitely related but not for this PR, I'm afraid. In all honesty, those issues could probably be closed as I doubt the helpfulness of a receipt for revocation anyway.

KPrasch:
Can you include a CLI integration test here: <https://github.com/nucypher/nucypher/blob/master/tests/cli/test_alice.py> ?

KPrasch:
Also: Can you update the docs here: <https://github.com/nucypher/nucypher/blob/master/docs/source/guides/character_control_guide.rst#id6> ?

nucypher/cli/characters/alice.py
```diff
@@ -186,7 +188,11 @@ def alice(click_config,
         return ALICE.controller.grant(request=grant_request)
 
     elif action == "revoke":
-        raise NotImplementedError  # TODO: Implement revoke entry point
+        revoke_request = {
+            'policy_encrypting_key': policy_encrypting_key
+        }
+
+        return ALICE.controller.revoke(request=revoke_request)
```
KPrasch: Actually, isn't this supposed to a URL parameter?

nucypher/characters/control/interfaces.py
```diff
@@ -99,7 +100,21 @@ def grant(self,
...
+    def revoke(self, policy_encrypting_key):
+        policy = self.character.active_policies[policy_encrypting_key]
+
+        failed_revocations = self.character.revoke(policy)
+        if len(failed_revocations) > 0:
+            for node_id, attempt in failed_revocations.items():
+                revocation, fail_reason = attempt
+                if fail_reason == NotFound:
+                    del(failed_revocations[node_id])
+        elif len(failed_revocations) <= (policy.n - policy.treasure_map.m + 1):
```
michwill: Wouldn't this condition always be satisfied though?
E.g. the first `if` makes it so that we get here only if `len(failed_revocations) == 0`, which is (of course!) always `<= n - m + 1`.

tests/characters/test_character_control.py
```diff
@@ -83,6 +83,32 @@ def test_alice_character_control_grant(alice_control_test_client, federated_bob)
...
+    response_data = json.loads(response.data)
+    policy_pubkey = response_data['result']['policy_encrypting_key']
+
+    response = alice_control_test_client.delete(f'/revoke/{policy_pubkey}')
+    assert response.status_code == 200
+
+    response_data = json.loads(response.data)
+    assert 'result' in response_data
+    assert 'failed_revocations' in response_data['result']
+    assert response_data['result']['failed_revocations'] == 0
```
michwill: Would be great to have tests with having failed revocations also (because there could be bugs, as mentioned in a comment above). Not necessarily in this PR, but...

nucypher/characters/control/interfaces.py
cygnusv: Perhaps @tuxxy didn't mean an `elif` but an `if`?

```suggestion
        if len(failed_revocations) <= (policy.n - policy.treasure_map.m + 1):
```
So the `if` in L109 is first cleaning the list of failed revocations, and then this line comes after that once the list only contains revocations with a valid reason? @tuxxy

tests/characters/test_character_control.py
KPrasch: Opened #879

nucypher/cli/characters/alice.py
KPrasch: Related to #875 and #833

nucypher/characters/control/interfaces.py
tuxxy: @cygnusv Yeah, this is a good catch.

KPrasch:
References <https://github.com/nucypher/nucypher/issues/773#issuecomment-476725670>

KPrasch:
@tuxxy  - Also still wondering about my above comment regarding the unused `response` here: <https://github.com/nucypher/nucypher/pull/795/files#diff-a1320afcd2301567f40adfe049e2616cR270>

tuxxy:
@KPrasch Yeah, the response isn't used. The only problematic responses are captured in the exceptions below.

KPrasch:
My sense is that this is also related to #504, #569 - Which still seems like a good Idea to me. I'd really like to see this response do something more useful.  Perhaps in a later PR, if you prefer, but if so, let's leave a TODO marker with the Issue number so it does not look unintentional.

tuxxy:
It's not really related to receipts. Only in the manner that you could use it to validate it, but in my current opinion, receipts are basically useless.

This is basically just an artifact of debugging, but I've always thought capturing responses for any HTTP request, regardless if you use it, makes code more readable to be clear about what is happening.

KPrasch:
Yeah - I need to think about the usefulness of receipts more. (move to the Issue ticket?)  In your thinking - how does Alice know if revocation was successful?

tuxxy:
Yeah, this is more related to the ticket. I'd prefer to discuss on Discord though. GitHub comments aren't really a useful communication device at this level.

KPrasch:
see https://github.com/nucypher/nucypher/issues/875

PR #821, opened 2019-03-09
IPC-RPC Character Control

KPrasch:
### Overview

Implements character control an inter-process communication protocol around JSON RPC 2.0 transported over standard I/O. Uses the twisted reactor to listen for and respond to RPC messages using a `JSONRPCLineReceiver` over a `stdio.StandardIO` interface. 

### Motivation

This is especially helpful for controlling a character from another process using a message passing approach.  This can be used as the message passing mechanism for building a NuCypher web browser extension.

### API Summary

Following the controller-emitter abstraction, this PR introduces `JSONRPCController` and `JSONRPCStdoutEmitter`.  A new IPC-RPC controll process can be started by calling `JSONRPCController.make_control_transport`.  Alternately a convenience method exists on characters for instantiating a new RPC controller: `Character.make_rpc_controller`

```python
def make_rpc_controller(drone, crash_on_error: bool = False):
    app_name = bytes(drone.stamp).hex()[:6]
    controller = JSONRPCController(app_name=app_name,
                                   character_controller=drone.controller,
                                   crash_on_error=crash_on_error)
```

A dedicated CLI entry point is used to launch a character with IPC enabled (`--json-rpc` flag as an argument to `nucypher`) - for example:

```
nucypher --json-rpc alice run
```

###### References

* JSON RPC 2.0 Spec: <https://www.jsonrpc.org/specification>
* Native Messaging MDN Docs: <https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Native_messaging>


nucypher/characters/control/controllers.py
```diff
@@ -164,42 +167,225 @@ def __init__(self,
...
+            valid_method_name = True
+        else:
+            valid_method_name = False
+
+        is_valid = all((valid_method_name, ))
+
+        if not is_valid:
+            raise self.emitter.InvalidRequest
+
+        return is_valid
```
cygnusv: I think lines 279 to 291 can be changed to something simpler like:
```python
        try:
            int(method_name)  # must not be a number
        except ValueError:
            return True
        else:
            raise self.emitter.InvalidRequest
```

nucypher/characters/control/controllers.py
```diff
@@ -164,42 +167,225 @@ def __init__(self,
...
+
+        try:
+            input_fields = set(request.keys())
+        except AttributeError:
+            raise self.emitter.InvalidRequest
+
+        contains_required_fields = required_fileds.issubset(input_fields)
+
+        uniquie_fields = all_fields - input_fields - optional_fields
+        contains_valid_fields = not bool(uniquie_fields)
```
cygnusv: I think this logic is unnecessarily complex and I don't see immediately what it's trying to achieve. 

Since `required_fileds` and `optional_fields` are disjoint sets (otherwise their names would be inconsistent) and `all_fields = required_fileds | optional_fields`, then the expression `all_fields - input_fields - optional_fields` is always equal to `required_fields - input_fields`.

If `contains_required_fields` is True (i.e., `required_fileds` is a subset of `input_fields`), then the latter expression is always the empty set, so `contains_valid_fields` is True too; on the contrary, if `contains_required_fields` is False, then the expression is not empty and `contains_valid_fields` is False. 

Hence, it's always the case that `contains_required_fields == contains_valid_fields`, so we can get rid of the `contains_valid_fields` check.

Also: `uniquie` --> `unique`

nucypher/characters/control/emitters.py
```diff
@@ -73,48 +79,144 @@ def _emit(self,
...
+
+        if as_bytes:
+            serialized_response = bytes(serialized_response, encoding='utf-8')  # type: bytes
 
-            # Serialize IPC Message
-            serialized_response = IPCStdoutEmitter.transport_serializer(response)
+        # Add delimiter
+        if delimiter:
+            if as_bytes:
+                delimiter = bytes(delimiter, encoding='utf=8')
```
cygnusv:
```suggestion
                delimiter = bytes(delimiter, encoding='utf-8')
```

nucypher/cli/characters/alice.py
```diff
@@ -129,14 +129,24 @@ def alice(click_config,
...
+        if click_config.json_ipc:
+            rpc_controller = ALICE.make_rpc_controller()
+            _transport = rpc_controller.make_control_transport()
+            rpc_controller.start()
+            return
+
+        # HTTP
+        else:
+            ALICE.controller.emitter(message=f"Alice Verifying Key {bytes(ALICE.stamp).hex()}", color="green", bold=True)
+            controller = ALICE.make_web_controller(crash_on_error=click_config.debug)
```
vepkenez: This seems like we are further overloading the "debug" argument.  Maybe that's fine?

nucypher/characters/control/emitters.py
KPrasch: :heavy_check_mark:

nucypher/characters/control/controllers.py
```diff
@@ -164,42 +162,223 @@ def __init__(self,
...
+        # Call the internal interface | pipe to output
+        return interface(request=method_params, request_id=request_id, *args, **kwargs)  # < ------- INLET
+
+    def validate_request(self, request: dict):
+
+        #
+        # Phase 1 - Metadata
+        #
+
+        required_fileds = {'jsonrpc', 'method'}
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/characters/control/controllers.py
KPrasch: Here, I am detecting the possibility of an unknown field being present in `all_fields`.

`unique_fields = all_fields - input_fields - optional_fields` - This leaves me with a set containing only the unknown fields.

nucypher/cli/characters/alice.py
KPrasch: Yeah - Let's ensure `--debug` does not become "overloaded".  At the moment It:

- shows logs in the console
- allows otherwise swallowed exceptions to be raised and crash the character control server

KPrasch: See related #941

nucypher/characters/control/controllers.py
```diff
@@ -173,41 +176,225 @@ def __init__(self,
...
+            raise self.emitter.InvalidRequest
+
+        #
+        # Phase 2 - Content Type
+        #
+
+        method_name = request['method']
+
+        try:
+            int(method_name)  # must not be a number
```
michwill: What if it's `'1.1'`? `'1blah'`?
In Python 3, strings seem to have a useful method `isidentifier()`. Is it what we need here?
```
>>> 'x'.isidentifier()
True
>>> 'x123'.isidentifier()
True
>>> '1'.isidentifier()
False
>>> '1.1'.isidentifier()
False
>>> '1blah'.isidentifier()
False
```

nucypher/characters/control/controllers.py
```diff
@@ -173,41 +176,225 @@ def __init__(self,
...
+        except ValueError:
+            valid_method_name = True
+        else:
+            valid_method_name = False
+
+        is_valid = all((valid_method_name, ))
```
michwill: Why not simply `if not valid_method_name: raise ...`?

tests/characters/control/test_web_control.py
```diff
@@ -93,15 +77,15 @@ def test_alice_character_control_revoke(alice_control_test_client, federated_bob
         'n': 3,
         'expiration': (maya.now() + datetime.timedelta(days=3)).iso8601(),
```
michwill: Not at all IPC-specific (many other places also), but I think we may want to test if the API works correctly when we happen to have `0`-th microsecond. One can do that as `(maya.MayaDT(maya.now().epoch) + datetime.timedelta(days=3)).iso8601()`. The result would be something like `2019-06-20T03:07:16Z` (which we'll happen to have one time in million!). Would the tests still pass?

nucypher/characters/control/controllers.py
KPrasch: This acts as a placeholder for additional validation rules in the near-future.

nucypher/characters/control/controllers.py
KPrasch: Yeah - this bit is a bit awkward. : Hmm... `.isidentifier`... thinking:

PR #826, opened 2019-03-13
Docker dev

vepkenez:
I explained everything in the README:
https://github.com/vepkenez/nucypher/blob/cdeda759f3b5a90ebf4de97c24eadd278bb26dd4/dev/docker/README.md

I could use some help adding some more example scenarios


jMyles:
:boom: 

Awesome man; I really look forward to taking a look.  Stand by.

README.md
```diff
@@ -12,18 +12,18 @@
...
 ----
 
-The NuCypher network uses the [Umbral](https://github.com/nucypher/pyUmbral) 
-threshold proxy re-encryption scheme to provide cryptographic access control 
-for distributed apps and protocols. 
-Applications can use the NuCypher network to facilitate end-to-end encrypted 
-data sharing via sharing policies. Access permissions are baked into the 
-underlying encryption, and access can only be explicitly granted by the data owner. 
-Consequently, the data owner has ultimate control over access to their data. 
-At no point is the data decrypted nor can the underlying private keys be
```
KPrasch: Did you use an automated to to remove extra spaces? I'd like to use it too!

dev/docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+ADD . /code
+
+# install reqs and solc
+RUN pip install --upgrade pip \
+    && pip install pipenv \
+    && pip3 install -r dev-requirements.txt --src /usr/local/src \
+    && ./scripts/installation/install_solc.sh
+
+# finish install with some local volume ops
+CMD ["./dev/docker/scripts/install/entrypoint.sh"]
```
KPrasch: Looks like GH is complaining about several missing newlines at EOF.

dev/docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+EXPOSE 11500
+
+RUN mkdir /code
+WORKDIR /code
+ADD . /code
+
+# install reqs and solc
+RUN pip install --upgrade pip \
+    && pip install pipenv \
+    && pip3 install -r dev-requirements.txt --src /usr/local/src \
```
KPrasch: Are development requirements required to run this Dockerfile?  Which dev dependencies are being used?

dev/docker/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+
+EXPOSE 11500
+
+RUN mkdir /code
+WORKDIR /code
+ADD . /code
+
+# install reqs and solc
+RUN pip install --upgrade pip \
+    && pip install pipenv \
```
KPrasch: Perhaps skip pipenv all together, and use *only* pip.

dev/docker/README.md
```diff
@@ -0,0 +1,34 @@
+### Developing with Docker
+
+The intention of the Docker configurations in this directory is to enable anyone to develop and test NuCypher on all major operating systems with minimal prerequisites and installation hassle.
```
KPrasch: @vepkenez - What do you think about moving this work into either `deploy/docker` or `examples/something`?

dev/docker/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
...
+  nucypher-dev:
+    volumes:
+      - ../..:/code
+    ports:
+      - 11500
+    build:
+      context: ../..
+      dockerfile: dev/docker/Dockerfile
+    image: dev:nucypher
+    container_name: nucypher-dev
```
KPrasch: another missing newline

dev/docker/docker-compose.yml
```diff
@@ -0,0 +1,13 @@
+version: '3'
+
+services:
+  nucypher-dev:
+    volumes:
+      - ../..:/code
```
KPrasch: Does this line make this docker-compose file `pwd` dependent?

dev/docker/scripts/ursula.ps1
```diff
@@ -0,0 +1 @@
+docker run -it dev:nucypher nucypher ursula run --dev --federated-only
```
KPrasch: What is the best way to view file logging inside this process?

README.md
vepkenez: Yes.  I wasn't sure and was considering turning it off because of those unintended changes but... if you approve :)

dev/docker/Dockerfile
vepkenez: Yes because the intention is to be able to _do_ development with this setup.  Running the test suite etc.  

I am not sure which are being used.  You would know more!

dev/docker/Dockerfile
vepkenez: I had that in there because the install_solc script uses it.  I could see if it works without it.

dev/docker/README.md
vepkenez: Yeah... I am open to anything and was waiting for some suggestions like this.  The thing is, it is totally distinct from *deploy* right now.  Like... this is for development.  The /docker inside deploy is different and will be developed differently.

dev/docker/docker-compose.yml
vepkenez: nope

dev/docker/scripts/ursula.ps1
vepkenez: this basically will open an ursula "console" just as if you ran that `nucypher nucypher ursula run --dev --federated-only` command in a regular shell.  You would know the answer to that better than I would and I'd love your help setting all this stuff up to work optimally.

dev/docker/Dockerfile
KPrasch: It depends how you envision this fleet being employed.  Most of the `dev-requirements` are aimed at running tests.

dev/docker/Dockerfile
KPrasch: Ah - gotcha. That's a cheap pipenv trick anyways.  You may be able to call the install script directly.

dev/docker/README.md
KPrasch: This part is still unclear to me - it seems the only distinction is only slight, right? (which network ports and domains are being used, for example).  When developing dApps, I mostly imagine the running development fleet as the background to the tests, nearly indistinguishable from production Ursulas.

dev/docker/Dockerfile
KPrasch: Point being, unless you aim to run tests *inside* the containers, then they dont necessarily need to have dev-dependencies installed; and you can still run tests that *connect* to this docker fleet fleet (anyone correct me if I'm wrong - I'm still trying to think of a solid advantage I'm missing out on).

I also dont know that there is a big *disadvantage* to using dev-dependencies here - However if we can share some common dev and deploy docker works, that seems efficient. :-)

KPrasch: Also - these containers use a local volume binding - so I can update the application code running inside the containers?

dev/docker/scripts/ursula.ps1
KPrasch: OK - got it - I think I was confused as to where this is running.

dev/docker/Dockerfile
vepkenez: yes exactly.  when you change the local code, the containers will be running that code.  Figuring that out was some of the hard part of this!

dev/docker/README.md
vepkenez: The way it is installed is totally different.  This is all about `-e`.  The regular docker deploy actually pulls our docker image just does some basic setup.

dev/docker/3-ursulas.yml
```diff
@@ -0,0 +1,61 @@
+version: '3'
```
KPrasch: So - more specifically, this script can possibly use a non-dev / standard requirements version of the codebase.

vepkenez: yes exactly, or like we were talking about using postgres and we could have a version that pre-installs postgres and uses that.

dev/docker/README.md
```diff
@@ -0,0 +1,34 @@
...
+* install [Docker Compose](https://docs.docker.com/compose/install/)
+* cd to dev/docker (where this README is located)
+* `docker-compose up` **this must be done once to complete install**
+
+
+Then you can do things like:
+* run the tests:
+`docker run -it dev:nucypher pytest`
+* start up an ursula:
+`docker run -it -p 10151:10151 dev:nucypher nucypher ursula run --dev --federated-only"`
```
mswilkison:
```suggestion
`docker run -it -p 10151:10151 dev:nucypher nucypher ursula run --dev --federated-only`
```

KPrasch:
I had a smooth install experience :+1: Nonetheless - I am able to run the tests via pytest and the local volume binding works nicely.

however, I cannot run the tests more than once (the first version is run each time), due to what seems to be pytest caching?  I've dome poking around related GH Issues, but I have had no success (https://github.com/afourmy/flask-gentelella/issues/3, https://github.com/pytest-dev/pytest/issues/3557, for example).

Here's the full output of my session: <https://gist.github.com/KPrasch/74097e28ac3f7a34ce0d23f9ec1b6817>

vepkenez:
> I had a smooth install experience 👍 I am able to run the tests via pytest and the local volume binding works nicely.
> 
> However, I cannot run the tests more than once (the first version is run each time), due to what seems to be pytest caching? I've dome poking around related GH Issues, but I have had no success ([afourmy/flask-gentelella#3](https://github.com/afourmy/flask-gentelella/issues/3), [pytest-dev/pytest#3557](https://github.com/pytest-dev/pytest/issues/3557), for example).
> 
> Here's the full output of my session: https://gist.github.com/KPrasch/74097e28ac3f7a34ce0d23f9ec1b6817

Ok @KPrasch this was a real issue that I didn't catch and it required a significant reworking once I figured out what the problem was, but it's all good now.
Thanks!

KPrasch:
Looks a new failed test - any Idea what caused it?

vepkenez:
That test failing makes no sense to me.  I didn’t actually change anything in this PR.  I was going to try running it again.

KPrasch:
Rebuilding: <https://circleci.com/workflow-run/573cd1f5-78a3-4123-a5ec-5934aa9e89b7>

KPrasch:
..and it passed! :thinking:

vepkenez:
I would still like to just have a quick (maybe real-time/hangout) brainstorm about this PR, about where this stuff lives, it’s role in the documentation, different ways it can be used etc.

dev/docker/README.md
```diff
@@ -0,0 +1,38 @@
+### Developing with Docker
```
KPrasch: As part of Issue #901 - I'd like to see this moved into `docs/source/guides/docker-development-guide` or similar.

jMyles:
@vepkenez, I'm working through the docker/pycharm integration.

It's not clear to me - and I don't see any direct answer in the docs - whether to connect to Docker daemon with Unix socket, Docker Machine, or TCP socket.

PR #812, opened 2019-03-06
Felix Distribution Mechanism

KPrasch:
Based on #830  .... sounds like a 30-day workout program.

Also updates to Web3 `5.0.0a8`

Here is the new character class definition :
 <https://github.com/nucypher/nucypher/pull/812/files#diff-9f8c085f8e0727aed7eed6c267d99904R119>

*Note*: Still needs documentation; Draft steps to run Felix:

```
nucypher deploy contracts --provider-uri <URI> --poa (if poa network)
```

```
nucypher felix init --provider-uri <URI> --poa (if poa network) --network <DOMAIN> --checksum-address <ADDR OF WALLET WITH TOKENS TO DISTRIBUTE>
nucypher felix createdb
nucypher felix run
```

`POST` to `localhost:6151/register` | `body` `{'address':  YOUR CHECKSUM ADDRESS}`

*Note* - Running with geth may be blocked by gas management. See automated testing for other examples.

*Update*

Opens new Issues: #870 #880 #881 #882 #883 #884 #885 #886 #887 #888 #889

PR #830, opened 2019-03-14
Base Denominations Management

KPrasch:
Aims to:
- Relive mental strain of handling large numbers (Nu-Wei)
- Increase readability of Stakes and quantities of NuCypher tokens
- Improve divide stake UI

#### Notes
- Introduces: `NU`, `Stake`
- Moves `divide-stake` to `stake --divide`

#### *Update*
-  Fixes: #852 + #857 
- Renamed `M` to `NUNITS_PER_TOKEN`


KPrasch:
Going back on WIP to address RFCs in #812 .

KPrasch:
Instead of re-inventing this awesome code, let's use it: <https://github.com/ethereum/eth-utils/blob/1aa906827162a9e1d6879c27b57b29468e826a72/eth_utils/currency.py>

tests/blockchain/eth/interfaces/test_blockchain_utils.py
```diff
@@ -0,0 +1,118 @@
...
+    with pytest.raises(ValueError):
+        _pi_nuweis = NU('3.14', 'NUWei')
+
+    # A decimal amount of NU, which amounts to NUWei with decimals
+    with pytest.raises(ValueError):
+        _pi_nus = NU('3.14159265358979323846', 'NU')
+
+    # Positive Infinity
+    with pytest.raises(ValueError):
+        _inf = NU(float('infinity'), 'NU')
```
tuxxy: :+1:

tests/blockchain/eth/interfaces/test_blockchain_utils.py
```diff
@@ -0,0 +1,118 @@
...
+
+    # Negative Infinity
+    with pytest.raises(ValueError):
+        _neg_inf = NU(float('-infinity'), 'NU')
+
+    # Not a Number
+    with pytest.raises(InvalidOperation):
+        _nan = NU(float('NaN'), 'NU')
```
tuxxy: All of these are great. :+1:

vepkenez:
So... This seems good.  What's stopping us from merging this?

tuxxy:
Needs more review and consensus, but that's about it, I think. I'd like to see a better name than `NU-wei`. :)

vepkenez:
Oh, good.  I agree about "Nu Wei" but didn't want to be the one to say that :)

tests/blockchain/eth/interfaces/test_blockchain_utils.py
KPrasch: credit to @cygnusv

KPrasch:
Related: https://github.com/nucypher/nucypher/issues/852

nucypher/cli/painting.py
```diff
@@ -209,7 +211,7 @@ def paint_staged_stake(ursula,
 
     click.echo(f"""
 {ursula}
-~ Value      -> {stake_nu} NU ({stake_wei} NU-wei) 
+~ Value      -> {stake_value} ({Decimal(int(stake_value)):.2E} NU-Wei)
 ~ Duration   -> {duration} Days ({duration} Periods)
 ~ Enactment  -> {datetime_at_period(period=start_period)} (period #{start_period})
 ~ Expiration -> {datetime_at_period(period=end_period)} (period #{end_period})
```
cygnusv: The `datetime` produced by `datetime_at_period()` is based on `maya.now()`, which means that it changes with time. Therefore, the datetimes in the `Stake` class and the ones printed here will differ. Perhaps `datatime_at_period` should receive an input `datetime` to avoid this.

nucypher/blockchain/eth/actors.py
```diff
@@ -97,8 +97,9 @@ def eth_balance(self):
     @property
     def token_balance(self):
```
cygnusv:
```suggestion
    def token_balance(self) -> NU:
```

nucypher/blockchain/eth/actors.py
```diff
@@ -484,20 +495,24 @@ def initialize_stake(self,
 
         if expiration:
             lock_periods = calculate_period_duration(future_time=expiration)
+
         if entire_balance is True:
             amount = self.token_balance
 
+        amount = NU(int(amount), 'NUWei')
```
cygnusv: Why the NU -> int -> NU conversion here?

nucypher/blockchain/eth/currency.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def to_nu_wei(self) -> int:
+        """Returns an int value in NU-Wei"""
+        return int(self.__value)
+
+    def __eq__(self, other) -> bool:
+        return int(self) == int(other)
+
+    def __radd__(self, other) -> int:
+        return int(self) + other
```
cygnusv: I think `__radd__` and `__rmul__` should behave exactly as `__add__` and `__mul__` respectively. For example, if we have an `amount` in NU and we compute `2 * amount`, I'd expect the result is still `NU`. What's the motivation for returning `int` here?

nucypher/blockchain/eth/currency.py
```diff
@@ -0,0 +1,197 @@
...
+                 start_period: int,
+                 end_period: int):
+
+        # Stake Info
+        self.owner_address = owner_address
+        self.index = index
+        self.value = value
+        self.start_period = start_period
+        self.end_period = end_period
+        self.duration = self.end_period - self.start_period
```
cygnusv: This implies that if I want to stake for a single period, the duration is 0. We may want to either add 1 here or require that `end_period` is the first period after the end of the stake. Alternatively (and I like this option better), is to get rid of `end_period` as a constructor argument, and change it to `duration`.

nucypher/blockchain/eth/currency.py
```diff
@@ -0,0 +1,197 @@
...
+    @property
+    def id(self) -> str:
+        """TODO: Unique staking ID, currently unused"""
+        digest = b''
+        digest += eth_utils.to_canonical_address(address=self.owner_address)
+        digest += str(self.index).encode()
+        digest += str(self.start_period).encode()
+        digest += str(self.end_period).encode()
+        digest += str(self.value).encode()
+        stake_id = sha256(digest).hex()[:16]
```
cygnusv: This approach for encoding integer values is prone to collisions. Let's focus in the last 4 concatenated items in the digest (i.e., index, start_period, end_period, value). These two stakes (with the same owner_address) have a colliding ID:
* `"1" + "2345" + "6789" + "5555"`
* `"12" + "3456" + "7895" + "555"`
There are many collisions that can be achieved this way. An easy solution is to include a separator character that doesn't occur in any of the variables.

nucypher/blockchain/eth/currency.py
```diff
@@ -0,0 +1,197 @@
...
+        """Returns the number of periods remaining in the stake from now."""
+        current_period = datetime_to_period(datetime=maya.now())
+        return self.end_period - current_period
+
+    def time_remaining(self, slang: bool = False) -> Union[int, str]:
+        """Returns the time delta remaining in the stake from now."""
+        now = maya.now()
+        delta = self.end_datetime - now
+        if slang:
+            result = self.end_datetime.slang_date()
```
cygnusv:
```suggestion
            result = delta.slang_date()
```

nucypher/blockchain/eth/currency.py
```diff
@@ -0,0 +1,197 @@
...
+    __denominations = {'NUWei': 'wei',
+                       'NU': 'ether'}
+
+    def __init__(self, value: Union[int, float, str], denomination: str):
+
+        # Calculate smallest denomination and store it
+        wrapped_denom = self.__denominations[denomination]
+
+        # Validate Early
+        if '.' in str(value):
```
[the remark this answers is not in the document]
cygnusv: I think we can write an issue and move on wrt to this particular problem, BTW.

nucypher/blockchain/eth/actors.py
KPrasch: This is handling two input types: An `int` or  a `NU` instance, depending on weather or not this input was accepted interactively or via the `--value` flag.

nucypher/blockchain/eth/currency.py
KPrasch: Yeah - Thanks for your input on this method.  It is totally unused at the moment, just a drafted Idea.  So - I can just add a delimiter between values to avoid collisions?

cygnusv: Yes, as long as that character cannot occur on any of those vars. First three are plain ints, so any char would work, but `str(self.value)` is `NU`, right? Not so sure, here. Perhaps something like `"//"`, I don't know.

nucypher/blockchain/eth/currency.py
KPrasch: I had a misunderstanding of these dunder methods.  I'll make an update.

nucypher/cli/painting.py
KPrasch: :thinking: - What would the input datetime be?

nucypher/blockchain/eth/currency.py
KPrasch: Adding one seems like a simple solution.   I was compelled to accept end_period as a constructor argument, because an on-chain staking record consists of `(start_period, end_period, stake_value)`

nucypher/blockchain/eth/currency.py
KPrasch: Perhaps there is no need for this feature?  How about using the pipe character `|` ?

nucypher/blockchain/eth/currency.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/constants.py
```diff
@@ -38,8 +38,8 @@
 
 NULL_ADDRESS = '0x' + '0' * 40
 
-__subdigits = 18
-M = 10 ** __subdigits                                  # Unit designation
+SUBDIGITS = 18
+M = 10 ** SUBDIGITS                                    # Unit designation
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/currency.py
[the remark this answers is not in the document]
KPrasch: This boils down to not being able to properly validate a `NU` when instantiated with scientific notation.

For example  `NU(1e+99, 'NU')` will result in:
 `ValueError: Resulting wei value must be between 1 and 2**256 - 1` 

instead of:
 `ValueError("Cannot initialize with fractional wei value")`.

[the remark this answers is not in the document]
KPrasch: This approach seems like a clear improvement:

```
decimal_part = Decimal(value)%1
if not decimal_part.is_zero():
    # The number had decimals
    ...
```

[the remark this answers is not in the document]
KPrasch: https://github.com/nucypher/nucypher/issues/857

nucypher/blockchain/eth/currency.py
```diff
@@ -0,0 +1,197 @@
...
+
+            _, fraction = str(value).split('.')
```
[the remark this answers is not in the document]
KPrasch: https://github.com/nucypher/nucypher/issues/857

nucypher/blockchain/eth/currency.py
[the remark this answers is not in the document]
KPrasch: @cygnusv - :thinking: - Actually, why *wont* this work, since `str(value)` will always add the decimal?  

For example:

```
str(1e+12)
'1000000000000.0'
```

nucypher/cli/painting.py
KPrasch: Another approach here is to read on-chain period values from the `stake.owner.blockchain` and calculate from there.  

Ideally, we'd have an off-chain utility to calculate the current period, If possible (@szotov ), which Is what I've tried to build here.

nucypher/blockchain/eth/currency.py
KPrasch: That will result in an `AttributeError`: <https://circleci.com/gh/nucypher/nucypher/21089?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link>.

I think it is correct the way it is now, see: <https://github.com/kennethreitz/maya>

nucypher/blockchain/eth/currency.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/currency.py
KPrasch: :heavy_check_mark:

nucypher/cli/painting.py
KPrasch: @cygnusv - Since we use floor division with seconds for period calculation, with periods based on epoch, wont they always be the same?

nucypher/blockchain/eth/constants.py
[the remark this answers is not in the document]
cygnusv: Sorry, I was talking about the `M` constant. Although, `TOKEN_DECIMALS` is clearer and more correct than `SUBDIGITS`, so that change is good.

nucypher/blockchain/eth/currency.py
cygnusv: Ok I see. I got confused because this method returns either an interval of time (in seconds, when `slang=False`) or an absolute date (when `slang=True`).

nucypher/blockchain/eth/actors.py
cygnusv: 👍

KPrasch:
@tuxxy - Renamed currency -> token
@cygnusv - I changed the exception handling behavior, such that NU acts like the `int` floor-caster.

nucypher/blockchain/eth/token.py
```diff
@@ -0,0 +1,196 @@
...
+        return currency.from_wei(self.__value, unit='ether')
+
+    def to_nunits(self) -> int:
+        """Returns an int value in NuNit"""
+        return int(self.__value)
+
+    def __eq__(self, other) -> bool:
+        return int(self) == int(other)
+
+    def __radd__(self, other) -> 'NU':
```
michwill: I think, `__radd__` and `__rmul__` are not needed when you have `__add__` and `__mul__`?

KPrasch: I am entertaining the idea of using NU with `sum`, which, I think uses `__radd__`?

nucypher/blockchain/eth/actors.py
```diff
@@ -381,16 +380,25 @@ def locked_tokens(self):
         return self.miner_agent.get_locked_tokens(miner_address=self.checksum_public_address)
 
     @property
-    def total_staked(self) -> int:
+    def total_staked(self) -> NU:
         if self.stakes:
-            return sum(stake[-1] for stake in self.stakes)
-        return 0
+            return NU(sum(int(stake.value) for stake in self.stakes.values()), 'NuNit')
+        return NU(0, 'NuNit')
```
michwill: What's wrong with using `else`?

nucypher/blockchain/eth/actors.py
```diff
@@ -381,16 +380,25 @@ def locked_tokens(self):
...
+            stakes[index] = stake
+        self.__stakes = stakes
 
     @property
-    def stakes(self) -> Tuple[list]:
+    def stakes(self) -> dict:
         """Read all live stake data from the blockchain and return it as a tuple"""
-        stakes_reader = self.miner_agent.get_all_stakes(miner_address=self.checksum_public_address)
-        return tuple(stakes_reader)
+        return self.__stakes
```
michwill: I think, at this point it's not live stake but cached stake

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/actors.py

tests/blockchain/eth/interfaces/test_token_and_stake.py
```diff
@@ -0,0 +1,134 @@
...
+    difference = one_nu - one_nu_wei
+    assert not difference == zero_nu
+
+    actual = float(difference.to_tokens())
+    expected = 0.999999999999999999
+    assert actual == expected
+
+    # 3.14 NU is 3_140_000_000_000_000_000 NuNit
+    pi_nuweis = NU(3.14, 'NU')
+    assert NU('3.14', 'NU') == pi_nuweis.to_nunits() == NU(3_140_000_000_000_000_000, 'NuNit')
```
michwill: Wow, I didn't know that you can do underscores like that

PR #874, opened 2019-03-24
Bob handles failure cases with a bit more tact

### LINKED ISSUE #833: Bob cannot retrieve the same data twice

### LINKED ISSUE #851: NotFound is not being caught and converted into a 404; Bob is also not properly catching it.
During my investigation of #833, I thought it odd that, once a Policy has been revoked, we expect that Bob will encounter `NotEnoughUrsulas`:

https://github.com/nucypher/nucypher/blob/6bdde8d7c46f1dd7470f3d8c67349a004645f701/tests/characters/test_bob_joins_policy_and_retrieves.py#L116

Now... maybe, at a high-level, this is true - we expect that Bob will treat any 404 as insufficient, and ultimately conclude that there are Not Enough Ursulas to complete the retrieval.

The problem is that **this is not at all why this is happening**.  Instead, it "works" for a terrible reason - and it's the same reason that #833 is broken in the first place: that Bob *doesn't generate new work orders once he has retrieved*.  So if he generates zero work orders, he reaches out to zero Ursulas.  And of course zero is not enough!

jMyles:
Bob previously had trouble retrieving twice.  Ursula raised an error instead of 404'ing if a KFrag was `NotFound`.  Both of these are fixed, as well as smoothing out the retrieval flow and adding some test logic.

Fixes #833 
Fixes #851 

Opened #892 

**Note**: Presently, even though Bob *can* retrieve multiple times, he does not pass `cache=True` to `retrieve(...)` in any of the control APIs, so he will still fail in this case.  If we determine that this is the sane default, I'll add it in this PR.  If more discussion is in order, it can be added later or as an option for dapp developers.

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -561,40 +561,53 @@ class Bob(Character):
     def retrieve(self, message_kit, data_source, alice_verifying_key, label, cache=False):
 
         capsule = message_kit.capsule  # TODO: generalize for WorkOrders with more than one capsule
-        capsule.set_correctness_keys(
-            delegating=data_source.policy_pubkey,
-            receiving=self.public_keys(DecryptingPower),
-            verifying=alice_verifying_key)
 
         hrac, map_id = self.construct_hrac_and_map_id(alice_verifying_key, label)
         _unknown_ursulas, _known_ursulas, m = self.follow_treasure_map(map_id=map_id, block=True)
 
-        # TODO: Consider blocking until map is done being followed.
+        already_retrieved = len(message_kit.capsule._attached_cfrags) >= m
 
-        work_orders = self.generate_work_orders(map_id, capsule)
+        if already_retrieved:
+            if cache:
+                must_do_new_retrieval = False
+            else:
+                raise TypeError("Not using cached retrievals, but the MessageKit's capsule has attached CFrags.  Not sure what to do.")
+        else:
+            must_do_new_retrieval = True
 
         cleartexts = []
-        work_orders = work_orders.values()
-        for work_order in work_orders:
-            try:
-                cfrags = self.get_reencrypted_cfrags(work_order)
-            except requests.exceptions.ConnectTimeout:
-                continue
 
-            cfrag = cfrags[0]  # TODO: generalize for WorkOrders with more than one capsule
-            try:
-                message_kit.capsule.attach_cfrag(cfrag)
-                if len(message_kit.capsule._attached_cfrags) >= m:
-                    break
-            except UmbralCorrectnessError:
-                evidence = self.collect_evidence(capsule=capsule,
-                                                 cfrag=cfrag,
-                                                 ursula=work_order.ursula)
+        if must_do_new_retrieval:
+            capsule.set_correctness_keys(
+                delegating=data_source.policy_pubkey,
+                receiving=self.public_keys(DecryptingPower),
+                verifying=alice_verifying_key)
 
-                # TODO: Here's the evidence of Ursula misbehavior. Now what? #500
-                raise self.IncorrectCFragReceived(evidence)
-        else:
-            raise Ursula.NotEnoughUrsulas("Unable to snag m cfrags.")
+            # TODO: Consider blocking until map is done being followed.
+
+            work_orders = self.generate_work_orders(map_id, capsule, cache=cache)
+            work_orders = work_orders.values()  # TODO: A little pedantic, isn't it?
+
+            for work_order in work_orders:
+                try:
+                    cfrags = self.get_reencrypted_cfrags(work_order)
+                except (requests.exceptions.ConnectTimeout, NotFound):
+                    continue
+
+                cfrag = cfrags[0]  # TODO: generalize for WorkOrders with more than one capsule
+                try:
+                    message_kit.capsule.attach_cfrag(cfrag)
+                    if len(message_kit.capsule._attached_cfrags) >= m:
+                        break
+                except UmbralCorrectnessError:
+                    evidence = self.collect_evidence(capsule=capsule,
+                                                     cfrag=cfrag,
+                                                     ursula=work_order.ursula)
+
+                    # TODO: Here's the evidence of Ursula misbehavior. Now what? #500
+                    raise self.IncorrectCFragReceived(evidence)
+            else:
+                raise Ursula.NotEnoughUrsulas("Unable to snag m cfrags.")
 
         delivered_cleartext = self.verify_from(data_source, message_kit, decrypt=True)
         cleartexts.append(delivered_cleartext)
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -527,9 +527,8 @@ class Bob(Character):
 
             capsules_to_include = []
             for capsule in capsules:
-                # if not capsule in self._saved_work_orders[node_id]:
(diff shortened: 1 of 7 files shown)


nucypher/characters/lawful.py
```diff
@@ -548,43 +549,56 @@ def join_policy(self, label, alice_pubkey_sig, node_list=None, block=False):
...
+                        break
+                except UmbralCorrectnessError:
+                    evidence = self.collect_evidence(capsule=capsule,
+                                                     cfrag=cfrag,
+                                                     ursula=work_order.ursula)
+
+                    # TODO: Here's the evidence of Ursula misbehavior. Now what? #500
+                    raise self.IncorrectCFragReceived(evidence)
+            else:
+                raise Ursula.NotEnoughUrsulas("Unable to snag m cfrags.")
```
KPrasch: This has been a popular error message - Can we provide a stronger message and escape path?

nucypher/characters/lawful.py
```diff
@@ -548,43 +549,56 @@ def join_policy(self, label, alice_pubkey_sig, node_list=None, block=False):
...
-        capsule.set_correctness_keys(
-            delegating=data_source.policy_pubkey,
-            receiving=self.public_keys(DecryptingPower),
-            verifying=alice_verifying_key)
 
         hrac, map_id = self.construct_hrac_and_map_id(alice_verifying_key, label)
         _unknown_ursulas, _known_ursulas, m = self.follow_treasure_map(map_id=map_id, block=True)
 
-        # TODO: Consider blocking until map is done being followed.
+        already_retrieved = len(message_kit.capsule._attached_cfrags) >= m
```
KPrasch: Why follow the TMap here - Is it possible to push this check earlier in this method?

tests/fixtures.py
```diff
@@ -270,9 +270,19 @@ def enacted_blockchain_policy(idle_blockchain_policy, blockchain_ursulas):
 
 @pytest.fixture(scope="module")
 def capsule_side_channel(enacted_federated_policy):
-    enrico = Enrico(policy_encrypting_key=enacted_federated_policy.public_key)
-    message_kit, _signature = enrico.encrypt_message(b"Welcome to the flippering.")
-    return message_kit, enrico
+    class _CapsuleSideChannel:
```
KPrasch: :+1:

nucypher/characters/lawful.py
```diff
@@ -527,7 +527,8 @@ def generate_work_orders(self, map_id, *capsules, num_ursulas=None):
                     arrangement_id, capsules_to_include, ursula, self)
                 generated_work_orders[node_id] = work_order
                 # TODO: Fix this. It's always taking the last capsule
-                self._saved_work_orders[node_id][capsule] = work_order
+                if cache:
+                    self._saved_work_orders[node_id][capsule] = work_order
```
KPrasch: Is this a possible unhanded `KeyError`?

tests/characters/test_bob_handles_frags.py
```diff
@@ -147,7 +147,7 @@ def test_bob_can_issue_a_work_order_to_a_specific_ursula(enacted_federated_polic
 
     # We'll test against just a single Ursula - here, we make a WorkOrder for just one.
     # We can pass any number of capsules as args; here we pass just one.
-    capsule = capsule_side_channel[0].capsule
+    capsule = capsule_side_channel()[0].capsule
```
KPrasch: phew!

nucypher/characters/lawful.py
jMyles: Yeah - wanna make an issue for it?  I don't think it belongs in this PR, do you?

nucypher/characters/lawful.py
jMyles: That's interesting.  It of course implies that we do have (or at least have the possibility of having) cached cfrags.  Wanna make an Issue?

nucypher/characters/lawful.py
jMyles: I don't immediately see how - what's your thinking?

tests/characters/test_bob_handles_frags.py
jMyles: Yeah, it's a little ungood there ain't it.  It's a test, though.

KPrasch:
Needs rebase

nucypher/characters/lawful.py
KPrasch: https://github.com/nucypher/nucypher/issues/929

nucypher/characters/lawful.py
KPrasch: Nothing specific, just a general curiosity whenever I see this kind of pattern.  In this case - it looks like there is only setter behavior, so looking good I think.

nucypher/characters/lawful.py
jMyles: :+1:

nucypher/characters/lawful.py
KPrasch: Following up : <https://github.com/nucypher/nucypher/issues/934>

KPrasch:
Are you still thinking that `cache=False` is a good default for control?

jMyles:
@KPrasch: As for `cache=False` being a default - I honest-to-goodness feel like it's a toss-up.  

On one hand, it's more confusing, without a doubt.  I think that it's very reasonable for users to expect the behavior they'll get with `cache=True` (and in fact it's reasonable for them to never even consider that there might be another possibility).

On the other hand, it's possibly more secure for some (albeit very rare) real-world scenarios.  Obviously, in these scenarios, we hope that people RTFM and set `cache=False` themselves.  I think that we'll feel somewhat bad (although I don't think that we'll be at fault in any logically or morally defensible way) if someone ever tries to revoke access for a known compromised device, not realizing that they had stored cached CFrags on it, while continuing to encrypt for a "double-chimney" sym key.

I truly can go either way on it.

nucypher/network/server.py
```diff
@@ -304,10 +304,14 @@ def revoke_arrangement(id_as_hex):
     def reencrypt_via_rest(id_as_hex):
         from nucypher.policy.models import WorkOrder  # Avoid circular import
         arrangement_id = binascii.unhexlify(id_as_hex)
-
-        with ThreadedSession(db_engine) as session:
-            policy_arrangement = datastore.get_policy_arrangement(arrangement_id=id_as_hex.encode(),
-                                                                  session=session)
+        work_order = WorkOrder.from_rest_payload(arrangement_id, request.data)
+        log.info("Work Order from {}, signed {}".format(work_order.bob, work_order.receipt_signature))
```
tuxxy: Nice logging addition.

nucypher/characters/lawful.py
```diff
@@ -556,43 +557,56 @@ def join_policy(self, label, alice_pubkey_sig, node_list=None, block=False):
...
+        if must_do_new_retrieval:
+            # TODO: Consider blocking until map is done being followed.
+
+            work_orders = self.generate_work_orders(map_id, capsule, cache=cache)
+            work_orders = work_orders.values()  # TODO: A little pedantic, isn't it?
+
+            for work_order in work_orders:
+                try:
+                    cfrags = self.get_reencrypted_cfrags(work_order)
+                except (requests.exceptions.ConnectTimeout, NotFound):
```
tuxxy: No logging here?

KPrasch:
Merging *into* epic.

PR #891, opened 2019-03-28
Give responses an application/json content type

drdgvhbh:


tuxxy:
Hi, @drdgvhbh! Is there any context for this PR? Is this due to a bug you experienced? If so, can you provide a traceback or something so we can ensure that this is a good solution to it.

drdgvhbh:
If you do `{{base_uri}}:{{alice_port}}/derive_policy_encrypting_key/{{label}}` or `{{base_uri}}:{{bob_port}}/public_keys` in Postman, the response is in `text/html` format. The response is clearly JSON, so its nicer to have it automatically formatted.

KPrasch:
Looks like this is fixing #891

nucypher/characters/control/interfaces.py
```diff
@@ -45,7 +45,9 @@ def wrapped(instance, request=None, *args, **kwargs) -> bytes:
         response_with_metadata = instance.serializer.build_response_metadata(response=response, duration=duration)
 
         # Emit
-        return instance.emitter(response=response_with_metadata)
+        emitted_response = instance.emitter(response=response_with_metadata)
+        emitted_response.headers["Content-Type"] = "application/json"
```
KPrasch: At the moment, `emitted_response` may be an `int` of the number of bytes written.  Instead, set the headers in the `WebEmitter` class here: <https://github.com/nucypher/nucypher/blob/master/nucypher/characters/control/emitters.py#L140>

drdgvhbh: Fixed. Also note, the response for `{{base_uri}}:{{enrico_port}}/encrypt_message` is still using `text/html` not sure where that comes from

drdgvhbh:
The person whos merging should be able to do that I think

tuxxy:
@drdgvhbh we don't have that enabled on our repos because we don't like to do squash-merges. It's better if you rebase and squash so you can preserve _your_ commit history.

PR #893, opened 2019-03-28
Enable CORS support

drdgvhbh:
CORS is necessary if you want to do any requests non-server side. i.e. from the browser. Pretty useful for the character API at this point. I.e. ur webapp tells the user run an alice node and paste in the URL in the browser.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -20,6 +20,7 @@ requests = "*"
 sqlalchemy = "*"
 maya = "*"
 flask = "*"
+flask-cors = "*"
 flask_sqlalchemy = "*"
 # Third-Party Ethereum
 py-evm = "==0.2.0a39"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
-            "sha256": "4824969533b567e06ed0c8721cda4da63bcdca46a0155cf032ce44c0d8bea8e1"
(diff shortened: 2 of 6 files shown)


tuxxy:
Hey, @drdgvhbh. We haven't enabled CORS for a very important reason -- security.

Once we enable CORS, _any_ website can make a request to the nucypher API, which is something we definitely don't want to happen right now. Exposing the API via an open network endpoint or via CORS is probably a bad idea. If your application requires this functionality, I would suggest iterating and trying a new architecture.

Feel free to jump into Discord and ask questions on change sets like these. We're pretty much always available. Discord is also a great place to get help for architecture questions that may arise when you're building stuff on our stack.

drdgvhbh:
I think I asked on discord but nobody responded. The use case for me is using the character API. If I'm building a webapp, ideally Bob and Alice should be running their respective nodes on their own computer... wouldnt make sense to deploy these on a server since there would only be one of each. I want my web browser to be able to call these local nodes that they're running

jMyles:
@drdgvhbh: Thanks for opening this PR.  It's a big discussion right now.

> What would be the security risks of just opening up the character endpoints?

Well, obviously you don't want just anybody to be able to grant, revoke, retrieve, etc.  So, in your thinking on this, how is this prevented?  If the character endpoints are open and can be accessed from any domain, don't you expect that other domains will attack these endpoints (presumably via malicious javascript)?

PR #899, opened 2019-03-31
Improve management of ETH test accounts. Test coverage power-up.

### LINKED ISSUE #894: TESTING_ETH_AIRDROP_AMOUNT is 0
https://github.com/nucypher/nucypher/blob/3f94b93df37a1f6e5b90d2c624f7c0aea7806198/nucypher/utilities/sandbox/constants.py#L81

```
>>> TESTING_ETH_AIRDROP_AMOUNT
0
```
It should be: `int(Web3().toWei(100, 'ether'))`

Corrected in WIP PR #899

### LINKED ISSUE #895: Improve management of ETH test accounts
I'm having a lot of problems with test account management, probably there are some things that I'm doing wrong, but I also suspect there's something weird happening. 

By default eth_tester creates 10 accounts, and so far we have managed with that. 
however, that's insufficient if we really intend to try 10 test ursulas, which in principle is the default. The truth is that we are not testing that correctly. If one account is used as etherbase, and a couple of them are used to alice, bob, whatever, then the remaining are not enough.
An example of this is the `blockchain_ursulas` fixture:
```
def blockchain_ursulas(three_agents, ursula_decentralized_test_config):
    token_agent, miner_agent, policy_agent = three_agents
    etherbase, alice, bob, *all_yall = token_agent.blockchain.interface.w3.eth.accounts

    assert len(all_yall) >= NUMBER_OF_URSULAS_IN_DEVELOPMENT_NETWORK
    ursula_addresses = all_yall[:NUMBER_OF_URSULAS_IN_DEVELOPMENT_NETWORK]
```
I added that assertion, which fails. In reality, the last line gets less than `NUMBER_OF_URSULAS_IN_DEVELOPMENT_NETWORK` ursulas, so tests are not actually testing the intended number of Ursulas. 

In principle, `TesterBlockchain` can be configured to create more accounts automatically, but if you do that, these accounts behave differently than the default 10 accounts by eth_tester. So, if in a test you are using the 10 test ursulas, the first part of them use the last default eth_tester accounts, and the remaining use newly created accounts, with a different behavior wrt to unlocking. From that moment on, everything seems to break.

We need a clever way to manage ETH test accounts, that considers not only this problem, but also a clear way of knowing which of the `interface.w3.eth.accounts` belong to who. By default we're assuming that etherbase is the first, but after that, it's kind of adhoc.

### LINKED ISSUE #919: Missing coverage reports for CLI
Name says it all... The circleCI yml configuration does not upload coverage results.

cygnusv:
* Fixes a subtle set of errors in our CI config that caused 15% of missing coverage. Just by fixing this, we increase coverage from 69% to 84%. 
* Fixes #895: 
   * ETH test accounts were not used correctly in some tests. 
   * Test accounts are now generated exclusively with `eth_tester`.  
   * Creates named accounts for etherbase, alice, bob and ursulas in `TesterBlockchain` and uses them in tests when possible.
* Includes web3 v5 upgrades (see https://web3py.readthedocs.io/en/latest/v5_migration.html)
* Fixes #894 (constant TESTING_ETH_AIRDROP_AMOUNT is 0)
* Fixes #919 (missive coverage reports of CLI tests)


KPrasch:
Nice! Let's include the `web3>=5.0.0a8` upgrades in the PR too.

KPrasch:
Still reviewing - but I noticed another related Issue that can be addressed in this PR: <https://github.com/nucypher/nucypher/issues/919>

nucypher/blockchain/eth/interfaces.py
```diff
@@ -234,9 +235,10 @@ def add_provider(self,
                     from nucypher.utilities.sandbox.constants import PYEVM_GAS_LIMIT
                     genesis_params = PyEVMBackend._generate_genesis_params(overrides={'gas_limit': PYEVM_GAS_LIMIT})
                     pyevm_backend = PyEVMBackend(genesis_parameters=genesis_params)
+                    pyevm_backend.reset_to_genesis(genesis_params=genesis_params,
```
KPrasch: :+1: - Excellent to see this come around full-circle

KPrasch: Also - As a future optimization, this method might be a way to speed up tests by avoiding airdrops and bypassing contract deployment.

nucypher/blockchain/eth/interfaces.py
```diff
@@ -391,7 +393,7 @@ def call_backend_verify(self, pubkey: PublicKey, signature: Signature, msg_hash:
     def unlock_account(self, address, password, duration):
         if 'tester' in self.provider_uri:
             return True  # Test accounts are unlocked by default.
-        return self.w3.personal.unlockAccount(address, password, duration)
+        return self.w3.geth.personal.unlockAccount(address, password, duration)
```
KPrasch: This deserves a moment of pause.

nucypher/config/keyring.py
```diff
@@ -422,13 +422,6 @@ def _generate_key_filepaths(public_key_dir: str,
 
         return __key_filepaths
 
-    def _export_wallet_to_node(self, blockchain, password):  # TODO: Deprecate with geth.parity signing EIPs
```
KPrasch: Is this TODO addressed in this PR, too?

nucypher/blockchain/eth/interfaces.py
cygnusv: Perhaps something we should describe in an issue or TODO?

nucypher/blockchain/eth/interfaces.py
cygnusv: I see your point, and although I agree in principle, in our codebase and docs we seem to talk only about `geth`. Also, `web3.parity` doesn't seem to support locking accounts; although we're not using that functionality right now, we have a related pending TODO in `BlockchainPower`. A possible way forward is to describe this (i.e., potential support for Parity or other Ethereum clientes) as an issue.

nucypher/config/keyring.py
cygnusv: It's not addressed. The `_export_wallet_to_node` method is unused and the TODO suggests that the method should be deprecated, so I thought I might as well remove it.

cygnusv:
> Still reviewing - but I noticed another related Issue that can be addressed in this PR: #919

This was addressed in this PR, commit cd1eb11d369cb71732add8b1bcf47cf180f3bc10

nucypher/utilities/sandbox/blockchain.py
```diff
@@ -75,20 +79,15 @@ def __init__(self, test_accounts=None, poa=True, airdrop=False, *args, **kwargs)
         self.attach_middleware(w3=self.interface.w3, poa=poa)
 
         # Generate additional ethereum accounts for testing
-        population = test_accounts if test_accounts is not None else NUMBER_OF_URSULAS_IN_DEVELOPMENT_NETWORK
-
+        population = test_accounts
```
KPrasch: :+1:

nucypher/blockchain/eth/interfaces.py
KPrasch: Following up: #933

KPrasch:
:cowboy_hat_face:

Issue #807, opened 2019-02-28
Missing "Contract deployment guide"


cygnusv:
When building the docs, it complains that:
```
/nucypher/docs/source/index.rst:96: WARNING: toctree contains reference to nonexisting document 'guides/contract_deployment_guide'
```

KPrasch:
Closed in #805

Issue #907, opened 2019-04-05
Cannot learn from devnet with domain AttributeError


KPrasch:
```
2019-04-04T17:31:25-0700 [learning-loop#warn] Unhandled error during node learning: Traceback (most recent call last):
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/base.py", line 1272, in run
	    self.mainLoop()
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/base.py", line 1281, in mainLoop
	    self.runUntilCurrent()
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/base.py", line 902, in runUntilCurrent
	    call.func(*call.args, **call.kw)
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/task.py", line 239, in __call__
	    d = defer.maybeDeferred(self.f, *self.a, **self.kw)
	--- <exception caught here> ---
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
	    result = f(*args, **kw)
	  File "/home/kieran/Git/nucypher/nucypher/network/nodes.py", line 534, in keep_learning_about_nodes
	    self.learn_from_teacher_node(eager=False)
	  File "/home/kieran/Git/nucypher/nucypher/network/nodes.py", line 746, in learn_from_teacher_node
	    if not self.learning_domains.intersection(node.serving_domains):
	builtins.AttributeError: 'list' object has no attribute 'intersection'
	

```

KPrasch:
Fixed in #916

Issue #896, opened 2019-03-29
--federated-only not read from configuration file


KPrasch:
`--federated-only` must be passed during `nucypher ursula run`.  When working properly, running `nucypher ursula init --federated-only` will write the value `federated_only: true` to the on-disk character configuration file, which is parsed and used as a default with `run` commands.

This behavior causes a misleading scenario wherein the user believes they are running in federated mode, but are not, resulting in the following error when connecting to a federated fleet:

```
WARNING: Running in Federated mode
...
nucypher.network.nodes.InvalidNode
```

As a temporary work-around: always pass `--federated-only` during run commands, and key an eye on your logging output.

PR #908, opened 2019-04-05
Fix Issues #807 and #896

KPrasch:


KPrasch:
Fix Issues #907 and #896

KPrasch:
Looks like this was only a partial fix, since it is still happening: <https://sentry.io/organizations/nucypher/issues/973841904/?project=1310685&statsPeriod=14d&utc=false>

Issue #909, opened 2019-04-05
Cannot start Ursula with PermissionError


KPrasch:
```
Apr 05 18:30:54 ip-172-31-12-213 nucypher[22959]: Connecting to devnet
Apr 05 18:30:54 ip-172-31-12-213 nucypher[22959]: [Errno 13] Permission denied: '/usr/local/lib/python3.6/dist-packages/hendrix/utils/8000_Flask.pid'
Apr 05 18:30:54 ip-172-31-12-213 nucypher[22959]: PermissionError [Errno 13] Permission denied: '/usr/local/lib/python3.6/dist-packages/hendrix/utils/8000_Flask.pid'
```

Hendrix appears to be making Flask PID files in a location where it does not have read permission.

KPrasch:
Escalating this Issue to a blocker because it is preventing new system deployments to production, as well as effecting new user installs.

KPrasch:
Manually changing the permission of the PID file or directory lead me to another exception, which is being swallowed by a twisted logging exception when handling unhandled exceptions.  (...say that 10x fast)

```
Apr 05 19:51:18 ip-172-31-7-153 nucypher[9504]: Traceback (most recent call last):
Apr 05 19:51:18 ip-172-31-7-153 nucypher[9504]:   File "/usr/local/lib/python3.6/dist-packages/twisted/internet/defer.py", line 654, in _runCallbacks
Apr 05 19:51:18 ip-172-31-7-153 nucypher[9504]:     current.result = callback(current.result, *args, **kw)
Apr 05 19:51:18 ip-172-31-7-153 nucypher[9504]:   File "/home/ubuntu/code/nucypher/network/nodes.py", line 497, in handle_learning_errors
Apr 05 19:51:18 ip-172-31-7-153 nucypher[9504]:     self.log.warn("Unhandled error during node learning: {}".format(failure.getTraceback()))
Apr 05 19:51:18 ip-172-31-7-153 nucypher[9504]:   File "/usr/local/lib/python3.6/dist-packages/twisted/logger/_logger.py", line 238, in warn
Apr 05 19:51:18 ip-172-31-7-153 nucypher[9504]:     self.emit(LogLevel.warn, format, **kwargs)
Apr 05 19:51:18 ip-172-31-7-153 nucypher[9504]:   File "/usr/local/lib/python3.6/dist-packages/twisted/logger/_logger.py", line 144, in emit
[...]

PR #825, opened 2019-03-12
[WIP] Giving codecov a spin.

jMyles:

PR #910, opened 2019-04-06
Update and re-lock dependencies; Reintroduce Codecov

KPrasch:
Includes hendrix v3.2.3 in response to Issue #909 ; Derived from commits in #825, removes coveralls and reintroduces codecov GH integration via circleCI orb. Also opens follow-up Issue: <https://github.com/nucypher/nucypher/issues/911>

Implements codecov orb: <https://circleci.com/blog/making-code-coverage-easy-to-see-with-the-codecov-orb/>
Applies codecov configuration / usage fixes documented here: <https://pytest-cov.readthedocs.io/en/latest/reporting.html>
see codecov results here: <https://codecov.io/gh/nucypher/nucypher>


Pipfile
```diff
@@ -22,7 +22,7 @@ maya = "*"
 flask = "*"
 flask_sqlalchemy = "*"
 # Third-Party Ethereum
-py-evm = "==0.2.0a39"
+py-evm = ">=0.2.0a39"
```
michwill: Does it really work with this newer version? I think, there were issues with `eth-tester`?

KPrasch: I think it remained locked at `0.2.0a39`

michwill: Sure... But here in Pipfile it's updated, and then `--skip-lock` will install 0.2.0a42 or whatever, and eth-tester will fail?
Unless they shipped a newer version

KPrasch: I think with the presence of a `Pipfile.lock` - there is no need to use `--skip-lock` and the best way to install is `pipenv sync`, although I think `pipenv install` will have a similar effect

michwill: Well, at my machine, it tries to relock for some reason, and that happens forever.
Why not keep it ==0.2.0a39 as long as it is not known to work at higher versions?

michwill: (until eth-tester gets updated)

michwill: upd: Pipfile.lock works indeed. But still, needs to be workable even with --skip-lock

KPrasch: Reverted - :heavy_check_mark:

KPrasch:
Also waiting on <https://circleci.com/workflow-run/3caf0197-fcb5-4061-891b-d35b09386736>

PR #921, opened 2019-04-07
Test showing 920.

jMyles:


jMyles:
Since this is just a test, and it tests stuff that obviously needs to go in here, I'm merging.

jMyles:
Is this issue closed?

I think that, pursuant to our discussions on Discord (starting [here](https://ptb.discordapp.com/channels/411401661714792449/411401661714792451/587144270318927887) but continuing only intermittently), it's fair to say that @tuxxy, @michwill, and @cygnusv are of a mindset that's in the ballpark of "this issue is fixed by `62c7df293a03347b16967fe7de7924199c38f819` and similar patterns."

Put another way, these bytes-hungry colleagues of mine are content to understand that anybody who calls `grant` (currently, if we're taking attendance, that's 1) the CLI, 2) the HTTP endpoints, 3) the IPC / unix tooling, our users who awesomely choose to use the Python API directly) is supposed to make a design decision about how to turn unicode characters into bytes.

In the 3 cases over which we have control, we have chosen UTF-8 (although @tuxxy correctly points out that UTF-16 takes an even broader view of lingual inclusion, albeit at a huge memory cost).

My preference, although I don't feel terribly strongly about it, is that `grant` itself is better understood as the conjoining place where unicode is accepted - in other words, that the proper type hint for `label` in `grant(...)` is `str`.

This is chiefly to avoid a scenario wherein the person sharing data does, eg, the following:

`nucypher alice grant ... ❽◯ ▲ ▶ஔ`

...and then shares that label, in unicode, with a would-be recipient, who is using an application developed for the same ecosystem, but which is similarly agnostic regarding encoding as @tuxxy and @michwill and @cygnusv want to be.

Now, they are left guessing - which encoding are they supposed to use?  UTF-8 is of course the correct answer, but how are they to know that?  Well, they have to check our documentation of course.

And if we're *forcing* them to use UTF-8 (ie, we have decided that our own interfaces are gnostic on this matter) then why don't we simply make this a default?

Put another way, if you reeeaaaalllly want to make a policy with a UTF-8 non-compliant label, then you can do that via the `create_policy` Python API.  Otherwise, we assume that you're in a unicode zone.

But, if that's actually the design we want (people to need to guess about what encoding to use in the Python API to have compatibility with the other interfaces, but also the flexibility to use whatever bytes they want) then I guess I can live with that.  I just want to hear everybody acknowledge that that's the price here.

PR #924, opened 2019-04-09
Updates README: Slight touch to first paragraph. Codecov badge

cygnusv:
* Slight touch to the README, since the very first sentence was a reference to Umbral and proxy re-encryption. Instead of saying *how* we do it (i.e., PRE), we should say *what* we do (i.e., E2E access control, etc).
* Switches coverall badge to codecov

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -31,8 +31,19 @@ NuCypher
 
 .. _Umbral: https://github.com/nucypher/pyUmbral
 
-The NuCypher network uses the Umbral_ threshold proxy re-encryption scheme
-to provide cryptographic access controls for distributed apps and protocols.
+The NuCypher network facilitates end-to-end encrypted data sharing
+for distributed apps and protocols.
+Access permissions are baked into the underlying encryption,
+and access can only be explicitly granted by the data owner via sharing policies.
+Consequently, the data owner has ultimate control over access to their data.
+At no point is the data decrypted nor can the underlying private keys be
+determined by the NuCypher network.
+
+Under the hood, the NuCypher network uses the Umbral_
+threshold proxy re-encryption scheme to provide cryptographic access control.
+
+How does NuCypher work?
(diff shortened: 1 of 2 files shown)


docs/source/index.rst
```diff
@@ -31,8 +31,19 @@ NuCypher
 
 .. _Umbral: https://github.com/nucypher/pyUmbral
 
-The NuCypher network uses the Umbral_ threshold proxy re-encryption scheme
-to provide cryptographic access controls for distributed apps and protocols.
+The NuCypher network facilitates end-to-end encrypted data sharing
+for distributed apps and protocols.
+Access permissions are baked into the underlying encryption,
```
arjunhassard: This is a feature of our cryptography we probably don't emphasise enough, but I would find the concept a little confusing if this was the second sentence I'd read about NuCypher. 

How about something like: 

To initiate a NuCypher sharing flow, the data owner chooses recipients and adds their public key to a _sharing policy_. This bakes the corresponding permissions into the encryption of the data payload itself. Hence, from the moment the data leaves the owner's device, it can only ever be accessed by the data owner and their chosen recipients. The network enforces these access rights without ever being trusted with the plaintext version of the data.

michwill: Hm.. Idk. The permissions are not baked into data payload itself, they exist separately as re-encryption keys.

I like how the originally proposed change sounds. Maybe can add something like "... via sharing policies, created using data owner's private key and recepient's public key."

arjunhassard: Right, this is what I find confusing about the sentence – 
`Access permissions are baked into the underlying encryption` 
– because 'encryption' without any other context carries a connotation of doing something directly to the plaintext data. 
To be fair, the underlying data is encrypted such that it pertains to a specific sharing policy and set of re-encryption keys and permissions, so arguably permissions are (indirectly) 'baked in' at this step.

Issue #937, opened 2019-04-17
Alice character control to be able to decrypt


michwill:
Currently, Alice character control can only grant permissions, but cannot decrypt Alice's own data. Meanwhile, Pythonic Alice character is able to do it. We need to implement this method for character control also

robin-thomas:
+1 for this, as I had to write a lot of code to get around this in my application.

PR #939, opened 2019-04-19
[WIP] decryption endpoint for alice 

vepkenez:
resolves https://github.com/nucypher/nucypher/issues/937


nucypher/characters/lawful.py
```diff
@@ -283,6 +283,25 @@ def revoke(self, policy) -> Dict:
                     failed_revocations[node_id] = (revocation, UnexpectedResponse)
         return failed_revocations
 
+    def decrypt_message_kit(self, message_kit, data_source, alice_verifying_key, label):
```
KPrasch: How about just `def decrypt`?

tests/cli/test_cli_lifecycle.py
```diff
@@ -262,8 +298,9 @@ def _bob_retrieves(_grant_result):
 
     # Run the Callbacks
     d = threads.deferToThread(enrico_encrypts)  # scene 4
-    d.addCallback(_run_teacher)                 # scene 5 (preamble)
-    d.addCallback(_grant)                       # scene 5
-    d.addCallback(_bob_retrieves)               # scene 6
+    d.addCallback(_alice_decrypts)              # scene 5 (uncertainty)
+    d.addCallback(_run_teacher)                 # scene 6 (preamble)
+    d.addCallback(_grant)                       # scene 7
+    d.addCallback(_bob_retrieves)               # scene 8
```
KPrasch: :+1:

nucypher/characters/lawful.py
KPrasch: Also let;s include this type hint: `-> List[bytes]`

nucypher/characters/lawful.py
```diff
@@ -283,6 +283,30 @@ def revoke(self, policy) -> Dict:
                     failed_revocations[node_id] = (revocation, UnexpectedResponse)
         return failed_revocations
 
+    def decrypt_message_kit(
+        self,
+        message_kit: UmbralMessageKit,
+        data_source: Character,
+        alice_verifying_key: bytes,
```
cygnusv: Same.

nucypher/characters/control/interfaces.py
```diff
@@ -125,6 +125,38 @@ def revoke(self, label: bytes, bob_verifying_key: bytes):
         response_data = {'failed_revocations': len(failed_revocations)}
         return response_data
 
+    def decrypt(self,
+                label: bytes,
+                policy_encrypting_key: bytes,
+                alice_verifying_key: bytes,
```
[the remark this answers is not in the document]
tuxxy: Good idea.

nucypher/characters/control/interfaces.py
```diff
@@ -125,6 +125,37 @@ def revoke(self, label: bytes, bob_verifying_key: bytes):
         response_data = {'failed_revocations': len(failed_revocations)}
         return response_data
 
+    def decrypt(self,
+                label: bytes,
+                policy_encrypting_key: bytes,
```
cygnusv: Alice can derive herself the `policy_encrypting_key` using `self.character.get_policy_pubkey_from_label(label)`, so you can get rid of this arg.

nucypher/characters/lawful.py
cygnusv: Actually, it seems this argument is never used. You can remove it.

nucypher/cli/characters/alice.py
```diff
@@ -35,6 +35,9 @@
 @click.option('--dev', '-d', help="Enable development mode", is_flag=True)
 @click.option('--force', help="Don't ask for confirmation", is_flag=True)
 @click.option('--dry-run', '-x', help="Execute normally without actually starting the node", is_flag=True)
+@click.option('--policy-encrypting-key', help="Encrypting Public Key for Policy as hexadecimal string", type=click.STRING)
+@click.option('--alice-verifying-key', help="Alice's verifying key as a hexadecimal string", type=click.STRING)
```
cygnusv: Reassess if you need these two lines, knowing that you don't need the policy encrypting key nor alice's verifying to in order to decrypt by Alice.

nucypher/characters/control/interfaces.py
```diff
@@ -125,6 +125,37 @@ def revoke(self, label: bytes, bob_verifying_key: bytes):
...
+        """
+        Character control endpoint to allow Alice to decrypt her own data.
+        """
+
+        from nucypher.characters.lawful import Enrico
+
+        policy_encrypting_key = UmbralPublicKey.from_bytes(
+            policy_encrypting_key)
+
+        alice_pubkey_sig = self.character.get_policy_pubkey_from_label(label)
```
cygnusv: This is incorrect. `alice_pubkey_sig` should be Alice's _signing_ public key, but this line gets the policy's encrypting key. In the end, however, it seems that `alice_pubkey_sig` is not used by `alice.decrypt_message_kit()` (see comment below) so nothing actually fails. I think what you intended was:
```python
policy_encrypting_key = self.character.get_policy_pubkey_from_label(label)
```

nucypher/characters/control/serializers.py
```diff
@@ -85,6 +103,17 @@ def dump_derive_policy_encrypting_key_output(response: dict):
         response_data = {'policy_encrypting_key': policy_encrypting_key_hex, 'label': unicode_label}
         return response_data
 
+    def load_decrypt_input(self, request: dict) -> dict:
+        parsed_input = dict(label=request['label'].encode(),
+                            policy_encrypting_key=bytes.fromhex(request['policy_encrypting_key']),
```
cygnusv: See comment above.

nucypher/characters/control/specifications.py
```diff
@@ -73,14 +72,18 @@ class AliceSpecification(CharacterSpecification):
     __revoke = (('label', 'bob_verifying_key', ),  # In
                 ('failed_revocations',))     # Out
 
+    __decrypt = (('label', 'policy_encrypting_key', 'message_kit'),  # In
```
cygnusv: See comment above.

nucypher/cli/characters/alice.py
```diff
@@ -192,5 +197,20 @@ def alice(click_config,
...
+    elif action == "decrypt":
+
+        if not all((label, policy_encrypting_key, message_kit)):
+            input_specification, output_specification = ALICE.controller.get_specifications(interface_name='decrypt')
+            required_fields = ', '.join(input_specification)
+            raise click.BadArgumentUsage(f'{required_fields} are required flags to decrypt')
+
+        request_data = {
+            'label': label,
+            'policy_encrypting_key': policy_encrypting_key,
```
cygnusv: Same.

nucypher/characters/control/interfaces.py
vepkenez: Ah.  Thanks!  These different keys are still a little fuzzy to me.  Will fix!

nucypher/characters/control/serializers.py
```diff
@@ -16,6 +16,24 @@ class SerializerError(ValueError):
...
+    __message_kit_transport_decoder = b64decode
+
+    def set_message_encoder(self, encoder: Callable):
+        self.__message_kit_transport_encoder = encoder
+
+    def set_message_decoder(self, decoder: Callable):
+        self.__message_kit_transport_decoder = decoder
+
+    def encode(self, plaintext: bytes) -> str:
+        return MessageHandlerMixin.__message_kit_transport_encoder(plaintext).encode()
```
cygnusv: It seems this method is unused. Shouldn't be it used in symmetry with `decode()`?

vepkenez: I actually didn't do anything with that class, I just moved it from lower in the code so I could re-use it in AliceControlJSONSerializer as it was originally only used in BobControlJSONSerializer.

I make no claim to knowing if it is or how it is used.

Issue #931, opened 2019-04-10
Challenge Actor - "Investigator" Bob


KPrasch:
The `MiningAdjudicator` contract introduces the contract method `evaluateCfrag` - following the Agent-Actor abstraction - we need a `Challenger` Actor and `MiningAdjudicatorAgent` to provide a high-level python API for evaluating `capsule`s and `cFrags`.

cygnusv:
@szotov has been calling this actor the "Investigator". I think that in most cases it will be Bob, and there doesn't seem to be much incentive to be someone else, but it's fine to have the Investigator actor in order to continue our Agent-Actor abstraction.

PR #866, opened 2019-03-20
MiningAdjudicatorDeployer, TokenEconomics and Stake Integration

KPrasch:
### Overview

- Introduces deployment API for new contract: `MiningAdjudicator`
- Integrates `TokenEconomics` into contract deployment
- Dissolves `blockchain.constants` into more formal interfaces
- Relocate blockchain staking I/O to `Stake` class
- Carefully considers on-chain staking cache invalidation scenarios
- Improved logging and error handling for failed deployment transactions

###### References

Fixes: #624 #354  #932
Opens: #931 #936 

Original work by @michwill <https://github.com/nucypher/nucypher/pull/325>

Issue #855, opened 2019-03-17
Python Deployers for Slashing Contracts


KPrasch:

KPrasch:
Fixed in #866

Issue #500, opened 2018-10-28
Bob finds out Ursula cheated. Now what?


cygnusv:
Bob retrieves the CFrags, checks them, and one of them is incorrect. He collects all the evidence.
Now what should be the workflow?

michwill:
I think, we should configuration in Bob's client, such as "always slash" or not

cygnusv:
Subsumed by #305

PR #947, opened 2019-04-25
Vodka aftertaste

### LINKED ISSUE #259: Define format for re-encryption metadata
This issue was first raised in https://github.com/nucypher/pyUmbral/issues/120.
This is continuing the discussion of the KMS specific stuff here.

What should this metadata contain? It must be serialized as a bytestring. In what order should it be created?

### LINKED ISSUE #626: Fix tests of MiningAdjudicator and ReEncryptionValidator
_Originally posted by @cygnusv in https://github.com/nucypher/nucypher/pull/509_

### LINKED ISSUE #831: Constant ONE_YEAR_IN_SECONDS is wrong!
For some reason, our constant `ONE_YEAR_IN_SECONDS` is wrong: https://github.com/nucypher/nucypher/blob/bf79bb135b471096b0bba352b8de1979c630e109/nucypher/blockchain/eth/constants.py#L19

This figure is used in several places:
https://github.com/nucypher/nucypher/search?q=31540000&unscoped_q=31540000

The correct number is 31536000:
```python
>>> 60*60*24*365
31536000
```

I'd suggest to use the `SECONDS_PER_PERIOD` constant defined later in the same module:
```python
ONE_YEAR_IN_SECONDS = 365 * SECONDS_PER_PERIOD
```

Curiously, we're not the only ones making this same mistake. There's even [poetry](https://www.powerpoetry.org/poems/31540000-seconds) using the same wrong number of seconds. Weird.

cygnusv:
This PR gets us very close to having real slashing, fully-compatible with Ursulas' work. Mainly, it crystallizes the re-encryption metadata format (#259) into something actually useful for `MiningAdjudicator`. There are, however, some small missing pieces in order to finish the integration of slashing (e.g., #962, #931, #855, #500)

It also completes the testing framework for `MiningAdjudicator` and `ReEncryptionValidator` (Close #626)

Other:
* Improvement and simplification of some fixtures
* Improved API for `Character.from_public_bytes()` (f078026eb8166276626893458595a50bbe10f65e)
* Adds missing commits to #866 that were lost after a rebase. Closes #831.
* References #354

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -844,6 +844,8 @@ class Ursula(Teacher, Character, Miner):
                          timestamp=timestamp,
                          identity_evidence=identity_evidence,
                          substantiate_immediately=is_me and not federated_only,
+                         # FIXME: When is_me and not federated_only, the stamp is substantiated twice
+                         # See line 728 above.
                          )
 
         #
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -390,8 +390,8 @@ class Bob(Character):
         """
         Raised when Bob detects incorrect CFrags returned by some Ursulas
         """
-        def __init__(self, grievances):
-            self.grievances = grievances
+        def __init__(self, evidence: List):
+            self.evidence = evidence
 
     def __init__(self, controller=True, *args, **kwargs) -> None:
         Character.__init__(self, *args, **kwargs)
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -390,12 +390,12 @@ class Bob(Character):
 
     _default_crypto_powerups = [SigningPower, DecryptingPower]
 
-    class IncorrectCFragReceived(Exception):
+    class IncorrectCFragsReceived(Exception):
         """
-        Raised when Bob detects an incorrect CFrag returned by some Ursula
+        Raised when Bob detects incorrect CFrags returned by some Ursulas
         """
-        def __init__(self, evidence):
-            self.evidence = evidence
+        def __init__(self, grievances):
+            self.grievances = grievances
 
     def __init__(self, controller=True, *args, **kwargs) -> None:
         Character.__init__(self, *args, **kwargs)
@@ -609,6 +609,7 @@ class Bob(Character):
         work_orders = self.generate_work_orders(map_id, capsule)
 
         cleartexts = []
+        the_airing_of_grievances = []
         work_orders = work_orders.values()
         for work_order in work_orders:
             try:
@@ -616,28 +617,30 @@ class Bob(Character):
             except requests.exceptions.ConnectTimeout:
                 continue
 
-            cfrag = cfrags[0]  # TODO: generalize for WorkOrders with more than one capsule
+            cfrag = cfrags[0]  # TODO: generalize for WorkOrders with more than one capsule/task
             try:
                 message_kit.capsule.attach_cfrag(cfrag)
                 if len(message_kit.capsule._attached_cfrags) >= m:
                     break
             except UmbralCorrectnessError:
-                evidence = self.collect_evidence(capsule=capsule,
-                                                 cfrag=cfrag,
-                                                 ursula=work_order.ursula)
-
-                # TODO: Here's the evidence of Ursula misbehavior. Now what? #500
-                raise self.IncorrectCFragReceived(evidence)
+                task = work_order.tasks[0]  # TODO: generalize for WorkOrders with more than one capsule/task
+                from nucypher.policy.models import IndisputableEvidence
+                evidence = IndisputableEvidence(task=task, work_order=work_order)
+                # I got a lot of problems with you people ...
+                the_airing_of_grievances.append(evidence)
         else:
             raise Ursula.NotEnoughUrsulas("Unable to snag m cfrags.")
+
+        if the_airing_of_grievances:
+            # ... and now you're gonna hear about it!
+            raise self.IncorrectCFragsReceived(the_airing_of_grievances)
+            # TODO: Find a better strategy for handling incorrect CFrags #500
+            #  - There maybe enough cfrags to still open the capsule
+            #  - This line is unreachable when NotEnoughUrsulas
 
         delivered_cleartext = self.verify_from(data_source, message_kit, decrypt=True)
         cleartexts.append(delivered_cleartext)
         return cleartexts
-
-    def collect_evidence(self, capsule, cfrag, ursula):
-        from nucypher.policy.models import IndisputableEvidence
-        return IndisputableEvidence(capsule, cfrag, ursula)
 
     def make_web_controller(drone_bob, crash_on_error: bool = False):
 
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -20,11 +20,7 @@ from base64 import b64encode
 from collections import OrderedDict
 from functools import partial
 from json.decoder import JSONDecodeError
-from typing import Dict
-from typing import Iterable
-from typing import List
-from typing import Set
-from typing import Tuple
+from typing import Dict, Iterable, List, Set, Tuple
 
 import maya
 import requests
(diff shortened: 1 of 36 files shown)


cygnusv:
@KPrasch I still need to close the loop between the new format for signatures and metadata (#259) and the tests for `MiningAdjudicator`.

nucypher/crypto/utils.py
```diff
@@ -47,39 +50,89 @@ def canonical_address_from_umbral_key(public_key: UmbralPublicKey) -> bytes:
...
+    :param message: Signed message
+    :param signature: The signature from which the pubkey is recovered
+    :param public_key: The public key for verifying the signature
+    :param is_prehashed: True if the message is already pre-hashed. Default is False, and message will be hashed with SHA256
+    :return: The compressed byte-serialized representation of the recovered public key
+    """
+
+    signature = bytes(signature)
+    ecdsa_signature_size = Signature.expected_bytes_length()
+    if not len(signature) == ecdsa_signature_size:
```
michwill: I guess, doesn't matter here, but what's wrong with using `!=`?

nucypher/crypto/utils.py
```diff
@@ -47,39 +50,89 @@ def canonical_address_from_umbral_key(public_key: UmbralPublicKey) -> bytes:
...
+    else:
+        raise ValueError("Signature recovery failed. "
+                         "Either the message, the signature or the public key is not correct")
+
+
+def get_coordinates_as_bytes(point: Union[Point, UmbralPublicKey, SignatureStamp],
+                             x_coord=True,
+                             y_coord=True) -> bytes:
+
+    try:
```
michwill: What's happening in this `try` block?
Is what meant this?
```
if isinstance(point, Point):
    ...
elif isinstance(point, UmbralPublicKey):
    ...
else:
    ...
```

nucypher/crypto/utils.py
cygnusv: Yeah, nothing wrong. In the codebase, it seems the `not foo == bar` pattern is more used, but I can change it to `foo != bar` if that's more readable.

nucypher/crypto/utils.py
cygnusv: The idea is that, whatever you pass into the function (Point, UmbralPublicKey or SignatureStamp), we need its uncompressed point representation at some moment. For the first two cases you can just call `.to_bytes(is_compressed=False)`, but in case it's a stamp then you need the public key first. Some Pythonistas (like @jMyles and @KPrasch) preach that it's [easier to ask for forgiveness than permission](https://docs.python.org/2/glossary.html#term-eafp) (which in this case implies the use of `try/except` rather than `isinstance`) and I'm starting to believe it 😛

michwill: Eh, I think in this particular case it makes the code less readable: too implicit for even humans to understand. A random reader would not be able to guess that it's SignatureStamp which has `as_umbral_pubkey` method.

Isn't this one more clear?
```
if isinstance(point, SignatureStamp):
    point = point.as_umbral_pubkey()

coordinates_as_bytes = point.to_bytes(is_compressed=False)[1:]
...
```

cygnusv: Gotcha. I'll make the changes 👍

cygnusv: ✅

nucypher/crypto/utils.py
cygnusv: ✅

nucypher/crypto/utils.py
```diff
@@ -41,45 +44,93 @@ def construct_policy_id(label: bytes, stamp: bytes) -> bytes:
...
+    """
+
+    signature = bytes(signature)
+    ecdsa_signature_size = Signature.expected_bytes_length()
+    if len(signature) != ecdsa_signature_size:
+        raise ValueError(f"The signature size should be {ecdsa_signature_size} B.")
+
+    kwargs = dict(hasher=None) if is_prehashed else {}
+    for v in (0, 1):
+        v_byte = bytes([v])
```
tuxxy: :+1:

nucypher/crypto/utils.py
```diff
@@ -41,45 +44,93 @@ def construct_policy_id(label: bytes, stamp: bytes) -> bytes:
...
+def get_coordinates_as_bytes(point: Union[Point, UmbralPublicKey, SignatureStamp],
+                             x_coord=True,
+                             y_coord=True) -> bytes:
+    if isinstance(point, SignatureStamp):
+        point = point.as_umbral_pubkey()
+
+    coordinates_as_bytes = point.to_bytes(is_compressed=False)[1:]
+    middle = len(coordinates_as_bytes)//2
+    if x_coord and y_coord:
+        return coordinates_as_bytes
```
tuxxy: Would it be better to return this as a tuple? Seems like this would be a better function that way, but this works too.

tests/blockchain/eth/contracts/lib/test_signature_verifier.py
```diff
@@ -52,36 +52,27 @@ def test_recover(testerchain, signature_verifier):
...
+    v = get_signature_recovery_value(message, signature, umbral_pubkey)
+    recoverable_signature = bytes(signature) + v
+
     # Check recovery method in the contract
     assert signer_address == to_normalized_address(
         signature_verifier.functions.recover(message_hash, recoverable_signature).call())
 
     # Also numbers 27 and 28 can be used for v
-    recoverable_signature = recoverable_signature[0:-1] + bytes([recoverable_signature[-1] + 27])
+    recoverable_signature = recoverable_signature[:-1] + bytes([recoverable_signature[-1] + 27])
```
tuxxy: This is beginning to smell like we need to probably add this functionality to `Signature`.

nucypher/characters/lawful.py
```diff
@@ -390,12 +386,12 @@ class Bob(Character):
...
-    class IncorrectCFragReceived(Exception):
+    class IncorrectCFragsReceived(Exception):
         """
-        Raised when Bob detects an incorrect CFrag returned by some Ursula
+        Raised when Bob detects incorrect CFrags returned by some Ursulas
         """
-        def __init__(self, evidence):
-            self.evidence = evidence
+        def __init__(self, grievances):
+            self.grievances = grievances
```
tuxxy: I noticed this in my first pass through, but it's not really sitting well with me.

nucypher/characters/lawful.py
```diff
@@ -609,36 +605,39 @@ def retrieve(self, message_kit, data_source, alice_verifying_key, label):
         work_orders = self.generate_work_orders(map_id, capsule)
 
         cleartexts = []
+        the_airing_of_grievances = []
```
tuxxy: I do on the other hand enjoy this.

nucypher/characters/lawful.py
cygnusv: Yeah, I also had second thoughts with these two lines in particular. I'll revert it.

nucypher/crypto/utils.py
cygnusv: Not really, since this function is used to get a single bytes sequence with the coords.

tests/blockchain/eth/contracts/lib/test_signature_verifier.py
cygnusv: Can you clarify this? This line in particular does something very specific for the test, nothing actually useful.

nucypher/characters/lawful.py
KPrasch: Wow - This is Grim

tests/blockchain/eth/contracts/lib/test_signature_verifier.py
tuxxy: Why not add recovery methods to the `Signature` object? Why add them to `utils`?

cygnusv: Mainly because `Signature` is from `umbral`, although I agree that from an OOP perspective it makes sense to do it in `Signature`. Should I do this in Umbral instead?

nucypher/characters/lawful.py
cygnusv: ✅

KPrasch:
Needs rebase over master

PR #955, opened 2019-05-06
Enable learning tests in CI test workflow

cygnusv:


.circleci/config.yml
```diff
@@ -386,24 +337,15 @@ jobs:
             pipenv run pytest --cov=nucypher --cov-report xml:reports/coverage.xml $(circleci tests glob "tests/blockchain/eth/contracts/**/test_*.py" | circleci tests split --split-by=timings)
       - capture_test_results
 
-  config:
+  basics:
```
KPrasch: Heh - that's a funny lump of tests.  What made you group these together?

.circleci/config.yml
```diff
@@ -416,36 +358,34 @@ jobs:
...
       - run:
-          name: Character Tests
+          name: Nucypher CLI Tests
           command: |
-            pipenv run pytest --cov=nucypher --cov-report xml:reports/coverage.xml $(circleci tests glob "tests/characters/**/test_*.py" | circleci tests split --split-by=timings)
+            pipenv run pytest --cov=nucypher --cov-report xml:reports/coverage.xml $(circleci tests glob "tests/cli/**/test_*.py" | circleci tests split --split-by=timings)
       - capture_test_results
 
-  learning:
+  tests_ok:
```
KPrasch: Also interesting - Why the extra step?  Is the green / red success not enough indication, or is this a matter of making the "fan-in" logic more maintainable?

cygnusv: Yeah, it's purely cosmetic, to collect all test jobs in a single entry point for subsequent jobs.

.circleci/config.yml
cygnusv: I tried to reduce the number of containers. Previous jobs bundled in basics wasted a total of 4 or 5 containers, and now it's reduced to only one. They were short test modules abreast anyway. My heuristic was to get all jobs to last for at least 1 minute; otherwise, they were sufficiently small to be bundled.

KPrasch: Gotcha - so more or less these are all the fastest tests.

PR #700, opened 2019-01-29
[WIP] Make Ursula web status hideous in a different way

KPrasch:
Use default bootstrap styles to improve visual presentation of the Ursula web status page.

PR #959, opened 2019-05-08
I feel pretty...Make Moe/Ursula status pages slightly prettier

derekpierre:
seinfeld

Improve the look of Moe and Ursula status pages.

Used Plotly Dash to improve the pages, use common functionality, and remove the need for templates. Moe status page is updated via a Hendrix web socket notification system - it can run for a long time. Ursula status pages are periodically updated using a time interval (currently every 2s).

Related to PR #700 which was abandoned.

*Moe*
Screen Shot 2019-05-08 at 3 21 48 PM

*Ursula*
Screen Shot 2019-05-08 at 4 19 30 PM


nucypher/network/status/assets/hendrix-update.js
```diff
@@ -0,0 +1,32 @@
+window.onload = function () {
+    const origin_hostname = window.location.hostname;
+    const socket = new WebSocket("ws://"+ origin_hostname +":9000");
```
derekpierre: I was hoping to make the port number be configurable by adding another hidden element and getting the value from it eg. for a hidden div `ws_port = document.getElementById('...').innerHTML`. However, there seems to be some timing issue with `window.onload` and when the element is actually rendered - this is also the reason for the null checks below for the hidden buttons.

This script is specified in the footer of the page automatically since the file is in the `assets` folder, so it is weird that when this function is called, the element is not yet loaded - some Dash issue most likely.

vepkenez: to just get it to work you could probably do something like
```
window.onload = function () {
const socket;
const origin_hostname = window.location.hostname;
setTimeout(function() {
    socket = new WebSocket("ws://"+ origin_hostname +":9000");
}, 500 <or whatever>);
}
```

derekpierre: I get connected to the socket, whose default port is 9000 and that works fine. 

The issue is if anyone ran Moe and used the `--ws_port` option to use a different port, that would be a problem since the `9000` is hardcoded.

Unless you mean setting a timeout for the WebSocket is just good practice anyway...

vepkenez: No the setTimeout is definitely _not_ a good practice but it usually hackily solves issues like what you're seeing.

I don't know Dash at all really but couldn't you do something like 
`html.Script("var websocketport={}".format(MoeSocketPort))` in your header function?

vepkenez: Also, in my opinion we should make all these hardcoded ports into environment variables.

derekpierre: Thanks! I'm definitely not a Dash expert.

There seems to be some issues with the html.Script() component, https://github.com/plotly/dash/issues/93 and it wasn't in their recommendation in https://dash.plot.ly/external-resources, so I haven't really tried to use it. That being said, perhaps I'll give it a whirl.

vepkenez: ugh.  good luck!

derekpierre: Turns out it was better to just add the script to the page template itself.

Thanks for the discussion!

nucypher/characters/chaotic.py
```diff
@@ -46,8 +47,7 @@ class MonitoringTracker(FleetStateTracker):
         def record_fleet_state(self, *args, **kwargs):
             new_state_or_none = super().record_fleet_state(*args, **kwargs)
             if new_state_or_none:
-                checksum, new_state = new_state_or_none
-                hey_joe.send({checksum: self.abridged_state_details(new_state)}, "states")
+                hey_joe.send(None, "states")
```
KPrasch: Sending `None` over the websocket?  What's going on here?

vepkenez:
I think None in that context might be an argument to hey joe that means
“broadcast to all channels”?

nucypher/characters/chaotic.py
derekpierre: `None` is the message payload. I don't really need to pass any data to the subscriber (i.e. the UI), just a notification that the status UI page for Moe should update its 'states', based on a change. Same idea for 'nodes' notification. The UI will then know to update what is being shown.

nucypher/characters/lawful.py
```diff
@@ -793,6 +794,9 @@ def __init__(self,
                     suspicious_activity_tracker=self.suspicious_activities_witnessed,
                     serving_domains=domains,
                 )
+                # attach status app to rest_app
+                UrsulaStatusPage(self, self.nickname, rest_app, '/status/')
```
derekpierre: This could be improved i.e. added within the `make_rest_app` call by David's PR #961 since I needed an instance of Ursula to pass in.

tuxxy:
@derekpierre Instead of pushing this into master, we're going to create an epic for Ursula's frontend, which this PR will be the first in there.

derekpierre:
@tuxxy ok. Once that happens, is it best to create a new PR merging into the epic, and then link and close this one?

tuxxy:
@derekpierre No, we can just change the branch that this one goes into.

PR #960, opened 2019-05-10
WorkLock

szotov:
Introduces `WorkLock` implementation. 
Distribution consists of 4 phases:
* Before bidding, nothing can do
* Bidding. Everyone can transfer ETH for tokens (multiple times) using `bid()`
* Claiming. After bidding is over stakers can claim their token as deposit in `StakingEscrow`
* Refund. After any work is done staker can ask for refund ETH

API changes:
* Removed unused `preDeposit` method from `StakingEscrow`
* `prolongStake` function in `StakingEscrow`
* `getCompletedWork` function in `StakingEscrow` (not necessary to have it in agent)
* additional `deposit(address _miner, uint256 _value, uint16 _periods)` to stake for somebody in `StakingEscrow`
* `WorkLock` contract + `setWorkLock` in `StakingEscrow`


nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -0,0 +1,147 @@
...
+    * @notice Refund ETH for the work done
+    **/
+    function refund() public returns (uint256 refundETH) {
+        WorkInfo storage info = workInfo[msg.sender];
+        require(info.claimed, "Tokens are not claimed");
+        require(info.depositedETH > 0, "Nothing deposited");
+        uint256 currentWork = escrow.getWorkDone(msg.sender);
+        uint256 workDone = currentWork.sub(info.workDone);
+        require(workDone > 0, "No work has been done.");
+        refundETH = workDone.div(refundRate);
```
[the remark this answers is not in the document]
arjunhassard: Currently inflationary rewards earned (in NU) since the last ETH withdrawal are the primary input to generate `workDone`. Then, unless I'm mistaken, the initial sum of NU acquired via the WorkLock (`claimedTokens`) is not incorporated into the calculation to produce `refundETH `.

Given that it is measured as rewards earned in absolute terms (i.e. # of NU tokens), the ability/power to produce 'work' is varies enormously between large and small stakers. Hence a constant (`refundRate`) cannot suitably cover the range of different stake sizes – large stakers will be able to get their ETH back many times faster than small stakers.

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    NuCypherToken public token;
+    MinersEscrow public escrow;
+    uint256 public startBidDate;
+    uint256 public endBidDate;
+    // ETH -> NU
+    uint256 public depositRate;
+    // Work (reward in NU) -> ETH
+    uint256 public refundRate;
```
arjunhassard: However we end up calculating `refundETH`, it may be prudent to hold off specifying associated parameters (like `refundRate`) until we've seen work activity on the network for some months. Otherwise we're shooting blindly, and it would justifiably upset stakers if we were forced to adjust it after the fact – it actually could be better to leave a variable deliberately TBD than to change it later (unless this change emerges via a community-driven process). 

Of course this is part of a wider debate about governance and the role of the NuCypher team following the launch of the network. There is (rightly) skepticism about networks that remain almost entirely centralized with vague promises of foundation control loosening later. At the same time, you can make the argument that in at least the first months after launch, there will be some dependence on the NuCypher team regardless (e.g. efforts to drive network adoption). Furthermore, stakers may prefer the reassurance that some key parameters can be adjusted to avoid network shocks or capitulation. 

@mswilkison @michwill

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -0,0 +1,147 @@
...
+        workDone = refundETH.mul(refundRate);
+        info.workDone = info.workDone.add(workDone);
+        emit Refund(msg.sender, refundETH, workDone);
+        msg.sender.transfer(refundETH);
+    }
+
+    /**
+    * @notice Get remaining work to full refund
+    **/
+    function getRemainingWork(address _miner) public view returns (uint256) {
```
arjunhassard: @szotov is this function for stakers to check how much more work they need to do to get back the rest of their deposit? Does it serve some other purpose?

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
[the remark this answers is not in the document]
szotov: I'm not sure that I understand whole idea, few moments:
* `refundETH` and `claimedTokens` are already connected through `refundRate`. If `refundRate` == `depositRate` then reward tokens have same value as claimed tokens.
* `relativeWork` in suggetions almost allways zero, because `workDone` will be less than claimed token initially
* Speed of refund in percents will be the same independatly of initial stake, so depending on state of whole net big and small stakers need same amount of time to get 100% of ETH back

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
szotov: Right now only for checking by staker or anyone else

michwill:
>     * Making your ETH refund, given some work/compensation, a function of your initial stake size. This is a compromise between two more extreme positions: a) refunds are a function of your _current_ stake size and b) refunds are driven only by how many block rewards you've earned, regardless of your stake size. These are both flawed – a) could disincentivize buying NU and increasing your stake and b) is very unfair on small stakers – see suggestion to WorkLock.sol
>

I actually disagree on how (b) is unfair. Your compensation is proportional to your stake size, so the more initial stake you have, the more work per day you are able to produce. And it will take exactly the same number of days to fully unlock (for both large and small staker).

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
michwill: IMO, this parameter mainly depends on the inflation rate. E.g. if this refund rate is the same as deposit rate, you'r have ~1 year to get fully refunded (depends on how you stake though) if you have 100% initial emission rate.
If this is too aggressive and we make it lower, that, imo, should affect the refund rate.

How about we make the refund rate linked to emission rate, so that the full refund time is 1 year, for example?

Otherwise, if we left it TBD, I think participants would be very meh on it

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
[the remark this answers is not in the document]
michwill: @arjunhassard percent-wise, there is no difference between large and small stakers, really.

The only problem I currently see is the minimal stake (e.g. very small stakers either cannot stake, or we have gas wars)

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -481,16 +486,27 @@ contract StakingEscrow is Issuer {
...
 
     /**
     * @notice Deposit tokens
     * @param _staker Staker
+    * @param _payor Owner of tokens
     * @param _value Amount of tokens to deposit
     * @param _periods Amount of periods during which tokens will be locked
     **/
-    function deposit(address _staker, uint256 _value, uint16 _periods) internal isInitialized {
+    function deposit(address _staker, address _payor, uint256 _value, uint16 _periods) internal isInitialized {
```
KPrasch: Did you mean `payer`?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -351,9 +373,29 @@ contract StakingEscrow is Issuer {
         return workerToStaker[_worker];
     }
 
+    /**
+    * @notice Get work that done by the staker
+    **/
+    function getWorkDone(address _staker) public view returns (uint256) {
```
KPrasch: This function name reads more like a verb than a getter.

tests/blockchain/eth/contracts/contracts/WorkLockTestSet.sol
```diff
@@ -0,0 +1,59 @@
+pragma solidity ^0.5.3;
```
KPrasch: Is this still the pragma we want to specify?

szotov: it works more like restriction >= 0.5.3, < 0.6.0, means that you can use 0.5.3 if you want, so it's still ok, maybe I'll change it to higher version in one of the next PR

cygnusv: Isn't it the version we're using in all our contracts?

szotov: We have 0.5.9 in installation script, main reason for 0.5.3 in contracts is that no changes in syntax after 0.5.3 (or maybe even earlier). But I'll try different versions from optimization point of view in next PR

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
szotov: ✔️

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
szotov: ✔️

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -501,7 +517,7 @@ contract StakingEscrow is Issuer {
             policyManager.register(_staker, getCurrentPeriod());
         }
         info.value = info.value.add(_value);
-        token.safeTransferFrom(_staker, address(this), _value);
+        token.safeTransferFrom(_payer, address(this), _value);
```
cygnusv: Why are you introducing a `_payer` here, apart from the `_staker`?

szotov: I split staker and who transfer tokens (payer) inside internal method because worklock will call deposit in escrow and so it is payer, but staker will have other address as a worklock user. Also added method if somebody wants stake token for anyone else

PR #963, opened 2019-05-13
[EPIC] Doubtfire: Modifications for Monitoring Applications

KPrasch:
Monitor crawler and dashboard changes moved here: <https://github.com/nucypher/nucypher-monitor>

### What this does

- Preserves History of Dash implementation for the nucypher monitoring services.
- Understand that "Moe' is not a character.
- No more Moe, Completely deprecated (relocated to external repo).
- Introduces SQLite based `NodeStorage` subclass
- Allow `Learner` to be directly subclassed; Thin `verify_from`
- Improve Handling of self-signed certificate generation for node discovery vs web-hosting-only.
- Additional stake start and end time properties for statistical use


cygnusv:
What about we dust off this?

derekpierre:
@cygnusv I'm game if others are interested - @jMyles @tuxxy @KPrasch ?

Let me know and I can look into rebasing.

KPrasch:
Yeah - Im in!

Pipfile
```diff
@@ -33,6 +33,11 @@ click = ">=7.0"
 colorama = "*"
 boto3 = "*"
 flask-limiter = "*"
+jinja2 = ">=2.10.1"
+dash = "==0.34.0"
+dash-core-components = "==0.41.0"
+dash-html-components = "==0.13.4"
+dash-dangerously-set-inner-html = "==0.0.2"
```
KPrasch: Hmm, I'm a little concerned with some of these deps, like this one.  Whats the story here?

Pipfile
```diff
@@ -58,6 +63,9 @@ sphinx_rtd_theme = "*"
 aafigure = "*"
 # CLI
 nucypher = {editable = true,path = "."}
+# UI testing
+pytest-dash = "==2.1.0"
+instapy-chromedriver = "*"
```
KPrasch: Also here - Is this used for installing chromedriver?  Is this the recommended install?

Pipfile
derekpierre: This was used in `status_page.py` for the fleet state icon. 
It's basically a way to specify html within dash.

The `fleet_state_icon` property of the node_dict returns html - and the individual values cannot be obtained through object calls.

Pipfile
derekpierre: It is used for installing the chromedriver - it handles platform-specific installation of the driver.

KPrasch:
After rebasing over master - I am getting:

```
(nucypher) k@nova:~/Git/nucypher$ nucypher ursula run --dev -F 
...
  File "/home/k/Git/nucypher/nucypher/network/server.py", line 49, in <module>
    with open(os.path.join(TEMPLATES_DIR, "basic_status.j2"), "r") as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/k/Git/nucypher/nucypher/network/templates/basic_status.j2'

```

derekpierre:
@KPrasch working through it...basically that template was removed in favour of flash/dash.

cygnusv:
Two comments:
* The "Checksum" column is wrong. It's showing the worker address, instead of the checksum address.
* In the "Num Stakers over the previous 30 days" chart, can you make the y-axis start from 0?

derekpierre:
> * In the "Num Stakers over the previous 30 days" chart, can you make the y-axis start from 0?

@cygnusv currently, the y-axis range is adjusted based on the values plotted. Any reason, in particular, you want the y-axis to start from 0? Currently, the lowest value since we started tracking the network was '12' stakers.

cygnusv:
> @cygnusv currently, the y-axis range is adjusted based on the values plotted. Any reason, in particular, you want the y-axis to start from 0? Currently, the lowest value since we started tracking the network was '12' stakers.

Charts where the y-axis don't start from 0 can be misleading, as visually, a growing trend can be wildly exaggerated. E.g., If we have 1000 nodes for some time, and suddenly 10 more nodes show up, the chart may look like the number of nodes has sky-rocketed, but in reality it's just 1% higher. 

More info here (it's just a quick search, perhaps there's a better reference out there):
https://uxdesign.cc/designing-charts-principles-every-designer-should-know-part-2-ce1e06af56fc

derekpierre:
> Charts where the y-axis don't start from 0 can be misleading, as visually, a growing trend can be wildly exaggerated. E.g., If we have 1000 nodes for some time, and suddenly 10 more nodes show up, the chart may look like the number of nodes has sky-rocketed, but in reality it's just 1% higher.
> 
> More info here (it's just a quick search, perhaps there's a better reference out there):
> https://uxdesign.cc/designing-charts-principles-every-designer-should-know-part-2-ce1e06af56fc

@cygnusv Thanks for the link - that was an interesting read. I can understand their reasoning for wanting the 0 line for the y-axis. My concern is that if we end up in a scenario where there are 50-100 stakers for the entirety of the 30-day range, and we use the 0-line, then half of the graph becomes useless. Any thoughts there?

derekpierre:
@cygnusv I'm sold for setting the y-axis range to start at 0 for all graphs - the relative changes of staking NU tokens and num stakers (currently side by side graphs) is less clear when both use different ranges. Should be resolved via #1437

nucypher/blockchain/eth/token.py
```diff
@@ -586,6 +587,15 @@ def __init__(self,
...
 
+    @property
+    def initial_period(self) -> int:
+        return self.__initial_period
+
+    @property
+    def terminal_period(self) -> int:
+        return self.__terminal_period
+
+    @validate_checksum_address
```
derekpierre: validate not needed since a checksum is not passed as a parameter.
```suggestion
```

nucypher/crypto/api.py
```diff
@@ -199,6 +206,29 @@ def generate_self_signed_certificate(host: str,
...
+        raise ValueError("checksum address cannot be used to generate standard self-signed certificates.")
+    cert = __generate_self_signed_certificate(checksum_address=None, *args, **kwargs)
+    return cert
+
+
+def read_certificate_pseudonym(certificate: Certificate):
+    try:
+        pseudonym = certificate.subject.get_attributes_for_oid(NameOID.PSEUDONYM)[0]
+    except IndexError:
+        raise RuntimeError("Invalid teacher certificate encountered: No checksum address present as pseudonym.")
```
derekpierre: I know that `RuntimeError` is in line with what was done previously,  any chance we could modify the exception to be `ValueError`? same for the invalid checksum check below?

tests/cli/test_moe.py
```diff
@@ -19,9 +19,7 @@ def test_run_moe(click_runner, federated_ursulas):
 
     _ursula_output = yield threads.deferToThread(start_pytest_ursula_services, ursula=ursula)
 
-    test_ws_port = select_test_port()
```
derekpierre: we can delete this entire file since we removed Moe - deleted myself in https://github.com/nucypher/nucypher/pull/963/commits/e3acb7788b21e10709775c0a813119694881b467

nucypher/config/storages.py
```diff
@@ -178,17 +179,17 @@ class ForgetfulNodeStorage(NodeStorage):
     _name = ':memory:'
     __base_prefix = "nucypher-tmp-certs-"
 
-    def __init__(self, *args, **kwargs) -> None:
+    def __init__(self, parent_dir: str = None, *args, **kwargs) -> None:
```
michwill: `certificate_dir`? `node_info_dir`?

PR #951, opened 2019-04-27
[EPIC] Decentralized Characters & Integrated Ethereum Node Providers

KPrasch:
Chaperones: @KPrasch 

### Epic Goals - The Decentralized Cast

- Blockchain deployment operations and test-staking
- Cross-compatibility for ethereum clients 
- Version `E` EIP-191 signature and verification support
- On-chain staking verification as part of standard node verification & signature check
- Test Token Faucet
- Decentralized domains
- CLI Character Controller Operations 

Efforts:

- [x] Handle genesis block configuration for private chains
- [x] Save Deployment Metadata to Disk (#785)
- [x] Handle Contract Registry Publication and Fetch (#461)
- [x] CLI-launched contract upgrades (#460)
- [x] Felix Ansible Provisioning 
- [x] Implement Geth child process management API
- [x] #976 
- [X] #1028 
- [X] #1027 
- [X] #1017 
- [X] #1016 
- [X] #1011 
- [X] #980, #970 
- [X] #991
- [X] #969
- [X] #1025
- [X] #1026 
- [x] Ethereum Client Signature Integration for Node Discovery - #1031 
- [X] #1040 
- [X] #1048 
-------------------------------
###### Hawksbeard

> Crepis, commonly known in some parts of the world as hawksbeard or hawk's-beard (but not to be confused with the related genus Hieracium with a similar common name), is a genus of annual and perennial flowering plants of the family Asteraceae superficially resembling the dandelion

PR #984, opened 2019-05-17
Casual Alice with Policy and Payment Defaults

### LINKED ISSUE #975: NU instance equality
@cygnusv Indicates that:

```
>>> NU(0, 'NU') is NU.ZERO()
False
>>> NU(0, 'NU') == NU.ZERO()
True
```

This may lead to some confusing behavior. What can we do to mitigate accidentally incorrect comparing?

### LINKED ISSUE #1015: Policy value can't be 0
Current behavior of `PolicyManager` contract is to require the value to be greater than 0. However, this is not enforced by `BlockchainPolicy`.

KPrasch:
#### What this does 

- Follow up for #951
- Decentralized Grant
- Save / Load policy defaults for M, N, duration and value on-disk via `AliceConfiguration`
- Allow policy defaults to be used or overridden on Alice APIs
- Use the CLI to initialize and run an Alice using an internally managed geth process (`--geth`)
- Use a common function for making CLI characters (reduces character CLI length)
- Removes policy mocking from decentralized grant testing
- Decentralized `test_cli_lifecycle` test
- `checksum_public_address` -> `checksum_address`
- Introduces `Agency` metaclass (makes agent subclasses singletons)
- Style and convention improvements for character CLI modules

#### Examples
```

# Configure a new persistent Alice with an integrated geth node
# using recommended testnet defaults, Save chain syncing for later.
$ nucypher alice init --geth --no-sync

# Run an Alice IPC Control Server using an integrated geth node
$ nucypher alice run --geth

# Use default M, N, and Rate to Grant
$ nucypher alice grant --geth --bob-encrypting-key ... --bob-verifying-key ... --label ...
```

Fixes #975 
Fixes #1015


nucypher/config/node.py
```diff
@@ -623,62 +629,61 @@ def initialize(self, password: str, download_registry: bool = True) -> str:
...
-                checksum_address = self.blockchain.interface.w3.eth.accounts[0]  # TODO: Make this a configurable default in config files
+                    # Get or create wallet address (geth etherbase)
+                    checksum_address = self.provider_process.ensure_account_exists(password=password)
 
-            # Addresses read from some node keyrings (clients) are *not* returned in checksum format.
-            checksum_address = to_checksum_address(checksum_address)
+                # "Formal Geth" - Manual Web3 Provider, We assume is already running and available
+                else:
+                    self.connect_to_blockchain()
+                    if not self.blockchain.interface.w3.eth.accounts:
```
derekpierre: I believe this was added to a function in the `Web3Client`. If not, we should add the function.

cygnusv:
Two questions:
* Do you want to merge this on master or on #1029?
* Was there a consensus on changing `checksum_public_address` to `checksum_address`? It doesn't seem as something critical. Also, I'm not a big fan that this type of renames that impact dozens of files are part of bigger PRs. Now there will be lots of conflicts for all the in-parallel work done in other branches.

KPrasch:
@cygnusv - Yeah I think this changeset is best merged *under* #1029 .  wrt `checksum_address`, If you prefer I can move the renaming to it's own PR, but I dont mind doing the subsequent rebases.

nucypher/characters/lawful.py
```diff
@@ -173,34 +206,74 @@ def create_policy(self,
 
         return policy
 
+    def generate_m_of_n(self, m: int = None, n: int = None):
+        m = m or self.m
+        n = n or self.n
+        payload = dict(m=m, n=n)
```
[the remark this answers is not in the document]
KPrasch: I just prefer named access to the attributes in the dict ie `mof_n['n']`.  Perhaps a `namedtuple`? :-)

nucypher/characters/lawful.py
```diff
@@ -173,34 +206,74 @@ def create_policy(self,
...
+                                   value: int = None,
+                                   rate: int = None,
+                                   first_period_value: int = None,
+                                   first_period_rate: float = None) -> dict:
+        """
+        Construct policy creation from parameters or overrides.
+        """
+
+        m_of_n = self.generate_m_of_n(m=m, n=n)
+        m, n = m_of_n['m'], m_of_n['n']
```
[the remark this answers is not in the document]
KPrasch: I'm sold on a `namedtuple`

nucypher/characters/lawful.py
```diff
@@ -71,23 +72,48 @@ class Alice(Character, PolicyAuthor):
     _default_crypto_powerups = [SigningPower, DecryptingPower, DelegatingPower]
 
     def __init__(self,
-                 is_me=True,
-                 federated_only=False,
+                 checksum_address: str = None,
+                 m: int = None,
```
michwill: Great that we allow to specify default `m` and `n`. Should we not offer some values here?

KPrasch: I've got the default values specified on `AliceConfiguration`, leaving character decoupled from config.

nucypher/characters/lawful.py
```diff
@@ -119,51 +151,52 @@ def generate_kfrags(self, bob: 'Bob', label: bytes, m: int, n: int) -> List:
         :param n: Total number of kfrags to generate
         """
 
-        self.revocation_kits = dict()
-        bob_pubkey_enc = bob.public_keys(DecryptingPower)
+        params = self.generate_m_of_n(m=m, n=n)
```
michwill: Hmm... I think, methods for such small things are only confusing the reader who doesn't like jumping over the code. Why not simply:
```
m = m or self.m
n = n or self.n
...
return delegating_power.generate_kfrags(..., m=m, n=n)
```
More readable, no? Or do we want to have more involved logic for selecting m and n eventually?

nucypher/characters/lawful.py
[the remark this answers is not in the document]
michwill: dict is understandable. Why a special method for this at all?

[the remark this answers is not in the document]
KPrasch: I... actually do not remember now, If we can do without that's clearly better, unless there is any additional logic we want to implement here in the future.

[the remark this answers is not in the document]
michwill: I'd probably reduce the number of calls when unnecessary (which I think is the case here). The code which has overly too many functions is IMO less readable (even when you have an IDE!)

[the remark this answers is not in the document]
KPrasch: Yes, of course.  This was part of an effort to reduce the number of occurrences of the code found in  `generate_policy_params`.

nucypher/characters/lawful.py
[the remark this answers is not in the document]
KPrasch: Actually, never mind I'm going to remove it

nucypher/cli/characters/alice.py
```diff
@@ -26,10 +25,11 @@
 @click.option('--config-file', help="Path to configuration file", type=EXISTING_READABLE_FILE)
 @click.option('--provider-uri', help="Blockchain provider's URI", type=click.STRING)
 @click.option('--geth', '-G', help="Run using the built-in geth node", is_flag=True)
+@click.option('--poa', help="Inject POA middleware", is_flag=True, default=None)
 @click.option('--no-registry', help="Skip importing the default contract registry", is_flag=True)
 @click.option('--registry-filepath', help="Custom contract registry filepath", type=EXISTING_READABLE_FILE)
-@click.option('--bob-encrypting-key', help="Bob's encrypting key as a hexideicmal string", type=click.STRING)
-@click.option('--bob-verifying-key', help="Bob's verifying key as a hexideicmal string", type=click.STRING)
+@click.option('--bob-encrypting-key', help="Bob's encrypting key as a hexadeicmal string", type=click.STRING)
+@click.option('--bob-verifying-key', help="Bob's verifying key as a hexadeicmal string", type=click.STRING)
```
cygnusv: Almost there!

nucypher/cli/characters/alice.py
```diff
@@ -13,59 +10,103 @@
...
+@click.option('--pay-with', help="Run with a specified account", type=EIP55_CHECKSUM_ADDRESS)
+@click.option('--bob-encrypting-key', help="Bob's encrypting key as a hexadecimal string", type=click.STRING)
+@click.option('--bob-verifying-key', help="Bob's verifying key as a hexadecimal string", type=click.STRING)
 @click.option('--label', help="The label for a policy", type=click.STRING)
 @click.option('--m', help="M-Threshold KFrags", type=click.INT)
 @click.option('--n', help="N-Total KFrags", type=click.INT)
-@click.option('--dev', '-d', help="Enable development mode", is_flag=True)
-@click.option('--force', help="Don't ask for confirmation", is_flag=True)
-@click.option('--dry-run', '-x', help="Execute normally without actually starting the node", is_flag=True)
+@click.option('--value', help="Total policy value", type=click.FLOAT)
```
cygnusv:
```suggestion
@click.option('--value', help="Total policy value (in Wei)", type=click.INT)
```
If you track this value all the way down to the transaction in `BlockchainPolicy` you'll see there's no conversion anywhere. So either you apply this suggestion, or you convert the input to Wei. Either way, the `help` text here must make explicit what denomination it expects.

nucypher/cli/characters/alice.py
```diff
@@ -106,102 +163,113 @@ def alice(click_config,
...
 
-    elif action == "view":
-        """Paint an existing configuration to the console"""
-        configuration_file_location = config_file or alice_config.config_file_location
-        response = AliceConfiguration._read_configuration_file(filepath=configuration_file_location)
-        return ALICE.controller.emitter(response=response)         # TODO: Uses character control instead
+    elif action == "destroy":
+        """Delete all configuration files from the disk"""
+        if dev:
+            message = "'nucypher ursula destroy' cannot be used in --dev mode"
```
cygnusv:
```suggestion
            message = "'nucypher alice destroy' cannot be used in --dev mode"
```

nucypher/cli/characters/alice.py
```diff
@@ -165,71 +177,87 @@ def alice(click_config,
...
                                                  "required options to create a new policy.")
 
+        # Request
         create_policy_request = {
             'bob_encrypting_key': bob_encrypting_key,
             'bob_verifying_key': bob_verifying_key,
             'label': label,
             'm': m,
             'n': n,
         }
```
cygnusv: The value is required for `BlockchainPolicies`. Anyway, I think we should get rid of the "create-policy" command. You cannot use the result for anything. If I'm not wrong, we already discussed this in Málaga with @vepkenez @jMyles and you.

nucypher/cli/characters/alice.py
```diff
@@ -165,71 +177,87 @@ def alice(click_config,
...
+        # Request
         return ALICE.controller.derive_policy_encrypting_key(label=label)
 
     elif action == "grant":
+
+        # Validate
+        if not all((bob_verifying_key, bob_encrypting_key, label)):
+            raise click.BadArgumentUsage(message="--bob-verifying-key, --bob-encrypting-key, and --label are "
+                                                 "required options to grant (optionally --m, --n, and --expiration).")
+
```
cygnusv: The presence of `--value` should be validated in case of decentralized Alice.

nucypher/cli/characters/bob.py
```diff
@@ -84,75 +91,87 @@ def bob(click_config,
...
-    elif action == "view":
-        """Paint an existing configuration to the console"""
-        response = BobConfiguration._read_configuration_file(filepath=config_file or bob_config.config_file_location)
-        return BOB.controller.emitter(response=response)
+    elif action == "destroy":
+        """Delete Bob's character configuration files from the disk"""
+
+        # Validate
+        if dev:
+            message = "'nucypher ursula destroy' cannot be used in --dev mode"
```
cygnusv:
```suggestion
            message = "'nucypher bob destroy' cannot be used in --dev mode"
```

tests/cli/test_cli_lifecycle.py
```diff
@@ -85,16 +92,23 @@ def test_cli_lifecycle(click_runner,
...
     # Alice performs an installation for the first time
     alice_init_args = ('alice', 'init',
-                       '--federated-only',
                        '--network', TEMPORARY_DOMAIN,
                        '--config-root', alice_config_root)
 
+    if federated:
+        alice_init_args += ('--federated-only', )
+    else:
+        alice_init_args += ('--provider-uri', 'tester://pyevm')
```
cygnusv:
```suggestion
        alice_init_args += ('--provider-uri', TEST_PROVIDER_URI)
```

tests/cli/test_cli_lifecycle.py
```diff
@@ -242,15 +262,20 @@ def _grant(teacher_uri):
...
+                      '--n', 3,
+                      '--expiration', (maya.now() + datetime.timedelta(days=3)).iso8601(),
                       '--label', random_label,
                       '--bob-encrypting-key', bob_encrypting_key,
                       '--bob-verifying-key', bob_verifying_key)
 
+        if federated:
+            grant_args += ('--federated-only',)
+        else:
+            grant_args += ('--provider-uri', 'tester://pyevm')
```
cygnusv:
```suggestion
            grant_args += ('--provider-uri', TEST_PROVIDER_URI)
```

nucypher/characters/lawful.py
```diff
@@ -173,34 +206,74 @@ def create_policy(self,
...
+            if not value and not rate:
+                value = int(self.rate * duration)
+                rate = self.rate
+            elif rate:
+                value = int(rate * duration)
+            elif value:
+                rate = value // duration
+
+            first_period_rate = first_period_rate or self.first_period_rate
+            first_period_value = first_period_value or int(rate * first_period_rate)
```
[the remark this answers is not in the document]
michwill: What is the `first_period_value` though? Is it the value for the time until the end of the period?
If so, it's logical if it is something like `rate * first_period_duration`, where `first_period_duration = second_period_time - now()`

nucypher/blockchain/eth/policies.py
```diff
@@ -255,20 +257,17 @@ def make_arrangements(self, network_middleware: RestMiddleware, *args, **kwargs)
 
     def publish(self, **kwargs) -> str:
 
-        if self.value is NON_PAYMENT:
-            self.value = 0
-
-        payload = {'from': self.author.checksum_public_address,
+        payload = {'from': self.author.checksum_address,
                    'value': self.value,
+                   'gas': 500_000,  # TODO: Gas management
```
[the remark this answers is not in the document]
KPrasch: Not a constant, per se, but a class attribute.  This value is specific to this expensive transaction - Needs follow up in a PR that allows for configurable gas management.

nucypher/blockchain/eth/policies.py
```diff
@@ -255,20 +257,17 @@ def make_arrangements(self, network_middleware: RestMiddleware, *args, **kwargs)
...
 
-        prearranged_ursulas = list(a.ursula.checksum_public_address for a in self._accepted_arrangements)
-
-        txhash = self.author.policy_agent.contract.functions.createPolicy(self.hrac()[:16],
-                                                                          self.lock_periods,
-                                                                          self.initial_reward,
-                                                                          prearranged_ursulas,
-                                                                          ).transact(payload)
+        prearranged_ursulas = list(a.ursula.checksum_address for a in self._accepted_arrangements)
+        policy_args = (self.hrac()[:16],     # bytes16 _policyID
```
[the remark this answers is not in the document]
KPrasch: Yeah - Good catch, this policy ID was set this way quite some time ago.  How do we proceed?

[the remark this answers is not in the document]
cygnusv: Perhaps just an issue to investigate later? It doesn't seem to be affecting current operation and it's not critically related to this PR.
Unless you want to tackle this now, of course.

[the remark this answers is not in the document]
KPrasch: I think it falls a just little out of scope for this PR.  I'm +1 on an issue.

nucypher/cli/characters/alice.py
KPrasch: Yeah, OK.  What can we check here?

nucypher/config/node.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/clients.py
```diff
@@ -186,37 +190,24 @@ def check_for_timeout(t):
...
+        self.log.info(f"Waiting for {self.chain_name.capitalize()} chain synchronization to begin")
+        while not self.syncing:
+            time.sleep(0)
+            check_for_timeout(t=120)
+
+        while self.syncing:
+
+            # current =
+            self.log.info(f"Syncing {self.syncing['currentBlock']}/{self.syncing['highestBlock']}")
+            time.sleep(5)
```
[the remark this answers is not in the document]
derekpierre: I have no concept of how long we expect this operation to be. If, as you said, it could be in such a wide range, seconds -> hours, then perhaps sleeping for 5 or 10s is sufficient. Just wasn't sure if there would be cases of us waiting for a long window when it took somewhere in the order of ms.

nucypher/config/node.py
```diff
@@ -623,62 +629,61 @@ def initialize(self, password: str, download_registry: bool = True) -> str:
...
-                checksum_address = self.blockchain.interface.w3.eth.accounts[0]  # TODO: Make this a configurable default in config files
+                    # Get or create wallet address (geth etherbase)
+                    checksum_address = self.provider_process.ensure_account_exists(password=password)
 
-            # Addresses read from some node keyrings (clients) are *not* returned in checksum format.
-            checksum_address = to_checksum_address(checksum_address)
+                # "Formal Geth" - Manual Web3 Provider, We assume is already running and available
+                else:
+                    self.connect_to_blockchain()
+                    if not self.blockchain.interface.client.accounts:
```
derekpierre: 💥

nucypher/cli/characters/alice.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/actors.py
```diff
@@ -493,7 +493,7 @@ def start_staking_loop(self, now=True) -> None:
     @property
     def is_staking(self) -> bool:
         """Checks if this Miner currently has active stakes / locked tokens."""
-        return bool(self.locked_tokens > NU.ZERO())
+        return bool(self.stakes)
```
cygnusv: Just curious: why `bool(self.stakes)` is the response here? What's the underlying logic?

nucypher/characters/control/specifications.py
```diff
@@ -60,7 +60,7 @@ def validate_response(self, interface_name: str, response: dict) -> bool:
 
 class AliceSpecification(CharacterSpecification):
 
-    __create_policy = (('bob_encrypting_key', 'bob_verifying_key', 'm', 'n', 'label'),  # In
+    __create_policy = (('bob_encrypting_key', 'bob_verifying_key', 'm', 'n', 'label', 'expiration'),  # In
```
cygnusv: How can we handle conditionally-required inputs in specifications? In this case, `value` is mandatory when not federated. Seems kind of tricky to do it.

KPrasch: Good eye! I've been thinking about this too, I think it's possible to allow values to be passes as `None`  all the way into character, with some modification to specification handling seen here.

nucypher/blockchain/eth/actors.py
KPrasch: ` self.__read_stakes()  # "load-in":  Read on-chain stakes` "syncs" the onchain stakes into the local staking cache on `Staker`, and will be set to `NO_STAKES.bool_value(False)` if there are no active stakes.

nucypher/characters/control/specifications.py
KPrasch: It might second Issue, but here's one <https://github.com/nucypher/nucypher/issues/1065>

PR #1002, opened 2019-05-23
[WIP] Abstract Web3 Clients

vepkenez:


vepkenez:
@derekpierre I tweaked your stuff and it's all working now.  What do you think?

nucypher/blockchain/eth/clients.py
```diff
@@ -15,16 +16,109 @@
...
+        return self.web3_instance.manager.request_blocking("parity_netPeers", [])
+
+    def unlock_account(self, address, password):
+        return self.web3_instance.parity.unlockAccount.unlockAccount(address, password)
+
+
+class GanacheClient(Web3ClientBase):
+
+    def peers(self):
+        return []
```
derekpierre: Are the return values for `peers` and `unlock_account` a result of how Ganache works or should we return a `NotImplementedError` instead?

nucypher/blockchain/eth/clients.py
```diff
@@ -15,16 +16,109 @@
...
+            }[node_technology]
+            return gclass(w3, *client_data)
+
+        except KeyError:
+            raise NotImplementedError(node_technology)
+
+
+class Web3ClientBase(object):
+
+    def __init__(self, w3, node_technology, version, backend, **kwargs):
```
derekpierre: `client_data` can sometimes split into a list of size 4 not 3 - eg. for Geth and Parity client version strings; does that cause a problem here? This was the reason for the tricky splitting logic dependent on node_technology.

nucypher/blockchain/eth/interfaces.py
```diff
@@ -210,7 +206,8 @@ def client_version(self) -> str:
             return "Unknown"
 
         if self.is_connected:
-            node_info = self.w3.clientVersion.split(os.sep)
+            node_info = self.w3.clientVersion.split('/')
```
derekpierre: do you need to do this, can the technology be obtained from `self.client`? It would be nice to keep that stuff obfuscated within the client

derekpierre: Also, I really wanted to rename the `client_version` method on line 204 - it's more of a `client_type` than a `client_version`, what do you think?

nucypher/blockchain/eth/clients.py
```diff
@@ -15,16 +16,109 @@
...
+        PARITY = 'Parity'
+        GANACHE = 'EthereumJS TestRPC'
+
+        try:
+            gclass = {
+                GETH: GethClient,
+                PARITY: ParityClient,
+                GANACHE: GanacheClient,
+            }[node_technology]
+            return gclass(w3, *client_data)
```
derekpierre: 👍 - ooh nice!

derekpierre:
@vepkenez - good stuff; I like the cleanup. Left some comments.

nucypher/blockchain/eth/clients.py
vepkenez: Yes I left this for when we have a way to test the various providers

nucypher/blockchain/eth/clients.py
vepkenez: I just assumed “don’t crash the whole thing” in these cases rather than raise an exception.  Maybe we will want to.

nucypher/blockchain/eth/interfaces.py
vepkenez: Yes that should be gone.  As for client_version as long as it isn’t being accessed anywhere I don’t see why not.

nucypher/blockchain/eth/clients.py
vepkenez: Maybe the thing to do is write the access to these methods in a way that these methods will never be hit if the provider doesn’t support it.

derekpierre: There may be workarounds for those methods. Previously, I couldn't find an endpoint for `peers` but there was one for `unlock_account` - see `personal_unlockAccount` in https://github.com/trufflesuite/ganache-cli. It may not need to be implemented in this PR however - but something to think about.

@KPrasch ?

nucypher/blockchain/eth/clients.py
derekpierre: Ok - but wouldn't this currently produce an error due to the number of parameters provided based on Geth and Parity `clientVersion` strings?

nucypher/blockchain/eth/clients.py
derekpierre: It may be best to crash for now, until we figure out what to do. Purposely not hitting these methods could produce some ugly code 🤢

vepkenez: What I mean is right now I think peers is called in sync() somewhere and in the case of ganache there is no syncing, so we could just implement sync on these client classes and ganacheclient would never call its peers() so it wouldn’t matter.  But yeah in that case raising is fine

derekpierre: 👍  - yeah implementing `sync` in the clients themselves and having it called from `Blockchain.sync()` would be good. Assuming Ganache does not require syncing by nature then sure the `sync` method for the GanacheClient would do nothing.

I just didn't want `Blockchain.sync()` to have a bunch of `if` statements checking client type.

nucypher/blockchain/eth/clients.py
```diff
@@ -15,16 +16,109 @@
...
+        #
+        client_data = w3.clientVersion.split('/')
+        node_technology = client_data[0]
+
+        GETH = 'Geth'
+        PARITY = 'Parity'
+        GANACHE = 'EthereumJS TestRPC'
+
+        try:
+            gclass = {
```
KPrasch: `cls`?

nucypher/blockchain/eth/clients.py
KPrasch: Perhaps move this return statement into an `else` block (matching the `try`) in order to avoid executing the instantiation of `gclass` within the `try` block

nucypher/blockchain/eth/clients.py
```diff
@@ -15,16 +16,109 @@
...
+                GETH: GethClient,
+                PARITY: ParityClient,
+                GANACHE: GanacheClient,
+            }[node_technology]
+            return gclass(w3, *client_data)
+
+        except KeyError:
+            raise NotImplementedError(node_technology)
```
KPrasch: :+1:

nucypher/blockchain/eth/clients.py
KPrasch: Yeah - We'll need to get started on a corresponding mock web3 clients to see it in action.  Some of these init params can be optional with a constant default if needed

nucypher/blockchain/eth/clients.py
```diff
@@ -15,16 +16,109 @@
...
+    def peers(self):
+        return self.web3_instance.geth.admin.peers()
+
+    def unlock_account(self, address, password):
+        return self.web3_instance.geth.personal.unlockAccount(address, password)
+
+
+class ParityClient(Web3ClientBase):
+
+    def peers(self) -> str:
```
KPrasch: This returns a `str`?

nucypher/blockchain/eth/interfaces.py
```diff
@@ -155,6 +156,7 @@ def __init__(self,
         #
 
         self.w3 = NO_BLOCKCHAIN_CONNECTION
```
KPrasch: We might be able to condense this API further and remove this attribute entirely.

nucypher/blockchain/eth/interfaces.py
KPrasch: Yeah I agree - I think the can be forwarded to the web3 client abstraction at `self.client`

nucypher/blockchain/eth/interfaces.py
```diff
@@ -226,27 +223,13 @@ def connect(self):
         # Connect
         web3_instance = Web3(provider=self.__provider)  # Instantiate Web3 object with provider
         self.w3 = web3_instance
+        self.client = Web3Client(self.w3)
```
KPrasch: Perhaps this line at the web3 instantiation above can be combined into one line:

`self.client = Web3Client(Web3(provider=self.__provider))`

nucypher/blockchain/eth/interfaces.py
```diff
@@ -524,20 +520,11 @@ def call_backend_verify(self, pubkey: PublicKey, signature: Signature, msg_hash:
         return is_valid_sig and (sig_pubkey == pubkey)
 
     def unlock_account(self, address, password, duration=60*30):
-        # TODO: Parity
-        # TODO: Duration
 
         if 'tester' in self.provider_uri:
-            return True  # Test accounts are unlocked by default.
+            return True
```
KPrasch: Hey - I liked that comment :-)

nucypher/blockchain/eth/interfaces.py
```diff
@@ -524,20 +520,11 @@ def call_backend_verify(self, pubkey: PublicKey, signature: Signature, msg_hash:
...
 
-        elif self.client_version == 'geth':
-            return self.w3.geth.personal.unlockAccount(address, password, duration)
-
-        elif self.client_version == 'parity-ethereum':
-            return self.w3.parity.personal.unlockAccount(address, password, None)
-
-        else:
-            raise self.InterfaceError(f'{self.client_version} is not a supported ETH node backend.')
+        return self.client.unlock_account(address, password)
```
KPrasch: :boom:

nucypher/blockchain/eth/clients.py
```diff
@@ -15,16 +16,109 @@
...
+    pass
+
+
+class Web3ClientUnexpectedVersionString(Web3ClientError):
+    pass
+
+
+class Web3Client(object):
+
+    def __new__(self, w3: Web3, *args):
```
KPrasch: What is the advantage of using `__new__` over a `classmethod` in this situation?

I can imagine this same logic in `Web3Client.from_version_string`.  I do not feel strongly either way, but alternate -constructor `@classmethods` are popular in the nucypher codebase.

nucypher/blockchain/eth/clients.py
```diff
@@ -15,16 +16,109 @@
...
+
+        GETH = 'Geth'
+        PARITY = 'Parity'
+        GANACHE = 'EthereumJS TestRPC'
+
+        try:
+            gclass = {
+                GETH: GethClient,
+                PARITY: ParityClient,
+                GANACHE: GanacheClient,
```
KPrasch: We need to add support for PyEVM - or none of the blockchian tests will run.  The node technology string will be `'EthereumTester'`, I think.   see <https://circleci.com/gh/nucypher/nucypher/28751>

nucypher/blockchain/eth/chains.py
```diff
@@ -78,62 +75,27 @@ def interface(self) -> Union[BlockchainInterface, BlockchainDeployerInterface]:
...
-            # Continue until done
-            while self.syncing:
-                current = self.syncing['currentBlock']
-                total = self.syncing['highestBlock']
-                self.log.info(f"Syncing {current}/{total}")
-                time.sleep(1)
-                check_for_timeout()
-
-            return True
+        return self.interface.client.sync(timeout=timeout)
```
derekpierre: 👍 - nice!

nucypher/blockchain/eth/clients.py
```diff
@@ -54,69 +55,124 @@ def __new__(self, w3: Web3, *args):
...
                 GETH: GethClient,
                 PARITY: ParityClient,
                 GANACHE: GanacheClient,
             }[node_technology]
-            return gclass(w3, *client_data)
-
         except KeyError:
             raise NotImplementedError(node_technology)
 
+        return cls(w3, *client_data)
```
derekpierre: This is what I'm worried about failing with the 3 or 4 value split. Currently, I think this line will fail for Geth and Parity.

nucypher/blockchain/eth/clients.py
```diff
@@ -54,69 +55,124 @@ def __new__(self, w3: Web3, *args):
...
 
 class GanacheClient(Web3ClientBase):
 
-    def peers(self):
-        return []
-
     def unlock_account(self, address, password):
         return True
 
+    def sync(self, *args, **kwargs):
```
derekpierre: Esto es perfecto!

nucypher/blockchain/eth/clients.py
vepkenez: yep

Pipfile
```diff
@@ -60,6 +60,8 @@ sphinx_rtd_theme = "*"
 aafigure = "*"
 # CLI
 nucypher = {editable = true,path = "."}
+rope = "*"
```
KPrasch: Whoops!

Issue #1003, opened 2019-05-23
Re-design Ursula's identity evidence


cygnusv:
We have to come up with a new design of identity evidence that fulfills with, at least, the following requirements:
* Allows Bob (and other characters) to check that the Ursula node they want to validate belongs to a valid staker. This is currently not checked anywhere.
* Allows stakers to be smart contracts, which can't sign stuff. This is what #949 does, but we will probably need to rethink that approach.
----
**Current approach (also followed by #996):**
* Ursulas have a non-ETH signing key (_the stamp_), which is signed with the ETH key of the staker to produce her _identity evidence_. The stamp is used for signing network protocol messages. 
* Ursulas inherit the address associated to the staker's ETH key (_the checksum address_) and use it as their ID. Ursulas don't need to control the staker's ETH key.
* When Bob validates an Ursula node, he only checks that the signature is valid, but he doesn't check that the staker's account is valid (e.g., it can be a signature from an arbitrary ETH account, with no tokens at all).

**#949's approach:**
* Ursulas have a stamp and an ETH key (which results in a worker address)
* Stakers have an ETH account (with or without keys, since they can be contracts).
* `MinersEscrow` has a list of the mapping between Ursulas' addresses and stakers' addresses.
* This new scheme is not connected to the network-level identity information, so it's not consistent at all. The fact that this doesn't raise any problems is bad, since it shows that we weren't really validating anything with our current identity evidence approach.
-----
**New approach?**
We can mix both approaches in something like this:
* Ursula does not need an ETH address, only her stamp, like it is now.
* `MinersEscrow` has a list of the mapping between Ursulas' stamps and stakers addresses.
* Bob, Alice and other characters should always check with `MinersEscrow` that a node they have learnt about is actually connected to a staking miner. This means characters need access to an ETH node for this validation.
* The ETH signature as proof of identity is not useful anymore, as even if it's valid, you have to check `MinersEscrow`.
* Ursula can continue using the staker's address as her `checksum_public_address`, even if it's a contract.

michwill:
Hm... If Ursula changes her non-ETH signing key: does it mean that since the old non-ETH signing key was broadcasted in the past, that key can also be (maliciously) used?

jMyles:
I'm digesting this right now, but first things first: this is a beautiful issue and a model for design decision making going forward.

jMyles:
On the "new approach":

1.  Can you walk me through how a smart contract can stake?  If this happens, who performs the work?

[...]

PR #1004, opened 2019-05-24
[Do Not Merge] [WIP] BadSignature fix for Pete's sake

tuxxy:
Obviously, this is a work in progress.

This patch fixes the problem, but it isn't great because we don't have a way of knowing if we're running a test or not. During tests, we don't have a need to prefix `Ethereum Signed Message:\n`, but in production we do.

I have confirmed the fix works on isogeny, but there is no great way to test this. Eth signatures suck.


KPrasch:
This may become willnotfix depending on the outcome of #1003

PR #1012, opened 2019-05-26
Resolve Dependency Conflicts

vepkenez:


cygnusv:
What made you go back to pytest <4? We had to do that in the past for some reason but not anymore. Is it perhaps because of the warnings?

dev-requirements.txt
```diff
@@ -1,82 +1,86 @@
...
+pytest-cov==2.7.1
 pytest-forked==1.0.2
-pytest-mock==1.10.3
+pytest-mock==1.10.4
 pytest-mypy==0.3.2
 pytest-twisted==1.10
-pytest-xdist==1.28.0
-pytest==4.4.0
+pytest-xdist==1.27.0
+pytest==3.10.1
```
KPrasch: We'll take it for passing tests - but I'm left wondering why this pin is causing Issues

vepkenez: see this PR:  https://github.com/nucypher/py-solc/pulls

vepkenez: @cygnusv @KPrasch or this one
https://github.com/nucypher/py-solc/pull/6

I believe either of these will work, but if we want to use pytest4  we need to choose the 2nd one.

cygnusv: @vepkenez As far as I understand, `pytest` is a dev dependency of pysolc, so it shouldn't affect nucypher.

vepkenez: Both of those PRs fix this issue:
```
item = <Function test_compile_empty_folder>

    def pytest_runtest_setup(item):
>       if (item.get_marker('requires_standard_json') is not None and
            not solc_supports_standard_json_interface()):
E           AttributeError: 'Function' object has no attribute 'get_marker'

src/py-solc/tests/conftest.py:86: AttributeError
```

cygnusv: Yeah, I understand that, but isn't that a problem only for pysolc tests? It's not nucypher's problem so I don't see why it should affect nucypher's dependencies.

vepkenez: If we merge nucypher/py-solc#6  and update our requirements to pull that commit (because we are sourcing Nucypher's fork of py-solc), then we can put back pytest="*".

vepkenez: Ok so it appears that that error is actually coming from a py-solc test that is running because pytest is finding it.  I will look into why it is there.

PR #1016, opened 2019-05-27
Abstract/Cleanup BlockChainInterface

vepkenez:


KPrasch:
Follow-up fixes incoming in #1031

PR #1031, opened 2019-05-30
You Sign Me (Geth EIP-191 Support); Controller, CLI Bug Fixes

KPrasch:
- Implements geth client signing via `w3.eth.sign` in NuCypher client abstractions
- Verify wallet signature as part of standard node verification
- Labels are Unicode, not Base64
- Remove keccack of hexadecimal map ID in requests
- Handle internal geth process management in-test for signature integration
- Remove "pubkey_sig" from the codebase
- Use common deploy interactive input fixture


nucypher/blockchain/eth/clients.py
```diff
@@ -38,97 +38,121 @@ class Web3ClientUnexpectedVersionString(Web3ClientError):
...
+            raise ValueError(f"Invalid client version string. Got '{w3.clientVersion}'")
+
         except KeyError:
-            raise NotImplementedError(node_technology)
+            raise NotImplementedError(f'{w3.clientVersion} is not a supported ethereum client')
+
+        client_kwargs = {
+            'node_technology': node_technology,
+            'version': client_data[1],
+            'backend': client_data[-1],
```
derekpierre: Nice logic!

nucypher/blockchain/eth/clients.py
```diff
@@ -38,97 +38,121 @@ class Web3ClientUnexpectedVersionString(Web3ClientError):
...
     @property
-    def syncing(self):
+    def syncing(self) -> Union[bool, dict]:
         return self.w3.eth.syncing
 
-    def unlock_account(self, address, password):
-        raise NotImplementedError
-
-    def unlockAccount(self, address, password):
+    def unlock_account(self, address, password) -> bool:
```
vepkenez: I like this convention of using underscore_methods to make them obviously distinct from w3 calls

nucypher/characters/base.py
```diff
@@ -217,7 +227,7 @@ def __hash__(self):
         return int.from_bytes(bytes(self.stamp), byteorder="big")
 
     def __repr__(self):
-        r = "({})⇀{}↽ ({})"
+        r = self._display_name_template
```
vepkenez: 👍

nucypher/characters/control/interfaces.py
```diff
@@ -83,7 +83,7 @@ def create_policy(self,
         return response_data
 
     def derive_policy_encrypting_key(self, label: bytes) -> dict:
-        policy_encrypting_key = self.character.get_policy_pubkey_from_label(label)
+        policy_encrypting_key = self.character.get_policy_encrypting_key_from_label(label)
```
vepkenez: these name changes make it much clearer to me what is going on here!
💯

nucypher/cli/characters/alice.py
```diff
@@ -170,7 +169,7 @@ def alice(click_config,
         grant_request = {
             'bob_encrypting_key': bob_encrypting_key,
             'bob_verifying_key': bob_verifying_key,
-            'label': b64encode(label.encode()).decode(),
+            'label': label,
```
vepkenez: 🥂

nucypher/policy/models.py
```diff
@@ -490,7 +490,8 @@ def public_id(self):
         Ursula will refuse to propagate this if it she can't prove the payload is signed by Alice's public key,
         which is included in it,
         """
-        return keccak_digest(bytes(self._verifying_key) + bytes(self._hrac)).hex()
+        _id = keccak_digest(bytes(self._verifying_key) + bytes(self._hrac)).hex()
+        return _id
```
vepkenez: to new standards! 🥂

cygnusv:
@KPrasch I have an ugly workaround for the finnegan's wake issue. It's in here, in case you want to cherry-pick and have this PR with tests passing.
ed65dce3218e5a90306522cff8f1b1265749cabd

nucypher/characters/base.py
```diff
@@ -124,12 +123,23 @@ def __init__(self,
...
         self._checksum_address = checksum_public_address
+
+        # Fleet and Blockchain Connection (Everyone)
+        if not domains:
+            domains = (NodeConfiguration.DEFAULT_DOMAIN, )
+
+        # Needed for on-chain verification
+        if not self.federated_only:
+            self.blockchain = blockchain or Blockchain.connect()
+            self.miner_agent = MinerAgent(blockchain=blockchain)
```
cygnusv: This logic is somewhat weird, since all characters, even Enrico will get a miner agent. But for the moment I guess it's OK! 

Perhaps a TODO?

nucypher/characters/base.py
```diff
@@ -355,53 +365,68 @@ def verify_from(self,
         :param message_kit: the message to be (perhaps decrypted and) verified.
         :param signature: The signature to check.
         :param decrypt: Whether or not to decrypt the messages.
+        :param: label: A Label used for verifying the results of PRE
```
cygnusv:
```suggestion
        :param label: A label used for decrypting messages encrypted under its associated policy encrypting key
```

nucypher/characters/base.py
```diff
@@ -355,53 +365,68 @@ def verify_from(self,
...
         if signature_to_use:
-            is_valid = signature_to_use.verify(message, sender_pubkey_sig)
+            is_valid = signature_to_use.verify(message, sender_verifying_key)  # FIXME: Message is undefined here
             if not is_valid:
-                raise stranger.InvalidSignature(
-                    "Signature for message isn't valid: {}".format(signature_to_use))
+                raise InvalidSignature("Signature for message isn't valid: {}".format(signature_to_use))
         else:
-            raise self.InvalidSignature("No signature provided -- signature presumed invalid.")
+            raise InvalidSignature("No signature provided -- signature presumed invalid.")
```
cygnusv: Why are we changing here the type of `InvalidSignature`? Previous one was our own, and now it's the one from the cryptography, even if the `Signature` is ours.

nucypher/characters/lawful.py
```diff
@@ -292,16 +297,13 @@ def decrypt_message_kit(
...
-                label=label
-            )
-        )
+        cleartexts = [self.verify_from(
+            data_source,
+            message_kit,
+            signature=message_kit.signature,
+            decrypt=True,
+            label=label
+        )]
```
cygnusv: Why are we returning here a list of cleartexts, when `message_kit` always corresponds to a single message?
I know that `Bob.retrieve()` returns a list of cleartexts, but only because it's associated to `WorkOrders`, which may have more than one capsule (although currently they don't). 
This may not be an issue to solve here, but perhaps we should consider a better API for retrieve/decrypt methods: it doesn't make any sense to have methods that accept only a single capsule, but that output a list of cleartexts (of length 1). Either we make the input also a list, or we return the only cleartext.

nucypher/cli/characters/alice.py
cygnusv: 🥂 indeed

nucypher/network/server.py
```diff
@@ -359,18 +359,18 @@ def receive_treasure_map(treasure_map_id):
...
         else:
             do_store = treasure_map.public_id() == treasure_map_id
 
         if do_store:
             log.info("{} storing TreasureMap {}".format(this_node.stamp, treasure_map_id))
+
             # TODO 341 - what if we already have this TreasureMap?
-            this_node.treasure_maps[keccak_digest(binascii.unhexlify(treasure_map_id))] = treasure_map
+            treasure_map_index = bytes.fromhex(treasure_map_id)
+            this_node.treasure_maps[treasure_map_index] = treasure_map
```
cygnusv: 🎉

tests/fixtures.py
```diff
@@ -173,7 +174,7 @@ def alice_blockchain_test_config(blockchain_ursulas, testerchain):
                                 provider_uri=TEST_PROVIDER_URI,
                                 checksum_public_address=testerchain.alice_account,
                                 network_middleware=MockRestMiddleware(),
-                                known_nodes=blockchain_ursulas,
+                                known_nodes=blockchain_ursulas[:-1],  # TODO: 1035
```
cygnusv: `[:-1]` <-- nice guy!

tests/learning/test_fault_tolerance.py
```diff
@@ -61,26 +63,27 @@ def warning_trapper(event):
...
     Response = namedtuple("MockResponse", ("content", "status_code"))
     response = Response(content=crazy_bytes_representation, status_code=200)
     learner.network_middleware.get_nodes_via_rest = lambda *args, **kwargs: response
     learner.learn_from_teacher_node()
 
-    assert len(warnings) == 2
-    # TODO: Why no assert? Is this in progress?
-    warnings[1]['log_format'] == learner.unknown_version_message.format(new_node, new_node.TEACHER_VERSION,
-                                                                        learner.LEARNER_VERSION)
+    # TODO: Fails because the above mocked Response is unsigned, and the API now enforces interface signatures
```
cygnusv:
```suggestion
    # TODO: #1039 - Fails because the above mocked Response is unsigned, and the API now enforces interface signatures
```

nucypher/characters/lawful.py
michwill: That's true: some hackathon participants wondered about the same thing

nucypher/characters/base.py
KPrasch: https://github.com/nucypher/nucypher/issues/1041

nucypher/characters/base.py
KPrasch: See related PR #1040

PR #1040, opened 2019-06-03
Intermode node verification support

KPrasch:
Adds some careful handling of node verification and staking checks during intermode learning.

### DIFF
--- a/nucypher/crypto/api.py
+++ b/nucypher/crypto/api.py
@@ -22,7 +22,7 @@ from typing import Tuple
 import sha3
 from constant_sorrow import constants
 from cryptography import x509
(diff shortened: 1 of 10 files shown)


nucypher/crypto/api.py
```diff
@@ -22,7 +22,7 @@
 import sha3
 from constant_sorrow import constants
 from cryptography import x509
-from cryptography.exceptions import InvalidSignature  # TODO: Use nucypher exceptions
+from cryptography.exceptions import InvalidSignature
```
KPrasch: Im getting a circular dependency when trying to import from `nucypher.crypto.signing`  - Unsure how to proceed exactly

tuxxy: Hmmm, tough one. Can't import it in the spot you want to raise specifically?

PR #1029, opened 2019-05-29
[EPIC] Alhambra Verde - Introduce Staker and Worker Roles

KPrasch:
Chaperones: @cygnusv, @KPrasch 

This PR addresses the problems of having a sole entity for staking and working. The solution is to separate both roles, as sketched in #1003.

Issues to fix?
- Worker class #1054, #1033
- Re-design identity evidence #1003
- Staking Agent in Character? #1041
- Missing steps towards identity validation for slashing (Initiated in closed PR #996)

Pull requests:
- [X] #949 
- [x] #1020 
- [x] #1023 
- [x] #1046 
- [x] #1058 
- [x] #1066 
- [x] #1089 
- [x] #1093 

A word cloud of the new lines added by this PR (thanks @jMyles for the idea):
image

PR #1048, opened 2019-06-04
Decentralized Identity Verification Fixes and Naming

### LINKED ISSUE #1044: Revisit Teacher.stamp_is_valid() logic
Weird things that were merged too fast with #1040:
* What happens if `self.federated_only` and `self._identity_evidence is NOT_SIGNED`? Currently, the method simply finishes without returning a value (hence, returning `None`).
* The logic for decentralized validation is the following:
```python
            # Off-chain signature verification
            if self._stamp_has_valid_wallet_signature():
                self.verified_stamp = True
                return True

            # On-chain staking check
            if verify_staking:
                self._is_valid_worker()
```
It seems that the on-chain checks are never hit when `self._stamp_has_valid_wallet_signature()`, since it returns `True`. Or conversely, the only way to do the on-chain check is to fail the stamp's signature verification, which is wrong.

In any case, I'm working in #1033 in `alhambra-verde` and I'll fix this too.

KPrasch:
Fixes #1044, Needs follow up post #1029  as part of #1003

### DIFF
--- a/nucypher/characters/unlawful.py
+++ b/nucypher/characters/unlawful.py
@@ -61,7 +61,7 @@ class Vladimir(Ursula):
                        checksum_public_address = cls.fraud_address,
                        ######### Asshole.
                        timestamp=target_ursula._timestamp,
-                       interface_signature=target_ursula._interface_signature_object,
+                       interface_signature=target_ursula.__interface_signature,
                        ######### 
                        )
 
--- a/nucypher/characters/unlawful.py
+++ b/nucypher/characters/unlawful.py
@@ -58,10 +58,10 @@ class Vladimir(Ursula):
                        rest_port=target_ursula.rest_information()[0].port,
                        certificate=target_ursula.rest_server_certificate(),
                        network_middleware=cls.network_middleware,
-                       checksum_public_address = cls.fraud_address,
+                       checksum_public_address=cls.fraud_address,
                        ######### Asshole.
                        timestamp=target_ursula._timestamp,
-                       interface_signature=target_ursula.__interface_signature,
+                       interface_signature=target_ursula._Teacher__interface_signature,
                        ######### 
                        )
 
--- a/nucypher/characters/unlawful.py
+++ b/nucypher/characters/unlawful.py
@@ -61,7 +61,7 @@ class Vladimir(Ursula):
                        checksum_public_address=cls.fraud_address,
                        ######### Asshole.
                        timestamp=target_ursula._timestamp,
-                       interface_signature=target_ursula._Teacher__interface_signature,
+                       interface_signature=target_ursula._interface_signature,
                        ######### 
                        )
 
--- a/nucypher/crypto/api.py
+++ b/nucypher/crypto/api.py
@@ -84,29 +84,29 @@ def keccak_digest(*messages: bytes) -> bytes:
     :rtype: bytes
     :return: bytestring of digested data
     """
-    hash = sha3.keccak_256()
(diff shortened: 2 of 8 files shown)


nucypher/crypto/api.py
```diff
@@ -118,24 +118,24 @@ def verify_eip_191(address: object, message: object, signature: object) -> objec
     if signature_is_valid:
         return True
     else:
-        raise InvalidSignature
+        return False
```
fjarri: Why not just `return signature_is_valid`?

nucypher/network/nodes.py
```diff
@@ -1021,56 +1077,61 @@ def verify_node(self,
...
         signature = blockchain_power.sign_message(bytes(self.stamp))
-        self._identity_evidence = signature
+        self.__decentralized_identity_evidence = signature
 
     #
     # Interface
     #
 
-    def interface_is_valid(self) -> bool:
+    def validate_interface(self):
```
cygnusv:
```suggestion
    def validate_interface(self) -> bool:
```

nucypher/characters/unlawful.py
```diff
@@ -58,10 +58,10 @@ def from_target_ursula(cls,
                        rest_port=target_ursula.rest_information()[0].port,
                        certificate=target_ursula.rest_server_certificate(),
                        network_middleware=cls.network_middleware,
-                       checksum_public_address = cls.fraud_address,
+                       checksum_public_address=cls.fraud_address,
                        ######### Asshole.
                        timestamp=target_ursula._timestamp,
-                       interface_signature=target_ursula._interface_signature_object,
+                       interface_signature=target_ursula._Teacher__interface_signature,
```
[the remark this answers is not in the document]
KPrasch: I did not introduce the property, but since it's there, I'm trying to enforce it.

nucypher/network/nodes.py
KPrasch: :heavy_check_mark:

nucypher/characters/unlawful.py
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/network/nodes.py
```diff
@@ -923,79 +980,78 @@ def update_snapshot(self, checksum, updated, number_of_known_nodes):
...
 
         if self.federated_only:
-            if self._identity_evidence is not NOT_SIGNED:
+            if self.__decentralized_identity_evidence is not NOT_SIGNED:
                 message = "This node cannot be verified in this manner, " \
                           "but is OK to use in federated mode if you "    \
                           "have reason to believe it is trustworthy."
                 raise self.WrongMode(message)
+            else:
+                # TODO: Verify the stamp "correctness" by serialization?
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/network/nodes.py
```diff
@@ -854,38 +871,63 @@ def __init__(self,
...
+        #
+
+        self._timestamp = timestamp
+        self.certificate = certificate
+        self.certificate_filepath = certificate_filepath
+        self.__interface_signature = interface_signature
+        self.__decentralized_identity_evidence = constant_or_bytes(decentralized_identity_evidence)
+
+        # Assume unverified
+        self.verified_stamp = False
```
jMyles: Why move these settings to `__init__`?  Doesn't this pattern just put more bloat in `__init__`?

PR #1058, opened 2019-06-07
Extract Worker actor from Staker

### LINKED ISSUE #800: Find a better metaphor than "miner"
I think this has been eloquently described by @jMyles:

> (...) it is more clear to me than ever that `Miner` is a broken metaphor.  I have never liked this name, but reading this, I am given pause that there is a number mismatch between the work types for a given node and the practices surrounding this work types.  For example, the inflation economics apply to a "Staker" or "Worker" (half of what we currently call `Miner`).  Since this is a work token, `Worker` makes sense to me.  However, the _work_ being performed pursuant to this work token, as it relates to the adjudication, is not of only one kind.  Presently, we expect `Ursula` to perform PRE and understand that she'll be slashed if she fails to perform.  Let's imagine a future in which we task a different character to perform SSS or NuFHE under the auspices of the same work token.  Is this person still a `Miner`?  It seems to me that `Worker` is a concept that might _utilize_ `Ursulas` and SSS/NuFHE-characters and is not properly understood as a "mining mixin."  Normally, I am not terribly concerned about these sorts of future-proofing exercises, but I think that we'll have crossed an ~epitaph~ rubicon if we deploy these smart contracts with the word "Miner" - it will be hard to educate our audience on what we really mean.

### LINKED ISSUE #962: How should Bob discover and validate Ursula's Identity Evidence?
_'Identity evidence'_ (sometimes called _'Evidence of decentralized identity'_) is a signature of Ursula's stamp with the checksum address.

Perhaps I'm too slow today, but I can't figure out what's the best way for Bob to discover this information in a way that gives him assurance that it's valid and compatible with the information contained in `MinersEscrow` contract.

Bob needs this information in order to request a CFrag evaluation and Ursula's slashing.

### LINKED ISSUE #1003: Re-design Ursula's identity evidence

### LINKED ISSUE #1033: Check valid staker->worker relation in Learner._is_valid_worker() against StakingEscrow
This can be done by checking `StakingEscrow.workerToStaker(worker_address) == staker_address` or `StakingEscrow.stakerInfo[staker_address].worker == worker_address`.

The former seems cheaper.

### LINKED ISSUE #1054: Extract Worker actor from current Staker
With the introduction of a Worker role in our smart contracts as part of the Alhambra Verde EPIC, we need to separate our `Staker` actor class in two parts: The real `Staker`, and the `Worker`. 

For starters, most (all?) methods in `Staker` marked with the `@only_me` decorator seems like  suitable candidates to remain in `Staker`. However, it seems reasonable to keep some of the "read-only" methods in `Staker` for `Worker` (i.e., `get_locked_tokens`).

Perhaps a "stranger" Staker (i.e., `is_me=False`) can be part of a `Worker`?

cygnusv:
The goal of this PR is to crystallize the implications of #1003, i.e., the separation of Ursula in two: Worker and Staker. Now Ursulas are worker nodes, while NU holders are stakers (either of human or contract nature).

Most drastic changes occur at the actor level (since now the old `Staker` actor was separated into `Staker` and `Worker`) and at the `Teacher` level, since now teacher nodes (which inherit from `Worker`) have to implement the new ways of identity validation (as per #1003), which include off-chain and on-chain means.

Side note: this was a fun PR. @KPrasch and I were working on it in turns, taking advantage of our almost complementary time zones. There wasn't too much need to talk. Just wake up in the morning, drink some coffee, `git remote update` and continue from where the other left. 

- Closes #1003 
- Closes #1033 
- Closes #1054
- Closes #962 
- Closes #800


nucypher/blockchain/eth/actors.py
```diff
@@ -686,11 +645,155 @@ def collect_policy_reward(self, collector_address=None, policy_agent: PolicyAgen
...
         self._transaction_cache.append((datetime.utcnow(), collection_txhash))
         return collection_txhash
 
 
+class Worker(NucypherTokenActor):
+    """
+    Ursula baseclass for blockchain operations, practically carrying a pickaxe.
+    """
+
+    __current_period_sample_rate = 60*60  # seconds
```
KPrasch: Perhaps - Let's change this is be based on a `TokenEconomics` instance?

tests/characters/test_ursula_prepares_to_act_as_mining_node.py
```diff
@@ -124,3 +137,34 @@ def test_vladimir_uses_his_own_signing_key(blockchain_alice, blockchain_ursulas)
     # However, the actual handshake proves him wrong.
     with pytest.raises(vladimir.InvalidNode):
         vladimir.verify_node(blockchain_alice.network_middleware, certificate_filepath="doesn't matter")
+
+
+# TODO: Change name of this file
+def test_blockchain_ursulas_reencrypt(blockchain_ursulas, blockchain_alice, blockchain_bob, policy_value):
```
KPrasch: :boom:

KPrasch:
Follow up to `@transaction` agency API - https://github.com/nucypher/nucypher/issues/1086

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -357,25 +357,33 @@ contract StakingEscrow is Issuer {
...
+            // Remove relation between the old worker and the staker, if there was one
+            if (info.worker != address(0)) {
+                workerToStaker[info.worker] = address(0);
+            }
+            info.worker = _worker;
+            info.workerStartPeriod = currentPeriod;
+            // Set new worker->staker relation
+            workerToStaker[_worker] = msg.sender;
+        } else { // Staker wants to unset her worker
+            // In this case, we don't require a minimum elapsed time with the worker
```
szotov: When staker changes worker - he potentially could avoid punishment. I think only setting from zero address to non-zero can be without checking period

Issue #842, opened 2019-03-15
Configurable gas limits


KPrasch:
In tests, we set the gas price to 0 for calculation accuracy checks, then separately execute gas estimation metrics. 

We need:
- Integration tests against ethereum clients using default gas price strategy
- Hot-swapable gas expenditure per-transaction (think metamask integration)
- Configurable per-method default gas
- Configurable minimum and maximum gas

cygnusv:
Consider also how to handle `gas_limit` in multi-transaction operations (e.g. deploying StakingEscrow)
See https://github.com/nucypher/nucypher/issues/1113

KPrasch:
Closed in #1750

KPrasch:
Advanced by #2445

cygnusv:
I think we can close it, as we have a fairly reduced surface of APIs that expose TXs now. We can reopen if the need arises again.

PR #1064, opened 2019-06-08
Introduce Stakeholder

KPrasch:
PR goes into epic.  Introduces new Issue https://github.com/nucypher/nucypher/issues/1119

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -32,7 +32,8 @@ from constant_sorrow.constants import (
     NO_FUNDING_ACCOUNT
 )
 from eth_tester.exceptions import TransactionFailed
-from eth_utils import keccak, is_checksum_address
+from eth_utils import is_checksum_address
+from eth_utils import keccak
 from twisted.logger import Logger
 from web3 import Web3
 
@@ -724,8 +725,20 @@ class StakeHolder(BaseConfiguration):
 
         self.__get_accounts()
 
-        if sync_now:
+        if sync_now and not offline_mode:
             self.read_onchain_stakes()  # Stakes
+
+    def __str__(self) -> str:
+        r = f'{self.__class__.__name__}(funding_account={self.funding_account})'
+        return r
+
+    def __repr__(self) -> str:
+        r = f'{self.__class__.__name__}(funding_account={self.funding_account})'
+        return r
+
+    #
+    # Configuration
+    #
 
     def static_payload(self) -> dict:
         """Values to read/write from stakeholder JSON configuration files"""
@@ -756,6 +769,18 @@ class StakeHolder(BaseConfiguration):
             transacting_power = TransactingPower(blockchain=self.blockchain, account=checksum_address)
             self.__transacting_powers[checksum_address] = transacting_power
         transacting_power.activate(password=password)
+
+    def to_configuration_file(self, *args, **kwargs) -> str:
+        filepath = super().to_configuration_file(modifier=self.funding_account, *args, **kwargs)
+        return filepath
+
+    def connect(self, blockchain: BlockchainInterface = None) -> None:
+        """Go Online"""
+        if not self.staking_agent:
+            self.staking_agent = StakingEscrowAgent(blockchain=blockchain)
+        if not self.token_agent:
+            self.token_agent = NucypherTokenAgent(blockchain=blockchain)
+        self.blockchain = self.token_agent.blockchain
 
     #
     # Account Utilities
@@ -916,7 +941,9 @@ class StakeHolder(BaseConfiguration):
         # Send new staker account NU
         _result = self.token_agent.transfer(amount=amount.to_nunits(),
                                             sender_address=self.funding_account,
-                                            target_address=staker.checksum_address)
+                                            target_address=staker.checksum_address,
+                                            auto_approve=True)
+        return staker
 
     def initialize_stake(self,
                          amount: NU,
@@ -966,6 +993,13 @@ class StakeHolder(BaseConfiguration):
             receipts[staker.checksum_address] = receipt
         self.to_configuration_file(override=True)
         return receipts
+
+    def calculate_rewards(self) -> dict:
+        rewards = dict()
+        for staker in self.stakers:
+            reward = staker.calculate_reward()
+            rewards[staker.checksum_address] = reward
+        return rewards
 
     def collect_rewards(self,
                         staker_address: str,
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -145,8 +145,6 @@ class Deployer(NucypherTokenActor):
 
     def __init__(self,
                  blockchain: BlockchainDeployerInterface,
-                 client_password: str = None,
-                 device = NO_STAKING_DEVICE,
                  deployer_address: str = None,
                  client_password: str = None,
                  bare: bool = True
@@ -174,7 +172,6 @@ class Deployer(NucypherTokenActor):
 
         # TODO: Does this class want to be a Character implementing PowerUp consumption?
         self.transacting_power = TransactingPower(blockchain=blockchain,
-                                                  device=device,
                                                   password=client_password,
                                                   account=deployer_address)
         self.transacting_power.activate()
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -753,12 +753,14 @@ class StakeHolder(BaseConfiguration):
     def from_configuration_file(cls, filepath: str = None, **overrides) -> 'StakeHolder':
         filepath = filepath or cls.default_filepath()
         payload = cls._read_configuration_file(filepath=filepath)
+
+        # Sub config
         blockchain_payload = payload.pop('blockchain')
         blockchain = BlockchainInterface.from_dict(payload=blockchain_payload)
-        instance = cls(filepath=filepath,
-                       blockchain=blockchain,
-                       **payload, **overrides)
+        payload.update(dict(blockchain=blockchain))
 
+        payload.update(overrides)
+        instance = cls(filepath=filepath, **payload)
         return instance
 
     @validate_checksum_address
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -53,6 +53,7 @@ from nucypher.blockchain.eth.interfaces import BlockchainInterface
 from nucypher.blockchain.eth.registry import AllocationRegistry
 from nucypher.blockchain.eth.token import NU, Stake, StakeTracker
 from nucypher.blockchain.eth.utils import datetime_to_period, calculate_period_duration
+from nucypher.config.base import BaseConfiguration
 from nucypher.config.constants import DEFAULT_CONFIG_ROOT
 from nucypher.crypto.powers import TransactingPower
 
@@ -656,3 +657,83 @@ class Investigator(NucypherTokenActor):
         receipt = self.adjudicator_agent.evaluate_cfrag(evidence=evidence,
                                                         sender_address=self.checksum_address)
         return receipt
+
+
+class StakeHolder(BaseConfiguration):
+
+    def __init__(self,
+                 blockchain: BlockchainInterface = None,
+                 staking_agent: StakingEscrowAgent = None,
+                 trezor: bool = False,
+                 *args, **kwargs):
+
+        if not staking_agent:
+            staking_agent = StakingEscrowAgent(blockchain=blockchain)
+
+        self.staking_agent = staking_agent
+        self.blockchain = staking_agent.blockchain
+
+        self.trezor = trezor
+        self.device = NO_STAKING_DEVICE
+        self.setup_device()
+
+        self.__accounts = list()
+        self.__stakers = dict()
+
+        super().__init__(*args, **kwargs)
+
+    def static_payload(self) -> dict:
+        payload = dict(trezor=self.trezor)
+        return payload
+
+    def setup_device(self) -> None:
+        device = NO_STAKING_DEVICE
+        if self.trezor:
+            device = NotImplemented  # Trezor()
+        self.device = device
+
+    def add_account(self, address: str) -> None:
+        self.blockchain.interface.client.unlock_account(address=address)
+        self.__accounts.append(address)
+
+    def get_accounts(self):
+        if self.device is not NO_STAKING_DEVICE:
+            self.__accounts.extend(self.device.accounts)
+
+    def get_stakers(self) -> None:
+        for account in self.__accounts:
+            staker = Staker(is_me=True, checksum_address=account)
+            self.__stakers[account] = staker
+
+    def get_staker(self, address: str) -> Staker:
+        try:
+            staker = self.__stakers[address]
+        except KeyError:
+            if address not in self.__accounts:
+                raise RuntimeError(f"Unknown or locked account {address}")
+            staker = Staker(is_me=True, checksum_address=address)
+        return staker
+
+    def set_worker(self, index: int, worker_address: str):
+        stake = self.__stakers[index]
+        result = self.staking_agent.set_worker(staker_address=stake.owner_address, worker_address=worker_address)
+        return result
+
+    def initialize_stake(self, address: str, amount: NU, duration: int):
+        staker = self.get_staker(address=address)
+        new_stake = Stake.initialize_stake(staker=staker, amount=amount)
+        return new_stake
+
+    def divide_stake(self, address: str, index: int):
+
+        try:
+            staker = self.__stakers[address]
+        except KeyError:
+            raise RuntimeError(f"No active stakes for {address}")
+
+        try:
+            stake = staker.stakes[index]
+        except IndexError:
+            raise
+
+        stake.divide(target_value=0)
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -32,8 +32,9 @@ from constant_sorrow.constants import (
     NO_FUNDING_ACCOUNT
 )
 from eth_tester.exceptions import TransactionFailed
-from eth_utils import keccak
+from eth_utils import keccak, is_checksum_address
 from twisted.logger import Logger
+from web3 import Web3
 
 from nucypher.blockchain.economics import TokenEconomics
 from nucypher.blockchain.eth.agents import (
@@ -143,7 +144,9 @@ class Deployer(NucypherTokenActor):
         pass
 
     def __init__(self,
-                 blockchain: BlockchainInterface,
+                 blockchain: BlockchainDeployerInterface,
+                 client_password: str = None,
+                 device = NO_STAKING_DEVICE,
                  deployer_address: str = None,
                  client_password: str = None,
                  bare: bool = True
@@ -168,6 +171,13 @@ class Deployer(NucypherTokenActor):
                                                         password=client_password)
         blockchain.transacting_power.activate()
         self.log = Logger("Deployment-Actor")
+
+        # TODO: Does this class want to be a Character implementing PowerUp consumption?
+        self.transacting_power = TransactingPower(blockchain=blockchain,
+                                                  device=device,
+                                                  password=client_password,
+                                                  account=deployer_address)
+        self.transacting_power.activate()
 
     def __repr__(self):
         r = '{name}({blockchain}, {deployer_address})'.format(name=self.__class__.__name__,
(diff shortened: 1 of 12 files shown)


nucypher/blockchain/eth/token.py
```diff
@@ -163,7 +170,7 @@ def __init__(self,
 
         # Agency
         self.staking_agent = None
-        self.token_agent = NucypherTokenAgent()
+        self.token_agent = NucypherTokenAgent()  # TODO: Use Agency
```
cygnusv: Do you really need to use the Agency? Isn't this way good enough as the objects are supposed to be singletons automagically, due to Agency being the metaclass?

KPrasch: The function signature of agent requires the `blockchain` parameter, so except for the *first* instantiation, it can be misleading to do:
- `StakingEscrowAgent(blockchain=blockchain)` (will not use this blockchain) 
- `StakingEscrowAgent()` (missing the required arg) 

As an alternate API - I've started using `Agency.get_agent(NucypherTokenAgent)` - perhaps the blockchain instance used for agent instances can be better managed there, too. 

What do you think about this pattern?

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +676,349 @@ def request_evaluation(self, evidence):
...
+        payload = list()
+        for staker in self.stakers:
+            stake_info = [stake.to_stake_info() for stake in staker.stakes]
+            worker_address = staker.worker_address or NO_WORKER_ASSIGNED
+            staker_funds = {'ETH': int(staker.eth_balance), 'NU': int(staker.token_balance)}
+            staker_payload = {'staker': staker.checksum_address,
+                              'balances': staker_funds,
+                              'worker': worker_address,
+                              'stakes': stake_info}
+            payload.append(staker_payload)
```
derekpierre: It could be cleaner to have the `Staker` class return the payload representation based on its own properties, and encompasses all the data you are using above. This could reduce the code here to something like:
```python
payload = list()
for staker in self.stakers:
    payload.append(staker.get_payload())

return payload
```

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +676,349 @@ def request_evaluation(self, evidence):
...
+                                               worker_address=worker_address)
+
+        self.to_configuration_file(override=True)
+        return result
+
+    def __prepare_new_staker(self, staker: Staker, amount: NU):
+        """Creates a new ethereum account, then transfers it ETH and NU for staking."""
+        if self.funding_tokens < amount:
+            delta = amount - staker.token_balance
+            raise self.ConfigurationError(f"{self.funding_account} Insufficient NU (need {delta} more).")
```
derekpierre:
```suggestion
            raise self.ConfigurationError(f"{self.funding_account} Insufficient NU (need {delta} more to transfer {amount}).")
```

nucypher/blockchain/eth/interfaces.py
```diff
@@ -304,7 +318,9 @@ def send_transaction(self,
         payload.update({'chainId': int(self.client.chain_id),
                         'nonce': nonce,
                         'from': sender_address,
-                        'gasPrice': self.client.w3.eth.gasPrice})
+                        'gasPrice': self.client.w3.eth.gasPrice,
```
derekpierre:
```suggestion
                        'gasPrice': self.client.gas_price,
```

tests/blockchain/eth/entities/actors/test_stakeholder.py
```diff
@@ -0,0 +1,219 @@
...
+    original_stake, new_stake = staking_software_stakeholder.divide_stake(address=stake.owner_address,
+                                                                          password=INSECURE_DEVELOPMENT_PASSWORD,
+                                                                          index=0,
+                                                                          duration=10,
+                                                                          value=target_value)
+
+    staking_agent = Agency.get_agent(StakingEscrowAgent)
+    stakes = list(staking_agent.get_all_stakes(staker_address=stake.owner_address))
+    assert len(stakes) == 2
+    assert new_stake.value == target_value
```
derekpierre: Is it possible to also check the value of `original_stake` after the divide. Something akin to (?):
```python
total_stake_value = staking_software_stakeholder.total_stake()
# ... do the divide
assert original_stake.value == (total_stake_value - target_value)
```

nucypher/blockchain/eth/clients.py
```diff
@@ -266,8 +270,22 @@ def peers(self):
         return self.w3.geth.admin.peers()
 
     def unlock_account(self, address, password):
+        # if self.is_local:
+        #     return True
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/interfaces.py
KPrasch: :heavy_check_mark:

tests/blockchain/eth/entities/actors/test_stakeholder.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/actors.py
```diff
@@ -483,6 +497,13 @@ def set_worker(self, worker_address: str) -> str:
                                                 worker_address=worker_address)
         return receipt
 
+    @property
+    def worker_address(self) -> str:
+        worker_address = self.staking_agent.get_worker_from_staker(staker_address=self.checksum_address)
+        if worker_address is BlockchainInterface.NULL_ADDRESS:
+            return NO_WORKER_ASSIGNED
+        return worker_address
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +696,342 @@ def request_evaluation(self, evidence):
...
+        if account:
+            accounts = [account]
+        else:
+            accounts = self.__accounts
+
+        for account in accounts:
+            stakes = list(self.staking_agent.get_all_stakes(staker_address=account))
+            if stakes:
+                staker = Staker(is_me=True, checksum_address=account, blockchain=self.blockchain)
+                self.__stakers[staker] = staker.stakes
```
cygnusv: You're using `Staker` as dict key here, but as far as I can tell, this class is not Hashable. Can this be problematic somehow?

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +696,342 @@ def request_evaluation(self, evidence):
...
+                     'NU': self.token_agent.get_balance(account)}
+            balances.update({account: funds})
+        return balances
+
+    @property
+    def staker_balances(self) -> dict:
+        balances = dict()
+        for staker in self.stakers:
+            staker_funds = {'ETH': staker.eth_balance, 'NU': staker.token_balance}
+            balances = {staker.checksum_address: staker_funds}
```
cygnusv:
```suggestion
            balances[staker.checksum_address] = staker_funds
```

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +696,342 @@ def request_evaluation(self, evidence):
...
+        if self.funding_power.device:
+            # TODO: More formal check for wallet here?
+            # With devices, the last account is is always an unsed one.
+            new_account = self.blockchain.client.accounts[-1]
+        else:
+            if not password:
+                raise self.ConfigurationError("No password supplied to create new staking account.")
+            new_account = self.blockchain.client.new_account(password=password)
+        self.__accounts.append(new_account)
+        staker = Staker(blockchain=self.blockchain, is_me=True, checksum_address=new_account)
```
cygnusv: Shouldn't this staker be added to `self.__stakers` too?

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +696,342 @@ def request_evaluation(self, evidence):
...
+    def create_worker_configuration(self, worker_address: str, password: str, **configuration) -> str:
+        """Generates a worker JSON configuration file for a given staking address."""
+        from nucypher.config.characters import UrsulaConfiguration
+        worker_configuration = UrsulaConfiguration.generate(worker_address=worker_address,
+                                                            password=password,
+                                                            config_root=self.config_root,
+                                                            federated_only=False,
+                                                            provider_uri=self.blockchain.provider_uri,
+                                                            **configuration)
+        return worker_configuration
```
cygnusv: :heart_eyes:

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +696,342 @@ def request_evaluation(self, evidence):
...
+            message = f"{self.funding_account} Insufficient NU (need {delta} more to transfer a total of {amount})."
+            raise self.ConfigurationError(message)
+        if not self.funding_eth:
+            raise self.ConfigurationError(f"{self.funding_account} has no ETH")
+
+        # Transfer NU
+        self.funding_power.activate()
+
+        # Send new staker account ETH
+        tx = {'to': staker.checksum_address, 'from': self.funding_account, 'value': self.eth_funding}
```
cygnusv: Why does it have to send a default amount of ETH?

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +696,342 @@ def request_evaluation(self, evidence):
...
+                                     target_value=value)
+        self.to_configuration_file(override=True)
+        return result
+
+    def sweep(self) -> dict:
+        """Collect all rewards from all staking accounts"""
+        receipts = dict()
+        for staker in self.stakers:
+            self.attach_transacting_power(checksum_address=staker.checksum_address)
+            receipt = self.collect_rewards(staker_address=staker.checksum_address)
```
cygnusv: No password is passed to `self.attach_transacting_power()` and to `self.collect_rewards()`, which means this assumes that the transacting power is already cached with a password; otherwise, this won't work. Perhaps you can pass a list of passwords as input to `sweep()`? (I know I know.....yuck... 🤢 )

nucypher/blockchain/eth/actors.py
KPrasch: The new staker account needs to perform transactions, and pay for gas.

nucypher/blockchain/eth/actors.py
KPrasch: Sweep is deprecated (for now) in `68426a32076f60be232a5b50b458f5da56d06f31`

nucypher/blockchain/eth/actors.py
KPrasch: Yes. Fixed here: `68426a32076f60be232a5b50b458f5da56d06f31`

nucypher/blockchain/eth/actors.py
KPrasch: Also fixed in `68426a32076f60be232a5b50b458f5da56d06f31`

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/clients.py
```diff
@@ -323,6 +326,10 @@ def unlock_account(self, address, password) -> bool:
     def sync(self, *args, **kwargs):
         return True
 
+    def new_account(self, password: str):
+        insecure_account = self.w3.provider.ethereum_tester.add_account(os.urandom(32).hex(), password=password)
```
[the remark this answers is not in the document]
KPrasch: :check

nucypher/blockchain/eth/actors.py
```diff
@@ -476,10 +505,21 @@ def initialize_stake(self,
...
+    @property
+    def worker_address(self) -> str:
+        if self.__worker_address:
+            return self.__worker_address
+        else:
+            worker_address = self.staking_agent.get_worker_from_staker(staker_address=self.checksum_address)
+            self.__worker_address = worker_address
+            if self.__worker_address is BlockchainInterface.NULL_ADDRESS:
+                return NO_WORKER_ASSIGNED
+            return self.__worker_address
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/actors.py
```diff
@@ -164,6 +171,12 @@ def __init__(self,
         blockchain.transacting_power.activate()
         self.log = Logger("Deployment-Actor")
 
+        # TODO: Does this class want to be a Character implementing PowerUp consumption?
```
vepkenez: Would would it mean to make this change?  Moving this class into either chaotic.py or lawful.py (and deciding which one)?

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +697,342 @@ def request_evaluation(self, evidence):
...
+                 offline_mode: bool = False,
+                 sync_now: bool = True,
+                 *args, **kwargs):
+
+        super().__init__(*args, **kwargs)
+
+        self.log = Logger(f"stakeholder")
+        self.blockchain = blockchain
+        self.offline_mode = offline_mode
+        self.eth_funding = Web3.toWei('0.1', 'ether')  # TODO: How much ETH to use while funding new stakes.
```
vepkenez: this should be a constant/cli arg/envvar right?

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +697,342 @@ def request_evaluation(self, evidence):
...
+        """Values to read/write from stakeholder JSON configuration files"""
+        payload = dict(provider_uri=self.blockchain.provider_uri,
+                       blockchain=self.blockchain.to_dict(),
+                       funding_account=self.funding_account,
+                       accounts=self.__accounts,
+                       stakers=self.__serialize_stakers())
+        return payload
+
+    @classmethod
+    def from_configuration_file(cls, filepath: str = None, **overrides) -> 'StakeHolder':
```
vepkenez: we already have this method in node.py don't we?  I guess this is how we know if it should be a Character?

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +697,342 @@ def request_evaluation(self, evidence):
...
+        # Sub config
+        blockchain_payload = payload.pop('blockchain')
+        blockchain = BlockchainInterface.from_dict(payload=blockchain_payload)
+        payload.update(dict(blockchain=blockchain))
+
+        payload.update(overrides)
+        instance = cls(filepath=filepath, **payload)
+        return instance
+
+    @validate_checksum_address
```
vepkenez: 👍

nucypher/blockchain/eth/actors.py
KPrasch: Yeah - I think into `lawful.py` - However, I'm not fully convinced.

nucypher/blockchain/eth/actors.py
KPrasch: Yes to all. Not sure what makes sense as a default.  Perhaps an amount based on the duration or another way to estimate future gas consumption?

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +697,342 @@ def request_evaluation(self, evidence):
...
+    @property
+    def accounts(self) -> list:
+        return self.__accounts
+
+    def __get_accounts(self) -> None:
+        accounts = self.blockchain.client.accounts
+        self.__accounts.extend(accounts)
+
+    def set_funding_account(self, address: str):
+        self.__funding_account = address
```
vepkenez: these getters and setters are a little...  why have them instead of
`self.funding_account = <whatever>`

nucypher/blockchain/eth/actors.py
KPrasch: It's one way for sure, but not the most defining characteristic (no pun indented) is the usage of `TransactingPower` in the `StakeHolder.__init__`, and the analogous layers to that of `Worker -> Ursula` & `Staker -> StakeHolder`

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +697,342 @@ def request_evaluation(self, evidence):
...
+        """Creates a new ethereum account, then transfers it ETH and NU for staking."""
+        if self.funding_tokens < amount:
+            delta = amount - staker.token_balance
+            message = f"{self.funding_account} Insufficient NU (need {delta} more to transfer a total of {amount})."
+            raise self.ConfigurationError(message)
+        if not self.funding_eth:
+            raise self.ConfigurationError(f"{self.funding_account} has no ETH")
+
+        # Transfer NU
+        self.funding_power.activate()
```
vepkenez: didn't we already activate this in the __init__?

nucypher/blockchain/eth/actors.py
KPrasch: I want to perform checksum address validation on setter here.

nucypher/blockchain/eth/actors.py
KPrasch: Yes, however, its possible through normal usage to switch `TransactingPower`s with one belonging to another account, like a `Staker`.  We might be able to remove this with some careful inspection, but this is ensuing that we indeed have the funding power attached.

nucypher/blockchain/eth/actors.py
```diff
@@ -656,3 +697,342 @@ def request_evaluation(self, evidence):
...
+        blockchain = BlockchainInterface.from_dict(payload=blockchain_payload)
+        payload.update(dict(blockchain=blockchain))
+
+        payload.update(overrides)
+        instance = cls(filepath=filepath, **payload)
+        return instance
+
+    @validate_checksum_address
+    def attach_transacting_power(self, checksum_address: str, password: str = None) -> None:
+        try:
```
vepkenez: I feel like the right way to do this is more like
```
if checksum_addresss in self.__transacting_powers:
    transacting_power = self.__transacting_powers[checksum_address]
else:
  transacting_power = TransactingPower(blockchain=self.blockchain, account=checksum_address)
  self.__transacting_powers[checksum_address] = transacting_power
transacting_power.activate(password=password)
```

I guess it's an ongoing debate...
https://softwareengineering.stackexchange.com/questions/351110/are-exceptions-for-flow-control-best-practice-in-python

KPrasch: Heh - Yeah I even say that I prefer the exception handling as flow control, from the old idiom "Don't look before you leap"

nucypher/blockchain/eth/actors.py
vepkenez: to me, these little fixes are a reason not to do the whole
```
@property
def thing(self)

set_thing(self, thing)

get_thing(self, thing)

self.__thing
```

just another bug to have

nucypher/blockchain/eth/agents.py
```diff
@@ -44,6 +47,13 @@ def clear(mcs):
     def agents(cls):
         return cls.__agents
 
+    @classmethod
+    def get_agent(mcs, cls):
+        try:
```
vepkenez: what is an "mcs"?

nucypher/blockchain/eth/actors.py
KPrasch: Heh - Just generally prefer statelessness?

nucypher/blockchain/eth/agents.py
KPrasch: it's a conventional first-arg name for a meta-classmethod.

nucypher/blockchain/eth/clients.py
```diff
@@ -197,20 +197,20 @@ def coinbase(self) -> str:
         return self.w3.eth.coinbase
```
vepkenez: I don't see any tests for a lot of the new methods in this file

KPrasch: Which new methods?  I do see one: `new_account`.

nucypher/blockchain/eth/actors.py
vepkenez: python objects already have a thing called `__dict__` and methods called `setattr` and `getattr` that help manage state.

self.thing = new_value
print (self.thing)

that is state enough no?
I guess if we want to do some kind of validation on the attribute being set it makes sense, but we're not really.

Also, there is nothing stopping a developer from directly modifying these attributes anyway.

nucypher/blockchain/eth/actors.py
vepkenez: well that's all the more reason to make it modifiable without editing the code eh?

KPrasch: Indeed.  I want to bring this forth for discussion before implementing it.  If there's not already an open Issue, lets make one.

nucypher/blockchain/eth/clients.py
vepkenez: Yes that one, but I am pretty sure that some of these will fail to work in some of the various clients which will necessitate another `fix-ganache-issues` branch.

nucypher/blockchain/eth/actors.py
KPrasch: Sometime I want to perform some logic during `object.attr = value`.  In python this can be implemented with the `@property.getter` and `@property.setter` methods, but I typically prefer to use named methods for this sort of operation.

nucypher/blockchain/eth/actors.py
KPrasch: Loosely Related to #842

PR #1067, opened 2019-06-11
@transaction - Transaction management decorator

KPrasch:


KPrasch:
Commits included in PR #1058

PR #1074, opened 2019-06-13
Standardize pip reqs files

vepkenez:
Our requirements.txt and dev-requirements.txt files have undergone many massive but seemingly arbitrary revisions over the last few months, mostly depending on who the last person was to generate them.

This PR  adds a test in Circle which ensures that requirements files are generated by consistent methods.

Note:  It will also fail if sub-dependencies are updated without an accompanying update of reqs files.
I think this is probably a good thing.

### DIFF
--- a/scripts/circle/compare_reqs.sh
+++ b/scripts/circle/compare_reqs.sh
@@ -2,22 +2,40 @@
 
 set -e
 
+# update lock and build requirements files
 yes | ./scripts/installation/rebuild_pipenv.sh circlereqs
 
-if [ $(md5 -q requirements.txt) == $(md5 -q circlereqs.txt) ]; then
-    echo "requirements.txt is valid"
+echo "---- validating requirements.txt ----"
+REQSHASH=$(md5sum requirements.txt | cut -d ' ' -f1)
+TESTHASH=$(md5sum circlereqs.txt | cut -d ' ' -f1)
+
+echo "- $REQSHASH"
+echo "- $TESTHASH"
+if [ $REQSHASH == $TESTHASH ]; then
+    echo "- requirements.txt is valid ...."
 
 else
-    echo "requirements.txt contains inconsistencies"
+    echo "- requirements.txt contains inconsistencies ...."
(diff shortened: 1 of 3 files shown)


scripts/circle/compare_reqs.sh
```diff
@@ -0,0 +1,41 @@
...
+    echo "- requirements.txt contains inconsistencies ...."
+    echo "- you may want to run `pipenv sync --dev` and then ./scripts/installation/rebuild_pipenv.sh ...."
+    echo "- which will rebuild your *requirements.txt files ...."
+    diff requirements.txt circlereqs.txt
+    exit 2
+fi
+
+echo "---- validating dev-requirements.txt ----"
+REQSHASH=$(md5sum dev-requirements.txt | cut -d ' ' -f1)
+TESTHASH=$(md5sum dev-circlereqs.txt | cut -d ' ' -f1)
```
KPrasch: :boom:

scripts/installation/rebuild_pipenv.sh
```diff
@@ -1,24 +1,28 @@
...
+echo
 if [[ $REPLY =~ ^[Yy]$ ]]
 then
     pip install --upgrade pip
     pip install --upgrade setuptools
 fi
 
-rm requirements.txt
-rm dev-requirements.txt
+rm $PREFIX.txt
```
derekpierre:
```suggestion
rm -f $PREFIX.txt
```

scripts/installation/rebuild_pipenv.sh
```diff
@@ -1,24 +1,28 @@
...
+rm dev-$PREFIX.txt
```
derekpierre:
```suggestion
rm -f dev-$PREFIX.txt
```

cygnusv:
I did something very similar some months ago (1657026f2955958952811a879ce5de45db10a92e), and it was on master for a while, but I don't know exactly why it was removed at some point. I think @jMyles was involved? Maybe we can get a comment on what was the problem then.

vepkenez:
Hmm interesting. So much more concise!   And this new one will actually
fail a test if there is a difference.  I would like to see what he says.

Issue #1018, opened 2019-05-27
Implement CLI Verbosity Levels and Toggle


KPrasch:
Use the click verbosity counting option to replace --quiet, --no-logs, --no-fun, etc. and replace them with verbosity levels.   The name --no-logs is misleading, and actually disables file logging.

cygnusv:
What happen to `--fun=N` to specify the level of fun?

fjarri:
Verbosity will be the new fun (maximum fun by default?). I don't think we need a fourth option governing log output.

jMyles:
Presumably (and I commented as much on #941) the verbosity count is something we'll use as a way to control the volume of console echoing, right?  Not the actual level of logs which are emitted in the first place?

Operating with this presumption, I'm also adding a `--log-level` as a CLI parameter.  If there's discussion to have about that, and it's seen as a different issue, I'm happy to open another one.  Until then, I'll point changes here.

fjarri:
My understanding was that verbosity would act exactly as log level - that's how it usually works, isn't it? The way I imagine it is:

- `--verbose` - sets the log level
- `--quiet` - disables console output
- `--no-logs` - disables file output
- `--no-sentry` - disables output to Sentry
- `--no-fun` - disables banners
- `--debug` - sets some combination of options that is commonly used for debugging (whatever that set may be). The idea is that it is only a shortcut for a bunch of other command line options and does not do anything special. 

(the names can be changed, of course)

mswilkison:
Fixed in #37c9b75

KPrasch:
closed by #1146

PR #1079, opened 2019-06-15
[WIP] CLI options refactoring

fjarri:
This PR intends to simplify CLI option processing (mostly the ones related to logging) and remove duplicate options, which will in turn pave way for fixing issue #1018

*Note:* this branch started as a fix for issue #1018, but got transformed into CLI option rearrangement. Perhaps the last commit should be a separate PR

Public API changes:

1. All options are now specified after commands. That is, if before one would call

        $ nucypher alice --debug init --federated-only

    after the PR the syntax would be 

        $ nucypher alice init --debug --federated-only

    Cons:
    - this goes somewhat against `click` philosophy (see [this `click` issue](https://github.com/pallets/click/issues/108)). 
    - there is some amount of "magic" in `config.py` required for this to work.

    Pros:
    - this helps avoid code duplication (for option definition and processing).
    - this helps user discover these options in `--help`. That is, if before, in order to find out about `--debug`, one would have to call `nucypher --help` in addition to `nucypher alice --help`, now just `nucypher alice --help` will be enough - all the options will be there.

2. Added `--no-sentry` option to turn off Sentry logging.

3. `--debug` now just sets/unsets certain otherwise available CLI options (**TODO:** to be determined which ones exactly), acting as a shortcut.

4. `NUCYPHER_SENTRY_LOGS` and `NUCYPHER_FILE_LOGS` have stricter requirements for their contents (whatever [`strtobool`](https://docs.python.org/3/distutils/apiref.html#distutils.util.strtobool) can recognize)

4. **TODO:** `--verbose` sets the logging level

5. **TODO:** add an option to turn off banners

Internal changes:

1. `GlobalLogger` class manages all logging and observers.

2. **TODO:** get rid of `emit` methods in random classes and use standard existing logging facilities?

### DIFF
--- a/examples/finnegans_wake_demo/finnegans-wake-demo.py
+++ b/examples/finnegans_wake_demo/finnegans-wake-demo.py
@@ -9,7 +9,7 @@ from umbral.keys import UmbralPublicKey
 from nucypher.characters.lawful import Alice, Bob, Ursula
 from nucypher.characters.lawful import Enrico as Enrico
 from nucypher.network.middleware import RestMiddleware
-from nucypher.utilities.logging import ConsoleLoggingObserver
+from nucypher.utilities.logging import GlobalLogger
 from nucypher.utilities.sandbox.constants import TEMPORARY_DOMAIN
 
 ######################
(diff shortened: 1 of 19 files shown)


fjarri:
Closing in favor of reworked and rebased PR #1103

PR #1094, opened 2019-06-19
Speed up CI builds

vepkenez:
Trims about 2 minutes off total CI build with potential for additional time savings each time docker is used for future tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -439,7 +439,7 @@ jobs:
     working_directory: ~/nucypher
     <<: *python_36_base
     steps:
-      - pipenv_install
+      - prepare_environment
       - run:
           name: Run Requirements comparison
           command: ./scripts/circle/compare_reqs.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -437,7 +437,7 @@ jobs:
 
   validate_reqs_files:
     working_directory: ~/nucypher
-    <<: *python_37_base
+    <<: *python_36_base
     steps:
       - pipenv_install
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,20 +95,26 @@ workflows:
 #            - agents     # These jobs are left out to not block execution of workflow
 #            - actors
 #            - contracts
+      - build_dev_docker_images:
+          filters:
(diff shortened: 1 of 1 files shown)


tuxxy:
Why is the python3.6 base faster?

vepkenez:
It's not that.  It's because it's using a "workspace" which has a saved state with all the pip installs done already.  Which all of our tests use and are built on the python_36 base

vepkenez:
that `prepare_environment` thing...

tuxxy:
Oh, I see... neat.

.circleci/config.yml
```diff
@@ -95,20 +95,26 @@ workflows:
 #            - agents     # These jobs are left out to not block execution of workflow
 #            - actors
 #            - contracts
+      - build_dev_docker_images:
+          filters:
+            tags:
+              only: /.*/
+          requires:
+            - basics
```
cygnusv: I see that you are starting this task as soon as `basics` is done. I think it's fine as long as we agree that we are prioritizing CI performance over the "ideal" (?) order of the CI test pipeline (i.e., setup -> tests -> demos/docs -> distribution). I personally prioritize performance, but I don't know about others. 

If, on the contrary, we want to preserve this staged approach to the test pipeline, then you can use `build_dev_docker_images` as a replacement of `tests_ok`, which was use as a marker that the test stage was over. I played with this in this commit c62a2c698ff32d4a2b3579f86c58b7b928d0380c in case you want to cherry-pick it. This is an example of how it looks: https://circleci.com/workflow-run/a1a0b017-d66b-4807-bd45-09d3a7c3c141

An intermediate option is just to leave it as it is in your PR, but require `network` and `characters` besides `basics`.

vepkenez: @cygnusv this step is really not a distribution step, it's really just a build step required by a set of tests which should just as well be run in parallel to other tests.

The `build_dev_docker_images` is currently just building the docker image for the finnegan's wake and heartbeat demo tests to run, so it makes sense to me for it to run as soon as `basics` is done, otherwise we end up with an additional 3 minutes of waiting after the other tests are finished just to run 2 (+ whatever other tests we end up running in docker) more tests.

PR #1101, opened 2019-06-21
Revise and improve Trezor backend

tuxxy:
### What this does:

1. Moves all device-based code to `nucypher.device` namespace.
2. Places the `trezorlib` mocks in `tests.conftest`.
3. Moves the device tests to `tests/device`.
4. Puts `TrustedDevice` in `nucypher.device.base`.
5. Caches a specified number of addresses from the Trezor in the `Trezor` class during `__init__`.
6. Adds a `get_address` device call method to retrieve those addresses.
7. Creates a `Web3` compatible ethereum transaction signing API.

Big thanks/credit to @KPrasch for delivering a concept demonstrating how these things would work and go together.


nucypher/device/base.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    An Abstract Base Class for implementing wallet-like functions for stakers
+    utilizing trusted hardware devices (e.g. trezor).
+    This class specifies a few basic functions required for staking on the
+    NuCypher network.
+
+    TODO: Define an abstractmethod for BIP44 derivation?
+    """
+
+    ETH_BIP44_PATH = "m/44'/60'/0'/0/{address_index}"
```
KPrasch: Do we want to use a `chainID` here for separation of testnet keys (the first `0` can be a `5` for example)?

nucypher/device/base.py
```diff
@@ -0,0 +1,96 @@
...
+    def sign_message(self, message: bytes, address_index: int = 0):
+        """
+        Abstract method for signing any arbitrary message via a device's API.
+
+        TODO: What format should signatures output?
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def verify_message(self, signature: bytes, messsage: bytes, checksum_address: str):
```
cygnusv:
```suggestion
    def verify_message(self, signature: bytes, message: bytes, checksum_address: str):
```

nucypher/device/trezor.py
```diff
@@ -0,0 +1,178 @@
...
+        def wrapped_call(inst, *args, **kwargs):
+            try:
+                result = device_func(inst, *args, **kwargs)
+            except USBErrorNoDevice:
+                raise inst.DeviceError("The client cannot communicate to the TREZOR USB device. Was it disconnected?")
+            except USBErrorBusy:
+                raise inst.DeviceError("The TREZOR USB device is busy.")
+            else:
+                return result
+        return wrapped_call
```
cygnusv: This is a cool decorator 😎 Just a question, why making it "private"?

nucypher/device/trezor.py
```diff
@@ -0,0 +1,178 @@
...
+              data that is passed in?
+        """
+        try:
+            bip44_path = self.__addresses[checksum_address]
+        except KeyError:
+            raise self.DeviceError(f'{checksum_address} is not available as a cached address on this device.')
+
+        # Handle Web3.py -> trezor API formatting
+        assert_valid_fields(transaction)
+        transaction['to'] = to_canonical_address(transaction['to'])
```
cygnusv: This line is mutating the `transaction` dict, which may have side-effects outside this method.

nucypher/device/trezor.py
tuxxy: Shouldn't be used outside of `TrusteDevice` subclasses, imho.

nucypher/device/base.py
tuxxy: I was thinking about this, but I'm unsure if I want to support it quite yet. Ideally, you configure a device _for a testing chain_ and _do not_ allow the user to use the _same_ device on _different_ chains.

KPrasch: In this case - why default to chain ID `0`?

tuxxy: Because there is no interface for a device config yet and it doesn't really matter which chain is derived -- it works on both.

tuxxy: We can certainly hardcode to Goerli, if that makes more sense for now. I'm not sure if web wallets actually follow this, though...

nucypher/device/trezor.py
KPrasch: This line causes the Trezor to freeze up, unless the address is in the checksum or normalized format.  However, the below `Transaction` object expects `canonical_address` bytes, so this call needs to be move *after* the call to `sign_tx`.

nucypher/device/trezor.py
```diff
@@ -0,0 +1,178 @@
...
+
+        TODO: Is there any input validation required for the transaction
+              data that is passed in?
+        """
+        try:
+            bip44_path = self.__addresses[checksum_address]
+        except KeyError:
+            raise self.DeviceError(f'{checksum_address} is not available as a cached address on this device.')
+
+        # Handle Web3.py -> trezor API formatting
```
KPrasch: There is an additional transaction formatting step that needs to be performed here, since incoming transaction data is a hex `str`, but the `sign_tx` interface expects `bytes`.: 

```
        if unsigned_transaction['data']:
            unsigned_transaction['data'] = Web3.toBytes(hexstr=unsigned_transaction['data'])
```

nucypher/device/trezor.py
```diff
@@ -0,0 +1,178 @@
...
+                          'gasPrice': 'gas_price',
+                          'chainId': 'chain_id'}
+        trezor_format_tx = dict(apply_key_map(trezor_tx_keys, transaction))
+
+        tx_v, tx_r, tx_s = trezor_eth.sign_tx(client=self.client, n=bip44_path,
+                                              **trezor_format_tx)
+
+        # The chain_id is not necessary to include, the Trezor API creates an
+        # EIP-155 signature if needed.
+        del transaction['chainId']
```
KPrasch: Additionally, we can take a non-mutative approach to manipulating transaction `dict` values:

```
from cytoolz.dicttoolz import dissoc

unsigned_transaction = dissoc(unsigned_transaction, 'chainId')
```

nucypher/device/trezor.py
```diff
@@ -0,0 +1,178 @@
...
+        address = trezor_eth.get_address(self.client, hd_path)
+        return address
+
+    @_handle_device_call
+    def sign_eth_transaction(self, checksum_address: str, transaction: dict) -> bytes:
+        """
+        Signs an Ethereum transaction via the Trezor ethereum sign_tx API
+        and returns the signed, RLP encoded transaction.
+
+        TODO: Is there any input validation required for the transaction
```
KPrasch: With the formatting suggestions applied, and considering the call to `assert_valid_fields` - I think you can remove this `TODO`.

tuxxy:
Closing due to recent Geth additions.

https://media.giphy.com/media/4LsN0YwgIsvaU/giphy.gif

PR #1056, opened 2019-06-06
Extract Stakeholder CLI and Deploy Actions

KPrasch:
This PR goes into the Sekanjabin Epic (#1107); Refactors the staking CLI to suit the `StakeHolder` and makes improvements to the usability of development deployment.

Based on #1064 

Fixes #1119 
Contributes to fixing #1117 

Introduces #1125 

#### Status
Still has failing mixed-configuration tests, possibly related to federated vs. non-federated keyring attachment methods on Ursula.  Needs interactive reviews and Issue-opening follow-ups.

###### 1- Run Development Blockchain
```
geth --dev --dev.period 2
```

###### 2 - Deploy Contracts
```
nucypher-deploy contracts --provider-uri ipc:///tmp/geth.ipc --poa --dev
```

###### 3 - Create StakeHolder

```
nucypher stake new-stakeholder --provider-uri file:///tmp/geth.ipc --registry-filepath  ~/.local/share/nucypher/dev_contract_registry.json  --poa
```

###### 4 - Initialize Stake

```
nucypher stake init
```

PR #1108, opened 2019-06-27
Worker-Ursula Configuration

KPrasch:


KPrasch:
Dissolved into #1064 and #1056

PR #1111, opened 2019-07-01
Improved deployers tests

cygnusv:
The goal of this PR is to improve the deployer tests. Initially, the goal included to improve actor tests (hence `drama-school`), but I prefer to reduce the scope of the PR for the moment. Nevertheless, it does include some improvements in some actor tests (e.g., a rough test of Investigator).

Hopefully, this will increase coverage a little.


nucypher/blockchain/eth/deployers.py
```diff
@@ -503,6 +504,12 @@ def retarget(self, new_target: str, existing_secret_plaintext: bytes, new_secret
         if new_target == self._contract.address:
             raise self.ContractDeploymentError(f"{self.contract_name} {self._contract.address} cannot target itself.")
 
+        # TODO: Make this logic a decorator for retarget and upgrade
```
KPrasch: :+1: - Good Idea. - You like decorators these days? :-)

nucypher/blockchain/eth/decorators.py
```diff
@@ -70,3 +72,36 @@ def wrapped(*args, **kwargs):
...
+            message = f"Incorrect use of decorator: '{parameter_name}' not found. " \
+                      f"Found parameters are: {found_params}"
+            log.debug(message)
+            raise TypeError(message)
+
+        # Validate
+        secret_hash = self.contract.functions.secretHash().call()
+        is_valid_secret = secret_hash == keccak_digest(secret)
+
+        if is_valid_secret:  # Yay!  \ (•◡•) /
```
KPrasch: Haha :confetti_ball:

nucypher/blockchain/eth/actors.py
```diff
@@ -131,7 +131,7 @@ def token_balance(self) -> NU:
         return nu_balance
 
 
-class Deployer(NucypherTokenActor):
+class DeployerActor(NucypherTokenActor):
```
KPrasch: It wants another name. Any ideas?

nucypher/blockchain/eth/actors.py
```diff
@@ -112,6 +113,10 @@ def __repr__(self):
         r = r.format(class_name, self.checksum_address)
         return r
 
+    def __eq__(self, other) -> bool:
+        """Actors are equal if they have the same address."""
+        return bool(self.checksum_address == other.checksum_address)
```
jMyles: This will raise `AttributeError` is the other doesn't have a `checksum_address`.  This might be OK, but also maybe we want to catch it and raise `TypeError` instead?

nucypher/blockchain/eth/actors.py
jMyles: Can it not stay `Deployer`?  

Who is an actor which deploys a token or contract?  `Bursar`?  `Secretary`?  `Executive`?

nucypher/blockchain/eth/actors.py
```diff
@@ -136,10 +141,15 @@ class Deployer(NucypherTokenActor):
...
 
-    contract_names = tuple(a.registry_contract_name for a in EthereumContractAgent.__subclasses__())
+    __secrets = ('staker_secret',
+                 'policy_secret',
+                 'user_escrow_proxy_secret',
+                 'adjudicator_secret')
+    Secrets = collections.namedtuple('Secrets', __secrets)
 
-    __interface_class = BlockchainDeployerInterface
+    contract_names = tuple(a.registry_contract_name for a in EthereumContractAgent.__subclasses__())
```
jMyles: Quasimodo in here.  Maybe this needs to be a `classmethod` on `EthereumContractAgent`?

nucypher/blockchain/eth/actors.py
```diff
@@ -248,39 +258,65 @@ def deploy_user_escrow(self, allocation_registry: AllocationRegistry):
...
+        deployment_receipts = dict()
+        gas_limit = 8_000_000  # TODO: Gas management
+
+        # NuCypherToken
+        token_receipts, token_deployer = self.deploy_contract(contract_name=NucypherTokenDeployer.contract_name,
+                                                              gas_limit=gas_limit)
+        paint_contract_deployment(contract_name=NucypherTokenDeployer.contract_name,
+                                  receipts=token_receipts,
+                                  contract_address=token_deployer.contract_address)
+        deployment_receipts[NucypherTokenDeployer.contract_name] = token_receipts
```
jMyles: So, the key space of the dict will be values of `NucypherTokenDeployer.contract_name`... does this value mutate then?  It's not obvious how to think about this.

nucypher/blockchain/eth/actors.py
```diff
@@ -248,39 +258,65 @@ def deploy_user_escrow(self, allocation_registry: AllocationRegistry):
...
+            click.pause(info="Press any key to continue")
+
+        # StakingEscrow
+        staking_receipts, staking_deployer = self.deploy_contract(contract_name=StakingEscrowDeployer.contract_name,
+                                                                  plaintext_secret=secrets.staker_secret,
+                                                                  gas_limit=gas_limit)
+        paint_contract_deployment(contract_name=StakingEscrowDeployer.contract_name,
+                                  receipts=staking_receipts,
+                                  contract_address=staking_deployer.contract_address)
+        deployment_receipts[StakingEscrowDeployer.contract_name] = staking_receipts
```
jMyles: Same comment as above - a reader has to assume that `StakingEscrowDeployer.contract_name`, but since this doesn't occur in a loop, the domain of the key space of this dict is difficult to grok.

nucypher/blockchain/eth/actors.py
```diff
@@ -969,25 +1029,50 @@ def initialize_stake(self,
...
+                    raise ValueError(f"{checksum_address} is an unknown wallet address.")
+                else:
+                    staker = Staker(is_me=True, checksum_address=checksum_address, blockchain=self.blockchain)
+
+        # New Staker address (no address passed)
         else:
-            staker = Staker(is_me=True, checksum_address=checksum_address, blockchain=self.blockchain)
+            if not password:
+                raise ValueError(f'No checksum address or password supplied to initialize new stake.')
+            staker = self.__create_staker(password=password)
```
jMyles: I need to dive in w/ a debugger, but I'm imagining that a random wallet is generated here?  Maybe that needs to be called out in at least a comment on this layer?

nucypher/blockchain/eth/actors.py
KPrasch: No - they do not mutate - they are hard-coded class attributes on the Deployer claas.  This is an effort to have a single source of contract names in the codebase.  In general, this method can be code-optimized somewhat with a loop and/or function.

nucypher/blockchain/eth/actors.py
KPrasch: I agree - more comments are needed here.  This area needs some thinking, ( see related Issue #1128 and #1129 ).  In the case of a software wallet - the RPC endpoint for account creation is used.  In the cased of a hardware wallet, an unused account is derived and used (which is always the last index geth loads for hardware wallets).

nucypher/blockchain/eth/actors.py
jMyles: Maybe uhhh... :thinking:... Maybe use a constant?  `NucypherTokenDeployer.contract_name` is a hard thing for the reader to resolve in terms of understanding the keyspace here.

nucypher/blockchain/eth/actors.py
jMyles: Yeah, so this is a bit of a lava bridge.  I think it's a little nerve-racking for `self.__create_staker` to, without affirmative direction to do so, use a heretofore unused checksum address.

nucypher/blockchain/eth/decorators.py
```diff
@@ -70,3 +72,36 @@ def wrapped(*args, **kwargs):
...
+            return func(self, *args, **kwargs)
+        else:
+            message = f"The secret for upgrade/rollback {self.contract_name} is not {secret}."
```
derekpierre: Do we actually want to include the `secret` here? What if it is close to the actual secret itself?

derekpierre:
@cygnusv your tests are so clear to read 🎉 !

nucypher/blockchain/eth/deployers.py
```diff
@@ -360,13 +375,13 @@ def rollback(self, existing_secret_plaintext: bytes, new_secret_hash: bytes):
...
+                                                        new_secret_hash=new_secret_hash,
+                                                        gas_limit=gas_limit)
 
-        txhash = rollback_receipt['transactionHash']
-        return txhash
+        return rollback_receipt
 
     def make_agent(self) -> EthereumContractAgent:
-        self.__check_policy_manager()  # Ensure the PolicyManager contract has already been initialized
+        #self.__check_policy_manager()  # Ensure the PolicyManager contract has already been initialized
```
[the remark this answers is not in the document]
cygnusv: It's on purpose. I don't see very consistent disallowing the creation of an agent when there's no policy manager set, while at the same time we create an StakingEscrow agent even before deploying PolicyManager, in the constructor of its deployer.

nucypher/blockchain/eth/decorators.py
cygnusv: That's a fair comment. I think we can remove that.

derekpierre: @KPrasch: to add since @cygnusv is on vacation:
```suggestion
            message = f"The secret for upgrade/rollback {self.contract_name} is not valid."
```

tuxxy: I'd do `provided ({secret})...` for visual confirmation of what was entered.

tests/blockchain/eth/clients/test_geth_integration.py
```diff
@@ -60,17 +59,17 @@ def test_geth_deployment_integration(instant_geth_dev_node):
 
     # Make Deployer
     etherbase = to_checksum_address(instant_geth_dev_node.accounts[0].decode())  # TODO: Make property on nucypher geth node instances?
-    deployer = Deployer(blockchain=blockchain,
-                        deployer_address=etherbase,
-                        client_password=None)  # dev accounts have no password.
+    deployer = DeployerActor(blockchain=blockchain,
```
michwill: This test fails for me due to exceeding the gas limits.
How do we currently cope with that? I mean, it's certainly possible to set a new target limit and wait, but that's slow...

michwill: One idea could be to run geth with mining-at-transaction-time, and manually mine few hundreds blocks after spinning up geth

michwill: One way to do it is to add `--targetgaslimit 8000000` to get command, and make some 100 transfers of eth from `accounts[0]` to `accounts[0]` (this'd take a few seconds).

But I wonder: what's the current way to test it? Older geth?

nucypher/blockchain/eth/actors.py
KPrasch: @jMyles - This name clashes with the namespace in `nucypher.blockchain.eth.deployers.Deployer`.

nucypher/blockchain/eth/actors.py
KPrasch: Yeah I can see it that way.  You propose we have a constants module with all the contract name strings defined?

nucypher/blockchain/eth/actors.py
KPrasch: From a brid's-eye view does it make sense to implement the feature at all? (#1129 ).  What's the user's experience like if they need to create accounts to stake with?

nucypher/blockchain/eth/decorators.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/actors.py
KPrasch: @jMyles  - Also, If you read above, the user is prompted for consent to use a new account.

tests/blockchain/eth/entities/deployers/test_user_escrow_deployer.py
```diff
@@ -16,58 +16,107 @@
...
+    testerchain = session_testerchain
+    deployer_account = testerchain.etherbase_account
+
+    linker_deployer = LibraryLinkerDeployer(blockchain=testerchain,
+                                            deployer_address=deployer_account,
+                                            target_contract=user_escrow_proxy_deployer.contract,
+                                            bare=True)
+    linker_address = linker_deployer.contract_address
+
+    for index in range(NUMBER_OF_USERS):
```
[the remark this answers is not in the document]
cygnusv: We mean users as in UserEscrow.

nucypher/blockchain/eth/actors.py
KPrasch: My initial impression is that returning `False` is desirable is that situation.

tests/blockchain/eth/entities/deployers/test_user_escrow_deployer.py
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

tests/blockchain/eth/clients/test_geth_integration.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/actors.py

tests/blockchain/eth/clients/test_geth_integration.py
KPrasch: There was an inadvertent deployment transaction gas limit set to 8 Million just above this line causing block gas limit to be exceeded.

PR #1116, opened 2019-07-10
fix issues preventing contract deploys and felix instantiation

vepkenez:
* adds gas-limit cli arg to nucypher-deploy contracts
* implements transaction signing for ganache accounts using envvar
* fixes exception raised when casting constant sorrow to boolean
* adds a couple of optional envvars for some hard-coded constants

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -246,12 +246,13 @@ class Deployer(NucypherTokenActor):
                                  policy_secret: str,
                                  adjudicator_secret: str,
                                  user_escrow_proxy_secret: str,
+                                 gas_limit: int,
                                  ) -> Tuple[dict, dict]:
         """
         Musketeers, if you will; Deploy the "big three" contracts to the blockchain.
         """
 
-        token_txs, token_deployer = self.deploy_contract(contract_name='NuCypherToken')
+        token_txs, token_deployer = self.deploy_contract(contract_name='NuCypherToken', gas_limit=gas_limit)
         staking_txs, staking_deployer = self.deploy_contract(contract_name='StakingEscrow', plaintext_secret=staker_secret)
         policy_txs, policy_deployer = self.deploy_contract(contract_name='PolicyManager', plaintext_secret=policy_secret)
         adjudicator_txs, adjudicator_deployer = self.deploy_contract(contract_name='Adjudicator', plaintext_secret=adjudicator_secret)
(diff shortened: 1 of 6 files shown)


nucypher/blockchain/eth/actors.py
```diff
@@ -246,12 +246,13 @@ def deploy_network_contracts(self,
...
                                  adjudicator_secret: str,
                                  user_escrow_proxy_secret: str,
+                                 gas_limit: int,
                                  ) -> Tuple[dict, dict]:
         """
         Musketeers, if you will; Deploy the "big three" contracts to the blockchain.
         """
 
-        token_txs, token_deployer = self.deploy_contract(contract_name='NuCypherToken')
+        token_txs, token_deployer = self.deploy_contract(contract_name='NuCypherToken', gas_limit=gas_limit)
```
szotov: Maybe set `gas_limit` to `None` by default? It will help when gas estimation works properly

vepkenez: 👍

nucypher/blockchain/eth/clients.py
```diff
@@ -250,6 +250,21 @@ def sign_message(self, account: str, message: bytes) -> str:
         """
         return self.w3.eth.sign(account, data=message)
 
+    def get_signing_key(self, account):
```
KPrasch: I do not think we want to expose local key signing here, since it's handled by the client's keyring.  We apply this logic below to `EthereumTester` because it does not expose the `signTransaction` RPC endpoints, but wed like to test our code using those calls.

nucypher/blockchain/eth/clients.py
```diff
@@ -288,6 +303,13 @@ def unlock_account(self, address, password):
     def sync(self, *args, **kwargs):
         return True
 
+    def get_signing_key(self, account):
```
KPrasch: does Ganache not support the `signTransaction` and `sendRawTransaction` endpoints?

nucypher/cli/deploy.py
```diff
@@ -67,6 +68,7 @@ def deploy(click_config,
            registry_outfile,
            no_compile,
            amount,
+           gas_limit,
```
KPrasch: Is this gas limit applied to only deployment transactions?

nucypher/blockchain/eth/clients.py
vepkenez: @KPrasch apparently not... :|

https://github.com/trufflesuite/ganache-core/issues/408

KPrasch: OK!  We need to handle the local private key then, much like `EthereumTesterClient`; However, let's remove all the private key handling from the base class, since we only want to do so in-tests.

KPrasch:
Left RFCs regarding composition of local key handling (Perhaps a local key handler mixin?).  Also needs rebase to resolve merge conflicts.

PR #1121, opened 2019-07-13
[APPROVED - DON'T MERGE YET] Contracts size optimizations

vzotova:
* Decreases size of `StakingEscrow` contract (decreases gas usage for deploy)
* Decreases gas usage in `createPolicy`, moves some data handling to revoke/refund methods
* Increases gas refund for revoke/refund methods, deacreases result cost of those methods
* Refactores `estimate_gas` script, more accurate gas estimations and using gas refunds in statistics

Some of those optimization are slightly dangerous due to removal of `SafeMath` protection. Reason for removing is that we have multiple layers of those checks in different methods. 
Major ideas:
* If we work with value that represents amount of tokens outside token contract then we can just add (and sometimes substract) value without check
* Contracts are not really work near last period value (~2150 year?), so no reason to check overflow for `somePeriod+1`
* Zero period makes no sense (1970 year), so also no check for `somePeriod-1`
* Checks in constructor is free from gas usage perspective (while deploy) 
* `require` messages are not cheap (especially if more than 32 bytes). So removed some messages from one-time functions


nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -56,11 +56,20 @@ contract Issuer is Upgradeable {
...
             _hoursPerPeriod != 0 &&
             _lockedPeriodsCoefficient != 0 &&
             _rewardedPeriods != 0);
+        uint256 maxLockedPeriods = _rewardedPeriods + _lockedPeriodsCoefficient;
+        require(maxLockedPeriods > _rewardedPeriods &&
+            // worst case for `totalLockedValue * k2`, when totalLockedValue == totalSupply
+            totalSupply * miningCoefficient / totalSupply == miningCoefficient &&
+            // worst case for `(totalSupply - currentSupply) * lockedValue * (k1 + allLockedPeriods)`,
+            // when currentSupply == 0, lockedValue == totalSupply
+            totalSupply * totalSupply * maxLockedPeriods / totalSupply / totalSupply == maxLockedPeriods,
```
mswilkison: I don't understand what these checks are doing... won't `totalSupply`always cancel out to `maxLockedPeriods == maxLockedPeriods` and `miningCoefficient == miningCoefficient`? So when can this require fail?

szotov: It will fail in case of overflow after multiplication, `totalSupply * totalSupply * maxLockedPeriods` will be less than `totalSupply` or `maxLockedPeriods`

michwill:
Yeah, that's a very dense PR, requires a very deep thinking about why wouldn't optimizations introduce vulnerabilities (it's clear when you think a bit though).
Maybe a call is needed to walk through?

mswilkison:
> Yeah, that's a very dense PR, requires a very deep thinking about why wouldn't optimizations introduce vulnerabilities (it's clear when you think a bit though).
> Maybe a call is needed to walk through?

Worth having auditors take a look during the worklock audit?

michwill:
I think it is worth checking out, yes. Note that there are some substitutions of `.add()` safe methods into "unsafe" ones (to save gas, in case overflow cannot happen). The fact that we don't shoot ourselves in the foot would indeed benefit from the second pair of eyes

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
michwill: Very nice things to check :+1:

nucypher/blockchain/eth/sol/compile.py
```diff
@@ -105,12 +105,13 @@ def compile(self) -> dict:
 
         self.log.info("Compiling with import remappings {}".format(", ".join(remappings)))
 
-        optimization_runs = 10  # TODO: Move..?
+        optimization_runs = 200  # TODO: Move..?
```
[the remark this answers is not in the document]
szotov: I just restored how it really was - wrong parameters were in `compile_files` (`optimize` instead of `optimize_runs`), so function worked with default 200 runs,. I decided to change it here otherwise it confusing

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -159,9 +159,9 @@ contract PolicyManager is Upgradeable {
         Policy storage policy = policies[_policyId];
         policy.client = msg.sender;
         uint16 currentPeriod = getCurrentPeriod();
-        policy.startPeriod = currentPeriod.add16(1);
+        policy.startPeriod = currentPeriod + 1;
         policy.lastPeriod = currentPeriod.add16(_numberOfPeriods);
-        policy.rewardRate = msg.value.div(_nodes.length).sub(_firstPartialReward).div(_numberOfPeriods);
+        policy.rewardRate = msg.value.div(_nodes.length).sub(_firstPartialReward) / _numberOfPeriods;
```
cygnusv: Why there's still a `div()` and `sub()` here?

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -244,36 +251,34 @@ contract PolicyManager is Upgradeable {
             } else if (endPeriod < minPeriod) {
                 continue;
             }
-            downtimePeriods = downtimePeriods.add16(
-                AdditionalMath.min16(maxPeriod, endPeriod)
-                .sub16(AdditionalMath.max16(minPeriod, startPeriod))
-                .add16(1));
+            downtimePeriods += AdditionalMath.min16(maxPeriod, endPeriod)
+                .sub16(AdditionalMath.max16(minPeriod, startPeriod)) + 1;
```
cygnusv: There's still a `sub16()` in here. Is that intentional?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -177,37 +172,40 @@ contract StakingEscrow is Issuer {
...
     * @notice Set worklock address
     **/
     function setWorkLock(WorkLockInterface _workLock) external onlyOwner {
-        // Two-part require...
-        require(address(workLock) == address(0) &&  // Can't workLock once it is set.
-            _workLock.escrow() == address(this));  // This is the escrow for the new workLock.
+        // WorkLock can be set only once
+        require(address(workLock) == address(0));
+        // This escrow must be the escrow for the new worklock
+        require(_workLock.escrow() == address(this));
```
cygnusv: Are two requires better than a single two-part require?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -527,7 +536,7 @@ contract StakingEscrow is Issuer {
     * @param _value Amount of tokens which will be locked
     * @param _periods Amount of periods during which tokens will be locked
     **/
-    function lock(uint256 _value, uint16 _periods) public onlyStaker {
+    function lock(uint256 _value, uint16 _periods) external onlyStaker {
```
cygnusv: Why all these changes from `public` to `external`?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -538,23 +547,22 @@ contract StakingEscrow is Issuer {
...
             _periods >= minLockedPeriods);
 
         uint16 currentPeriod = getCurrentPeriod();
-        uint16 nextPeriod = currentPeriod.add16(1);
+        uint16 nextPeriod = currentPeriod + 1;
         StakerInfo storage info = stakerInfo[_staker];
         uint256 lockedTokens = getLockedTokens(info, currentPeriod, nextPeriod);
-        require(_value.add(lockedTokens) <= info.value &&
-            _value.add(lockedTokens) <= maxAllowableLockedTokens);
+        uint256 requestedLockedTokens = _value.add(lockedTokens);
```
cygnusv: Change `add()` by `+`?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -618,17 +624,18 @@ contract StakingEscrow is Issuer {
         uint256 oldValue = subStake.lockedValue;
         subStake.lockedValue = oldValue.sub(_newValue);
         require(subStake.lockedValue >= minAllowableLockedTokens);
-        saveSubStake(info, subStake.firstPeriod, 0, subStake.periods.add16(_periods), _newValue);
+        uint16 requestedPeriods = subStake.periods.add16(_periods);
```
cygnusv: `add16()` --> `+`?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -752,36 +758,26 @@ contract StakingEscrow is Issuer {
...
         uint16 startPeriod = getStartPeriod(info, currentPeriod);
         uint256 reward = 0;
         if (info.confirmedPeriod1 != EMPTY_CONFIRMED_PERIOD &&
             info.confirmedPeriod1 < info.confirmedPeriod2) {
-            reward = reward.add(mint(_staker, info, 1, currentPeriod, startPeriod));
+            reward = mint(_staker, info, 1, currentPeriod, startPeriod);
         } else if (info.confirmedPeriod2 != EMPTY_CONFIRMED_PERIOD &&
             info.confirmedPeriod2 < info.confirmedPeriod1) {
-            reward = reward.add(mint(_staker, info, 2, currentPeriod, startPeriod));
+            reward = mint(_staker, info, 2, currentPeriod, startPeriod);
```
cygnusv: Don't lines 765 and 768 need `reward += ... ` ?

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
szotov: only `!=0` need to check for `div()`, `_numberOfPeriods` was checked in first `require`

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
szotov: It's intentional, mostly for unreal situation when `getPastDowntime` will return something unpredictable

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
szotov: Not really, they are the same, it's just unification with other similar functions

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
szotov: I prefer when function has the most strict modifier, and those methods logically only for calling from outside. No reason from optimization perspective

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
szotov: here `_value` comes from outside, if somebody set too high `_value` then will be overflow

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
szotov: same as previous with `_value`, `_period` could be too big

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
szotov: it was just mistake, before `if` reward is zero, so no need in `+`

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -538,16 +534,15 @@ contract StakingEscrow is Issuer {
     * @param _periods Amount of periods during which tokens will be locked
     **/
     function lock(address _staker, uint256 _value, uint16 _periods) internal {
-        require(_value <= token.balanceOf(address(this)) &&
```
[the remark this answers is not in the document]
szotov: Good question: idea was that locking amount was compared with `info.value` (line 545, `requestedLockedTokens <= info.value`) and `info.value` can be increased only through `deposit()` function. So `info.value` is allways less than balance of contract and not neccesery do this check `_value <= token.balanceOf(address(this))`

tests/blockchain/eth/contracts/main/staking_escrow/test_staking.py
```diff
@@ -735,3 +753,16 @@ def test_slashing(testerchain, token, escrow_contract, token_economics, deploy_c
     assert 10 == event_args['penalty']
     assert investigator == event_args['investigator']
     assert 0 == event_args['reward']
+
+    # Next test: optimization does not brake the saving old sub stake
```
cygnusv:
```suggestion
    # Next test: optimization does not break saving old sub stake
```

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -425,15 +434,15 @@ contract StakingEscrow is Issuer {
...
                 "Not enough time has passed since the previous setting worker");
             // Remove the old relation "worker->staker"
             workerToStaker[info.worker] = address(0);
         }
 
-        if (_worker != address(0)){
+        if (_worker != address(0)) {
             require(workerToStaker[_worker] == address(0), "Specified worker is already in use");
             require(stakerInfo[_worker].subStakes.length == 0 || _worker == msg.sender,
                 "Specified worker is a staker");
```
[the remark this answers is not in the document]
vzotova: Yep, I removed from one-time method on purpose, 
but I hope this require messages will be useful or it will be just a dead weight for next optimization time

PR #1123, opened 2019-07-16
[WIP] Adds static file service to all web server endpoints

vepkenez:
Adds Twisted/Hendrix static file resources to our web server characters.

This will allow us to build better/nicer admin/monitor/status pages without resorting to hosting statics in centralized ways.

~Much of the code in this PR could be removed if these two PRs to hendrix were merged https://github.com/hendrix/hendrix/pulls~

Much of the code in /network/protocols.py could be removed if we could resolve some dependency issues and use the latest version of Hendrix

### DIFF
--- a/nucypher/characters/chaotic.py
+++ b/nucypher/characters/chaotic.py
@@ -29,10 +29,10 @@ from nucypher.config.constants import TEMPLATES_DIR
 from nucypher.crypto.powers import SigningPower, TransactingPower
 from nucypher.keystore.threading import ThreadedSession
 from nucypher.network.nodes import FleetStateTracker
-from nucypher.network.protocols import get_statics, HendrixDeployWithStatics
+from nucypher.network.server import NonTLSHost
 
 
-class Moe(Character):
+class Moe(Character, NonTLSHost):
     """
     A monitor (lizard?)
     """
@@ -108,16 +108,16 @@ class Moe(Character):
         # Server
         #
 
-        deployer = HendrixDeployWithStatics(action="start", options={"wsgi": rest_app, "http_port": http_port})
-        deployer.resources = get_statics()
-
+        deployer = self.get_deployer(
+            '127.0.0.1', http_port,
+            options={"wsgi": rest_app, "http_port": http_port})
         deployer.add_non_tls_websocket_service(websocket_service)
 
         if not dry_run:
             deployer.run()
 
 
-class Felix(Character, NucypherTokenActor):
+class Felix(Character, NucypherTokenActor, NonTLSHost):
     """
     A NuCypher ERC20 faucet / Airdrop scheduler.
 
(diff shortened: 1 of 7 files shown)


vepkenez:
These tests pass locally... I think there is some kind of file system/path issue on circle that I need to figure out.

nucypher/characters/chaotic.py
```diff
@@ -108,7 +108,9 @@ def status():
         # Server
         #
 
-        deployer = HendrixDeploy(action="start", options={"wsgi": rest_app, "http_port": http_port})
+        deployer = HendrixDeployWithStatics(action="start", options={"wsgi": rest_app, "http_port": http_port})
+        deployer.resources = get_statics()
```
KPrasch: Can this call be embedded inside of the above `HendrixDeployWithStatics` init?

vepkenez: I left it there because if we merge those changes in hendrix, then statics will be added like this
```
HendrixDeploy(action="start", options={"wsgi": rest_app, "http_port": http_port, "resources": get_statics()})
```

nucypher/network/server.py
```diff
@@ -421,3 +421,19 @@ def __init__(self,
         elif public_certificate_filepath:
             kwargs['keypair'] = HostingKeypair(certificate_filepath=public_certificate_filepath, host=host)
         super().__init__(*args, **kwargs)
+
+
+class NonTLSHost():
```
tuxxy: I'm just gonna go ahead and reject on the basis that TLS should always be enabled on _any_ host or endpoint for us.

tuxxy: "Why? There's no sensitive information going over the wire."

Without TLS, anyone can inject anything into the traffic, local or external network. This violates our threat model, imho.

vepkenez: I agree with you and I don’t know the history of why it was done this way in the 1st place.  But this isn’t the “upgrade Moe and Felix to https” pull request. Should I do that?

vepkenez: i always assumed that for these “chaotic” characters they could just have their https endpoints be a load balancer or something

tuxxy: Hmmm, I see your point about the load balancer, but I don't think that's a good assumption. We shouldn't be building a pretty major foot-gun like that.

I would much prefer to approve a PR that used TLS than not.

vepkenez: I think it should be optional because  self signed TLS may not be preferable to a ca signed load balancer TLS which might be easier with a normal http endpoint.

tuxxy: Unless you're going to build support in and make it a workable default (ie: working with LetsEncrypt), then I'm not sure there's currently any secure preference at all because a user wouldn't see a `NonTLSHost` class as a secure default.

vepkenez: My opinion about this is that unless it's a CA signed domain validated cert, users will see **THIS SITE IS INSECURE!!! GO BACK!**.  right?

KPrasch: Consider that ethereum client nodes run an RPC server, which can be interfaced with the HTTP web3.py provders here: <https://web3py.readthedocs.io/en/stable/providers.html#httpprovider>.  

For example, `geth` by default runs it's RPC server over a (non TLS) HTTP transport.  What is the thinking behind this design? 

On driver of my understanding here is the imperfect match between HTTP and the "web3 universe".  The web3 model of security is different in some distinct ways, since any blockchain read operation requires a cryptographic signature and all the transport has to do is pass along RPC calls.

vepkenez: Right.  That is our universe.

PR #1132, opened 2019-07-23
fixed emitter error for bob

pranav-singhal:
in line 146 of bob.py

it says:
`click_config.emitter(message=f"Bob Verifying Key {bytes(BOB.stamp).hex()}", color='green', bold=True)`

this line was throwing error when I tried to run: 
`nucypher bob run -d -F --teacher-uri localhost:10151 --http-port 3002`

i think it should be: 
`click_config.emit(message=f"Bob Verifying Key {bytes(BOB.stamp).hex()}", color='green', bold=True)`

### DIFF
--- a/nucypher/cli/characters/bob.py
+++ b/nucypher/cli/characters/bob.py
@@ -143,7 +143,7 @@ def bob(click_config,
             rpc_controller.start()
             return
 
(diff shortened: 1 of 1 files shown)


KPrasch:
Addressed upstream in #1103

Issue #1137, opened 2019-07-24
`nucypher alice run`:  TypeError: an integer is required (got type NoneType)


KPrasch:
Looks like http controller port's type is not being properly handled via CLI flags.

Issue #1128, opened 2019-07-18
Reliable and thin device auto-detection for Transacting Power and CLI.


KPrasch:

cygnusv:
Related to #1385

PR #1146, opened 2019-07-25
CLI Cleanup

### LINKED ISSUE #1139: Interactive account selection for alice --pay-with
At the moment, if `--pay-with` is not passed to `nucypher alice init`  (decentralized mode) - `nucypher.config.base.ConfigurationError: No checksum address provided for decentralized configuration.
` is raised.

KPrasch:
Fixes #1139 #1137 #1148 #1133 #1155

### DIFF
--- a/nucypher/cli/actions.py
+++ b/nucypher/cli/actions.py
@@ -27,6 +27,7 @@ from nacl.exceptions import CryptoError
 from twisted.logger import Logger
 
 from nucypher.blockchain.eth.clients import NuCypherGethGoerliProcess
+from nucypher.blockchain.eth.decorators import validate_checksum_address
 from nucypher.blockchain.eth.token import Stake
 from nucypher.characters.lawful import Ursula
 from nucypher.cli import painting
@@ -65,10 +66,17 @@ class UnknownIPAddress(RuntimeError):
     pass
 
 
-def get_password(confirm: bool = False) -> str:
-    keyring_password = os.environ.get("NUCYPHER_KEYRING_PASSWORD", NO_PASSWORD)
+@validate_checksum_address
+def get_client_password(checksum_address) -> str:
+    prompt = f"Enter password to unlock account {checksum_address}"
+    keyring_password = click.prompt(prompt, hide_input=True)
+    return keyring_password
+
+
+def get_nucypher_password(confirm: bool = False, envvar="NUCYPHER_KEYRING_PASSWORD") -> str:
+    keyring_password = os.environ.get(envvar, NO_PASSWORD)
     if keyring_password is NO_PASSWORD:  # Collect password, prefer env var
-        prompt = "Enter keyring password"
+        prompt = "Enter nucypher keyring password"
         keyring_password = click.prompt(prompt, confirmation_prompt=confirm, hide_input=True)
     return keyring_password
 
@@ -243,7 +251,7 @@ def make_cli_character(character_config,
         character_config.attach_keyring()
         unlock_nucypher_keyring(emitter,
                                 character_configuration=character_config,
-                                password=get_password(confirm=False))
+                                password=get_nucypher_password(confirm=False))
 
     # Handle Teachers
     teacher_nodes = None
--- a/nucypher/cli/characters/alice.py
+++ b/nucypher/cli/characters/alice.py
@@ -19,7 +19,7 @@ from nucypher.config.characters import AliceConfiguration
 @click.option('--teacher', 'teacher_uri', help="An Ursula URI to start learning from (seednode)", type=click.STRING)
 @click.option('--min-stake', help="The minimum stake the teacher must have to be a teacher", type=click.INT, default=0)
 @click.option('--discovery-port', help="The host port to run node discovery services on", type=NETWORK_PORT)
-@click.option('--controller-port', help="The host port to run Alice HTTP services on", type=NETWORK_PORT)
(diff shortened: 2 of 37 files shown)


nucypher/cli/actions.py
```diff
@@ -65,10 +66,17 @@ class UnknownIPAddress(RuntimeError):
...
-    keyring_password = os.environ.get("NUCYPHER_KEYRING_PASSWORD", NO_PASSWORD)
+@validate_checksum_address
+def get_client_password(checksum_address) -> str:
+    prompt = f"Enter password to unlock account {checksum_address}"
+    keyring_password = click.prompt(prompt, hide_input=True)
+    return keyring_password
+
+
+def get_nucypher_password(confirm: bool = False, envvar="NUCYPHER_KEYRING_PASSWORD") -> str:
+    keyring_password = os.environ.get(envvar, NO_PASSWORD)
```
fjarri: If this is a boolean var, it might be better to use `cli/config.py:get_env_bool()` here - stricter, but less surprises (the function is perhaps better moved to `utilities` somewhere in that case)

nucypher/crypto/powers.py
```diff
@@ -146,7 +146,7 @@ def account(self) -> str:
 
     def activate(self, password: str = None):
         """Be Consumed"""
-        self.blockchain.connect(sync_now=False)
+        self.blockchain.connect(fetch_registry=True, sync_now=False)
```
fjarri: Since `False` is the default now, this may not be necessary (depending on whether the intention here is "do the default action" or "never sync")

cygnusv: `fetch_registry=True` may clash with a fix for #1166. Is it truly necessary to always fetch registry here?

KPrasch: I think this can be handled elsewhere, no critical reason to be here other than being prudent.

nucypher/cli/actions.py
KPrasch: You know - I'm thinking of restoring the original state of this code since its not being implemented anywhere.  I had started to do this thinking I was going to share one method for password collection of both client and nucypher keyrings.  Instead, I ended up writing multiple functions.

nucypher/cli/characters/alice.py
```diff
@@ -171,11 +185,28 @@ def alice(click_config,
...
+        return actions.destroy_configuration(emitter, character_config=alice_config, force=force)
+
+    #
+    # Produce Alice
+    #
+
+    # TODO: OH MY.
+    client_password = None
+    if not alice_config.federated_only:
+        if (not hw_wallet or not dev) and not click_config.json_ipc:
```
michwill: Is `hw_wallet` option used at all except here?

nucypher/cli/characters/bob.py
```diff
@@ -146,7 +152,12 @@ def bob(click_config,
         # Start Controller
         controller = BOB.make_web_controller(crash_on_error=click_config.debug)
         BOB.log.info('Starting HTTP Character Web Controller')
-        return controller.start(http_port=http_port, dry_run=dry_run)
+        return controller.start(http_port=controller_port , dry_run=dry_run)
```
michwill: " " after `controller_port` here

KPrasch: Huh?

nucypher/cli/characters/alice.py
KPrasch: This is the only reason the flag is used at the moment.  See related #1128

nucypher/cli/characters/alice.py
```diff
@@ -117,10 +119,22 @@ def alice(click_config,
...
 
+        if not provider_uri and not federated_only:
+            raise click.BadOptionUsage(option_name='--provider',
+                                       message="--provider is required to create a new decentralized alice.")
+
         if not config_root:                         # Flag
             config_root = click_config.config_file  # Envvar
 
-        new_alice_config = AliceConfiguration.generate(password=get_password(confirm=True),
+        if not pay_with and not federated_only:
```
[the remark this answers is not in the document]
KPrasch: Sorry - I dont mean to ignore any comments, just a bit of alert fatigue.  Backstory here:  originally the flag was `--checksum-address` for all characters.  This started to feel a little meaningless in regards to alice, who only uses the blockchain to pay for policies.  Bob CLI was developed last, and the CLI flag from alice was copied and pasted into bob, leaving `--pay-with` as a refactoring artifact which I initially accepted.  Bob's only reason to have a ethereum wallet, is to challenge incorrect re-encryptions.  Is it appropriate to call this `--pay-with`?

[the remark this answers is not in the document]
derekpierre: No worries - I definitely didn't take it personally and figured you just missed it 😄 

It does seem like `pay-with` is a bit more descriptive than `checksum-address`. Perhaps `wallet-address` or something of the like could be good...?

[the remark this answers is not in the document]
KPrasch: "pay with" was easy to name since it only does just that.  The other CLI modules have a more broad use of this flag across different actions.  However, I am certainly interested in an alternate name for `--checksum-address`.  At the moment this name reflects that of a character (`Character.checksum_address`), and may actually be a "federated checksum address"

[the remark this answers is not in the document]
derekpierre: 👍

PR #1175, opened 2019-07-31
Bug fixes, CLI enhancements, and other touchups

### LINKED ISSUE #1169: Running tests overwrites contract registry in my local nucypher installation
For some reason, when I run `tests/characters/test_ursula_prepares_to_act_as_mining_node.py`, my existing contract registry in my local nucypher installation (located in `~/Library/Application\ Support/nucypher/contract_registry.json` in OS X) is overwritten with the one from GH (which opportunely still refers to `MinersEscrow`). Obviously, running tests shouldn't mess with a local installation. 

I guess it's as combination of `fetch_registry` in the wrong place (see #1166), and a wrong registry filepath during tests (which should never use the default path). Also, I don't think it's anything particular to that test module.

### LINKED ISSUE #1186: "Press any key to continue..."
During the deploy process, let's list the actual contract about to be deployed.

cygnusv:
* Fixes #1169: Tests were overwriting local installation's registry
* CircleCI: 
    * Logs from test runs are now stored as CI artifacts, in case they're need for inspection
    * Switch to codecov@1.0.5 orb
* When `--etherscan` in the CLI, some selected actions (currently only contract deployment) open a new tab in the browser with the TX or contract from Etherscan. Need opinions on this one.
* Refinement of `set_worker` and `detach_worker` in the CLI
* Minor stuff:
     * Fixes #1186


nucypher/blockchain/eth/actors.py
```diff
@@ -286,14 +288,16 @@ def deploy_network_contracts(self,
...
+            total_deployment_transactions += len(deployer_class.deployment_steps)
 
         first_iteration = True
-        with click.progressbar(length=total_deployment_transactions, label="Deployment progress") as bar:
+        with click.progressbar(length=total_deployment_transactions,
+                               label="Deployment progress",
+                               show_eta=False) as bar:
             bar.short_limit = 0
             for deployer_class in self.deployer_classes:
                 if interactive and not first_iteration:
```
derekpierre: Since we are prompting before each contract and providing the name of the contract, do we want to prompt before the deployment of the first iteration contract "NuCypherToken"? As is, we deploy "NuCypherToken" contract without prompting, but then prompt for the others. Perhaps we don't - but just checking.

nucypher/blockchain/eth/deployers.py
```diff
@@ -142,7 +144,7 @@ class NucypherTokenDeployer(ContractDeployer):
 
     agency = NucypherTokenAgent
     contract_name = agency.registry_contract_name
-    number_of_deployment_transactions = 1
+    deployment_steps = ('contract_deployment', )
```
derekpierre: 👍

tests/blockchain/eth/entities/deployers/test_adjudicator_deployer.py
```diff
@@ -46,12 +46,11 @@ def test_adjudicator_deployer(session_testerchain, slashing_economics, deploymen
...
 
-    assert len(deployment_receipts) == 3
     # deployment steps must match expected number of steps
-    assert deployment_progress.num_steps == deployer.number_of_deployment_transactions
+    assert deployment_progress.num_steps == len(deployer.deployment_steps) == len(deployment_receipts) == 3
 
-    for title, receipt in deployment_receipts.items():
-        assert receipt['status'] == 1
+    for step in deployer.deployment_steps:
+        assert deployment_receipts[step]['status'] == 1
```
derekpierre: 🎉

nucypher/cli/deploy.py
```diff
@@ -140,7 +150,7 @@ def deploy(action,
 
     password = None
     if not hw_wallet and not blockchain.client.is_local:
-        password = get_nucypher_password(confirm=False)
+        password = get_client_password(confirm=False)
```
derekpierre: @cygnusv `get_client_password` takes a `checksum` as a parameter and not `confirm`. Did you mean `get_nucypher_password`?
I'm getting the following traceback:
```
  File "/Users/derek/.local/share/virtualenvs/nucypher-M0kpWrFa/bin/nucypher-deploy", line 11, in <module>
    load_entry_point('nucypher', 'console_scripts', 'nucypher-deploy')()
  File "/Users/derek/.local/share/virtualenvs/nucypher-M0kpWrFa/lib/python3.7/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/Users/derek/.local/share/virtualenvs/nucypher-M0kpWrFa/lib/python3.7/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/Users/derek/.local/share/virtualenvs/nucypher-M0kpWrFa/lib/python3.7/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/derek/.local/share/virtualenvs/nucypher-M0kpWrFa/lib/python3.7/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/Users/derek/Documents/Github/repos/forks/david/nucypher/nucypher/cli/deploy.py", line 153, in deploy
    password = get_client_password(confirm=False)
  File "/Users/derek/Documents/Github/repos/forks/david/nucypher/nucypher/blockchain/eth/decorators.py", line 37, in wrapped
    params = inspect.getcallargs(func, *args, **kwargs)
  File "/usr/local/Cellar/python/3.7.3/Frameworks/Python.framework/Versions/3.7/lib/python3.7/inspect.py", line 1358, in getcallargs
    (f_name, kw))
TypeError: get_client_password() got an unexpected keyword argument 'confirm'
```

cygnusv: Fixed, thanks!

nucypher/blockchain/eth/actors.py
cygnusv: I personally found it weird to type `"DEPLOY"`, waiting the countdown,... and then being interrupted with the requirement of pressing a key. We do however print `"Deploying NuCypherToken ..."`, so it's not like the user doesn't know what's happening.

nucypher/cli/stake.py
```diff
@@ -164,7 +180,54 @@ def stake(click_config,
                                          password=password,
                                          worker_address=worker_address)
 
-        emitter.echo(f"OK | Receipt: {receipt['transactionHash'].hex()}", color='green')
+        # TODO: Double-check dates
+        current_period = STAKEHOLDER.staking_agent.get_current_period()
+        bounded_date = datetime_at_period(period=current_period)
```
mswilkison: is it meant to be bonded?

nucypher/cli/actions.py
```diff
@@ -305,15 +310,13 @@ def select_client_account(emitter, blockchain, prompt: str = None, default=0) ->
...
-        confirmed_chain_id = int(click.prompt("Enter the Chain ID to confirm deployment", type=click.INT))
-        expected_chain_id = int(deployer.blockchain.client.chain_id)
-        if confirmed_chain_id != expected_chain_id:
-            abort_message = f"Chain ID not a match ({confirmed_chain_id} != {expected_chain_id}) Aborting Deployment"
-            emitter.echo(abort_message, fg='red', bold=True)
-            raise click.Abort()
+        expected_chain_name = deployer.blockchain.client.chain_name
+
+    if click.prompt(f"Type '{expected_chain_name}' to continue") != expected_chain_name:
+        emitter.echo("Aborting Deployment", fg='red', bold=True)
```
fjarri: The keyword for coloring is `color`, not `fg` - this must be a remainder I forgot to fix when in the emitter refactoring.

nucypher/cli/config.py
```diff
@@ -133,6 +134,7 @@ def set_options(self,
         self.no_registry = no_registry
         self.debug = debug
         self.json_ipc = json_ipc
+        self.etherscan = etherscan
```
fjarri: Is it used anywhere?

nucypher/cli/painting.py
```diff
@@ -301,22 +303,61 @@ def paint_staged_stake_division(emitter,
...
+                              transaction_type=tx_name)
+
+    if open_in_browser:
+        try:
+            url = etherscan_url(item=contract_address,
+                                network=chain_name,
+                                is_token=is_token_contract)
+        except ValueError as e:
+            emitter.log.info("Failed Etherscan URL construction: " + str(e))
+        else:
```
michwill: I would try to import `webbrowser` right here (in case we deploy from a commandline-only remote machine)

nucypher/cli/stake.py
michwill: I think, `bounded`?

cygnusv: Ah very confusing: there's `bonded`, `bound`, and `bounded`. But yes, it is meant to be "bonded".

nucypher/cli/config.py
cygnusv: Yes, currently only in `nucypher-deploy contracts` as a proof of concept, but the idea is to use it in other commands too, but it needs your opinion :)

nucypher/cli/painting.py
cygnusv: I tried in my remote machine, and there's no problem at all. It simply outputs `False`
```python
>>> webbrowser.open_new_tab("https://github.com/")
False
```

nucypher/cli/actions.py
cygnusv: Good catch! Fixed.

nucypher/cli/painting.py
michwill: Huh, didn't realize it's a built-in module. I think, all good in that case

nucypher/cli/config.py
fjarri: My understanding was that there are two `etherscan` options defined: one in `cli/deploy.py::deploy()` (which is used), and the other in `cli/config.py::nucypher_click_config()` (which isn't). I think it might be better to postpone adding the latter one until some code actually uses `.etherscan` attribute of `NucypherClickConfig`.

PR #1181, opened 2019-08-01
Update Dependency Locks

KPrasch:


KPrasch:
WIth this lock of dependencies installed I am currently getting `cannot import name 'validate_address' from 'ethpm.validation' `

Issue #1199, opened 2019-08-06
Make it possible to work without `--no-sync` if Ethereum node was started before


michwill:
It's very likely that Ethereum node was synced beforehand. Currently, we fail to detect it, and all the commands have to ignore the detection logic using `--no-sync`.

Here's how we can determine if the node was synced:
```
if time.time() - w3.eth.getBlock('latest')['timestamp'] < 100 and not w3.eth.syncing:
    print('Yay, we are synced!')
else:
    print('Not so much')
```

vzotova:
Closed in #1203

PR #1203, opened 2019-08-07
Improved Blockchain Syncing UX

vepkenez:
Fix and test for https://github.com/nucypher/nucypher/issues/1199
Also adds nice progress bar to report syncing progress on all CLI sync scenarios.

### DIFF
--- a/nucypher/blockchain/eth/clients.py
+++ b/nucypher/blockchain/eth/clients.py
@@ -263,11 +263,14 @@ class Web3Client:
                 check_for_timeout(t=self.SYNC_TIMEOUT_DURATION*2)
 
             while True:
+                #  TODO:  Should this timeout eventually?
                 syncdata = self.syncing
                 if not syncdata:
-                    break
+                    raise StopIteration
+
                 self.log.info(f"Syncing {syncdata['currentBlock']}/{syncdata['highestBlock']}")
                 time.sleep(self.SYNC_SLEEP_DURATION)
+                yield syncdata
 
         return True
 
--- a/nucypher/blockchain/eth/clients.py
+++ b/nucypher/blockchain/eth/clients.py
@@ -2,6 +2,7 @@ import json
 import os
 import shutil
 import time
+import datetime
 from typing import Union
 
 import maya
@@ -229,27 +230,37 @@ class Web3Client:
         now = maya.now()
         start_time = now
 
+        def has_latest_block():
+            # check that our local chain data is up to date
+            return (
+                datetime.datetime.now() -
+                datetime.datetime.fromtimestamp(
+                    self.w3.eth.getBlock(self.w3.eth.blockNumber)['timestamp']
+                )
+            ).seconds < 30
+
         def check_for_timeout(t):
             last_update = maya.now()
             duration = (last_update - start_time).seconds
             if duration > t:
                 raise self.SyncTimeout
 
-        # Check for ethereum peers
-        self.log.info(f"Waiting for Ethereum peers ({len(self.peers)} known)")
-        while not self.peers:
-            time.sleep(0)
-            check_for_timeout(t=60)
+        while not has_latest_block():
+            # Check for ethereum peers
+            self.log.info(f"Waiting for Ethereum peers ({len(self.peers)} known)")
+            while not self.peers:
+                time.sleep(0)
+                check_for_timeout(t=60)
 
-        # Wait for sync start
-        self.log.info(f"Waiting for {self.chain_name.capitalize()} chain synchronization to begin")
-        while not self.syncing:
-            time.sleep(0)
-            check_for_timeout(t=120)
+            # Wait for sync start
+            self.log.info(f"Waiting for {self.chain_name.capitalize()} chain synchronization to begin")
+            while not self.syncing:
+                time.sleep(0)
+                check_for_timeout(t=120)
 
-        while self.syncing:
-            self.log.info(f"Syncing {self.syncing['currentBlock']}/{self.syncing['highestBlock']}")
-            time.sleep(5)
+            while self.syncing:
+                self.log.info(f"Syncing {self.syncing['currentBlock']}/{self.syncing['highestBlock']}")
+                time.sleep(5)
 
         return True
 
--- a/nucypher/blockchain/eth/clients.py
+++ b/nucypher/blockchain/eth/clients.py
@@ -70,6 +70,15 @@ class Web3Client:
     ALT_PARITY = 'Parity-Ethereum'
     GANACHE = 'EthereumJS TestRPC'
     ETHEREUM_TESTER = 'EthereumTester'  # (PyEVM)
+    SYNC_TIMEOUT_DURATION = 60 # seconds to wait for various blockchain syncing endeavors
+    SYNC_SLEEP_DURATION = 5
+
+    class ConnectionNotEstablished(RuntimeError):
+        pass
+
+    class SyncTimeout(RuntimeError):
+        pass
+
 
     def __init__(self,
                  w3,
@@ -131,12 +140,6 @@ class Web3Client:
 
         instance = ClientSubclass(w3, **client_kwargs)
         return instance
-
-    class ConnectionNotEstablished(RuntimeError):
-        pass
-
-    class SyncTimeout(RuntimeError):
-        pass
 
     @property
     def peers(self):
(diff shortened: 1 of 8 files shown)


nucypher/blockchain/eth/clients.py
```diff
@@ -229,27 +230,37 @@ def sync(self,
...
         start_time = now
 
+        def has_latest_block():
+            # check that our local chain data is up to date
+            return (
+                datetime.datetime.now() -
+                datetime.datetime.fromtimestamp(
+                    self.w3.eth.getBlock(self.w3.eth.blockNumber)['timestamp']
+                )
+            ).seconds < 30
```
michwill: I think this is wrong. (1 day + 1 second).seconds = 1, and it is clearly less than 30 (but the timedelta is more than 1 day!).

Why not do `time.time() - ...['timestamp'] < 30`?
(alternatively, you can do `timedelta(seconds=30)` and compare the object with that

michwill: Besides, I'd also check that `self.w3.eth.syncing` is False in addition to this

vepkenez: that gets checked further down

vepkenez: The issue this solves (assuming correct time math) is that we weren't checking for _synced_ chains, so the `while not self.syncing:` was hanging.

michwill: Yep, right

tests/blockchain/eth/clients/test_mocked_clients.py
```diff
@@ -126,3 +202,65 @@ def test_ganache_web3_client():
     assert interface.client.platform is None
     assert interface.client.backend == 'ethereum-js'
     assert interface.client.is_local
+
+
+def test_datetime_logic():
```
vepkenez: @michwill have a look at this test.  I think it proves that this datetime logic works as I originally assumed

tests/blockchain/eth/clients/test_mocked_clients.py
```diff
@@ -50,14 +116,24 @@ def clientVersion(self):
...
 
-#
+
+class SyncingMockWeb3(SyncedMockWeb3):
+
+    net = SyncingMockW3Eth()
+    eth = SyncingMockW3Eth()
+
+
+class SyncingMockWeb3Nopeers(SyncingMockWeb3):
```
mswilkison:
```suggestion
class SyncingMockWeb3NoPeers(SyncingMockWeb3):
```

tests/blockchain/eth/clients/test_mocked_clients.py
```diff
@@ -126,3 +202,65 @@ def test_ganache_web3_client():
...
+        assert interface.client.sync()
+
+
+def test_no_peers_unsynced_geth_client():
+
+    GethClient.SYNC_TIMEOUT_DURATION = 2
+
+    class NonSyncedNoPeersBlockchainInterface(GethClientTestBlockchain):
+
+        Web3 = SyncingMockWeb3Nopeers
```
mswilkison:
```suggestion
        Web3 = SyncingMockWeb3NoPeers
```

nucypher/blockchain/eth/clients.py
```diff
@@ -217,6 +222,15 @@ def send_transaction(self, transaction: dict) -> str:
...
         return self.w3.eth.sendRawTransaction(raw_transaction=transaction)
 
+    def _has_latest_block(self):
+        # check that our local chain data is up to date
+        return (
+            datetime.datetime.now() -
+            datetime.datetime.fromtimestamp(
+                self.w3.eth.getBlock(self.w3.eth.blockNumber)['timestamp']
+            )
+        ).seconds < 30
```
michwill: Looks like the comment got lost, so repeating here. This one doesn't work properly because this condition is satisfied also for "1 day and 3 seconds" or "10 days and 5 seconds". Replace with:
```
            (datetime.datetime.now() -
            datetime.datetime.fromtimestamp(
                self.w3.eth.getBlock(self.w3.eth.blockNumber)['timestamp'])) < timedelta(seconds=30)
```
or
```
time.time() - self.w3.eth.getBlock(self.w3.eth.blockNumber)['timestamp'] < 30
```

tests/blockchain/eth/clients/test_mocked_clients.py
derekpierre: @vepkenez I think if you change the test, you might get a different result - which is what @michwill is pointing out:

```
In [19]: now = datetime.datetime.now()

In [20]: now
Out[20]: datetime.datetime(2019, 8, 8, 8, 36, 14, 323883)

In [21]: past_date = now - datetime.timedelta(days=3,seconds=3)

In [22]: past_date
Out[22]: datetime.datetime(2019, 8, 5, 8, 36, 11, 323883)

In [23]: (now - past_date).seconds
Out[23]: 3
```

nucypher/blockchain/eth/clients.py
```diff
@@ -229,27 +243,35 @@ def sync(self,
...
-
-        while self.syncing:
-            self.log.info(f"Syncing {self.syncing['currentBlock']}/{self.syncing['highestBlock']}")
-            time.sleep(5)
+        while not self._has_latest_block():
+            # Check for ethereum peers
+            self.log.info(f"Waiting for Ethereum peers ({len(self.peers)} known)")
+            while not self.peers:
+                time.sleep(0)
+                check_for_timeout(t=self.PEERING_TIMEOUT)
```
derekpierre: @vepkenez did you intend for this timeout to go from a 60s timeout previously to a 5s timeout now (self.PEERING_TIMEOUT)?

nucypher/blockchain/eth/interfaces.py
```diff
@@ -249,14 +251,48 @@ def connect(self, fetch_registry: bool = True, sync_now: bool = False):
...
+                try:
+                    emitter.echo("Beginning sync...")
+                    initial_state = next(sync_state)
+                except StopIteration:  # will occur if no syncing needs to happen
+                    emitter.echo("Local blockchain data is already synced.")
+                    return True
+
+                prior_state = initial_state
+                total_blocks_to_sync = int(initial_state.get('highestBlock', 0)) - int(initial_state.get('currentBlock', 0))
+                with click.progressbar(
```
derekpierre: 👍 . Also, if you don't want ETA to be shown with the `progressbar` because of variability in syncing of blocks, you could also pass `show_eta=False`

tests/blockchain/eth/clients/test_mocked_clients.py
derekpierre: Using `total_seconds()` may be the way to go:
```
In [26]: (now - past_date).total_seconds()
Out[26]: 259203.0
```

nucypher/blockchain/eth/clients.py
derekpierre: I think using `.total_seconds()` instead of `.seconds` may solve the problem. See my comment below in the unit test file and https://docs.python.org/3/library/datetime.html

tests/blockchain/eth/clients/test_mocked_clients.py
vepkenez: Ah yes.  I just got lucky there... total_seconds it is!

nucypher/blockchain/eth/interfaces.py
vepkenez: In testing yesterday it is actually turned out to be very accurate, other than speeding up a bit at the end.

tests/blockchain/eth/clients/test_mocked_clients.py
michwill: That's another way, yes.
Although it doesn't quite matter: why datetime objects at all!

nucypher/blockchain/eth/clients.py
vepkenez: I don't think this timeout stuff is working at all right now.  Still working on that.

nucypher/blockchain/eth/clients.py
```diff
@@ -217,6 +221,21 @@ def send_transaction(self, transaction: dict) -> str:
...
+        eth-tester signing interface to do so.
+        """
+        return self.w3.eth.sign(account, data=message)
+
+    def _has_latest_block(self):
+        # check that our local chain data is up to date
+        return (
+            time.time() -
+            self.w3.eth.getBlock(self.w3.eth.blockNumber)['timestamp']
+        ) < 30
```
michwill: :+1:

PR #1222, opened 2019-08-13
Cleanup and test chain_name/chain_id

vepkenez:


nucypher/blockchain/eth/clients.py
```diff
@@ -188,7 +188,7 @@ def inject_middleware(self, middleware, **kwargs):
 
     @property
     def chain_id(self) -> int:
-        return int(self.w3.eth.chainId, 16)
+        return int(self.w3.eth.chainId)
```
cygnusv: The `web3.py` docs say the value is already an integer, so perhaps we should get rid of that casting directly. 
https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.chainId
```suggestion
        return self.w3.eth.chainId
```

cygnusv: Oh nevermind, it seems it's a hex string after all: `ValueError: invalid literal for int() with base 10: '0x3d'`

This means web3.py docs are wrong...

vepkenez: ah so the int() is just to base it 16... putting that back.

vepkenez: one weird thing is that in @mswilkison's gist, it is also casting to string which I don't see anywhere in master or horchata.  Maybe it's a different branch?

cygnusv: For completeness, the original JSON-RPC endpoint return a [hex string](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-695.md#abstract). Web3.py docs imply they parse it, but apparently, they don't.

vepkenez: yes, there are also variations between some of the providers sometimes that aren't documented.

derekpierre: This may be a versioning issue (?)
```
v5.0.0
Released August 1, 2019

Features
web3.eth.chainId now returns an integer instead of hex - #1394
```

https://github.com/ethereum/web3.py/pull/1394

Perhaps @mswilkison was using a newer version that you guys?

mswilkison: I'm v1.9.1-stable

derekpierre: @mswilkison the version of web3.py

Try running:
```
pipenv graph | grep "web3"
```

cygnusv: Ahh, I saw the error in our CI build, which is currently using `v5.0.0b2` (a prerelease), and this was fixed in a recent stable release (`v5.0.0`). So I guess we need to update the `Pipfile.lock` file to the stable version of this package.

mswilkison: Oh, sorry read to fast! v5.0.0b2 which is what's in requirements.txt

vepkenez: wow.  well our version is pinned to one older than that so we will need to update this code when the time comes.

michwill: Or `python -c "import web3; print(web3.__version__)"`

vepkenez: maybe @mswilkison installed with `--skip-lock`?

PR #1241, opened 2019-08-20
Add nucypher version to Ursula status page

### LINKED ISSUE #1235: Include version information in node status page
We need to know the version of nucypher being run by a given node by browsing to its status page.  Ideally, we also want this information available as JSON so that we can aggregate it.  It may even be useful to simply include it in all node metadata so that we can include it on `Moe`.

cygnusv:
Fixes #1235

### DIFF
--- a/nucypher/network/templates/basic_status.j2
+++ b/nucypher/network/templates/basic_status.j2
@@ -105,6 +105,7 @@
 <div id="this-node">
     <h2>{{ this_node.nickname}}</h2>
     {{ this_node.nickname_icon }}
+    <h4>v{{ version }}</h4>
     <h4>Domains: {% for domain in domains %}{{ domain }} {% endfor %}</h4>
 
     <h3>Fleet State</h3>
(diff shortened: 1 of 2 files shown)


derekpierre:
Adding my comment from Discord here:
@cygnusv  what do you think of `nucypher version: 0.1.0-alpha.24` -> `v0.1.0-alpha.24` and put it above the "Domains" line?

tuxxy:
@derekpierre I'm +1 on that change, waiting for @cygnusv to finish lunch.

nucypher/network/templates/basic_status.j2
```diff
@@ -105,6 +105,7 @@
 <div id="this-node">
     <h2>{{ this_node.nickname}}</h2>
     {{ this_node.nickname_icon }}
+    <h4>{{ version }}</h4>
```
derekpierre: Add a `v`?

```suggestion
    <h4>v{{ version }}</h4>
```

PR #1259, opened 2019-08-24
Add staker info tools to `nucypher status`

cygnusv:
_If you have ideas to improve these commands (or new commands), don't be shy! Maybe I can add them._

Usage:
* To see evolution of overall locked tokens of the network for next `N` days (default is 90): 
```nucypher status locked-tokens --provider <GETH> --poa [--periods N]```
image

* To see all stakers info (optionally a staking address can be supplied): 
```nucypher status stakers --provider <GETH> --poa [--staking-address 0xfoobar]```
image

Issue #1061, opened 2019-06-08
Make sample selection buffer configurable


KPrasch:
See: <https://github.com/nucypher/nucypher/pull/984/files#diff-5b3bd97d75d539a8b1311b82d491cb58R119>

KPrasch:
@fjarri - Considering `WeightedSampler`'s design, what do you think about exposing any part sample selection to user configuration / CLI flag, or instead closing this issue?

fjarri:
If I understand correctly, the analogous logic is now contained in `policies.py::PrefetchStrategy`. I don't think we need to make it configurable right now, since with the parallel proposals it is stable enough as it is, as long as a significant part of nodes is available. Here is a simple simulation for 2000 total nodes, assuming a proposal takes 1s and the timeout is 20s:

`n=3`:
![t_3](https://user-images.githubusercontent.com/30720/109410439-97bc2800-794f-11eb-9f8a-605e09c4ae67.png)

`n=5`:
![t_5](https://user-images.githubusercontent.com/30720/109410441-9c80dc00-794f-11eb-941c-8ca094e62ec5.png)

(code: https://gist.github.com/fjarri/e8584987e40477108b83de5e48dad763)

PR #1280, opened 2019-08-28
Touches to StakingEscrow.sample()

### LINKED ISSUE #1019: Test correct probability distribution of MinerAgent.sample()
Find a way to test that the sampling probability distribution produced by MinerAgent.sample() is consistent with the expected behavior (i.e. the probability of a node being selected is proportional to its stake).

### LINKED ISSUE #1090: For N>=4, test_blockchain_ursulas_reencrypt fails with certain probability
After rebasing alhambra-verde, these tests were [failing for no apparent reason](https://circleci.com/gh/nucypher/nucypher/33609#tests/containers/3). Both are from `tests/characters/test_ursula_prepares_to_act_as_mining_node.py`. I had to rerun the build twice, but in the end the tests passed without changing anything 🤷‍♂ 

For the first test, the issue may be related to the sampling parameters (see #1061).
For the second test...no idea.

cygnusv:
Corrects a minor bug in `StakingEscrow.sample()` that was affecting sampling (Fixes #1090)

Also adds a test that checks that sampling produces an acceptable distribution (Closes #1019)

### DIFF
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -359,7 +359,7 @@ class StakingEscrowAgent(EthereumContractAgent):
         stakers_population = self.get_staker_population()
         n_select = math.ceil(quantity * additional_ursulas)  # Select more Ursulas
         if n_select > stakers_population:
-            raise self.NotEnoughStakers(f'There are {stakers_population} published stakers, need a total of {n_select}.')
+            raise self.NotEnoughStakers(f'There are {stakers_population} active stakers, need at least {n_select}.')
 
         system_random = random.SystemRandom()
         n_tokens = self.contract.functions.getAllLockedTokens(duration).call()
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -367,13 +367,10 @@ class StakingEscrowAgent(EthereumContractAgent):
             raise self.NotEnoughStakers('There are no locked tokens for duration {}.'.format(duration))
 
         for _ in range(attempts):
-            points = [0] + sorted(system_random.randrange(n_tokens) for _ in range(n_select))
+            points = sorted(system_random.randrange(n_tokens) for _ in range(n_select))
+            self.log.debug(f"Sampling {n_select} stakers with random points: {points}")
 
-            deltas = []
-            for next_point, previous_point in zip(points[1:], points[:-1]):
-                deltas.append(next_point - previous_point)
-
-            addresses = set(self.contract.functions.sample(deltas, duration).call())
+            addresses = set(self.contract.functions.sample(points, duration).call())
             addresses.discard(str(BlockchainInterface.NULL_ADDRESS))
 
             if len(addresses) >= quantity:
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -16,6 +16,7 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
 
+import math
 import random
 from typing import Generator, List, Tuple, Union
 
@@ -356,13 +357,12 @@ class StakingEscrowAgent(EthereumContractAgent):
         """
 
         stakers_population = self.get_staker_population()
-        if quantity > stakers_population:
-            raise self.NotEnoughStakers(f'There are {stakers_population} published stakers, need a total of {quantity}.')
+        n_select = math.ceil(quantity * additional_ursulas)  # Select more Ursulas
+        if n_select > stakers_population:
+            raise self.NotEnoughStakers(f'There are {stakers_population} published stakers, need a total of {n_select}.')
 
         system_random = random.SystemRandom()
-        n_select = round(quantity*additional_ursulas)            # Select more Ursulas
         n_tokens = self.contract.functions.getAllLockedTokens(duration).call()
-
         if n_tokens == 0:
             raise self.NotEnoughStakers('There are no locked tokens for duration {}.'.format(duration))
 
--- a/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
@@ -839,16 +839,20 @@ contract StakingEscrow is Issuer {
 
     /**
     * @notice Get active stakers based on input points
-    * @param _points Array of absolute values
+    * @param _points Array of absolute values. Must be sorted in ascending order.
     * @param _periods Amount of periods for locked tokens calculation
     *
-    * @dev Sampling iterates over an array of stakers and input points.
-    * Each iteration checks if the current point is contained within the current staker stake.
-    * If the point is greater than or equal to the current sum of stakes,
-    * this staker is skipped and the sum is increased by the value of next staker's stake.
-    * If a point is less than the current sum of stakes, then the current staker is appended to the resulting array.
-    * Secondly, the sum of stakes is decreased by a point;
-    * The next iteration will check the next point for the difference.
+    * @dev This method implements the Probability Proportional to Size (PPS) sampling algorithm,
+    * but with the random input data provided in the _points array.
+    * In few words, the algorithm places in a line all active stakes that have locked tokens for
+    * at least _periods periods; a staker is selected if an input point is within its stake.
+    * For example:
+    *
+    * Stakes: |----- S0 ----|--------- S1 ---------|-- S2 --|---- S3 ---|-S4-|----- S5 -----|
+    * Points: ....R0.......................R1..................R2...............R3...........
+    *
+    * In this case, Stakers 0, 1, 3 and 5 will be selected.
+    *
     * Only stakers which confirmed the current period (in the previous period) are used.
     * If the number of points is more than the number of active stakers with suitable stakes,
     * the last values in the resulting array will be zeros addresses.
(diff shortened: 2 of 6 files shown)


nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -865,28 +865,22 @@ contract StakingEscrow is Issuer {
...
                 stakerIndex += 1;
-                addMoreTokens = true;
                 continue;
             }
-            if (addMoreTokens) {
-                sumOfLockedTokens = sumOfLockedTokens.add(getLockedTokens(info, currentPeriod, nextPeriod));
-            }
-            if (sumOfLockedTokens > point) {
+            uint256 stakerTokens = getLockedTokens(info, currentPeriod, nextPeriod);
+            if (sumOfLockedTokens + stakerTokens > point) {
```
michwill: Need to think if we need also `sumOfLockedTokens <= point`, otherwise all good

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -862,31 +866,30 @@ contract StakingEscrow is Issuer {
...
                 stakerIndex += 1;
-                addMoreTokens = true;
                 continue;
             }
-            if (addMoreTokens) {
-                sumOfLockedTokens = sumOfLockedTokens.add(getLockedTokens(info, currentPeriod, nextPeriod));
-            }
-            if (sumOfLockedTokens > point) {
+            uint256 stakerTokens = getLockedTokens(info, currentPeriod, nextPeriod);
+            uint256 nextSumValue = sumOfLockedTokens.add(stakerTokens);
```
michwill: Interesting. So, instead of post-selecting points which don't belong to one staker, you switch to the next staker if you selected a point. I think, it makes sense (in the limit of `stake << point_distance`), so I'm ok to do it this way. But also good run the simulations you had

michwill: I'm not 100% convinced though that it won't distort the probability distribution too much. It won't break things, so I am +1 on merging. But need to do tests for sure (comparing it with the situation when sample can select multiple points per staker

nucypher/blockchain/eth/agents.py
```diff
@@ -356,24 +357,20 @@ def sample(self, quantity: int, duration: int, additional_ursulas: float = 1.7,
         """
 
         stakers_population = self.get_staker_population()
-        if quantity > stakers_population:
-            raise self.NotEnoughStakers(f'There are {stakers_population} published stakers, need a total of {quantity}.')
+        n_select = math.ceil(quantity * additional_ursulas)  # Select more Ursulas
+        if n_select > stakers_population:
+            raise self.NotEnoughStakers(f'There are {stakers_population} active stakers, need at least {n_select}.')
```
derekpierre: Should we clarify that we need a `sample` that may be larger than the quantity specified. As a user I may be confused that I get a `NotEnoughStakers` exception if I specify `n` = 6 and there are 6 `active stakers` that randomly says some value eg. `need at least 9` (assuming `additional_ursulas` is 1.5)

Perhaps our message could be:
```suggestion
            raise self.NotEnoughStakers(f'There are {stakers_population} active stakers - for {quantity} stakers we need a sample size of at least {n_select}.')
```

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -839,16 +839,20 @@ contract StakingEscrow is Issuer {
...
-    * Secondly, the sum of stakes is decreased by a point;
-    * The next iteration will check the next point for the difference.
+    * @dev This method implements the Probability Proportional to Size (PPS) sampling algorithm,
+    * but with the random input data provided in the _points array.
+    * In few words, the algorithm places in a line all active stakes that have locked tokens for
+    * at least _periods periods; a staker is selected if an input point is within its stake.
+    * For example:
+    *
+    * Stakes: |----- S0 ----|--------- S1 ---------|-- S2 --|---- S3 ---|-S4-|----- S5 -----|
+    * Points: ....R0.......................R1..................R2...............R3...........
```
derekpierre: 👍

nucypher/blockchain/eth/agents.py
cygnusv: Yes, I think that's a good suggestion. Thanks @derekpierre !

nucypher/policy/policies.py
```diff
@@ -378,23 +378,16 @@ def sample_essential(self, quantity: int, handpicked_ursulas: Set[Ursula] = None
...
-
-        if actual_sample_quantity > 0:
-            selected_ursulas = self.sample_essential(quantity=actual_sample_quantity,
-                                                     handpicked_ursulas=handpicked_ursulas)
-            handpicked_ursulas.update(selected_ursulas)
+        target_sample_quantity = self.n - len(selected_ursulas)
+        if target_sample_quantity > 0:
+            sampled_ursulas = self.sample_essential(quantity=target_sample_quantity,
+                                                    handpicked_ursulas=handpicked_ursulas)
+            selected_ursulas.update(sampled_ursulas)
```
tuxxy: :+1:

tests/blockchain/eth/contracts/main/staking_escrow/test_sampling.py
```diff
@@ -43,16 +42,18 @@ def test_sampling(testerchain, token, escrow_contract):
         testerchain.wait_for_receipt(tx)
         amount = amount // 2
 
-    # Cant't use sample without points or with zero periods value
-    with pytest.raises((TransactionFailed, ValueError)):
+    # Can't use sample without points, with zero periods value, or not sorted in ascending order.
+    with pytest.raises(TransactionFailed):
         escrow.functions.sample([], 1).call()
-    with pytest.raises((TransactionFailed, ValueError)):
```
KPrasch: Just a note - By removing the handling of `ValueError`, the test suite is less compatible with other backends, like geth, since the exception is EthereumTester-specific.   Unsure if this is a reasonable goal for us, but I'd like to remain open to the possibility of running the test suite without eth-tester.

nucypher/blockchain/eth/agents.py
KPrasch: Does `additional_ursulas` still make sense as a static value?  For such small m of n values, a proportional, smaller value might be more usable.

tests/blockchain/eth/contracts/main/staking_escrow/test_sampling_distribution.py
```diff
@@ -0,0 +1,111 @@
...
+
+    #
+    # Test sampling distribution
+    #
+
+    ERROR_TOLERANCE = 0.05  # With this tolerance, all sampling ratios should between 5% and 15% (expected is 10%)
+    SAMPLES = 100
+    quantity = 3
+    import random
+    from collections import Counter
```
KPrasch: Wondeful :+1:

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -862,31 +866,30 @@ contract StakingEscrow is Issuer {
...
             }
-            if (addMoreTokens) {
-                sumOfLockedTokens = sumOfLockedTokens.add(getLockedTokens(info, currentPeriod, nextPeriod));
-            }
-            if (sumOfLockedTokens > point) {
+            uint256 stakerTokens = getLockedTokens(info, currentPeriod, nextPeriod);
+            uint256 nextSumValue = sumOfLockedTokens.add(stakerTokens);
+
+            uint256 point = _points[pointIndex];
+            require(point >= previousPoint);  // _points must be a sorted array
```
KPrasch: :+1:

KPrasch:
I've reviewed this changeset, however I'd like to go through it with @szotov before giving approval.

nucypher/policy/policies.py
```diff
@@ -378,23 +378,16 @@ def sample_essential(self, quantity: int, handpicked_ursulas: Set[Ursula] = None
...
 
     def sample(self, handpicked_ursulas: Set[Ursula] = None) -> Set[Ursula]:
-        if not handpicked_ursulas:
-            handpicked_ursulas = set()
-        else:
-            handpicked_ursulas = set(handpicked_ursulas)
+        selected_ursulas = set(handpicked_ursulas) if handpicked_ursulas else set()
 
         # Calculate the target sample quantity
-        ADDITIONAL_URSULAS = self.selection_buffer
```
szotov: Just to clarify - no more selection buffer for federated policy, right?

nucypher/blockchain/eth/agents.py
cygnusv: @derekpierre sorry I didn't include this in #1280. I'll do that in #1259.
@KPrasch Although in the method signature there's a static default, in reality it's always called explicitly with the `BlockchainPolicy.selection_buffer` value as defined in `BlockchainPolicy.sample_essential()`. We however don't configure `BlockchainPolicy.selection_buffer` yet, so for the moment it always uses the default value of `1.5`. That's why #1061 is still open.

tests/blockchain/eth/contracts/main/staking_escrow/test_sampling.py
cygnusv: Oh so that's why! I always found weird those lines. In any case, if that's the reason, isn't it better that we have something like:

```python
TransactionException = TransactionFailed if using_eth_tester() else ValueError
...
with pytest.raises(TransactionException):
    foo.bar()
```

Otherwise, there may be the case that in our current setting the method being tested is currently raising a legitimate`ValueError` not caused by a failed transaction and we're accepting that as valid.

PR #1295, opened 2019-09-04
Resolve issues running local development fleet scripts/files

derekpierre:
Some of our local fleet scripts/files have had some neglect and were not working.


derekpierre:
This was on my Mac. I don't really use the solc script since last I checked only the linux executable was supported.

PR #1296, opened 2019-09-04
Relock Dependencies; Use Pip throughout CI

### LINKED ISSUE #997: Reuse pip installation environment in CI workflow
We can reduce 1-2 minutes the overall CI workflow by reusing the pip installation environment when necessary.

KPrasch:
##### Chagelog

- Includes updates to pyumbral, hendrix, and py-evm.  
- Do not update py-evm beyond v0.3.0a1.
- Deprecates `boto`, `moto`, Introduces `py-geth`
- Accommodates upstream chain ID reporting bug fixes
- Re-implements `Mock`s as `mocker.Mock`
- Use Pip-based user installation in circleCI workflows
- Use circleCI dependency caching for workflow persistence
- Fixes missing zeppelin package data
- Fixes #997 
- Speeds up Workflows

##### Notes 

-  Using caching is a trade-off between reliability and performance.  One new potential issue is that contracts are treated like dependencies, stored in site-packages, so it is possible that multiple subsequent runs after making edits to smart contract code will result in the need to manually increment the cache key version.

- Install by running `pip install -e . -r dev-requirements` or `pip install -e .`

-  Despite depending on the `setup.py` integration for click CLI entrypoints, Pipenv is still relevant because we use it as a means of generating compatible `requirements.txt` / `dev-requirements.txt`, and use the `Pipfile.lock` a representation of *all* dependencies in one file - used by checksum-digest for dependency cache invalidation on circleCI.


nucypher/blockchain/eth/clients.py
```diff
@@ -189,7 +189,12 @@ def inject_middleware(self, middleware, **kwargs):
 
     @property
     def chain_id(self) -> int:
-        return int(self.w3.eth.chainId, 16)
+        try:
+            # from hex-str
+            return int(self.w3.eth.chainId, 16)
+        except TypeError:
```
michwill: :+1:

tests/blockchain/eth/contracts/integration/test_intercontract_integration.py
```diff
@@ -160,15 +159,15 @@ def worklock(testerchain, token, escrow, deploy_contract):
     return contract
 
 
-def mock_ursula(testerchain, account):
+def mock_ursula(testerchain, account, mocker):
```
cygnusv: Just curious: what's the advantage of using mocking via pytest-mock spy?

cygnusv:
Just to summarize my post above, current changes to CI are introducing both false negatives and false positives in our tests:
* False negatives: when there's broken logic but the CI tests doesn't catch it (e.g. [this build that had a broken import but the build "was OK"](https://circleci.com/workflow-run/45d09b1e-3a83-46b4-97a7-153b4ea44e31))
* False positives: when the logic is good but the CI tests fail (e.g., [current CI build that's using old code](https://circleci.com/workflow-run/51879d5a-0c5a-4235-bc46-a62bfd7bf3ca))

nucypher/blockchain/eth/clients.py
derekpierre: @KPrasch do we need to be careful here? - in newer versions chainId could already be in decimal: See https://github.com/ethereum/web3.py/pull/1394. See release note for v5.0.0

cygnusv: I think the intention is to be backwards-compatible

derekpierre: Ah right!

KPrasch:
Good Catch - I think I've unintentionally introduced source caching, by using a standard pip installation.  I added the `-e` flag to try and avoid this condition in the future.

nucypher/blockchain/eth/clients.py
KPrasch: :heavy_check_mark:

.circleci/config.yml
```diff
@@ -244,38 +234,39 @@ commands:
   pip_install:
     description: "Install NuCypher with Pip"
     steps:
-      - checkout
       - run:
           name: Install Python Dependencies with Pip
-          command: pip3 install --user .
+          command: pip3 install --user -e .
```
tuxxy: :+1:

.circleci/config.yml
```diff
@@ -31,56 +31,56 @@ workflows:
...
-            - pipenv_install_36
+            - pip_install_37
       - cli:
           context: "NuCypher Tests"
           filters:
             tags:
               only: /.*/
           requires:
-            - pipenv_install_36
+            - pip_install_37
```
tuxxy: :+1:

Issue #846, opened 2019-03-15
Gracefully handle BobInterface.retrieve decryption failure


KPrasch:
```
...
policy_encrypting_key = UmbralPublicKey.from_bytes(policy_encrypting_key)
alice_pubkey_sig = UmbralPublicKey.from_bytes(alice_verifying_key)
...
 message_kit = UmbralMessageKit.from_bytes(message_kit)  <<<< DECRYPTION FAILURE
```
Either `UnknownOpenSSLError` or `InvalidTag` may be raised on this line.

Perhaps it is better to provide a NuCypher exception here?

cygnusv:
Just a reminder I have an old stash with tests for this and a fix. It's caused by the difference between serializing `MessageKits` via `bytes()` vs `.to_bytes()`.

cygnusv:
This may be a duplicate of #776

PR #1297, opened 2019-09-04
[WIP] EthBerlin reconciliation

jMyles:


cygnusv:
hey @jMyles 
I'm closing #1309, which addressed the fix to #846, since it seems you already do this here. Can you add some unit tests for MessageKits? There's a small unit test in #1309 in case you want to use that.

jMyles:
No longer relevant.

Issue #1144, opened 2019-07-25
Windows compatibility 


KPrasch:

KPrasch:
Closing for non-specificity: re-open Issues pertaining to windows compatibility on a per-domain basis ie. windows development, enrico-on-windows, etc.

PR #1304, opened 2019-09-07
Increasing windows compatibility in tests

szotov:
Refs #1144


szotov:
In tests I see errors only with py-geth, because it uses unix sockets, I don't think there is real reason to solve it

michwill:
I guess, we just don't need to run geth tests in Windows

PR #1309, opened 2019-09-09
[ON-HOLD] Give UmbralMessageKit some love. Fixes #846

### LINKED ISSUE #846: Gracefully handle BobInterface.retrieve decryption failure

cygnusv:
- Fixes serialization issues with `UmbralMessageKit` (Fixes #846)
- Base64 support for UmbralMessageKit
- Basic unit tests

### DIFF
--- a/nucypher/crypto/kits.py
+++ b/nucypher/crypto/kits.py
@@ -15,76 +15,90 @@ You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
-
 from bytestring_splitter import BytestringSplitter
-from constant_sorrow import constants
+from constant_sorrow.constants import NOT_SIGNED
 
-from nucypher.crypto.splitters import key_splitter, capsule_splitter
-
-
-class CryptoKit:
-    splitter = None
-
-    @classmethod
-    def split_bytes(cls, some_bytes):
-        if not cls.splitter:
-            raise TypeError("This kit doesn't have a splitter defined.")
-
-        return cls.splitter(some_bytes,
-                            return_remainder=cls.return_remainder_when_splitting)
-
-    @classmethod
-    def from_bytes(cls, some_bytes):
-        constituents = cls.split_bytes(some_bytes)
-        return cls(*constituents)
+from nucypher.crypto.splitters import key_splitter, capsule_splitter, signature_splitter
 
 
-class MessageKit(CryptoKit):
+KIT_NOT_SIGNED = b"\x00"
+KIT_SIGNED_BUT_NO_SIGNATURE = b"\x01"  # includes verifying key
+KIT_INCLUDES_SIGNATURE = b"\x02"  # includes verifying key
+
+
+class UmbralMessageKit:
+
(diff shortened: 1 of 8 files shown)


nucypher/crypto/kits.py
```diff
@@ -15,76 +15,112 @@
...
-    @classmethod
-    def split_bytes(cls, some_bytes):
-        if not cls.splitter:
-            raise TypeError("This kit doesn't have a splitter defined.")
 
-        return cls.splitter(some_bytes,
-                            return_remainder=cls.return_remainder_when_splitting)
+KIT_NOT_SIGNED = b"\x00"
+KIT_SIGNED_BUT_NO_SIGNATURE = b"\x01"  # includes verifying key
+KIT_INCLUDES_SIGNATURE = b"\x02"  # includes verifying key
```
derekpierre: did you mean `# includes signature`

nucypher/crypto/kits.py
```diff
@@ -15,76 +15,112 @@
...
+        some_bytes = base64.b64decode(some_b64_str)
+        message_kit = cls.from_bytes(some_bytes)
+        return message_kit
 
-    return_remainder_when_splitting = True
-    splitter = capsule_splitter + key_splitter
+    def __eq__(self, other):
+        eq_conditions = (self.ciphertext == other.ciphertext,
+                         self.capsule == other.capsule,
+                         self.sender_verifying_key == other.sender_verifying_key)
```
derekpierre: Do we need to match on `signature` as well or is it purposely ignored?

nucypher/crypto/kits.py
cygnusv: Yeah, right.

nucypher/crypto/kits.py
cygnusv: It was on purpose, and loosely related to this interchange between @jMyles and me:

> dnunez: something that I'd like your opinion here is the relation between the MessageKit and it's signature. It's very weird since (1) the signature is included in the constructor, but (2) it's not serialized together with the MessageKit (nor deserialized obviously), and (3) it's passed separately in the output of encrypt_and_sign() (and therefore in the output of Enrico.encrypt())
> jmyles: I agree that this area needs some love (in fact see my PR, which makes potentially useful changes - I'm still hoping to get that merged along with the demo material).  I don't per se agree that including the signature in the (or at least in some) constructor but then not serializing it as a separate component is strange.

cygnusv:
I'm putting this PR on hold for the moment. It seems @jMyles already worked on this following a different direction (see #1297), and as long as it fixes #846 and provides unit tests, I don't see a reason to merge this instead of @jMyles' approach.

cygnusv:
I'm closing this for the moment, as it seems #1297 already handles it.

PR #1313, opened 2019-09-10
Moe self-signed certificate for HTTPS + WSS

KPrasch:
Based on #1311

### DIFF
--- a/nucypher/crypto/api.py
+++ b/nucypher/crypto/api.py
@@ -185,8 +185,11 @@ def generate_self_signed_certificate(host: str,
     now = datetime.datetime.utcnow()
     subject = issuer = x509.Name([
         x509.NameAttribute(NameOID.COMMON_NAME, host),
-        x509.NameAttribute(NameOID.PSEUDONYM, checksum_address)
     ])
+
+    if checksum_address:
+        x509.NameAttribute(NameOID.PSEUDONYM, checksum_address)
+
     cert = x509.CertificateBuilder().subject_name(subject)
     cert = cert.issuer_name(issuer)
     cert = cert.public_key(public_key)
--- a/nucypher/network/status_app/moe.py
+++ b/nucypher/network/status_app/moe.py
@@ -43,7 +43,7 @@ class MoeStatusApp(NetworkStatusPage):
         template_path = os.path.join(self.TEMPLATE_DIR, 'moe.html')
         with open(template_path, 'r') as file:
             moe_template = file.read()
-            self.dash_app.index_string = Template(moe_template).substitute(ws_port=ws_port)
+            self.dash_app.index_string = Template(moe_template).substitute(ws_port=ws_port, ws_host='0.0.0.0')  # TODO
 
         self.dash_app.layout = html.Div([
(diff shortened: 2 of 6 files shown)


nucypher/network/status/moe.html
```diff
@@ -13,7 +13,7 @@
             window.onload = function () {
                 const ws_port = $ws_port;
                 const origin_hostname = window.location.hostname;
```
derekpierre: @KPrasch Do we need to template hostname if it can be passed as a parameter to Moe?

nucypher/network/status/moe.html
```diff
@@ -12,8 +12,8 @@
         <script>
             window.onload = function () {
                 const ws_port = $ws_port;
-                const origin_hostname = window.location.hostname;
-                const socket = new WebSocket("wss://"+ origin_hostname + ":" + ws_port);
+                const ws_hostname = '$ws_host';
```
derekpierre: Ah! here is the templating (separate commit?)! 😄

nucypher/crypto/api.py
```diff
@@ -185,8 +185,11 @@ def generate_self_signed_certificate(host: str,
     now = datetime.datetime.utcnow()
     subject = issuer = x509.Name([
         x509.NameAttribute(NameOID.COMMON_NAME, host),
-        x509.NameAttribute(NameOID.PSEUDONYM, checksum_address)
     ])
+
```
KPrasch: Note for review - please look here.

nucypher/network/status_app/moe.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, moe, ws_port: int, *args, **kwargs):
         template_path = os.path.join(self.TEMPLATE_DIR, 'moe.html')
         with open(template_path, 'r') as file:
             moe_template = file.read()
-            self.dash_app.index_string = Template(moe_template).substitute(ws_port=ws_port)
+            self.dash_app.index_string = Template(moe_template).substitute(ws_port=ws_port, ws_host='0.0.0.0')  # TODO
```
derekpierre: Would we want to pass in `ws_host` to `__init__(...)` and then use its value here instead of just using the default, which can be overridden in the cli?

derekpierre: or maybe `moe.host` can just be used?

KPrasch: :heavy_check_mark:

nucypher/network/status_app/moe.py
```diff
@@ -43,7 +42,8 @@ def __init__(self, moe, ws_port: int, *args, **kwargs):
         template_path = os.path.join(self.TEMPLATE_DIR, 'moe.html')
         with open(template_path, 'r') as file:
             moe_template = file.read()
-            self.dash_app.index_string = Template(moe_template).substitute(ws_port=ws_port, ws_host='0.0.0.0')  # TODO
+            self.dash_app.index_string = Template(moe_template).substitute(ws_port=moe.websocket_port,
+                                                                           ws_host=self.moe.host)
```
derekpierre: 👍

PR #1318, opened 2019-09-12
Contracts tests time optimization

szotov:
* Reset state for chain each module
* Stub for gas estimation while `transact`

### DIFF
--- a/tests/fixtures.py
+++ b/tests/fixtures.py
@@ -74,7 +74,6 @@ from nucypher.utilities.sandbox.constants import (
     MOCK_REGISTRY_FILEPATH,
     TEST_GAS_LIMIT,
     INSECURE_DEPLOYMENT_SECRET_HASH,
-    NUMBER_OF_ETH_TEST_ACCOUNTS,
 )
 from nucypher.utilities.sandbox.middleware import MockRestMiddleware
 from nucypher.utilities.sandbox.policy import generate_random_label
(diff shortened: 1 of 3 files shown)


tests/fixtures.py
```diff
@@ -72,6 +74,7 @@
     MOCK_REGISTRY_FILEPATH,
     TEST_GAS_LIMIT,
     INSECURE_DEPLOYMENT_SECRET_HASH,
+    NUMBER_OF_ETH_TEST_ACCOUNTS,
```
mswilkison: I think you can remove this since you reverted the line where it's used in 800876

szotov: ✔️

tests/fixtures.py
```diff
@@ -366,6 +368,14 @@ def _make_testerchain():
     eth._utils.headers.GAS_LIMIT_MINIMUM = TEST_GAS_LIMIT
     eth._utils.headers.GENESIS_GAS_LIMIT = TEST_GAS_LIMIT
     eth.vm.forks.frontier.headers.GENESIS_GAS_LIMIT = TEST_GAS_LIMIT
+
+    # Monkey patch to prevent gas estimates
+    def _get_buffered_gas_estimate(web3, transaction, gas_buffer=100000):
+        return TEST_GAS_LIMIT
+
+    import web3
+    web3.eth.get_buffered_gas_estimate = _get_buffered_gas_estimate
```
KPrasch: How does this differ from setting a Web3 gas strategy and why not use that API?

szotov: It's related to gas estimations itself, gas price is independent from that

PR #1321, opened 2019-09-12
Token burning

vzotova:
* Burn tokens in `StakingEscrow` (for future reward)
* Ability to burn remaining tokens in `WorkLock`
* Explicit initialization for `StakingEscrow` and `WorkLock`

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
@@ -18,7 +18,7 @@ contract WorkLock {
     event Bid(address indexed staker, uint256 depositedETH, uint256 claimedTokens);
     event Claimed(address indexed staker, uint256 claimedTokens);
     event Refund(address indexed staker, uint256 refundETH, uint256 completedWork);
-    event Burn(address indexed sender, uint256 value);
+    event Burnt(address indexed sender, uint256 value);
 
     struct WorkInfo {
         uint256 depositedETH;
@@ -164,7 +164,8 @@ contract WorkLock {
         require(remainingTokens > 0, "There are no tokens that can be burned");
         token.approve(address(escrow), remainingTokens);
         escrow.burn(remainingTokens);
-        emit Burn(msg.sender, remainingTokens);
+        emit Burnt(msg.sender, remainingTokens);
+        remainingTokens = 0;
     }
 
 }
--- a/tests/blockchain/eth/contracts/base/test_issuer.py
+++ b/tests/blockchain/eth/contracts/base/test_issuer.py
@@ -74,14 +74,14 @@ def test_issuer(testerchain, token, deploy_contract):
     events = issuer.events.Initialized.createFilter(fromBlock='latest')
 
     # Give staker tokens for reward and initialize contract
-    tx = token.functions.transfer(issuer.address, economics.erc20_reward_supply).transact({'from': creator})
+    tx = token.functions.approve(issuer.address, economics.erc20_reward_supply).transact({'from': creator})
     testerchain.wait_for_receipt(tx)
 
     # Only owner can initialize
     with pytest.raises((TransactionFailed, ValueError)):
-        tx = issuer.functions.initialize().transact({'from': ursula})
+        tx = issuer.functions.initialize(0).transact({'from': ursula})
         testerchain.wait_for_receipt(tx)
-    tx = issuer.functions.initialize().transact({'from': creator})
+    tx = issuer.functions.initialize(economics.erc20_reward_supply).transact({'from': creator})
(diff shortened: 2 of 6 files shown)


tests/blockchain/eth/contracts/main/staking_escrow/test_staking_escrow_additional.py
```diff
@@ -452,7 +453,7 @@ def test_worker(testerchain, token, escrow_contract, deploy_contract):
     worker_log = escrow.events.WorkerSet.createFilter(fromBlock='latest')
 
     # Initialize escrow contract
-    tx = escrow.functions.initialize().transact({'from': creator})
+    tx = escrow.functions.initialize(0).transact({'from': creator})
```
KPrasch: `0`?

szotov: Just for tests where mining calculation is not necessary

mswilkison:
The most likely case is that the WorkLock will be uncapped. Since there will be no predetermined exchange rate and the WorkLock allocation will be spread pro rata across all participants, we won't actually use the Burn function.

This could change, so perhaps it makes sense to leave the Burn function in.

michwill:
@mswilkison Well, even in the case of uncapped - we do need burning.
Imagine that someone got # tokens less than minimum needed for staking, and they choose to get their ETH back rather than tokens. What to do with those tokens if not burn?

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -160,6 +163,16 @@ contract Issuer is Upgradeable {
         currentSupply2 = currentSupply2.sub(_amount);
     }
 
+    /**
+    * @notice Burn sender's tokens. Amount of tokens will be returned for future minting
+    * @param _value Amount to burn
+    **/
+    function burn(uint256 _value) public {
```
arjunhassard:
```suggestion
    function burn(uint256 _value) public {
    require(currentMintingPeriod <= 100, "Sorry, you waited too long – get some more Nu and start staking!");
```
If the primary purpose of the burn function is to allow Nu holders who don't meet minimum stake size requirements to get their ETH back, then maybe one should only be able to call it up until a predefined moment in time – e.g. within the first 100 periods of the reward schedule

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -160,6 +163,16 @@ contract Issuer is Upgradeable {
...
+        token.safeTransferFrom(msg.sender, address(this), _value);
```
arjunhassard: Alternatively, or in addition, we could limit the use of this function to participants that end up with less than the min stake size

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -144,4 +156,16 @@ contract WorkLock {
...
 
+    /**
+    * @notice Burn unclaimed tokens
+    **/
+    function burnRemaining() public {
+        require(block.timestamp >= endBidDate, "Burning tokens allowed when bidding is over");
+        require(remainingTokens > 0, "There are no tokens that can be burned");
+        token.approve(address(escrow), remainingTokens);
+        escrow.burn(remainingTokens);
+        emit Burnt(msg.sender, remainingTokens);
```
arjunhassard: Is there any practical difference in outcome between actively burning your Nu and not claiming them?

szotov: For now if staker does not claim Nu (in worklock) then can do it latter

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
szotov: Initial idea was for using from worklock and who else wants, but of course we can restrict this method for other Nu holders

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -87,15 +90,15 @@ contract Issuer is Upgradeable {
     /**
     * @notice Initialize reserved tokens for reward
     **/
-    function initialize() public onlyOwner {
+    function initialize(uint256 _reservedReward) public onlyOwner {
         require(currentSupply1 == 0);
+        token.safeTransferFrom(msg.sender, address(this), _reservedReward);
```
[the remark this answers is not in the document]
szotov: More control - explicitly set amount of reward, I'm feeling much safer with setting this parameter in function (same for `WorkLock`)

PR #1322, opened 2019-09-12
[WIP] [Epic] Susskartoffel -- Revocations and policy info serializations

tuxxy:
WIP

### DIFF
--- a/nucypher/policy/collections.py
+++ b/nucypher/policy/collections.py
@@ -14,8 +14,10 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
-import binascii
 
+
+import binascii
+import json
 import maya
 import msgpack
 from bytestring_splitter import BytestringSplitter, VariableLengthBytestring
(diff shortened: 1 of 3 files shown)


nucypher/policy/collections.py
```diff
@@ -209,6 +211,67 @@ def __len__(self):
...
+    @classmethod
+    def from_json(cls, data: str):
+        """
+        Deserializes the PolicyCredential from JSON.
+        """
+        cred_json = json.loads(data)
+
+        alice_verifying_key = UmbralPublicKey.from_bytes(
+                                    cred_json['alice_verifying_key'],
+                                    decoder=bytes().fromhex)
```
KPrasch: Extra set of parenthesis after `bytes` here - passing the `fromhex` method bound to an empty bytestring.

nucypher/policy/policies.py
```diff
@@ -277,6 +277,22 @@ def publish(self, network_middleware: RestMiddleware) -> dict:
         """
         return self.publish_treasure_map(network_middleware=network_middleware)
 
+    def credential(self, with_treasure_map=True):
```
KPrasch: I think `make_credential` or similar is more descriptive here, since we are not *using* the credential here.

michwill:
I think, rather than returning a credential _with_ every grant, we need to make the output of `grant` being a credential (there is not much difference: just slightly more info in the credential)

nucypher/policy/policies.py
michwill: I like more concise option. Or what, in your opinion, does the `credential()` method name say intuitively?

michwill:
Hodl on, I think repearing the same info in credential _and_ grant output is ugly. Refactoring a bit

nucypher/policy/collections.py
tuxxy: Not an extra set, this is intentional. :)

nucypher/policy/policies.py
tuxxy: Credential is already a verb. When you _credential_ someone, you are giving them credential.

tuxxy:
@michwill we don't return a credential with every grant. Nothing has changed on that, only in the character control is that what happens because we can't return a `Policy` object.

In the Python API, we still only return a `policy`, which you can call `policy.credential()` to have the credential.

nucypher/policy/collections.py
KPrasch: @tuxxy - Why do this / What effect does it have? Why not use the class method directly `bytes.from_hex`?

nucypher/policy/policies.py
KPrasch: Using the verb form of `credential` sounds to my ear like you are *presenting it to gain access*, not merely possessing it.  `Policy` can be used to produce a credential, but It sounds  to me like `Alice` / `Bob` do the actual credentialing.

michwill:
Well, ok, then I'd be ok with this one as a temporary measure. If I touch this, I'm afraid I'll be rebuilding a good part of the codebase, heh.

nucypher/policy/collections.py
michwill: Yeah, I think `bytes.fromhex` works just as well

KPrasch:
Rebased over `64465d6f7a8536f6d7b52e9d8daa9482c311925b`

KPrasch:
Moved to #1367

PR #1323, opened 2019-09-12
Refactor Alice's revoke method to accept a `TreasureMap`

tuxxy:
### What this does:

1. Refactors `Alice.revoke` to take a `TreasureMap` affiliated with an active policy instead of a `Policy` directly.
2. The `RevocationKit` now also takes a ` TreasureMap` instead of a `Policy`.
3. Removes `add_active_policy` and `active_policies`, as well as the `revocation_kits` dict.
4. Adds two methods `__serialize` and `__deserialize` to `TreasureMap`.
    - This is not super fine work, but it's a hacky way to get going in the direction that we eventually need to be in.
5. The `grant` character control method additionally returns a `PolicyCredential` for Alice (or Bob) to use as needed.

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -139,19 +139,6 @@ class Alice(Character, BlockchainPolicyAuthor):
         self.log = Logger(self.__class__.__name__)
         self.log.info(self.banner)
 
-        self.active_policies = dict()
-        self.revocation_kits = dict()
-
-    def add_active_policy(self, active_policy):
-        """
-        Adds a Policy object that is active on the NuCypher network to Alice's
-        `active_policies` dictionary by the policy ID.
-        The policy ID is a Keccak hash of the policy label and Bob's stamp bytes
-        """
-        if active_policy.id in self.active_policies:
-            raise KeyError("Policy already exists in active_policies.")
-        self.active_policies[active_policy.id] = active_policy
-
     def generate_kfrags(self,
                         bob: 'Bob',
                         label: bytes,
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -59,7 +59,7 @@ from nucypher.characters.control.controllers import (
 from nucypher.config.storages import NodeStorage, ForgetfulNodeStorage
 from nucypher.crypto.api import keccak_digest, encrypt_and_sign
 from nucypher.crypto.constants import PUBLIC_KEY_LENGTH, PUBLIC_ADDRESS_LENGTH
-from nucypher.crypto.kits import UmbralMessageKit
+from nucypher.crypto.kits import UmbralMessageKit, RevocationKit
 from nucypher.crypto.powers import SigningPower, DecryptingPower, DelegatingPower, TransactingPower, PowerUpError
 from nucypher.crypto.signing import InvalidSignature
 from nucypher.keystore.keypairs import HostingKeypair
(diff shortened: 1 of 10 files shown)


nucypher/characters/lawful.py
```diff
@@ -292,28 +279,30 @@ def get_policy_encrypting_key_from_label(self, label: bytes) -> UmbralPublicKey:
         policy_pubkey = alice_delegating_power.get_pubkey_from_label(label)
         return policy_pubkey
 
-    def revoke(self, policy) -> Dict:
+    def revoke(self, treasure_map) -> Dict:
```
michwill: So, do we want want to revoke by treasure_map or by revocation_kit?

michwill: Ah, I see, it's probably better to keep only the minimum needed here indeed

nucypher/policy/collections.py
```diff
@@ -163,6 +167,29 @@ def public_id(self):
         _id = keccak_digest(bytes(self._verifying_key) + bytes(self._hrac)).hex()
         return _id
 
+    def __serialize(self):
+        """
+        WAIT! This is probably not what you want. In 9/10 times, you'll
```
KPrasch: It's not clear what's happening here to me - What are the differences between this and the other serilizaers?

michwill: The "normal" serializer also encrypts for Bob (yeah, very convoluted logic). This one serializes a plaintext TreasureMap

michwill:
> Help me understand at a high-level - What is the motivation for making this change?

Basically, Alice may not have enough information at the time she revokes to construct a whole policy object. But you only need the treasure map to actually revoke.

I do agree though that it's logical to revoke by policy. Maybe we can have a separate method?

nucypher/policy/collections.py
tuxxy: Yes, essentially, when you call `bytes(...)` on a `TreasureMap`, it assumes you want it encrypted for Bob. Attempting to change this causes massive ripple effects.

tuxxy:
@KPrasch From an API design standpoint, it may be more clear _by name only_.

Once you start digging into the internals of `Policy`, it's really obvious that using it, serializing it, and passing it around places is exceptionally burdensome. It's becoming far more apparent that `Policy` needs some refactoring to get it in a place where it may be useful for anything like this. As of right now, it feels extremely inappropriate to use a `Policy` at an API level, unfortunately.

When we start reducing what we _need_ from a `Policy` to use in the API, we essentially construct a `PolicyCredential`. Though, this may be evolving into as @jMyles a parent class.

michwill:
Hmm, would it actually revoke on chain?

nucypher/policy/collections.py
```diff
@@ -163,6 +167,29 @@ def public_id(self):
...
+        probably want to use `bytes(my_treasure_map)`. So... do that.
+        """
+        return self._public_signature + self._hrac + self._m.to_bytes(1, 'big') + self.nodes_as_bytes()
+
+    @classmethod
+    def __deserialize(cls, bytes_representation):
+        """
+        WAIT! This is probably not what you want. In 9/10 times, you'll probably
```
KPrasch: Hmm - can you be more specific bout the condition in which you *do* want to use this method in the doctring?

tests/characters/control/test_web_control.py
```diff
@@ -99,12 +99,15 @@ def test_alice_character_control_revoke(alice_web_controller_test_client, federa
     response = alice_web_controller_test_client.put('/grant', data=json.dumps(grant_request_data))
     assert response.status_code == 200
 
+    policy_credential = PolicyCredential.from_json(
+                                response.json['result']['policy_credential'])
```
KPrasch: This style of formatting (odd newline placement, IMO) is throughout your PR - are running black?

michwill: That's the way to please PEP8 gods with the 80-character rule!

KPrasch: in this case (and in many) you can get a shorter, more readable block by adding a variable:

```
    credential_json = response.json['result']['policy_credential']
    policy_credential = PolicyCredential.from_json()
```

KPrasch:
Included in #1367

tuxxy:
Closing in favor of #1367

PR #1327, opened 2019-09-13
[EPIC] Biznaga: Integration of testnet with staking contracts

### LINKED ISSUE #856: Pre-allocation CLI Entry Points (UserEscrow)

### LINKED ISSUE #1084: Update/Rethink @validate_checksum_address decorator
With the new changes in #1058 sometimes the checksum addresses used in functions' arguments are called "staker_address" or "worker_address". The decorator currently ignores these addresses as it only looks for "checksum_address".

### LINKED ISSUE #1110: Is UserEscrowProxy a good name?
At first sight, the name `UserEscrowProxy` seems to hint that this contract act as a proxy _to_ `UserEscrow` (i.e. it intermediates between a client and `UserEscrow`). However, it acts as a proxy _for_ `UserEscrow` _to_ several staker-related functionalities of various contracts, namely, `NuCypherToken`, `StakingEscrow` and `PolicyManager`, so its name can be somewhat misleading.

What is more, `UserEscrow` is, hopefully, just the first of many types of contracts acting as stakers in the NuCypher network (see the term `PuppetStaker` in #1057). So, from that viewpoint, the current name `UserEscrowProxy` is clearly unambitious. 

In light of both reasons, what's a good name for `UserEscrowProxy`? I was thinking something like `StakingProxy`, but perhaps there's something better out there.

Bonus question: What's a good name for `UserEscrowLibraryLinker`? As a reminder, this contract simply acts as a pointer to the incumbent `UserEscrowProxy`. My suggestion would be something like `StakingProxyPointer` or `StakingProxyRegistry`.

### LINKED ISSUE #1115: Check that staker has minimum stake in node validation, instead of >0

### LINKED ISSUE #1128: Reliable and thin device auto-detection for Transacting Power and CLI.

### LINKED ISSUE #1351: Redesign allocation registry
Current `AllocationRegistry` is the output of `nucypher-deploy allocations` and is a JSON file containing a dictionary of beneficiary addresses and "contract data", which is just a tuple of a contract address (i.e. the `UserEscrow` address for that beneficary) and the corresponding contract ABI.

Some inefficiencies of such design for the allocation registry:
* The contract ABI is the same for all `UserEscrow` contracts, so the registry is very redundant.
* It presumes that the same file that is the output of `nucypher-deploy allocations` will be used by allocation users when `nucypher stake`, which is not very convenient.
* The file contains the beneficiary address and `UserEscrow` address of all allocation users, which is only marginally useful for the deployer, but not for allocation users.

The work in #1227 that introduced `UserEscrow` staking with the CLI continues to use said design for `AllocationRegistry` as a compromise to not change too much in a single PR, but this must change in a subsequent PR, part of the `biznaga` EPIC. 

**Proposal**
1. Instead/In addition of current `allocation_registry.json` file, the `nucypher-deploy` command will output:
   - A file with just the ABI of the `UserEscrow` contract, which recall it's the same for all allocation users. Let's call this file `allocation_contract_ABI.json`.
   - A small file for each allocation user containing the beneficiary address and the corresponding `UserEscrow` contract address. Let's call this file `allocation_user_0xfoobar.json`, where `0xfoobar` is the beneficiary.

1. `allocation-contract-ABI.json` is uploaded to the "master registry repo" (currently in GH), together with the contract registry (`contract_registry.json`).
1. `allocation_user_0xfoobar.json` is sent to the allocation user `0xfoobar`.
1. When allocation users stake using `nucypher stake`, they'll have to provide either:
    - Both the `--beneficiary-address 0xfoobar` and `--contract-address 0xabc123`
    - A path to an `allocation_user_0xfoobar.json` file, which recall contains just a tuple with the beneficiary and `UserEscrow` contract addresses. 

Any comment or alternative proposal is welcome!

### LINKED ISSUE #1369: Staked tokens should be only the locked tokens, not locked + rewards

* In `Staker.current_stake`:
```python
@property
        """
        """
        return nu_stake
```

### LINKED ISSUE #1385: Potential incorrect initialization of TransactingPowers in StakeHolder

```ValueError: {'code': -32000, 'message': 'authentication needed: password or unlock'}```

### LINKED ISSUE #1418: Past stakes shouldn't be printed as "Active" with `nucypher stake list`
E.g.:
```

| 0 | 0x240e | 0x5d26 | 0 | 15000 NU | 31 periods . | Sep 03 11:04:10 CEST - Oct 04 11:04:10 CEST 
| 1 | 0x240e | 0x5d26 | 1 | 15000 NU | 360 periods  | Oct 17 11:04:10 CEST - Oct 11 11:04:10 CEST  
```

In this example, stake 0 already ended as for october 16th, but it's still printed as active.

### PR DESCRIPTION
**Pull request by cygnusv**
image

**Goals:**
- Integration of `UserEscrow` staking with the CLI
- Renaming of `UserEscrow`, `UserEscrowProxy` and `UserEscrowLibraryLinker` (see https://github.com/nucypher/nucypher/issues/1110)
- Improvement of UX of allocation process

**Ramp:**
- [x] #1319 
- [x] #1227 
- [x] #1421 
- [x] #1402 

**Fixes**:
- Fixes #856 
- Fixes #1084 
- Fixes #1385 
- Fixes #1128
- Closes #1115
- Fixes #1369
- Closes #1351
- Fixes #1418
- Fixes #1110 

Chaperone: @cygnusv

### CODE REVIEW

File: nucypher/blockchain/eth/actors.py [hunk 1]
```diff
@@ -349,43 +349,106 @@ def deploy_beneficiary_contracts(self,
...
-                continue
+        if interactive:
+            click.confirm("Continue with the allocation process?", abort=True)
 
-            else:
-                allocation_txhashes[allocation['beneficiary_address']] = txhashes
+        total_to_allocate = NU.from_nunits(sum(allocation['amount'] for allocation in allocations))
+        balance = ContractAgency.get_agent(NucypherTokenAgent, self.registry).get_balance(self.deployer_address)
+        if balance < total_to_allocate:
+            raise ValueError("Not enough tokens to allocate. We need at least {total_to_allocate}.")
```
mswilkison:
```suggestion
            raise ValueError(f"Not enough tokens to allocate. We need at least {total_to_allocate}.")
```

nucypher/cli/painting.py
```diff
@@ -418,10 +446,26 @@ def paint_staged_stake_division(emitter,
...
+                       unlock_period=new_end_period,
                        division_message=division_message)
 
 
+def paint_accounts(emitter, balances):
+    rows = list()
+    max_eth_len, max_nu_len = 0, 0
+    for address, balances in sorted(balances.items()):
+        eth = str(Web3.fromWei(balances['ETH'], 'ether')) + " ETH"
+        nu = str(NU.from_nunits(balances['NU']))
```
mswilkison: Do we want to append " NU" like we append " ETH" in the line above?

cygnusv: The `NU.__str__()` method already does that by default.

nucypher/blockchain/eth/registry.py
```diff
@@ -427,3 +449,38 @@ def read(self) -> list:
...
+        super().__init__(*args, **kwargs)
+
+        self.beneficiary_address = beneficiary_address
+        self.contract_address = contract_address
+
+        if not contract_abi:
+            # Download individual allocation template to extract contract_abi
+            individual_allocation_template = json.loads(self.fetch_latest_publication())
+            if len(individual_allocation_template) != 1:
+                raise self.IllegalRegistry("Individual allocation template must contain a single entry")
```
derekpierre: Minor nitpick: I would vote for `InvalidRegistry` instead of `IllegalRegistry`.

nucypher/cli/characters/stake.py
```diff
@@ -164,34 +172,42 @@ def accounts(click_config,
...
 @_worker_options
+@click.option('--force', help="Don't ask for confirmation", is_flag=True)
 @nucypher_click_config
 def set_worker(click_config,
 
                # Worker Options
-               poa, registry_filepath, config_file, provider_uri, staking_address, hw_wallet, worker_address):
+               poa, registry_filepath, config_file, provider_uri, staking_address, hw_wallet,
+               beneficiary_address, allocation_filepath,
+               worker_address, force):
```
derekpierre:
```suggestion
               worker_address, 
               
               # Other options
               force):
```

nucypher/cli/characters/stake.py
```diff
@@ -224,25 +242,33 @@ def set_worker(click_config,
...
 @_worker_options
+@click.option('--force', help="Don't ask for confirmation", is_flag=True)
 @nucypher_click_config
 def detach_worker(click_config,
 
                   # Worker Options
-                  poa, registry_filepath, config_file, provider_uri, staking_address, hw_wallet, worker_address):
+                  poa, registry_filepath, config_file, provider_uri, staking_address, hw_wallet,
+                  beneficiary_address, allocation_filepath,
+                  worker_address, force):
```
derekpierre:
```suggestion
                  worker_address, 
                  
                  # Other options
                  force):
```

nucypher/cli/painting.py
```diff
@@ -382,20 +393,37 @@ def prettify_stake(stake, index: int = None) -> str:
...
+        emitter.echo(header, bold=True)
+        emitter.echo(breaky, bold=True)
+
+        for stake in sorted([s for s in stakes if s not in active_stakes],
+                            key=lambda some_stake: some_stake.address_index_ordering_key):
+            row = prettify_stake(stake=stake, index=None)
+            emitter.echo(row, color='red')
+        emitter.echo('')  # newline
+    elif not active_stakes:
+        emitter.echo(f"There are no active stakes\n")
```
derekpierre: Just to clarify our logic here - we only log `There are no active stakes` if there are no active stakes **AND** `paint_inactive` is false. 

If there are no active stakes but `paint_inactive` is true, then we don't log?

cygnusv: L420 loops over non-active stakes. So if `paint_inactive is True` then it prints any existing non-active stake.

nucypher/blockchain/eth/registry.py
cygnusv: ✅

nucypher/cli/painting.py
derekpierre: Apologies if I'm missing something here.

Is it the case that the `stakes` parameter always has at least one stake in it? If `stakes` is empty, but `paint_inactive` is True, we will print the following:
```
========= Inactive Stakes ============
<---- headers----->
------|-------| ....
```
with no stakes listed, correct? So we show empty `Inactive Stakes` titled section?

cygnusv: Ah I see. Well, we have an open issue for this command option, we can rethink this to there.

cygnusv: Ah I see. Well, we have an open issue for this command option (https://github.com/nucypher/nucypher/issues/1436), we can rethink this to there.

derekpierre: 👍

nucypher/network/nodes.py
```diff
@@ -1012,9 +1013,19 @@ def _staker_is_really_staking(self, registry: BaseContractRegistry) -> bool:
         As a follow-up, this checks that the staker is, indeed, staking.
         """
         # Lazy agent get or create
-        staking_agent = ContractAgency.get_agent(StakingEscrowAgent, registry=registry)
-        locked_tokens = staking_agent.get_locked_tokens(staker_address=self.checksum_address)
-        return locked_tokens > 0  # TODO: Consider min stake size #1115
+        staking_agent = ContractAgency.get_agent(StakingEscrowAgent, registry=registry)  # type: StakingEscrowAgent
+
+        try:
+            economics = TokenEconomicsFactory.get_economics(registry=registry)
```
tuxxy: :+1:

PR #1328, opened 2019-09-13
Relock Dependencies

KPrasch:
Currently prevents new CI workflows from running.

See <https://circleci.com/gh/nucypher/nucypher/46185>

### DIFF
--- a/scripts/installation/relock_dependencies.sh
+++ b/scripts/installation/relock_dependencies.sh
@@ -1,0 +1,48 @@
+#!/usr/bin/env bash
+
+set -e
+
+# can change output file names with rebuild_pipenv.sh <prefix>
+PREFIX=${1:-requirements}
+
+read -p "Update your pip and setuptools? (type y or Y) " -n 1 -r
+echo
+if [[ $REPLY =~ ^[Yy]$ ]]
+then
+    pip install --upgrade pip
+    pip install --upgrade setuptools
+fi
+
+echo "Removing existing lock files..."
+pipenv --rm
+rm Pipfile.lock
+rm -f $PREFIX.txt
+rm -f dev-$PREFIX.txt
+
+read -p "Remove system pip and pipenv cache? (type y or Y) " -n 1 -r
+echo
+if [[ $REPLY =~ ^[Yy]$ ]]
+then
+    echo "Removing pip and pipenv system cache..."
+    rm -r ~/.cache/pip ~/.cache/pipenv
+fi
+
+echo "Rebuilding dependency locks..."
+
+echo "Bulding dev-$PREFIX.txt..."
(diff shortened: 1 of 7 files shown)


scripts/installation/relock_dependencies.sh
```diff
@@ -0,0 +1,48 @@
...
+
+echo "Installing nucypher into new virtual enviorment..."
+pipenv run pip install -e . -r dev-$PREFIX.txt
+
+echo "Checking nucypher CLI entry point..."
+pipenv run nucypher --version
+
+echo "Committing Result..."
+git add Pipfile Pipfile.lock $PREFIX.txt dev-$PREFIX.txt
+git commit -m "Relock dependencies"
```
derekpierre: I guess this is mainly for us to run. Although - how does this work when we want to add a dependency as part of a broader change - the "Relock dependencies" commit would be independently committed. Perhaps we could add a 'commit' option, whose default is to not commit?

KPrasch: Hmm - how about we always try to commit `Pipfile` as well?

derekpierre: If I understand correctly, running the script performs a git commit at the end, and with a comment just about relocking the dependency. If I run this before committing my code, then my code changes also get committed as part of the "relock dependencies" commit too, no?

vepkenez: That kind of thing gets annoying.

KPrasch: Well, it only runs `git add` against the 4 dependency filepaths, so for you to accidentally include other changes, you'd have to stage them first.

vepkenez: The current flow of validating that the requirements match should work fine.  And if your PR fails the reqs validation on circle, then you'd need to relock and push a new commit which is normal and fine.

KPrasch: I kept the original script in-tack, for circleCI to use, and relocated it to `scripts/circle`

vepkenez: yeah.  It all looks good to me.

derekpierre: I still feel funny about it, but I guess we can see how it plays out.

KPrasch: I am drawing a bit of inspiration from `bumpversion` - which performs a commit by default - but with a check for staged changes.   I think- it's a good follow up change to include one as well as support cache filepaths on other operating systems.  For now I've been holding on to that script for a while and need to get it off my local tree and am glad to see it merged. :-)

KPrasch: I think you are highlighting my critique via this PR - I dont see any reason to ever push a commit that includes both logic and dependency updates in our architecture.

PR #1331, opened 2019-09-14
CLI lifecycle test didn't work. When it does, tests didn't pass

michwill:


michwill:
The error now is:
```
E           eth_abi.exceptions.InsufficientDataBytes: Tried to read 32 bytes.  Only got 0 bytes
```
So, probably need some help.

tests/cli/test_cli_lifecycle.py
```diff
@@ -373,4 +375,4 @@ def _bob_retrieves(_grant_result):
     d.addCallback(_grant)                       # scene 7
     d.addCallback(_bob_retrieves)               # scene 8
 
-    yield d
+    return d
```
cygnusv: Oh, man, that's a good catch!
How do we know this same problem is not happening somewhere else?
Also, what's the explanation?

michwill: If you look at the function which calls this, it also yields. Basically it turns out that if it's a `yield` here, we'll return a future somewhere else instead of waiting for its completion, so the code inside doesn't get executed.

I think, places with `@pt.inlineCallbacks` should be checked extra carefully

PR #1344, opened 2019-09-17
Bare Deployment and Retargeting Workflow

KPrasch:
##### Motivation

Deployment tooling and UX response to state verification failures experienced during `StakingEscrow` upgrade.  Provides a workflow to help mitigate a *partially* completed contract upgrade.

##### Points of Interest

- Engagement of `_deploy_essential` without deploying a proxy any additional transactions via CLI (`--bare`).
- Proxy contract upgrade / retargeting without contract re-deployment (nucypher upgrade --retarget ...).
- Adds deployer methods for better direct-access to contracts both targeted and, untargeted by proxies.
- Allows for an alternate registry output file path to be used via depolyment CLI (`--registry-outfile`).
- Clean-up registry mocking and fixture usage.  (always use `MOCK_REGISTRY_FILEPATH`)

Based on #1345  and #1350 

##### Re-Attempt at Upgrade/Re-Target

without gas limit -> <https://gist.github.com/KPrasch/58b26d20dc2fef7505e4634e5680f49e>
with gas limit -> <https://gist.github.com/KPrasch/5977b5ce353c4ea832219d1a76f2faab>

##### Successful Upgrade

```
(nucypher) kieran@pegasus:~/Git/nucypher$ nucypher-deploy upgrade --contract-name StakingEscrow --provider ~/.ethereum/goerli/geth.ipc --poa --hw-wallet 
WARNING: --etherscan is disabled. If you want to see deployed contracts and TXs in your browser, activate --etherscan.
Configured to registry filepath /home/kieran/.local/share/nucypher/contract_registry.json
0 | 0xb3aD00CB3fEcbfe729cb496C70aC420FDc229AAC 
1 | 0xcC0678E51a8237b762c09d6548d2d07285609e98 
Select deployer account [0]: 0
Selected 0xb3aD00CB3fEcbfe729cb496C70aC420FDc229AAC - Continue? [y/N]: y


Deployer ETH balance: 4.266476613
Enter existing contract upgrade secret: 
Enter new contract upgrade secret: 
Repeat for confirmation: 
Confirm deploy new version of StakingEscrow and retarget proxy? [y/N]: y
Successfully deployed and upgraded StakingEscrow
OK | 0xcb45d757ada0bcb6680dd612e32801abbe6e229a50f813cce95871b8ba14eabe (6198413 gas)
Block #1331807 | 0x20e4b2a612adb87a04a58862c0ed65c331e1ae6dc67633730577a2cb97ba7886
[...]

PR #1339, opened 2019-09-17
Restaking Control

KPrasch:
Co-Author: @cygnusv 

Fixes #1269, Includes extracted changes from #1227 

##### New Staker CLI Commands

`nucypher stake restake --enable`
`nucypher stake restake --disable`
`nucypher stake restake --lock-until <PERIOD>`

##### Example 

```
$ nucypher stake restake --enable
 ____    __            __                      
/\  _`\ /\ \__        /\ \                     
\ \,\L\_\ \ ,_\    __ \ \ \/'\      __   _ __  
 \/_\__ \\ \ \/  /'__`\\ \ , <    /'__`\/\`'__\
   /\ \L\ \ \ \_/\ \L\.\\ \ \\`\ /\  __/\ \ \/ 
   \ `\____\ \__\ \__/.\_\ \_\ \_\ \____\\ \_\ 
    \/_____/\/__/\/__/\/_/\/_/\/_/\/____/ \/_/ 
   
The Holder of Stakes.                                      

======================================= Active Stakes =========================================

| ~ | Staker | Worker | # | Value    | Duration     | Enactment          
|   | ------ | ------ | - | -------- | ------------ | ----------------------------------------- 
| 0 | 0xEc39 | 0x985a | 0 | 45000 NU | 103 periods  | Aug 27 23:47:06 PDT - Dec 08 22:47:06 PST 

Select Stake: 0
Enter password to unlock account 0xEc391142E1007787101F328837C3ffa9eef767A2: 
Confirm enable restaking for staker 0xEc391142E1007787101F328837C3ffa9eef767A2? [y/N]: y
Successfully enabled restaking for 0xEc391142E1007787101F328837C3ffa9eef767A2
OK | 0x279c03b0bade60b0e499d85cce090d35bae2472aaa3f63250291f5b255086476 (31953 gas)
Block #1312292 | 0x7eab515f7a2e9ed98cbdbe6f2d4ff637e07274be1b9fd11bde7d0bbedc14a829
 See https://goerli.etherscan.io/tx/0x279c03b0bade60b0e499d85cce090d35bae2472aaa3f63250291f5b255086476

```

PR #1345, opened 2019-09-18
Fix StakingEscrow state for upgrading

szotov:
Deployed in testnet:
```
struct StakerInfo {
        uint256 value;
        uint16 confirmedPeriod1;
        uint16 confirmedPeriod2;
        bool reStake;
        uint16 lockReStakeUntilPeriod;
        address worker;
        // period when worker was set
        uint16 workerStartPeriod;
        // last confirmed active period
        uint16 lastActivePeriod;
        Downtime[] pastDowntime;
        SubStakeInfo[] subStakes;
}
```
and in master:
```
struct StakerInfo {
        uint256 value;
        uint16 confirmedPeriod1;
        uint16 confirmedPeriod2;
        bool reStake;
        uint16 lockReStakeUntilPeriod;
        address worker;
        // period when worker was set
        uint16 workerStartPeriod;
        // last confirmed active period
        uint16 lastActivePeriod;
        bool measureWork;
        uint256 completedWork;
        Downtime[] pastDowntime;
        SubStakeInfo[] subStakes;
}
```
and `require` with `delegateGet(_testTarget, "getPastDowntimeLength(address)", staker)` fails in `verifyState` because `pastDowntime` in new contract in a wrong slot
`measureWork` and `completedWork` must be after `subStakes`

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
@@ -90,10 +90,10 @@ contract StakingEscrow is Issuer {
         uint16 workerStartPeriod;
         // last confirmed active period
         uint16 lastActivePeriod;
-        bool measureWork;
-        uint256 completedWork;
         Downtime[] pastDowntime;
         SubStakeInfo[] subStakes;
+        bool measureWork;
+        uint256 completedWork;
     }
 
     /*
@@ -1216,6 +1216,11 @@ contract StakingEscrow is Issuer {
         bytes32 memoryAddress = delegateGetData(_target, "stakerInfo(address)", 1, _staker, 0);
         assembly {
             result := memoryAddress
+            // copy data to the right position after the array pointer place
+            // measureWork
+            mstore(add(memoryAddress, 0x140), mload(add(memoryAddress, 0x100)))
+            // completedWork
+            mstore(add(memoryAddress, 0x160), mload(add(memoryAddress, 0x120)))
         }
     }
 
--- a/tests/blockchain/eth/contracts/main/staking_escrow/test_staking_escrow_additional.py
+++ b/tests/blockchain/eth/contracts/main/staking_escrow/test_staking_escrow_additional.py
@@ -35,6 +35,7 @@ secret2 = (654321).to_bytes(32, byteorder='big')
 def test_upgrading(testerchain, token, token_economics, deploy_contract):
     creator = testerchain.client.accounts[0]
     staker = testerchain.client.accounts[1]
+    worker = testerchain.client.accounts[2]
 
     secret_hash = keccak(secret)
     secret2_hash = keccak(secret2)
@@ -86,6 +87,8 @@ def test_upgrading(testerchain, token, token_economics, deploy_contract):
     tx = contract.functions.setWorkLock(worklock.address).transact()
     testerchain.wait_for_receipt(tx)
 
+    tx = token.functions.transfer(contract.address, token_economics.erc20_reward_supply).transact({'from': creator})
+    testerchain.wait_for_receipt(tx)
     tx = contract.functions.initialize().transact({'from': creator})
     testerchain.wait_for_receipt(tx)
     tx = token.functions.transfer(staker, 1000).transact({'from': creator})
@@ -100,6 +103,13 @@ def test_upgrading(testerchain, token, token_economics, deploy_contract):
     tx = contract.functions.lockReStake(contract.functions.getCurrentPeriod().call() + 1).transact({'from': staker})
     testerchain.wait_for_receipt(tx)
     tx = worklock.functions.setWorkMeasurement(staker, True).transact()
+    testerchain.wait_for_receipt(tx)
+    tx = contract.functions.setWorker(worker).transact({'from': staker})
+    testerchain.wait_for_receipt(tx)
+    tx = contract.functions.confirmActivity().transact({'from': worker})
+    testerchain.wait_for_receipt(tx)
+    testerchain.time_travel(hours=2)
+    tx = contract.functions.confirmActivity().transact({'from': worker})
     testerchain.wait_for_receipt(tx)
 
     # Upgrade to the second version


KPrasch:
Failing CI/CLI test is resolved in #1339

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -1216,6 +1216,11 @@ contract StakingEscrow is Issuer {
         bytes32 memoryAddress = delegateGetData(_target, "stakerInfo(address)", 1, _staker, 0);
         assembly {
             result := memoryAddress
+            // copy data to the right position after the array pointer place
+            // measureWork
+            mstore(add(memoryAddress, 0x140), mload(add(memoryAddress, 0x100)))
```
michwill: Is this direct memory addressing the result of previous version not having those two fields? E.g. is it how we will handle similar upgrades in future?

szotov: It's independent from previous version, issue that structure has array pointers between fields. And function `stakerInfo` returns without array pointers but without gap. To correct translate from returned value to `StakerInfo` struct need to move tail of memory. In future need to add copying for each new slot of memory

KPrasch: Still not understanding fully - So this is specific to the upgrade issue with different struct layouts?  If we deploy on another network, with this as the initially deployed contract, then this code will have no effect (rather what effect will it have)?

cygnusv: I'm afraid I too don't understand. Why is `measureWork` (L1220) relevant here?

cygnusv: Oh, is it because `pastDowntime` and `subStakes` are arrays? and these arrays in the struct have a different memory structure in storage and when returned by `stakerInfo`?

szotov: Every layout (in every contract) such as 
```
uint256 value1;
uint256[] array1;
uint256 value2;
```
needs memory copying in this type of function.
Mostly we try to have this type of layout:
```
uint256 value1;
uint256 value2;
uint256[] array1;
```
because no need to copy memory
But backward compatibility requires add new field in the end of the structure.
So in new network/deployment this contract still requires  memory copying but we have opportunity to change layout before deploy if we want and remove unnesessary code

szotov: @cygnusv 
Storage layout (https://solidity.readthedocs.io/en/latest/miscellaneous.html?highlight=layout#layout-of-state-variables-in-storage)
`uint256 value;` -> 1st slot, `memoryPointer`
`uint16 confirmedPeriod1;`  and `uint16 confirmedPeriod2;` -> 2nd slot , `memoryPointer` + 0x20
`bool reStake;` and `uint16 lockReStakeUntilPeriod;` -> 3rd slot, `memoryPointer` + 0x40
`address worker;` -> 4th slot, `memoryPointer` + 0x60
`uint16 workerStartPeriod;` and `uint16 lastActivePeriod;` -> 5th slot, `memoryPointer` + 0x80
pointer to `Downtime[] pastDowntime;` in 6th slot, `memoryPointer` + 0x100
pointer to `SubStakeInfo[] subStakes;` in 7th slot, `memoryPointer` + 0x120
`bool measureWork;` in 8th slot, `memoryPointer` + 0x140
`uint256 completedWork;` in 9th slot, `memoryPointer` + 0x160

Can't say for sure why only two values in one slot for small types

`stakerInfo` returns without array pointers and `measureWork` `completedWork` are shifted to the left. To assign returned value to structure need to shift them to the right

cygnusv: OK, so this part of the code address the problem of memory copying because of current structure that contains arrays in the middle of the struct. Your point is that in the future we have the chance to deploy a Testnet V2 with a simpler memory structure. Is that correct?

szotov: Yes, it's correct. I prefer not to change this, but we can

michwill: Yeah, I'd certainly prefer to start mainnet without those unobvious pointers, which means that we need to have some testnet without those pointers.

But good to test those backward compatibility measures in practice!

KPrasch: > Yeah, I'd certainly prefer to start mainnet without those unobvious pointers

> But good to test those backward compatibility measures in practice!

I agree!

cygnusv: Does it make sense then, instead of doing this fix, have a hard fork with the correct structure of `stakerInfo`?

KPrasch: This means we'd also need to redeploy the token, start with 0 stakes, redistribute, etc.  considering our current standing - I think this fix is still the best course of action for *this* deployment set.

KPrasch: One of the best additions we can make it some tooling to help mitigate this class of problem in the future.  How can we prevent this in CI?  (Perhaps another discussion)

szotov: I'm for having this code maybe even in production, because it shows narrow places and we will meet with those type of upgrading on mainnet.
Other idea - do hardfork test (before mainnet), not just redeploy but also break contract and copy information from old to new one

cygnusv: Well, we don't need to be attached for this deployment set. If we deploy this new contract then at some point we'll still have to do hard fork if we want to deploy in mainnet with the correct structure from the start.

My point is that now is the moment to do unorthodox changes, not just before mainnet.

michwill: @cygnusv I totally agree with that. Nevertheless, it is very valuable to test, how we will do upgrades in similar cases.
I think, incentivized testnet is a good candidate for a "clean" deployment without this patch

michwill:
Since this fix is deployed anyway... Do we want to go and merge it?

cygnusv:
I think we will merge it together 1344. However, there's something I still don't have clear: we're going to merge this but we will deploy at some moment before mainnet (e g. Incentivized testnet) a new version with the correct memory layout from the start, and hence, reverting this fix. Is my understanding correct? In that case, we'd like to open an issue

michwill:
@cygnusv Yes, that's correct. We need an issue

cygnusv:
Created issue https://github.com/nucypher/nucypher/issues/1363, and assigned it a milestone for next testnet release.

tuxxy:
Did we ever figure out tooling as a check for this in the future?

KPrasch:
Included in #1344

PR #1364, opened 2019-09-27
Extract sample() from solidity to python

szotov:
Extracting `sample()` increases speed of sampling, decreases size of `StakingEscrow` and helps to easier maintain sampling in future


nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -327,22 +327,36 @@ contract StakingEscrow is Issuer {
...
+    * @return activeStakers Array of stakers and their locked tokens. Stakers addresses stored as uint256
     **/
-    function getAllLockedTokens(uint16 _periods)
-        external view returns (uint256 lockedTokens)
+    function getAllActiveStakers(uint16 _periods)
+        external view returns (uint256 allLockedTokens, uint256[2][] memory activeStakers)
     {
         require(_periods > 0);
+
+        uint256 stakersLength = stakers.length;
```
michwill: Oh interesting. In solidity, doing it through a local variable is faster?

szotov: It should slightly decrease size of contract if I remember correctly. Need to check it to be sure

nucypher/blockchain/eth/agents.py
```diff
@@ -456,16 +469,37 @@ def sample(self, quantity: int, duration: int, additional_ursulas: float = 1.7,
...
+            previous_point = 0
+            point_index = 0
+            sum_of_locked_tokens = 0
+            staker_index = 0
+            stakers_len = len(stakers)
+            while staker_index < stakers_len and point_index < n_select:
+                current_staker = stakers[staker_index][0]
+                staker_tokens = stakers[staker_index][1]
+                if staker_tokens == 0:
+                    break
```
KPrasch: why not throw an exception here?

szotov: Zero means that no more stakers in an array, but result still could be more than `quantity`

nucypher/blockchain/eth/agents.py
```diff
@@ -456,16 +469,37 @@ def sample(self, quantity: int, duration: int, additional_ursulas: float = 1.7,
...
+                next_sum_value = sum_of_locked_tokens + staker_tokens
+
+                point = points[point_index]
+                assert point >= previous_point
```
fjarri: Probably not necessary given that `points` are sorted.

nucypher/blockchain/eth/agents.py
```diff
@@ -456,16 +469,37 @@ def sample(self, quantity: int, duration: int, additional_ursulas: float = 1.7,
...
 
         for _ in range(attempts):
             points = sorted(system_random.randrange(n_tokens) for _ in range(n_select))
             self.log.debug(f"Sampling {n_select} stakers with random points: {points}")
 
-            addresses = set(self.contract.functions.sample(points, duration).call())
-            addresses.discard(str(BlockchainInterface.NULL_ADDRESS))
+            addresses = set()
+
+            previous_point = 0
```
fjarri: Just curious, how important the performance of this algorithm is, and whether the choice of linear search was intentional, or just the simplest one? As far as I understand it, the situation is as follows:

Let `M = stakers_len` and `N = n_select`. Then for different weighted selection algorithms we will have:
* linear search: `O(N log N) + O(M)`
* binary search: `O(M) + O(N log M)`
* hopscotch: `O(M log M) + O(N log M)` (but the second term has a much smaller constant factor than that in binary search, so it often turns out much faster)
In the application, `M` is probably large, and `N` is small. So if `getAllActiveStakers()` can be cached, the last two approaches will be faster. Although it is a smart contract call, so I'm guessing it can't?

michwill: Good points here!
I think, `getAllActiveStakers` can be cached for 1 day (for the current period) since this list doesn't change during this time. And it's a time-consuming call too! After it's done, optmizing weighted selection algorithm may help further

nucypher/blockchain/eth/agents.py
cygnusv: Fixed in a later commit.

nucypher/blockchain/eth/agents.py
cygnusv: While there's room for optimization here, I have some comments:
1) `getAllActiveStakers(_periods)` takes a `_periods` parameter for the duration of the policy, so that makes cache difficult and not that useful, specially if it only makes sense for that day.
2) Cached where? Caching in the smart contract costs gas, while current approach is free since it's a call.
3) Binary search makes a lot of sense, actually. Current approach was linear since it's a direct translation of the original sampling method in the smart contract, where sorting the stakers array is not convenient. But it does make sense to just traverse the array of random points, and for each point, perform binary search on the stakers array. However, we'd have to modify `getAllActiveStakers(_periods)` to return a sorted list of stakers.
4) Where can I find a description of hopscotch search? Is it related to [hopscotch hashing](https://en.wikipedia.org/wiki/Hopscotch_hashing)?

fjarri: 4. I read about it here: https://blog.bruce-hill.com/a-faster-weighted-random-choice

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
cygnusv: If I'm not wrong the explanation is that each access to `stakers.length` is actually an `SLOAD` instruction since `stakers` is and storage variable and its length is located as the first slot, and has a gas cost of 200. Access to a local variable is just a `POP`, with a cost of 2 gas.

nucypher/blockchain/eth/agents.py
michwill: 2. Yeah, indeed, caching doesn't work well here since the policy duration could be anything. Only if we have many policies of the same duration

jMyles:
OK, so at the outset, first thougths:  the rubicon that we're crossing here is that we won't be able to, in the future, economically force Alice to use a cohort of stake-weighted Ursulas.  

Are we cool with that?  I mean, I am.  But is this in keeping with our vision?  There's no good reason to believe, going forward, that higher-staking Ursulas will end up getting more work if Alice decides to use any kind of custom decision-making.

Of course, Alice was already capable of this, but if we are committing to doing this in python in perpetuity, she'll always be able to handpick.

cygnusv:
The previous algorithm didn't enforced that either, as it was based on Alice randomness, which obviously she can control. Unless we use some sourceb of on chain randomness, this in the smart contact

michwill:
Alice was handpicking Ursulas from the very beginning (and by design). Yes, a good part of sampling was in a "smart contract", but it was a read-only function, merely to read the blockchain.

arjunhassard:
@jMyles perhaps the best way to discourage users handpicking Ursulas by address is to eventually offer something more useful – a matching algorithm that incorporates user preferences such as proximity/latency, capacity,  % ignored requests (moving average). We already match based on policy duration

PR #1366, opened 2019-09-28
Allow optional arguments in controllers

michwill:
- [x] Optional arguments in RPC controllers
- [x] Pass value/rate/first_period_reward arguments optionally to make blockchain grant working over RPC (how did that work at all?)
- [x] Tests RPC with blockchain characters


tests/characters/control/blockchain/test_rpc_control_b.py
```diff
@@ -0,0 +1,107 @@
+import pytest
```
KPrasch: Are these new intentionally test modules (added the letter `b`)?

tests/fixtures.py
```diff
@@ -296,6 +298,26 @@ def reset(self):
     return _CapsuleSideChannel()
 
 
+@pytest.fixture(scope="module")
+def capsule_side_channel_blockchain(enacted_blockchain_policy):
```
KPrasch: Interesting - What Incompatibility did you encounter with the Federated Side Channel?

tests/characters/control/federated/test_rpc_control_f.py
```diff
@@ -8,21 +8,22 @@
...
 def test_alice_rpc_character_control_create_policy(alice_rpc_test_client, create_policy_control_request):
+    alice_rpc_test_client.__class__.MESSAGE_ID = 0
     method_name, params = create_policy_control_request
     request_data = {'method': method_name, 'params': params}
     rpc_response = alice_rpc_test_client.send(request=request_data)
     assert rpc_response.success is True
     assert rpc_response.id == 1
 
-    _input_fields, required_output_fileds = alice_specification.get_specifications(interface_name=method_name)
+    _input_fields, _optional, required_output_fileds = alice_specification.get_specifications(interface_name=method_name)
```
KPrasch: I wonder if this API wants to be access-by-key instead of unpacking now.

nucypher/characters/control/specifications.py
```diff
@@ -22,11 +22,13 @@ def get_specifications(cls, interface_name: str) -> tuple:
         if cls._specifications is NotImplemented:
             raise NotImplementedError("Missing specifications for character")
         try:
-            input_specification, output_specification = cls.specifications()[interface_name]
+            spec = cls.specifications()[interface_name]
         except KeyError:
             raise cls.SpecificationError(f"{cls.__class__.__name__} has no such control interface: '{interface_name}'")
 
-        return input_specification, output_specification
+        return spec.get('in', tuple()),\
```
KPrasch: This bit reads a little awkward on the return line, especially.

nucypher/characters/control/specifications.py
```diff
@@ -50,35 +53,37 @@ def __validate(specification: tuple, data: dict, error_class):
...
+        _, _, output_specification = self.get_specifications(interface_name=interface_name)
         return self.__validate(specification=output_specification, data=response, error_class=self.InvalidInputField)
 
 
 class AliceSpecification(CharacterSpecification):
 
-    __create_policy = (('bob_encrypting_key', 'bob_verifying_key', 'm', 'n', 'label', 'expiration'),  # In
-                       ('label', 'policy_encrypting_key'))                              # Out
+    __create_policy = {'in': ('bob_encrypting_key', 'bob_verifying_key', 'm', 'n', 'label', 'expiration'),
+                       'optional': ('value', 'first_period_reward', 'rate'),
```
KPrasch: :+1: - Excellent improvement

tests/fixtures.py
```diff
@@ -296,6 +298,26 @@ def reset(self):
...
+        def __call__(self):
+            enrico = Enrico(policy_encrypting_key=enacted_blockchain_policy.public_key)
+            message = "Welcome to flippering number {}.".format(len(self.messages)).encode()
+            message_kit, _signature = enrico.encrypt_message(message)
+            self.messages.append((message_kit, enrico))
+            return message_kit, enrico
+
+        def reset(self):
+            self.messages = []
+            self()
```
KPrasch: :upside_down_face:

nucypher/characters/control/specifications.py
vepkenez: Yes, we have an "official" standard which is to declare variables before they get returned. right @KPrasch ?

KPrasch:
Well - I personally do prefer to have named objects before returning them for debugging ease and readability. :-)

tests/fixtures.py
michwill: I think, they have different labels, so it pretty much appears to be the wrong capsule (and yielding a decryption error). It's not incompatibility per se, I think fixtures can be worked on. I just didn't focus on that in this PR - wanted them minimally working.

tests/characters/control/federated/test_rpc_control_f.py
michwill: Yes, that can be done. Do you think it's strictly better?

tests/characters/control/blockchain/test_rpc_control_b.py
michwill: Somehow pytest isn't happy when they have the same filename, albeit in different directories. Thus, `_b` / `_f`

nucypher/characters/control/specifications.py
michwill: Why? Because all `in` should be present in the spec?

tests/characters/control/federated/test_rpc_control_f.py
KPrasch: No - not strictly better - In fact I think the *primary* use case is to unpack 'em all. So I think I like it as-is.

tests/characters/control/blockchain/test_rpc_control_b.py
KPrasch: Can we combine them into a single test - or does this need to be a stand alone module?

tests/fixtures.py
michwill: Ah?
I think, I just copied it :-) Could be done better, yes

michwill:
@KPrasch ok, will return dict perhaps. I understand how that is better

michwill:
Ok, as per conversation, will do `namedtuple`!

PR #1372, opened 2019-09-30
Add @pytest.mark.nightly marker to specify nightly tests

### LINKED ISSUE #1288: Secondary CI workflows
Some types of tests could be safely ran in additional, secondary CI workflows that could be executed, say, nightly. 

One example of this test is the one I just created in #1280 that checks for a correct sampling distribution. In order for the test to be statistically meaningful, it has to use a lot of samples and it can be very slow, but we don't need it to run for all PRs, all the time.

Other examples of tests/scripts are big demos, stress tests, gas estimations, etc.

cygnusv:
* Adds support in our pytest config to the `@pytest.mark.nightly` marker, similar to  `@pytest.mark.slow` but for tests that are specifically designed for nightly builds.
* Uses this marker for running a more robust statistical test for sampling (Closes #1288)
* Schedules tests to run at **06:00 UTC** (08:00 CEST, 23:00 PDT).
   * East coast nucypherinos should be sleeping
   * West coast nucypherinos should be brushing their teeth
  * European nucypherinos are snoozing the alarm for the third time

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -184,7 +184,7 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "0 6 * * *"   # Scheduled for 0800 CEST / 2300 PDT
+          cron: "0 6 * * *"   # Scheduled for 0600 UTC (0800 CEST, 2300 PDT)
           filters:
             branches:
               only:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -184,13 +184,11 @@ workflows:
   nightly:
     triggers:
       - schedule:
-          cron: "45 * * * *" # Just for testing configs
-          # cron: "0 6 * * *"   # Scheduled for 0800 CEST / 2300 PDT
+          cron: "0 6 * * *"   # Scheduled for 0800 CEST / 2300 PDT
```
michwill: Wow. CircleCI can do even that?

PR #1376, opened 2019-10-01
Add commands for MacOS and fixed typos.

afrancht:
- Added commands for MacOS
- Fixed some typos

### DIFF
--- a/docs/source/guides/staking_guide.rst
+++ b/docs/source/guides/staking_guide.rst
@@ -114,13 +114,22 @@ off).
 
 If you don't have a hardware wallet, you can create a software one:
 
+Whilst running the initialized node:
+
 .. code:: bash
 
+    Linux:
     $ geth attach /home/<username>/.ethereum/goerli/geth.ipc
     > personal.newAccount();
     > eth.accounts
     ["0x287a817426dd1ae78ea23e9918e2273b6733a43d"]
-
+    
+    MacOS:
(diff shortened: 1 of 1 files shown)


docs/source/guides/staking_guide.rst
```diff
@@ -114,13 +114,22 @@ off).
 
 If you don't have a hardware wallet, you can create a software one:
 
+Whilst running the initialized node:
```
cygnusv: Maybe put this line together with the previous one:

"If you don't have a hardware wallet, you can create a software one. Whilst running the initialized node:"

docs/source/guides/staking_guide.rst
```diff
@@ -215,8 +226,8 @@ attached a Worker node to your Staker, so you still have to do it!
 Bond an Ursula to a Staker
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-After you create an Ethereum node for your worker (see below about the worker),
-you can set the worker. Stakes which don't yet have a worker will be
+After you create an Ethereum node for your staker (see below about the worker),
```
mswilkison: I think it's actually correct to say the worker has the Ethereum node, right?

cygnusv: YEs, but this step is referred to the Staker, since she is setting now her worker.

KPrasch: It's subtle - but the original intention is to indicate that the staker must know about the worker's address to call `set-worker`.

PR #1365, opened 2019-09-27
Docs for fundamental character concepts/processes including diagrams

derekpierre:
Fixes #1347 .

[X] Alice.grant - How Alice communicates with Ursula and the blockchain for policies
[X] Enrico.encrypt - How Alice and Enrico communicate, and Enrico's use of storage (capsule)
[X] Bob.retrieve - How Bob communicates with Ursula for re-encryptions
[X] Ursula.reencrypt - Receive kFrag from Alice, receive capsule and produce cFrag

Holding off on Alice.revoke() for now until the mechanics of it are finalized ( #1322 )

PR #1377, opened 2019-10-01
Glossary Page for Docs

### LINKED ISSUE #28: Glossary of Terms
We need a glossary of terms for new contributors to look at and understand the internal development lingo (Alice, Bob, Ursula, kFrag, etc).

derekpierre:
While I'm playing with the docs for #1365, I figured we could add some initial work on a glossary that could be updated over time.

To reference the glossary term anywhere in the docs, you can use :term:\`Glossary Term\` - see https://sublime-and-sphinx-guide.readthedocs.io/en/latest/glossary.html.

Fixes #28 .

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -1,0 +1,78 @@
+Glossary
+========
+
+.. _Umbral: https://github.com/nucypher/umbral-doc/blob/master/umbral-doc.pdf
+
+.. glossary::
+
+    Alice
+      *"The Data Owner"* :term:`Character` - retains full control over the data encrypted for her and determines whom to share the data with.
+
+    Bob
+      *"The Data Recipient"* :term:`Character` - the data recipient that :term:`Alice` intends to share data with.
+
+    Capsule
+      Encrypted symmetric key (:term:`KEM`) that is eventually re-encrypted.
+
+    cFrag
+      A fragment of ciphertext that is a partial re-encryption produced by a :term:`kFrag` operation on a :term:`Capsule`.
+
+    Character
+      A common term for any entity fulfilling a particular role in our cryptographic narrative.
+
+    DEM
+      Data encapsulation mechanism - data encrypted with a symmetric key.
+
+    Enrico
+      *"The Encryptor"* :term:`Character` - a data source that encrypts data on behalf of :term:`Alice` and produces a :term:`MessageKit`.
+
+    Felix
+      *"The Faucet"* :term:`Character` - provides *testnet* NU tokens for nodes on the test NuCypher Network.
+
+    KEM
+      Key encapsulation mechanism - a symmetric key encrypted with an asymmetric key
+
+    kFrag
+      A fragment of a :term:`Re-encryption Key`.
+
+    Label
+      A title for a classification/categorization of data according to how it is intended to be shared.
+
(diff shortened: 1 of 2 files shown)


docs/source/glossary.rst
```diff
@@ -0,0 +1,78 @@
...
+    Character
+      A common term for any entity fulfilling a particular role in our cryptographic narrative.
+
+    DEM
+      Data encapsulation mechanism - data encrypted with a symmetric key.
+
+    Enrico
+      *"The Encryptor"* :term:`Character` - a data source that encrypts data on behalf of :term:`Alice` and produces a :term:`MessageKit`.
+
+    Felix
```
mswilkison: Do we want to include Felix in the glossary? Is he important enough to make the cut?

docs/source/glossary.rst
```diff
@@ -0,0 +1,78 @@
...
+    kFrag
+      A fragment of a :term:`Re-encryption Key`.
+
+    Label
+      A title for a classification/categorization of data according to how it is intended to be shared.
+
+    MessageKit
+      The :term:`KEM` and :term:`DEM` encrypted data that is stored together.
+
+    Moe
```
mswilkison: Same question as for Felix.

docs/source/glossary.rst
```diff
@@ -0,0 +1,78 @@
...
+      Public-key encryption.
+
+    PRE
+      Proxy re-encryption.
+
+    Re-encryption Key
+      A key that facilitates the transformation of ciphertext from one encryption key to another.
+
+    Stake
+      A quantity of tokens and escrow duration in periods.
```
mswilkison: Should we define period?

docs/source/glossary.rst
```diff
@@ -0,0 +1,78 @@
...
+      Data encapsulation mechanism - data encrypted with a symmetric key.
+
+    Enrico
+      *"The Encryptor"* :term:`Character` - a data source that encrypts data on behalf of :term:`Alice` and produces a :term:`MessageKit`.
+
+    Felix
+      *"The Faucet"* :term:`Character` - provides *testnet* NU tokens for nodes on the test NuCypher Network.
+
+    KEM
+      Key encapsulation mechanism - a symmetric key encrypted with an asymmetric key
```
cygnusv: KEM and DEM are "approaches" or "methodologies". They are not objects. I think it's better to just not include them in the glossary.

derekpierre: @cygnusv I do reference them as part of the definition for `Capsule`, `MessageKit`. Does that change your mind or no? If not, that's fine - I'll remove the `:term:` references. wdyt?

docs/source/glossary.rst
derekpierre: I would say yes, since outputs of the NuBot faucet funding includes the term "Felix" - also maybe for incentivized testnet etc. we include temporary docs about the faucet and funding process...?

docs/source/glossary.rst
derekpierre: Less strong feeling about "Moe" but people will see it in the code... - at this point feels like why not

docs/source/glossary.rst
derekpierre: ✔️

PR #1378, opened 2019-10-01
[APPROVED - DON'T MERGE YET] Fix potential problem with EIP-1884

vzotova:
[EIP-1884](https://github.com/ethereum/EIPs/blob/dcc573e74adc0e6dd25821ddaabf862e8f85e107/EIPS/eip-1884.md)

Potential issue: 
Problem may appear in calling `someAddress.transfer(ethAmount)` because it uses fixed gas limit (2300). Transfering ETH to some contract executes fallback function which can consume more than 2300 gas and probability of exceeding gas limit becomes greater with EIP-1884. For our network it means that not every intermediary contract can withdraw ETH (policy reward, worklock refund) from our contracts.
Fix:
Instead of not adjustable `transfer()` function use `call("")` with all initial gas limit

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.5.3;
 import "zeppelin/token/ERC20/SafeERC20.sol";
 import "zeppelin/math/SafeMath.sol";
 import "zeppelin/math/Math.sol";
+import "zeppelin/utils/Address.sol";
 import "contracts/lib/AdditionalMath.sol";
 import "contracts/StakingEscrow.sol";
 import "contracts/NuCypherToken.sol";
@@ -19,6 +20,7 @@ contract PolicyManager is Upgradeable {
     using AdditionalMath for uint256;
     using AdditionalMath for int256;
     using AdditionalMath for uint16;
+    using Address for address payable;
 
     event PolicyCreated(
         bytes16 indexed policyId,
@@ -216,9 +218,7 @@ contract PolicyManager is Upgradeable {
         uint256 reward = node.reward;
         require(reward != 0);
         node.reward = 0;
-        // Transfer reward to _recipient
-        (bool success, ) = _recipient.call.value(reward)("");
-        require(success);
+        _recipient.sendValue(reward);
         emit Withdrawn(msg.sender, _recipient, reward);
         return reward;
     }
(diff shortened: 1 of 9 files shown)


KPrasch:
https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1884.md

tests/blockchain/eth/contracts/contracts/PolicyManagerTestSet.sol
```diff
@@ -86,15 +86,25 @@ contract StakingEscrowForPolicyMock {
 
     /**
     * @notice Emulate mint method
+    * @param _staker Staker on behalf of whom minting will be
     * @param _startPeriod Start period for minting
     * @param _numberOfPeriods Number periods for minting
     **/
-    function mint(uint16 _startPeriod, uint16 _numberOfPeriods) external {
+    function mint(address _staker, uint16 _startPeriod, uint16 _numberOfPeriods) public {
```
michwill: Uh, maybe I missed something. Why not `msg.sender`?

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -216,7 +216,8 @@ contract PolicyManager is Upgradeable {
         uint256 reward = node.reward;
         require(reward != 0);
         node.reward = 0;
-        _recipient.transfer(reward);
+        (bool success, ) = _recipient.call.value(reward)("");
```
michwill: It's, of course, terrible if we have to write it this way. But testnet is on Istanbul now.
Is it indeed a problem? And new solc doesn't compile it properly? If so, I'd vote for a merge.
If it 's to be merged, can we add a comment like `// Transfer reward to _recipient`?

tests/blockchain/eth/contracts/contracts/PolicyManagerTestSet.sol
vzotova: It's just for tests

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
vzotova: I don't think that is terrible, looks a little bit ugly, but I don't think "transfer" was the best protection from reentrancy. Even without Istanbul it removes restriction on complexity of fallback function in recepient contract

vzotova: ✔️

vzotova: It will increase size definitely, but size is problem for StakingEscrow and a little bit for Adjudicator, so it'll ok for this 4 places

vzotova: ✔️

PR #1387, opened 2019-10-03
Stake Graphing and Layout; Dark Theme; Organize Status Apps

KPrasch:


nucypher/network/status_app/crawler.py
```diff
@@ -0,0 +1,184 @@
...
+                                                 database=self.DB_NAME,
+                                                 default=True)
+        else:
+            self.log.info(f'Database {self.DB_NAME} already exists, no need to create it')
+
+    def _learn_about_contracts(self):
+        period_range = range(1, 365+1)
+        token_counter = {day: NU.from_nunits(self._moe.staking_agent.get_all_locked_tokens(day)).to_tokens()
+                         for day in period_range}
+        self.__snapshot['future_locked_tokens'] = token_counter
```
derekpierre: Clever! 🤓 . Alternatively, it could be stored in a different table in the DB if you like - then you could do things like, highest vs lowest periods etc., averages (and periods below and above the average) - may/may not be overkill haha.

PR #1388, opened 2019-10-03
Auto-detect HW wallet accounts in TransactingPower 

### LINKED ISSUE #1115: Check that staker has minimum stake in node validation, instead of >0

### LINKED ISSUE #1128: Reliable and thin device auto-detection for Transacting Power and CLI.

### LINKED ISSUE #1369: Staked tokens should be only the locked tokens, not locked + rewards

### LINKED ISSUE #1385: Potential incorrect initialization of TransactingPowers in StakeHolder

cygnusv:
* `TransactingPower` now auto-detects that the account comes from a HW, using the data provided by geth endpoint `personal_listWallets`. Fix #1385. Fix #1128 
* Check min staking size when learning. Also accepts staking for next period, which makes possible creating a stake and running a discoverable Ursula in the same period. Closes #1115 
* Fix confusion between staked and locked+rewards. Fixes #1369 
* Fix lack of client account selection for `nucypher stake collect-reward`
* Complete CLI tests for staking via contracts, except for divide.

### DIFF
--- a/nucypher/crypto/powers.py
+++ b/nucypher/crypto/powers.py
@@ -19,6 +19,7 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 import inspect
 from typing import List, Tuple, Optional
 
+from eth_utils import to_checksum_address
 from constant_sorrow.constants import NO_BLOCKCHAIN_CONNECTION
 from hexbytes import HexBytes
 from umbral import pre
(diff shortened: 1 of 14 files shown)


nucypher/crypto/powers.py
```diff
@@ -119,8 +120,19 @@ def __init__(self, account: str, password: str = None, cache: bool = False):
         self.blockchain = BlockchainInterfaceFactory.get_interface()
         self.__account = account
 
-        # TODO: Is there a better way to design this Flag? See #1385
-        self.device = True if not password else False
+        # TODO: Temporary fix for #1128 and #1385. It's ugly af, but it works. Move somewhere else?
```
michwill: Alternatively, could be in BlockchainInterface

cygnusv:
Included in #1402

PR #1390, opened 2019-10-03
Optional `treasure_map` argument for Bob.retrieve

michwill:
Currently Ursulas store all treasure maps which is not feasible in the long term.
One way to fix that is to let Bob use treasure maps from some side channel. This PR implements `treasure_map` as an optional argument for Bob.retrieve.
Question: do we want to cache "oriented" (decrypted) treasure maps after the first use by Bob?
- [x] `treasure_map` in `Bob.retrieve`;
- [x] Tests for `Bob.retrieve`;
- [x] `treasure_map` in Bob's character control
- [x] Tests for character control with `treasure_map`


cygnusv:
Does it make sense to include this somehow as part of "birdcage", which seems to address a lot of Bob stuff? What do you think @michwill, @jMyles?

tests/characters/control/blockchain/test_web_control_blockchain.py
```diff
@@ -186,8 +189,25 @@ def test_bob_web_character_control_retrieve(bob_web_controller_test_client, retr
...
+    method_name, params = make_retrieve_control_request().values()
+    wrong_tmap = TreasureMap(m=0)
+    wrong_tmap.prepare_for_publication(
+            blockchain_bob.public_keys(DecryptingPower),
+            blockchain_bob.public_keys(SigningPower),
+            blockchain_alice.stamp,
+            b'Wrong!')
+    wrong_tmap = b64encode(bytes(wrong_tmap)).decode()
+    params['treasure_map'] = wrong_tmap
+    with pytest.raises(Exception):
```
derekpierre: Is there a more specific `Exception` that can be checked here? Or maybe an error response code is more appropriate - I guess this is the question in the comment below...

tests/characters/control/federated/test_web_control_federated.py
```diff
@@ -186,9 +186,6 @@ def test_bob_web_character_control_retrieve(bob_web_controller_test_client, retr
     response = bob_web_controller_test_client.post(endpoint, data=json.dumps({'bad': 'input'}))
     assert response.status_code == 400
 
-    del(params['alice_verifying_key'])
```
derekpierre: What was the reason for removing these lines?

tests/characters/test_bob_joins_policy_and_retrieves.py
```diff
@@ -161,3 +161,26 @@ def test_treasure_map_serialization(enacted_federated_policy, federated_bob):
...
+    message_kit, data_source = capsule_side_channel()
+    treasure_map = enacted_federated_policy.treasure_map
+    alice_verifying_key = enacted_federated_policy.alice.stamp
+
+    # Teach Bob about the network
+    federated_bob.remember_node(list(federated_ursulas)[0])
+    federated_bob.learn_from_teacher_node(eager=True)
+
+    # Deserialized treasure map
+    federated_bob.retrieve(
```
derekpierre: Are we able to check the returned cleartext?

jMyles:
@michwill, I wonder also with @cygnusv if you might consider pointing this into `birdcage` so that I can manage a merge there with the various changes to `retrieve(...)`.  Does this make sense to you?

tests/characters/control/blockchain/test_web_control_blockchain.py
cygnusv: Yes, we should catch a more specific error here.

michwill: Let me see if I can make a treasure map with 1 Ursula instead

tests/characters/test_bob_joins_policy_and_retrieves.py
michwill: Yes we are (fixed)

nucypher/characters/control/interfaces.py
```diff
@@ -192,7 +192,8 @@ def retrieve(self,
                  label: bytes,
                  policy_encrypting_key: bytes,
                  alice_verifying_key: bytes,
-                 message_kit: bytes):
+                 message_kit: bytes,
+                 treasure_map: bytes = None):
```
tuxxy: Can we use a `Union` here to specify that you can pass `bytes` or a `TreasureMap`?

nucypher/characters/lawful.py
```diff
@@ -595,14 +594,20 @@ def get_treasure_map_from_known_ursulas(self, network_middleware, map_id):
 
         return treasure_map
 
-    def generate_work_orders(self, map_id, *capsules, num_ursulas=None, cache=False):
+    def generate_work_orders(self, map_or_id, *capsules, num_ursulas=None, cache=False):
```
tuxxy: Not sure if I like this as `map_or_id`. I think I'd rather see a more explicit param for a treasure_map.

tests/characters/control/blockchain/conftest.py
```diff
@@ -114,17 +115,19 @@ def join_control_request(blockchain_bob, enacted_blockchain_policy):
...
+        message_kit, data_source = capsule_side_channel_blockchain()
+
+        params = {
+            'label': enacted_blockchain_policy.label.decode(),
+            'policy_encrypting_key': bytes(enacted_blockchain_policy.public_key).hex(),
+            'alice_verifying_key': bytes(enacted_blockchain_policy.alice.stamp).hex(),
+            'message_kit': b64encode(message_kit.to_bytes()).decode(),
+        }
+        return OrderedDict([('method', method_name), ('params', params)])
+    return f
```
tuxxy: :+1:

nucypher/characters/lawful.py
michwill: How do you imagine that? Ideally, I can see `generate_work_orders_from_tmap` in addition to this method (which would take the id). But that's such a terrible name! Suggestions?

michwill: Wouldn't like to have both `map_id` and `treasure_map` as keyword arguments though where you ned just one - that's terrible!

michwill: Anyway. I think, I still like `map_or_id` more than alternatives, but adding a type hint

cygnusv: I like the `map_or_id` thing. Anyway, a type hint can't hurt.

michwill: Anyway, looks like after the rebase it's gonna be the way @tuxxy wants anyway!

michwill:
Hmm, need to make it working after a long rebase. Also birdcage dependencies seem to be broken

michwill:
Hmm, weird. Fails in CI (e.g. py-evm and eth-tester versions) but works locally...

michwill:
Ah, need `requirements.txt`

PR #1402, opened 2019-10-08
Auto-detect HW accounts in TransactingPower. Finalize basic staking via contract CLI functionality.

### LINKED ISSUE #1115: Check that staker has minimum stake in node validation, instead of >0

### LINKED ISSUE #1128: Reliable and thin device auto-detection for Transacting Power and CLI.

### LINKED ISSUE #1351: Redesign allocation registry

### LINKED ISSUE #1369: Staked tokens should be only the locked tokens, not locked + rewards

### LINKED ISSUE #1385: Potential incorrect initialization of TransactingPowers in StakeHolder

### LINKED ISSUE #1418: Past stakes shouldn't be printed as "Active" with `nucypher stake list`

cygnusv:
* `TransactingPower` now auto-detects that the account comes from a HW, using the data provided by geth endpoint `personal_listWallets`. Fix #1385. Fix #1128 
* Check min staking size when learning. Also accepts staking for next period, which makes possible creating a stake and running a discoverable Ursula in the same period. Closes #1115 
* Fix confusion between staked and locked+rewards. Fixes #1369 
* Fix lack of client account selection for `nucypher stake collect-reward`
* Complete CLI tests for staking via contracts, except for divide.
* Revamp output files of allocation process and how allocation users stake. Closes #1351 
* Fix datetimes printed by `nucypher stake`.
* `nucypher stake list` now only prints active stakes. Fixes #1418

### DIFF
--- a/docs/source/guides/staking_guide.rst
+++ b/docs/source/guides/staking_guide.rst
@@ -392,30 +392,31 @@ since the contract can't perform transactions, and it's the preallocation user
 (also known as the "`beneficiary`" of the contract)
 who has to perform staking operations.
 
+As part of the preallocation process, beneficiaries receive an allocation file,
+containing the ETH addresses of their beneficiary account and corresponding
+preallocation contract.
+
 In general, preallocation users can use all staking-related operations offered
-by the CLI in the same way as described above, except that they have to include
-the ``--escrow`` flag to indicate they are acting as beneficiaries.
+by the CLI in the same way as described above, except that they have to specify
+the path to the allocation file using the option ``--allocation-filepath PATH``.
 
 For example, to create a stake:
 
 .. code:: bash
 
-    (nucypher)$ nucypher stake create --hw-wallet --escrow
+    (nucypher)$ nucypher stake create --hw-wallet --allocation-filepath PATH
 
 
 Or to set a worker:
 
 .. code:: bash
 
-    (nucypher)$ nucypher stake set-worker --hw-wallet --escrow
+    (nucypher)$ nucypher stake set-worker --hw-wallet --allocation-filepath PATH
 
 
-As an alternative to the ``--escrow`` flag, preallocation users can directly specify
-their beneficiary address with the ``--beneficiary-address ADDRESS`` flag.
-
-The CLI will automatically look for a file named ``allocation_registry.json`` in
-the default installation path, although a custom path can be used with the flag
-``--allocation-filepath /path/to/allocation_registry.json``.
+As an alternative to the ``--allocation-filepath`` flag, preallocation users
+can directly specify their beneficiary and staking contract addresses with the
+``--beneficiary-address ADDRESS`` and ``--staking-address ADDRESS``, respectively.
 
 
 Inline Method
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -64,7 +64,12 @@ from nucypher.blockchain.eth.registry import (
 from nucypher.blockchain.eth.token import NU, Stake, StakeList, WorkTracker
 from nucypher.blockchain.eth.utils import datetime_to_period, calculate_period_duration, datetime_at_period
 from nucypher.characters.control.emitters import StdoutEmitter
-from nucypher.cli.painting import paint_contract_deployment, paint_input_allocation_file, paint_deployed_allocations
+from nucypher.cli.painting import (
+    paint_contract_deployment,
+    paint_input_allocation_file,
+    paint_deployed_allocations,
+    write_deployed_allocations_to_csv
+)
 from nucypher.config.constants import DEFAULT_CONFIG_ROOT
 from nucypher.crypto.powers import TransactingPower
 
@@ -426,6 +431,8 @@ class ContractAdministrator(NucypherTokenActor):
         template_parent_path = Path(allocation_registry.filepath).parent  # Use same folder as allocation registry
         template_filepath = os.path.join(template_parent_path, template_filename)
         AllocationRegistry(filepath=template_filepath).write(registry_data=allocation_template)
+        if emitter:
+            emitter.echo(f"Saved allocation template file to {template_filepath}", color='blue', bold=True)
 
         # Deploy each allocation contract
         with click.progressbar(length=total_deployment_transactions,
@@ -482,6 +489,12 @@ class ContractAdministrator(NucypherTokenActor):
 
             if emitter:
                 paint_deployed_allocations(emitter, allocated, failed)
+
+            csv_filename = f'allocations-{self.deployer_address[:6]}-{maya.now().epoch}.csv'
+            csv_filepath = os.path.join(template_parent_path, csv_filename)
+            write_deployed_allocations_to_csv(csv_filepath, allocated, failed)
+            if emitter:
+                emitter.echo(f"Saved allocation summary CSV to {csv_filepath}", color='blue', bold=True)
 
             if failed:
                 # TODO: More with these failures: send to isolated logfile, and reattempt
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -584,8 +584,8 @@ class Staker(NucypherTokenActor):
         self.__worker_address = None
 
         # Blockchain
-        self.policy_agent = ContractAgency.get_agent(PolicyManagerAgent, registry=self.registry)
-        self.staking_agent = ContractAgency.get_agent(StakingEscrowAgent, registry=self.registry)
+        self.policy_agent = ContractAgency.get_agent(PolicyManagerAgent, registry=self.registry)  # type: PolicyManagerAgent
+        self.staking_agent = ContractAgency.get_agent(StakingEscrowAgent, registry=self.registry)  # type: StakingEscrowAgent
         self.economics = TokenEconomicsFactory.get_economics(registry=self.registry)
 
         # Staking via contract
@@ -835,9 +835,13 @@ class Staker(NucypherTokenActor):
             receipt = self.staking_agent.mint(staker_address=self.checksum_address)
         return receipt
 
-    def calculate_reward(self) -> int:
+    def calculate_staking_reward(self) -> int:
         staking_reward = self.staking_agent.calculate_staking_reward(staker_address=self.checksum_address)
         return staking_reward
+
+    def calculate_policy_reward(self) -> int:
+        policy_reward = self.policy_agent.get_reward_amount(staker_address=self.checksum_address)
+        return policy_reward
 
     @only_me
     @save_receipt
@@ -857,7 +861,7 @@ class Staker(NucypherTokenActor):
     def collect_staking_reward(self) -> str:
         """Withdraw tokens rewarded for staking."""
         if self.is_contract:
-            reward_amount = self.staking_agent.calculate_staking_reward(staker_address=self.checksum_address)
+            reward_amount = self.calculate_staking_reward()
             self.log.debug(f"Withdrawing staking reward ({NU.from_nunits(reward_amount)}) to {self.checksum_address}")
             receipt = self.preallocation_escrow_agent.withdraw_as_staker(value=reward_amount)
         else:
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -421,15 +421,15 @@ class ContractAdministrator(NucypherTokenActor):
 
         # Create an allocation template file, containing the allocation contract ABI and placeholder values
         # for the beneficiary and contract addresses. This file will be shared with all allocation users.
-        empty_user_escrow_deployer = UserEscrowDeployer(registry=self.registry)
-        allocation_contract_abi = empty_user_escrow_deployer.get_contract_abi()
+        empty_allocation_escrow_deployer = PreallocationEscrowDeployer(registry=self.registry)
+        allocation_contract_abi = empty_allocation_escrow_deployer.get_contract_abi()
         allocation_template = {
             "BENEFICIARY_ADDRESS": ["ALLOCATION_CONTRACT_ADDRESS", allocation_contract_abi]
         }
 
+        parent_path = Path(allocation_registry.filepath).parent  # Use same folder as allocation registry
         template_filename = IndividualAllocationRegistry.REGISTRY_NAME
-        template_parent_path = Path(allocation_registry.filepath).parent  # Use same folder as allocation registry
-        template_filepath = os.path.join(template_parent_path, template_filename)
+        template_filepath = os.path.join(parent_path, template_filename)
         AllocationRegistry(filepath=template_filepath).write(registry_data=allocation_template)
         if emitter:
             emitter.echo(f"Saved allocation template file to {template_filepath}", color='blue', bold=True)
@@ -474,9 +474,20 @@ class ContractAdministrator(NucypherTokenActor):
 
                 else:
                     allocation_receipts[beneficiary] = receipts
-                    principal_address = deployer.contract_address
-                    self.log.info(f"Created PreallocationEscrow contract at {principal_address} for beneficiary {beneficiary}.")
-                    allocated.append((allocation, principal_address))
+                    allocation_contract_address = deployer.contract_address
+                    self.log.info(f"Created {deployer.contract_name} contract at {allocation_contract_address} "
+                                  f"for beneficiary {beneficiary}.")
+                    allocated.append((allocation, allocation_contract_address))
+
+                    # Create individual allocation file
+                    individual_allocation_filename = f'allocation-{beneficiary}.json'
+                    individual_allocation_filepath = os.path.join(parent_path, individual_allocation_filename)
+                    individual_allocation_file_data = {
+                        'beneficiary_address': beneficiary,
+                        'contract_address': allocation_contract_address
+                    }
+                    with open(individual_allocation_filepath, 'w') as outfile:
+                        json.dump(individual_allocation_file_data, outfile)
 
                     if emitter:
                         blockchain = BlockchainInterfaceFactory.get_interface()
@@ -486,12 +497,14 @@ class ContractAdministrator(NucypherTokenActor):
                                                   emitter=emitter,
                                                   chain_name=blockchain.client.chain_name,
                                                   open_in_browser=False)
+                        emitter.echo(f"Saved individual allocation file to {individual_allocation_filepath}",
+                                     color='blue', bold=True)
 
             if emitter:
                 paint_deployed_allocations(emitter, allocated, failed)
 
             csv_filename = f'allocations-{self.deployer_address[:6]}-{maya.now().epoch}.csv'
-            csv_filepath = os.path.join(template_parent_path, csv_filename)
+            csv_filepath = os.path.join(parent_path, csv_filename)
             write_deployed_allocations_to_csv(csv_filepath, allocated, failed)
             if emitter:
                 emitter.echo(f"Saved allocation summary CSV to {csv_filepath}", color='blue', bold=True)
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -21,6 +21,7 @@ import os
 from datetime import datetime
 from decimal import Decimal
 from json import JSONDecodeError
+from pathlib import Path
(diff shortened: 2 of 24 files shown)


KPrasch:
looks like it's passing :-) why else do you you anticipate including?

nucypher/cli/actions.py
```diff
@@ -288,8 +293,9 @@ def make_cli_character(character_config,
         CHARACTER = character_config(known_nodes=teacher_nodes,
                                      network_middleware=character_config.network_middleware,
                                      **config_args)
-    except CryptoError as e:
-        raise character_config.keyring.AuthenticationFailed(str(e))
+    except CryptoError:
+        raise character_config.keyring.AuthenticationFailed("Failed keyring unlocking. "
```
derekpierre:
```suggestion
        raise character_config.keyring.AuthenticationFailed("Failed to unlock keyring. "
```

nucypher/cli/characters/stake.py
```diff
@@ -152,25 +150,35 @@ def stake(click_config,
...
+            initial_address = individual_allocation.beneficiary_address
+        elif beneficiary_address and staking_address:
+            individual_allocation = IndividualAllocationRegistry(beneficiary_address=beneficiary_address,
+                                                                 contract_address=staking_address)
+            initial_address = beneficiary_address
+        else:
+            option = "--beneficiary_address" if beneficiary_address else "--staking-address"
+            raise click.BadOptionUsage(option_name=option,
+                                       message=f"You must specify both --beneficiary-address and --staking-address. "
+                                               f"Only {option} was provided. As an alternative you can input just an "
```
derekpierre:
```suggestion
                                               f"Only {option} was provided. As an alternative, you can simply provide an "
```

nucypher/network/nodes.py
```diff
@@ -1012,9 +1013,19 @@ def _staker_is_really_staking(self, registry: BaseContractRegistry) -> bool:
...
+        try:
+            economics = TokenEconomicsFactory.get_economics(registry=registry)
+        except Exception:
+            raise  # TODO: Get StandardEconomics
+
+        min_stake = economics.minimum_allowed_locked
+
+        stake_current_period = staking_agent.get_locked_tokens(staker_address=self.checksum_address, periods=0)
+        stake_next_period = staking_agent.get_locked_tokens(staker_address=self.checksum_address, periods=1)
+        is_staking = max(stake_current_period, stake_next_period) >= min_stake
```
derekpierre: Just confirming what we are checking for here - Is our goal to make sure that the stake for this period and the next are both greater than `min_stake`? If so, should we instead use min? i.e.:
```suggestion
        is_staking = min(stake_current_period, stake_next_period) >= min_stake
```

Or is it that the stake for the current period could be 0 (not staking until UTC midnight), and the stake for the next period is correctly set, in which case `max` will usually just produce the stake for the next period?

nucypher/blockchain/eth/actors.py
```diff
@@ -838,7 +869,7 @@ def collect_staking_reward(self) -> str:
     def withdraw(self, amount: NU) -> str:
         """Withdraw tokens (assuming they're unlocked)"""
         if self.is_contract:
-            receipt = self.user_escrow_agent.withdraw_as_staker(amount=int(amount))
+            receipt = self.user_escrow_agent.withdraw_as_staker(value=int(amount))
```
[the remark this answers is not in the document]
cygnusv: We tend to use NU in higher-level methods, while directly ints (in nunits) at the agent level. Note that we cast to int in line 872.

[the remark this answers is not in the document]
derekpierre: Ah! didn't realize the `int(...)` cast does the conversion... 👍 .

nucypher/network/nodes.py
cygnusv: It's the latter, yes.

.circleci/config.yml
```diff
@@ -267,7 +267,9 @@ commands:
     steps:
       - run:
           name: Install Python Dependencies with Pip
-          command: pip3 install --user -e .
+          command: |
+            python setup.py install_egg_info --install-dir=/home/circleci/.local/lib/python3.7/site-packages
```
mswilkison:
```suggestion
            python setup.py install_egg_info --install-dir=/home/circleci/.local/lib/python3.7/site-packages
```
This is #1411 - should we be duplicating it here?

cygnusv: Right, we don't. It's just a placeholder before we merge #1411 to master. Historically, I tried this here before, and once it worked I created #1411, but I will remove this commit once that's merged.

nucypher/cli/actions.py
cygnusv: ✅

nucypher/cli/characters/stake.py
cygnusv: ✅

tests/cli/ursula/test_stake_via_allocation_contract.py
```diff
@@ -161,9 +208,69 @@ def test_stake_set_worker(click_runner,
...
+
+    # Ok ok, let's set the worker again.
+
+    init_args = ('stake', 'set-worker',
+                 '--config-file', stakeholder_configuration_file_location,
+                 '--allocation-filepath', MOCK_INDIVIDUAL_ALLOCATION_FILEPATH,
+                 '--worker-address', manual_worker,
+                 '--force')
+
+    user_input = f'{INSECURE_DEVELOPMENT_PASSWORD}'
```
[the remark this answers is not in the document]
cygnusv: ✅

.circleci/config.yml
cygnusv: ✅

nucypher/blockchain/eth/registry.py
```diff
@@ -376,20 +379,26 @@ def search(self, beneficiary_address: str = None, contract_address: str = None):
...
+                result = contract_data
             except KeyError:
                 raise self.UnknownBeneficiary
 
         elif contract_address:
             records = list()
             for beneficiary_address, contract_data in allocation_data.items():
-                contract_address, contract_abi = contract_data['address'], contract_data['abi']
-                records.append(dict(address=contract_address, abi=contract_abi))
+                if contract_address == contract_data[0]:
```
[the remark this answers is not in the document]
cygnusv: It's a good idea, but since it should involve the same change to the contract registry, besides the allocation registry, I think it falls out of the scope of this PR.

nucypher/cli/characters/stake.py
```diff
@@ -63,6 +63,7 @@
 @click.option('--lock-until', help="Period to release re-staking lock", type=click.IntRange(min=0))
 @click.option('--beneficiary-address', help="Address of a pre-allocation beneficiary", type=EIP55_CHECKSUM_ADDRESS)
 @click.option('--allocation-filepath', help="Path to individual allocation file", type=EXISTING_READABLE_FILE)
+@click.option('--all', help="List all stakes, including inactive", is_flag=True)
```
michwill: `--all` sounds like a too general word for this option. Maybe `--show-inactive`, or `--with-inactive`?

cygnusv: I agree, although @szotov told me something today about how this are not really "inactive", so perhaps we have to rethink this option altogether.

docs/source/guides/staking_guide.rst
```diff
@@ -392,30 +392,31 @@ since the contract can't perform transactions, and it's the preallocation user
...
 
-As an alternative to the ``--escrow`` flag, preallocation users can directly specify
-their beneficiary address with the ``--beneficiary-address ADDRESS`` flag.
 
-The CLI will automatically look for a file named ``allocation_registry.json`` in
-the default installation path, although a custom path can be used with the flag
-``--allocation-filepath /path/to/allocation_registry.json``.
+As an alternative to the ``--allocation-filepath`` flag, preallocation users
+can directly specify their beneficiary and staking contract addresses with the
+``--beneficiary-address ADDRESS`` and ``--staking-address ADDRESS``, respectively.
```
michwill: So, staking address is address of the contract which will stake for this beneficiary?

nucypher/blockchain/eth/registry.py
```diff
@@ -184,7 +187,7 @@ def search(self, contract_name: str = None, contract_address: str = None) -> tup
             raise self.UnknownContract(contract_name)
 
         if contract_address and len(contracts) > 1:
-            m = "Multiple records returned for address {}"
+            m = f"Multiple records returned for address {contract_address}"
             self.log.critical(m)
             raise self.IllegalRegistry(m.format(contract_address))
```
michwill: You mean, `raise self.IllegalRegistry(m)`?

nucypher/blockchain/eth/registry.py
```diff
@@ -440,3 +449,32 @@ def read(self) -> list:
...
+    def __init__(self, beneficiary_address: str, contract_address: str, contract_abi=None, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self.beneficiary_address = beneficiary_address
+        self.contract_address = contract_address
+
+        if not contract_abi:
+            # Download individual allocation template to extract contract_abi
+            individual_allocation_template = json.loads(self.fetch_latest_publication())
+            contract_data = [*individual_allocation_template.values()].pop()
```
michwill: Ugh, what's happening here?
Upon closer inspection,
```
inidividual_allocation_template = {'BENEFICIARY_ADDRESS':
    ['ALLOCATION_CONTRACT_ADDRESS', <abi>}
```
Can `BENEFICIARY_ADDRESS` be an address, or is it always a template if we are here, and can be accessed by key?

michwill: Main concern with the method above is that the dict returned by json.load is not ordered (although in python 3.7 it is). But it has only one value...

nucypher/blockchain/eth/utils.py
```diff
@@ -31,22 +31,33 @@ def datetime_to_period(datetime: maya.MayaDT, seconds_per_period: int) -> int:
...
-    if delta_periods:
-        target_period = now + maya.timedelta(days=delta_periods)
+        # +
+        if delta_periods:
+            target_datetime = now + maya.timedelta(days=delta_periods)
 
-    # -
-    else:
-        target_period = now - maya.timedelta(days=delta_periods)
+        # -
```
michwill: Wait... `else` works only when `delta_periods == 0`!

Did you mean simply always `now + timedelta(days=delta_periods)`? Timedelta understands negative time too.

Also why days when we have `seconds_per_period` (what if it's not 1 day)? Maybe `maya.timedelta(seconds=1) * seconds_per_period * delta_periods` or something?

nucypher/cli/actions.py
```diff
@@ -306,8 +312,10 @@ def make_cli_character(character_config,
 
 def select_stake(stakeholder, emitter) -> Stake:
     stakes = stakeholder.all_stakes
-    enumerated_stakes = dict(enumerate(stakes))
-    painting.paint_stakes(stakes=stakes, emitter=emitter)
+    active_stakes = sorted((stake for stake in stakes if stake.is_active),
```
michwill: :+1: 
Much needed to sort them!

michwill: Btw I am wondering: is it possible to have stakes sorted in the order addresses are derived by geth?

nucypher/crypto/powers.py
```diff
@@ -123,8 +124,19 @@ def __init__(self,
...
-        self.device = True if not password else False
+        # TODO: Temporary fix for #1128 and #1385. It's ugly af, but it works. Move somewhere else?
+        try:
+            wallets = self.blockchain.client.wallets
+        except AttributeError:
+            is_from_hw_wallet = False
+        else:
+            HW_WALLET_URL_PREFIXES = ('trezor', 'ledger')
+            hw_accounts = [w['accounts'] for w in wallets if w['url'].startswith(HW_WALLET_URL_PREFIXES)]
+            hw_addresses = [to_checksum_address(account['address']) for sublist in hw_accounts for account in sublist]
```
michwill: Almost feels wasteful that we do this cycle for every account. But shouldn't be slow, I guess?

nucypher/utilities/sandbox/constants.py
```diff
@@ -129,6 +129,8 @@ def select_test_port() -> int:
 
 MOCK_ALLOCATION_REGISTRY_FILEPATH = os.path.join(BASE_TEMP_DIR, f'{BASE_TEMP_PREFIX}test-allocation-registry-{str(datetime.now().strftime(DATETIME_FORMAT))}.json')
 
+MOCK_INDIVIDUAL_ALLOCATION_FILEPATH = os.path.join(BASE_TEMP_DIR, f'{BASE_TEMP_PREFIX}test-individual-allocation-{str(datetime.now().strftime(DATETIME_FORMAT))}.json')
```
michwill: I don't think `str()` is needed here - `strftime()` returns a string anyway

docs/source/guides/staking_guide.rst
cygnusv: Yes, exactly.

nucypher/blockchain/eth/registry.py
cygnusv: ✅

nucypher/utilities/sandbox/constants.py
cygnusv: ✅

nucypher/crypto/powers.py
cygnusv: It seems too much but in most cases (non hw accounts) these sublists only have one element. They have more than one just when they are hw accounts.

nucypher/cli/actions.py
cygnusv: I guess we could try to preserve that order, but I don't know if geth enforces some ordering criteria. I think that alphabetical order is fine here, tbh.

nucypher/blockchain/eth/registry.py
cygnusv: Yes, the template is always of the form you wrote above, with just one entry. I added a check that raises an exception otherwise.

nucypher/blockchain/eth/registry.py
```diff
@@ -440,3 +449,34 @@ def read(self) -> list:
...
+
+        self.beneficiary_address = beneficiary_address
+        self.contract_address = contract_address
+
+        if not contract_abi:
+            # Download individual allocation template to extract contract_abi
+            individual_allocation_template = json.loads(self.fetch_latest_publication())
+            if len(individual_allocation_template) != 1:
+                raise self.IllegalRegistry("Individual allocation template must contain a single entry")
+            contract_data = list(individual_allocation_template.values()).pop()
```
michwill: What I've meant is: would it be more appropriate (and readable) to do:
```
contract_data = individual_allocation_template['BENEFICIARY_ADDRESS']
```
?

nucypher/blockchain/eth/token.py
```diff
@@ -220,8 +222,14 @@ def __repr__(self) -> str:
         return r
 
     def __eq__(self, other) -> bool:
+        # TODO: Is this right? Two stakes from different accounts, durations, etc, but of the same value, are equal?
```
derekpierre: Yes, this seems off...

nucypher/blockchain/eth/utils.py
derekpierre: +1 on the use of seconds instead

nucypher/blockchain/eth/actors.py
```diff
@@ -804,10 +835,14 @@ def mint(self) -> Tuple[str, str]:
             receipt = self.staking_agent.mint(staker_address=self.checksum_address)
         return receipt
 
-    def calculate_reward(self) -> int:
+    def calculate_staking_reward(self) -> int:
```
tuxxy: :+1: good rename

nucypher/blockchain/eth/registry.py
```diff
@@ -262,8 +265,8 @@ def _destroy(self) -> None:
     @classmethod
     def from_dict(cls, payload: dict, **overrides) -> 'LocalContractRegistry':
         payload.update({k: v for k, v in overrides.items() if v is not None})
-        blockchain = cls(filepath=payload['filepath'])
-        return blockchain
+        registry = cls(filepath=payload['filepath'])
+        return registry
```
tuxxy: :+1:

nucypher/blockchain/eth/token.py
```diff
@@ -205,9 +205,11 @@ def __init__(self,
 
         # Time
         self.start_datetime = datetime_at_period(period=first_locked_period,
-                                                 seconds_per_period=self.economics.seconds_per_period)
+                                                 seconds_per_period=self.economics.seconds_per_period,
+                                                 start_of_period=True)
```
tuxxy: This is a bit confusing for a boolean flag. Would it be better as `is_start_of_period`?

nucypher/cli/actions.py
tuxxy: Would also agree that alphabetical is fine here.

nucypher/cli/characters/stake.py
tuxxy: Yeah, this needs a rename. Does this have an issue open?

nucypher/cli/characters/stake.py
```diff
@@ -152,25 +152,35 @@ def stake(click_config,
...
+            initial_address = beneficiary_address
+        else:
+            option = "--beneficiary_address" if beneficiary_address else "--staking-address"
+            raise click.BadOptionUsage(option_name=option,
+                                       message=f"You must specify both --beneficiary-address and --staking-address. "
+                                               f"Only {option} was provided. As an alternative, you can simply "
+                                               f"provide an individual allocation with --allocation-file <PATH>")
     else:
-        allocation_registry = None
+        individual_allocation = None
```
tuxxy: :+1:

nucypher/blockchain/eth/utils.py
cygnusv: opened https://github.com/nucypher/nucypher/issues/1435

PR #1404, opened 2019-10-08
Add "TreasureMap" to the Glossary.

### LINKED ISSUE #1403: Glossary needs a definition for Treasure Map

derekpierre:
Also, general cleanup.

Fixes #1403

### DIFF
--- a/docs/source/glossary.rst
+++ b/docs/source/glossary.rst
@@ -39,7 +39,7 @@ Glossary
       A title for a classification/categorization of data according to how it is intended to be shared.
 
     MessageKit
-      The :term:`KEM` and :term:`DEM` encrypted data that are stored together.
+      The ciphertext (data encrypted with a symmetric key) and the :term:`Capsule` (encrypted symmetric key) that are stored together.
 
     Moe
       *"The Monitor"* :term:`Character` - provides a high-level view of the NuCypher Network.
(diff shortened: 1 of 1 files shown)


docs/source/glossary.rst
```diff
@@ -39,7 +39,7 @@ Glossary
       A title for a classification/categorization of data according to how it is intended to be shared.
 
     MessageKit
-      The :term:`KEM` and :term:`DEM` encrypted data that is stored together.
+      The :term:`KEM` and :term:`DEM` encrypted data that are stored together.
```
cygnusv: Something that was missing from a previous PR is to remove references to KEM and DEM. I think it's not NuCypher specific, and additionally, it's not being used 100% correct. I think we can safely define MessageKits this in terms of Capsule and Ciphertext, as Umbral does.

docs/source/glossary.rst
```diff
@@ -71,11 +71,14 @@ Glossary
     Stamp
       The public key for a :term:`Character`'s signing key pair.
 
+    Treasure Map
+      The locations of :term:`Ursulas<Ursula>` that have the relevant :term:`kFrags<kFrag>` for a policy. :term:`Bob` will use the treasure map to determine which :term:`Ursulas<Ursula>` to contact when he wants to retrieve the data :term:`Alice` has shared with him.
```
cygnusv: From this wording, it seems that Ursulas have the data that Bob needs. Perhaps just change "retrieve" by "decrypt"?

docs/source/glossary.rst
derekpierre: @KPrasch added a comment to the previous PR to indicate that it made sense to define KEM/DEM usage in NuCypher's specific case - https://github.com/nucypher/nucypher/pull/1377#discussion_r330188704.

Updated the definition to use `Capsule` and `ciphertext`

derekpierre: ✔️

docs/source/glossary.rst
derekpierre: ✔️ 

Used `re-encrypt`

PR #1126, opened 2019-07-18
[Do Not Merge] WorkLock Deployer, Agent, Economics and CLI

KPrasch:
#### Overview

- Introduces `WorkLockDeployer`
- Introduces `WorkLockAgent`
- Introduces `TestEconomics`
- Adds WorkLock economics to `BaseEconomics`
- Introduces WorkLock CLI: `nucypher worklock <ACTION> <OPTIONS>`

###### Injected

- worklock supply - Total number of NU available via WorkLock commitment
- start_date - Bidding window open time
- end_date - Bidding window close time
- locked_periods - Commitment duration of stake
- deposit_rate - Reward Ratio in wei (1 means 1 wei:1 NuNit)
- refund rate - Rate per period in wei of refunds.

###### Calculated

- maximum bid
- minimum bid

#####  WorkLock Deployment

WorkLock is included in the standard `DEPLOYMENT_SERIES`, now defined on `ContractAdministrator`.

Outstanding Question: Do we want to include WorkLock funding with the deployment series?

###### Status
View the current on-chain status of the WorkLock contract in detail:
`nucypher worklock status --provider <PROVIDER_URI>`

###### Bid
Bid for NU in ETH, transferring the bid to the WorkLock contract:
`nucypher worklock bid --value <VALUE> --bidder-address <BIDDER_ADDRESS> ...`

###### Claim
After the bidding widow closes, bidders claim tokens in the form of a stake:
`nucypher worklock claim --bidder-address <BIDDER_ADDRESS> ...`

###### Refund
After the staker's worker has performed work, ETH can be refunded:
`nucypher worklock refund --bidder-address <BIDDER_ADDRESS> ...`

##### WorkLock Claims

Note that participation in the Worklock commits the bidding address to become a staker address.
This effects how new stakes are created via CLI. 

After running `nucypher worklock claim ...` `StakeHolder`'s can proceed directly to `nucypher set-worker ...`, skipping `nucypher stake create ...`.

PR #1407, opened 2019-10-09
WorkLock updates

vzotova:
Included #1321 
Opens #1540, #1529, #1508

* Burn tokens in `StakingEscrow` (for future reward)
* Ability to burn remaining tokens in `WorkLock`
* Explicit initialization for `StakingEscrow` and `WorkLock`
* New function - `cancelBid`
* Tokens locked in `PreallocationEscrow` in `claim`
* Ability to change `WorkLock` address in `StakingEcrow` if test flag was set <-- (what do you think, is it ok?)

### DIFF
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -598,8 +598,8 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
         the_escrow_contract = wrapped_escrow_contract
 
         # 3 - Transfer the reward supply tokens to StakingEscrow #
-        reward_function = self.token_contract.functions.transfer(the_escrow_contract.address,
-                                                                 self.economics.erc20_reward_supply)
+        reward_function = self.token_contract.functions.approve(the_escrow_contract.address,
+                                                                self.economics.erc20_reward_supply)
 
         # TODO: Confirmations / Successful Transaction Indicator / Events ??  - #1193, #1194
         reward_receipt = self.blockchain.send_transaction(contract_function=reward_function,
@@ -609,7 +609,7 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
             progress.update(1)
 
         # 4 - Initialize the StakingEscrow contract
-        init_function = the_escrow_contract.functions.initialize()
+        init_function = the_escrow_contract.functions.initialize(self.economics.erc20_reward_supply)
 
         init_receipt = self.blockchain.send_transaction(contract_function=init_function,
                                                         sender_address=self.deployer_address,
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -598,14 +598,14 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
         # Switch the contract for the wrapped one
         the_escrow_contract = wrapped_escrow_contract
 
-        # 3 - Transfer the reward supply tokens to StakingEscrow #
-        reward_function = self.token_contract.functions.approve(the_escrow_contract.address,
-                                                                self.economics.erc20_reward_supply)
+        # 3 - Approve transfer the reward supply tokens to StakingEscrow #
+        approve_reward_function = self.token_contract.functions.approve(the_escrow_contract.address,
+                                                                        self.economics.erc20_reward_supply)
 
         # TODO: Confirmations / Successful Transaction Indicator / Events ??  - #1193, #1194
-        reward_receipt = self.blockchain.send_transaction(contract_function=reward_function,
-                                                          sender_address=self.deployer_address,
-                                                          payload=origin_args)
+        approve_reward_receipt = self.blockchain.send_transaction(contract_function=approve_reward_function,
+                                                                  sender_address=self.deployer_address,
+                                                                  payload=origin_args)
         if progress:
             progress.update(1)
 
@@ -619,7 +619,7 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
             progress.update(1)
 
         # Gather the transaction receipts
-        ordered_receipts = (deploy_receipt, dispatcher_deploy_receipt, reward_receipt, init_receipt)
+        ordered_receipts = (deploy_receipt, dispatcher_deploy_receipt, approve_reward_receipt, init_receipt)
         deployment_receipts = dict(zip(self.deployment_steps, ordered_receipts))
 
         # Set the contract and transaction receipts #
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -483,7 +483,7 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
 
     agency = StakingEscrowAgent
     contract_name = agency.registry_contract_name
-    deployment_steps = ('contract_deployment', 'dispatcher_deployment', 'reward_transfer', 'initialize')
+    deployment_steps = ('contract_deployment', 'dispatcher_deployment', 'approve_reward_transfer', 'initialize')
     _proxy_deployer = DispatcherDeployer
 
     def __init__(self, test_mode: bool = False, *args, **kwargs):
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -509,7 +509,8 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
             "_minLockedPeriods": args[4],
             "_minAllowableLockedTokens": args[5],
             "_maxAllowableLockedTokens": args[6],
-            "_minWorkerPeriods": args[7]
+            "_minWorkerPeriods": args[7],
+            "_isTestContract": False
         }
         constructor_kwargs.update(overrides)
         constructor_kwargs = {k: v for k, v in constructor_kwargs.items() if v is not None}
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -486,13 +486,14 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
     deployment_steps = ('contract_deployment', 'dispatcher_deployment', 'reward_transfer', 'initialize')
     _proxy_deployer = DispatcherDeployer
 
-    def __init__(self,  *args, **kwargs):
+    def __init__(self, test_mode: bool = False, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.__dispatcher_contract = None
 
         token_contract_name = NucypherTokenDeployer.contract_name
         self.token_contract = self.blockchain.get_contract_by_name(registry=self.registry,
                                                                    contract_name=token_contract_name)
+        self.test_mode = test_mode
 
     def __check_policy_manager(self):
         result = self.contract.functions.policyManager().call()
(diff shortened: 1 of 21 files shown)


nucypher/blockchain/eth/registry.py
```diff
@@ -88,6 +88,7 @@ def __repr__(self) -> str:
     def id(self) -> str:
         """Returns a hexstr of the registry contents."""
         blake = hashlib.blake2b()
+        blake.update(self.__class__.__name__.encode())
```
michwill: Oh wow... Was it supposed to be in all the time?

vzotova: I think in production it's impossible to have two different registries classes with the same content, but in test it was all the time. 
Also interesting moment - locally I still have old behavior with ordering which "good" for tests, but one commit changed it for CircleCI

nucypher/blockchain/eth/deployers.py
```diff
@@ -467,7 +467,7 @@ def __check_policy_manager(self):
             raise RuntimeError("PolicyManager contract is not initialized.")
 
     def _deploy_essential(self, gas_limit: int = None):
-        escrow_constructor_args = (self.token_contract.address, *self.economics.staking_deployment_parameters)
+        escrow_constructor_args = (self.token_contract.address, *self.economics.staking_deployment_parameters, False)
```
KPrasch: Heh - this lonely `False` is the worklock test flag?

vzotova: Yep, and it look ugly, don't know how to do it better

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -2,148 +2,245 @@ pragma solidity ^0.5.3;
...
+    function ethToWork(uint256 _ethAmount) public view returns (uint256) {
+        return _ethAmount.mul(tokenSupply).mul(SLOWING_REFUND).divCeil(ethSupply.mul(boostingRefund));
+    }
+
+    /**
+    * @notice Calculate amount of ETH that will be refund for completing specified amount of work
+    * @dev This value will be fixed only after end of bidding
+    **/
+    function workToETH(uint256 _completedWork) public view returns (uint256) {
+        return _completedWork.mul(ethSupply).mul(boostingRefund).div(tokenSupply.mul(SLOWING_REFUND));
```
arjunhassard: Say for sake of argument that `SLOWING_REFUND` = `boostingRefund` – this would mean stakers must earn the exact same amount in inflationary rewards as their initial stake in order to reclaim all their locked ETH.

That sounds ok, except that the time it will take to get to reach this amount can vary enormously and, crucially, independently of the decisions of that staker. Let's assume an annual base inflation rate of 4% and a staker who locks their tokens to receive maximum rewards (i.e. duration >> 12 months), and restakes 100% of those rewards. This staker would face the following variance in the number of months ('#M') to reclaim the original ETH they locked in the Worklock, based on the participation rate ('PR'):
PR: 30% , #M: 7
PR: 50% , #M: 12
PR: 70% , #M: 20
PR: 90% , #M: 32
This isn't ideal, because the opportunity cost of locking up ETH for 1 year is very different to 2.5 years. The ETH price has dropped by 90% in 1 year before. 

If we want to give Worklock participants greater control/predictability over getting their ETH back, then this function should adjust based on the participation rate. In other words, if the participation rate goes up, then so should `boostingRefund`.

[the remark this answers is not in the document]
arjunhassard: Also, choosing a constant for `boostingRefund` ahead of mainnet (like what I assume is the placeholder value of _100_ for `SLOWING_REFUND`) suffers from the same unpredictability with respect to the participation rate – unless we plan to adjust this via some governance process every so often?

nucypher/blockchain/eth/deployers.py
KPrasch: Perhaps the value can be an Init parameter to the Deployer instance itself  (`test_mode`) - this flag is lacking any naming within the python API, which I think will contribute to it being forgotten / neglected.

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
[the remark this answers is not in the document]
vzotova: hmm, first what do you mean by participation rate?
external factor (for staker) for produced amount of work is how many tokens are staked in all network for current period. So if staker restakes + active + stake >= 12 month then only big changes in global stakes can increase time to unlock

nucypher/blockchain/eth/deployers.py
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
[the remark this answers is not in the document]
arjunhassard: participation rate ('PR') = tokens locked and therefore eligible for inflationary rewards, as % of all tokens

i suppose we can split this into two issues of uncertainty: 
- not knowing, at the point of transferring your ETH, when you will get it back – even if you work/stake maximally. for example, the Edgeware lockdrop allowed you to choose concrete unlock periods (3, 6, 12 months etc) that do not change. we also assume that duration certainty is necessary when staking/re-staking NU. 
- big changes to the participation rate that move your future reclaim date forward or back, _during_ your tenure as a worker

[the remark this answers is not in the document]
vzotova: about PR%: then more you have PR then faster you get ETH back, so % in first message confused me
What ideal way to calculate time/work do you see? What parameters should be used?

nucypher/blockchain/eth/deployers.py
KPrasch: OK, I'm in - but I do think we may need to control this parameter via CLI flags, and make a section about it in our deployment documentation so that it is not forgotten.

KPrasch:
Phew! This is going to make #1126 an adventure to bring up-to-speed. Good Work.

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -168,6 +171,16 @@ contract Issuer is Upgradeable {
         currentSupply2 = currentSupply2 - _amount;
     }
 
+    /**
+    * @notice Burn sender's tokens. Amount of tokens will be returned for future minting
+    * @param _value Amount to burn
+    */
+    function burn(uint256 _value) public {
```
cygnusv: Is it really "burn"? Perhaps a term like `giveBack` is more correct.

nucypher/blockchain/eth/deployers.py
```diff
@@ -452,21 +452,24 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
     deployment_steps = ('contract_deployment', 'dispatcher_deployment', 'reward_transfer', 'initialize')
```
cygnusv:
```suggestion
    deployment_steps = ('contract_deployment', 'dispatcher_deployment', 'approve_reward_transfer', 'initialize')
```

nucypher/blockchain/eth/deployers.py
```diff
@@ -543,19 +546,19 @@ def deploy(self,
         # Switch the contract for the wrapped one
         the_escrow_contract = wrapped_escrow_contract
 
-        # 3 - Transfer the reward supply tokens to StakingEscrow #
-        reward_function = self.token_contract.functions.transfer(the_escrow_contract.address,
-                                                                 self.economics.erc20_reward_supply)
+        # 3 - Approve transfer the reward supply tokens to StakingEscrow #
+        approve_reward_function = self.token_contract.functions.approve(the_escrow_contract.address,
+                                                                        self.economics.erc20_reward_supply)
```
cygnusv: I like this, specially considering the problem described in #1532. Just approving and not sending makes it even more secure and organized.

tests/blockchain/eth/contracts/main/worklock/test_worklock.py
```diff
@@ -25,6 +28,54 @@ def token(testerchain, token_economics, deploy_contract):
...
+               to_canonical_address(worklock.address) + \
+               bytes.fromhex('83') + \
+               nonce.to_bytes(3, byteorder='big')
+    else:
+        data = bytes.fromhex('da') + \
+               bytes.fromhex('94') + \
+               to_canonical_address(worklock.address) + \
+               bytes.fromhex('84') + \
+               nonce.to_bytes(4, byteorder='big')
+    return to_checksum_address(keccak(data)[12:32])
```
cygnusv: Wow, what's going on here?

nucypher/blockchain/eth/registry.py
cygnusv: Instead of the class name we should use the registry type:
```suggestion
        blake.update(self.REGISTRY_TYPE.encode())
```
We don't want the ID to be different depending if it's a `LocalContractRegistry`, a `TemporaryContractRegistry` or an `InMemoryContractRegistry`, even though all of them are contract registries.

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
vzotova: not necessary back, because you can "burn" even tokens that you have with airdrop

tests/blockchain/eth/contracts/main/worklock/test_worklock.py
vzotova: Calculating future contract address that will be deployed by worklock

nucypher/blockchain/eth/registry.py
vzotova: while tests we create two same registries Local and Memory and they have same ID, it leads to bugs depending on ordering in cache dictionary. I see this ID not for content but for instance

tests/blockchain/eth/contracts/main/worklock/test_worklock.py
cygnusv: Can you provide a comment with a link or description of this process?

vzotova: ✔️
I simplified calculation and added link to wiki
Also useful links:
* https://ethereum.stackexchange.com/questions/760/how-is-the-address-of-an-ethereum-contract-computed
* https://ethereum.stackexchange.com/questions/24248/how-to-calculate-an-ethereum-contracts-address-during-its-creation-using-the-so

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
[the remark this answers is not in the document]
arjunhassard: Possible approaches to this:
1) normalise the workToETH function with the participation rate, such that maximal staking (restaking + active + stake > 12m) always means ETH is refunded in 1 year. this approach is complicated by the fact the PR does not stay constant and each staker has a set of unique periods (sub-stakes) over which they received rewards
2) more simple: use a constant to enforce a time ceiling – e.g. everyone who works maximally should get their ETH back in ~1 year. So taking a worst case scenario for the average PR, we can choose a value for `boostingRefund` accordingly
for example, assuming the highest imaginable PR is 90% , then `boostingRefund` should be ~2
*this example assumes a base monthly inflation of 4% and that **all** stakers lock their tokens for >= 12 months and set their stakes to 'auto-prolong'

nucypher/blockchain/eth/registry.py
cygnusv: Well, but if this is a problem for tests, then we should monkey-patch it in tests then. I don't see impossible that in production we could have two registries with the same content but different classes (e.g., an InMemory registry persisted to a LocalRegistry)
@KPrasch what do you think?

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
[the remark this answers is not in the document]
vzotova: Opened #1540

nucypher/blockchain/eth/registry.py
cygnusv: See #1545

PR #1411, opened 2019-10-11
Adiós, dependency conflicts! 👋

cygnusv:
See https://github.com/pypa/pip/issues/4537#issuecomment-373386803

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -267,7 +267,9 @@ commands:
     steps:
       - run:
           name: Install Python Dependencies with Pip
-          command: pip3 install --user -e .
+          command: |
+            python setup.py install_egg_info --install-dir=/home/circleci/.local/lib/python3.7/site-packages
+            pip3 install --user -e .
       - check_nucypher_entrypoints   # Ensure Standard Installation Entry-points Work
       - run:
           name: Install Python Development Dependencies with Pip


michwill:
We want this in, like, yesterday!

cygnusv:
As far as I can tell this PR:
* Solves the problem caused by conflicts between cached dependencies and latest releases (which we were solving manually by relocking dependencies)
* Still respects cached dependencies that didn't cause a conflict

However, I don't fully understand what this is doing, so I prefer to have reviews of our pythonistas (@KPrasch , @michwill , @jMyles )

michwill:
Heh, I am already rebasing on this to have the CI working

jMyles:
So, with this change, *when* does a dependency's version get updated?

In other words, what does one have to do in order to change the version of a dependency?

cygnusv:
@jmyles Relock dependencies, as we have been doing for the last weeks, since we use pip instead of pipenv.

When? Whenever you feel it's right. Something I've been saying is using the nightly build to test the latest dependencies.

cygnusv:
More info on the problem. It seems it's still open and the workaround is the same.

https://github.com/pypa/pip/issues/6275

jMyles:
It's still not obvious to me what this is doing.  When using `install_egg_info`, from where is the list of dependencies being read?  The latest *released* version, no?  Does this mean that one must build a release in order to tick a dependency version?

I hope I'm misunderstanding this.

derekpierre:
@jMyles I'm still a bit fuzzy on what exactly is occurring.

However, based on some reading of code in `<python installation>/site-packages/setuptools/install_egg_info.py` and `<python installation>/site-packages/setuptools/egg_info.py` it seems that the dependencies used for `install_egg_info` are read from the `requirements.txt` file.

To be sure, I changed the version of a package in the `requirements.txt` file to an older package and then ran `python setup.py install_egg_info` locally. The produced `nucypher.egg-info/requires.txt` contained the modified version number for that package. Therefore, I think you should be able to advance the version of the dependency through the requirements file.

How the egg_info is subsequently used is outlined here - https://github.com/pypa/pip/issues/6275#issuecomment-466332992, but it's still not really clear to me - there seems to be some hierarchical check for requirement versions and `egg_info` perhaps gets precedent...(?)

Issue #1412, opened 2019-10-12
Characters try to use keyring_root property, but it is now hidden


michwill:
```
>>> alice = Alice(keyring=keyring, known_nodes=[ursula], provider_uri='~/.ethereum/goerli/geth.ipc')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.6/site-packages/nucypher/characters/lawful.py", line 124, in __init__
    *args, **kwargs)
  File "/usr/lib/python3.6/site-packages/nucypher/characters/base.py", line 131, in __init__
    keyring_root, checksum_address = keyring.keyring_root, keyring.checksum_address
AttributeError: 'NucypherKeyring' object has no attribute 'keyring_root'
```
The issue is because now `NucypherKeyring` has a property `__keyring_root` instead of `keyring_root`

KPrasch:
Closed and tested in #1413

PR #1413, opened 2019-10-12
Keyring update prevented character instantiation

michwill:
After moving `keyring_root` to a hidden property, a `@property` decorated method was not created. Characters were never tested with a keyring, so the bug (https://github.com/nucypher/nucypher/issues/1412) was unnoticed.

This PR fixes the issue and tests if Alice can be created with a keyring


michwill:
Heh, indeed: can add them to tests too

PR #1417, opened 2019-10-15
Improve CLI commands to better detect invalid option/commands and provide context-specific help

### LINKED ISSUE #1286: Unknown staking CLI commands do not crash properly
We need a way to more eagerly check that a CLI action / command is valid before attempting further operations, like loading configurations from the disk.  An informative error message and exit code will help too.

derekpierre:
Fixes #1286 .

In our current CLI setup we lay out all of the options for all commands even though some options only apply to specific "commands", and not to others. Furthermore, we can't do something like:
```
nucypher <character> <command> --help
```
and get specific help for that specific command.

One way for us to get around this could be to restructure our CLI commands a bit to instead use `click.group` and `click.command`. It would allow us to ensure that required options for a specific command/action are provided by specifying `required=True` for options tied to those specific commands/actions, or that options that aren't applicable to specific commands, can't be specified (a recurring example in our code is the `dev` option)

This would improve UX by providing more context-specific help, and inherently fixes the problem of detecting unrecognized commands.

The drawback of the above strategy is that common code/objects needed across different commands need to be refactored into common methods, but this may not be such a bad thing.

### DIFF
--- a/nucypher/cli/characters/alice.py
+++ b/nucypher/cli/characters/alice.py
@@ -270,10 +270,13 @@ def public_keys(click_config,
 
 
 @alice.command('derive-policy-pubkey')
-@_api_options
 @click.option('--label', help="The label for a policy", type=click.STRING, required=True)
+@_api_options
 @nucypher_click_config
 def derive_policy_pubkey(click_config,
+
+                         # Other (required)
+                         label,
 
                          # API Options
                          geth,
@@ -288,9 +291,7 @@ def derive_policy_pubkey(click_config,
                          hw_wallet,
                          teacher_uri,
                          min_stake,
-
-                         # Other
-                         label):
+                         ):
     """
     Get a policy public key from a policy label.
     """
@@ -308,7 +309,6 @@ def derive_policy_pubkey(click_config,
 
 
 @alice.command()
-@_api_options
 @click.option('--bob-encrypting-key', help="Bob's encrypting key as a hexadecimal string", type=click.STRING,
               required=True)
 @click.option('--bob-verifying-key', help="Bob's verifying key as a hexadecimal string", type=click.STRING,
@@ -318,8 +318,14 @@ def derive_policy_pubkey(click_config,
 @click.option('--n', help="N-Total KFrags", type=click.INT)
 @click.option('--expiration', help="Expiration Datetime of a policy", type=click.STRING)  # TODO: click.DateTime()
 @click.option('--value', help="Total policy value (in Wei)", type=types.WEI)
+@_api_options
 @nucypher_click_config
 def grant(click_config,
+          # Other (required)
+          bob_encrypting_key, bob_verifying_key, label,
+
+          # Other
+          m, n, expiration, value,
 
           # API Options
           geth,
@@ -334,9 +340,7 @@ def grant(click_config,
           hw_wallet,
           teacher_uri,
           min_stake,
-
-          # Other
-          bob_encrypting_key, bob_verifying_key, label, m, n, expiration, value):
+          ):
     """
     Create and enact an access policy for some Bob.
     """
@@ -365,12 +369,15 @@ def grant(click_config,
 
 
 @alice.command()
-@_api_options
 @click.option('--bob-verifying-key', help="Bob's verifying key as a hexadecimal string", type=click.STRING,
               required=True)
 @click.option('--label', help="The label for a policy", type=click.STRING, required=True)
+@_api_options
 @nucypher_click_config
 def revoke(click_config,
+
+           # Other (required)
+           bob_verifying_key, label,
 
            # API Options
            geth,
@@ -385,9 +392,7 @@ def revoke(click_config,
            hw_wallet,
            teacher_uri,
            min_stake,
-
-           # Other
-           bob_verifying_key, label):
+           ):
     """
     Revoke a policy.
     """
@@ -406,12 +411,15 @@ def revoke(click_config,
 
 
 @alice.command()
-@_api_options
 @click.option('--label', help="The label for a policy", type=click.STRING, required=True)
 @click.option('--message-kit', help="The message kit unicode string encoded in base64", type=click.STRING,
               required=True)
+@_api_options
 @nucypher_click_config
 def decrypt(click_config,
+
+            # Other (required)
+            label, message_kit,
 
             # API Options
             geth,
@@ -426,9 +434,7 @@ def decrypt(click_config,
             hw_wallet,
             teacher_uri,
             min_stake,
-
-            # Other
-            label, message_kit):
+            ):
     """
     Decrypt data encrypted under an Alice's policy public key.
     """
--- a/nucypher/cli/characters/alice.py
+++ b/nucypher/cli/characters/alice.py
@@ -46,7 +46,7 @@ def _api_options(func):
 @click.group()
 def alice():
     """
-    Alice the Policy Authority" management commands.
+    "Alice the Policy Authority" management commands.
     """
     pass
 
@@ -208,7 +208,7 @@ def run(click_config,
                                                    geth, network, pay_with, provider_uri, registry_filepath)
     #############
 
-    ALICE = _create_alice(alice_config, click_config, dev, emitter, hw_wallet, min_stake, teacher_uri)
+    ALICE = _create_alice(alice_config, click_config, dev, emitter, hw_wallet, teacher_uri, min_stake)
 
     try:
         # RPC
@@ -263,7 +263,7 @@ def public_keys(click_config,
                                                    geth, network, pay_with, provider_uri, registry_filepath)
     #############
 
-    ALICE = _create_alice(alice_config, click_config, dev, emitter, hw_wallet, min_stake, teacher_uri)
+    ALICE = _create_alice(alice_config, click_config, dev, emitter, hw_wallet, teacher_uri, min_stake)
 
     response = ALICE.controller.public_keys()
     return response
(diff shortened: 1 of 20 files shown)


nucypher/cli/characters/alice.py
```diff
@@ -11,171 +12,411 @@
...
+    def wrapper(*args, **kwargs):
+        return func(*args, **kwargs)
+    return wrapper
+
+
+# Args (geth, provider_uri, federated_only, dev, pay_with, network, registry_filepath, config_file, discovery_port,
+#       hw_wallet, teacher_uri, min_stake)
+def _api_options(func):
+    @_admin_options
+    @click.option('--config-file', help="Path to configuration file", type=EXISTING_READABLE_FILE)
```
fjarri: May be a good idea to make repeating options like this one a variable and then use it everywhere it's needed, i.e.

    config_file_option = click.option('--config-file', help="Path to configuration file", type=EXISTING_READABLE_FILE)

and

    @_admin_options
    @config_file_option
    ...etc

nucypher/cli/characters/alice.py
```diff
@@ -11,171 +12,411 @@
...
+@alice.command()
+@_api_options
+@click.option('--controller-port', help="The host port to run Alice HTTP services on", type=NETWORK_PORT,
+              default=AliceConfiguration.DEFAULT_CONTROLLER_PORT)
+@click.option('--dry-run', '-x', help="Execute normally without actually starting the node", is_flag=True)
+@nucypher_click_config
+def run(click_config,
+
+        # API Options
+        geth, provider_uri, federated_only, dev, pay_with, network, registry_filepath,
```
fjarri: These repeating argument lists kind of bother me. Do you think it would be better to gather them into structures e.g.

```
def _admin_options2(func):
    @click.option('--geth', '-G', help="Run using the built-in geth node", is_flag=True)
    @click.option('--provider', 'provider_uri', help="Blockchain provider's URI", type=click.STRING)
    @click.option('--federated-only', '-F', help="Connect only to federated nodes", is_flag=True)
    @click.option('--dev', '-d', help="Enable development mode", is_flag=True)
    @click.option('--pay-with', help="Run with a specified account", type=EIP55_CHECKSUM_ADDRESS)
    @click.option('--network', help="Network Domain Name", type=click.STRING)
    @click.option('--registry-filepath', help="Custom contract registry filepath", type=EXISTING_READABLE_FILE)
    @functools.wraps(func)
    def wrapper(
            geth,
            provider_uri,
            federated_only,
            dev,
            pay_with,
            network,
            registry_filepath,
            **kwargs):

        opts = namedtuple(
            'AdminOptions',
            'geth',
            'provider_uri',
            'federated_only',
            'dev',
            'pay_with',
            'network',
            'registry_filepath',
            )

        return func(admin_options=opts, **kwargs)

    return wrapper


def _api_options2(func):
    @_admin_options2
    @click.option('--config-file', help="Path to configuration file", type=EXISTING_READABLE_FILE)
    @click.option('--discovery-port', help="The host port to run node discovery services on", type=NETWORK_PORT)
    @click.option('--hw-wallet/--no-hw-wallet', default=False)
    @click.option('--teacher', 'teacher_uri', help="An Ursula URI to start learning from (seednode)", type=click.STRING)
    @click.option('--min-stake', help="The minimum stake the teacher must have to be a teacher", type=click.INT,
                  default=0)
    @functools.wraps(func)
    def wrapper(
            admin_options,
            config_file,
            discovery_port,
            hw_wallet,
            teacher,
            min_stake,
            **kwargs):

        opts = namedtuple(
            'ApiOptions',
            'admin_options',
            'discovery_port',
            'hw_wallet',
            'teacher',
            'min_stake',
            )

        return func(api_options=opts, **kwargs)

    return wrapper
```

and then

```
@alice.command()
@_admin_options2
@click.option('--config-root', help="Custom configuration directory", type=click.Path())
@click.option('--poa', help="Inject POA middleware", is_flag=True, default=None)
@click.option('--m', help="M-Threshold KFrags", type=click.INT)
@click.option('--n', help="N-Total KFrags", type=click.INT)
@click.option('--rate', help="Policy rate per period in wei", type=click.FLOAT)
@click.option('--duration-periods', help="Policy duration in periods", type=click.FLOAT)
@nucypher_click_config
def init(click_config,
         admin_options,
         # Other
         config_root, poa, m, n, rate, duration_periods):
```

Or is it too much decorator magic?

nucypher/cli/characters/enrico.py
```diff
@@ -7,73 +9,81 @@
 from nucypher.cli.types import NETWORK_PORT
 
 
-@click.command()
-@click.argument('action')
+# Args (policy_encrypting_key)
+def _common_options(func):
```
fjarri: If there's no more descriptive name than "common options", it may be better to just define 
```
policy_encrypting_key_option = click.option('--policy-encrypting-key', help="Encrypting Public Key for Policy as hexadecimal string", type=click.STRING, required=True)
```

and then using it as decorator.

nucypher/cli/characters/enrico.py
```diff
@@ -7,73 +9,81 @@
...
+    # Request
+    encryption_request = {'message': message}
+    response = ENRICO.controller.encrypt_message(request=encryption_request)
+    return response
 
-    if not policy_encrypting_key:
-        raise click.BadArgumentUsage('--policy-encrypting-key is required to start Enrico.')
 
-    # Banner
+def _create_enrico(click_config, policy_encrypting_key):
```
fjarri: I think it is better separated into `_setup_emitter()` and `_create_enrico()` proper, similarly to other characters.

tests/cli/ursula/test_run_ursula.py
```diff
@@ -172,14 +172,15 @@ def test_persistent_node_storage_integration(click_runner,
     assert result.exit_code == 0
 
 
+# TODO: test can probably be removed (--dev is not a valid option for `init`)
```
fjarri: I agree, I think it can be safely removed even in this PR

nucypher/cli/characters/bob.py
```diff
@@ -11,106 +12,268 @@
...
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        return func(*args, **kwargs)
+    return wrapper
+
+
+@click.group()
+def bob():
+    """
+    "Bob" management commands.
```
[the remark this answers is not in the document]
cygnusv: Oh @jMyles is not going to like this suggestion...

cygnusv:
> The drawback of the above strategy is that common code/objects needed across different commands need to be refactored into common methods, but this may not be such a bad thing.

It's not that bad. In fact, after having spent a lot of time with `stake.py`, `alice.py` and `nucypher-deploy.py`, I think it's the opposite. This allow us to customize the setups for each command as needed, while in the previous approach it was really messy.

nucypher/cli/characters/alice.py
cygnusv: I like this suggestion.

vepkenez:
I agree that this isn’t perfect but let’s make issues for cleaning up the repeated options and things like that so we can merge this.  We have a lot of new users now and this PR makes Nucypher so much more useable.

cygnusv:
> I agree that this isn’t perfect but let’s make issues for cleaning up the repeated options and things like that so we can merge this. We have a lot of new users now and this PR makes Nucypher so much more useable.

I'd simply address the really minor comments (e.g., fixes to command help texts) and perhaps open an issue for the others.

nucypher/cli/characters/alice.py
derekpierre: 👍  - Filed #1431 .

nucypher/cli/characters/alice.py
derekpierre: 👍  - It would be cleaner and prevent copy and pasting - Filed #1431 .

tests/cli/ursula/test_run_ursula.py
derekpierre: ✔️

nucypher/cli/characters/enrico.py
derekpierre: ✔️

nucypher/cli/characters/enrico.py
derekpierre: ✔️

nucypher/cli/characters/enrico.py
```diff
@@ -7,73 +9,81 @@
...
+    @click.option('--policy-encrypting-key', help="Encrypting Public Key for Policy as hexadecimal string", type=click.STRING, required=True)
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
+        return func(*args, **kwargs)
+    return wrapper
+
+@click.group()
+def enrico():
+    """
+    Enrico the Encryptor" management commands.
```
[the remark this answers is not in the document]
derekpierre: ✔️  - Included in subsequent RFC commit.

nucypher/cli/characters/bob.py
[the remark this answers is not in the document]
derekpierre: ✔️  - Included in subsequent RFC commit.

derekpierre:
> We've got rid of `--sync`, right?

Yep! 🙂

cygnusv:
> > We've got rid of `--sync`, right?
> 
> Yep! 🙂

🤔 That's something we should study: I noticed the other day that the `--sync` flag was unused in most commands (for current `master`). Is it because it's not a supported functionality anymore in blockchain connections?

PR #1425, opened 2019-10-19
Hardcoding a felix NU transfer gas limit.

KPrasch:


KPrasch:
I dont know if this is a fix yet - putting it here for now.

michwill:
Ah no, it must be a transaction count. Well, in such case, we have to have nonce=None, and get it from blockchain when it's None, otherwise increase by 1 every next transaction.

michwill:
Ok, I think that the problem is not gas usage, it's probably broadcasting several transactions with the same nonce (because transaction counter in geth accounts only for transactions which are already on blockchain?).
Here's an attempt to fix it, and, I think, we can try to test it on Felix
https://github.com/nucypher/nucypher/pull/1427

michwill:
Simpler fix here https://github.com/nucypher/nucypher/pull/1430

PR #1427, opened 2019-10-19
[WIP] Use counters for nonces

michwill:
When there are many transactions submitted into one block, current approach we use fails. Let's fix that


michwill:
Note that I didn't do any locks for the case of multi-threading. Without them, some edge cases are still possible.
Also, I think, the approached should be tested with Felix first

michwill:
Actually, moving that functionality to the client rather than interface

michwill:
Ok, there is a simpler way

PR #1430, opened 2019-10-20
Include pending transactions when calculating a nonce

michwill:


michwill:
(Felix is currently doing that - will see what happens)

michwill:
@cygnusv this one addresses the nonce calculation. I suppose, for getting the balance it's not as important.
I would consider locking between getting the nonce and broadcasting the transaction as a further improvement though

PR #1434, opened 2019-10-22
CLI options cleanup

### LINKED ISSUE #1431: Cleanup repeated options and repeated arguments for cli commands
Refactor repeated options and arguments for commands introduced in PR #1417 . 

See comments https://github.com/nucypher/nucypher/pull/1417#discussion_r336756976 and https://github.com/nucypher/nucypher/pull/1417#discussion_r336757499 for potential solutions to explore.

### LINKED ISSUE #1503: Sync progress bar stdout being swallowed into a black hole

### LINKED ISSUE #1556: Logging is not configurable for nucypher-deploy
Currently, logging for `nucypher-deploy` is not active. Plus, by default, it's set on `info` level, but for deployment it's useful to have it as `debug`.

fjarri:
A continuation of PR #1417, fixes #1431

- Moved repeating CLI options to a separate file
- Collected groups of options into objects to avoid long parameter lists
- Removed `etherscan` option from `config.py` - it is not used anywhere
- Moved `mock_networking` option from `config.py` to the characters that use it
- Got rid of `NucypherClickConfig` import in `CharacterConfiguration` constructor; fixes #1503 
- Removed assignments to `NucypherClickConfig` attributes which did nothing, and replaced these assignments with `GlobalLoggerSettings` method calls in `conftest.py`. **Note:** I'm assuming here that `NucypherClickConfig` was only used in CLI calls, which were all started with `click.invoke()` and therefore setting `NucypherClickConfig` attributes in the main process had no effect.
- Used `NucypherClickConfig` (now `GroupGeneralConfig`) in CLI `deploy` command, instead of creating an emitter manually; fixes #1556
- Removed caching from `get_json_file_observer()` and `get_text_file_observer()` in `logging.py`. These functions are called very rarely, and have side effects, so caching is not needed and can actually lead to failures there.
- **CHANGED:** `NUCYPHER_CONFIG_FILE` envvar renamed to `NUCYPHER_CONFIG_ROOT` (according to how it is used in the code)

Minor:
- Removed an unused method `GlobalLoggerSettings.pause_console_logging_while()`
- Removed an unneeded repetition of the command name in `stake.py::preallocation()`

Notes:
- Currently, some of the flag-type options (`federated_only` and `poa`) default to `None` because they can be loaded from the config file. I added the corresponding comments in this PR, but I would prefer a more explicit way to handle it. I already spent some time debugging test fails because I accidentally omitted `default=None` for `federated_only`.

### DIFF
--- a/nucypher/characters/control/emitters.py
+++ b/nucypher/characters/control/emitters.py
@@ -1,3 +1,4 @@
+import io
 import json
 import sys
 from functools import partial
@@ -8,6 +9,10 @@ from flask import Response
 from twisted.logger import Logger
 
 import nucypher
+
+
+def null_stream():
+    return open(os.devnull, 'w')
 
 
 class StdoutEmitter:
@@ -63,6 +68,12 @@ class StdoutEmitter:
             e_str = str(e)
             click.echo(message=e_str)
             self.log.info(e_str)
+
+    def get_stream(self, verbosity: int = 0):
+        if verbosity <= self.verbosity:
+            return click.get_text_stream('stdout')
+        else:
+            return null_stream()
 
 
 class JSONRPCStdoutEmitter(StdoutEmitter):
@@ -179,6 +190,9 @@ class JSONRPCStdoutEmitter(StdoutEmitter):
         # self.log.info(f"Error {e.code} | {e.message}")  # TODO: Restore this log message
         return size
 
+    def get_stream(self, *args, **kwargs):
+        return null_stream()
+
 
 class WebEmitter:
 
@@ -227,3 +241,6 @@ class WebEmitter:
         # ---------- HTTP OUTPUT
         response = drone_character.sink(response=serialized_response, status=200, content_type="application/javascript")
         return response
+
+    def get_stream(self, *args, **kwargs):
+        return null_stream()
--- a/nucypher/cli/characters/alice.py
+++ b/nucypher/cli/characters/alice.py
@@ -477,7 +477,7 @@ def _create_alice(alice_config, general_config, dev, emitter, hw_wallet, teacher
             client_password = get_client_password(checksum_address=alice_config.checksum_address)
     try:
         ALICE = actions.make_cli_character(character_config=alice_config,
-                                           general_config=general_config,
+                                           emitter=general_config.emitter,
                                            unlock_keyring=not dev,
                                            teacher_uri=teacher_uri,
                                            min_stake=min_stake,
--- a/nucypher/cli/characters/alice.py
+++ b/nucypher/cli/characters/alice.py
@@ -41,24 +41,158 @@ from nucypher.utilities.sandbox.constants import TEMPORARY_DOMAIN
 option_bob_verifying_key = click.option(
     '--bob-verifying-key', help="Bob's verifying key as a hexadecimal string", type=click.STRING,
     required=True)
+option_pay_with = click.option(
+    '--pay-with', help="Run with a specified account", type=EIP55_CHECKSUM_ADDRESS)
 
 
-group_admin = group_options(
-    'admin',
-    geth=option_geth,
-    provider_uri=option_provider_uri(),
-    federated_only=option_federated_only,
+class AliceConfigOptions:
+
+    __option_name__ = 'config_options'
+
+    def __init__(
+            self, dev, network, provider_uri, geth, federated_only, discovery_port,
+            pay_with, registry_filepath):
+
+        if federated_only and geth:
+            raise click.BadOptionUsage(
+                option_name="--geth",
+                message="--federated-only cannot be used with the --geth flag")
+
+        # Managed Ethereum Client
+        eth_node = NO_BLOCKCHAIN_CONNECTION
+        if geth:
+            eth_node = actions.get_provider_process()
+            provider_uri = ETH_NODE.provider_uri(scheme='file')
+
+        self.dev = dev
+        self.domains = {network} if network else None
+        self.provider_uri = provider_uri
+        self.geth = geth
+        self.federated_only = federated_only
+        self.eth_node = eth_node
+        self.pay_with = pay_with
+        self.discovery_port = discovery_port
+        self.registry_filepath = registry_filepath
+
+    def create_config(self, middleware, config_file):
+
+        if self.dev:
+
+            # Can be None as well, meaning it is unset - no error in this case
+            if self.federated_only is False:
+                raise click.BadOptionUsage(
+                    option_name="--federated-only",
+                    message="--federated-only cannot be explicitly set to False when --dev is set")
+
+            return AliceConfiguration(
+                dev_mode=True,
+                network_middleware=middleware,
+                domains={TEMPORARY_DOMAIN},
+                provider_process=self.eth_node,
+                provider_uri=self.provider_uri,
+                federated_only=True)
+
+        else:
+            try:
+                return AliceConfiguration.from_configuration_file(
+                    dev_mode=False,
+                    network_middleware=middleware,
+                    domains=self.domains,
+                    provider_process=self.eth_node,
+                    provider_uri=self.provider_uri,
+                    filepath=config_file,
+                    rest_port=self.discovery_port,
+                    checksum_address=self.pay_with,
+                    registry_filepath=self.registry_filepath)
+            except FileNotFoundError:
+                return actions.handle_missing_configuration_file(
+                    character_config_class=AliceConfiguration,
+                    config_file=config_file)
+
+    def generate_config(self, emitter, config_root, poa, light, m, n, duration_periods, rate):
+
+        if self.dev:
+            raise click.BadArgumentUsage("Cannot create a persistent development character")
+
+        if not self.provider_uri and not self.federated_only:
+            raise click.BadOptionUsage(
+                option_name='--provider',
+                message="--provider is required to create a new decentralized alice.")
+
+        pay_with = self.pay_with
+        if not pay_with and not self.federated_only:
+            pay_with = select_client_account(
+                emitter=emitter,
+                provider_uri=self.provider_uri)
+
+        return AliceConfiguration.generate(
+            password=get_nucypher_password(confirm=True),
+            config_root=config_root,
+            checksum_address=pay_with,
+            domains=self.domains,
+            federated_only=self.federated_only,
+            provider_uri=self.provider_uri,
+            provider_process=self.eth_node,
+            registry_filepath=self.registry_filepath,
+            poa=poa,
+            light=light,
+            m=m,
+            n=n,
+            duration_periods=duration_periods,
+            rate=rate)
+
+
+group_config_options = group_options(
+    AliceConfigOptions,
     dev=option_dev,
-    pay_with=click.option('--pay-with', help="Run with a specified account", type=EIP55_CHECKSUM_ADDRESS),
     network=option_network,
-    registry_filepath=option_registry_filepath)
-
-
-group_api = group_options(
-    'api',
-    admin=group_admin,
-    config_file=option_config_file,
+    provider_uri=option_provider_uri(),
+    geth=option_geth,
+    federated_only=option_federated_only,
     discovery_port=option_discovery_port(),
+    pay_with=option_pay_with,
+    registry_filepath=option_registry_filepath,
+    )
+
+
+class AliceCharacterOptions:
+
+    __option_name__ = 'character_options'
+
+    def __init__(self, config_options, hw_wallet, teacher_uri, min_stake):
+        self.config_options = config_options
+        self.hw_wallet = hw_wallet
+        self.teacher_uri = teacher_uri
+        self.min_stake = min_stake
+
+    def create_character(self, emitter, config_file, middleware, json_ipc, load_seednodes=True):
+
+        config = self.config_options.create_config(middleware, config_file)
+
+        client_password = None
+        if not config.federated_only:
+            if (not self.hw_wallet or not config.dev_mode) and not json_ipc:
+                client_password = get_client_password(checksum_address=config.checksum_address)
+
+        try:
+            ALICE = actions.make_cli_character(character_config=config,
+                                               emitter=emitter,
+                                               unlock_keyring=not config.dev_mode,
+                                               teacher_uri=self.teacher_uri,
+                                               min_stake=self.min_stake,
+                                               client_password=client_password,
+                                               load_preferred_teachers=load_seednodes,
+                                               start_learning_now=load_seednodes)
+
+            return ALICE
+        except NucypherKeyring.AuthenticationFailed as e:
+            emitter.echo(str(e), color='red', bold=True)
+            click.get_current_context().exit(1)
+
+
+group_character_options = group_options(
+    AliceCharacterOptions,
+    config_options=group_config_options,
     hw_wallet=option_hw_wallet,
     teacher_uri=option_teacher_uri,
     min_stake=option_min_stake,
@@ -74,7 +208,7 @@ def alice():
 
 
 @alice.command()
-@group_admin
+@group_config_options
 @option_config_root
 @option_poa
 @option_light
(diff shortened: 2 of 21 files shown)


derekpierre:
@fjarri Replying to your Discord comments here concerning `test_character_command_options_and_parameters_match`

```
@derek , as for your points: 1. I'm not sure if it's necessary to explicitly check that click_config is present since we use it to create an emitter anyway. 2. If the option names and parameters do not match, we will get a click error when we call the function (and the ordering does not matter, since these parameters are used as keywords by click). Also, why does it matter if click_config is the bottom decorator or not? It does not interfere with other options.
Speaking of click_config, it is indeed the same option grouping, and group_options can accommodate it - if it takes an existing class instead of creating a nametuple on the fly.
As for moe, perhaps we should actually test its functionality instead of testing for a specific way we choose to write code?
Anyway, perhaps the discussion is better moved to the PR comments
(taking an existing class in group_options will also help with your concerns about opaqueness of grouped options - this way the structures can actually be documented)
```

Remember my goal here was that the test is just a sanity check:
1. An emitter for new cli commands can easily be created without that `click_config` option using `emitter = StdoutEmitter()` eg. in `deploy.py`. I'm not sure why that was, but I left the functionality the same as before my refactor.
2. I only mentioned the ordering for `click_config` because as part of the test I use an ordered `set` and ordering matters for the comparison. `click_config` was easier to replace with the full list of options because it was at the end. If you kept the test but decided to replace other named options not at the end of the set, then it would require more intricate code to maintain ordering.
3. Moe was just an example, I wasn't saying don't write tests, I'm just pointing out that assuming that cli commands are tested may not always be correct - some commands are difficult to test or get skipped for some reason (maybe temporarily), or only run on CI (might as well catch things early)
4. There was a case where I messed up an option/parameter setup and my cli character tests failed but it wasn't immediately clear why it failed - unfortunately I don't remember the exact scenario; some parameter used the wrong option value or something like that - which is originally why I wrote the test.

Facilitating a common layout of commands allows for these refactors to be easier, but if the test is no longer feasible based on your changes across the cli commands, then you can either modify or remove it.

nucypher/cli/characters/stake.py
```diff
@@ -101,133 +190,84 @@ def stake():
...
 
-    client_account, staking_address = handle_client_account_for_staking(emitter=emitter,
-                                                                        stakeholder=STAKEHOLDER,
-                                                                        staking_address=staking_address,
-                                                                        individual_allocation=STAKEHOLDER.individual_allocation,
-                                                                        force=force)
+    client_account, staking_address = handle_client_account_for_staking(
+        emitter=emitter,
+        stakeholder=STAKEHOLDER,
+        staking_address=character2_options.character1_options.staking_address,
```
cygnusv: I don't understand what's going on here? What do `Character1` and `Character2` mean?

fjarri: As I mentioned in the general description, the names are subject to change. Basically, the data flow in `stake.py` implies that there are two option groups used to create a character, one (the smaller one) used in `list` in `accounts`, and the other (the bigger one) in the rest of the commands. I suspect there may be some actual terms that can be used to denote them.

KPrasch: ...So this is an "option group"?

fjarri: Yep. The option group thing is essentially unchanged from the Derek's PR, but they're passed as a single object to the command functions instead of all the constituents separately. The common pattern seems to be [[options used to create/find a config] options used to create the character], but in `stake.py` there are two scenarios of character creation.

KPrasch: Do the names "Configuration Options" and "Character Options" fit?

nucypher/cli/characters/stake.py
```diff
@@ -34,63 +34,155 @@
...
+
+class StakeHolderBasicCharacterOptions:
+
+    __option_name__ = 'basic_character_options'
+
+    def __init__(self, config_options, staking_address):
+        self.config_options = config_options
+        self.staking_address = staking_address
+
+    def create_object(self, emitter, config_file, beneficiary_address=None, allocation_filepath=None):
```
derekpierre: So the difference between `StakeHolderBasicCharacterOptions` and `StakeHolderCharacterOptions` is the stuff needed for a preallocation staker.

Unfortunately, the `beneficiary_address` and `allocation_filepath` still leak into `StakeHolderBasicCharacterOptions`, albeit as `None` parameters but still feels a little weird since most of the `create_object` function is made up of logic for preallocation staker. Without the preallocation staker functionality, this method seems to boil down to the following:

```python
def create_object(self, emitter, config_file):
    stakeholder_config = self.config_options.create_config(emitter, config_file)
    return stakeholder_config.produce(
            initial_address=self.staking_address) 
```
^ *(I think the `individual_allocation` parameter for `produce` defaults to `None` so not needed to be specified)*

Perhaps it makes sense to just have two separate `create_object` methods for the two types of Option classes `StakeHolderBasicCharacterOptions` and `StakeHolderCharacterOptions`.

Also, would it make sense to rename the old`StakeHolderBasicCharacterOptions` to `StakeHolderCharacterOptions` and then rename the old `StakeHolderCharacterOptions` to `PreallocationStakeHolderCharacterOptions` (or something similar)?

@cygnusv wdyt? You would know about the nuance of this code better than I would.

nucypher/cli/characters/stake.py
```diff
@@ -678,56 +626,40 @@ def collect_reward(click_config,
 
 
 @stake.command('preallocation')
```
derekpierre: This was a previous change done by @cygnusv , but this line is not needed since the function name is the same as the command.

tests/cli/ursula/test_stake_via_allocation_contract.py
```diff
@@ -535,7 +535,6 @@ def test_collect_rewards_integration(click_runner,
 
     # Collect Policy Reward
     collection_args = ('stake', 'collect-reward',
-                       '--mock-networking',
```
derekpierre: I'm a little confused why this option is no longer needed, could you clarify the solution? 

Also, it was previously available to all cli commands as part of the `general_config`, but now it is specified as a `middleware` config option for specific characters - was it that it was not as general as we previously thought?

nucypher/characters/control/emitters.py
```diff
@@ -64,6 +65,12 @@ def error(self, e):
             click.echo(message=e_str)
             self.log.info(e_str)
 
+    def get_stream(self, verbosity: int = 0):
+        if verbosity <= self.verbosity:
+            return click.get_text_stream('stdout')
+        else:
+            return io.StringIO() # TODO: is there a null stream?
```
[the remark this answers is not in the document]
fjarri: Thanks, the `open()` one seems like a good solution - returns the same type as `get_text_stream()`

nucypher/cli/characters/alice.py
```diff
@@ -7,44 +7,172 @@
...
+
+    __option_name__ = 'character_options'
+
+    def __init__(self, config_options, hw_wallet, teacher_uri, min_stake):
+        self.config_options = config_options
+        self.hw_wallet = hw_wallet
+        self.teacher_uri = teacher_uri
+        self.min_stake = min_stake
+
+    def create_object(self, emitter, config_file, json_ipc, load_seednodes=True):
```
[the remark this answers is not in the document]
fjarri: I agree, it was a temporary choice that I forgot to change. Perhaps even just `create()` (and `generate()` for config groups) will be enough?

nucypher/cli/characters/bob.py
```diff
@@ -128,129 +167,77 @@ def run(click_config,
...
+def view(general_config, config_file):
     """
     View existing Bob's configuration.
     """
-    emitter = _setup_emitter(click_config)
-    bob_config = _get_bob_config(click_config, dev, provider_uri, network, registry_filepath, checksum_address,
-                                 config_file, discovery_port)
-    #############
+    emitter = _setup_emitter(general_config)
     filepath = config_file or bob_config.config_file_location
```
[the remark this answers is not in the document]
fjarri: Uh oh, good catch. I'm relying on tests too much. It was one of the parts that changed before the last rebase, and I wasn't attentive enough.

nucypher/cli/characters/stake.py
```diff
@@ -102,149 +194,84 @@ def stake():
...
-                     provider_uri,
-
-                     # Other
-                     config_root, force,
-
-                     # Admin Options
-                     poa, light, registry_filepath):
+@option_config_root
+@option_force
+@group_config_options
```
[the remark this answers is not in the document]
fjarri: There's an `assert` in `StakeHolderConfigOptions.generate_config()` to check that, but I agree, it's not ideal. The abstraction is leaking here... I may need to move it out of the option group completely.

nucypher/cli/characters/stake.py
fjarri: Thanks, I'll remove it.

tests/cli/ursula/test_stake_via_allocation_contract.py
fjarri: As far as I understand, the `stake collect-reward` command does not use this option. Of all the characters, only Alice, Bob and Felix use it. Plus, my intention was to only keep output and logging-related options in general config.

nucypher/cli/characters/stake.py
fjarri: Yes, I wasn't sure about the names because I don't understand the area well. `PreallocationStakeHolderCharacterOptions` sounds good to me.

> but still feels a little weird since most of the create_object function is made up of logic for preallocation staker.

Yes, that's a good point. I'll try to rearrange the logic there.

nucypher/characters/control/emitters.py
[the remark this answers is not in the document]
fjarri: Using `open()` in the new revision

nucypher/cli/characters/alice.py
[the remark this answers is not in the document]
fjarri: I went with `create_character()`/`create_actor()` in the new revision.

nucypher/cli/characters/bob.py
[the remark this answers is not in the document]
fjarri: Fixed in the new revision.

nucypher/cli/characters/stake.py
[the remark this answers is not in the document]
fjarri: Replaced `assert` with `click.BadOptionUsage()` in the new revision.

nucypher/cli/characters/stake.py
fjarri: Changed in the new revision, both the names and the logic placement.

nucypher/cli/characters/stake.py
fjarri: Removed in the new revision.

nucypher/cli/deploy.py
```diff
@@ -30,42 +30,147 @@
...
+        self.registry_outfile = registry_outfile
+        self.hw_wallet = hw_wallet
+        self.dev = dev
+        self.force = force
+        self.config_root = config_root
+        self.etherscan = etherscan
+        self.poa = poa
+        self.se_test_mode = se_test_mode
+
+    def create_actor(self, emitter):
```
derekpierre: 👍

tests/cli/ursula/test_stake_via_allocation_contract.py
derekpierre: @cygnusv would know more about that test.

Many of the characters use the `MockRestMiddleware()` setup for non-CLI tests, so it's possible that the option is useful for CLI tests as well but just not being utilized based on the current tests written. In any case, it's probably fine for now 👍 , and can be moved later if deemed necessary.

nucypher/cli/characters/stake.py
```diff
@@ -335,43 +350,29 @@ def detach_worker(click_config,
...
-@nucypher_click_config
-def create(click_config,
-
-           # Stake Options
-           poa, light, registry_filepath, config_file, provider_uri, staking_address, hw_wallet,
-           beneficiary_address, allocation_filepath,
-
-           # Other
-           force, value, lock_periods):
+@group_prealloc_character_options
```
cygnusv: I think this group of options should be named something like `@group_staker_options`, since it works for both preallocation and normal stakers.

nucypher/cli/characters/stake.py
```diff
@@ -34,63 +34,164 @@
...
+group_config_options = group_options(
+    StakeHolderConfigOptions,
+    provider_uri=option_provider_uri(),
+    poa=option_poa,
+    light=option_light,
+    registry_filepath=option_registry_filepath,
+    )
+
+
+class StakeHolderCharacterOptions:
```
cygnusv: I think a renaming is in order here. `Stakeholder` is no longer a character, nor it's actually the relevant actor here (in other words, Stakeholder is something basically internal). 
I'd name this `StakerOptions`, simply, although since we have to differentiate here with the other one (currently named `PreallocationStakeHolderCharacterOptions`), and the main difference between them is the possibility of being an "authenticated" staker (i.e., being able to transact) or not (i.e., just to use `list` or `accounts`), maybe the naming should reflect this.
For instance, this group can be `StakerOptions`, while the other can be `TransactingStakerOptions`. 
@KPrasch what do you think?

nucypher/cli/characters/stake.py
```diff
@@ -34,63 +34,164 @@
...
+
+
+group_character_options = group_options(
+    StakeHolderCharacterOptions,
+    config_options=group_config_options,
+    staking_address=option_staking_address,
+    )
+
+
+class PreallocationStakeHolderCharacterOptions:
```
cygnusv: I think this group needs a renaming. See my comment above.

nucypher/cli/common_options.py
```diff
@@ -0,0 +1,194 @@
...
+
+# Defaults to None since it can be overridden by the config file
+option_federated_only = click.option(
+    '--federated-only', '-F', help="Connect only to federated nodes", is_flag=True, default=None)
+
+option_force = click.option('--force', help="Don't ask for confirmation", is_flag=True)
+
+option_geth = click.option('--geth', '-G', help="Run using the built-in geth node", is_flag=True)
+
+# TODO: Make True by default or deprecate
```
cygnusv:
```suggestion
# TODO: Make True by default or deprecate - see #1439 
```

nucypher/cli/characters/alice.py
```diff
@@ -7,44 +7,170 @@
 from nucypher.characters.banners import ALICE_BANNER
 from nucypher.cli import actions, painting, types
 from nucypher.cli.actions import get_nucypher_password, select_client_account, get_client_password
-from nucypher.cli.config import nucypher_click_config
+from nucypher.cli.common_options import *
```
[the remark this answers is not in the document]
fjarri: Fixed in the new rebase.

nucypher/cli/characters/alice.py
```diff
@@ -7,44 +7,170 @@
...
+
+        pay_with = self.pay_with
+        if not pay_with and not self.federated_only:
+            pay_with = select_client_account(
+                emitter=emitter,
+                provider_uri=self.provider_uri)
+
+        return AliceConfiguration.generate(
+            password=get_nucypher_password(confirm=True),
+            config_root=config_root, # FIXME: or DEFAULT_CONFIG_ROOT?
```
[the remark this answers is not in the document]
fjarri: Fixed in the new rebase. `or` is not needed here, since `config_root` is never `None` (it is set to the value from the config object in the calling function).

nucypher/cli/characters/alice.py
```diff
@@ -7,44 +7,170 @@
...
+
+        if self.dev:
+            return AliceConfiguration(
+                emitter=emitter,
+                dev_mode=True,
+                network_middleware=self.middleware,
+                domains={TEMPORARY_DOMAIN},
+                provider_process=self.eth_node,
+                provider_uri=self.provider_uri,
+                federated_only=True) # FIXME: add error if --federated-only is not True?
```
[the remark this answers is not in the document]
fjarri: Added a check and an error. The error is only triggered when `federated_only` is set to `False` explicitly.

nucypher/cli/characters/alice.py
```diff
@@ -56,153 +182,76 @@ def alice():
...
-                                                   light=light,
-                                                   provider_uri=provider_uri,
-                                                   m=m,
-                                                   n=n,
-                                                   duration_periods=duration_periods,
-                                                   rate=rate)
+    emitter = _setup_emitter(general_config)
+
+    if not config_root:
+        config_root = general_config.config_file # FIXME: better called `config_root`?
```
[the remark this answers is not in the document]
fjarri: Added a commit that renames `config_file` to `config_root` (and renames the corresponding envvar as well).

nucypher/cli/characters/bob.py
```diff
@@ -6,42 +6,112 @@
 from nucypher.characters.banners import BOB_BANNER
 from nucypher.cli import actions, painting
 from nucypher.cli.actions import get_nucypher_password, select_client_account
-from nucypher.cli.config import nucypher_click_config
+from nucypher.cli.common_options import *
```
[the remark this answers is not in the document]
fjarri: Fixed in the new rebase.

nucypher/cli/characters/enrico.py
```diff
@@ -3,13 +3,9 @@
...
 from nucypher.characters.lawful import Enrico
-from nucypher.cli.config import nucypher_click_config
+from nucypher.cli.config import group_general_config
 from nucypher.cli.types import NETWORK_PORT
-
-
-policy_encrypting_key_option =  \
-    click.option('--policy-encrypting-key', help="Encrypting Public Key for Policy as hexadecimal string",
-                 type=click.STRING, required=True)
+from nucypher.cli.common_options import *
```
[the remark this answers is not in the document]
fjarri: Fixed in the new rebase.

nucypher/cli/characters/stake.py
fjarri: I agree with the reasoning, but I'll hold the renaming until we all agree on the final names.

nucypher/cli/common_options.py
fjarri: Fixed in the new rebase.

fjarri:
As far as I understand, the original `test_character_command_options_and_parameters_match()` checked two things:
1. `click_config` being present in CLI commands list.
2. The parameters of the CLI command correspond to its decorators.

I doubt the usefulness of 2. If there is a decorator not corresponding to a parameter, or, conversely, a parameter not corresponding to a decorator, there will be an exception thrown on call. The only possible event where an error can go unnoticed if there's an option in the decorator list which is already present in one of the groups. But do we really want to test for that? Seems to me that instead of testing a specific implementation we are better off just testing functionality.

Now as for 1., while it is slightly more reasonable (although it didn't prevent all the `deploy` commands miss logging options for a long time), I don't think it is the way to go either. I wonder if it is possible in `click` to define a set of options that will apply to all subcommands, without needing to specify them before subcommands on call? (That is, I wouldn't want to return to `nucypher --debug alice ...`)

(By the way, that reminds me, if the test is removed, I should remove `_check_command_options_and_parameters()` as well)

nucypher/cli/characters/stake.py
fjarri: Renamed to `StakerOptions` and `TransactingStakerOptions`.

nucypher/cli/characters/alice.py
```diff
@@ -7,44 +7,201 @@
...
-                  default=0)
-    @functools.wraps(func)
-    def wrapper(*args, **kwargs):
-        return func(*args, **kwargs)
-    return wrapper
+        # Managed Ethereum Client
+        eth_node = NO_BLOCKCHAIN_CONNECTION
+        if geth:
+            eth_node = actions.get_provider_process()
+            provider_uri = ETH_NODE.provider_uri(scheme='file')
```
derekpierre: @fjarri - I was working on top of this branch and noticed the that `ETH_NODE` is not defined - it should be `eth_node`.

PR #1437, opened 2019-10-22
Cleanup of Moe Status Page regarding checksum, graphs, and states

derekpierre:
Comments from @cygnusv in #963 :
- [x] The "Checksum" column is wrong. It's showing the worker address, instead of the checksum address.
- [x] In the "Num Stakers over the previous 30 days" chart, can you make the y-axis start from 0 (Discussion here: https://github.com/nucypher/nucypher/pull/963#issuecomment-544681150 and https://github.com/nucypher/nucypher/pull/963#issuecomment-544701507.)
- [x] The network status page shows the previous fleet states, but it doesn't show the current fleet state.

### DIFF
--- a/nucypher/network/status_app/base.py
+++ b/nucypher/network/status_app/base.py
@@ -13,7 +13,7 @@ import nucypher
 from nucypher.blockchain.eth.agents import ContractAgency, StakingEscrowAgent
 from nucypher.blockchain.eth.interfaces import BlockchainInterface
 from nucypher.characters.base import Character
-from nucypher.network.nodes import Learner
+from nucypher.network.nodes import Learner, FleetStateTracker
 
 
 class NetworkStatusPage:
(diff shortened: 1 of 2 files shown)


derekpierre:
@vepkenez expressed his views on Discord (below) regarding replacing the line chart for *number of stakers* with a bar chart. My feeling is to leave the line chart for now, but we can file an issue to change it to a bar chart if there are strong feelings about it.

```
damonCToday at 10:15 AM
I actually think you should switch them
since the bar chart represents integer number of stakers which are static per period much better, and the line chart represents the fractional NU values better
the line chart on stakers is kind of misleading because it interpolates throughout the course of a day
 
...
 
damonCToday at 11:18 AM
right. the dots are accurate, but the line itself is misleading (i think)
which is why a bar graph might be better for that case
like if we actually listed the number of stakers every minute of the day it wouldn't make the same shape as that line
it would be flat all day
 
derekToday at 11:19 AM
ah, I see your point. You view the line as the linear change of stakers throughout the course of previous day - whereas really its just relative to the previous end of day
 
damonCToday at 11:19 AM
and then the next day jump to the next number, like the bar graph
yes exactly
and it's a nitpick but for someone new to the space (which I relate to) it kinda opens the door to misunderstanding how things work
 
derekToday at 11:24 AM
Could that be fixed with labelling you think? That's a general feature of line graphs - If that's the case we shouldn't use line-graphs for anything not real-time
 
damonCToday at 11:28 AM
I'm no Edward Tufte or anything but, I kind of agree that maybe line graphs are not ideal for non realtime/interpolatable data.  Why not a bar graph?
...
michwillToday at 11:32 AM
I like the line graph, to be fair
 
damonCToday at 11:32 AM
but it lies!
yes it looks good
 
derekToday at 11:32 AM
Wouldn't active policies over time have the same problem?
 
michwillToday at 11:32 AM
Depends on how you interpret the number on the x axis?
 
damonCToday at 11:33 AM
policies can be made at any time during the day
so if I sample them at midnight each day, the sloping line interpolation is sort of a useful guess at what the number would be between the samples
which is completely not true with # of active stakers
also, the fact that the # of stakers is always a known integer value, but the line shows an infinite range of intermediate values that are never actually true
sorry, I'm starting to descend into insanity... I'll stop now 
```

PR #1438, opened 2019-10-24
Fix "invalid escape sequence" warnings for regexes in test_status

fjarri:
Just a small fix for annoying warnings. Turn regex f-strings into raw f-strings.


cygnusv:
This is fixed in biznaga, so perhaps it's better to just wait for it.

fjarri:
Uhh, sorry, should have checked. These epics sure have everything in them.

KPrasch:
Looks ready to merge but needs a rebase.

KPrasch:
Included in Biznaga.

PR #1446, opened 2019-11-07
Touchups to the CLI allocation process: support for CSV, bugfixes, etc.

cygnusv:
Command to use the allocation CLI:
```nucypher-deploy allocations --provider <PROVIDER> --poa  --allocation-infile <ALLOCATION_FILE>```

Example of allocation file in CSV:
```
"beneficiary_address","name","amount","duration_seconds"
"0x1715a31490371EBC469D775d6b384C354Be803C7","Aunt Irma",15000000000000000000,259200
```
Note you need those exact column names.

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -518,7 +518,7 @@ class ContractAdministrator(NucypherTokenActor):
             if emitter:
                 paint_deployed_allocations(emitter, allocated, failed)
 
-            csv_filename = f'allocations-{self.deployer_address[:6]}-{maya.now().epoch}.csv'
+            csv_filename = f'allocation-summary-{self.deployer_address[:6]}-{maya.now().epoch}.csv'
             csv_filepath = os.path.join(parent_path, csv_filename)
             write_deployed_allocations_to_csv(csv_filepath, allocated, failed)
             if emitter:
@@ -575,12 +575,12 @@ class ContractAdministrator(NucypherTokenActor):
         os.makedirs(DEFAULT_CONFIG_ROOT, exist_ok=True)
         with open(filepath, 'w') as file:
             data = dict()
-            for contract_name, receipts in receipts.items():
+            for contract_name, contract_receipts in receipts.items():
                 contract_records = dict()
-                for tx_name, receipt in receipts.items():
+                for tx_name, receipt in contract_receipts.items():
                     # Formatting
-                    receipt = {item: str(result) for item, result in receipt.items()}
-                    contract_records.update({tx_name: receipt for tx_name in receipts})
+                    pretty_receipt = {item: str(result) for item, result in receipt.items()}
+                    contract_records[tx_name] = pretty_receipt
                 data[contract_name] = contract_records
             data = json.dumps(data, indent=4)
             file.write(data)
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: nucypher/blockchain/eth/actors.py [hunk 1]
```diff
@@ -518,11 +533,22 @@ def deploy_beneficiary_contracts(self,
...
+                allocation_data = list()
+                reader = csv.DictReader(allocation_file)
+                for entry in reader:
+                    entry['amount'], entry['duration_seconds'] = int(entry['amount']), int(entry['duration_seconds'])
+                    allocation_data.append(entry)
+            else:
+                data = allocation_file.read()
+                try:
+                    allocation_data = json.loads(data)
+                except JSONDecodeError:
```
fjarri: Hm, and ignore any other errors? I understand that this `try` block came from the original version, but perhaps we can just remove it - the CSV part doesn't have anything like that.

nucypher/blockchain/eth/actors.py
```diff
@@ -383,13 +384,20 @@ def deploy_beneficiary_contracts(self,
                                      emitter: StdoutEmitter = None,
                                      ) -> Dict[str, dict]:
         """
-        The allocation file is a JSON file containing a list of allocations. Each allocation has a:
+        The allocation file is a CSV file containing a list of allocations. Each allocation has a:
```
fjarri: "A CSV or a JSON file", perhaps? Also, judging by the code, the CSV file is recognized by extension, and anything else is considered JSON - perhaps is worth mentioning too.

nucypher/blockchain/eth/actors.py
cygnusv: Current catch block doesn't achieve anything, so yes, perhaps it's better to simply remove it.

cygnusv: ✅

nucypher/blockchain/eth/actors.py
cygnusv: ✅

nucypher/blockchain/eth/actors.py
```diff
@@ -518,11 +533,19 @@ def deploy_beneficiary_contracts(self,
     @staticmethod
     def __read_allocation_data(filepath: str) -> list:
         with open(filepath, 'r') as allocation_file:
-            data = allocation_file.read()
-            try:
+            if filepath.endswith(".csv"):
+                reader = csv.DictReader(allocation_file)
+                allocation_data = list(reader)
+            else:  # Assume it's JSON by default
+                data = allocation_file.read()
```
michwill: Why not `json.load(allocation_file)`?

cygnusv: Why not indeed.

cygnusv:
@michwill I think it definitely falls within the scope of the PR. Let me try to add that too.

cygnusv:
> Not sure if it's within the scope of the PR... But what if we save non-failed allocations after every deployment to some csv file, so that if something terrible happened (CryptoKitty attack, hardware wallet broke down, ...), we can easily continue from the file which contains what's left to deploy?

Oh I just forgot that I already did that! 😅 
It doesn't create a separate file, but instead it reports failed allocations in the summary file (CSV format) which is created at the end of the process. This file cannot be directly used as input to the allocation process (since it also contains successful allocations) but it's a very good start to create a new file by hand. I think the human touch in such moments is important so perhaps it's good to not automate more than necessary.

michwill:
Sure, if you prefer this way.
Also, sometimes geth may report the transaction as "failed" when it succeeds (when there is a chain reorg, for example). It's actually a good idea to check if something is deployed or not already. Not sure if possible though

PR #1447, opened 2019-11-08
allow for decryption of intentionally unsigned data

vepkenez:
I guess this may be controversial, but if we allow for the encryption of unsigned data [here](https://github.com/nucypher/nucypher/blob/master/nucypher/crypto/api.py#L224), we should allow for that data to be decrypted as well right?

### DIFF
--- a/nucypher/characters/base.py
+++ b/nucypher/characters/base.py
@@ -30,7 +30,8 @@ from constant_sorrow.constants import (
     SIGNATURE_TO_FOLLOW,
     SIGNATURE_IS_ON_CIPHERTEXT,
     STRANGER,
-    FEDERATED_ONLY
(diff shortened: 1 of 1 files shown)


tuxxy:
While I'm open to allowing the decryption of unsigned data, we need to figure out how to handle this UX. Previous experiences have demonstrated that a warning is not enough. We need to be able to present the risks to the User and have them understand and, with their consent, proceed to decrypt the data.

tuxxy:
Then again, there could be hardly any risks to it and we just need to think very carefully about it.

michwill:
Any tests?

KPrasch:
Moving to https://github.com/nucypher/nucypher/issues/1563

PR #1456, opened 2019-11-12
Pagination for `getAllActiveStakers` method

vzotova:
* Pagination for `getAllActiveStakers()` to mitigate problems with light node/client (geth/parity)
* Added "light" parameter to the blockchain interface
* Use "light" marker to get default pagination in `StakingEscrowAgent`

### DIFF
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -185,6 +185,8 @@ class StakingEscrowAgent(EthereumContractAgent):
     registry_contract_name = STAKING_ESCROW_CONTRACT_NAME
     _proxy_name = DISPATCHER_CONTRACT_NAME
 
+    DEFAULT_PAGINATION_SIZE = 30    # TODO test on goerli
+
     class NotEnoughStakers(Exception):
         pass
 
@@ -228,28 +230,33 @@ class StakingEscrowAgent(EthereumContractAgent):
 
         return active_stakers, pending_stakers, missing_stakers
 
-    def get_all_active_stakers(self, periods: int, pagination_size: int = 0) -> Tuple[int, list]:
+    def get_all_active_stakers(self, periods: int, pagination_size: int = None) -> Tuple[int, list]:
         """Only stakers which confirmed the current period (in the previous period) are used."""
         if not periods > 0:
             raise ValueError("Period must be > 0")
 
-        if pagination_size != 0:
-            stakers_length = self.contract.functions.getStakersLength().call()
+        if pagination_size is None:
+            pagination_size = StakingEscrowAgent.DEFAULT_PAGINATION_SIZE if self.blockchain.is_light else 0
+        elif pagination_size < 0:
+            raise ValueError("Pagination size must be >= 0")
+
+        if pagination_size > 0:
+            num_stakers = self.get_staker_population()
             start_index = 0
             n_tokens = 0
             stakers = list()
-            while start_index < stakers_length:
+            while start_index < num_stakers:
                 temp_locked_tokens, temp_stakers = \
-                    self.contract.functions.getAllActiveStakers(periods, start_index, pagination_size).call()
+                    self.contract.functions.getActiveStakers(periods, start_index, pagination_size).call()
                 n_tokens += temp_locked_tokens
                 stakers += temp_stakers
                 start_index += pagination_size
         else:
-            n_tokens, stakers = self.contract.functions.getAllActiveStakers(periods, 0, 0).call()
+            n_tokens, stakers = self.contract.functions.getActiveStakers(periods, 0, 0).call()
 
         return n_tokens, stakers
 
-    def get_all_locked_tokens(self, periods: int, pagination_size: int = 0) -> int:
+    def get_all_locked_tokens(self, periods: int, pagination_size: int = None) -> int:
         all_locked_tokens, _stakers = self.get_all_active_stakers(periods=periods, pagination_size=pagination_size)
         return all_locked_tokens
 
@@ -486,7 +493,7 @@ class StakingEscrowAgent(EthereumContractAgent):
                duration: int,
                additional_ursulas:float = 1.5,
                attempts: int = 5,
-               pagination_size: int = 0
+               pagination_size: int = None
                ) -> List[str]:
         """
         Select n random Stakers, according to their stake distribution.
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -228,12 +228,29 @@ class StakingEscrowAgent(EthereumContractAgent):
 
         return active_stakers, pending_stakers, missing_stakers
 
-    def get_all_locked_tokens(self, periods: int) -> int:
-        """Returns the current period"""
+    def get_all_active_stakers(self, periods: int, pagination_size: int = 0) -> Tuple[int, list]:
+        """Only stakers which confirmed the current period (in the previous period) are used."""
         if not periods > 0:
             raise ValueError("Period must be > 0")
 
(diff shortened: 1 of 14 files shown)


nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -358,6 +364,9 @@ contract StakingEscrow is Issuer {
                 allLockedTokens = allLockedTokens.add(lockedTokens);
             }
         }
+        assembly {
+            mstore(activeStakers, resultIndex)
```
michwill: What is this doing?

cygnusv: It's storing the array length in the first memory address

michwill: More of a question is: why `mstore` in assembly instead of addressing `activeStakers` as an array? Optimization?

vzotova: There is no such thing like dynamic memory arrays in solidity, so this line is set real length for result memory array (removes zeros at the end of the array)

michwill: Ah I see. Very cool!

nucypher/blockchain/eth/agents.py
```diff
@@ -228,12 +230,34 @@ def partition_stakers_by_activity(self) -> Tuple[List[str], List[str], List[str]
...
+            n_tokens = 0
+            stakers = list()
+            while start_index < num_stakers:
+                temp_locked_tokens, temp_stakers = \
+                    self.contract.functions.getActiveStakers(periods, start_index, pagination_size).call()
+                n_tokens += temp_locked_tokens
+                stakers += temp_stakers
+                start_index += pagination_size
+        else:
+            n_tokens, stakers = self.contract.functions.getActiveStakers(periods, 0, 0).call()
```
michwill: One interesting way could be to catch an error here and start paginating if the "heavy" call fails

vzotova: Interesting idea, I'd prefer more predictable solution and this method could fail with other errors (not in reality, just potentially), but auto pagination could work
if we don't need "light" flag then it definitely will be better

nucypher/blockchain/eth/agents.py
```diff
@@ -228,12 +230,34 @@ def partition_stakers_by_activity(self) -> Tuple[List[str], List[str], List[str]
 
         return active_stakers, pending_stakers, missing_stakers
 
-    def get_all_locked_tokens(self, periods: int) -> int:
-        """Returns the current period"""
+    def get_all_active_stakers(self, periods: int, pagination_size: int = None) -> Tuple[int, list]:
```
tuxxy: nice

nucypher/blockchain/eth/interfaces.py
```diff
@@ -95,6 +95,7 @@ class UnknownContract(InterfaceError):
 
     def __init__(self,
                  poa: bool = True,
+                 light: bool = False,
```
tuxxy: I don't like having `is_light` vs `light`, but I guess we can't really avoid it with click.

vzotova: you don't like the name of the parameter or the parameter in itself?

nucypher/blockchain/eth/agents.py
cygnusv: I'd love to help on a dynamic pagination size. Sounds like a fun tiny project!

nucypher/blockchain/eth/interfaces.py
tuxxy: Just the name. I don't like how click forces us to use a different param name. :P

nucypher/blockchain/eth/agents.py
vzotova: It could be handy, but from development perspective I don't like it: for agent user better to understand what doing each method, but this method will be too smart. Also to do it correct we need to distinguish `TransactionFailed` from timeout problems

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: In this way better to return old name. I just thought there are not too many calls for policy creation right now

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -329,22 +329,28 @@ contract StakingEscrow is Issuer {
...
     * @return allLockedTokens Sum of locked tokens for active stakers
     * @return activeStakers Array of stakers and their locked tokens. Stakers addresses stored as uint256
     **/
-    function getAllActiveStakers(uint16 _periods)
+    function getActiveStakers(uint16 _periods, uint256 _startIndex, uint256 _maxStakers)
         external view returns (uint256 allLockedTokens, uint256[2][] memory activeStakers)
     {
         require(_periods > 0);
 
         uint256 stakersLength = stakers.length;
```
cygnusv: BTW, just a minor additional comment: this variable should be renamed to `endIndex` or something similar.

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -329,22 +329,28 @@ contract StakingEscrow is Issuer {
...
+        require(_startIndex < stakersLength);
         activeStakers = new uint256[2][](stakersLength);
```
cygnusv: Wouldn't be better to allocate the array with `stakersLength - _startIndex` length (after `stakersLength` is updated later, of course)? It'd reduce gas, which is one of the reasons we need this PR, even if it's a call.

vzotova: I don't think so, because memory array initialization - get empty memory pointer and write length variable.
But I'll try it, maybe it will reduce gas

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: agree with renaming, it's `endIndex` + 1, which name will be more understandable?

cygnusv: In several languages, the end index is assumed to be exclusive, so not included in the array.

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: It gave a little bit gas savings, thanks for suggestions
✔️

nucypher/blockchain/eth/agents.py
```diff
@@ -228,12 +230,34 @@ def partition_stakers_by_activity(self) -> Tuple[List[str], List[str], List[str]
...
+        """Only stakers which confirmed the current period (in the previous period) are used."""
         if not periods > 0:
             raise ValueError("Period must be > 0")
 
-        all_locked_tokens, _stakers = self.contract.functions.getAllActiveStakers(periods).call()
+        if pagination_size is None:
+            pagination_size = StakingEscrowAgent.DEFAULT_PAGINATION_SIZE if self.blockchain.is_light else 0
```
KPrasch: This is the *light mode" entry point... yeah?

vzotova: yep, I see this as - all calls that will be broken in light mode will use something like this

Issue #1406, opened 2019-10-09
Make sure we support Policy authors to be smart contracts


cygnusv:
Let's say there's a DApp that want to reduce UX friction for Alice, abstracting the need for ETH funding to participate in NuCypher as policy authors. For example, this DApp can accept whatever payment is better for Alices and act as a proxy to the NuCypher Network.

In an eye-opening discussion with @szotov, he points out that this can be done by the DApp by setting up a contract acting as the policy author. Since Alice's ETH key is not tied anyway with her cryptographic keys (i.e, signing, encryption, delegation), it seems that in principle this approach has a lot of sense for this use case.

On the other hand, this type of use case doesn't feel like something we should implement in our code base – in other words, it's not our problem. Except that if we want to increase traction, then we should make sure that it's doable, even if we don't do it.

The question is: is there something in the codebase preventing this use case from being possible? 

Pinging @afrancht for the original discussion that led to this issue.

afrancht:
I think that @cygnusv is right in stating that this is not something that NuCypher should implement but should make sure it is possible. 

Here is a small diagram of how I would see it implemented:
![NuCypher with a Policy Manager](https://user-images.githubusercontent.com/19917183/66505000-5b6dfa80-eac2-11e9-92da-a7555d8d1b37.png)

### Why NuCypher shouldn't implement it
In the context of the NuCypher network (as I understand it), it makes sense that each Alice takes care of their policies. This is because proxy re-encryption empowers "the user that has something to lose" (their plaintext/secret) to revoke a policy whenever she sees fit. Therefore, it would contradict your narrative as having a policy manager "out of context" introduces the issue of Alice having to trust the policy manager to create the policy right and with your current narrative, it is clear that this is not the case. 

Additionally, "the Policy Manager" abstracts the user away from managing policies for the sake of a better UX and UX varies from implementation to implementation thus creating one solution to fit all may not be possible. Example: in my implementation, I perhaps do not want to use the nucypher keyring and I want to provide the keys myself (as a user may make use of the NuCypher network from different devices). With the policy manager it would be a bit similar if how you create the policy is somewhat different to what I need, I will find myself writing my policy manager anyways.   

DISCLAIMER: I am not 100% sure I can use the NuCypher network without using the NuCypher keyring. I know that is possible with Umbral as you can provide a key as a parameter but if not that would also be a possible feature that increases usability and traction. 

### Why it should be possible
[...]

Issue #1188, opened 2019-08-02
Choose a start date for a sharing policy that's not today/tomorrow 


arjunhassard:
Currently `create_policy` accepts an expiration datetime and a number of duration periods, but by default the policy becomes active either a) when the grant function executes or b) at the start of the next period (i.e. within 24h) 

Some Alices, across various use cases, may wish to future-date their sharing policies. For example, a staking-as-a-service company setting up a new validator instance wants its access to signing keys (and the unique conditions for that access) to commence at a specific date in the future, such that this access is synchronised with other processes/infrastructure. 

Arguably this future-dating could be configured on Alice's machine, since she needs to be online and connected to the network for the policy creation in any case. On the other hand, if a re-encryption key can be generated by Alice and held by Ursulas in advance, including the condition of delayed access to a Bob, this would surely enhance the _Disappearing Alice_ source of value (@jMyles – that's for you <3)

This would impact the `first_period_value` & `first_period_rate` – see #1063 

Future ideas: sharing policy start dates based on arbitrary conditions, not just a datetime. For example, if we end up with Ursulas actively ingesting block heights from various networks, then reaching a certain epoch could trigger access for a set of pre-designated validators.

derekpierre:
I like the idea of the feature.

Question: How does that affect the cost of the policy? Is the cost of the policy just the time from start date to expiration OR does it need to also include the time between issuance of the grant and the start date as well? At issuance, Ursulas would be storing the kfrags, and actually enforcing a policy by not letting the policy be applicable before the start date arrives. Should the cost for the period between issuance of grant and the start-date be less since there isn't any re-encryption being done, just a witholding of execution of the policy until start-time?

arjunhassard:
Following verbal discussion with @szotov. 
Note: hereafter the time between policy creation and policy start date is the **pre-activation period**.

_Fees_
As @derekpierre points out, Ursulas are still responsible for holding the kfrags during the pre-activation period, so some retainer fee is required at a minimum. Indeed, if Alice couldn't deliver the kfrag in advance, the value of this feature is severely diminished. 

However, Ursula do not technically need to be online in this period (acknowledging that in practice, they are likely to have other active policies requiring high uptime), nor perform any re-encryptions. 

Hence, a lower fee, also based on duration, is likely the best answer. 
Downside: extra gas costs for storing two payment metrics; shouldered by Alice, since she's invoking an advanced feature.

_Revocation_
[...]

PR #1459, opened 2019-11-16
Policy refactoring

### LINKED ISSUE #1063: How to determine default `first_period_value`
see: <https://github.com/nucypher/nucypher/pull/984/files#r291809731>

vzotova:
Based on #1470
Refs #1188 
Closes #1063 
Refs #1406
Refs #1492

| WARNING: Not upgradeable changes |
| --- |

* discrete payment for all policy periods: if policy starts today, ends tomorrow then fee must be for 2 periods
* flexible end timestamp
* Ursula pays more for confirmActivity (~15k-20k gas) to minimize Alice gas fee (save ~15k per node)
* split creator role into creator and owner
* revocation using owner's signature 

PolicyManager `v1.1.2` -> `v 2.1.1`
StakingEscrow `v1.4.1` -> `v2.1.1`

### DIFF
--- a/docs/source/architecture/contracts.md
+++ b/docs/source/architecture/contracts.md
@@ -21,14 +21,14 @@ For a guide of how to deploy these contracts automatically, see the [Deployment 
 2. Deploy `StakingEscrow` with a dispatcher targeting it
 3. Deploy `PolicyManager` with its own dispatcher, also targeting it
 4. Deploy `Adjudicator` with a dispatcher
-5. Deploy `WorkLock` contract
-6. Transfer reward tokens to the `StakingEscrow` contract. These tokens are future mining rewards and initial allocations
-7. Run the `initialize()` method to initialize the `StakingEscrow` contract
-8. Set the address of the `PolicyManager` contract  in the `StakingEscrow` by using the `setPolicyManager(address)`
-9. Set the address of the `Adjudicator` contract  in the `StakingEscrow` by using the `setAdjudicator(address)`
-10. Set the address of the `WorkLock` contract  in the `StakingEscrow` by using the `setWorkLock(address)`
-11. Transfer tokens for distribution to the `WorkLock` contract
-12. Deploy `StakingInterface` with `StakingInterfaceRouter` targeting it
+5. Deploy `StakingInterface` with `StakingInterfaceRouter` targeting it
+6. Deploy `WorkLock` contract
+7. Set the address of the `PolicyManager` contract  in the `StakingEscrow` by using the `setPolicyManager(address)`
+8. Set the address of the `Adjudicator` contract  in the `StakingEscrow` by using the `setAdjudicator(address)`
+9. Set the address of the `WorkLock` contract  in the `StakingEscrow` by using the `setWorkLock(address)`
+10. Approve tokens transfer to the `StakingEscrow` contract. These tokens are future mining rewards
+11. Run the `initialize(uint256)` method to initialize the `StakingEscrow` contract
+12. Approve tokens transfer for distribution to the `WorkLock` contract and call `tokenDeposit(uint256)` method
 13. Pre-deposit tokens to the `PreallocationEscrow`:
 	* Create new instance of the `PreallocationEscrow` contract 
 	* Transfer ownership of the instance of the `PreallocationEscrow` contract to the user
@@ -41,22 +41,24 @@ For a guide of how to deploy these contracts automatically, see the [Deployment 
 
 Alice uses a network of Ursula stakers to deploy policies.
 In order to take advantage of the network, Alice chooses stakers and deploys policies with fees for those stakers.
-Alice can choose stakers by herself ("handpicked") or by using `StakingEscrow.sample(uint256[], uint16)` - This is  known as ("sampling").
-`sample` parameters are:
-* The array of absolute values
+Alice can choose stakers by herself ("handpicked") or select from the result of `StakingEscrow.getActiveStakers(uint16, uint256, uint256)` method - This is  known as ("sampling").
+`getActiveStakers` parameters are:
 * Minimum number of periods during which tokens are locked
+* Start index for looking in stakers array 
+* Max stakers for looking
 This method will return only active stakers.
 
-In order to place the fee for a policy, Alice calls the method `PolicyManager.createPolicy(bytes16, uint16, uint256, address[])`,
-specifying the staker's addresses, the policy ID (off-chain generation), the policy duration in periods, and the first period's reward.
-Payment should be added to the transaction in ETH and the amount is `firstReward * stakers.length + rewardRate * periods * stakers.length`.
-The reward rate must be greater than or equal to the minimum reward for each staker in the list. The first period's reward is not refundable, and can be zero.
+In order to place the fee for a policy, Alice calls the method `PolicyManager.createPolicy(bytes16, address, uint64, address[])`,
+specifying the staker's addresses, the policy ID (off-chain generation), the policy owner (could be zero address), and the end timestamp of the policy.
+Payment should be added to the transaction in ETH and the amount is `rewardRate * periods * stakers.length`, where `periods` is `endTimestampPeriod - currentPeriod + 1`.
+The reward rate must be greater than or equal to the minimum reward for each staker in the list.
 
 ### Alice Revokes a Blockchain Policy
 
 When Alice wants to revoke a policy, she calls the `PolicyManager.revokePolicy(bytes16)` or `PolicyManager.revokeArrangement(bytes16, address)`.
 Execution of these methods results in Alice recovering all fees for future periods, and also for periods when the stakers were inactive.
 Alice can refund ETH for any inactive periods without revoking the policy by using the method `PolicyManager.refund(bytes16)` or `PolicyManager.refund(bytes16, address)`.
+If Alice doesn't have ability to execute on-chain transaction or wants to share ability to revoke then she can sign revocation parameters. Anyone who have this signature will be able to revoke policy using `PolicyManager.revoke(bytes16, address, bytes)`
 
 
 ## Staker's Contract Interaction
@@ -118,7 +120,7 @@ The staker can set a minimum reward rate for a policy. For that, the staker shou
 ### NuCypher Partner Ursula Staking
 Some users will have locked but not staked tokens.
 In that case, an instance of the `PreallocationEscrow` contract will hold their tokens (method `PreallocationEscrow.initialDeposit(uint256, uint256)`).
-All tokens will be unlocked after a specified time and the user can retrieve them using the `PreallocationEscrow.withdraw(uint256)` method.
+All tokens will be unlocked after a specified time and the user can retrieve them using the `PreallocationEscrow.withdrawTokens(uint256)` method.
 When the user wants to become a staker - they use the `PreallocationEscrow` contract as a proxy for the `StakingEscrow` and `PolicyManager` contracts.
 
 
--- a/docs/source/architecture/contracts.md
+++ b/docs/source/architecture/contracts.md
@@ -20,13 +20,13 @@ For a guide of how to deploy these contracts automatically, see the [Deployment 
 1. Deploy `NuCypherToken` with all future supply tokens
 2. Deploy `StakingEscrow` with a dispatcher targeting it
 3. Deploy `PolicyManager` with its own dispatcher, also targeting it
-4. Deploy `Adjudicator` with a dispatcher
-5. Deploy `StakingInterface` with `StakingInterfaceRouter` targeting it
-6. Deploy `WorkLock` contract
-7. Set the address of the `PolicyManager` contract  in the `StakingEscrow` by using the `setPolicyManager(address)`
-8. Set the address of the `Adjudicator` contract  in the `StakingEscrow` by using the `setAdjudicator(address)`
+4. Set the address of the `PolicyManager` contract  in the `StakingEscrow` by using the `setPolicyManager(address)`
+5. Deploy `Adjudicator` with a dispatcher
+6. Set the address of the `Adjudicator` contract  in the `StakingEscrow` by using the `setAdjudicator(address)`
+7. Deploy `StakingInterface` with `StakingInterfaceRouter` targeting it
+8. Deploy `WorkLock` contract
 9. Set the address of the `WorkLock` contract  in the `StakingEscrow` by using the `setWorkLock(address)`
-10. Approve tokens transfer to the `StakingEscrow` contract. These tokens are future mining rewards
+10. Approve tokens transfer to the `StakingEscrow` contract. These tokens are future staking rewards
 11. Run the `initialize(uint256)` method to initialize the `StakingEscrow` contract
 12. Approve tokens transfer for distribution to the `WorkLock` contract and call `tokenDeposit(uint256)` method
 13. Pre-deposit tokens to the `PreallocationEscrow`:
@@ -44,12 +44,12 @@ In order to take advantage of the network, Alice chooses stakers and deploys pol
 Alice can choose stakers by herself ("handpicked") or select from the result of `StakingEscrow.getActiveStakers(uint16, uint256, uint256)` method - This is  known as ("sampling").
 `getActiveStakers` parameters are:
 * Minimum number of periods during which tokens are locked
-* Start index for looking in stakers array 
-* Max stakers for looking
+* Start index in stakers array 
+* Maximum number of stakers
 This method will return only active stakers.
 
 In order to place the fee for a policy, Alice calls the method `PolicyManager.createPolicy(bytes16, address, uint64, address[])`,
-specifying the staker's addresses, the policy ID (off-chain generation), the policy owner (could be zero address), and the end timestamp of the policy.
+specifying the policy ID (off-chain generation), the policy owner (could be zero address), the end timestamp of the policy and the stakers' addresses.
 Payment should be added to the transaction in ETH and the amount is `rewardRate * periods * stakers.length`, where `periods` is `endTimestampPeriod - currentPeriod + 1`.
 The reward rate must be greater than or equal to the minimum reward for each staker in the list.
 
@@ -58,7 +58,7 @@ The reward rate must be greater than or equal to the minimum reward for each sta
 When Alice wants to revoke a policy, she calls the `PolicyManager.revokePolicy(bytes16)` or `PolicyManager.revokeArrangement(bytes16, address)`.
 Execution of these methods results in Alice recovering all fees for future periods, and also for periods when the stakers were inactive.
 Alice can refund ETH for any inactive periods without revoking the policy by using the method `PolicyManager.refund(bytes16)` or `PolicyManager.refund(bytes16, address)`.
-If Alice doesn't have ability to execute on-chain transaction or wants to share ability to revoke then she can sign revocation parameters. Anyone who have this signature will be able to revoke policy using `PolicyManager.revoke(bytes16, address, bytes)`
+If Alice can't execute an on-chain transaction or wants to share the ability to revoke, then she can sign revocation parameters. Anyone who has this signature will be able to revoke policy using `PolicyManager.revoke(bytes16, address, bytes)`
 
 
 ## Staker's Contract Interaction
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1048,7 +1048,6 @@ class BlockchainPolicyAuthor(NucypherTokenActor):
                  checksum_address: str,
                  rate: int = None,
                  duration_periods: int = None,
-                 first_period_reward: int = None,
                  *args, **kwargs):
         """
         :param policy_agent: A policy agent with the blockchain attached;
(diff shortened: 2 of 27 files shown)


tests/cli/ursula/test_stakeholder_and_ursula.py
```diff
@@ -375,7 +375,9 @@ def test_collect_rewards_integration(click_runner,
     blockchain_alice.selection_buffer = 1
 
     M, N = 1, 1
-    expiration = maya.now() + datetime.timedelta(days=3)
+    days = 3
+    now = testerchain.w3.eth.getBlock(block_identifier='latest').timestamp
+    expiration = maya.MayaDT(now).add(days=days-1)
```
derekpierre: I don't know if I fully grasp this - we use `days-1` as opposed to `days` because the current day counts as one i.e. is included?

nucypher/blockchain/eth/actors.py
```diff
@@ -1045,23 +1037,26 @@ def generate_policy_parameters(self,
...
-                                            seconds_per_period=self.economics.seconds_per_period)
+                                            seconds_per_period=self.economics.seconds_per_period,
+                                            start_of_period=True)
+            expiration -= 1  # Get the last second of the target period
+        else:
+            now = self.staking_agent.blockchain.w3.eth.getBlock(block_identifier='latest').timestamp
+            duration_periods = calculate_period_duration(now=maya.MayaDT(now),
+                                                         future_time=expiration,
+                                                         seconds_per_period=self.economics.seconds_per_period)
+            duration_periods += 1  # Number of all included periods
```
derekpierre: Are we using `+=1` because the current day counts as one? Should `calculate_period_duration` make this adjustment, instead of having to manually do it?

nucypher/policy/policies.py
```diff
@@ -498,55 +496,42 @@ def __init__(self,
...
-        rate_per_period = ((self.value // self.n) - self.first_period_reward) // self.duration_periods  # wei
-        if rate_per_period <= self.first_period_reward:
-            raise ValueError(
-                f"Policy rate ({rate_per_period} wei) is less then the initial reward ({self.first_period_reward})")
-
-        recalculated_value = ((self.duration_periods * rate_per_period) + self.first_period_reward) * self.n
+        rate_per_period = self.value // self.n // self.duration_periods  # wei
+        recalculated_value = self.duration_periods * rate_per_period * self.n
         if recalculated_value != self.value:
             raise ValueError(f"Invalid policy value calculation.")  # TODO: Make a better suggestion.
```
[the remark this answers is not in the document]
vzotova: I think this PR is good place to fix this, any suggestions?

tests/cli/ursula/test_stakeholder_and_ursula.py
vzotova: Days here means how many days must be paid, duration could be one period from noon to noon but need to pay for two periods.

nucypher/blockchain/eth/actors.py
vzotova: Yeah, you're right, `calculatePeriodDuration` returns `periodAt(expiration) - periodAt(now)` and we add current period.
> Should `calculate_period_duration` make this adjustment, instead of having to manually do it?

Good question, I'm still not sure what behaviour is more straightforward

nucypher/policy/policies.py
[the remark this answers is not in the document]
derekpierre: I'm unsure what the subsequent suggestion should be for the developer (I assume this is not expected to occur) but perhaps something like...

```suggestion
            raise ValueError(f"Mismatched policy value calculation when applying rate per period ({recalculated_value}) vs expected overall value ({self.value}).")
```

tests/cli/ursula/test_stakeholder_and_ursula.py
derekpierre: 👍

nucypher/blockchain/eth/actors.py
derekpierre: It would seem that it should to me, but it may depend on where else it is used, and whether that same `+=1` logic applies.

nucypher/policy/policies.py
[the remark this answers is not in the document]
vzotova: Error more related to situation when user specifies `value` and this value can't be divided without remainder.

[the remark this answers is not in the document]
derekpierre: Would this be better then?:

```suggestion
            raise ValueError(f"Invalid policy value calculation - {self.value} can't be divided into {self.n} staker payments per period for {self.duration_periods} periods without a remainder")
```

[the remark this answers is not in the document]
vzotova: I like it

docs/source/architecture/contracts.md
```diff
@@ -41,22 +41,24 @@ For a guide of how to deploy these contracts automatically, see the [Deployment
...
+specifying the staker's addresses, the policy ID (off-chain generation), the policy owner (could be zero address), and the end timestamp of the policy.
+Payment should be added to the transaction in ETH and the amount is `rewardRate * periods * stakers.length`, where `periods` is `endTimestampPeriod - currentPeriod + 1`.
+The reward rate must be greater than or equal to the minimum reward for each staker in the list.
 
 ### Alice Revokes a Blockchain Policy
 
 When Alice wants to revoke a policy, she calls the `PolicyManager.revokePolicy(bytes16)` or `PolicyManager.revokeArrangement(bytes16, address)`.
 Execution of these methods results in Alice recovering all fees for future periods, and also for periods when the stakers were inactive.
 Alice can refund ETH for any inactive periods without revoking the policy by using the method `PolicyManager.refund(bytes16)` or `PolicyManager.refund(bytes16, address)`.
+If Alice doesn't have ability to execute on-chain transaction or wants to share ability to revoke then she can sign revocation parameters. Anyone who have this signature will be able to revoke policy using `PolicyManager.revoke(bytes16, address, bytes)`
```
derekpierre:
```suggestion
If Alice can't execute an on-chain transaction or wants to share the ability to revoke, then she can sign revocation parameters. Anyone who has this signature will be able to revoke policy using `PolicyManager.revoke(bytes16, address, bytes)`
```

docs/source/architecture/contracts.md
```diff
@@ -41,22 +41,24 @@ For a guide of how to deploy these contracts automatically, see the [Deployment
 
 Alice uses a network of Ursula stakers to deploy policies.
 In order to take advantage of the network, Alice chooses stakers and deploys policies with fees for those stakers.
-Alice can choose stakers by herself ("handpicked") or by using `StakingEscrow.sample(uint256[], uint16)` - This is  known as ("sampling").
-`sample` parameters are:
-* The array of absolute values
+Alice can choose stakers by herself ("handpicked") or select from the result of `StakingEscrow.getActiveStakers(uint16, uint256, uint256)` method - This is  known as ("sampling").
+`getActiveStakers` parameters are:
 * Minimum number of periods during which tokens are locked
+* Start index for looking in stakers array
```
derekpierre:
```suggestion
* Start index in stakers array 
```

docs/source/architecture/contracts.md
```diff
@@ -41,22 +41,24 @@ For a guide of how to deploy these contracts automatically, see the [Deployment
...
 Alice uses a network of Ursula stakers to deploy policies.
 In order to take advantage of the network, Alice chooses stakers and deploys policies with fees for those stakers.
-Alice can choose stakers by herself ("handpicked") or by using `StakingEscrow.sample(uint256[], uint16)` - This is  known as ("sampling").
-`sample` parameters are:
-* The array of absolute values
+Alice can choose stakers by herself ("handpicked") or select from the result of `StakingEscrow.getActiveStakers(uint16, uint256, uint256)` method - This is  known as ("sampling").
+`getActiveStakers` parameters are:
 * Minimum number of periods during which tokens are locked
+* Start index for looking in stakers array 
+* Max stakers for looking
```
derekpierre:
```suggestion
* Maximum number of stakers

File: docs/source/architecture/contracts.md [hunk 7]
```diff
@@ -41,22 +41,24 @@ For a guide of how to deploy these contracts automatically, see the [Deployment
...
 This method will return only active stakers.
 
-In order to place the fee for a policy, Alice calls the method `PolicyManager.createPolicy(bytes16, uint16, uint256, address[])`,
-specifying the staker's addresses, the policy ID (off-chain generation), the policy duration in periods, and the first period's reward.
-Payment should be added to the transaction in ETH and the amount is `firstReward * stakers.length + rewardRate * periods * stakers.length`.
-The reward rate must be greater than or equal to the minimum reward for each staker in the list. The first period's reward is not refundable, and can be zero.
+In order to place the fee for a policy, Alice calls the method `PolicyManager.createPolicy(bytes16, address, uint64, address[])`,
+specifying the staker's addresses, the policy ID (off-chain generation), the policy owner (could be zero address), and the end timestamp of the policy.
```
derekpierre:
```suggestion
specifying the stakers' addresses, the policy ID (off-chain generation), the policy owner (could be zero address), and the end timestamp of the policy.
```

docs/source/architecture/contracts.md
vzotova: ✔️

docs/source/architecture/contracts.md
vzotova: ✔️

docs/source/architecture/contracts.md
vzotova: ✔️

docs/source/architecture/contracts.md
vzotova: ✔️

nucypher/blockchain/eth/actors.py
vzotova: `calculatePeriodDuratio` used in `initialize_stake` without adding 1

tests/blockchain/eth/contracts/integration/test_intercontract_integration.py
```diff
@@ -33,7 +33,7 @@
 
 DISABLE_RE_STAKE_FIELD = 3
 
-DISABLED_FIELD = 5
+DISABLED_FIELD = 4
```
jMyles: I'm sure it's supposed to be obvious, but help me catch up: why did this change?

nucypher/blockchain/eth/actors.py
```diff
@@ -1062,23 +1054,26 @@ def generate_policy_parameters(self,
...
-            duration_periods = calculate_period_duration(future_time=expiration,
-                                                         seconds_per_period=self.economics.seconds_per_period)
-        else:
-            duration_periods = duration_periods or self.duration_periods
+        if duration_periods:
+            # Duration equals one period means that expiration date is the last second of the current period
             expiration = datetime_at_period(self.staking_agent.get_current_period() + duration_periods,
-                                            seconds_per_period=self.economics.seconds_per_period)
+                                            seconds_per_period=self.economics.seconds_per_period,
+                                            start_of_period=True)
```
jMyles: The comment says "expiration date is the last second of the current period", but here we pass `start_of_period=True` - what am I missing?

jMyles:
OK, after a first full read-through, here's what I think I'm seeing - correct me where I'm wrong and add what I'm missing:

* Policies don't have clients, they have creators
* Policy creation and revocation events don't have clients, they have senders
* ...no wait, policy *creation* doesn't have a sender, it has a creator and an owner
* We use timestamps instead of periods for Policy beginnings and endings
* We check revocation by checking the signature of policy owner.

nucypher/blockchain/eth/actors.py
vzotova: We take start of the next period and subtract 1 second (next line)

tests/blockchain/eth/contracts/integration/test_intercontract_integration.py
vzotova: Changed index of `disabled` slot in PolicyManager.Policy struct, because `firstPartialReward` was removed

vzotova:
@jMyles All good, I'll add some remarks:
* `client` -> `creator` and `client` -> `sender` are just renames without changing meaning because `client` initially was not good name for this role
* Before the changes, Alice/client/creator could choose whether she would pay for the current period and how much, now always have to pay the same price as for other periods.

tests/blockchain/eth/contracts/integration/test_intercontract_integration.py
jMyles: Ahh, roger, got it.

docs/source/architecture/contracts.md
```diff
@@ -21,14 +21,14 @@ For a guide of how to deploy these contracts automatically, see the [Deployment
...
-7. Run the `initialize()` method to initialize the `StakingEscrow` contract
-8. Set the address of the `PolicyManager` contract  in the `StakingEscrow` by using the `setPolicyManager(address)`
-9. Set the address of the `Adjudicator` contract  in the `StakingEscrow` by using the `setAdjudicator(address)`
-10. Set the address of the `WorkLock` contract  in the `StakingEscrow` by using the `setWorkLock(address)`
-11. Transfer tokens for distribution to the `WorkLock` contract
-12. Deploy `StakingInterface` with `StakingInterfaceRouter` targeting it
+5. Deploy `StakingInterface` with `StakingInterfaceRouter` targeting it
+6. Deploy `WorkLock` contract
+7. Set the address of the `PolicyManager` contract  in the `StakingEscrow` by using the `setPolicyManager(address)`
+8. Set the address of the `Adjudicator` contract  in the `StakingEscrow` by using the `setAdjudicator(address)`
```
cygnusv: This two steps currently happen immediately after deploying the corresponding dispatcher. So it's:
3. Deploy PolicyManager and dispatcher
4. SetPolicyManager
5. Deploy Adjudicator and dispatcher
6. SetAdjuticator

In principle, it doesn't matter, but perhaps we should change it to reflect our current workflow.

docs/source/architecture/contracts.md
```diff
@@ -21,14 +21,14 @@ For a guide of how to deploy these contracts automatically, see the [Deployment
...
+9. Set the address of the `WorkLock` contract  in the `StakingEscrow` by using the `setWorkLock(address)`
+10. Approve tokens transfer to the `StakingEscrow` contract. These tokens are future mining rewards
```
cygnusv:
```suggestion
10. Approve tokens transfer to the `StakingEscrow` contract. These tokens are future staking rewards
```

docs/source/architecture/contracts.md
```diff
@@ -21,14 +21,14 @@ For a guide of how to deploy these contracts automatically, see the [Deployment
...
+11. Run the `initialize(uint256)` method to initialize the `StakingEscrow` contract
```
cygnusv: This is a goal (#1532), but it's not currently that way. I'm working on it. I'm OK if we leave it as it is for the moment, though.

docs/source/architecture/contracts.md
```diff
@@ -41,22 +41,24 @@ For a guide of how to deploy these contracts automatically, see the [Deployment
...
+* Start index in stakers array 
+* Maximum number of stakers
 This method will return only active stakers.
 
-In order to place the fee for a policy, Alice calls the method `PolicyManager.createPolicy(bytes16, uint16, uint256, address[])`,
-specifying the staker's addresses, the policy ID (off-chain generation), the policy duration in periods, and the first period's reward.
-Payment should be added to the transaction in ETH and the amount is `firstReward * stakers.length + rewardRate * periods * stakers.length`.
-The reward rate must be greater than or equal to the minimum reward for each staker in the list. The first period's reward is not refundable, and can be zero.
+In order to place the fee for a policy, Alice calls the method `PolicyManager.createPolicy(bytes16, address, uint64, address[])`,
+specifying the stakers' addresses, the policy ID (off-chain generation), the policy owner (could be zero address), and the end timestamp of the policy.
```
cygnusv: The `address` argument before the policy ID is missing.

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -25,16 +26,21 @@ contract PolicyManager is Upgradeable {
 
     event PolicyCreated(
         bytes16 indexed policyId,
-        address indexed client
+        address indexed creator,
```
cygnusv: Rather than `creator`, it should be something like `sponsor`, `payer`, etc.

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -374,42 +449,83 @@ contract PolicyManager is Upgradeable {
...
 
     /**
-    * @notice Refund part of fee by client
+    * @notice Get unsigned hash for revocation
+    * @param _policyId Policy id
+    * @param _node Node that will be excluded
+    * @return Revocation hash
+    */
+    function getRevocationHash(bytes16 _policyId, address _node) public view returns (bytes32) {
+        return SignatureVerifier.hashEIP191(abi.encodePacked(_policyId, _node), byte(0x45));
```
cygnusv: I think this line deserves a comment explaining we're using EIP191 version `0x45 ('E')` for `personal_sign`

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -1178,11 +1186,6 @@ contract StakingEscrow is Issuer {
         bytes32 memoryAddress = delegateGetData(_target, "stakerInfo(address)", 1, _staker, 0);
         assembly {
             result := memoryAddress
-            // copy data to the right position after the array pointer place
-            // measureWork
-            mstore(add(memoryAddress, 0x140), mload(add(memoryAddress, 0x100)))
-            // completedWork
-            mstore(add(memoryAddress, 0x160), mload(add(memoryAddress, 0x120)))
```
cygnusv: 🎉

tests/cli/ursula/test_stake_via_allocation_contract.py
```diff
@@ -464,7 +464,9 @@ def test_collect_rewards_integration(click_runner,
     blockchain_alice.selection_buffer = 1
 
     M, N = 1, 1
-    expiration = maya.now() + datetime.timedelta(days=3)
+    days = 3
+    now = testerchain.w3.eth.getBlock(block_identifier='latest').timestamp
+    expiration = maya.MayaDT(now).add(days=days-1)
```
cygnusv: This whole manually subtracting/adding a day is not intuitive at all.

michwill: How about `expiration = maya.now() - timedelta(days=1)`?

vzotova: how about to rename `days` to `days_to_pay` or to something similar?

nucypher/blockchain/eth/actors.py
jMyles: I'm sorry, maybe I'm being thick here, but I still don't get it.  How does subtracting one second from the `MayaDT` get us to the last second of the current period?  And how does `start_of_period=True` relate to that?

vzotova: maybe comment is misleading, current period will be only for case when `duration_periods==1`, sort of example.
In that case `start_of_period=True` will return zero/first second of the next period and subtracting 1 second we will come to the last second of the previous period. 
If `duration_periods==2` then `start_of_period=True` will be period after next period and `expiration` after subtracting will be last second of the next period
`start_of_period=True` is sort of rounding, returns the beggining of calculated period

docs/source/architecture/contracts.md
vzotova: ✔️

docs/source/architecture/contracts.md
vzotova: ✔️

docs/source/architecture/contracts.md
vzotova: I wanted to keep contract docs as ideal picture how to work with, so I'm for leaving

docs/source/architecture/contracts.md
vzotova: Not sure about `address`, maybe question with order in description? I've put stakers addresses to the end of the sentence

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
vzotova: ✔️

tests/cli/ursula/test_stake_via_allocation_contract.py
vzotova: I think I need to explain it better what is core of these changes:
Before this PR duration of policy was equal to periods for which Alice pays. Also was `firstPartialReward`, kind of alternative way to pay for current period.
Now duration in seconds and not in periods. But payment still in periods. Sort of border between Alice and staker/worker regarding fees. Also duration includes current period by default (start timestamp is `block.timestamp`), so we get payment periods == duration periods + 1 (I mean duration periods as ceiling of `(endTimestamp - block.timestamp)/secondsPerPeriod`). Of course it's possible that  payment periods == duration periods in case when `block.timestamp` == first second of current period and end timestamp == last second of any period but this situation is not used in tests

PR #1462, opened 2019-11-18
Self-Availability Checks; Ursula Rescheduling

### LINKED ISSUE #1416: Raise error in Ursula if the rest_server isn't reachable
Related issue is #1414.

When an Ursula is `is_me`, there's a very simple check we that can be done to prevent Ursulas, who aren't visible to the rest of the network, from getting online.

After the `ProxyRESTServer` for an Ursula is created and running, we should call a method that performs a request on the `rest_host` and `rest_port`. If the connection is refused, then we should raise an error like `UnreachableNodeError` or something. This prevents stakers from running Ursula nodes that aren't reachable to the rest of the network.

Furthermore, this check should be made as a looping call to check at a set interval (perhaps once per day). If the check fails, the node should shutdown.

Steps to close this issue:

1. Write a method on Ursula (or somewhere else more appropriate, if applicable) that accepts an IP and a port as input. The method should make a request to the given IP and port. If the request connection is refused or fails, it should raise an error.
2. Call this method _after_ the `self.rest_server` is created in `nucypher.characters.lawful.Ursula` on L876.
3. Register this method as a looping call for a specific interval. If the error is raised, it should cause a node to shutdown.
4. Catch the `UnreachableNodeError` in the CLI where appropriate to improve UX for stakers.

KPrasch:
Fixes #1416


nucypher/network/protocols.py
```diff
@@ -89,3 +93,140 @@ def __radd__(self, other):
...
+    def score(self) -> float:
+        failed_records = list(record.result for record in self._records if not record.result)
+        return len(failed_records) / self.retention
+
+    def issue_warnings(self) -> None:
+        for threshold, action in self.warnings.items():
+            if self.score >= threshold:
+                action()
+
+    def sample(self, quantity: int) -> list:
```
michwill: Fine for now but... Is it possible to use Alice's sample?

Though it requires some thinking, so let's start with this, simple one

nucypher/characters/lawful.py
```diff
@@ -942,6 +951,25 @@ def __init__(self,
...
+            worker: bool = True):
+
+        """Schedule Async Loops"""
+
+        if learning:
+            self.start_learning_loop(now=True)
+        if availability:
+            self._availability_sensor.start(now=False)  # wait...
+        if worker and not self.federated_only:
+            self.work_tracker.start(act_now=True)
```
michwill: Is it possible for the work tracker to actually check what the availability sensor shows (if the availability sensor exists at all), and confirm activity only if the sensor shows that everything is fine?

nucypher/network/protocols.py
```diff
@@ -89,3 +94,149 @@ def __radd__(self, other):
...
+    def severe_warning(self, reason=None) -> None:
+        self.log.warn(f'{self._ursula.rest_url} is unreachable')
+        try:
+            if reason:
+                raise reason(reason.message)
+            raise self.Unreachable(f'{self._ursula} is unreachable.')
+        finally:
+            self.shutdown_everything()
+
+    def shutdown_everything(self):
```
tuxxy: Insert the control rods!

tuxxy:
What's left to change this from a concept to a draft/WIP?

KPrasch:
> What's left to change this from a concept to a draft/WIP?

Im still working on a way to implement some basic tests here to weed out remaining design flaws; It will also need some testing on an experimental worker in the wild.

nucypher/characters/lawful.py
KPrasch: Okay, as a start - I pushed an update that allows for the worker to accept a confirmation requirement condition based on the sensor's status.

nucypher/network/protocols.py
KPrasch: Yes, it's possible - What do you think are the advantages?

michwill: Just a more objective measurement. But I think, the advantages are diminishingly small, as I'm thinking now

michwill: They have a minimum stake presumably anyway, that's good enough

michwill: Do they have the minimum stake though? Or could they be stakeless nodes?
If they can be stakeless, someone can spin up a bunch of nodes w/o NU and shut down the whole network by always responding that every node is down

nucypher/blockchain/eth/token.py
```diff
@@ -559,6 +575,12 @@ def _do_work(self) -> None:
         if interval > 0:
             self.log.warn(f"MISSED CONFIRMATIONS - {interval} missed staking confirmations detected.")
 
+        # Only perform work this round if the requirements are met
+        if not self.__check_work_requirement():
```
michwill: Yeah, great!

What is the value of `__check_work_requirement()` when the node just starts, how quickly does it turn `True`?

michwill:
```python
measurement = 0
N = 0
threshold = 5
...
while True:
    current = ask_random_node_if_I_am_online()
    # current can be 1, 0, or None if the feature is not supported
    if current is not None:
        N += 1
        measurment = current + 0.9 * measurement
    ...
    if N > 20:
        if measurement < threshold:
            scream()
            fail()
            burn() 
```

michwill:
Will we be able to roll this up at incentivized testnet, or should we introduce that as an upgrade, and make sure that people can do the upgrade well (having that as a part of the incentivized "challenge")?

nucypher/network/sensors.py
```diff
@@ -0,0 +1,186 @@
...
+
+    def handle_measurement_errors(self, *args, **kwargs) -> None:
+        failure = args[0]
+        cleaned_traceback = failure.getTraceback().replace('{', '').replace('}', '')  # FIXME: Amazing.
+        self.log.warn("Unhandled error during availability check: {}".format(cleaned_traceback))
+        failure.raiseException()
+
+    def status(self) -> bool:
+        """Returns current indication of availability"""
+        return self.score > self.SENSOR_SENSITIVITY
```
derekpierre:
```suggestion
        return self.score > (self.SENSOR_SENSITIVITY * self.MAXIMUM_SCORE)
```

nucypher/network/sensors.py
```diff
@@ -0,0 +1,186 @@
...
+            d = self.__task.start(interval=self.FAST_INTERVAL, now=now)
+            d.addErrback(self.handle_measurement_errors)
+
+    def stop(self) -> None:
+        if self.running:
+            self.__task.stop()
+
+    def maintain(self) -> None:
+        self.log.debug(f"Measuring node availability current score is {self.score}.")
+        known_nodes_is_smaller_than_sample_size = len(self._ursula.known_nodes) < self.SAMPLE_SIZE
```
derekpierre: Does a sample size of 1 make sense? Is this statement likely, since the default teacher node is our own? I'm sure that I'm missing something here - just need a little clarification

nucypher/network/sensors.py
```diff
@@ -0,0 +1,186 @@
...
+    def score(self) -> float:
+        return self.__score
+
+    def record(self, result: bool = None) -> None:
+        """Score the result and cache it."""
+        if result is None:
+            return
+        score = int(result) + self.CHARGE_RATE * self.__score
+        if score >= self.MAXIMUM_SCORE:
+            return
```
derekpierre: Can we just set score as MAXIMUM_SCORE or do you leave the original score (higher than MAXIMUM_SCORE) for the smoothing out of subsequent scores over time?

tests/network/test_availability.py
```diff
@@ -0,0 +1,56 @@
+import maya
```
derekpierre: We could also verify `ursula._availability_sensor.status()` as part of the tests. See my code suggestion for the bug in the `status()` method

nucypher/network/sensors.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/token.py
```diff
@@ -550,6 +568,12 @@ def _do_work(self) -> None:
             # TODO: #1516 Follow-up actions for downtime
             self.log.warn(f"MISSED CONFIRMATIONS - {interval} missed staking confirmations detected.")
 
+        # Only perform work this round if the requirements are met
+        if not self.__check_work_requirement():
+            self.log.warn(f'CONFIRMATION PREVENTED - There are unmet confirmation requirements.')
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/network/sensors.py
```diff
@@ -0,0 +1,196 @@
...
+    CHARGE_RATE = 0.9          # Measurement Multiplier
+
+    class Unreachable(RuntimeError):
+        pass
+
+    class Solitary(Unreachable):
+        message = "Cannot connect to any teacher nodes."
+
+    class Lonely(Unreachable):
+        message = "Cannot connect to enough teacher nodes."
```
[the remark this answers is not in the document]
KPrasch: Lonely is not *enough* nodes for an operation.  Solitary is truly zero.

nucypher/network/protocols.py
KPrasch: We're depending on the LL to verify staking here.

nucypher/blockchain/eth/token.py
KPrasch: Fast, and assumes 10 at startup.

nucypher/network/sensors.py
KPrasch: This is just a simple optimization mechanism, anticipating some nodes being incompatible.

nucypher/network/sensors.py
KPrasch: :heavy_check_mark:

tests/network/test_availability.py
KPrasch: :heavy_check_mark:

nucypher/network/middleware.py
```diff
@@ -234,6 +238,22 @@ def send_work_order_payload_to_ursula(self, work_order):
             path=f"kFrag/{id_as_hex}/reencrypt",
             data=payload, timeout=2)
 
+    def node_information(self, host, port, certificate_filepath=None):
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/network/middleware.py
```diff
@@ -234,6 +238,22 @@ def send_work_order_payload_to_ursula(self, work_order):
...
+        response = self.client.get(host=host, port=port,
+                                   path="public_information",
+                                   timeout=2,
+                                   certificate_filepath=certificate_filepath)
+        return response.content
+
+    def check_rest_availability(self, requesting_ursula, responding_ursula, certificate_filepath=None):
+        response = self.client.post(host=responding_ursula.rest_interface.host,
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/network/server.py
```diff
@@ -121,6 +123,26 @@ def public_information():
 
         return response
 
+    @rest_app.route("/ping")
+    def ping():
+        """
+        Returns network information about the accessor's connection to
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

tests/network/test_availability.py
```diff
@@ -0,0 +1,109 @@
...
+    def maintain():
+        sensor = ursula._availability_sensor
+        sensor.maintain()
+
+        # The node goes offline for some time...
+        for _ in range(10):
+            ursula._availability_sensor.record(False, reason={'error': 'fake failure reason'})
+
+        assert sensor.score < 4
+
```
derekpierre: just for testing purposes, we could also include calls to sensor.status() next to checking of score values. eg
```suggestion
        assert not sensor.status()
        
```
in this case `4` which is less than the sensitivity threshold, hence false.

Alternatively, to be more flexible (if say the threshold changes later on) you could do the following:
```suggestion
        assert sensor.status() == (sensor.score > (sensor.SENSOR_SENSITIVITY * sensor.MAXIMUM_SCORE))
        
```

This can be repeated in other spots as well.

tests/network/test_availability.py
```diff
@@ -0,0 +1,109 @@
...
+        # to keep this test fast, were just checking for a single entry
+        # (technically there will be 10, but resolution is one second.)
+        assert len(sensor._AvailabilitySensor__excuses) > 0
+
+    def raise_to_maximum():
+        for i in range(150):
+            ursula._availability_sensor.record(True)
+        assert ursula._availability_sensor.score > 9.98
+
```
derekpierre: Another example of what was in my comment above.

```suggestion
        assert ursula._availability_sensor.status()
        
```

KPrasch:
@jMyles - Are you still experiencing unexpected blocking behavior?

scripts/circle/docker-compose.yml
```diff
@@ -41,7 +41,7 @@ services:
       - 11500
     image: circle:nucypher
     depends_on:
-      - circleusula1
+      - circleusula2
```
vepkenez: I like that you maintained this spelling mistake.

nucypher/network/server.py
```diff
@@ -121,6 +124,49 @@ def public_information():
...
+        #
+        # Make a Sandwich
+        #
+
+        try:
+            # Fetch and store requester's teacher certificate.
+            certificate = this_node.network_middleware.get_certificate(host=requesting_ursula_address,
+                                                                       port=requesting_ursula_port)
+            certificate_filepath = this_node.node_storage.store_node_certificate(certificate=certificate)
+            result = requests.get(f"https://{requesting_ursula.rest_interface}/public_information",
```
jMyles: We want to use middleware here, right?

jMyles:
Following up on some Discord discussion here: can somebody try to duplicate:

1) inconsistent reentry in `ping`
2) error when trying to run in pycharm thread visualizer

nucypher/network/server.py
KPrasch: :heavy_check_mark:

tests/network/test_availability.py
KPrasch: :heavy_check_mark:

nucypher/network/sensors.py
```diff
@@ -0,0 +1,196 @@
...
+        return ursulas
+
+    @property
+    def score(self) -> float:
+        return self.__score
+
+    def record(self, result: bool = None, reason: dict = None) -> None:
+        """Score the result and cache it."""
+        if (not result) and reason:
+            self.__excuses[maya.now().epoch] = reason
```
derekpierre: @KPrasch - do we reduce the score somewhere when there is a problem? I looked through the PR some time ago - just rechecking.

nucypher/network/sensors.py
```diff
@@ -0,0 +1,196 @@
...
+        if result is None:
+            return  # Actually nevermind, dont score this one...
+        score = int(result) + self.CHARGE_RATE * self.__score
```
derekpierre: @KPrasch So if result is 'false' then `score = 0 + 0.9 * <previous_score>" - so good.

However, do we ever actually hit 0 for a shutdown, or we just get very close to it over time?

Issue #1463, opened 2019-11-19
JSONDecodeError in test_cli_lifecycle


cygnusv:
There are some tests that sometimes, at random, fail due to a `JSONDecodeError`. The problem is that that piece of code is expacting a JSON string which is the output of a json-rpc call, but it's also getting a stdout output like if it were using the `--debug` flag.

Example: https://circleci.com/gh/nucypher/nucypher/56027#tests/containers/5

```
self = <json.decoder.JSONDecoder object at 0x7f9060df2cd0>
s = 'Unlocking keyring.\nFinished unlocking.\nStarting Learning Loop.\nFinished learning about all seednodes.\n\n\n    / \...-11-19_11-23-05.009732/known_nodes/metadata\nOK | Responded to IPC request #1574162584 with None bytes, took 0:00:40\n'
idx = 0

    def raw_decode(self, s, idx=0):
        """Decode a JSON document from ``s`` (a ``str`` beginning with
        a JSON document) and return a 2-tuple of the Python
        representation and the index in ``s`` where the document ended.
    
        This can be used to decode a JSON document from a string that may
        have extraneous data at the end.
    
        """
        try:
            obj, end = self.scan_once(s, idx)
        except StopIteration as err:
>           raise JSONDecodeError("Expecting value", s, err.value) from None
E           json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

I have no idea why is this happening. Perhaps some misconfiguration of our logging (and its associated cache) during tests.

michwill:
Perhaps some logging is going to stdout instead of stderr?

cygnusv:
BTW, cannot reproduce locally. So far, I only encounter this on CircleCI.

KPrasch:
Digging a bit deeper in the stack trace there is a `Unhandled error during node learning.`

I am able to reproduce this by running this group of tests: `pytest tests/cli/ursula/ tests/cli/test_cli_lifecycle.py `

```
Unhandled error during node learning: Traceback (most recent call last):
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.6/site-packages/twisted/internet/base.py", line 1283, in run
    self.mainLoop()
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.6/site-packages/twisted/internet/base.py", line 1292, in mainLoop
    self.runUntilCurrent()
[...]

PR #1467, opened 2019-11-20
Bump version: 1.1.0-beta.0 → 1.2.0-beta.1

KPrasch:


KPrasch:
Blocked by <https://github.com/nucypher/nucypher/issues/1463> and <https://github.com/nucypher/nucypher/issues/1469>

cygnusv:
> Blocked by #1463 and #1469

Is it a reason to block? I recall seen this bug for at least weeks, and it seems we've only observed it in CI: not locally and not in the wild.

KPrasch:
Well - it will block CI - Unless we want to skip this test for now...

PR #1470, opened 2019-11-22
Versioning of contracts and upgradability test

### LINKED ISSUE #1464: Versioning of contracts
Add way to set version for contract and have contracts with different versions deployable in one `DeployerInterface`.
Idea:
Set in `@dev` doc for the contract

### LINKED ISSUE #1465: Test for contracts upgradability between master and PRs

vzotova:
Closes #1464
Closes #1465 

* Version for upgradeable contracts inside contract doc (`@dev`), for example `|v1.2.3|`
* Multi-source solidity compilation
* Extracting version inside `SolidityCompiler`
* Storing version in contract registry
* Using version while deployment
* Test for upgrading possibility between current commit and master

Version format is `|vi.j.k|`, where `i` - fork (I hope will be always `1`), `j` - upgradeable signature/state changes, `k` - small changes inside existent function

### DIFF
--- a/docs/source/architecture/contracts.md
+++ b/docs/source/architecture/contracts.md
@@ -126,5 +126,5 @@ When the user wants to become a staker - they use the `PreallocationEscrow` cont
 Upgradeable contracts, such as `Adjudicator`, `StakingEscrow`, `PolicyManager` and `StakingInterface`, have their version specified in contract doc inside @dev.  
 Version format is `|vi.j.k|`, where `i` - major version, `j` - minor version, `k` - patch, for example `|v1.2.3|`:  
 * Different major versions mean different forks and they are not upgradeable
-* Minor versions relate to any signatures or state changes inside contract, contracts are upgradeable between minor versions, but have different ABI and as follows different agent layers 
-* Patch is changes inside function(s) with signature untouched. All patches inside one minor version can be upgraded from one to another without other changes
+* Minor versions relate to any signatures or state changes inside contract, contracts are upgradeable between minor versions, but have different ABI and follows different agent layers 
+* Patches involve changes inside function(s) with signature(s) untouched. All patches with a common minor version can be upgraded from one to another without other changes
--- a/docs/source/architecture/contracts.md
+++ b/docs/source/architecture/contracts.md
@@ -122,3 +122,9 @@ All tokens will be unlocked after a specified time and the user can retrieve the
 When the user wants to become a staker - they use the `PreallocationEscrow` contract as a proxy for the `StakingEscrow` and `PolicyManager` contracts.
 
 
+## Contracts Versioning
+Upgradeable contracts, such as `Adjudicator`, `StakingEscrow`, `PolicyManager` and `StakingInterface`, have their version specified in contract doc inside @dev.  
+Version format is `|vi.j.k|`, where `i` - major version, `j` - minor version, `k` - patch, for example `|v1.2.3|`:  
+* Different major versions mean different forks and they are not upgradeable
+* Minor versions relate to any signatures or state changes inside contract, contracts are upgradeable between minor versions, but have different ABI and as follows different agent layers 
+* Patch is changes inside function(s) with signature untouched. All patches inside one minor version can be upgraded from one to another without other changes
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -154,7 +154,7 @@ class BaseContractDeployer:
             raise self.ContractDeploymentError(message)
         return True
 
-    def deploy(self, gas_limit: int, progress) -> dict:
+    def deploy(self, gas_limit: int, progress, **overrides) -> dict:
         """
         Provides for the setup, deployment, and initialization of ethereum smart contracts.
         Emits the configured blockchain network transactions for single contract instance publication.
@@ -361,7 +361,7 @@ class NucypherTokenDeployer(BaseContractDeployer):
     _upgradeable = False
     _ownable = False
 
-    def deploy(self, gas_limit: int = None, progress=None) -> dict:
+    def deploy(self, gas_limit: int = None, progress=None, **overrides) -> dict:
         """
         Deploy and publish the NuCypher Token contract
         to the blockchain network specified in self.blockchain.network.
@@ -371,11 +371,13 @@ class NucypherTokenDeployer(BaseContractDeployer):
         self.check_deployment_readiness()
 
         # Order-sensitive!
+        constructor_kwargs = {"_totalSupply": self.economics.erc20_total_supply}
+        constructor_kwargs.update({k: v for k, v in overrides.items() if v is not None})
         contract, deployment_receipt = self.blockchain.deploy_contract(self.deployer_address,
                                                                        self.registry,
                                                                        self.contract_name,
                                                                        gas_limit=gas_limit,
-                                                                       _totalSupply=self.economics.erc20_total_supply)
+                                                                       **constructor_kwargs)
         if progress:
             progress.update(1)
         self._contract = contract
@@ -465,14 +467,27 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
         if result == self.blockchain.NULL_ADDRESS:
             raise RuntimeError("PolicyManager contract is not initialized.")
 
-    def _deploy_essential(self, gas_limit: int = None):
-        escrow_constructor_args = (self.token_contract.address, *self.economics.staking_deployment_parameters)
+    def _deploy_essential(self, gas_limit: int = None, **overrides):
+        args = self.economics.staking_deployment_parameters
+        constructor_kwargs = {
+            "_hoursPerPeriod": args[0],
+            "_miningCoefficient": args[1],
+            "_lockedPeriodsCoefficient": args[2],
+            "_rewardedPeriods": args[3],
+            "_minLockedPeriods": args[4],
+            "_minAllowableLockedTokens": args[5],
+            "_maxAllowableLockedTokens": args[6],
+            "_minWorkerPeriods": args[7]
+        }
+        constructor_kwargs.update({k: v for k, v in overrides.items() if v is not None})
+        # Force use of the token address from the registry
+        constructor_kwargs.update({"_token": self.token_contract.address})
         the_escrow_contract, deploy_receipt = self.blockchain.deploy_contract(
             self.deployer_address,
             self.registry,
             self.contract_name,
             gas_limit=gas_limit,
-            *escrow_constructor_args,
+            **constructor_kwargs
         )
 
         return the_escrow_contract, deploy_receipt
@@ -481,7 +496,8 @@ class StakingEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin, Owna
                initial_deployment: bool = True,
                secret_hash: bytes = None,
                gas_limit: int = None,
-               progress=None
+               progress=None,
+               **overrides
                ) -> dict:
         """
         Deploy and publish the StakingEscrow contract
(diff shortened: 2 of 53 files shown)


nucypher/blockchain/eth/sol/compile.py
```diff
@@ -74,20 +92,43 @@ def install_compiler(self, version: str=None):
...
+        interfaces = dict()
+        for root_source_dir, other_source_dirs in self.source_dirs:
+            if root_source_dir is None:
+                self.log.warn("One of the root directories is None")
+                continue
+
+            raw_interfaces = self._compile(root_source_dir, other_source_dirs)
+            for name, data in raw_interfaces.items():
+                # Extract contract version from docs
+                version_search = re.search(r'\"details\":\".*?\|(v\d+\.\d+\.\d+)\|.*?\"', data['devdoc'])
```
KPrasch: Can you expand this regex into the verbose format?  

Example:

```
p = re.compile(r"""

(\+\d{1,2})?               # Optional Country Code +1 or perhaps +22
(-|\()?                    # Optional hyphen or open parenths for the area code.
(\d{10}                    # Try and match ten digits exactly
|                          # Else
(\d{3}                     # Match three digits
[\-|\)]{0,2})              # Hypen or close parenths zero through 2 times (to account for -))  
{2}                        # Repeat to match the next three digits
\d{4})                     # The remaining four digits
$                          # Ensure nothing else follows 


""", re.VERBOSE)

```

nucypher/blockchain/eth/sol/source/contracts/staking_contracts/StakingInterface.sol
```diff
@@ -168,7 +169,7 @@ contract StakingInterface {
     * @notice Prolong active sub stake
     * @param _index Index of the sub stake
     * @param _periods Amount of periods for extending sub stake
-    **/
+    */
```
KPrasch: Hahah :joy:

nucypher/blockchain/eth/sol/compile.py
vzotova: ✔️

nucypher/blockchain/eth/sol/compile.py
```diff
@@ -105,7 +105,22 @@ def compile(self) -> dict:
             raw_interfaces = self._compile(root_source_dir, other_source_dirs)
             for name, data in raw_interfaces.items():
                 # Extract contract version from docs
-                version_search = re.search(r'\"details\":\".*?\|(v\d+\.\d+\.\d+)\|.*?\"', data['devdoc'])
+                version_search = re.search(r"""
```
KPrasch: :+1: Badass

tests/blockchain/eth/contracts/integration/test_contracts_upgradability.py
```diff
@@ -0,0 +1,144 @@
+"""
```
KPrasch: :rocket:  - This is a relief, good work.  I think this test can be emphasized in our CI workflow by running it in its own job.

tests/blockchain/eth/entities/deployers/test_interdeployer_integration.py
```diff
@@ -97,10 +97,10 @@ def test_deploy_ethereum_contracts(testerchain,
 
     with pytest.raises(BaseContractDeployer.ContractDeploymentError):
         assert policy_manager_deployer.contract_address is constants.CONTRACT_NOT_DEPLOYED
-    assert not policy_manager_deployer.is_deployed
+    assert not policy_manager_deployer.is_deployed()
```
KPrasch: Is `PolicyManagerDeployer.is_deployed` no longer a @property method?

vzotova: I added version as parameter to this method

nucypher/cli/deploy.py
```diff
@@ -116,12 +116,13 @@ def inspect(provider_uri, config_root, registry_infile, deployer_address, poa):
 @click.option('--retarget', '-d', help="Retarget a contract's proxy.", is_flag=True)
 @click.option('--target-address', help="Recipient's checksum address for token or ownership transference.",
               type=EIP55_CHECKSUM_ADDRESS)
+@click.option('--ignore-deployed', help="Ignore already deployed contracts if exist.", is_flag=True)
```
KPrasch: return of "bare" mode? I like your name better.

tests/blockchain/eth/interfaces/test_solidity_compiler.py
```diff
@@ -14,14 +14,54 @@
...
+
+
+def test_multi_source_compilation(testerchain):
+    solidity_compiler = SolidityCompiler(source_dirs=[
+        (SolidityCompiler.default_contract_dir(), None),
+        (SolidityCompiler.default_contract_dir(), {TesterBlockchain.TEST_CONTRACTS_DIR})
+    ])
+    interfaces = solidity_compiler.compile()
+
+    # Remove AST because id in tree node depends on compilation scope
```
KPrasch: Woah - can you expand a bit on this?

vzotova: Each element in AST has ID field. This value depends on scope of compiled files (all contracts). First source for SolidityCompiler includes only main contracts, second - also tests. Merging two results lead that main contracts has AST from first compilation, test contracts from the second. Interface has contracts with the same AST as in second compilation for all contracts. We don't use AST yet and main part for comparing is bytecode, so I remove it

nucypher/cli/deploy.py
vzotova: It's related to "bare" but much simpler, turns off 'is_deployed' check to redeploy same contract multiple times (first of all for test)

tests/blockchain/eth/contracts/integration/test_contracts_upgradability.py
vzotova: ✔️

docs/source/architecture/contracts.md
```diff
@@ -122,3 +122,9 @@ All tokens will be unlocked after a specified time and the user can retrieve the
 When the user wants to become a staker - they use the `PreallocationEscrow` contract as a proxy for the `StakingEscrow` and `PolicyManager` contracts.
 
 
+## Contracts Versioning
+Upgradeable contracts, such as `Adjudicator`, `StakingEscrow`, `PolicyManager` and `StakingInterface`, have their version specified in contract doc inside @dev.  
+Version format is `|vi.j.k|`, where `i` - major version, `j` - minor version, `k` - patch, for example `|v1.2.3|`:  
+* Different major versions mean different forks and they are not upgradeable
+* Minor versions relate to any signatures or state changes inside contract, contracts are upgradeable between minor versions, but have different ABI and as follows different agent layers 
+* Patch is changes inside function(s) with signature untouched. All patches inside one minor version can be upgraded from one to another without other changes
```
derekpierre:
```suggestion
* Patches involve changes inside function(s) with signature(s) untouched. All patches with a common minor version can be upgraded from one to another without other changes
```

nucypher/cli/deploy.py
derekpierre: Just for my own understanding: is this flag essentially a `force` type option to deploy the contract even if previously deployed?

vzotova: Correct, one remark - `force` for the same contract with the same version

derekpierre: 👍

nucypher/blockchain/eth/interfaces.py
```diff
@@ -606,39 +615,80 @@ def deploy_contract(self,
...
+            major, minor, patch = [int(v) for v in version[1:].split(".", 3)]
+            if current_version_parsed[0] == -1 or \
+                requested_version == 'latest' and \
+                (major > current_version_parsed[0] or
+                 major == current_version_parsed[0] and minor > current_version_parsed[1] or
+                 major == current_version_parsed[0] and minor == current_version_parsed[1] and patch > current_version_parsed[2]) or \
+                requested_version == 'earliest' and \
+                (major < current_version_parsed[0] or
+                 major == current_version_parsed[0] and minor < current_version_parsed[1] or
+                 major == current_version_parsed[0] and minor == current_version_parsed[1] and patch < current_version_parsed[2]):
```
cygnusv: Wow, a 9-line condition. This is though!
I think this part needs some touches to improve readability. As far as I can tell, you're trying to check lexicographical ordering between tuples, which [Python implements already](https://docs.python.org/3/tutorial/datastructures.html#comparing-sequences-and-other-types). So, e.g., lines 667 to 669 can be changed to `(major, minor, patch) > current_version_parsed`; similarly, lines 671 to 673 can be changed to `(major, minor, patch) < current_version_parsed`

nucypher/cli/deploy.py
cygnusv: Does this have a use beyond tests?

nucypher/blockchain/eth/interfaces.py
derekpierre: @cygnusv That's pretty cool - didn't know you could do that.

vzotova: Thanks, really cool suggestion. 
✔️

nucypher/cli/deploy.py
vzotova: I see at least one use: redeploying same contract but with different constructor parameters, for example, disable/enable test mode of StakingEscrow (for worklock)

cygnusv: But are we going to do that on a real deployment? because as far as I understand, the test mode is only for tests. Or are you planning on trying that on testnet?

vzotova: test mode for testnet, to change worklock

.circleci/config.yml
```diff
@@ -456,7 +463,18 @@ jobs:
...
+
+  contracts_upgradeability:
+    <<: *python_37_base
+    parallelism: 1
+    steps:
+      - prepare_environment
+      - run:
+          name: Ethereum Contract Upgradeability Test
+          command: |
+            pytest $(circleci tests glob "tests/blockchain/eth/contracts/test_contracts_upgradeability.py" | circleci tests split --split-by=timings)
```
cygnusv: After checking the performance of the CI build, I think having a separate job for this test is unnecessary. It blocks a CI box for some time, while if it were included in the `contracts` job, it will probably be executed without any overhead due to current parallelism.

nucypher/blockchain/eth/sol/compile.py
cygnusv: 🆒

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -10,7 +10,8 @@ import "contracts/lib/AdditionalMath.sol";
 
 /**
 * @notice Contract for calculate issued tokens
-**/
+* @dev |v1.1.2|
```
cygnusv: What about considering `v.1.1.1` for all contracts?

nucypher/blockchain/eth/deployers.py
```diff
@@ -105,15 +106,25 @@ def contract(self):
     def dispatcher(self):
         return self.__proxy_contract
 
-    @property
-    def is_deployed(self) -> bool:
-        return bool(self._contract is not CONTRACT_NOT_DEPLOYED)
-
```
cygnusv: Yes! I had a very similar change in a WIP PR. This property was just checking if the contract was deployed with that very same instance of the deployer, which of course is not very useful in use cases like upgrades or rollbacks. Glad this is changed :)

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
vzotova: I just tried to remember how many changes we did before, but really it doesn't matter because of testnet redeploy and new versions `v2.1.1`

.circleci/config.yml
vzotova: The main reason for extracting was readability of test results, like in 1459, where only one contract test failed, I like how it looks

cygnusv: Right, I understand, but we have a limited number of CI boxes and using one of them for a single test that runs for 21 seconds doesn't seem a good idea to me.

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
cygnusv: 👍

.circleci/config.yml
vzotova: makes sense, I'll change it back

vzotova: ✔️

PR #1480, opened 2019-11-27
Updates to PreallocationEscrow and allocation CLI operations

cygnusv:
* Allow preallocation users to collect tokens _from_ the preallocation contract. So far there only was tooling to collect rewards from `StakingEscrow` to the allocation contract, but not that last step.
* Fix a bug in `PreallocationEscrow` that prevented users to collect rewards if they weren't superior to the initial lock
* Add a copy of the registry to the local installation, and implement a fallback mechanism so we can use that if the Github registry is not accessible.


nucypher/blockchain/eth/sol/source/contracts/staking_contracts/PreallocationEscrow.sol
```diff
@@ -62,7 +73,18 @@ contract PreallocationEscrow is AbstractStakingContract, Ownable {
     * @param _value Amount of token to withdraw
     **/
     function withdrawTokens(uint256 _value) public onlyOwner {
-        require(token.balanceOf(address(this)).sub(getLockedTokens()) >= _value);
+        uint256 balance = token.balanceOf(address(this));
+        uint32 secondsPerPeriod = stakingEscrow.secondsPerPeriod();
+        uint16 currentPeriod = uint16(block.timestamp / secondsPerPeriod);
+        uint16 endLockPeriod = uint16(endLockTimestamp / secondsPerPeriod);
+        if (currentPeriod <= endLockPeriod) {
+            uint256 stakedTokens = stakingEscrow.getLockedTokens(address(this), endLockPeriod - currentPeriod);
```
vzotova: Maybe it's too strict, because only counted tokens that locked for whole period. How about `getAllTokens` function? it will return all tokens that belong to current `PreallocationEscrow` and will simplify this logic

nucypher/characters/lawful.py
```diff
@@ -1433,7 +1433,9 @@ def __init__(self,
             # If an initial address was passed,
             # it is safe to understand that it has already been used at a higher level.
             if initial_address not in self.wallet:
-                raise self.StakingWallet.UnknownAccount
+                message = f"Account {initial_address} is not known by this Ethereum client. Is it a HW account? " \
+                          f"If so, make sure your device is plugged and you use the --hw-wallet flag"
```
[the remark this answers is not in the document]
cygnusv: ✅

docs/source/guides/staking_guide.rst
```diff
@@ -418,19 +423,49 @@ As an alternative to the ``--allocation-filepath`` flag, preallocation users
 can directly specify their beneficiary and staking contract addresses with the
 ``--beneficiary-address ADDRESS`` and ``--staking-address ADDRESS``, respectively.
 
+Finally, note that collected staking rewards always end up in the original
```
derekpierre:
```suggestion
Finally, note that collected staking rewards are always placed in the original
```

nucypher/cli/deploy.py
```diff
@@ -72,8 +72,9 @@ def deploy():
 @deploy.command(name='download-registry')
 @click.option('--config-root', help="Custom configuration directory", type=click.Path())
 @click.option('--registry-outfile', help="Output path for contract registry file", type=click.Path(file_okay=True))
+@click.option('--network', help="", type=click.Choice(['goerli']), default='goerli')  # TODO: #1496
```
derekpierre: Is it possible to use `CanonicalRegistrySource.networks` values for the choice type, instead of needing to manually add to the choice when mainnet rolls around?

cygnusv: Ah, that's a great idea! Although strictly better to current approach, I don't know if when the moment comes we'd end up extracting the networks list to some other class. But yeah, in the meantime, let's do this.

nucypher/blockchain/eth/sol/source/contracts/staking_contracts/PreallocationEscrow.sol
vzotova: In the beginning `x` tokens belong to  `PreallocationEscrow` (PE). 
Next step is staking (transfer+lock), `x = y + z`, `y` -> PE, `z` -> locked in `StakingEscrow` (SE), `balance = y`, `stakedTokens = z`. 
After some time we get reward or some of `z` amount become unlocked, `x = y + t + v + w`, `y` -> PE, `t` -> locked in SE, `v` -> reward in SE, `w` -> unlocked in SE (we can track only sum `v + w`). Still `balance = y`, `stakedTokens = t` but `v + w` ignored. While this not locked tokens in SE they are not taken in calculation for PE, they become withdrawable only after transfering back . I think it's to strict and complex. We can use all amount `t + v + w` just calling `getAllTokens`, all this tokens belong to PE and can't be withdrawn except through PE so this's safe. Also this method much simplier then `getLockedTokens` (just returns `stakerInfo.value`) and not need to think about periods.

nucypher/cli/characters/stake.py
```diff
@@ -679,7 +674,7 @@ def preallocation(click_config,
     #############
     # Unauthenticated actions: status
 
-    if status:
+    if action == 'status':
```
KPrasch: :+1:

PR #1493, opened 2019-12-05
Ensure all tests can run in a singe session; Docker adjustments

KPrasch:
Fixes #1472 #1486  #945 #1229 , Unblocks progress on #1479 .

I completed `pytest` locally (and in docker) in just under 15 min. including slow tests.

PR #1494, opened 2019-12-05
[WIP] Connect to preferred seednodes at startup

### LINKED ISSUE #1201: --teacher is ignored when network is goerli or mainnet
Method `load_seednodes()` is ignoring the provided `teacher_uris`, since in L121, it directly overwrites them with `teacher_uris = TEACHER_NODES[domain]`, whose current values are:
```
TEACHER_NODES = {
    'goerli': [],
    'mainnet': [],
}
```
The net effect is that `--teacher TEACHER_URI` doesn't work.

What should be the intended behavior of `--teacher TEACHER_URI` when the network is mainnet or goerli? I think that it should consider whatever nodes are specified in `TEACHER_NODES` plus the `TEACHER_URI` provided.

KPrasch:
Fixes #1201 good.

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -1027,10 +1027,12 @@ class Ursula(Teacher, Character, Worker):
                          min_stake: int,
                          network_middleware: RestMiddleware = None,
                          registry: BaseContractRegistry = None,
+                         retry_attempts: int = 2,
+                         retry_interval: int = 2
                          ) -> 'Ursula':
 
-        def __attempt(attempt=1, interval=10) -> Ursula:
-            if attempt > 3:
+        def __attempt(attempt=1, interval=retry_interval) -> Ursula:
+            if attempt >= retry_attempts:
                 raise ConnectionRefusedError("Host {} Refused Connection".format(teacher_uri))
 
             try:
--- a/nucypher/cli/actions.py
+++ b/nucypher/cli/actions.py
@@ -48,6 +48,7 @@ from nucypher.cli import painting
 from nucypher.cli.types import IPV4_ADDRESS
 from nucypher.config.constants import DEFAULT_CONFIG_ROOT
 from nucypher.config.node import CharacterConfiguration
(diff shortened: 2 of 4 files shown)


nucypher/characters/lawful.py
```diff
@@ -1027,10 +1027,12 @@ def from_teacher_uri(cls,
                          min_stake: int,
                          network_middleware: RestMiddleware = None,
                          registry: BaseContractRegistry = None,
+                         retry_attempts: int = 2,
+                         retry_interval: int = 2
```
derekpierre: Any reason for the retry interval changing from 10 to 2? Just to make it faster?

Also the retry attempts went from 3 to 2...?

nucypher/cli/actions.py
```diff
@@ -138,11 +140,15 @@ def load_seednodes(emitter,
             teacher_uris.extend(seednode_uris)
```
derekpierre: Instead of the log message at L118, what do you think of adding the following here at L141
```suggestion

            emitter.message(f'Connecting to preferred seednodes for {domain}: {teacher_uris}', color='yellow')
```

nucypher/characters/lawful.py
KPrasch: Yeah, just in an effort to increase(decrease..?) tolerance against failed connections to hardcoded nodes and take a little less time blocking.  I *think* we want the user to not really notice if the hardcoded seeds are unreachable, and be allowed to proceed unobstructed... We do want to raise to bar to entry too high by enforcing preferred nodes too strictly.

Additional thoughts?

derekpierre: Makes sense - I guess testing will tell.

KPrasch:
This PR was unintentionally included in the merge of #1493 (one commit) - There was additional work I wanted to include, but Ill represent those as Issues instead.

KPrasch:
Follow-up https://github.com/nucypher/nucypher/issues/1528

Issue #567, opened 2018-11-20
Buckets.  We need Buckets.


jMyles:
Currently, we have just a single bucket for nodes - `known_nodes`. 

But we need to bucket them separately.  Nodes that are palpably offline don't belong in this bucket.  Nor do those whose metadata is out of date.  Or probably even those who have a known-but-out-of-date fleet state.

KPrasch:
Guess I better roll up my sleeves!  I'd like to work on this Issue this week since I anticipate it will need several iterations for a minimum viable solution.

cygnusv:
Moved to Mainnet month 1 milestone
```
dnunez Today at 5:36 PM
what if we go to mainnet without buckets? @jMyles @kprasch I don't immediately see it's urgent, but I admit I don't fully understand the extend os this issue
jMyles Today at 5:37 PM
@dnunez: It's an option worth considering.  It can be introduced later.  It makes me a little uncomfortable, but it won't be a complete disaster.
```

KPrasch:
Finally I'm going to close this issue since it has grown to be too broad and is partially fixed since nodes now track a verification state of each peer effectively creating two "buckets".  Any additional buckets represent enhancements and optimizations for stability/performance/accuracy.  Related to #355

PR #1520, opened 2019-12-08
Tolerate non-staking nodes in storages; Handle NotStaking in learning

### LINKED ISSUE #1519: Non-staking nodes in storage cause crash
When reading a node from persistent storage that is no longer staking, The uncaught exception `NotStaking` is raised leading to a crash and preventing nodes from starting up.

``` 
 File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/network/nodes.py", line 1059, in validate_worker
    raise self.NotStaking(f"Staker {self.checksum_address} is not staking")
nucypher.network.nodes.NotStaking: Staker 0x980fFE805d00E7C446E33C9b4240CDaE5697Bf4A is not staking
```

KPrasch:
Fixes #1519 - Patch to fix unhandled `NotStaking` exception raised in `Learner.remember_node`

Related to https://github.com/nucypher/nucypher/issues/567

### DIFF
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -440,11 +440,18 @@ class Learner:
                              network_middleware=self.network_middleware,
                              registry=self.registry)  # composed on character subclass, determines operating mode
         except SSLError:
-            return False  # TODO: Bucket this node as having bad TLS info - maybe it's an update that hasn't fully propagated?
+            # TODO: Bucket this node as having bad TLS info - maybe it's an update that hasn't fully propagated?
+            return False
 
         except NodeSeemsToBeDown:
             self.log.info("No Response while trying to verify node {}|{}".format(node.rest_interface, node))
-            return False  # TODO: Bucket this node as "ghost" or something: somebody else knows about it, but we can't get to it.
+            # TODO: Bucket this node as "ghost" or something: somebody else knows about it, but we can't get to it.
+            return False
+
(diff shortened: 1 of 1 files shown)


nucypher/network/nodes.py
```diff
@@ -440,11 +440,17 @@ def remember_node(self, node, force_verification_check=False, record_fleet_state
...
 
         except NodeSeemsToBeDown:
             self.log.info("No Response while trying to verify node {}|{}".format(node.rest_interface, node))
-            return False  # TODO: Bucket this node as "ghost" or something: somebody else knows about it, but we can't get to it.
+            # TODO: Bucket this node as "ghost" or something: somebody else knows about it, but we can't get to it.
+            return False
+
+        except node.NotStaking:
+            # TODO: Bucket this node as inactive, and potentially safe to forget.
+            self.log.info(f'Staker:Worker {node.checksum_address}:{node.worker_address} is not actively staking, skipping.')
```
derekpierre: Do you need to return False, so that the node isn't saved?

KPrasch: Yes - Good catch!

KPrasch: :heavy_check_mark:

KPrasch:
@michwill > So, no need to remove the node metadata when this is merged?

Nope - its back compatible.  This will tolerate even existing non-staking nodes on users disks.

PR #1533, opened 2019-12-12
Static Nodes from Disk

KPrasch:
Follow up for accidentally merged commits in #1494

### DIFF
--- a/nucypher/cli/actions.py
+++ b/nucypher/cli/actions.py
@@ -19,7 +19,7 @@ import json
 import os
 import shutil
 from json import JSONDecodeError
-from typing import List, Tuple
+from typing import List, Tuple, Dict, Set, Optional
 
 import click
 import requests
@@ -50,6 +50,7 @@ from nucypher.config.constants import DEFAULT_CONFIG_ROOT
 from nucypher.config.node import CharacterConfiguration
 from nucypher.network.exceptions import NodeSeemsToBeDown
 from nucypher.network.middleware import RestMiddleware
+from nucypher.network.nodes import Teacher
 from nucypher.network.teachers import TEACHER_NODES
 
 NO_BLOCKCHAIN_CONNECTION.bool_value(False)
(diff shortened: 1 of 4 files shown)


nucypher/cli/actions.py
```diff
@@ -108,6 +108,40 @@ def unlock_nucypher_keyring(emitter, password: str, character_configuration: Cha
...
+            data = file.read()
+    except FileNotFoundError:
+        # If there is no static nodes file, there are no results
+        return list()
+
+    # Decode
+    try:
+        static_nodes = json.loads(data)
+    except JSONDecodeError:
+        raise RuntimeError(f"Static nodes configuration at '{filepath}' contains invalid JSON.")
```
derekpierre: If you want, you could save some lines by doing:
```python
try:
    with open(filepath, 'r') as file:
        static_nodes = json.load(file)
except FileNotFoundError:
    return list()
except JSONDecodeError:
    raise RuntimeError(f"Static nodes configuration at '{filepath}' contains invalid JSON.")
```

nucypher/cli/actions.py
```diff
@@ -129,8 +173,13 @@ def load_seednodes(emitter,
         teacher_uris = list()
 
     for domain in network_domains:
+
+        # Static nodes from JSON file
+        static_nodes = load_static_nodes(domain=domain)
+        teacher_uris.extend(static_nodes)
+
+        # Hardcoded teachers from module
```
derekpierre: If network domains is ever > 1, isn't the `for` loop on line 192 re-adding already added teacher nodes for subsequent domains? If so, that's a big bug that gets worse the higher the number of domains. It feels like the loop on line 192 should be outside of the `for domain in network_domains:` loop (Line 175).

KPrasch: :heavy_check_mark:

nucypher/cli/actions.py
KPrasch: :heavy_check_mark: :+1:

nucypher/cli/actions.py
```diff
@@ -108,6 +109,47 @@ def unlock_nucypher_keyring(emitter, password: str, character_configuration: Cha
...
+    return filtered_static_nodes
+
+
+def aggregate_seednode_uris(domains: set, highest_priority: Optional[List[str]] = None) -> List[str]:
+
+    # Read from the disk
+    static_nodes = load_static_nodes(domains=domains)
+
+    # Priority 1 - URI passed via --teacher
+    uris = highest_priority or list()
```
derekpierre: Out of curiosity - should this be a set i.e. may not matter, but can a node be part of more than one domain?

KPrasch: No such thing as an "ordered set".  It's true that a valid code path includes a node twice, if its specified as part of two domains.  This may indicate that the overall approach can be better implemented by not storing static nodes in buckets of unique domains.

Instead of static-nodes.json :
```
STATIC_NODES = {'domain-name': 'https://uri:9151'}
```

We can do `static-nodes.conf`:
``` 
https://uri:9151 domian1
https://uri:9151 domian1, domain2, domain3
https://uri:9151 domian1, domain2
```

What do you think?

michwill: Can the same node be serving for two different domains?
Do domains mean different testnets, federated nets, mainnet?

derekpierre: You could do an ordered dict which, while redundant, acts as an ordered set. Something like:

```python
uris = OrderedDict()
...
uris['https://uri:9151'] = <some_value, eg. 'https://uri:9151'>
```

@KPrasch, your `static-nodes.conf` idea is good, but maybe a tad less efficient since all uris would need to be iterated over regardless of whether they apply to the set of domains or not...? wdyt?
Also, @michwill 's question above is pertinent - is this even a possible scenario?

KPrasch: > Can the same node be serving for two different domains?
Yes.

> Do domains mean different testnets, federated nets, mainnet? 
Yes, but not necessarily.  They are domains of node discovery.

KPrasch: > ... you could just leave the code as is...wdyt?

Yep - reasonable and allows for future iteration.

Also, @michwill 's question above is pertinent - is this even a possible scenario?

Yes - See above comment.

nucypher/cli/actions.py
```diff
@@ -108,6 +109,47 @@ def unlock_nucypher_keyring(emitter, password: str, character_configuration: Cha
         raise character_configuration.keyring.AuthenticationFailed
 
 
+def load_static_nodes(domains: Set[str], filepath: Optional[str] = None) -> Dict[str, 'Ursula']:
+    """
+    Non-invasively read teacher-uris from a JSON configuration file keyed by domain name.
```
mswilkison:
```suggestion
    Non-invasively read teacher-uris from a JSON configuration file keyed by domain name
```

KPrasch:
I have one additional follow-up commit to push before this can be merged, please!

Issue #1349, opened 2019-09-19
Enable re-staking by default


KPrasch:

cygnusv:
We have 3 options here:

**1. Leave Re-staking as False by default**
* This is our current status
* If user wants to restake, she has to use the CLI to do it.
* This implies an additional transaction, but that's for the user to decide.

**2. Leave Re-staking as False by default in the contract, but guide the user towards re-staking**
* When creating a stake, add an additional transaction that sets restaking to True
* That's 3 transactions in total when creating a stake.
* We can add a flag/prompt in the stake creation process to cancel the restaking step.

**3. Set Re-staking to True when creating the stake**
* Modify the `deposit()` function to accept a `restake` argument.
* Call `deposit()` with `restake=True` when creating the stake.

michwill:
(3) sounds like the cleanest to me. But otherwise, even the current approach is not too bad

szotov:
Also third approach could look like:
modify first deposit
```
function deposit
...
// initial stake of the staker
if (info.subStakes.length == 0) {
  stakers.push(_staker);
  policyManager.register(_staker, getCurrentPeriod());
  info.reStake = True;
...
}
```

KPrasch:
How can we make a move on this? I actually dont have any Issue with proposed solution 2.

PR #1542, opened 2019-12-18
Re-stake flag invert

vzotova:
Based on #1470 

| WARNING: Not upgradeable changes |
| --- |
*Because logic for using flag has changed*

* Inverts re-stake flag
* Re-stake is enabled by default

StakingEscrow `v1.5.1` -> `v2.1.1`


nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -84,7 +84,7 @@ contract StakingEscrow is Issuer {
         */
         uint16 confirmedPeriod1;
         uint16 confirmedPeriod2;
-        bool reStake;
+        bool disableReStake;
```
mswilkison: I personally find it easier to reason about `reStake` than `disableReStake`. disableReStake sounds like a function.

Why change the name of the variable? Why not just change reStake's default value from False to True?

vzotova: it will increase cost of first deposit plus need proper place to set default True.
False by default makes it much easier.
maybe choose other name?

vzotova: if it's not an option - then I prefer to not change anything and use second approach from #1349

mswilkison: I see... maybe just change it to `reStakeDisabled` or `reStakeOff` or something?

PR #1544, opened 2019-12-20
Wind-down parameter

### LINKED ISSUE #1535: Flag to control stakes winddown

vzotova:
Fixes #1535 

Adds wind-down parameter:
* If true then all sub stakes will be decreased each period in `confirmActivity()`
* If false then stake duration will be increased by 1 period after `confirmActivity()`. It guarantees that worker should call `confirmActivity()` method at least `stake-duration` times when winddown will be enabled
* Disabled by default


nucypher/blockchain/eth/agents.py
```diff
@@ -450,6 +450,24 @@ def get_restake_unlock_period(self, staker_address: str) -> int:
         restake_unlock_period = int(staker_info[4])  # TODO: #1348 Use constant or enum
         return restake_unlock_period
 
+    @validate_checksum_address
+    def is_winding_down(self, staker_address: str) -> bool:
+        staker_info = self.get_staker_info(staker_address)
```
michwill: Just a thought for future PRs: we could make `get_staker_info` return a namedtuple

nucypher/cli/painting.py
```diff
@@ -583,6 +584,10 @@ def paint_stakers(emitter, stakers: List[str], staking_agent, policy_agent) -> N
                 emitter.echo(f"{tab}  {'Re-staking:':10} Yes  (Unlocked)")
         else:
             emitter.echo(f"{tab}  {'Re-staking:':10} No")
+        if is_winding_down:
+            emitter.echo(f"{tab}  {'Winding down:':10} Yes")
+        else:
+            emitter.echo(f"{tab}  {'Winding down:':10} No")
```
derekpierre: You could save yourself some lines:

```python
emitter.echo(f"{tab}  {'Winding down:':10} {'Yes' if is_winding_down else 'No'}"
```

michwill:
Uh, wrong one approved. Now need to review this one, too :-)

nucypher/cli/actions.py
```diff
@@ -442,6 +442,14 @@ def confirm_enable_restaking(emitter, staking_address: str) -> bool:
     return True
 
 
+def confirm_enable_winding_down(emitter, staking_address: str) -> bool:
```
derekpierre: @vzotova One extra thought here: Do we need a more explicit blurb about winding down, here, and something explicit in the docs?

nucypher/blockchain/eth/agents.py
```diff
@@ -806,6 +828,17 @@ def lock_restaking(self, release_period: int) -> dict:
         # TODO: Handle ReStakeLocked event (see #1193)
         return receipt
 
+    def set_winding_down(self, value: bool) -> dict:
```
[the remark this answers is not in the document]
vzotova: `PreallocationEscrowAgent` has some code that looks same as in `StakingEscrowAgent`, but it different because uses beneficiary address, also `PreallocationEscrowAgent` and `StakingEscrowAgent` have different contracts 
I guess it's possible that `PreallocationEscrowAgent` will reuse `StakingEscrowAgent`  but it's not in scope of this PR

nucypher/cli/actions.py
vzotova: Definitely yes, but I need some help here, because it's not easy to describe it properly

nucypher/blockchain/eth/actors.py
```diff
@@ -805,6 +816,28 @@ def disable_restaking(self) -> dict:
...
         return receipt
 
+    @property
+    def is_winding_down(self) -> bool:
+        winding_down = self.staking_agent.is_winding_down(staker_address=self.checksum_address)
+        return winding_down
+
+    @only_me
+    @save_receipt
+    def _set_winding_down_value(self, value: bool) -> dict:
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -321,14 +330,14 @@ contract StakingEscrow is Issuer {
     }
 
     /**
-    * @notice Get the value of locked tokens for active stakers in (getCurrentPeriod() + _periods) period
+    * @notice Get the value of locked tokens for active stakersdin (getCurrentPeriod() + _periods) period
```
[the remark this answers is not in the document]

nucypher/cli/painting.py
vzotova: ✔️

nucypher/cli/actions.py
derekpierre: How about the following...?

```
Over time, as the locked stake duration decreases i.e. `winds down`, you will receive decreasing inflationary rewards. Instead, by disabling `wind down` (default) the locked stake duration can remain constant until you specify that `wind down` should begin. By keeping the locked stake duration constant, it ensures that you will receive maximum inflation compensation. 

If `wind down` was previously disabled, you can enable it at any point and the locked duration will decrease after each period.

For more information see ([docs link]).
```

^ @vzotova , @michwill , @cygnusv , @mswilkison, @arjunhassard  wdyt?

@vzotova , we should file a separate issue for adding a more in-depth explanation to the docs, perhaps including stake wind-down diagrams and examples.

michwill: Yeah, diagrams are much required to understand.

I think, it's a good explanation, but I still wouldn't necessarily get it from just text

vzotova: Opened #1551

derekpierre: @vzotova should we link to the docs in this blurb? - perhaps as part of the changes for #1558.

vzotova: Done 
✔️

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -92,8 +93,8 @@ contract StakingEscrow is Issuer {
         uint16 lastActivePeriod;
         bool measureWork;
         uint256 completedWork;
+        bool windDown;
 
-        uint256 reservedSlot1;
```
cygnusv: awesome to see this in use :)
BTW, does it make sense to keep an indication that the reservedSlot was of size 32B? I mention this in case a new `bool` field is added and that perhaps can still fit into the previous slot.

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -92,8 +93,8 @@ contract StakingEscrow is Issuer {
         uint16 lastActivePeriod;
         bool measureWork;
         uint256 completedWork;
+        bool windDown;
```
cygnusv: I thought that `windDown` was going to be `true` by default? Consequently, the field would be something like `windDownDisabled`.

cygnusv: Oh I see in your PR description that the default is False. If this was the consensus, then there's no comment here :)

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
cygnusv: Now that I think of, that depends on the size of all the previous slots, and it can become difficult to track. Nevermind then.

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: Yep, it was consensus, False by default

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: I could add comment that last slot with bool has some empty space (previous uint256 and next slot also uint256, so we have room for new value)

cygnusv: Great, maybe it can be useful in the future :)

PR #1299, opened 2019-09-05
[WIP] NuCypher ConnectNative Firefox Extension 

vepkenez:
* extracted from Demo and rebased on master
* An install script!
* A readme!
* Status check to tell you if it's installed correctly

requires #1555

PR #1555, opened 2020-01-05
Refactor Specifications/Serializers

vepkenez:
* refactors code around input and output specifications and serializers
* removes 100+ lines of controllers/serializers code
* adds new declarative format for io specification using py-marshmallow
* adds improved validation for different datatypes
* additional extensibility to validate keys, treasure maps, message kits etc.
* allows for specifications to indicate type which enables better connect native integration as well as improved compatibility with REST/OPTIONS interfaces
* required by #1299


nucypher/characters/control/specifications/fields/__init__.py
```diff
@@ -0,0 +1,9 @@
+from marshmallow import fields
+from .key import *
+from .treasuremap import *
+from .messagekit import *
+from .datetime import *
+from .label import *
```
derekpierre: Should we just add all field type definitions to the same `fields.py` file instead of one per file? Would that save us from needing this code?

nucypher/characters/control/specifications/fields/messagekit.py
```diff
@@ -0,0 +1,15 @@
...
+class MessageKitField(fields.Field):
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        return b64encode(bytes(value)).decode()
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, bytes):
+            return value
+        return b64decode(value)
+
```
derekpierre: How come `MessageKit` isn't validated? i.e. `def _validate(...)`

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -0,0 +1,12 @@
...
+from base64 import b64decode, b64encode
+
+class TreasureMapField(fields.Field):
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        return b64encode(bytes(value)).decode()
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        return b64decode(value)
+
```
derekpierre: Same question as `MessageKit`: 

How come TreasureMap isn't validated? i.e. def _validate(...)

nucypher/characters/control/specifications/enrico.py
```diff
@@ -0,0 +1,11 @@
+from .fields import fields
+from .base import BaseSchema
+
+class EncryptMessage(BaseSchema):
+
+    message = fields.Str(required=True, load_only=True)
+    message_kit = fields.MessageKit(dump_only=True)
+    signature = fields.Key(dump_only=True) # maybe we need a signature field?
```
derekpierre: I would think that the Signature is different from a Key...

nucypher/characters/control/specifications/bob.py
```diff
@@ -0,0 +1,31 @@
...
+    # but bob.join_policy outputs {'policy_encrypting_key': 'OK'}
+
+
+class Retrieve(BaseSchema):
+    label = fields.Label(required=True, load_only=True)
+    policy_encrypting_key = fields.Key(required=True, load_only=True)
+    alice_verifying_key = fields.Key(required=True, load_only=True, )
+    message_kit = fields.MessageKit(required=True, load_only=True)
+
+    cleartexts = fields.List(fields.Str(), dump_only=True)
```
derekpierre: This output is base64 though, right?

nucypher/characters/control/specifications/fields/messagekit.py
```diff
@@ -0,0 +1,15 @@
+from marshmallow import fields
+from base64 import b64decode, b64encode
+
+
+class MessageKitField(fields.Field):
```
derekpierre: I guess this is really `UmbralMessageKit` unless you intend to use the same field type for `RevocationKit`...?

nucypher/characters/control/interfaces.py
```diff
@@ -185,7 +190,7 @@ def join_policy(self, label: bytes, alice_verifying_key: bytes):
         Character control endpoint for joining a policy on the network.
         """
         self.bob.join_policy(label=label, alice_verifying_key=alice_verifying_key)
-        response = dict()  # {'policy_encrypting_key': ''}  # FIXME
+        response = {'policy_encrypting_key': 'OK'}  # FIXME
```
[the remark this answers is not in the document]
vepkenez: yeah, I didn't do that, just moved it.  I think it's because `bob.join_policy` doesn't output anything useful and maybe that's a big project.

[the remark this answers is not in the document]
vepkenez: Oh, and I just maintained it because the tests are testing for it.  I'm not changing the tests(yet).

nucypher/characters/control/controllers.py
```diff
@@ -37,122 +38,93 @@ class CharacterControllerBase(ABC):
...
         self.emitter = self._emitter_class()
-        self.emitter.transport_serializer = self.serializer
 
         # Disables request & response serialization
         self.serialize = serialize
 
+    def get_serializer(self, interface_name: str):
+        return self.specification.get_serializer(inteface_name)
+
+    def _perform_action(self, action: str, request: dict) -> dict:
```
mswilkison: 👏

nucypher/characters/control/controllers.py
```diff
@@ -37,122 +38,93 @@ class CharacterControllerBase(ABC):
...
     @character_control_interface
     def decrypt(self, request: dict):
-        result = super().decrypt(**self.serializer.load_decrypt_input(request=request))
-        response_data = self.serializer.dump_decrypt_output(response=result)
-        return response_data
+        return self._perform_action('decrypt', request)
 
     @character_control_interface
     def public_keys(self, request):
-        """
```
mswilkison: Should we keep the docstrings? (same question for other occurences)

nucypher/characters/control/specifications/exceptions.py
```diff
@@ -0,0 +1,11 @@
+class SpecificationError(ValueError):
+        """The protocol request is completely unusable"""
```
mswilkison:
```suggestion
    """The protocol request is completely unusable"""
```

nucypher/characters/control/specifications/exceptions.py
```diff
@@ -0,0 +1,11 @@
...
+
+class MissingField(SpecificationError):
+    """The protocol request can be deserialized by is missing required fields"""
+
+class InvalidInputField(SpecificationError):
+    """Response data does not match the output specification"""
```
mswilkison:
```suggestion
    """Response data does not match the input specification"""
```

nucypher/characters/control/specifications/exceptions.py
```diff
@@ -0,0 +1,11 @@
+class SpecificationError(ValueError):
+        """The protocol request is completely unusable"""
+
+class MissingField(SpecificationError):
+    """The protocol request can be deserialized by is missing required fields"""
```
mswilkison:
```suggestion
    """The protocol request cannot be deserialized because it is missing required fields"""
```

nucypher/characters/control/specifications/fields/label.py
```diff
@@ -0,0 +1,16 @@
...
+    def _serialize(self, value, attr, obj, **kwargs):
+        return value.decode('utf-8')
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, bytes):
+            return value
+        return value.encode()
+
+    def _validate(self, value):
+        return True
```
mswilkison: Is some validation logic supposed to happen here?

nucypher/characters/control/specifications/fields/messagekit.py
```diff
@@ -0,0 +1,24 @@
+from typing import Union
```
mswilkison: Unused import

```suggestion
```

tests/characters/test_specifications.py
```diff
@@ -0,0 +1,13 @@
+import pytest
+from nucypher.characters.control.specifications import AliceSpecification
+from nucypher.characters.control.specifications.exceptions import SpecificationError
+
+
+def test_alice_spec_validate(federated_alice, federated_bob):
```
mswilkison: Test other character specs too?

nucypher/characters/control/specifications/fields/datetime.py
```diff
@@ -0,0 +1,12 @@
...
+from nucypher.characters.control.specifications.fields.base import BaseField
+
+
+class DateTime(BaseField, fields.Field):
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        return value.iso8601()
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        return maya.MayaDT.from_iso8601(iso8601_string=value)
```
derekpierre: I'm unsure, but would a validate function here help? i.e. catch invalid input earlier or is it just redundant since you would use Maya for the validation anyway? What's the code path for `_validate`?

nucypher/characters/control/specifications/bob.py
```diff
@@ -0,0 +1,26 @@
...
+    # but bob.join_policy outputs {'policy_encrypting_key': 'OK'}
+
+
+class Retrieve(BaseSchema):
+    label = fields.Label(required=True, load_only=True)
+    policy_encrypting_key = fields.Key(required=True, load_only=True)
+    alice_verifying_key = fields.Key(required=True, load_only=True, )
+    message_kit = fields.UmbralMessageKit(required=True, load_only=True)
+
+    cleartexts = fields.List(fields.Cleartext(), dump_only=True)
```
derekpierre: 👍

nucypher/characters/control/interfaces.py
```diff
@@ -103,6 +107,7 @@ def grant(self,
               n: int,
               expiration: maya.MayaDT,
               value: int = None,
+              first_period_reward: int = None,
```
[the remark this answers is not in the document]
michwill: Yes, I think this parameter got removed

nucypher/characters/control/specifications/fields/datetime.py
vepkenez: Yeah, the _validate runs on the  _output_ Of the _deserialize anyway so it would have failed.  It would be more to check that the date itself met some characteristic.

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -0,0 +1,30 @@
...
+class TreasureMap(BaseField, fields.Field):
+
+    def _serialize(self, value, attr, obj, **kwargs):
+        return b64encode(bytes(value)).decode()
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        return b64decode(value)
+
+    def _validate(self, value):
```
derekpierre: Can validate here simply be (?)

```python
try:
    treasure_map = TreasureMap.from_bytes(value):
    return True
except:
    return False
```

I'm unsure whether the `verify` kwarg for `from_bytes` can remain as `True` (default) or needs to be set to `False`.

vepkenez: Yes, I'd love it if that worked but unfortunately importing `nucypher.policy.collections.Treasuremap` causes a circular import since Bob is imported into that same file... which I think should maybe be a TODO but maybe there is some reason I'm not aware of.

vepkenez:
> Unsure how I feel about the CLI integration - a bit of indirection - need to think a bit more about it.

I agree @derekpierre.  It's not ideal.  It was better before the cleanup-options PR.  I think in the long run, the path is going to be to remove some of the complexity and have all of the CLI options be specified through similar schemas so it will actually be very simple.  

so instead of 
```
alice_verifying_key = fields.Key(
         load_only=True, required=True,
         click=click.option(
             '--alice-verifying-key',
             help="Alice's verifying key as a hexadecimal string",
             required=True, type=click.STRING,))
```

we'll have
```
alice_verifying_key = fields.Key(load_only=True, required=True, help="Alice's verifying key as a hexadecimal string")
```
And then the click stuff would be generated from that. 
Because of the recent changes that happened concurrently to the work in this PR, I just folded it in in a way that would provide exactly matching results.

@KPrasch and I are planning to expand this concept to move most of the ad-hoc code that lives in the cli files into interfaces that can then be accessed through any entry point... but later.

KPrasch:
Rebased over `1234006780ad00952629b5825398625a0428319e`

nucypher/characters/control/interfaces.py
```diff
@@ -1,59 +1,31 @@
...
-        # Validate response
-        instance.validate_response(response=response, interface_name=interface_name)
+    def callable(func):
+        func._schema = schema()
 
-        # Record duration
-        responding = maya.now()
-        duration = responding - received
+        @functools.wraps(func)
+        def wrapped(*args, **kwargs):
```
KPrasch: This inner wrapper does not do anything?

nucypher/characters/control/interfaces.py
```diff
@@ -62,13 +34,27 @@ def __init__(self, character=None, *args, **kwargs):
         self.character = character
         super().__init__(*args, **kwargs)
 
+    @classmethod
+    def connect_cli(cls, action):
```
KPrasch: A comment here in is sorely needed.

nucypher/characters/control/controllers.py
```diff
@@ -178,27 +71,60 @@ def __init__(self,
         # Internals
         self._transport = None
 
-        # Hard-wire the character's output flow to the Emitter
-        self._internal_controller = character_controller
-        self._internal_controller.emitter = self.emitter
+        self.interface = interface
+
+        def set_method(name):
```
KPrasch: Needs follow up for deeper integration IMO - perhaps move to a method explaining the polymorphic approach and implicit routing implicitly occurring here.   I think this can be named `self.bind_interface()`

nucypher/characters/control/controllers.py
```diff
@@ -178,27 +71,60 @@ def __init__(self,
...
+
+            def wrapper(request=None, **kwargs):
+                request = request or kwargs
```
KPrasch: This is a shim for optionally passing kwargs to a controller rather than a dictionary (unsupported behaviors), which needs to be more formally enforced.

nucypher/characters/control/controllers.py
```diff
@@ -178,27 +71,60 @@ def __init__(self,
...
 
-        super().__init__(control_serializer=self._internal_controller.serializer)
+        for method_name in self._get_interfaces().keys():
+            set_method(method_name)
+
+        super().__init__(*args, **kwargs)
 
         self.log = Logger(app_name)
 
+    def _get_interfaces(self):
```
KPrasch: Let's embed this method on a base class of `Interfaces`

nucypher/characters/control/controllers.py
```diff
@@ -178,27 +71,60 @@ def __init__(self,
...
+        return {
+            name: method for name, method in
+            inspect.getmembers(
+                self.interface,
+                predicate=inspect.ismethod)
+            if hasattr(method, '_schema')
```
KPrasch: Perhaps move the suffix to somewhere more conspicuous like a class level variable `Interface.PREDICATE`.

nucypher/characters/control/controllers.py
```diff
@@ -178,27 +71,60 @@ def __init__(self,
...
     def test_client(self):
         return NotImplemented
 
 
+class CLIController(CharacterControlServer):
+
+    _emitter_class = StdoutEmitter
+
+    def make_control_transport(self):
+        return
```
KPrasch: Seems awkward to disable this method, perhaps indicating additional refactoring needed to fully realize the critique of this PR.

nucypher/cli/options.py
```diff
@@ -53,6 +51,9 @@
 option_teacher_uri = click.option('--teacher', 'teacher_uri', help="An Ursula URI to start learning from (seednode)", type=click.STRING)
 _option_middleware = click.option('-Z', '--mock-networking', help="Use in-memory transport instead of networking", count=True)
 
+# Avoid circular input
```
KPrasch: This is not a common option - So this is just a semi-temporary location for this flag until the full refactor can be performed.

tests/characters/control/blockchain/test_rpc_control_blockchain.py
```diff
@@ -56,6 +63,14 @@ def test_alice_rpc_character_control_create_policy(alice_rpc_test_client, create
         assert rpc_response.success is True
         assert rpc_response.id == response_id
 
+def test_alice_rpc_character_control_bad_input(alice_rpc_test_client, create_policy_control_request):
```
derekpierre: do we only test bad input for alice? what about others?

nucypher/characters/control/specifications/alice.py
```diff
@@ -36,9 +39,26 @@ class PolicyBaseSchema(BaseSchema):
...
+    )
+
     # output
     policy_encrypting_key = fields.Key(dump_only=True)
 
+    @validates_schema
+    def check_n_gte_m(self, data, **kwargs):
+        # ensure that n is greater than or equal to m
+        if not data['n'] >= data['m']:
+            raise InvalidArgumentCombo
```
derekpierre: We should be descriptive here i.e. include the reason text.

nucypher/characters/control/specifications/alice.py
```diff
@@ -36,9 +39,26 @@ class PolicyBaseSchema(BaseSchema):
...
+    rate = fields.Wei(
+        load_only=True,
+        required=False,
+        click=options.option_rate
+    )
+
     # output
     policy_encrypting_key = fields.Key(dump_only=True)
 
+    @validates_schema
```
derekpierre: 👍 - it's cool that all of the validation logic does not have to be done in one function.

tests/characters/test_specifications.py
```diff
@@ -1,10 +1,50 @@
...
+    result = GrantPolicy().load(data)
+    assert result['label'] == b'cats the animal'
+
+    # validate that m > n fails validation
+    data['m'] = data['n'] + 19
     with pytest.raises(SpecificationError) as e:
-        _result = CreatePolicy().load(dict())
+        GrantPolicy().load(data)
+
+    breakpoint()
```
derekpierre:
```suggestion
```

vepkenez: heh. yeah :|

tests/characters/control/blockchain/test_rpc_control_blockchain.py
vepkenez: It's not really in this PR's scope for me to write tests for every single character's input.  I'm just testing the implementation.  The code paths are the same for all the other characters so this test covers all of them to some degree.

nucypher/characters/lawful.py
```diff
@@ -384,64 +374,55 @@ def public_keys():
...
 
         @alice_flask_control.route('/derive_policy_encrypting_key/<label>', methods=['POST'])
         def derive_policy_encrypting_key(label) -> Response:
             """
             Character control endpoint for deriving a policy encrypting given a unicode label.
             """
-            response = controller(interface=controller._internal_controller.derive_policy_encrypting_key,
-                                  control_request=request,
-                                  label=label)
+            response = controller(method_name='derive_policy_encrypting_key', control_request=request, label=label)
```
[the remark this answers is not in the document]
vepkenez: This is some vestigial thing left over from the web controller maybe...  I'm just keeping it as is.

tests/characters/control/blockchain/test_rpc_control_blockchain.py
derekpierre: It feels like our list of follow-ups to this PR is growing. If not done already, perhaps creating and linking follow-up issue(s) would be best.
1. CLI parameter thing we talked about during the walkthrough
2. Expanded character tests, or modifying the tests to be generic/repeatable for other character specifications.

derekpierre:
@KPrasch , @vepkenez : The following comment is hindering me from approving this PR: https://github.com/nucypher/nucypher/pull/1555/files#r363467150 - any thoughts there? Perhaps I'm mistaken? Once resolved I'll approve.

nucypher/characters/base.py
```diff
@@ -116,7 +116,9 @@ def __init__(self,
 
         #
         # Operating Mode
-
+        if hasattr(self, '_interface_class'):  # TODO: have argument about meaning of 'lawful'
+            #                                         and whether maybe only Lawful characters have an interface
+            self.interface = self._interface_class(character=self)
```
KPrasch: This is only needed when using a controller, consider moving this out of init to prevent interface creation for characters that use just the plain python API.

nucypher/characters/control/interfaces.py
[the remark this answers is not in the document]
KPrasch: #1459 is already merged into matster, so if this PR is currently based over the latest, and this variable is unused within this function it can be safely removed.

nucypher/characters/control/interfaces.py
```diff
@@ -171,29 +162,28 @@ def decrypt(self, label: bytes, message_kit: bytes) -> dict:
...
 
+    @attach_schema(bob.JoinPolicy)
     def join_policy(self, label: bytes, alice_verifying_key: bytes):
         """
         Character control endpoint for joining a policy on the network.
         """
-        self.bob.join_policy(label=label, alice_verifying_key=alice_verifying_key)
-        response = dict()  # {'policy_encrypting_key': ''}  # FIXME
+        self.character.join_policy(label=label, alice_verifying_key=alice_verifying_key)
+        response = {'policy_encrypting_key': 'OK'}  # FIXME
```
KPrasch: Do we have the PEK from the result of the above call?

nucypher/characters/control/specifications/alice.py
KPrasch: Continued scoping: Handle errors and error output via `Emitters`

nucypher/characters/control/specifications/exceptions.py
```diff
@@ -0,0 +1,14 @@
...
+class MissingField(SpecificationError):
+    """The protocol request cannot be deserialized because it is missing required fields"""
+
+class InvalidInputField(SpecificationError):
+    """Response data does not match the input specification"""
+
+class InvalidOutputField(SpecificationError):
+    """Response data does not match the output specification"""
+
+class InvalidArgumentCombo(SpecificationError):
```
jMyles: This is interesting.  We've often discussed how to formalize "xor" arguments - we have different mutually exclusivity logic in different parts of the codebase.  Perhaps this is the beginning of something normal.

nucypher/characters/base.py
```diff
@@ -515,11 +517,20 @@ def _set_checksum_address(self):
 
         self._checksum_address = public_address
 
-    def make_rpc_controller(drone, crash_on_error: bool = False):
-        app_name = bytes(drone.stamp).hex()[:6]
+    def make_rpc_controller(self, crash_on_error: bool = False):
```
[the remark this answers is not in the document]
tuxxy: This. I really enjoyed reading `drone` as it made a lot of sense when reading.

nucypher/characters/control/emitters.py
```diff
@@ -232,10 +237,8 @@ def exception(drone_character,
             raise e
         return drone_character.sink(str(e), status=response_code)
 
-    def ipc(drone_character, response, request_id, duration) -> Response:
-        assembled_response = drone_character.assemble_response(response=response,
-                                                               request_id=request_id,
-                                                               duration=duration)
+    def respond(drone_character, response) -> Response:
```
tuxxy: Solid rename. :+1:

nucypher/characters/lawful.py
```diff
@@ -136,7 +136,7 @@ def __init__(self,
                                             checksum_address=checksum_address)
 
         if is_me and controller:
-            self.controller = self._controller_class(alice=self)
+            self.make_cli_controller()
```
tuxxy: I really like how this evolved. :+1:

nucypher/characters/control/specifications/alice.py
```diff
@@ -0,0 +1,121 @@
...
+
+    @validates_schema
+    def check_n_gte_m(self, data, **kwargs):
+        # ensure that n is greater than or equal to m
+        if not data['n'] >= data['m']:
+            raise InvalidArgumentCombo("N must be greater than or equal to M")
+
+    @validates_schema
+    def check_rate_or_value_not_both(self, data, **kwargs):
+        if data.get('rate') and data.get('value'):
```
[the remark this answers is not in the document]
vepkenez: hmm maybe.  I'll make that change!

tests/characters/test_specifications.py
```diff
@@ -0,0 +1,48 @@
...
+
+    # remove value and now it works
+    del data['value']
+    result = GrantPolicy().load(data)
+    assert result['label'] == b'cats the animal'
+
+    # validate that m > n fails validation
+    data['m'] = data['n'] + 19
+    with pytest.raises(SpecificationError) as e:
+        GrantPolicy().load(data)
```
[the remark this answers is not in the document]
vepkenez: yes this test is testing the specification validation, not the granting.

nucypher/characters/control/specifications/fields/label.py
cygnusv: > Is some validation logic supposed to happen here?

Do we want to accept empty labels (`""`)? Are there other encodings that are outside UTF8?

nucypher/characters/control/specifications/fields/misc.py
```diff
@@ -0,0 +1,34 @@
...
+
+class String(BaseField, fields.String):
+    pass
+
+
+class List(BaseField, fields.List):
+    pass
+
+
+class Integer(BaseField, fields.Integer):
```
cygnusv: I'd create a PositiveInteger subclass with a `_validate()`, as most (all?) integer fields are actually intended to be positive.

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -0,0 +1,30 @@
...
+        # TODO: we don't currently accept any treasuremap as input so this is not
+        # used.
```
cygnusv: This is going to change in birdcage. Should this PR be prepared for that? Unless @jmyles has already done this there.

nucypher/cli/commands/alice.py
```diff
@@ -393,7 +384,7 @@ def grant(general_config,
         if any((value, rate)):
             raise click.BadOptionUsage(option_name="--value, --rate",
                                        message="Can't use --value or --rate with a federated Alice.")
-    elif not (bool(value) ^ bool(rate)):
+    if bool(value) and bool(rate):
```
cygnusv: Why this change? Now it misses when `bool(value)==False` and `bool(rate)==False`

cygnusv:
Also, I think the commit history in this PR could be cleaner. For example, there are several commits that can be easily squashed.

nucypher/characters/control/specifications/fields/misc.py
vepkenez: already done.  next push!

derekpierre:
I agree with @cygnusv that some commits can probably be squashed for a cleaner history.

PR #1557, opened 2020-01-09
Incorrect recommendation for missing StakeholderConfiguration file

### LINKED ISSUE #1553: Incorrect recommendation in CLI when stakeholder configuration not found
```
$ nucypher stake divide --provider /Users/derek/Library/Ethereum/goerli/geth.ipc

 ____    __            __
/\  _`\ /\ \__        /\ \
\ \,\L\_\ \ ,_\    __ \ \ \/'\      __   _ __
 \/_\__ \\ \ \/  /'__`\\ \ , <    /'__`\/\`'__\
   /\ \L\ \ \ \_/\ \L\.\\ \ \\`\ /\  __/\ \ \/
   \ `\____\ \__\ \__/.\_\ \_\ \_\ \____\\ \_\
    \/_____/\/__/\/__/\/_/\/_/\/_/\/____/ \/_/

The Holder of Stakes.

Error: Could not open file /Users/derek/Library/Application Support/nucypher/stakeholder.json: No Stakeholder configuration file found.
To create a new persistent Stakeholder run: 'nucypher stakeholder init'
```

**`nucypher stakeholder init` is no longer a valid command**

derekpierre:
Fixes #1553 .

### DIFF
--- a/tests/cli/test_alice.py
+++ b/tests/cli/test_alice.py
@@ -1,10 +1,20 @@
 import os
+from unittest import mock
 
 from nucypher.cli.main import nucypher_cli
 from nucypher.config.characters import AliceConfiguration
 from nucypher.utilities.sandbox.constants import INSECURE_DEVELOPMENT_PASSWORD, \
     MOCK_IP_ADDRESS, MOCK_CUSTOM_INSTALLATION_PATH, TEMPORARY_DOMAIN
 from nucypher.cli.actions import SUCCESSFUL_DESTRUCTION
+
+
+@mock.patch('nucypher.config.characters.AliceConfiguration.default_filepath', return_value='/non/existent/file')
+def test_missing_configuration_file(default_filepath_mock, click_runner):
+    cmd_args = ('alice', 'run')
+    result = click_runner.invoke(nucypher_cli, cmd_args, catch_exceptions=False)
+    assert result.exit_code != 0
+    assert default_filepath_mock.called
+    assert "run: 'nucypher alice init'" in result.output
 
 
 def test_initialize_alice_defaults(click_runner, mocker):
--- a/nucypher/cli/actions.py
+++ b/nucypher/cli/actions.py
@@ -286,11 +286,12 @@ paid out in ethers retro-actively and on-demand.
 Accept ursula node operator obligation?""", abort=True)
 
 
-def handle_missing_configuration_file(character_config_class, config_file: str = None):
+def handle_missing_configuration_file(character_config_class, init_command_hint: str = None, config_file: str = None):
     config_file_location = config_file or character_config_class.default_filepath()
+    init_command = init_command_hint or f"{character_config_class._NAME} init"
     message = f'No {character_config_class._NAME.capitalize()} configuration file found.\n' \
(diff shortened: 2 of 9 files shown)


nucypher/config/characters.py
```diff
@@ -321,3 +321,9 @@ def generate(cls, *args, **kwargs):
 
     def to_configuration_file(self, override: bool = True, *args, **kwargs) -> str:
         return super().to_configuration_file(override=True, *args, **kwargs)
+
+    @classmethod
+    def init_cmd(cls):
+        # provide a direct link to underlying command so that if it ever changes again, it will be detected
+        from nucypher.cli.characters import stake
```
fjarri: I'm not sure I agree with this approach. First, there is a non-root-level import that hides a circular dependency - a very possible source of problems in the future (especially if we ever want to decouple CLI from the rest of the code). Second, if there is already a test that calls this command, and the command does change, we will detect it there - throwing an exception a little earlier will not change anything. Plus, the whole purpose of `init_cmd()` seems to be for testing, I can't imagine a scenario where a user would call it. So perhaps the testsuite is where it should reside.

derekpierre: I don't think I follow:

The underlying issue is not about calling the CLI command - the issue is that when a configuration file is missing, the **message** that we show about recommending a command to use to persist a character is incorrect (see #1553 output). Before "stake", all characters are initialized with `init` eg. `nucypher alice init`, `nucypher bob init` etc. However for stakeholder, we do not use init, but rather `nucypher stake init-stakeholder` i.e. different from all other characters, yet the message was still indicating that `nucypher stake init` should be used.

The reason for the `init_cmd()` function is for the cases when the default `init` is not the command to initialize some character from its configuration. I used a direct link to the non-default function so that if we decide to change it again to something else, either the test for StakeHolderConfiguration will fail because command is no longer available, OR it will continue to pass because the recommendation gets automatically updated.

For example, if the `init_stakeholder(...)` function name would go away and the test would fail. However, if the `init_stakeholder` function can remain but the @stake.command(name=...) decorator gets changed (which changes the cli name of the function, but not the code) then the message automatically gets updated.

fjarri: I understand the issue, what worries me is the function-local import, which backfires in 99% of cases. Would it work to just pass the required command to `handle_missing_configuration_file`? E.g.
```
            except FileNotFoundError:
                return actions.handle_missing_configuration_file(
                    character_name_hint="Alice",
                    command_name_hint="alice", # can be omitted if equal to lowercased `character_name_hint`
                    config_file=config_file)
```
So you can do something like `command_name_hint=init_stakeholder.name` to catch the change, because it is all confined to the same file.

tests/cli/test_alice.py
```diff
@@ -7,6 +8,15 @@
 from nucypher.cli.actions import SUCCESSFUL_DESTRUCTION
 
 
+@mock.patch('nucypher.config.characters.AliceConfiguration.default_filepath', return_value='/non/existent/file')
```
KPrasch: Nice - Inspired by the recent mocking work in birdcage? :-)

derekpierre: 😉

PR #1558, opened 2020-01-09
Sub-stakes description

### LINKED ISSUE #1551: Update docs with sub stakes description (re-staking, winding down diagrams)

vzotova:
Fixes #1551

Description of internal organization of sub-stakes with diagrams.
Feel free to change anything, even whole structure.


docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+
+
+Operations that modify the sub-stake array
+------------------------------------------
+
+Deposit and locking
+^^^^^^^^^^^^^^^^^^^
+*Methods* : ``deposit(uint256,uint16)``,  ``deposit(address,uint256,uint16)``,  ``lock(uint256,uint16)``
+
+To become a staker, tokens owner transfers tokens to ``StakingEscrow`` and lock them  using one of the ``deposit()`` methods. In case that staker already has some amount of unlocked tokens on the account in the contract then lock can be created using ``lock()`` method. As a result of executing on of the above methods, a new element will be created in the array of sub-stakes with the next period as starting date  and with duration equals input parameter.
```
derekpierre:
```suggestion
To become a staker, NU tokens must be transferred to the ``StakingEscrow`` contract and locked using one of the ``deposit()`` methods. If the staker already has unlocked tokens within the account in the contract, then the sub-stake can be locked using the ``lock()`` method. If successful, a new element will be created in the array of sub-stakes with the next period as the starting date, and the duration equal to the input parameter.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+            +---+-------------+--->
+            0   1             5
+
+
+
+Lock prolongation
+^^^^^^^^^^^^^^^^^
+*Methods* : ``prolongStake(uint256,uint16)``
+
+In order to increase the staking reward, as well as the possibility of obtaining policies with a longer time,  staker can increase the locking duration of the sub-stake using the ``prolongStake()`` method. The number of sub-stakes does not change, but the end period of locking in the specified sub-stake will be changed.
```
derekpierre:
```suggestion
In order to increase the staking reward, as well as the possibility of obtaining policies with a longer timeframe, stakers can increase the duration of their locked sub-stake using the ``prolongStake()`` method. The number of sub-stakes does not change, but the locked duration for the specified sub-stake will be increased.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+            |      1st    |  period
+            +-------------+--->   
+            0             7     
+			
+			
+Sub-stake division
+^^^^^^^^^^^^^^^^^^
+*Methods* : ``divideStake(uint256,uint256,uint16)``
+
+If necessary, staker can extend the locking duration only for the part of tokens in a sub-stake (``divideStake()``). Contract method splits desirable sub-stake into two elements: the old sub-stake with changed locked amount and the new sub-stake. New element has the extended locking duration (by the specified number of periods) and the same start period as the old one has. Locked amount in the new sub-stake equals to requested input parameter.
```
derekpierre:
```suggestion
If necessary, stakers can extend the locking duration for only a portion of their tokens in a sub-stake by using the ``divideStake()`` method. This method splits a sub-stake into two elements: the old sub-stake with the reduced locked amount and the new sub-stake with the specified amount. The new sub-stake has the specified locked amount and an extended lock duration, based on the specified number of periods, with the same start period as the old sub-stake.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+
+Flags that affect the sub-stake array
+-------------------------------------
+
+Re-staking
+^^^^^^^^^^
+*Used in methods* : ``confirmActivity()``, ``mint()``
+
+When re-staking is turned off, the number of locked tokens in sub-stakes does not change by itself.
+In case when re-staking parameter is on then all staking reward is locked as the part of each involved sub-stake (inside ``confirmActivity()`` and/or ``mint()``).  Accordingly, each such sub-stake has increased locked amount (by reward) and the number of sub-stakes stays unchanged.
```
derekpierre:
```suggestion
However, when re-staking is enabled then all staking rewards are re-locked as part of each relevant sub-stake (inside ``confirmActivity()`` and/or ``mint()``).  Consequently, each such sub-stake has an increased locked amount (by reward) and the number of sub-stakes remains unchanged.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+
+**Example:**
+
+    A staker has few sub-stakes and calls ``mint()``:
```
derekpierre: For this scenario, is it the case that 1st sub-stake produces 50 tokens and 2nd sub-stake produces 20 tokens as rewards? If so, we should state that explicitly for the scenario.

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+         700|     +----+
+            |     | 3rd|
+         600+-----+----+
+         500|          +-------------+
+            |    2nd   |     3rd     |
+         400+----------+-------------+----------+
+            |                                   |
+            |               1st                 |   period
+            +-----------------------------------+--->    
+            0     1    2             5          10
```
derekpierre: Since the `3rd` sub-stake is to start at period and last for 5 periods, should the end of the `3rd` sub-stake correspond to `6` instead of `5` on the timeline? - Same for the "After" diagram below?

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+
+
+Flags that affect the sub-stake array
+-------------------------------------
+
+Re-staking
+^^^^^^^^^^
+*Used in methods* : ``confirmActivity()``, ``mint()``
+
+When re-staking is turned off, the number of locked tokens in sub-stakes does not change by itself.
```
derekpierre: Should we indicate what the default is? - like we did for `wind-down`

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+            |                                   |   period
+            +-----------------------------------+--->    
+            0     1    2             5          10
+			
+
+Winding down
+^^^^^^^^^^^^
+*Used in methods* : ``confirmActivity()``
+
+Disabled winding down parameter (by default) guarantees that worker must call ``confirmActivity()`` at least N times after parameter will be turned on to unlock sub-stake, where N is locking duration of sub-stake. Thus unlocking date for each sub-stakes shifts by 1 period each period (duration remains the same). In case when winding down is enabled, each ``confirmActivity()`` (no more than once in a period) leads to decreasing of locking duration of each sub-stake. If worker calls ``confirmActivity()`` each period then unlocking date remains unchanged.
```
derekpierre:
```suggestion
A disabled "winding down" parameter (default) guarantees that the worker must call ``confirmActivity()`` at least N times after the parameter is enabled to unlock the sub-stake, where N is the locking duration of sub-stake. When disabled, the unlock date for each sub-stakes shifts forward by 1 period after each period i.e. the duration continues to remain the same until the parameter is enabled. Once the "winding down" parameter is enabled, each call to ``confirmActivity()`` (no more than once in a period) leads to a reduction of locking duration for each sub-stake, and the unlock date no longer changes.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
+Sub-stakes
+==========
+
+A staker may extend the unlock period for any number of portions of their total stake. This divides the stake into smaller parts, each with a unique unlock date in the future. Stakers may also acquire and lock new tokens. The total stake is represented as the sum of all the different sub-stakes active in a given cycle (new cycle every 24h), which includes locked sub-stakes, and any sub-stakes that have passed their unlock date, and can be freely withdrawn. Each sub-stake has a beginning and duration (lock time). When a staker confirms activity each day and winding down is enabled, the remaining lock time for relevant sub-stakes is reduced.
+
+Sub-stake remains active until become unlocked and staker get reward for last period of sub-stake by calling ``mint()`` or ``confirmActivity()`` after the last period is surpassed. Each staker can have no more than 30 active sub-stakes. All sub-stakes changes use slots of inactive sub-stakes in the first place and only after that - empty slots in the array. Therefore, obtaining data on past staker locks is not guaranteed.
```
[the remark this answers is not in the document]
vzotova: Few moments:
- Collecting reward is not necessary to make sub-stake inactive. We can "remove" element from sub-stakes array when we no need information about locking. After last period of sub stake become history we still need to calculate reward for this last period. It will happen inside mint() method wich work only with surpassed period.
- Not sure about "the inactive stake slots", for me sounds like slot is inactive (maybe I'm wrong). I'll try to rephrase idea: we have array with sub-stakes with fixed size (30). Some of this sub-stakes are not usefull for any calculation in StakingEscrow, they are inactive and we can reuse them. When we add new one - we just looking from left to right trying to find place where to save new element, inactive sub-stake or empty slots if there are any. Not sure that is more clear than before :)
- We allways can get information about locking for any staker for the current moment and for future, also we could try to retreive this information for some period in the past, and if were no any rewrites of inactive slots then this information still will be valid

docs/source/architecture/sub_stakes.rst
vzotova: I like sentence, just not sure about  `the sub-stake can be locked`, in my head it sounds like sub-stake can be only created and locking tokens is staking/creating satake. What do you think?

docs/source/architecture/sub_stakes.rst
vzotova: Yep, you're totally right, where to put it? in description of before? Or in "after" sections?

docs/source/architecture/sub_stakes.rst
vzotova: I'm glad that somebody asked this :)
So thing is that locking for 5 periods means 5 future period - 1,2,3,4,5.  3rd sub-stake will last up to 5th period and will end in 6th. Current period is already part of history from this perspective.

docs/source/architecture/sub_stakes.rst
vzotova: Thanks, I forgot to add this

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
derekpierre: I saw your update: 👍 .

docs/source/architecture/sub_stakes.rst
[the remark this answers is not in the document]
derekpierre: Thanks for the explanation - that makes things clearer. I guess the terminology `staker locks` is just a little weird to me.

```suggestion
A sub-stake remains active until it becomes unlocked, and a staker collects the reward for the last period by calling ``mint()`` or ``confirmActivity()`` once the last period is surpassed. Each staker can have no more than 30 active sub-stakes which are stored in an array. All sub-stake changes initially reuse slots of inactive sub-stakes for storage in the array, and if there are none, will instead use empty slots. Therefore, attempting to retrieve data about previous inactive sub-stakes is not guaranteed to be successful since the data could have been overwritten.
```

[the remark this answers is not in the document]
vzotova: Great! Just changed `collects` -> `gets`, because we have CLI command that collects reward and this command has other meaning

docs/source/architecture/sub_stakes.rst
vzotova: I took your suggestion with some modifications, tell me what do you think

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+
+**Example:**
+
+    A staker has few sub-stakes and calls ``mint()``:
+		- Before calling: 
+			* 1st sub-stake = 400 tokens for 10 periods
+			* 2nd sub-stake = 200 tokens for 2 periods
+			* 3rd sub-stake = 100 tokens locked starting from the next period and locked for 5 periods
+			* 100 tokens in an unlocked state
+		- After calling, re-staking is disabled:
```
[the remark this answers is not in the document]
vzotova: Agreed, I like it
✔️

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+Re-staking
+^^^^^^^^^^
+*Used in methods* : ``confirmActivity()``, ``mint()``
+
+When re-staking is turned off, the number of locked tokens in sub-stakes does not change by itself.
+However, when re-staking is enabled (default) then all staking rewards are re-locked as part of each relevant sub-stake (inside ``confirmActivity()`` and/or ``mint()``).  Consequently, each such sub-stake has an increased locked amount (by reward) and the number of sub-stakes remains unchanged.
+
+**Example:**
+
+    A staker has few sub-stakes and calls ``mint()``. Assume that thus far the 1st and 2nd sub-stakes will produced 50 tokens and 20 tokens respectively in rewards:
```
derekpierre: You can either do `will produce` (future tense) or `have produced` (past tense)

docs/source/architecture/sub_stakes.rst
derekpierre: 👍 added minor comment about tense.

docs/source/architecture/sub_stakes.rst
vzotova: I tried to show discrete figure, when there is no such thing like 1->2, for example lock for one period  will be shown as vertical line. 
In all calculation current period ignored, so almost everything starts from 1 on those graphs. So lock for 10 periods will be for 1,2,3...9,10 period number, 0 is not included in this row. I guess this's problem in the explanation, maybe add "future periods" everywhere?

docs/source/architecture/sub_stakes.rst
vzotova: Thanks, I tried to use future tense but failed :)

docs/source/architecture/sub_stakes.rst
vzotova: How about such figure:
image
I'm trying to use `aafigure` lib, it helps to maintain in future, but if still not good - I'll search something else

docs/source/architecture/slashing.rst
```diff
@@ -65,21 +65,21 @@ Penalty Scenarios:
 
     Result:
 
-        * 1st sub stake = 500 tokens locked for 10 periods
-        * 2nd sub stake = 200 tokens for 2 periods
-        * 3rd sub stake = 100 tokens locked starting from the next period
+        * 1st sub-stake = 500 tokens locked for 10 periods
+        * 2nd sub-stake = 200 tokens for 2 periods
+        * 3rd sub-stake = 100 tokens locked starting from the next period
         * 100 tokens in an unlocked state
```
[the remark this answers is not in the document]
vzotova: I'm not sure I agree with you.
I see this as help to reader with what to expect from the next scenarios

docs/source/architecture/slashing.rst
```diff
@@ -65,21 +65,21 @@ Penalty Scenarios:
...
 
 * *Scenario 2*: Staker incurs penalty calculated to be worth **300 tokens**:
 
-    The unlocked tokens can only cover 200 tokens. In the current period, 700 tokens are locked and 800 tokens are locked for the next period. Therefore, we should reduce amount of locked tokens for the next period and leave unchanged locked amount in the current period. The 3rd sub stake suits for this purpose but it's not the shortest one. So we take the 2nd sub stake (the shortest), reduce it to 100 tokens and add new sub stake with 100 tokens which active only in the current period.
+    The unlocked tokens can only cover 200 tokens. In the current period, 700 tokens are locked and 800 tokens are locked for the next period. Therefore, we should reduce amount of locked tokens for the next period and leave unchanged locked amount in the current period. The 3rd sub-stake suits for this purpose but it's not the shortest one. So we take the 2nd sub-stake (the shortest), reduce it to 100 tokens and add new sub-stake with 100 tokens which active only in the current period.
```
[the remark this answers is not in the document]
vzotova: I like your suggestion except one sentence `In this scenario, the next period's sum is reduced and the current period's sum remains unchanged.` Which related to your questions.
Before slashing staker has 1000 tokens, fine is 300 tokens, so after slashing maximum locked tokens must be equal or less 700. It's why current period is untouched

docs/source/architecture/slashing.rst
```diff
@@ -26,16 +26,16 @@ The entire stake consists of:
...
+.. _`Sub-stakes`: https://docs.nucypher.com/en/latest/architecture/sub_stakes.html
 
-Sub stakes get slashed in the order of their remaining lock time, beginning with the shortest – so the first portion of the stake to be slashed is the unlocked portion. After that, if necessary, locked sub-stakes are decreased – the shortest sub stake is decreased by the required amount; if the adjustment of that sub-stake is insufficient to fulfil the required punishment sum, then the next shortest sub-stake is decreased, and so on. Sub-stakes that begin in the next period are checked separately.
+`Sub-stakes`_ get slashed in the order of their remaining lock time, beginning with the shortest – so the first portion of the stake to be slashed is the unlocked portion. After that, if necessary, locked sub-stakes are decreased – the shortest sub-stake is decreased by the required amount; if the adjustment of that sub-stake is insufficient to fulfil the required punishment sum, then the next shortest sub-stake is decreased, and so on. Sub-stakes that begin in the next period are checked separately.
```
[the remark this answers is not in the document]
vzotova: Will be slashed shortest sub-stake - which starts from tomorrow to minimize influence of slashing

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,270 @@
...
+
+A sub-stake remains active until it becomes unlocked, and a staker gets the reward for the last period by calling ``mint()`` or ``confirmActivity()`` once the last period is surpassed. Each staker can have no more than 30 active sub-stakes which are stored in an array. All sub-stake changes initially reuse slots of inactive sub-stakes for storage in the array, and if there are none, will instead use empty slots. Therefore, attempting to retrieve data about previous inactive sub-stakes is not guaranteed to be successful since the data could have been overwritten.
```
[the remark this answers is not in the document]
vzotova: Can't accept this suggestion, some little things do changes in emphasize. Splitting first sentence into two makes other meaning that I want: active sub-stake is two condition but in suggetion looks like only one.
And don't want to put info about empty slots to parentheses

docs/source/architecture/slashing.rst
```diff
@@ -101,13 +101,13 @@ Penalty Scenarios:
    
 * *Scenario 3*: Staker incurs penalty calculated to be worth **400 tokens**:
 
-    The difference from the previous scenario is that should also decrease locked tokens in the current period. At the first step the 2nd sub stake is reduced to 100 tokens. Next step - adjustment for the next period. The shortest sub stake still the same - the 2nd. And we need to deacrese it from 100 to 0 only for the next period. Will be the same if we change duration of the 2nd sub stake from 2 periods to 1 and the other sub stakes remain unchanged.
+    The difference from the previous scenario is that should also decrease locked tokens in the current period. At the first step the 2nd sub-stake is reduced to 100 tokens. Next step - adjustment for the next period. The shortest sub-stake still the same - the 2nd. And we need to deacrese it from 100 to 0 only for the next period. Will be the same if we change duration of the 2nd sub-stake from 2 periods to 1 and the other sub-stakes remain unchanged.
```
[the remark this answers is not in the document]
vzotova: ✔️

docs/source/architecture/slashing.rst
```diff
@@ -127,13 +127,13 @@ Penalty Scenarios:
  
 * *Scenario 4*: Staker incurs penalty calculated to be worth **600 tokens**:
 
-    Reducing the unlocked remaining tokens, 3rd sub stakes, and the shortest sub stake (2nd) is not enough, so they are all removed. The next shortest sub stake is the 1st which is reduced from 500 to 400.
+    Reducing the unlocked remaining tokens, 3rd sub-stakes, and the shortest sub-stake (2nd) is not enough, so they are all removed. The next shortest sub-stake is the 1st which is reduced from 500 to 400.
```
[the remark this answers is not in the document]
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
derekpierre: @vzotova, 👍 I can get on board with that if that works with `aafigure`.

vzotova: Updated all diagrams 
waiting your review (also better to check version after compilation)

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,276 @@
...
+        :proportional:
+        :textual:
+
+            stake
+            ^
+            |
+         900|   +-------------------+
+            |   |        1st        |   period
+            +---+---+---+---+---+---+---->
+            | 0 | 1 | 2 | 3 | 4 | 5 | 6
```
derekpierre: 👍

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,276 @@
...
+                         Before             
+
+            stake
+            ^
+            |
+         900+-----------------------+
+            |          1st          |   period
+            +---+---+---+---+---+---+---->
+            | 0 | 1 | 2 | 3 | 4 | 5 | 6      
+
```
derekpierre: @vzotova This is where I get confused - (and apologies, if I'm just totally missing it) - but shouldn't the first sub-stake for the `before` scenario stop at the end of period 4, and not at the end of period 5? i.e. the 1st sub-stake includes the current period and is for periods 0, 1, 2, 3, 4 (total = 5 periods). 

This stake starts from the current period (0), no? or is it 5 periods starting the next period, like in the `Deposit and locking` scenario above, in which case it should start from period 1?

vzotova: current period does not count, it's sort of historical data, for example lock for 0 periods looks like
```
^
|
+---+
|   |
+---+--->
| 0 | 1
```

derekpierre: Perhaps it boils down to a wording issue - we should indicate whether the current period is included or not, eg.

```
* 1st sub-stake = 900 tokens for 5 periods (in addition to the current period)
```

wdyt? Open to alternative wording as well.

@arjunhassard wdyt about the wording?

vzotova: I'm for describing this somewhere in the first section, because current period always does not count when we talk about locking duration

derekpierre: 👍 - Adding it as an overall statement makes sense.

docs/source/architecture/slashing.rst
```diff
@@ -65,21 +65,21 @@ Penalty Scenarios:
...
 
-    The unlocked tokens can only cover 200 tokens. In the current period, 700 tokens are locked and 800 tokens are locked for the next period. Therefore, we should reduce amount of locked tokens for the next period and leave unchanged locked amount in the current period. The 3rd sub stake suits for this purpose but it's not the shortest one. So we take the 2nd sub stake (the shortest), reduce it to 100 tokens and add new sub stake with 100 tokens which active only in the current period.
+    The unlocked tokens can only cover 200 tokens worth of the penalty. Beyond that, the staker has 700 tokens currently locked and 100 tokens that will lock in the next period, meaning 800 tokens will be locked in total. In this scenario, we should reduce amount of locked tokens for the next period and leave unchanged locked amount in the current period. The 3rd sub-stake would be suitable to be reduced except that it's not the shortest, in terms of its unlock date. Instead, the 2nd sub-stake – the shortest (2 periods until unlock) – is reduced to 100 tokens and a new sub-stake with 100 tokens is added which is only active in the current period.
```
mswilkison: "Instead, the 2nd sub-stake – the shortest (2 periods until unlock) – is reduced to 100 tokens and a new sub-stake with 100 tokens is added which is only active in the current period."

Wait... why is a new sub-stake added? This leaves us with 800 tokens when there should only be 700 remaining.

docs/source/architecture/slashing.rst
```diff
@@ -101,13 +101,13 @@ Penalty Scenarios:
    
 * *Scenario 3*: Staker incurs penalty calculated to be worth **400 tokens**:
 
-    The difference from the previous scenario is that should also decrease locked tokens in the current period. At the first step the 2nd sub stake is reduced to 100 tokens. Next step - adjustment for the next period. The shortest sub stake still the same - the 2nd. And we need to deacrese it from 100 to 0 only for the next period. Will be the same if we change duration of the 2nd sub stake from 2 periods to 1 and the other sub stakes remain unchanged.
+    The difference between this and the previous scenario is that the current period's sum of locked tokens is also reduced. The first step is to reduce the 2nd sub-stake to 100 tokens. Then, the next period is adjusted – the shortest sub-stake is still the 2nd – and it is reduced from 100 to zero for the next period. Notably, this would have the same result if we changed the duration of the 2nd sub-stake from 2 periods to 1 and the other sub-stakes remained unchanged.
```
mswilkison: I think the way we're explaining this is very confusing. We have 1,000 initial tokens and in this and the last example, we slash X amount but the `Result` displays an amount (1,000 - X + 100) which I think will confuse everyone

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,276 @@
+Sub-stakes
+==========
+
+A staker may extend the unlock period for any number of portions of their total stake. This divides the stake into smaller parts, each with a unique unlock date in the future. Stakers may also acquire and lock new tokens. The total stake is represented as the sum of all the different sub-stakes active in a given cycle (new cycle every 24h), which includes locked sub-stakes, and any sub-stakes that have passed their unlock date, and can be freely withdrawn. Each sub-stake has a beginning and duration (lock time). When a staker confirms activity each day and winding down is enabled, the remaining lock time for relevant sub-stakes is reduced.
```
mswilkison: Should we use consistent terminology? Here we use `cycle` - is that the same thing as a `period`?

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,276 @@
...
+
+Flags that affect the sub-stake array
+-------------------------------------
+
+Re-staking
+^^^^^^^^^^
+*Used in methods* : ``confirmActivity()``, ``mint()``
+
+When re-staking is turned off, the number of locked tokens in sub-stakes does not change by itself.
+However, when re-staking is enabled (default) then all staking rewards are re-locked as part of each relevant sub-stake (inside ``confirmActivity()`` and/or ``mint()``).  Consequently, each such sub-stake has an increased locked amount (by reward) and the number of sub-stakes remains unchanged.
```
mswilkison:
```suggestion
However, when re-staking is enabled (default) then all staking rewards are re-locked as part of each relevant sub-stake (inside ``confirmActivity()`` and/or ``mint()``).  Consequently, each such sub-stake has an increased locked amount (by the accrued staking reward) and the number of sub-stakes remains unchanged.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,276 @@
...
+            +---+---+---+---+---+---+---+---+---+---->    
+            | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
+			
+
+
+Winding down
+^^^^^^^^^^^^
+*Used in methods* : ``confirmActivity()``
+
+A disabled "winding down" parameter (default) guarantees that the worker must call ``confirmActivity()`` at least N times after the parameter is enabled to unlock the sub-stake, where N is the locking duration of sub-stake. When disabled, the unlock date for each sub-stakes shifts forward by 1 period after each period i.e. the duration continues to remain the same until the parameter is enabled. Once the "winding down" parameter is enabled, each call to ``confirmActivity()`` (no more than once in a period) leads to a reduction of locking duration for each sub-stake, and the unlock date no longer changes.
```
mswilkison: Is it more intuitive to explain an enabled "winding down" parameter first?

Something like:
```suggestion
An enabled "winding down" parameter means that each call to ``confirmActivity()`` (no more than once in a period) leads to a reduction of locking duration for each sub-stake. It guarantees that the sub-stake will unlock after the worker calls ``confirm-activity()`` at least N times (no more than once in a period), where N is the locking duration of sub-stake. When disabled (default), the unlock date for each sub-stakes shifts forward by 1 period after each period i.e. the duration continues to remain the same until the parameter is enabled.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,276 @@
...
+    A staker has few sub-stakes, worker calls ``сonfirmActivity()`` each period:
+		- Current period: 
+			* 1st sub-stake = 400 tokens for 8 periods
+			* 2nd sub-stake = 100 tokens locked starting from the next period and locked for 5 periods
+		- Next period, if winding down is disabled:  
+			* 1st sub-stake = 400 tokens for 8 periods
+			* 2nd sub-stake = 100 tokens locked starting from the current period and locked for 5 future periods
+		- Next period, if winding down is enabled: 
+			* 1st sub-stake = 400 tokens for 7 periods
+			* 2nd sub-stake = 100 tokens locked starting from the current period and locked for 4 future periods
```
mswilkison: Is this right? Before this period, the 2nd sub-stake hadn't started. So there should be 5 remaining periods?

docs/source/architecture/sub_stakes.rst
mswilkison: I agree that it'd be nice for some additional clarity on this. You can see from my other comments that I'm quite confused about it 😆

docs/source/architecture/slashing.rst
vzotova: when we reduce 2nd sub-stake to 100 then we have 600 locked in current period and 700 in the next one, so to reestablish current period we create new sub-stake

docs/source/architecture/slashing.rst
vzotova: Could you point where is 1,000 - X + 100?

docs/source/architecture/sub_stakes.rst
vzotova: It's correct, will be exactly 5 `confirmActivity()` calls (including 0 period call)

docs/source/architecture/sub_stakes.rst
vzotova: I guess it will be better but something wrong with this particular suggestion (confused by guranties part)

docs/source/architecture/slashing.rst
derekpierre: @vzotova , it's the fact that we start with 1000 tokens, get slashed 400 tokens, but there are 700 (500 + 100 + 100) tokens accounted for in our `Result` statement.

It's unclear why the 2nd sub-stake still ends up with 100 tokens...? even though the full 200 tokens are slashed form it.

derekpierre: Is a timing thing, where the 3rd sub-stake has 100 tokens that won't be locked until the next period - so the same 100 tokens are instead staked for 1 period as part of the 2nd sub-stake?

vzotova: Yep, both current and next period will have 600 locked tokens, 
I was counting on the picture to help explain with this moment

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
vzotova: Changed to `period`
✔️

docs/source/architecture/sub_stakes.rst
vzotova: I'm a little bit stuck with formulating this, could you help me
What I want to say:
* duration (phrases "locked for ") means how many times need to call `confirmActivity()` so this leads to thought that duration is only future periods (current does not count)
* locking in the current period depends on waht situation was in past - if we step 1 period to the previous period and look into next one (which is current period in the current moment), if next period is locked for some sub-stake that means lock in the current period in the current moment and vice versa

derekpierre: So just to ensure that I understand, we define `locking duration` as the number of times `confirmActivity()` needs to be called.

If a stake has a locking duration of 5 periods it means that the stake will be active for periods: 0,1,2,3,4,5 because `confirmActivity()` would be called during periods 0,1,2,3,4 each time confirming activity for the subsequent period (1,2,3,4,5) respectively.

Is this right?

vzotova: almost perfect, 0th period lock is not guaranteed, so lock will be active for periods: 1,2,3,4,5 and 0th period should be specified additionally (which by default is locked and unlocked when said that lock starts in the next period)

vzotova: summing up, the correct explanation:
`If a stake has a locking duration of 5 periods it means that the stake will be active for periods: 1,2,3,4,5 because confirmActivity() would be called during periods 0,1,2,3,4 each time confirming activity for the subsequent period (1,2,3,4,5) respectively.`

docs/source/architecture/sub_stakes.rst
mswilkison:
```suggestion
An enabled "winding down" parameter means that each call to ``confirmActivity()`` (no more than once in a period) leads to a reduction of locking duration for each sub-stake. In other words, the sub-stake will unlock after the worker calls ``confirm-activity()`` at least N times (no more than once in a period), where N is the locking duration of sub-stake. When disabled (default), the unlock date for each sub-stakes shifts forward by 1 period after each period. In other words, the duration continues to remain the same until the "winding down" parameter is enabled.```

Comment (vzotova) [reply docs/source/architecture/sub_stakes.rst]: added!
Also if you could add somewhere that when winding down enabled - unlock date remains the same

File: docs/source/architecture/sub_stakes.rst (hunk 19 above)
Comment (derekpierre) [reply docs/source/architecture/sub_stakes.rst]: Right, so unless specified otherwise (example by stating "*starting from the next period *") the 0th period is considered active.

---

@vzotova what about something like the following at the top of the doc (after the sentence that ends "and can be freely withdrawn.").

```
[and can be freely withdrawn.]

Each sub-stake has a beginning and locking duration. For our terminology, we define `locking duration` to be the required number of times that `confirmActivity()` needs to be called so that the stake is registered as active for the subsequent period. For example, if a stake has a *locking duration of 5 periods* it means that the stake will be active for periods: 1,2,3,4,5 because `confirmActivity()` would be called during periods 0,1,2,3,4 each time confirming activity for the subsequent period (1,2,3,4,5) respectively. Unless the starting period is specified, the stake is considered active for the 0th (current) period by default.
```

Obviously you can reword as needed.

Comment (derekpierre) [reply docs/source/architecture/sub_stakes.rst]: I was also thinking we could remove the sentence at the top (Overview):

` When a staker confirms activity each day and winding down is enabled, the remaining lock time for relevant sub-stakes is reduced.` 

since wind-down will be explained appropriately later on.

Comment (vzotova) [reply docs/source/architecture/sub_stakes.rst]: Changed `active` to `locked` because we introduce `active` in the next paragraph with other meaning. and also `stake` -> `sub-stake` (here it's very important).
About winding down: sometimes better to say twice :) but if you sure - I can remove

File: docs/source/architecture/sub_stakes.rst [hunk 26]
```diff
@@ -0,0 +1,278 @@
...
+
+Each sub-stake has a beginning and locking duration. For our terminology, we define `locking duration` to be the required number of times that `confirmActivity()` needs to be called so that the sub-stake is registered as locked for the subsequent period. For example, if a sub-stake has a *locking duration of 5 periods* it means that the sub-stake will be locked for periods: 1,2,3,4,5 because `confirmActivity()` would be called during periods 0,1,2,3,4 each time confirming activity for the subsequent period (1,2,3,4,5) respectively. Unless the starting period is specified, the sub-stake is considered locked for the 0th (current) period by default. When a staker confirms activity each day and winding down is enabled, the remaining lock time for relevant sub-stakes is reduced.
```
derekpierre:
```suggestion
Each sub-stake has a starting period and a locking duration. For our terminology, we define "locking duration" to be the required number of times that ``confirmActivity()`` needs to be called so that the sub-stake is registered as locked for the subsequent period. For example, if a sub-stake has a *locking duration of 5 periods* it means that the sub-stake will be locked for periods: 1,2,3,4,5 because ``confirmActivity()`` would be called during periods 0,1,2,3,4 each time confirming activity for the subsequent period (1,2,3,4,5) respectively. Unless the starting period is specified, the sub-stake is considered locked for the 0th (current) period by default. When a staker confirms activity each day and winding down is enabled, the remaining lock time for relevant sub-stakes is reduced.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,278 @@
...
+^^^^^^^^^^^^^^^^^^^
+*Methods* : ``deposit(uint256,uint16)``,  ``deposit(address,uint256,uint16)``,  ``lock(uint256,uint16)``
+
+To become a staker, NU tokens must be transferred to the ``StakingEscrow`` contract and locked using one of the ``deposit()`` methods. If the staker already has unlocked tokens within the account in the contract, then the stake can be locked using the ``lock()`` method. If successful, a new element will be created in the array of sub-stakes with the next period as the starting date, and the duration equal to the input parameter.
+
+
+**Example:**
+
+    A staker deposits 900 tokens:
+        * 1st sub-stake = 900 tokens starting from the next period and locked for 5 periods
```
derekpierre: we should use our 'locking duration' term - similar comment for other spots on the page.
```suggestion
        * 1st sub-stake = 900 tokens starting from the next period and a locking duration of 5 periods
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,278 @@
...
+^^^^^^^^^^^^^^^^^
+*Methods* : ``prolongStake(uint256,uint16)``
+
+In order to increase the staking reward, as well as the possibility of obtaining policies with a longer timeframe, stakers can increase the duration of their locked sub-stake using the ``prolongStake()`` method. The number of sub-stakes does not change, but the locked duration for the specified sub-stake will be increased.
+
+**Example:**
+
+    A staker prolongs sub-stake for 2 additional periods up to 7:
+		- Before: 
+			* 1st sub-stake = 900 tokens for 5 periods
```
derekpierre: we should use our 'locking duration' term - similar comment for other spots in the page.

```suggestion
			* 1st sub-stake = 900 tokens with locking duration of 5 periods
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,278 @@
...
+
+
+Flags that affect the sub-stake array
+-------------------------------------
+
+Re-staking
+^^^^^^^^^^
+*Used in methods* : ``confirmActivity()``, ``mint()``
+
+When re-staking is turned off, the number of locked tokens in sub-stakes does not change by itself.
```
derekpierre: Since we used enabled later on, we should use disabled here.
```suggestion
When re-staking is disabled, the number of locked tokens in sub-stakes does not change by itself.
```

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,278 @@
...
+            +---+---+---+---+---+---+---+---+---+---->    
+            | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9
+			
+
+
+Winding down
+^^^^^^^^^^^^
+*Used in methods* : ``confirmActivity()``
+
+An enabled "winding down" parameter means that each call to ``confirmActivity()`` (no more than once in a period) leads to a reduction of locking duration for each sub-stake. In other words, the sub-stake will unlock after the worker calls ``confirm-activity()`` at least N times (no more than once in a period), where N is the locking duration of sub-stake. When disabled (default), the unlock date for each sub-stakes shifts forward by 1 period after each period. In other words, the duration continues to remain the same until the "winding down" parameter is enabled.
```
derekpierre:
```suggestion
An enabled "winding down" parameter means that each call to ``confirmActivity()`` (no more than once in a period) leads to a reduction of the locking duration for each sub-stake. In other words, the sub-stake will unlock after the worker calls ``confirm-activity()`` at least N times (no more than once in a period), where N is the locking duration of sub-stake. When disabled (default), the unlock date for each sub-stakes shifts forward by 1 period after each period. In other words, the duration continues to remain the same until the "winding down" parameter is enabled.
```

docs/source/architecture/slashing.rst
```diff
@@ -26,16 +26,16 @@ The entire stake consists of:
     * tokens which the staker can withdraw at any moment
```
derekpierre:
```suggestion
    * unlocked tokens which the staker can withdraw at any moment
```

docs/source/architecture/slashing.rst
derekpierre: Should we use our "locking duration" term here instead of "lock time"?

docs/source/architecture/slashing.rst
```diff
@@ -26,16 +26,16 @@ The entire stake consists of:
...
+`Sub-stakes`_ get slashed in the order of their remaining lock time, beginning with the shortest – so the first portion of the stake to be slashed is the unlocked portion. After that, if necessary, locked sub-stakes are decreased – the shortest sub-stake is decreased by the required amount; if the adjustment of that sub-stake is insufficient to fulfil the required punishment sum, then the next shortest sub-stake is decreased, and so on. Sub-stakes that begin in the next period are checked separately.
 
 **Example:**
 
     A staker has 1000 tokens:
-        * 1st sub stake = 500 tokens locked for 10 periods
-        * 2nd sub stake = 200 tokens for 2 periods
-        * 3rd sub stake = 100 tokens locked starting from the next period and locked for 5 periods. The 3rd sub stake is locked for the next period but has not yet been used as a deposit for "work" - not until the next period begins.
+        * 1st sub-stake = 500 tokens locked for 10 periods
```
derekpierre: Should we use "locking duration"...? here and in other spots in the document?
```suggestion
        * 1st sub-stake = 500 tokens with a locking duration of 10 periods
```

docs/source/architecture/slashing.rst
derekpierre: We'll need to explain this timing thing a more clearly. The diagram helps, but still needs a solid explanation as a lead in.

vzotova: For me it's difficult to rephrase this, too many time I work with slashing :)
Also keep in mind that ideally user will read "slashing" after "sub-stakes"

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

docs/source/architecture/slashing.rst
vzotova: ✔️

docs/source/architecture/slashing.rst
vzotova: For explanation of slashing we don't use strict periods, just comparison between sub-stakes, so I see that it's ok to write something simple like "lock for" or "lock time". But again if we want to have strict termenology everywhere - I'll change this

derekpierre: That's fair - changing it might introduce more confusion than necessary.

docs/source/architecture/slashing.rst
derekpierre: Can we add something general to the beginning explanation of our slashing methodology?

Basically what I've gleaned from our discussion is that we make every attempt to limit idle "not yet locked" tokens after slashing; these idle tokens will be used in earlier periods before their original lock time, to supplement a slashed locked stake to attempt to keep the pre-slashed total locked stake the same for earlier periods.

Assuming the above is correct, perhaps we can add something like:

```
Once a stake is slashed, tokens that are not currently locked but are set to be locked during future periods, i.e. currently "idle", can be used to replenish/supplement a slashed stake in an attempt to maintain the original pre-slashed total locked tokens for a period.  These "idle" tokens can only be used in periods before their starting lock period, and if necessary, a separate sub-stake is created to facilitate this action.
```

In our scenarios that involve the timing situation above (namely Scenario 2 & 3), we should indicate the available "idle" tokens that are then used to supplement slashed stakes, and the new sub-stake if created.

vzotova: I'm so confused with this explanation, I found somehow wrong some ideas from it. 
Next try :)
First I whant to point that stake is time function

I'll try to explain abstraction and how I was thinking when wrote code:
we have initial figure that consists of several sub-stakes + unlocked tokens, when we slash we move imaginary line from the top to bottom and this line "destroys" or "presses" sub-stakes starting from the shortest, but we can have sub-stakes only as rectangles so when imaginary line pushes - whole sub-stake becomes smaller

next step is restoring some sub-stakes where possible to minimize damage (time-issues in explanation come from this part): we still have this imagimary line that destroyed some sub-stakes and stoped but we will add some small sub-stakes which do not cross this line and restores some locks - this source for 4th sub-stake in 2nd scenario. 
In 3rd scenario imaginary line works as just press, 3rd sub-stake "stronger" then "second" (because longer) and 3rd "cuts" part of 2nd under the press. 

Not sure that'll help :)

derekpierre: I think we are in line with the understanding, seems like just figuring out a wording that works 😅 .

In Scenario 2: 
the 2nd sub-stake is reduced from 200 to 100 tokens, and the 100 "idle" tokens from the 3rd sub-stake and used to create the 4th sub-stake for staking for 1 period.

In Scenario 3:
The 2nd sub-stake is emptied from 200 to 0 tokens, and the 100 "idle" tokens from the 3rd sub-stake are used in the 2nd sub-stake directly (since it has no tokens) for 1 period.

Correct?

vzotova: can't say that is correct 
There is no such thing like "idle" tokens
Maybe this picture will show why I'm so confused:
```
            stake
            ^
            |
         700+-----+----+
            |     | 3rd|
         600|     +----+-------------+
            |    2nd   |     3rd     |
         500+----------+-------------+----------+
            |                                   |
            |               1st                 |
            |                                   |   period
            +-----------------------------------+--->
```

vzotova: I've added more diagrams in slashing doc

derekpierre:
@vzotova , see https://github.com/vzotova/nucypher/pull/5 as my initial stab at putting into words our discussion about sub-stake fixed token amount across periods and the nuances of the slashing algorithm.

docs/source/architecture/slashing.rst
derekpierre: @vzotova thanks for adding the intermediate steps. Let's add a little explanation to the last diagram for Scenario 2 explaining our creation of the 4th sub-stake. You can reuse some words from my attempt (😊 ) of an explanation in https://github.com/vzotova/nucypher/pull/5.

Similar kind of thing for Scenario 3 i.e. add an explanation that the 2nd sub-stake is emptied, but gets replenished for one period. So we didn't "create" a new sub-stake, but we did reorganize tokens so that the 2nd sub-stake had tokens 1 period.

docs/source/architecture/slashing.rst
```diff
@@ -19,23 +19,43 @@ TBD (https://github.com/nucypher/nucypher/issues/803)
 How slashing affects stake
 --------------------------
 
-The goal of slashing is to reduce the number of tokens that belongs to a staking offender.
-In this case, the main task is not to violate the logic of locking tokens.
-The entire stake consists of:
+.. _`Sub-stakes`: https://docs.nucypher.com/en/latest/architecture/sub_stakes.html
 
-    * tokens which the staker can withdraw at any moment
+The goal of slashing is to reduce the number of tokens that belongs to a staking offender. While the reduction of tokens is intended to be punitive, once the tokens have been slashed, the slashing algorithm attempts to preserve the most efficient use of the offenders' remaining tokens based on their existing `Sub-stakes`_. The main task is not to violate the logic of locking tokens.
```
mswilkison: "The main task is not to violate the logic of locking tokens."

Can you explain what you mean by this?

docs/source/architecture/slashing.rst
```diff
@@ -19,23 +19,43 @@ TBD (https://github.com/nucypher/nucypher/issues/803)
...
+In terms of how the stake is slashed, unlocked tokens are the first portion of the stake to be slashed. After that, if necessary, locked sub-stakes are decreased in order based on their remaining lock time, beginning with the shortest. The shortest sub-stake is decreased, and if the adjustment of that sub-stake is insufficient to fulfil the required punishment sum, then the next shortest sub-stake is decreased, and so on. Sub-stakes that begin in the next period are checked separately.
+
+Sub-stakes for past periods cannot be slashed, so only the periods from now onward can be slashed. However, by design sub-stakes can't have a starting period that is after the next period, so all future periods after the next period will always have an amount of tokens less than or equal to the next period. The current period still needs to be checked since its stake may be different than the next period. Therefore, only the current period and the next period need to be checked for slashing.
```
mswilkison:
```suggestion
Sub-stakes for past periods cannot be slashed, so only the periods from the current period onward can be slashed. However, by design sub-stakes can't have a starting period that is after the next period, so all future periods after the next period will always have an amount of slashable tokens less than or equal to the next period. The current period still needs to be checked since its stake may be different than the next period. Therefore, only the current period and the next period need to be checked for slashing.
```

docs/source/architecture/slashing.rst
```diff
@@ -19,23 +19,43 @@ TBD (https://github.com/nucypher/nucypher/issues/803)
...
+    d. The above property of sub-stakes means that there is the possibility that the total amount of locked tokens for a particular period could be reduced to even lower than the ``max_allowed_stake`` when unnecessary. Therefore, the slashing algorithm may create new sub-stakes on the staker's behalf to utilize tokens in the earlier period, when a sub-stake is needed to be reduced to an even lower value because of the next period. In the example above in c), the sub-stake was reduced to 5 tokens because of period ``n+1``, so there are 5 "extra" tokens `(10 - 5)` available in period ``n`` that can still be staked; hence, a new sub-stake with 5 tokens would be created to utilize these tokens in period ``n``. This benefits both the staker, by ensuring that their remaining tokens are efficiently utilized, and the network by maximizing its health.
```
mswilkison:
```suggestion
    d. The above property of sub-stakes means that there is the possibility that the total amount of locked tokens for a particular period could be reduced to even lower than the ``max_allowed_stake``. Therefore, the slashing algorithm may create new sub-stakes on the staker's behalf to utilize tokens in the earlier period, when a sub-stake is needed to be reduced to an even lower value because of the next period. In the example above in c), the sub-stake was reduced to 5 tokens because of period ``n+1``, so there are 5 "extra" tokens `(10 - 5)` available in period ``n`` that can still be staked; hence, a new sub-stake with 5 tokens would be created to utilize these tokens in period ``n``. This benefits both the staker, by ensuring that their remaining tokens are efficiently utilized, and the network by maximizing its health.
```

docs/source/architecture/slashing.rst
```diff
@@ -19,23 +19,43 @@ TBD (https://github.com/nucypher/nucypher/issues/803)
...
 
 **Example:**
 
     A staker has 1000 tokens:
-        * 1st sub stake = 500 tokens locked for 10 periods
-        * 2nd sub stake = 200 tokens for 2 periods
-        * 3rd sub stake = 100 tokens locked starting from the next period and locked for 5 periods. The 3rd sub stake is locked for the next period but has not yet been used as a deposit for "work" - not until the next period begins.
+        * 1st sub-stake = 500 tokens locked for 10 periods
+        * 2nd sub-stake = 200 tokens for 2 periods
+        * 3rd sub-stake = 100 tokens locked starting from the next period and locked for 5 periods. The 3rd sub-stake is locked for the next period but has not yet been used as a deposit for "work" - not until the next period begins.
```
mswilkison:
```suggestion
        * 3rd sub-stake = 100 tokens locked starting from the next period and locked for 5 periods. The 3rd sub-stake is locked for the next period but won't be used as a deposit for "work" until the next period begins.
```

docs/source/architecture/slashing.rst
vzotova: Partially added, removed `slashable` form `amount of slashable tokens less than or equal to the next period` because it's important that all tokens in the stake less or equal to stake in the next period

docs/source/architecture/slashing.rst
vzotova: ✔️

docs/source/architecture/slashing.rst
vzotova: ✔️

docs/source/architecture/slashing.rst
vzotova: we have such structure for staker in `StakingEscrow`
```
struct StakerInfo {
        uint256 value;
...
        SubStakeInfo[] subStakes;
    }
```
Technically slashing is just deacresing `value` field by slashing amount, because this `value` means how many tokens belong to staker. But what we call "slashing" here is to make all data still consistent.
If we just deacrese `value` and untouch `subStakes` array it could lead to such situation: sum if sub-stakes in some period in a future will be more than `value` and it means when staker will mint reward for that period he will get more tokens than he desirves. Also could be some other potential problems in the code but that is the main. So our task is to adjust values in  `subStakes` array

derekpierre: What about?

```suggestion
Slashing aims to reduce the overall number of tokens that belongs to a staking offender. While the reduction of tokens is intended to be punitive, once the tokens have been slashed, the slashing algorithm attempts to preserve the most efficient use of the offenders' remaining tokens based on their existing `Sub-stakes`_. The main goal is to decrease the tokens belonging to the staker while considering how tokens are allocated across different sub-stakes.
```

vzotova: I like it, what do you think @mswilkison ?

mswilkison: 👍

vzotova: ✔️

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,290 @@
...
+            +---+---+---+---+---+---+---+---+---->
+            + 0 + 1 + 2 + 3 + 4 + 5 + 6 + 7 + 8
+			
+			
+Sub-stake division
+^^^^^^^^^^^^^^^^^^
+*Methods* : ``divideStake(uint256,uint256,uint16)``
+
+If necessary, stakers can extend the locking duration for only a portion of their tokens in a sub-stake by using the ``divideStake()`` method. This method splits a sub-stake into two elements: the old sub-stake with the reduced locked amount and the new sub-stake with the specified amount. The new sub-stake has the specified locked amount and an extended lock duration, based on the specified number of periods, with the same start period as the old sub-stake.
+
```
cygnusv: A fundamental question for me is: why would a staker want to divide her stake? What's the use case or goal behind that? Can we give an example or motivation?

docs/source/architecture/sub_stakes.rst
```diff
@@ -0,0 +1,290 @@
+Sub-stakes
+==========
+
+A staker may extend the unlock period for any number of portions of their total stake. This divides the stake into smaller parts, each with a unique unlock date in the future. Stakers may also acquire and lock new tokens. The total stake is represented as the sum of all the different sub-stakes active in a given period (new period every 24h), which includes locked sub-stakes, and any sub-stakes that have passed their unlock date, and can be freely withdrawn. 
+
+Each sub-stake has:
+
+    * **Locked amount of tokens**
+
+        The locked amount of tokens is fixed for all applicable periods i.e. it is not possible for a single sub-stake to have a different amount of tokens locked for different periods. To facilitate such a scenario, sub-stakes would need to be divided, see `Sub-stake division`_.
```
cygnusv:
```suggestion
        The locked amount of tokens is fixed for all applicable periods, i.e., it is not possible for a single sub-stake to have a different amount of tokens locked for different periods. To facilitate such a scenario, sub-stakes would need to be divided, see `Sub-stake division`_.
```

docs/source/architecture/sub_stakes.rst
cygnusv: Also, substake division needs a link from the staking guide. See
https://github.com/nucypher/nucypher/commit/36204bd86cf692037dbc1bb0278be14700a50ae8

docs/source/architecture/slashing.rst
cygnusv: What about this comment from @derekpierre ?

vzotova: Actually that's resolved, see one resolved comment above

docs/source/architecture/sub_stakes.rst
vzotova: This supposed to be an answer: `can extend the locking duration for only a portion of their tokens in a sub-stake`. Any suggestions how to improve?

docs/source/architecture/sub_stakes.rst
vzotova: ✔️

PR #1373, opened 2019-10-01
[WIP] MultiSig API

KPrasch:
- [ ] MultiSigAgent
    - [ ] Agent
    - [ ] Tests
- [ ] MultiSigDeployer
    - [ ] Deployer
    - [ ] Tests
- [ ] MultiSig Actors
    - [ ] Executive
    - [ ] Trustee
    - [ ] Tests
- [ ] MultiSig Entities
    - [ ] Authorization
    - [ ] Board
    - [ ] Tests
- [ ] MultiSig CLI
    - [ ] Entry Point
    - [ ] Tests

PR #1560, opened 2020-01-12
Foundations for MultiSig API and tooling.

### LINKED ISSUE #1194: What about block confirmations?
Currently, we just care about getting successful receipts for transactions, but in order to have a higher assurance about the real status of the transaction, we should wait for a certain number of block confirmations. 

A secondary question is how many confirmations do we need, but that's just a parameter we can adjust.

cygnusv:
Based on #1373 

- [x] MultiSigAgent
    - [x] Agent
    - [x] Tests
- [x] MultiSigDeployer
    - [x] Deployer
    - [x] Tests
- [ ] MultiSig Actors
    - [ ] Executive
    - [ ] Trustee
    - [ ] Tests
- [ ] MultiSig Entities
    - [ ] Authorization
    - [ ] Board
    - [ ] Tests
- [x] MultiSig CLI
    - [x] Entry Point
    - [ ] Tests

Fixes:
* Closes #1194: Implements a basic block confirmation option in `BlockchainInterface.send_transaction()`


nucypher/blockchain/eth/agents.py
```diff
@@ -1122,3 +1124,90 @@ def dump(self) -> list:
...
+    def threshold(self) -> int:
+        threshold = self.contract.functions.required().call()
+        return threshold
+
+    @validate_checksum_address
+    def is_owner(self, checksum_address: str) -> bool:
+        result = self.contract.functions.isOwner(checksum_address).call()
+        return result
+
+    def add_owner(self, new_owner_address: str, sender_address: str) -> dict:
```
derekpierre: Should we validate checksum address here? And  on Line 1182, and 1198?

nucypher/cli/commands/deploy.py
```diff
@@ -477,6 +520,59 @@ def transfer_tokens(general_config, actor_options, target_address, value):
     paint_receipt_summary(emitter=emitter, receipt=receipt)
 
 
+@deploy.command(name='multisig')
```
derekpierre: Agreed that it may make sense to move into its own module - at a later date.

nucypher/blockchain/eth/agents.py
KPrasch: Let's allow for this in a follow up PR.

PR #1562, opened 2020-01-14
Removing PE from WorkLock

vzotova:
Changes time-based lock for claimed tokens to stake-based lock (simply by removing PE and directly staking tokens inside `claim()` method)
I want to emphasize - it's not work-based lock because doesn't matter how many work will be produced by staker.

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
@@ -6,7 +6,6 @@ import "zeppelin/token/ERC20/SafeERC20.sol";
 import "zeppelin/utils/Address.sol";
 import "contracts/NuCypherToken.sol";
 import "contracts/StakingEscrow.sol";
-import "contracts/staking_contracts/AbstractStakingContract.sol";
 import "contracts/lib/AdditionalMath.sol";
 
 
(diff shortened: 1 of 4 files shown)


michwill:
It's still work based in a sense that if you don't stake, you never unlock

vzotova:
Example - we have weird coefficients or it's end of inflation rate curve, then staker can't mint new tokens (which are "work") but can stake

michwill:
Ahh, I see. Well, practically, they can do enough work to unlock though

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -55,56 +52,51 @@ contract WorkLock {
...
 
     /**
     * @param _token Token contract
     * @param _escrow Escrow contract
-    * @param _router Router contract
     * @param _startBidDate Timestamp when bidding starts
     * @param _endBidDate Timestamp when bidding will end
     * @param _boostingRefund Coefficient to boost refund ETH
-    * @param _lockingDuration Duration of tokens locking
+    * @param _stakingPeriods Amount of periods during which tokens will be locked while claiming
```
cygnusv:
```suggestion
    * @param _stakingPeriods Amount of periods during which tokens will be locked after claiming
```

tests/blockchain/eth/contracts/main/worklock/test_worklock.py
```diff
@@ -29,22 +26,6 @@ def token(testerchain, token_economics, deploy_contract):
...
-@pytest.fixture()
-def router(testerchain, deploy_contract):
-    staking_interface, _ = deploy_contract('StakingInterfaceMock')
-    secret = os.urandom(32)
-    secret_hash = keccak(secret)
-    contract, _ = deploy_contract('StakingInterfaceRouter', staking_interface.address, secret_hash)
-    return contract
-
-
-def next_address(testerchain, worklock):
```
KPrasch: :joy:

vzotova: Yeah, goodbye cool stuff, I'll miss you

PR #1572, opened 2020-01-16
Fixes #1435 - Rethink datetime to period utils.

### LINKED ISSUE #1435: Rethink nucypher/blockchain/eth/utils/datetime_at_period
Wait... `else` works only when `delta_periods == 0`!

Did you mean simply always `now + timedelta(days=delta_periods)`? Timedelta understands negative time too.

Also why days when we have `seconds_per_period` (what if it's not 1 day)? Maybe `maya.timedelta(seconds=1) * seconds_per_period * delta_periods` or something?

_Originally posted by @michwill in https://github.com/nucypher/nucypher/pull/1402_

KPrasch:
Thanks to @michwill and @vzotova for the solution!

### DIFF
--- a/nucypher/blockchain/eth/utils.py
+++ b/nucypher/blockchain/eth/utils.py
@@ -48,15 +48,7 @@ def datetime_at_period(period: int, seconds_per_period: int, start_of_period: bo
         now = maya.now()
         current_period = datetime_to_period(datetime=now, seconds_per_period=seconds_per_period)
         delta_periods = period - current_period
-
-        # +
-        if delta_periods:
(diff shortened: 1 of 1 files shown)


nucypher/blockchain/eth/utils.py
```diff
@@ -14,6 +14,7 @@
 You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
+from datetime import datetime
```
michwill: This import is unused because we use everything through maya

KPrasch: :heavy_check_mark:

PR #1602, opened 2020-01-21
Adds support to --rate when granting, tweaks to the Staking Guide, and other beasts

### LINKED ISSUE #1262: Alice.controller does not parse `value` as part of the request when granting

### LINKED ISSUE #1601: Documentation says re-staking is disabled by default, while according to the team it is enabled
Need to update: https://docs.nucypher.com/en/latest/guides/staking_guide.html#manage-automatic-reward-re-staking

cygnusv:
* Fixes #1601 
* Closes #1262

### DIFF
--- a/nucypher/characters/control/interfaces.py
+++ b/nucypher/characters/control/interfaces.py
@@ -107,6 +107,7 @@ class AliceInterface(CharacterPublicInterface, AliceSpecification):
               n: int,
               expiration: maya.MayaDT,
               value: int = None,
+              rate: int = None,
               ) -> dict:
 
         from nucypher.characters.lawful import Bob
@@ -118,6 +119,7 @@ class AliceInterface(CharacterPublicInterface, AliceSpecification):
                                           m=m,
                                           n=n,
                                           value=value,
+                                          rate=rate,
                                           expiration=expiration)
 
         response_data = {'treasure_map': new_policy.treasure_map,
--- a/nucypher/characters/control/serializers.py
+++ b/nucypher/characters/control/serializers.py
@@ -123,7 +123,8 @@ class AliceControlJSONSerializer(CharacterControlJSONSerializer, MessageHandlerM
                             label=request['label'].encode(),
                             m=request['m'],
                             n=request['n'],
-                            value=request.get('value'),  # optional / unused in federation mode
+                            value=request.get('value'),  # optional / unused in federated mode
+                            rate=request.get('rate'),  # optional / unused in federated mode
                             expiration=maya.MayaDT.from_iso8601(iso8601_string=request['expiration']))
         return parsed_input
 
--- a/nucypher/cli/commands/alice.py
+++ b/nucypher/cli/commands/alice.py
@@ -45,6 +45,7 @@ option_bob_verifying_key = click.option(
 )
 
 option_pay_with = click.option('--pay-with', help="Run with a specified account", type=EIP55_CHECKSUM_ADDRESS)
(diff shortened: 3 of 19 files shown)


nucypher/cli/commands/alice.py
```diff
@@ -373,7 +398,7 @@ def grant(general_config,
     }
 
     if not ALICE.federated_only:
-        grant_request.update({'value': value})
+        grant_request.update({'value': value, 'rate': rate})
```
derekpierre: @cygnusv within `grant`, do we want either `value` or `rate` to be provided but not both? i.e. perform an xor check of those cli options.

tuxxy: This. You can't really provide both. You need either or.

nucypher/characters/control/interfaces.py
```diff
@@ -107,6 +107,7 @@ def grant(self,
               n: int,
               expiration: maya.MayaDT,
               value: int = None,
+              rate: int = None,
```
tuxxy: :+1:

nucypher/characters/control/serializers.py
```diff
@@ -123,7 +123,8 @@ def parse_grant_input(request: dict):
                             label=request['label'].encode(),
                             m=request['m'],
                             n=request['n'],
-                            value=request.get('value'),  # optional / unused in federation mode
+                            value=request.get('value'),  # optional / unused in federated mode
+                            rate=request.get('rate'),  # optional / unused in federated mode
                             expiration=maya.MayaDT.from_iso8601(iso8601_string=request['expiration']))
```
tuxxy: Unrelated to this PR, but we need a better UX for specifying this in the CLI.

nucypher/cli/commands/alice.py
```diff
@@ -174,8 +194,14 @@ def create_character(self, emitter, config_file, json_ipc, load_seednodes=True):
         config = self.config_options.create_config(emitter, config_file)
 
         client_password = None
-        if not config.federated_only:
-            if (not self.hw_wallet or not config.dev_mode) and not json_ipc:
+        eth_password_is_needed = not config.federated_only and not self.hw_wallet and not config.dev_mode
+        if eth_password_is_needed:
```
tuxxy: :+1:

nucypher/cli/commands/alice.py
```diff
@@ -218,7 +244,7 @@ def alice():
 @option_light
 @option_m
 @option_n
-@click.option('--rate', help="Policy rate per period in wei", type=click.FLOAT)
+@option_rate
```
tuxxy: :+1:

nucypher/cli/commands/alice.py
```diff
@@ -349,15 +375,14 @@ def grant(general_config,
           bob_encrypting_key, bob_verifying_key, label,
 
           # Other
-          m, n, expiration, value,
+          m, n, expiration, value, rate,
```
tuxxy: How does this work if you provide both?

nucypher/cli/commands/alice.py
cygnusv: That's already checked here, in `BlockchainPolicy`:
https://github.com/nucypher/nucypher/blob/345a09e2f83d2f6c288bb4171337571b16f326f8/nucypher/policy/policies.py#L507

nucypher/characters/control/serializers.py
cygnusv: Opened #1606

nucypher/cli/commands/alice.py
cygnusv: It fails in `BlockchainPolicy`. See https://github.com/nucypher/nucypher/pull/1602#discussion_r369300611

PR #1611, opened 2020-01-22
Small fixes to grant: input validation, correct formatting of stakers' data from StakingEscrow

### LINKED ISSUE #1608: Perhaps raise BadArgumentUsage when 'rate' and 'value' are provided for Alice.grant in the CLI
Related to #1602 .

Perhaps we should raise a `click.BadArgumentUsage` when both 'rate' and 'value' are provided for Alice.grant since only one is allowed (xor condition). 

Instead of currently having the `InvalidPolicyValue` exception raised, the `click.BadArgumentUsage` would be handled by the CLI more cleanly.

cygnusv:
* Fixes #1608


nucypher/cli/commands/alice.py
```diff
@@ -387,6 +388,14 @@ def grant(general_config,
 
     ALICE = character_options.create_character(emitter, config_file, general_config.json_ipc)
 
+    # Input validation
+    if ALICE.federated_only:
+        if any((value, rate)):
+            raise click.BadOptionUsage(option_name="--value, --rate",
+                                       message="Can't use --value or --rate with a federated Alice.")
+    elif not (bool(value) ^ bool(rate)):
+        raise click.BadOptionUsage(option_name="--rate", message="Can't use --value if using --rate")
```
derekpierre: Since we are saying "Can't use --value...", should this line instead be (?):
```suggestion
        raise click.BadOptionUsage(option_name="--value", message="Can't use --value if using --rate")
```

or alternatively

```suggestion
        raise click.BadOptionUsage(option_name="--rate", message="Can't use --rate if using --value")
```

PR #1622, opened 2020-01-25
Inline CLI character configuration

KPrasch:
Enables inline display and edit of default configuration values for all characters and entities.

Refactors `view` commands to `config`, Usage examples:

- view the existing config: `nucypher ursula config `
- set the default network: `nucypher stake config --network cassandra`
- change the worker address: `nucypher ursula config --worker-address <ADDR>`
- set federated-only to False: `nucypher ursula config --decentralized`
- modify the default provider: `nucypher ursula config --provider <URI>`
- set multiple values at once: `nucypher alice config --pay-with --m 2 --n 3`


nucypher/cli/commands/alice.py
```diff
@@ -358,31 +378,27 @@ def derive_policy_pubkey(general_config, label, character_options, config_file):
...
 @alice.command()
-@click.option('--bob-encrypting-key', help="Bob's encrypting key as a hexadecimal string", type=click.STRING,
-              required=True)
-@option_bob_verifying_key
-@option_label(required=True)
-@option_m
-@option_n
-@option_rate
-@click.option('--expiration', help="Expiration Datetime of a policy", type=click.STRING)  # TODO: click.DateTime()
-@click.option('--value', help="Total policy value (in Wei)", type=types.WEI)
```
mswilkison: Do we no longer expose `value` as a CLI command?

nucypher/cli/commands/alice.py
```diff
@@ -96,6 +96,12 @@ def __init__(
         self.discovery_port = discovery_port
         self.registry_filepath = registry_filepath
         self.middleware = middleware
+        self.poa = poa
+        self.light = light
+        self.m = m
+        self.n = n
+        self.rate = rate
```
cygnusv: This implies all these values are going to be applicable for all Alice commands. I also thought about doing something like this in my last PR, but in the end I decided against it because most of commands don't need them.

KPrasch: Yeah - this is indeed a bit awkward, I am having a hard time grouping the grant options more locally while still having the proper state to construct a configuration.  Perhaps @fjarri can take a look?

KPrasch: At the end of the day I'm trying to follow the configuration pattern from the other modules here, since m, n, rate are configuration values.

nucypher/cli/commands/alice.py
KPrasch: I moved it into one of the group decorators `group_character_options`.

nucypher/cli/commands/alice.py
fjarri: I don't think adding these options to the group is justified. `poa`, `light` and `rate` are only used by the `init` command, and `m` and `n` used in `grant` are technically not the same ones used to generate the config (that is, in `grant` they are not used for `generate()`, but for a different purpose).

nucypher/config/characters.py
```diff
@@ -157,9 +157,10 @@ def __init__(self,
         super().__init__(*args, **kwargs)
         self.m = m or self.DEFAULT_M
         self.n = n or self.DEFAULT_N
-        if not self.federated_only:
-            self.rate = rate
-            self.duration_periods = duration_periods
+
+        # if not self.federated_only:  # TODO: why not?
```
fjarri: Probably deserves an issue, or a resolution of some sort.

KPrasch:
> I don't think adding these options to the group is justified. poa, light and rate are only used by the init command, and m and n used in grant are technically not the same ones used to generate the config

I dont see a problem with using these here, and in fact I think it follows our new conventions here better. In this PR - The *only* place these options are implemented is in `init` and `grant` (try using the `--help`  commands).  @fjarri - Why do you say it's not the same option?  I see them as being identical.  Even though it's used to set the default with one action, but override it in another does not make it an entirely different option, In fact I view the re-usage as a strength in simplicity.

nucypher/config/characters.py
KPrasch: With the tests passing I'm inclined to just remove the TODO. objections?

nucypher/cli/commands/alice.py
KPrasch: > This implies all these values are going to be applicable for all Alice commands.

How so?

nucypher/cli/commands/alice.py
```diff
@@ -65,15 +65,15 @@
 
 option_pay_with = click.option('--pay-with', help="Run with a specified account", type=EIP55_CHECKSUM_ADDRESS)
 option_rate = click.option('--rate', help="Policy rate per period (in wei)", type=types.WEI)  # TODO: Is wei a sane unit here? Perhaps gwei?
+option_duration_periods = click.option('--duration-periods', help="Policy duration in periods", type=click.FLOAT)
```
derekpierre: Should this be of type INT instead of FLOAT?

nucypher/cli/commands/alice.py
```diff
@@ -358,31 +378,27 @@ def derive_policy_pubkey(general_config, label, character_options, config_file):
...
-@click.option('--expiration', help="Expiration Datetime of a policy", type=click.STRING)  # TODO: click.DateTime()
-@click.option('--value', help="Total policy value (in Wei)", type=types.WEI)
-@group_character_options
 @option_config_file
 @group_general_config
+@group_character_options
+@option_label(required=True)
+@option_bob_verifying_key
+@click.option('--bob-encrypting-key', help="Bob's encrypting key as a hexadecimal string", type=click.STRING, required=True)
+@click.option('--expiration', help="Expiration Datetime of a policy", type=click.STRING)  # TODO: click.DateTime()
```
derekpierre: Should we indicate the expected format of expiration date? - I think it's supposed to be in iso8601

nucypher/cli/commands/alice.py
```diff
@@ -392,13 +408,13 @@ def grant(general_config,
...
-        'm': m,
-        'n': n,
+        'm': config_options.m,
+        'n': config_options.n,
         'expiration': expiration,
     }
 
     if not ALICE.federated_only:
-        grant_request.update({'value': value, 'rate': rate})
+        grant_request.update({'rate': config_options.rate})
```
derekpierre: I thought we allowed either `value` or `rate`? Should we check which one is provided and then add one to the request?

nucypher/cli/types.py
```diff
@@ -37,7 +42,7 @@ def convert(self, value, param, ctx):
         try:
             _address = ip_address(value)
         except ValueError as e:
-            self.fail(str(e))
+            self.fail("Invalid IP Address")
```
derekpierre: Nice to know what the problem was:

```suggestion
            self.fail(f"Invalid IP Address: {str(e)}")
```

nucypher/cli/types.py
```diff
@@ -27,7 +27,12 @@ class ChecksumAddress(click.ParamType):
     name = 'checksum_address'
 
     def convert(self, value, param, ctx):
-        return to_checksum_address(value=value)  # TODO: More robust validation here?
+        try:
+            value = to_checksum_address(value=value)
+        except ValueError as e:
+            self.fail("Invalid ethereum address")
```
derekpierre: Nice to know what the problem was:

```suggestion
            self.fail(f"Invalid ethereum address: {str(e)}")
```

nucypher/cli/commands/alice.py
cygnusv: @mswilkison had the same comment. It seems that @KPrasch's intention is to disable value altogether, I guess for UX reasons. I must say I agree with this, since value is not very easy to compute for an end-user.

nucypher/cli/commands/alice.py
cygnusv:
```
└─ $ nucypher alice run --help
Usage: nucypher alice run [OPTIONS]

  Start Alice's web controller.

Options:
[...some lines removed...]
  --duration-periods FLOAT        Policy duration in periods
  --rate INTEGER RANGE            Policy rate per period (in wei)
  --n INTEGER                     N-Total KFrags
  --m INTEGER                     M-Threshold KFrags
  --light                         Indicate that node is light
  --poa                           Inject POA middleware
[...some lines removed...]
```
It happens the same for other commands, such as `public-keys` or `destroy`, which doesn't have much sense.

cygnusv: I don't get your assertion "_The only place these options are implemented is in init and grant_". These options are present in `AliceConfigOptions`, which in turn is part of `AliceCharacterOptions`, which are present in most of the commands via the `@group_character_options` decorator. 

Having said that, I was in need for having the `--light` option also present during granting, so Alice could decide to run as light, overriding her config if necessary.

nucypher/config/base.py
```diff
@@ -243,3 +243,13 @@ def deserialize(cls, payload: str, deserializer=json.loads) -> dict:
             raise cls.OldVersion(f"Configuration file is the wrong version "
                                  f"Expected version {cls.VERSION}; Got version {version}")
         return deserialized_payload
+
+    def update(self, filepath: str = None, modifier: str = None, **updates):
+        for field, value in updates.items():
+            try:
+                getattr(self, field)
+            except AttributeError:
+                raise self.ConfigurationError(f"Cannot update '{field}'. It is an invalid configuration field,")
```
derekpierre:
```suggestion
                raise self.ConfigurationError(f"Cannot update '{field}'. It is an invalid configuration field.")
```

vepkenez:
@KPrasch I don't understand why we configure characters in /cli/.  Is what is happening here only relevant to CLI invocations?

nucypher/cli/types.py
KPrasch: Meh - I think I prefer to just fail at this eager level.

nucypher/config/base.py
derekpierre: @KPrasch ^

KPrasch: :heavy_check_mark:

nucypher/cli/options.py
```diff
@@ -36,11 +36,11 @@
 option_db_filepath = click.option('--db-filepath', help="The database filepath to connect to", type=click.STRING)
 option_dry_run = click.option('--dry-run', '-x', help="Execute normally without actually starting the node", is_flag=True)
 option_etherscan = click.option('--etherscan/--no-etherscan', help="Enable/disable viewing TX in Etherscan")
-option_federated_only = click.option('--federated-only', '-F', help="Connect only to federated nodes", is_flag=True, default=None)
+option_federated_only = click.option('--federated-only/--decentralized', '-F', help="Connect only to federated nodes", is_flag=True, default=None)
```
cygnusv: 🚀

Issue #977, opened 2019-05-15
install-solc script fails on mac because libpcre is missing


derekpierre:
The script fails to set executable permissions even though it states that installation was successful:
```
 [9] → ./scripts/installation/install_solc.sh
Downloading solidity compiler binary to: /Users/<user>/.local/share/virtualenvs/nucypher-X1j1AHHb/bin/solc
dyld: Library not loaded: /usr/local/opt/pcre/lib/libpcre.1.dylib
  Referenced from: /usr/local/bin/wget
  Reason: image not found
./scripts/installation/install_solc.sh: line 8: 60770 Abort trap: 6           wget "https://github.com/ethereum/solidity/releases/download/v${SOLC_VER}/solc-static-linux" -O ${SOL_BIN_PATH}
Setting executable permission on /Users/<user>/.local/share/virtualenvs/nucypher-X1j1AHHb/bin/solc
chmod: /Users/<user>/.local/share/virtualenvs/nucypher-X1j1AHHb/bin/solc: No such file or directory
Successfully Installed solc 0.5.3
```

derekpierre:
Installing `brew install pcre` fixed the problem.

derekpierre:
The script is platform specific for linux and windows but not mac. 

```
OSError: [Errno 8] Exec format error: '/Users/derek/.local/share/virtualenvs/nucypher-X1j1AHHb/bin/solc'
```

I had to install the solidity compiler using `brew`, see https://solidity.readthedocs.io/en/v0.5.3/installing-solidity.html, and then replaced the copy of the solc binary in the virtual dev from `pipenv shell` with the freshly installed copy.

```
$ which solc
/usr/local/bin/solc

$ cp /usr/local/bin/solc /Users/<user>/.local/share/virtualenvs/nucypher-X1j1AHHb/bin/solc
```

vzotova:
@derekpierre Hey, could you try again? I hope py-solc-x can handle it

derekpierre:
Bingo - installation worked!

```bash
derek @ Dereks-MacBook-Pro-2 ~/Documents/Github/repos/nucypher
 [5] → pipenv run install-solc
Fetched solc version v0.6.7 from source configuration
Installing solidity dependencies on macOS 10.15 Catalina.
Updated 2 taps (homebrew/core and ethereum/ethereum).
...
solc, the solidity compiler commandline interface
Version: 0.6.7+commit.b8d736ae.Darwin.appleclang
[...]

PR #1627, opened 2020-01-27
Single version of solidity compiler

### LINKED ISSUE #1478: No single version of solidity compiler
There are three places with solidity version: `install_solc.sh`, `SolidityCompiler.__default_version` and `pragma solidity ...` in contracts. In all places only recommendation for version but not checking or forcing to use it. Could be situation when tests run with one version and deploy with another

vzotova:
Fixes #1478

*  Only one place for solc version: `nucypher/blockchain/eth/sol/__init__.py`
* `SolidityCompiler` cleanup
* `SolidityCompiler`  forces to use supported version


nucypher/cli/commands/deploy.py
```diff
@@ -62,6 +62,8 @@
 option_target_address = click.option('--target-address', help="Address of the target contract", type=EIP55_CHECKSUM_ADDRESS)
 option_gas = click.option('--gas', help="Operate with a specified gas per-transaction limit", type=click.IntRange(min=1))
 option_network = click.option('--network', help="Name of NuCypher network", type=click.Choice(NetworksInventory.networks))
+option_ignore_deployed = click.option('--ignore-deployed', help="Ignore already deployed contracts if exist.", is_flag=True)
+option_ignore_solidity = click.option('--ignore-solidity-check', help="Ignore solidity version compatibility check", is_flag=True, default=None)
```
derekpierre: Minor thought: I would be a bit more explicit here for the name of the option - `option_ignore_solidity_version`.

vzotova: ✔️

scripts/installation/install_solc.py
```diff
@@ -0,0 +1,38 @@
...
+response.raise_for_status()
+with open(solc_bin_path, 'wb') as f:
+    f.write(response.content)
+
+# Set executable permission
+print(f"Setting executable permission on {solc_bin_path}")
+file_stat = stat(solc_bin_path)
+chmod(solc_bin_path, 0o0755)
+
+print(f"Successfully Installed solc {solc_version}")
```
KPrasch: :boom:

scripts/installation/install_solc.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python
+
```
KPrasch: Perhaps a comment in here about this script - It does *not* need nucypher installed to operate, but *does* depend on requests and is cross-platform compatible, etc.

nucypher/blockchain/eth/sol/__init__.py
```diff
@@ -14,3 +14,5 @@
 You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
+
+SOLIDITY_COMPILER_VERSION = '0.5.16'
```
KPrasch: This is the canonical version - correct?

scripts/installation/install_solc.py
```diff
@@ -0,0 +1,38 @@
...
+import platform
+import shutil
+from os.path import dirname, join, abspath
+from os import stat, chmod
+import requests
+
+
+if platform.system() != 'Linux':
```
KPrasch: What is preventing support of other platforms - a lack of available binary?

scripts/installation/install_solc.py
```diff
@@ -0,0 +1,38 @@
...
+FILE_PATH = join(BASE_DIR, PACKAGE_NAME, "__init__.py")
+
+METADATA = dict()
+with open(FILE_PATH) as f:
+    exec(f.read(), METADATA)
+
+solc_version = METADATA['SOLIDITY_COMPILER_VERSION']
+
+# Get solc binary for linux
+solc_bin_path=join(dirname(shutil.which('python')), 'solc')
```
KPrasch: Some convention comments: Use the same case for names in the same scope - `SOLC_BIN_PATH` for example. Perhaps make a function to encapsulate the install functionality (`def main`) and include a call at the bottom of this file:
```
if __name__ == "__main__":
    main()
```

scripts/installation/install_solc.py
```diff
@@ -0,0 +1,38 @@
...
+
+response = requests.get(url)
+response.raise_for_status()
+with open(solc_bin_path, 'wb') as f:
+    f.write(response.content)
+
+# Set executable permission
+print(f"Setting executable permission on {solc_bin_path}")
+file_stat = stat(solc_bin_path)
+chmod(solc_bin_path, 0o0755)
```
KPrasch: Can you name this octal? :-)

nucypher/cli/commands/deploy.py
```diff
@@ -178,10 +183,17 @@ def create_actor(self, emitter):
...
     )
 
 
 @click.group()
+@click.option('--solidity-version',
+              help="Echo the supported solidity version.",
+              is_flag=True,
+              callback=echo_solidity_version,
+              expose_value=False,
+              is_eager=True)
```
KPrasch: Excellent

nucypher/blockchain/eth/sol/__init__.py
vzotova: yep, https://github.com/ethereum/solidity/releases/tag/v0.5.16

scripts/installation/install_solc.py
vzotova: Github also has link for windows, but that is zip archive which has exe and few dll libraries, so need aditional operations. For me much easier to maintain this manually. But if we need it, I can add this for windows and I think using information from #977 - for mac

vzotova:
> Just to clarify - What relationship does this new single version have on contract `pragma`?

`pragma` defines range of versions or one specific version, `SOLIDITY_COMPILER_VERSION` defines one  specific version within this range

scripts/installation/install_solc.py
```diff
@@ -0,0 +1,38 @@
...
+from os import stat, chmod
+import requests
+
+
+if platform.system() != 'Linux':
+    raise EnvironmentError("This installation script is only compatible with linux-gnu-based operating systems.")
+
+PACKAGE_NAME = join('nucypher', 'blockchain', 'eth', 'sol')
+BASE_DIR = dirname(dirname(dirname(abspath(__file__))))
+FILE_PATH = join(BASE_DIR, PACKAGE_NAME, "__init__.py")
```
KPrasch: I am having second thoughts about using `__init__.py` as the location since it implies that *no new code* can be added there without also considering the `exec`.  What do you think about using the same path, but the name `__version__.py` instead?

vzotova: `__version__.py` looks like version of sub-module. Maybe `conf.py`?

KPrasch: I think the dunder is warranted - `__conf__.py?`

scripts/installation/install_solc.py
vzotova: ✔️

scripts/installation/install_solc.py
vzotova: ✔️

scripts/installation/install_solc.py
vzotova: ✔️

scripts/installation/install_solc.py
vzotova: 👍 
✔️

nucypher/blockchain/eth/sol/compile.py
```diff
@@ -26,12 +26,7 @@
 import itertools
 import shutil
 
-try:
-    from solc import install_solc, compile_files
-    from solc.exceptions import SolcError
-except ImportError:
-    # TODO: Issue #461 and #758 & PR #1480 - Include precompiled ABI; Do not use py-solc in standard installation
-    pass
+from nucypher.blockchain.eth.sol.__conf__ import SOLIDITY_COMPILER_VERSION
```
KPrasch: If you want to preserve the importable from the namespace `eth` - you can move this line into `__init__.py` and instead have just `from nucypher.blockchain.eth import SOLIDITY_COMPILER_VERSION`

vzotova: good advice, thanks
✔️

.circleci/config.yml
```diff
@@ -299,9 +299,12 @@ commands:
       - save_cache:
           key: pip-v2-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
```
cygnusv:
```suggestion
          key: pip-v3-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
```
I think we should bump cache version, since the cache for `master` still includes solc stuff, and we may experience problems when we merge this. In any case, with this change we only will incur in 2 or 3 more minutes for that single build.

.circleci/config.yml
```diff
@@ -310,6 +313,8 @@ commands:
       - run: sudo chown -R circleci:circleci /usr/local/lib/python3.7/site-packages
       - restore_cache:  # ensure this step occurs *before* installing dependencies
           key: pip-v2-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
```
cygnusv:
```suggestion
          key: pip-v3-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
```

.circleci/config.yml
```diff
@@ -299,9 +299,12 @@ commands:
       - save_cache:
           key: pip-v2-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
           paths:
-            - "/usr/local/bin/solc"
             - "~/.local/bin"
             - "~/.local/lib/python3.7/site-packages"
+      - save_cache:
+          key: solc-v1-{{ checksum "nucypher/blockchain/eth/sol/__init__.py" }}
```
cygnusv:
```suggestion
          key: solc-v1-{{ checksum "nucypher/blockchain/eth/sol/__conf__.py" }}
```

.circleci/config.yml
```diff
@@ -310,6 +313,8 @@ commands:
       - run: sudo chown -R circleci:circleci /usr/local/lib/python3.7/site-packages
       - restore_cache:  # ensure this step occurs *before* installing dependencies
           key: pip-v2-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
+      - restore_cache:
+          key: solc-v1-{{ checksum "nucypher/blockchain/eth/sol/__init__.py" }}
```
cygnusv:
```suggestion
          key: solc-v1-{{ checksum "nucypher/blockchain/eth/sol/__conf__.py" }}
```

.circleci/config.yml
vzotova: ✔️

.circleci/config.yml
vzotova: ✔️

.circleci/config.yml
vzotova: nice catch, thanks
✔️

.circleci/config.yml
vzotova: ✔️

PR #1646, opened 2020-02-02
update command from apt to apt-get

SeanMikha:


docs/source/guides/installation_guide.md
```diff
@@ -26,7 +26,7 @@ or similar distros:
     - build-essential
         
 Here's a one-liner to install the above packages on linux:
-`sudo apt install python3-dev build-essential libffi-dev python3-pip`
+`sudo apt-get install python3-dev build-essential libffi-dev python3-pip`
```
KPrasch: @SeanMikha - Does simply `apt` not work for you system (which platform is it)?

michwill: I understand, `apt-get` is more universal... But I think, if one uses a system which doesn't have `apt`, they'd figure out

PR #1652, opened 2020-02-04
[Do Not Merge] Fixes sub-stake duration (Pending Upgrade)

### LINKED ISSUE #1614: stake prolong→ periods to extend (1-3999999999999999999999968)
when 

```nucypher stake prolong```

**Enter number of periods to extend (1-3999999999999999999999968)**

 3999999999999999999999968 is wrong, but in this case, the actual stake is setup to 32 periods. so, if max periods is 368 for a year, not sure if has to be 

1- Enter number of periods to extend (1-336): 336 prolong + 32 actual= 368 a year
2- Enter number of periods to extend (1-365)

```
The Holder of Stakes.                                      

| # | Account  ---------------------------------- | Balance -----
=================================================================
| 0 | 0x4570805BC9CcEEB0370b9f2F7ba1E860ca11Bb47  | 45000 NU
| 1 | 0x1Ef837917c26a2f9aa31652A36aa7EA5aF7582c8  | 180489 NU
Select staking account [0]: 1
Selected 1:0x1Ef837917c26a2f9aa31652A36aa7EA5aF7582c8

| 0 | 0x1Ef8 | 0x4570 | 0 | 44511 NU | 32 periods . | Jan 22 01:00 CET - Feb 23 01:00 CET 

Select Stake: 0
Enter number of periods to extend (1-3999999999999999999999968): 336


Publish stake extension of 336 period(s) to the blockchain? [y/N]: y
Enter password to unlock account 0x1Ef837917c26a2f9aa31652A36aa7EA5aF7582c8: 
Broadcasting Stake Extension...
Successfully Prolonged Stake
OK | 0x9235d0d626d6bc45b85cde3bb2027074cd90e0d19801a46f13cfed4d0df4edca (42277 gas)
Block #2046144 | 0xfb75165281578ac594aec10d589759b791d21b4d8c0f2f7ec41ab158c746b1d8
 See https://goerli.etherscan.io/tx/0x9235d0d626d6bc45b85cde3bb2027074cd90e0d19801a46f13cfed4d0df4edca

| 0 | 0x1Ef8 | 0x4570 | 0 | 44511 NU | 368 periods  | Jan 22 01:00 CET - Jan 24 01:00 CET 
```

### LINKED ISSUE #1621: Improve UX of stake duration
Currently, if someone has wind down disabled, their stake duration will *increase* (as measured from the date they initiated their stake). This often confuses stakers.

IMO it's more intuitive to display the `Remaining stake duration`. This way, winddown disabled would cause the `Remaining stake duration` to stay constant (e.g. 365) whereas winddown enabled would cause `Remaining stake duration` to decrement towards 0 every period.

### PR DESCRIPTION
**Pull request by vzotova**
Based on #1672 

* Fixes leak in sub-stake duration abstraction
* Cleanup in contracts
* Fixes #1614, fixes #1621 
* Updates duration range for `stake create`, `stake divide` and `stake prolong`. For example, max stake duration in `stake create` is `65535 - current_period`

Leak occurs when winding down is disabled. Each period happens this: before `confirmActivity()` durstion is `x`, after `confirmActivity()` duration is `x+1`. This PR fixes this behaviour with addtional logic in duration calculating, now duration will be fixed for current period and will change only when next period begins and winding down is enabled

Versions:
StakingEscrow `2.2.2` -> `2.2.3`
PolicyManager `2.1.2` -> `2.1.3`

### CODE REVIEW

File: nucypher/blockchain/eth/constants.py [hunk 1]
```diff
@@ -19,3 +19,7 @@
 
 ETH_ADDRESS_BYTE_LENGTH = 20
 ETH_ADDRESS_STR_LENGTH = 40
+
+
+# Solidity
+SOLIDITY_MAX_UINT16 = 65535
```
derekpierre: Based on subsequent usage, I would just remove the `SOLIDITY_` prefix
```suggestion
MAX_UINT16 = 65535
```

OR 

```suggestion
MAX_PERIOD_VALUE = 65535
```

nucypher/blockchain/eth/constants.py
```diff
@@ -19,3 +19,7 @@
 
 ETH_ADDRESS_BYTE_LENGTH = 20
 ETH_ADDRESS_STR_LENGTH = 40
+
+
+# Solidity
```
derekpierre:
```suggestion
```

nucypher/cli/commands/stake.py
```diff
@@ -431,9 +432,10 @@ def create(general_config, transacting_staker_options, config_file, force, value
 
     if not lock_periods:
         min_locktime = STAKEHOLDER.economics.minimum_locked_periods
-        max_locktime = STAKEHOLDER.economics.maximum_rewarded_periods
+        default_locktime = STAKEHOLDER.economics.maximum_rewarded_periods
+        max_locktime = SOLIDITY_MAX_UINT16 - STAKEHOLDER.staking_agent.get_current_period()
         prompt = f"Enter stake duration ({min_locktime} - {max_locktime})"
-        lock_periods = click.prompt(prompt, type=stake_duration_range, default=max_locktime)
+        lock_periods = click.prompt(prompt, type=stake_duration_range, default=default_locktime)
```
derekpierre: Is it ever possible for `default_locktime` to be greater than `max_locktime`? Feels like a very small possibility when current period is large enough (?), but I can't quite wrap my head around it...

tests/blockchain/eth/contracts/main/staking_escrow/test_staking_escrow_additional.py
```diff
@@ -843,7 +843,19 @@ def test_wind_down(testerchain, token, escrow_contract, token_economics):
     # Staker deposits some tokens and confirms activity
     sub_stake = token_economics.minimum_allowed_locked
     duration = 10
-    tx = token.functions.transfer(staker, sub_stake).transact({'from': creator})
+
+    def check_last_period():
+        assert sub_stake == escrow.functions.getLockedTokens(staker, duration).call(), "Sub-stake is already unlocked"
+        assert 0 == escrow.functions.getLockedTokens(staker, duration + 1).call(), "Sub-stake is still locked"
+
+    def check_events(value: bool, length: int):
```
derekpierre: Perhaps we can rename `value` to `wind_down` to be more explicit?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -229,7 +229,9 @@ contract StakingEscrow is Issuer {
         internal view returns (uint16)
     {
         // if the next period (after current) is confirmed
-        if (_info.confirmedPeriod1 > _currentPeriod || _info.confirmedPeriod2 > _currentPeriod) {
+        if (_info.windDown &&
```
derekpierre: Just to make sure I understand here, we add 1 because the next period has already been confirmed, and the duration (`periods`) has been adjusted (`-1 if winddown, and +1 if !winddown`)...?

nucypher/cli/commands/stake.py
vzotova: You're right, after 06/06/2148 it will become a problem :) I guess current code will face with other problems closer to this date so I prefer to keep it for a few decades :)

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: very close, but not precisely.
we have several invariants with winding down parameter:
- state when whinding down is disabled. in that we case we don't need to do anything with duration and `confirmedPeriod`. Duration is fixed and we use it straithforward in last period calculation.
- winding dowin is enabled: duration is fixed without confimrations, but when worker confirms then duration must be decresed and only starting from the next period. As a result we use duration straightforward without confirmation and +1 with.

Next step is switching between invariats (when change winding down parameter). Without confirmation both invariants work equaly, so no need aditional math. 
When next period is confirmed:
- from false to true. We have "deferred" decrease of duration for the next period, so we return it back, mean add 1 to duration.
- from true to false. Same but in other direction

Also `setWindDown` has some corner cases, but logic there is the same as above

nucypher/blockchain/eth/constants.py
vzotova: ✔️

nucypher/blockchain/eth/constants.py
vzotova: ✔️

tests/blockchain/eth/contracts/main/staking_escrow/test_staking_escrow_additional.py
vzotova: ✔️

nucypher/cli/commands/stake.py
derekpierre: Hahaha touché!

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -100,7 +100,7 @@ contract Issuer is Upgradeable {
     /**
     * @notice Initialize reserved tokens for reward
     */
-    function initialize(uint256 _reservedReward) public onlyOwner {
+    function initialize(uint256 _reservedReward) external onlyOwner {
```
KPrasch: This is a compiler change?

vzotova: just small restriction (can be called only outside of contract), as a result give small gas savings

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -60,6 +60,10 @@ contract PolicyManager is Upgradeable {
         address indexed sender,
         uint256 value
     );
+    event NodeBrokenState(
+        address indexed node,
+        uint16 period
+    );
```
KPrasch: Heh - an event we hope never happens.

PR #1653, opened 2020-02-04
use ipdb as debugger in nucypher-dev docker container

vepkenez:


michwill:
Is `ipdb` already better than `pdbpp`?

PR #1686, opened 2020-02-14
Extract options used only by alice-init and alice-config in a separate option group

### LINKED ISSUE #1682: Extract Alice Configuration Defaults from CLI commands
Presently, alice defaults m, n, value, etc. are improperly grouped into `AliceConfigOptions` - since they are implemented as part of `AliceConfiguration`.  We need to split these options into a smaller scope so that `--help` remains useful and informative for commands that cannot accept those flags.

example:

```
nucypher alice public-key

Usage: nucypher alice public-keys [OPTIONS]

  Obtain Alice's public verification and encryption keys.

Options:
  --min-stake INTEGER             The minimum stake the teacher must have to
                                  be a teacher
  --teacher TEXT                  An Ursula URI to start learning from
                                  (seednode)
  --hw-wallet / --no-hw-wallet
  --duration-periods INTEGER      Policy duration in periods
  --n INTEGER                     N-Total KFrags
  --m INTEGER                     M-Threshold KFrags
  --light                         Indicate that node is light
  --poa                           Inject POA middleware
  -Z, --mock-networking           Use in-memory transport instead of
 [...removed text...]

```

fjarri:
Fixes #1682

Also `grant()` now uses `m` and `n` from the schema, and not from the config - seems more logical to me, and they are already present there.

### DIFF
--- a/nucypher/cli/commands/alice.py
+++ b/nucypher/cli/commands/alice.py
@@ -71,7 +71,7 @@ class AliceConfigOptions:
     __option_name__ = 'config_options'
 
     def __init__(self, dev, network, provider_uri, geth, federated_only, discovery_port,
-                 pay_with, registry_filepath, middleware, poa, light, m, n, duration_periods):
+                 pay_with, registry_filepath, middleware):
 
         if federated_only and geth:
             raise click.BadOptionUsage(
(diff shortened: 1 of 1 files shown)


fjarri:
Not sure about the name for the new option group, but that's easy to change.

PR #1688, opened 2020-02-14
improve ansible deployment/management/status

vepkenez:
When restarting ursula after an update, EC2 nodes with 4GB of RAM occasionally spaz on scrypt/keychain decryptiont because of geth taking too much memory.

So lets stop geth and clear out some RAM before restarting ursula container.  This seems to solve it.

Also adds NuCypher version to status output
```
        "MediumTurquoise Clock Purple Heart",
        "nucypher version:      2.0.0-beta.5",
        "staker address:        0xB1171EE4fd7E54ed960ccF9ECa383E0f3B1C6B06",
        "worker address:        0x94fBC43fC56d7D3A3002E85C3575009316BcCcB9",
        "rest url:              https://52.34.68.134:9151",
        "missing_confirmations: 0",
        "last active period:    18307",
        "balances:",
        "   ETH:                0.045739414712468285",
        "   NU:                 0.0"
```

### DIFF
--- a/deploy/ansible/worker/include/check_running_ursula.yml
+++ b/deploy/ansible/worker/include/check_running_ursula.yml
@@ -20,6 +20,7 @@
       debug:
         msg:
           - "{{status_data.json.nickname}}"
+          - "nucypher version:      {{status_data.json.version}}"
           - "staker address:        {{status_data.json.staker_address}}"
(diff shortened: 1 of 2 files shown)


KPrasch:
:cowboy_hat_face:

PR #1693, opened 2020-02-14
[WIP] Have Ursula idle when there is no bonded worker

### LINKED ISSUE #1656: Ursula should idle instead of crash if there's no bonded staker
Currently Ursula crashes on startup if there's no corresponding Staker initiated beforehand. She should idle and wait for a stake instead.

tuxxy:
### What this does:
1. Creates a new blocking `staticmethod` on `Worker` called `block_until_bonded`.
    - Returns the staking address of the provided worker.
    - Blocks for a set `poll_rate` (by default, checks every 10 seconds)
2. Resolves #1656.

### TODO:
1. Remove prompts for the Staker address on `nucypher ursula init`.
    - We want to ask the blockchain for the staking address instead.

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1200,6 +1200,18 @@ class Worker(NucypherTokenActor):
             if start_working_now:
                 self.work_tracker.start(act_now=False)
 
+    @staticmethod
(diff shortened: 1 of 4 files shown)


KPrasch:
Moved to #1697

PR #1729, opened 2020-02-21
Disable Sentry

KPrasch:


nucypher/utilities/logging.py
```diff
@@ -37,7 +37,10 @@
 
 
 def initialize_sentry(dsn: str):
-    import sentry_sdk
+    try:
+        import sentry_sdk
+    except ImportError:
+        raise ImportError('Sentry SDK is not installed. Please install it and try again.')
```
derekpierre: If we aren't going to use sentry by default, should we also remove it from the Pipfile, or move it to the [dev-packages] section?

KPrasch: Yeah, Good Idea.

PR #1734, opened 2020-02-21
Ansible worker deploy tweak: Don't wait for geth sync if we are just updating the node

vepkenez:
When updating multiple machines, geth sync could take disparate amounts of time and cause the different nodes to actually _lose_ sync while waiting for a straggler.  Better to just restart them quickly.


vepkenez:
🤠

PR #1738, opened 2020-02-23
Allow to deploy a network in idle mode. Activate later

### LINKED ISSUE #1532: Don't initialize StakingEscrow by default
I have serious concerns about our deployment process and how it affects day 0 of mainnet. I think we need to carefully design the process to minimize potential problems and blockages.

Currently, the `StakingEscrowDeployer` directly creates, funds, and initializes the `StakingEscrow` contract. This means that, by default, `StakingEscrow` is up and running once deployed. From that moment on, the clock is ticking and the first stakers have to create their stakes and confirm activity during the first 24 hours. But it can get worse: our current deployment process for `PreallocationEscrow` does something similar (i.e., creation, funding, and ownership transfer) and it can be quite intensive. On top of that, these contracts require knowing in advance the `StakingEscrow` address, so we can't deploy them before. It can be quite difficult to manage all those deployments in one day and still leave enough time for stakers to stake.

The good news is that the contracts are well-prepared for this. So, we only need to design carefully the deployment process and make the proper changes to the deployers and CLI, which I don't think it's too much work. It's just a matter of thinking ahead now.

My proposal:
- Don't initialize `StakingEscrow` by default. This means we can still do all the previous deployment steps, and leave initialization to when it's the right time. This also imply we need a CLI option to initialize an existing `StakingEscrow`.
- Deploy the preallocation contracts in the interim between `StakingEscrow` deployment and initialization. Transfer ownership of the contracts after the initialization (although perhaps there's no risk about doing that together with the deployment, after all `StakingEscrow` is not active yet. After #1407, it even doesn't have tokens yet!)
- Encourage stakers to set up their worker nodes before they can create the stake. Maybe there should be a way to create and run an "empty" Ursula that can run the discovery loop and checks that it's accessible.

cygnusv:
Closes #1532.

Recipe to deploy a network in idle mode:

1. Deploy a NuCypher contract

```
nucypher-deploy contracts --contract-name NuCypherToken --provider $NUCYPHER_PROVIDER_URI --poa --registry-infile ...
```
2. Deploy a StakingEscrow in IDLE mode
```
nucypher-deploy contracts --contract-name StakingEscrow --mode idle --provider $NUCYPHER_PROVIDER_URI --poa --registry-infile ...
```
3. Deploy PolicyManager and Adjudicator
```
nucypher-deploy contracts --contract-name PolicyManager --provider $NUCYPHER_PROVIDER_URI --poa --registry-infile ...
nucypher-deploy contracts --contract-name Adjudicator --provider $NUCYPHER_PROVIDER_URI --poa --registry-infile ...
```
4. When you're ready, activate StakingEscrow
```
nucypher-deploy contracts --contract-name StakingEscrow --activate --provider $NUCYPHER_PROVIDER_URI --poa --registry-infile ...
```

### DIFF
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -19,7 +19,14 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 from collections import OrderedDict
 from typing import Tuple, Dict, List
 
-from constant_sorrow.constants import CONTRACT_NOT_DEPLOYED, NO_DEPLOYER_CONFIGURED, NO_BENEFICIARY
+from constant_sorrow.constants import (
+    CONTRACT_NOT_DEPLOYED,
+    NO_DEPLOYER_CONFIGURED,
+    NO_BENEFICIARY,
+    BARE,
+    IDLE,
+    FULL
+)
 from web3 import Web3
 from web3.contract import Contract
 
@@ -36,11 +43,11 @@ from nucypher.blockchain.eth.agents import (
     MultiSigAgent,
     ContractAgency
 )
-from nucypher.blockchain.eth.constants import DISPATCHER_CONTRACT_NAME, STAKING_INTERFACE_ROUTER_CONTRACT_NAME
+from nucypher.blockchain.eth.constants import DISPATCHER_CONTRACT_NAME
 from nucypher.blockchain.eth.decorators import validate_secret, validate_checksum_address
-from nucypher.blockchain.eth.interfaces import BlockchainDeployerInterface
-from nucypher.blockchain.eth.interfaces import BlockchainInterfaceFactory
 from nucypher.blockchain.eth.interfaces import (
+    BlockchainDeployerInterface,
+    BlockchainInterfaceFactory,
     VersionedContract
 )
 from nucypher.blockchain.eth.registry import AllocationRegistry
@@ -58,6 +65,8 @@ class BaseContractDeployer:
     _upgradeable = NotImplemented
     _ownable = NotImplemented
     _proxy_deployer = NotImplemented
+
+    can_be_idle = False
 
     class ContractDeploymentError(Exception):
         pass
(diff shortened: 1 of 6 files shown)


nucypher/cli/commands/deploy.py
```diff
@@ -359,20 +366,27 @@ def rollback(general_config, actor_options):
...
 @group_general_config
 @group_actor_options
-@click.option('--bare', help="Deploy a contract *only* without any additional operations.", is_flag=True)
 @option_gas
 @option_ignore_deployed
 @click.option('--confirmations', help="Number of required block confirmations", type=click.IntRange(min=0))
 @click.option('--parameters', help="Filepath to a JSON file containing additional deployment parameters",
               type=EXISTING_READABLE_FILE)
-def contracts(general_config, actor_options, bare, gas, ignore_deployed, confirmations, parameters):
+@click.option('--mode',
```
KPrasch: :+1:

nucypher/blockchain/eth/deployers.py
```diff
@@ -54,13 +61,13 @@ class BaseContractDeployer:
...
     contract_name = NotImplemented
     deployment_steps = NotImplemented
-    upgrade_steps = NotImplemented
-    rollback_steps = NotImplemented
 
     _upgradeable = NotImplemented
     _ownable = NotImplemented
     _proxy_deployer = NotImplemented
 
+    can_be_idle = False
```
derekpierre: Is this used somewhere?

nucypher/blockchain/eth/deployers.py
```diff
@@ -654,22 +674,39 @@ def deploy(self,
...
-        deployment_receipts = dict(zip(self.deployment_steps, ordered_receipts))
+        activation_receipts = dict(zip(self.activation_steps, (approve_reward_receipt, init_receipt)))
+        return activation_receipts
 
-        # Set the contract and transaction receipts #
-        self._contract = the_escrow_contract
-        self.deployment_receipts = deployment_receipts
-        return deployment_receipts
+    @property
+    def ready_to_activate(self) -> bool:
```
vzotova: I guess it's duplicated in the next method `is_active`, maybe remove this and keep other method?

nucypher/blockchain/eth/deployers.py
```diff
@@ -654,22 +674,39 @@ def deploy(self,
...
+    def is_active(self) -> bool:
+        try:
+            deployed_contract = self._get_deployed_contract()
+        except self.blockchain.UnknownContract:
+            return False
+
+        # TODO: Consider looking for Initialized event - see #1193
+        # This mimics isInitialized() modifier in Issuer (StakingEscrow's base contract)
+        current_supply_1 = deployed_contract.functions.currentSupply1().call()
+        return current_supply_1 != 0
```
vzotova: I'm for removing TODO and keeping check like now. If changes to `currentSupply1` is visible then block was added to chain

nucypher/blockchain/eth/deployers.py
cygnusv: Not really, this method deals with `current_supply_1 == 0`, while the next one is `current_supply_1 != 0`.

vzotova: I see, this confuses, especially `ready_to_activate` and 
 when `current_supply_1 != 0` which means contract activated and never will be ready to activate again

PR #1747, opened 2020-02-26
Auto Arrangement Pruning

KPrasch:
- Automatically deletes expired arrangements and kfrags from Ursulas's database
- `KeyStore` -> `Datastore`
- Complete type annotations for `Datastore`


nucypher/characters/lawful.py
```diff
@@ -1008,6 +1012,9 @@ def __init__(self,
                                                    rest_app=rest_app, datastore=datastore,
                                                    hosting_power=tls_hosting_power)
 
+                self._arrangement_pruning_task = LoopingCall(f=self.__prune_arrangements)
+                self._arrangement_pruning_task.start(interval=self._pruning_interval)  # TODO: Move to ursula.run ?
```
cygnusv: What about moving this to the WorkTracker? I can see it being related.

nucypher/characters/lawful.py
```diff
@@ -1064,6 +1071,12 @@ def __init__(self,
             message = "Initialized Stranger {} | {}".format(self.__class__.__name__, self)
             self.log.debug(message)
 
+    def __prune_arrangements(self) -> None:
+        result = self.datastore.del_expired_policy_arrangements(
+            now=datetime.fromtimestamp(self._arrangement_pruning_task.clock.seconds()))
+        self.log.debug(f"Pruned {result} policy arrangements.")
```
derekpierre: What do you think about only printing this when `result > 0` - no use flooding the logs (even in debug) with useless entries.

nucypher/characters/lawful.py
```diff
@@ -871,15 +873,17 @@ def retrieve():
...
 class Ursula(Teacher, Character, Worker):
+
     banner = URSULA_BANNER
     _alice_class = Alice
 
     # TODO: Maybe this wants to be a registry, so that, for example,  NRN
     # TLSHostingPower still can enjoy default status, but on a different class  NRN
     _default_crypto_powerups = [SigningPower, DecryptingPower]
 
+    _pruning_interval = 20
```
derekpierre: Any context for the decision for the default of `20 seconds`? If just pruning does it make sense to do it a little less often...per minute(?) - I mean who knows 🤔 😊 ... just asking the question. Presumably pruning is a fairly light operation anyway.

KPrasch: Its a fairly arbitrary guess trying to consider the duration of a very short policy, and the probability of the kfrag being pruned in a cohort of ursulas for one policy.  

Your sense is that this is a bit too slow (I think I agree)?

derekpierre: Not based on anything but a 60-second (per minute) interval seems fine, which would be less iterations of the task, but still effective enough - saves the task from having to run too often. But I defer to you on this one, you would know better than I would.

nucypher/keystore/keystore.py
```diff
@@ -145,9 +158,24 @@ def del_policy_arrangement(self, arrangement_id: bytes, session=None):
         session.query(PolicyArrangement).filter_by(id=arrangement_id).delete()
         session.commit()
 
+    def del_expired_policy_arrangements(self, session=None, now=None) -> int:
+        """
+        Deletes all expired PolicyArrangements from the Keystore.
+        """
+        session = session or self._session_on_init_thread
+        now = now or datetime.now()
+        result = session.query(PolicyArrangement).filter(PolicyArrangement.expiration < now)
```
tuxxy:
```suggestion
        result = session.query(PolicyArrangement).filter(PolicyArrangement.expiration <= now)
```

Gotta make this a `<=` comparison.

nucypher/keystore/keystore.py
```diff
@@ -145,9 +158,24 @@ def del_policy_arrangement(self, arrangement_id: bytes, session=None):
...
+        deleted_records = result.delete()
+        try:
+            session.commit()
+        except OperationalError:
+            session.rollback()
```
tuxxy: Should we add some logging here in the event that a rollback occurs?

tests/characters/test_bob_joins_policy_and_retrieves.py
```diff
@@ -198,3 +201,30 @@ def test_bob_retrieves_with_treasure_map(
         treasure_map=bytes(treasure_map))
 
     assert text1[0] == text2[0] == b'Welcome to flippering number 2.'
+
+
+def test_bob_retrieves_too_late(federated_bob, federated_ursulas,
```
tuxxy: A very solid test :+1:

nucypher/characters/lawful.py
tuxxy: I pointed out below that the session could rollback and not prune any arrangements but this would still be written to the log.

nucypher/characters/lawful.py
tuxxy: Since it's just a very simple query, it seems reasonable to do it every 20-30 seconds, imho.

nucypher/keystore/keystore.py
tuxxy: Maybe this should raise some sort of exception for `__prune_arrangements` to handle?

nucypher/characters/lawful.py
KPrasch: `WorkTracker` is only implemented for non-federated operation.

nucypher/characters/lawful.py
KPrasch: :heavy_check_mark:

nucypher/characters/lawful.py
KPrasch: :heavy_check_mark:

nucypher/keystore/keystore.py
KPrasch: :heavy_check_mark:

nucypher/keystore/keystore.py
KPrasch: :heavy_check_mark:

nucypher/characters/lawful.py
```diff
@@ -873,13 +876,16 @@ def retrieve():
...
 class Ursula(Teacher, Character, Worker):
+
     banner = URSULA_BANNER
     _alice_class = Alice
 
     # TODO: Maybe this wants to be a registry, so that, for example,  NRN
     # TLSHostingPower still can enjoy default status, but on a different class  NRN
     _default_crypto_powerups = [SigningPower, DecryptingPower]
 
+    _pruning_interval = 60  # seconds
```
tuxxy: :+1:

nucypher/characters/lawful.py
```diff
@@ -1049,21 +1064,42 @@ def __init__(self,
...
             message = "Initialized Stranger {} | {}".format(self.__class__.__name__, self)
             self.log.debug(message)
 
+    def __prune_arrangements(self) -> None:
+        """Deletes all expired arrangements and kfrags in the datastore."""
+        now = datetime.fromtimestamp(self._arrangement_pruning_task.clock.seconds())
+        try:
+            result = self.datastore.del_expired_policy_arrangements(now=now)
+        except OperationalError:
+            self.log.warn(f"Failed to prune policy arrangements; DB session rolled back.")
```
tuxxy: Yeah, this is good.

nucypher/datastore/datastore.py
```diff
@@ -53,7 +57,23 @@ def __init__(self, sqlalchemy_engine=None) -> None:
         # Best to treat like hot lava.
         self._session_on_init_thread = Session()
 
-    def add_key(self, key, is_signing=True, session=None) -> Key:
+    @staticmethod
+    def __commit(session) -> None:
```
tuxxy: Very nice addition here, as well.

nucypher/datastore/datastore.py
```diff
@@ -14,35 +14,39 @@
...
 class NotFound(Exception):
     """
-    Exception class for KeyStore calls for objects that don't exist.
+    Exception class for Datastore calls for objects that don't exist.
     """
     pass
 
 
-class KeyStore(object):
+class Datastore:
```
tuxxy: :+1:

nucypher/characters/lawful.py
derekpierre: 👍

nucypher/characters/lawful.py
```diff
@@ -1049,21 +1064,42 @@ def __init__(self,
...
+        else:
+            self.log.debug(f"Pruned {result} policy arrangements.")
```
derekpierre: To reduce the "noise" in logs:
```suggestion
            if result > 0:
                self.log.debug(f"Pruned {result} policy arrangements.")
```

KPrasch: :heavy_check_mark:

PR #1759, opened 2020-03-03
Offline Configuration Selection

KPrasch:
- Always include checksum address in ursula filenames (deprecates `ursula.json` as a default)
- "Offline and offchain" configuration files selection via filename and directory parsing
- Non-interactive offline config selection via `--worker-address`
- Interactive selection of Ursula configuration files only if > 1 config exists
- Implements interactive config selection for Ursula commands
- Interactive Network Selection


nucypher/cli/actions.py
```diff
@@ -609,26 +612,100 @@ def get_or_update_configuration(emitter, config_class, filepath: str, config_opt
...
+        # Display account info
+        headers = ['Account']
+        emitter.echo(tabulate(parsed_addresses, headers=headers, showindex='always'))
+
+        # Prompt the user for selection, and return
+        prompt = f"Select {config_class._NAME} configuration"
+        account_range = click.IntRange(min=0, max=len(config_files) - 1)
+        choice = click.prompt(prompt, type=account_range, default=0)
+        config_file = config_files[choice]
+        emitter.echo(f"Selected {choice}: {config_file}", color='blue')
```
derekpierre: 👍 🎉

nucypher/cli/commands/deploy.py
```diff
@@ -74,7 +74,7 @@
 option_target_address = click.option('--target-address', help="Address of the target contract", type=EIP55_CHECKSUM_ADDRESS)
 option_gas = click.option('--gas', help="Operate with a specified gas per-transaction limit", type=click.IntRange(min=1))
 option_gas_strategy = click.option('--gas-strategy', help="Operate with a specified gas price strategy", type=click.STRING)  # TODO: GAS_STRATEGY_CHOICES
-option_network = click.option('--network', help="Name of NuCypher network", type=click.Choice(NetworksInventory.networks))
+option_network = click.option('--network', help="Name of NuCypher network", type=click.Choice(NetworksInventory.NETWORKS))
```
derekpierre: Should we make this option required for deployment? Feels like it should be for deployment, and since we limit the type.

nucypher/cli/actions.py
```diff
@@ -609,26 +612,100 @@ def get_or_update_configuration(emitter, config_class, filepath: str, config_opt
...
 
-    if not worker_config_exists:
-        emitter.message("No Ursula configurations found.  run 'nucypher ursula init' then try again.", color='red')
+    #
+    # Scrape Disk Configurations
+    #
+
+    config_root = config_root or DEFAULT_CONFIG_ROOT
+    default_config_file = glob.glob(config_class.default_filepath(config_root=config_root))
+    glob_pattern = f'{config_root}/{config_class._NAME}-0x*.{config_class._CONFIG_FILE_EXTENSION}'
```
fjarri: Feels like `._NAME` is used often enough to warrant it being a public attribute/property.

nucypher/cli/actions.py
```diff
@@ -609,26 +612,100 @@ def get_or_update_configuration(emitter, config_class, filepath: str, config_opt
...
-    config_root = abspath(dirname(config_file)) if config_file else DEFAULT_CONFIG_ROOT
-    worker_config_exists = glob.glob(UrsulaConfiguration.default_filepath(config_root=config_root))
+        else:
+            raise ValueError(f"Cannot extract checksum from filepath '{filepath}'")
+    else:
+        try:
+
+            character_name, checksum_address = match.groups()
+        except ValueError:
+            raise  # TODO: Additional handling here?
```
fjarri: What kind of a situation can result in a `ValueError` here?

nucypher/cli/actions.py
```diff
@@ -637,3 +714,12 @@ def issue_stake_suggestions(value: NU = None, lock_periods: int = None):
         click.confirm(f"Wow, {value} - That's alot of NU - Are you sure this is correct?", abort=True)
```
fjarri: Might as well fix a typo here: "alot" -> "a lot" :)

nucypher/cli/actions.py
KPrasch: Too many of too few match groups being unpacked into the tuple... the regex pattern would have to match the wrong number of times.

fjarri: But if there are too many matches, `match()` will just pick the first one, and if there are too few (that is, zero). `match` will be `None`, and this branch won't be reached at all.

KPrasch: :heavy_check_mark:

nucypher/cli/actions.py
KPrasch: :heavy_check_mark:

nucypher/cli/actions.py
KPrasch: Hmm, yeah Let's make an Issue for that.

fjarri: Filed issue #1771

nucypher/cli/actions.py
```diff
@@ -609,31 +612,113 @@ def get_or_update_configuration(emitter, config_class, filepath: str, config_opt
...
+    else:
+        # Default: Only one config file, use it.
+        config_file = config_files[0]
+
     return config_file
 
 
 def issue_stake_suggestions(value: NU = None, lock_periods: int = None):
     if value and (value > NU.from_tokens(150000)):
-        click.confirm(f"Wow, {value} - That's alot of NU - Are you sure this is correct?", abort=True)
```
tuxxy: https://thewritepractice.com/wp-content/uploads/2012/05/Alot-vs-a-lot1-600x450.png

PR #1769, opened 2020-03-11
Forces ansible playbook to specifically use pip3 for python3

SeanMikha:
This adds greater portability for users with mixed environments, allowing ansible to explicitly state which pip version to use. Ansible will typically default to 'python/pip' whichever is set beforehand but that may not be the case for remote hosts where we want to deploy workers.

### DIFF
--- a/deploy/ansible/worker/include/setup_docker.yml
+++ b/deploy/ansible/worker/include/setup_docker.yml
@@ -27,6 +27,7 @@
     - name: Install Docker Module for Python
       pip:
         name: docker
+        executable: pip3
 
     - name: Ensure group "docker" exists
       group:
--- a/deploy/ansible/worker/inventory.yml
+++ b/deploy/ansible/worker/inventory.yml
@@ -43,6 +43,7 @@ all:
             geth_dir: '/home/nucypher/geth/.ethereum/goerli/'
             geth_container_datadir: "/root/.ethereum/goerli"
             etherscan_domain: goerli.etherscan.io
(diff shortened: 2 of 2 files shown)


KPrasch:
@vepkenez - Are there any compatibility changes in ansible wrt workers polling for bonding?

vepkenez:
> @vepkenez - Are there any compatibility changes in ansible wrt workers polling for bonding?
I don't foresee that being an issue.

vepkenez:
This doesn't really affect anything except installing docker.
I do want to test this before approving though.  I'll do that today.

vepkenez:
@SeanMikha are there any specific host environments where you have found this to be an issue?

SeanMikha:
@vepkenez yes on Azure I found that if the localhost or cloudshell (from Azure) in this case has  python2 installed if you don't specifically create symbolic link or set ansible to use python3 it will default to using the standard python pip command without a version which makes it use the default on the localhost where ansible is running (python2 in this case). This was causing errors when running the setup workers playbook, (the remote hosts were running canonical ubuntu 16.04).

SeanMikha:
@vepkenez I went back and checked on azure. The default/untouched/vanilla version of canonical Ubuntu 16.04 LTS comes with python version 2.7 out-of-the-box and as the default interpreter (python --version) so this is going to be a issue with host environments unless specifically put in.

vepkenez:
Right and we are specifically installing python3 and pip3 so this seems like obviously a good idea.

vepkenez:
💯

PR #1770, opened 2020-03-11
update inventory to include azure deployments

SeanMikha:
Not much is changed or is different than AWS, however default for azure users regardless of distribution is azureuser, you cannot deploy vms with base user root.


vepkenez:
This seems good.  I’m surprised by how normal azure’s Ubuntu support is.  🎉

Issue #1758, opened 2020-03-03
Worklock: Move enabling of claiming to ContractAdministrator and nucypher-deploy


vzotova:
`post-init`, `force_refund`, `verify_bidding_correctness`, `enable_claiming`  from `Bidder` actor and from `worklock` CLI

cygnusv:
I think we can leave as it is. What do you think?

vzotova:
I guess so
at least do this task with the smallest priority and if we will have time for this

cygnusv:
I'm afraid we don't have time 😅 . Wontfix.

PR #1773, opened 2020-03-14
🐳 Worklock updates (option 4)

### LINKED ISSUE #1508: What condition for bid canceling in worklock?
When bid canceling become available: while bidding, after bidding?
Who can cancel bid? Anyone or who will get only less then minimum NU?
Other conditions?

### LINKED ISSUE #1733: Worklock: Min stake expressed in ETH & no post-bid cancelling
For Worklock, denominate minimum threshold in ETH, auto-reject bids lower than this figure & prevent cancelling after bidding period ends. The min NU required to actually stake is set after the bidding period ends and equals the minimum ETH threshold (using the final NU/ETH worklock rate). 

Benefits:
- Impede 'whale vs. smaller players' attack wherein adversary floods worklock with ETH in order to drive down the sums of NU granted to smaller participants, knocking them below the min NU threshold and forcing them to cancel their bids. 
- Impedes 'inflation pool bloat' attack where adversary cancels a large bid after the final NU/ETH worklock rate is set, such that 'burned' tokens are sent to a pool to be distributed as inflationary rewards, which benefits larger stakers more.
- ETH has clear and established economic value, and therefore functions better as collateral (i.e. stronger loss aversion and disincentive against slashable offences)  
- Participants can join worklock without being at the mercy of the eventual NU/ETH rate. From a planning point of view, this reduces the risk of spending resources preparing an Ursula, only to be forced to cancel their bid. Without the change proposed, the only option is over-bidding to ensure satisfying the (uncertain) threshold, which biases worklock towards the deeper-pocketed.

### LINKED ISSUE #1778: Improvements to WorkLock status command output
Current status:
```
Time
======================================================
Bidding Start Date ................... 2020-03-18 22:10:00+00:00
Bidding End Date ..................... 2020-03-19 22:00:00+00:00
Cancellation Window End Date ......... 2020-03-20 00:00:00+00:00
Bidding Duration ..................... 23:50:00
Cancellation Window Duration ......... 1 day, 1:50:00
Bidding Time Remaining ............... 8:07:11
Cancellation Window Time Remaining ... 10:07:11
Claiming phase open .................. No

Economics
======================================================
Min allowed bid ....... 10 wei
ETH Pool .............. 3000000000000152235 wei
ETH Supply ............ 3000000000000152235 wei
Bonus ETH Supply ...... 3000000000000152025 wei

Lot Size .............. 283000000 NU
Bonus Lot Size ........ 282685000 NU

Boosting Refund ....... 400
Slowing Refund ........ 100
Bonus Refund Rate ..... 23557083.25
Bonus Deposit Rate .... 94228333
```

* The output of the WorkLock `status` command doesn't currently include the number of bids - something to think about.
* When using the `--bidder-address` option, we can try to give the current NU amount for the bid. In case it's above the max, we can state that the bid will likely be partially refunded.

vzotova:
Fixes #1733
Fixes #1508
Fixes #1778
Opens #1758

Differences from #1732:
* No max bid
* No updating of min / max stake in `StakingEscrow` (`enableClaiming()`)
* Bidder will get min stake for min bid and some bonus stake for everything that higher than min bid
* Only bonus part could be force refunded


nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -109,74 +143,324 @@ contract WorkLock {
...
+            info.depositedETH = resultBid;
+        }
+
+        // reset verification
+        nextBidderToCheck = 0;
+
+        // transfer a refund
+        for (uint256 i = 0; i < length; i++) {
+            address payable bidder = _biddersForRefund[i];
+            bidder.sendValue(refunds[i]);
```
cygnusv: There's no need to use the Withdrawal pattern here, right? My understanding is that it's not necessary since if a bidder that needs to be refunded is trying to do something fishy using a contract that doesn't accept ETH, what would happen is that it wouldn't be able to claim either. Just want to double-check.

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -109,74 +143,324 @@ contract WorkLock {
...
-        require(block.timestamp >= startBidDate && block.timestamp <= endBidDate,
-            "Bid is open during a certain period");
+        require(block.timestamp >= startBidDate, "Bidding is not open yet");
+        require(block.timestamp < endBidDate, "Bidding is already finished");
         WorkInfo storage info = workInfo[msg.sender];
+
+        // first bid
+        if (info.depositedETH == 0) {
+            require(msg.value >= minAllowedBid, "Bid must be at least minimum");
+            require(bidders.length < tokenSupply / minAllowableLockedTokens, "Not enough tokens for more bidders");
```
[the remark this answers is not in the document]
vzotova: Do you want this for readability? Or other reason?

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
cygnusv: After reviewing this a second time, I think we should follow the withdrawal pattern here. We should do the ETH transfers on the claiming function, on a per-bidder basis.

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -109,74 +143,324 @@ contract WorkLock {
...
+
+        // when all participants bid with the same minimum amount of eth
+        if (bonusETHSupply == 0) {
+            return tokenSupply / bidders.length;
+        }
+
+        uint256 bonusETH = _ethAmount - minAllowedBid;
+        if (bonusETH == 0) {
+            return minAllowableLockedTokens;
+        }
```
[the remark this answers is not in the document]
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -109,74 +143,324 @@ contract WorkLock {
...
         info.depositedETH = 0;
-        if (block.timestamp <= endBidDate) {
-            ethSupply = ethSupply.sub(refundETH);
-        } else {
-            unclaimedTokens = unclaimedTokens.add(ethToTokens(refundETH));
+
+        // remove from bidders array, move last bidder to the empty place
+        uint256 length = bidders.length;
+        if (info.index != length - 1) {
+            address lastBidder = bidders[length - 1];
```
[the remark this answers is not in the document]
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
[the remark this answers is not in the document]
cygnusv: Oh nevermind...I was thinking for a moment that that division was not an integer division.

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: Good thoughts, I think still could be one irrational behaviour - make receive eth impossible and it will freeze worklock. so I think we need withdraw pattern for that, something very simple - mapping (address => refund)

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -23,63 +23,92 @@ contract WorkLock {
     event Bid(address indexed sender, uint256 depositedETH);
     event Claimed(address indexed sender, uint256 claimedTokens);
     event Refund(address indexed sender, uint256 refundETH, uint256 completedWork);
-    event Burnt(address indexed sender, uint256 value);
     event Canceled(address indexed sender, uint256 value);
+    event BiddersChecked(address indexed sender, uint256 startIndex, uint256 endIndex);
+    event ClaimingEnabled(address indexed sender);
```
cryptoseal86: I don't see any emits of this event in the code. Should it be deleted?

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -109,74 +143,320 @@ contract WorkLock {
...
+    function shutdown() internal {
+        endCancellationDate = uint256(0) - 1; // "infinite" cancellation window
+        token.safeTransfer(creator, tokenSupply);
+    }
+
+    /**
+    * @notice Make force refund to bidders who can get tokens more than maximum allowed
+    * @param _biddersForRefund Sorted list of unique bidders. Only bidders who must receive a refund
+    */
+    function forceRefund(address payable[] calldata _biddersForRefund) external afterCancellationWindow {
```
cryptoseal86: should it have `onlyOwner` modificator?

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: Thanks, forgot to remove while rebasing, 
✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: nope, this can be called by any address

nucypher/cli/commands/worklock.py
```diff
@@ -126,15 +126,17 @@ def bid(general_config, worklock_options, registry_options, force, hw_wallet, va
...
+
+    registry = registry_options.get_registry(emitter, general_config.debug)
+    bidder = worklock_options.create_bidder(registry=registry, hw_wallet=hw_wallet)
+
     if not value:
         if force:
             raise click.MissingParameter("Missing --value.")
-        value = click.prompt("Enter bid amount in ETH", type=click.STRING)
+        minimum_bid = bidder.worklock_agent.minimum_allowed_bid
+        value = click.prompt(f"Enter bid amount in ETH (at least {Web3.fromWei(minimum_bid, 'ether')})", type=click.STRING)
```
derekpierre: Is the type `String` intended here? vs Float? Could help cleanup any errors that occur on line 138 - eg. descriptive error vs exception perhaps?

nucypher/cli/painting.py
```diff
@@ -854,41 +855,47 @@ def echo_solidity_version(ctx, param, value):
...
 ======================================================
-Start Date ........ {start}
-End Date .......... {end}
-Duration .......... {duration}
-Time Remaining .... {remaining} 
+Bidding Start Date ................... {bidding_start}
+Bidding End Date ..................... {bidding_end}
+Cancellation Window End Date ......... {cancellation_end}
+Bidding Duration ..................... {bidding_duration}
+Cancellation Window Duration ......... {cancellation_duration}
```
derekpierre: These two lines feel redundant since we have the start dates, the end dates, and the time remaining...

tests/blockchain/eth/entities/actors/test_bidder.py
```diff
@@ -78,14 +124,14 @@ def test_claim(testerchain, agency, token_economics, test_registry):
     with pytest.raises(Bidder.BidderError):
```
derekpierre: Should this be more specifically a `ClaimError`?

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: ✔️

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,77 +1653,185 @@ def __init__(self,
...
+            message = message or f'Bidding closed at {maya.MayaDT(end).slang_date()}'
+            raise self.BiddingIsClosed(message)
 
     def _ensure_bidding_is_closed(self, message: str = None) -> None:
         highest_block = self.worklock_agent.blockchain.w3.eth.getBlock('latest')
         now = highest_block['timestamp']
-        end = self.worklock_agent.end_date
+        end = self.worklock_agent.end_bidding_date
         if now < end:
             message = message or f"Bidding does not close until {end}"
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,77 +1653,185 @@ def __init__(self,
...
 
+    def _ensure_cancellation_window(self, ensure_closed: bool = True, message: str = None) -> None:
+        highest_block = self.worklock_agent.blockchain.w3.eth.getBlock('latest')
+        now = highest_block['timestamp']
+        end = self.worklock_agent.end_cancellation_date
+        if ensure_closed and now < end:
+            message = message or f"Operation cannot be performed while the cancellation window is still open (closes at {end})."
+            raise self.CancellationWindowIsOpen(message)
+        elif not ensure_closed and now >= end:
+            message = message or f"Operation is allowed only while the cancellation window is open (closed at {end})."
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,77 +1653,185 @@ def __init__(self,
...
     #
     # Transactions
     #
 
     def place_bid(self, value: int) -> dict:
         self._ensure_bidding_is_open()
+        minimum = self.worklock_agent.minimum_allowed_bid
+        if value < minimum:
+            raise self.BidderError(f"Too small value {prettify_eth_amount(value)} for bidding, "
+                                   f"bid must be at least {prettify_eth_amount(minimum)}")
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,77 +1653,185 @@ def __init__(self,
...
 
         # Require an active bid
         if not self.get_deposited_eth:
-            raise self.BidderError(f"No claims available for {self.checksum_address}")
-
-        # Ensure the claim is at least large enough for min. stake
-        minimum = self.economics.minimum_allowed_locked
-        if self.available_claim < minimum:
-            raise ValueError(f"Claim is too small. Claim amount must be worth at least {NU.from_nunits(minimum)})")
+            raise self.ClaimError(f"No claims available for {self.checksum_address}")
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,77 +1653,185 @@ def __init__(self,
...
+        self._all_bonus_bidders.update(whales)
+
+        return whales
+
+    def force_refund(self) -> dict:
+        self._ensure_cancellation_window(ensure_closed=True)
+
+        whales = self.get_whales()
+        if not whales:
+            raise self.BidderError(f"All bids are correct, force refund is not needed")
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,77 +1653,185 @@ def __init__(self,
...
+    # TODO better control: max iterations, interactive mode
+    def verify_bidding_correctness(self, gas_limit: int) -> dict:
+        self._ensure_cancellation_window(ensure_closed=True)
+
+        if self.worklock_agent.bidders_checked():
+            raise self.BidderError(f"Check was already done")
+
+        whales = self.get_whales()
+        if whales:
+            raise self.BidderError(f"There are some bidders which have too high bid: {whales}")
```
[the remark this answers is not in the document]

nucypher/cli/commands/worklock.py
vzotova: Changed to Float

tests/blockchain/eth/entities/actors/test_bidder.py
vzotova: ✔️

nucypher/cli/painting.py
vzotova: I'm not sure,  as for me - more information is better. But it's not very strong opinion.  What others think?

KPrasch: Inclusion of this data is useful, IMO.

nucypher/cli/commands/worklock.py
```diff
@@ -264,20 +266,84 @@ def refund(general_config, worklock_options, registry_options, force, hw_wallet)
...
 
 @worklock.command()
+@group_general_config
+@group_registry_options
+@group_worklock_options
+@option_force
+@option_hw_wallet
+@click.option('--gas-limit', help="Gas limit per each verification transaction", type=click.IntRange(min=60000))
+# TODO: Consider moving to administrator (nucypher-deploy) #1758
+def post_init(general_config, registry_options, worklock_options, force, hw_wallet, gas_limit):
```
KPrasch: Perhaps `def enable_claims`?

nucypher/cli/commands/worklock.py
```diff
@@ -264,20 +266,84 @@ def refund(general_config, worklock_options, registry_options, force, hw_wallet)
...
+    emitter = _setup_emitter(general_config)
+    if not worklock_options.bidder_address:  # TODO: Consider bundle this in worklock_options
+        worklock_options.bidder_address = select_client_account(emitter=emitter,
+                                                                provider_uri=registry_options.provider_uri,
+                                                                network=registry_options.network,
+                                                                show_balances=True)
+    registry = registry_options.get_registry(emitter, general_config.debug)
+    bidder = worklock_options.create_bidder(registry=registry, hw_wallet=hw_wallet)
+
+    whales = bidder.get_whales()
```
KPrasch: :joy:

nucypher/cli/commands/worklock.py
vzotova: I guess in this version of updates we can use this name, because no more `enable_claiming` function in contract

nucypher/cli/painting.py
derekpierre: 👍

nucypher/cli/commands/worklock.py
derekpierre: +1!

vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -23,63 +23,91 @@ contract WorkLock {
...
-    event Burnt(address indexed sender, uint256 value);
     event Canceled(address indexed sender, uint256 value);
+    event BiddersChecked(address indexed sender, uint256 startIndex, uint256 endIndex);
+    event ForceRefund(address indexed sender, address indexed bidder, uint256 refundETH);
 
     struct WorkInfo {
         uint256 depositedETH;
         uint256 completedWork;
         bool claimed;
+        uint256 index;
```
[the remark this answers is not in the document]
vzotova: Short types (less than 256) potentially could increase gas cost for reading. In this case we will slightly increase `cancelBid()` and deacrease `claim()` so it's good, I added it
✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,361 @@ contract WorkLock {
...
+    }
+
+    /**
+    * @notice Calculate amount of work that need to be done to refund specified amount of ETH
+    * @dev This value will be fixed only after end of bidding
+    * @param _ethAmount Specified amount of ETH
+    * @param _otherETHAmount Base part of ETH amount for correct calculation of work
+    * @dev ethAmount + otherETHAmount = depositedETH
+    */
+    function ethToWork(uint256 _ethAmount, uint256 _otherETHAmount) internal view returns (uint256) {
```
derekpierre: If I'm understanding the use of `_otherETHAmount`, would a rename to `_remainderOfDepositedETH` or something similar make sense?

Basically `_ethAmount` is the refund, and `_otherETHAmount` is the remaining amount of deposited ETH...

vzotova: I like `_remainderOfDepositedETH`
✔️

nucypher/blockchain/economics.py
```diff
@@ -51,34 +51,36 @@ class BaseEconomics:
     nunits_per_token = 10 ** __token_decimals  # Smallest unit designation
 
     # Period Definition
-    __default_hours_per_period = 24
+    _default_hours_per_period = 24
```
KPrasch: Moving to single underscore here for access via subclasses?

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
-        end = self.worklock_agent.end_date
+        end = self.worklock_agent.end_bidding_date
         if now < end:
-            message = message or f"Bidding does not close until {end}"
+            message = message or f"Bidding does not close until {maya.MayaDT(end).slang_date()}"
             raise self.BiddingIsOpen(message)
 
+    def _ensure_cancellation_window(self, ensure_closed: bool = True, message: str = None) -> None:
+        highest_block = self.worklock_agent.blockchain.w3.eth.getBlock('latest')
+        now = highest_block['timestamp']
```
KPrasch: Consider encapsulating `get_blocktime` in a method this might also help with future upstream API changes and test mocks.

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
+    def _get_max_bonus_bid_from_max_stake(self) -> int:
+        """Returns maximum allowed bid calculated from maximum allowed locked tokens"""
+        max_bonus_tokens = self.economics.maximum_allowed_locked - self.economics.minimum_allowed_locked
+        bonus_eth_supply = sum(self._all_bonus_bidders.values()) if self._all_bonus_bidders else self.worklock_agent.get_bonus_eth_supply()
+        bonus_worklock_supply = self.worklock_agent.get_bonus_lot_value()
+        max_bonus_bid = max_bonus_tokens * bonus_eth_supply // bonus_worklock_supply
+        return max_bonus_bid
+
+    def get_whales(self, force_read: bool = False) -> Dict[str, int]:
+        max_bonus_bid_from_max_stake = self._get_max_bonus_bid_from_max_stake()
```
KPrasch: Needs a docstring... something like: "Returns all worklock bidders over the whale threshold as a dictionary of addresses and bid values."

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
+        # first step - align at a minimum bid
+        if min_whale_bonus_bid != max_whale_bonus_bid:
+            whales = dict.fromkeys(whales.keys(), min_whale_bonus_bid)
+            self._all_bonus_bidders.update(whales)
+
+        bonus_eth_supply = sum(self._all_bonus_bidders.values())
+        bonus_worklock_supply = self.worklock_agent.get_bonus_lot_value()
+        max_bonus_tokens = self.economics.maximum_allowed_locked - self.economics.minimum_allowed_locked
+        if (min_whale_bonus_bid * bonus_worklock_supply) // bonus_eth_supply <= max_bonus_tokens:
+            raise self.BidderError(f"Internal error: At least one of bidders {whales} has allowable bid")
```
KPrasch: Hmm, this consistency check feels more like a form of `RuntimeError` - What is the recommended follow up action for this condition?

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
+
+        a = min_whale_bonus_bid * bonus_worklock_supply - max_bonus_tokens * bonus_eth_supply
```
KPrasch: ...`base_supply`? :sweat_smile:

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
+        self._all_bonus_bidders.update(whales)
+
+        return whales
+
+    def force_refund(self) -> dict:
+        self._ensure_cancellation_window(ensure_closed=True)
+
+        whales = self.get_whales()
+        if not whales:
+            raise self.BidderError(f"All bids are allowable, force refund is not needed")
```
KPrasch:
```suggestion
            raise self.BidderError(f"Force refund aborted: No whales detected and all bids qualify for claims.")
```
Better yet, an exception `class WhaleError` can be implemented several times here for a somewhat more clear  indication of the context.  :whale:

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
+
+        bonus_eth_supply = sum(self._all_bonus_bidders.values())
+        bonus_worklock_supply = self.worklock_agent.get_bonus_lot_value()
+        max_bonus_tokens = self.economics.maximum_allowed_locked - self.economics.minimum_allowed_locked
+        if (min_whale_bonus_bid * bonus_worklock_supply) // bonus_eth_supply <= max_bonus_tokens:
+            raise self.BidderError(f"Internal error: At least one of bidders {whales} has allowable bid")
+
+        a = min_whale_bonus_bid * bonus_worklock_supply - max_bonus_tokens * bonus_eth_supply
+        b = bonus_worklock_supply - max_bonus_tokens * len(whales)
+        refund = -(-a//b)  # div ceil
```
KPrasch: Unsure if there is an alternate (more semantic) approach, but I think this is an acceptable way of performing "ceiling division" (<https://stackoverflow.com/questions/14822184/is-there-a-ceiling-equivalent-of-operator-in-python?noredirect=1&lq=1>)

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
+        while new_whales:
+            whales.update(new_whales)
+            whales = self._reduce_bids(whales)
+            new_whales = self.get_whales()
+
+        receipt = self.worklock_agent.force_refund(checksum_address=self.checksum_address,
+                                                   addresses=list(whales.keys()))
+
+        if self.get_whales(force_read=True):
+            raise self.BidderError(f"Internal error: offline simulation differs from transaction results")
```
KPrasch: `raise WhaleError`

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
+        return receipt
+
+    # TODO better control: max iterations, interactive mode
+    def verify_bidding_correctness(self, gas_limit: int) -> dict:
+        self._ensure_cancellation_window(ensure_closed=True)
+
+        if self.worklock_agent.bidders_checked():
```
KPrasch: Is this reflective of on-chain state, or can a "force" mode be implemented?

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
         """Refund ethers for completed work"""
         if not self.available_refund:
             raise self.BidderError(f'There is no refund available for {self.checksum_address}')
         receipt = self.worklock_agent.refund(checksum_address=self.checksum_address)
         return receipt
 
+    def withdraw_compensation(self) -> dict:
+        """Withdraw compensation after force refund"""
+        if not self.available_compensation:
+            raise self.BidderError(f'There is no compensation available for {self.checksum_address}')
```
KPrasch: Perhaps more details here as an escape hatch: "...did you mean to call 'refund deposit'?"

nucypher/blockchain/eth/agents.py
```diff
@@ -1027,6 +1027,25 @@ def cancel_bid(self, checksum_address: str) -> dict:
...
+        addresses = sorted(addresses, key=str.casefold)
+        contract_function = self.contract.functions.forceRefund(addresses)
+        receipt = self.blockchain.send_transaction(contract_function=contract_function, sender_address=checksum_address)
+        return receipt
+
+    @validate_checksum_address
+    def verify_bidding_correctness(self,
+                                   checksum_address: str,
+                                   gas_limit: int,  # TODO - #842: Gas Management
+                                   gas_to_save_state: int = 30000) -> dict:
```
KPrasch: Can you remind me how `30000` became the default here?

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,361 @@ contract WorkLock {
...
+
+        uint256 baseETHSupply = bidders.length * minAllowedBid;
+        // when all participants bid with the same minimum amount of eth
+        if (bonusETHSupply == 0) {
+            return ethToWork(_ethAmount, tokenSupply, baseETHSupply);
+        }
+
+        uint256 baseETH = 0;
+        uint256 bonusETH = 0;
+        // only base part of bid and base part of work
```
arjunhassard:
```suggestion
        // If the staker's total remaining deposit (including the specified sum of ETH to reclaim) is lower than the minimum bid size, then only the base rate is used to calculate the work required to reclaim ETH
```

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,361 @@ contract WorkLock {
...
+    */
+    function ethToWork(uint256 _ethAmount, uint256 _remainderOfDepositedETH) internal view returns (uint256) {
+
+        uint256 baseETHSupply = bidders.length * minAllowedBid;
+        // when all participants bid with the same minimum amount of eth
+        if (bonusETHSupply == 0) {
+            return ethToWork(_ethAmount, tokenSupply, baseETHSupply);
+        }
+
+        uint256 baseETH = 0;
```
arjunhassard:
```suggestion
        uint256 baseRateETH = 0;
```

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,361 @@ contract WorkLock {
...
+        uint256 bonusETH = 0;
```
arjunhassard:
```suggestion
        uint256 bonusRateETH = 0;
```

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,361 @@ contract WorkLock {
...
+        // only base part of bid and base part of work
+        if (_ethAmount + _remainderOfDepositedETH <= minAllowedBid) {
+            baseETH = _ethAmount;
+        // remaining part of bid greater than min so _ethAmount is only bonus part
```
arjunhassard:
```suggestion
        // If the staker's remaining deposit, minus the specified sum of ETH to reclaim, is still greater than the minimum bid size, then only the bonus rate is used to calculate the work required to reclaim ETH
```

nucypher/blockchain/eth/actors.py
```diff
@@ -1643,84 +1653,201 @@ def __init__(self,
...
+        return whales
+
+    def force_refund(self) -> dict:
+        self._ensure_cancellation_window(ensure_closed=True)
+
+        whales = self.get_whales()
+        if not whales:
+            raise self.BidderError(f"All bids are allowable, force refund is not needed")
+
+        new_whales = whales
```
KPrasch: Hmm... unsure whether we're referencing or making a copy here :thinking: 

```suggestion
        new_whales = whales.copy()
```

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,361 @@ contract WorkLock {
...
+        } else if (_remainderOfDepositedETH >= minAllowedBid) {
+            bonusETH = _ethAmount;
+        // _ethAmount has both base and bonus parts
```
arjunhassard:
```suggestion
        // If the staker's remaining deposit, minus the specified sum of ETH to reclaim, is lower than the minimum bid size, then both the base rate and bonus rate must be used to calculate the work required to reclaim ETH
```

nucypher/blockchain/economics.py
vzotova: Maybe I did it wrong, I just thought that these values should be accessible from BaseEconomics._default

nucypher/blockchain/eth/actors.py
vzotova: This error could be if something wrong with formula or if input list of whales includes regular bidder

nucypher/blockchain/eth/actors.py
vzotova: Could you elaborate?

nucypher/blockchain/eth/actors.py
vzotova: Changed to `WhaleError`

nucypher/blockchain/eth/actors.py
vzotova: I'm for `RuntimeError` here, what do you think?

nucypher/blockchain/eth/actors.py
vzotova: I'm for not force, because after force refund will be onchain representation in cache and if we have whales in cache then something goes wrong and let raise exception

nucypher/blockchain/eth/agents.py
vzotova: It's safe amount to save state in verrify method in contract:
```
        if (index != nextBidderToCheck) {
            emit BiddersChecked(msg.sender, nextBidderToCheck, index);
            nextBidderToCheck = index;
        }
```
Value depends on gas costs for saving state and emitting event

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,361 @@ contract WorkLock {
...
+    function ethToWork(uint256 _ethAmount, uint256 _tokenSupply, uint256 _ethSupply)
+        internal view returns (uint256)
+    {
+        return _ethAmount.mul(_tokenSupply).mul(SLOWING_REFUND).divCeil(_ethSupply.mul(boostingRefund));
+    }
+
+    /**
+    * @notice Calculate amount of work that need to be done to refund specified amount of ETH
+    * @dev This value will be fixed only after end of bidding
+    * @param _ethAmount Specified amount of ETH
```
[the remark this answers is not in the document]
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,361 @@ contract WorkLock {
...
+    * @param _remainderOfDepositedETH Base part of ETH amount for correct calculation of work
```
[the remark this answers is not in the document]
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: Accepted with few changes
✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: Accepted with few changes
✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: Accepted with few changes
✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: I'm not sure about this changes, because it's base part of bid and difficult to see here rate

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: Same as above

nucypher/blockchain/eth/actors.py
KPrasch: Just trying to give names to `a` and `b`.

nucypher/blockchain/eth/agents.py
KPrasch: As a matter of admin UX - How can these values be determined for the user?

vzotova: only try to run with different values

nucypher/cli/painting.py
```diff
@@ -854,41 +855,47 @@ def echo_solidity_version(ctx, param, value):
...
-    remaining = end - maya.now()
+    bidding_start = MayaDT(worklock_agent.contract.functions.startBidDate().call())
+    bidding_end = MayaDT(worklock_agent.contract.functions.endBidDate().call())
+    cancellation_end = MayaDT(worklock_agent.contract.functions.endCancellationDate().call())
 
-    # TODO: Include calculated refund and deposit rates
+    bidding_duration = bidding_end - bidding_start
+    cancellation_duration = cancellation_end - bidding_start
+    now = maya.now()
+    bidding_remaining = bidding_end - now if bidding_end > now else timedelta()
```
KPrasch: Hmm - what is the intention of `else timedelta()` here?

nucypher/cli/painting.py
```diff
@@ -898,8 +905,8 @@ def paint_bidder_status(emitter, bidder):
     message = f"""
 WorkLock Participant {bidder.checksum_address}
 =====================================================
-Total Bid ............ {bidder.get_deposited_eth}
-Available Refund ..... {bidder.available_refund}
+Total Bid ............ {prettify_eth_amount(bidder.get_deposited_eth)}
+Available Refund ..... {prettify_eth_amount(bidder.available_refund)}
```
KPrasch: What about including available compensation also?

nucypher/cli/painting.py
```diff
@@ -854,41 +855,47 @@ def echo_solidity_version(ctx, param, value):
...
-Duration .......... {duration}
-Time Remaining .... {remaining} 
+Bidding Start Date ................... {bidding_start}
+Bidding End Date ..................... {bidding_end}
+Cancellation Window End Date ......... {cancellation_end}
+Bidding Duration ..................... {bidding_duration}
+Cancellation Window Duration ......... {cancellation_duration}
+Bidding Time Remaining ............... {bidding_remaining} 
+Cancellation Window Time Remaining ... {cancellation_remaining} 
+Claiming is available ................ {'Yes' if worklock_agent.is_claiming_available() else 'No'}
```
KPrasch:
```suggestion
Claiming available ................ {'Yes' if worklock_agent.is_claiming_available() else 'No'} 
```

nucypher/cli/commands/worklock.py
```diff
@@ -264,20 +266,84 @@ def refund(general_config, worklock_options, registry_options, force, hw_wallet)
 
 
 @worklock.command()
+@group_general_config
+@group_registry_options
+@group_worklock_options
+@option_force
+@option_hw_wallet
+@click.option('--gas-limit', help="Gas limit per each verification transaction", type=click.IntRange(min=60000))
+# TODO: Consider moving to administrator (nucypher-deploy) #1758
```
KPrasch: What's preventing this from being relocated to `nucypher-deploy` /  Is there some motive for inclusion here?

nucypher/cli/commands/worklock.py
```diff
@@ -115,7 +115,7 @@ def status(general_config, registry_options, worklock_options):
 @group_worklock_options
 @option_force
 @option_hw_wallet
-@click.option('--value', help="ETH value of bid", type=click.STRING)
+@click.option('--value', help="ETH value of bid", type=click.FloatRange(min=0))
```
KPrasch: Can this min. be based on economics?

tests/blockchain/eth/entities/actors/test_bidder.py
```diff
@@ -51,21 +56,90 @@ def test_cancel_bid(testerchain, agency, token_economics, test_registry):
...
+    assert not worklock_agent.is_claiming_available()
+
+
+def test_force_refund(testerchain, agency, token_economics, test_registry):
+    bidder_address = testerchain.client.accounts[0]
+    bidder = Bidder(checksum_address=bidder_address, registry=test_registry)
+    whales = bidder.get_whales()
+
+    # Simulate force refund
+    new_whales = whales
```
KPrasch: `whales.copy()` for consistency.

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -23,63 +23,93 @@ contract WorkLock {
...
+    uint256 public maxAllowableLockedTokens;
+    uint256 public minAllowableLockedTokens;
+
+    /**
+    * @dev Checks timestamp regarding cancellation window
+    */
+    modifier afterCancellationWindow()
+    {
+        require(block.timestamp >= endCancellationDate,
+            "Operation is allowed when cancellation phase is over");
```
KPrasch:
```suggestion
            "Operation not allowed during cancellation phase ");
```

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -23,63 +23,93 @@ contract WorkLock {
...
         uint256 totalSupply = _token.totalSupply();
-        require(totalSupply > 0 &&
-            _escrow.secondsPerPeriod() > 0 &&
-            _endBidDate > _startBidDate &&
-            _endBidDate > block.timestamp &&
-            _boostingRefund > 0 &&
-            _stakingPeriods >= _escrow.minLockedPeriods());
+        require(totalSupply > 0 &&                              // token contract is deployed and accessible
+            _escrow.secondsPerPeriod() > 0 &&                   // escrow contract is deployed and accessible
+            _endBidDate > _startBidDate &&                      // bidding period lasts some time
```
KPrasch: At *least* one second, haha.

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,370 @@ contract WorkLock {
...
     * @notice Deposit tokens to contract
     * @param _value Amount of tokens to transfer
     */
     function tokenDeposit(uint256 _value) external {
-        require(block.timestamp <= endBidDate, "Can't deposit more tokens after end of bidding");
+        require(block.timestamp < endBidDate, "Can't deposit more tokens after end of bidding");
         token.safeTransferFrom(msg.sender, address(this), _value);
         tokenSupply += _value;
         emit Deposited(msg.sender, _value);
+        require(tokenSupply / minAllowableLockedTokens <= uint128(0) - 1, "Potential overflow for bidder index");
```
KPrasch: ..."Potential" overflow?  What is this condition indicating?

nucypher/cli/commands/worklock.py
derekpierre: There was some discussion about this in the (option 2) PR: https://github.com/nucypher/nucypher/pull/1732#discussion_r387111897, and https://github.com/nucypher/nucypher/pull/1732#discussion_r386985913.

@vzotova filed a follow-up issue: https://github.com/nucypher/nucypher/issues/1758

nucypher/cli/painting.py
vzotova: to show zero value instead of negative

nucypher/cli/painting.py
vzotova: Not sure about this because it will be for very few addresses and may confuse other bidders

nucypher/cli/commands/worklock.py
vzotova: I have no idea how to inject token_economics here, because it will be available only after reading from blockchain

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: Hmm, we have also time before cancelation phase so it's why here "after"  but maybe it's ok to change to "during"

nucypher/cli/commands/worklock.py
KPrasch: Hah, right... perhaps we can set the type inside the scope of the function, lazily.

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
vzotova: It means that possible situation when number of bidders will be so big and will be overflow in indexes. I know it's not realistic but prefer to check such things

nucypher/cli/painting.py
vzotova: ✔️

tests/blockchain/eth/entities/actors/test_bidder.py
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
KPrasch: I appreciate your dislike of undefined behavior :-)

nucypher/blockchain/economics.py
KPrasch: You want access by *class*  - Is this implemented somewhere access by *instance* cannot be used?  The original intention of double underscore here is to ensure that economic instance attributes are not accidentally mutated during runtime; however, if this is presenting a challenge regarding the inheritance of defaults, it's acceptable to set the to single underscores IMO.

nucypher/cli/painting.py
KPrasch: Will there ever be a practical need to obtain this information as *any* kind of user?

nucypher/cli/painting.py
derekpierre: We should be careful here. Indicating that `Claims available` may make a bidder assume that they have bids that can be claimed. As opposed to `Claiming available` which seems to better apply to the more general window to perform claims.

I think I prefer `Claiming available` or `Claiming phase open`. Thoughts? @KPrasch , @vzotova ?

nucypher/blockchain/economics.py
vzotova: I rely on your opinion, so what is the best way to do this?

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,370 @@ contract WorkLock {
...
+    {
+        return _ethAmount.mul(_tokenSupply).mul(SLOWING_REFUND).divCeil(_ethSupply.mul(boostingRefund));
+    }
+
+    /**
+    * @notice Calculate amount of work that need to be done to refund specified amount of ETH
+    * @dev This value will be fixed only after end of bidding
+    * @param _ethToReclaim Specified sum of ETH staker wishes to reclaim following completion of work
+    * @param _restOfDepositedETH Remaining ETH in staker's deposit once ethToReclaim sum has been subtracted
+    * @dev ethAmount + remainderOfDepositedETH = depositedETH
```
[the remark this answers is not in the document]
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -109,74 +144,331 @@ contract WorkLock {
...
+        return block.timestamp >= endCancellationDate &&
+            nextBidderToCheck == bidders.length;
+    }
+
     /**
     * @notice Claimed tokens will be deposited and locked as stake in the StakingEscrow contract.
     */
-    function claim() external returns (uint256 claimedTokens) {
-        require(block.timestamp >= endBidDate, "Claiming tokens allowed after bidding is over");
+    function claim() external afterCancellationWindow returns (uint256 claimedTokens) {
```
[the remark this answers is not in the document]
vzotova: ✔️

nucypher/cli/painting.py
vzotova: Added printing of compensation if there is any

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,370 @@ contract WorkLock {
...
     function bid() external payable {
-        require(block.timestamp >= startBidDate && block.timestamp <= endBidDate,
-            "Bid is open during a certain period");
+        require(block.timestamp >= startBidDate, "Bidding is not open yet");
+        require(block.timestamp < endBidDate, "Bidding is already finished");
         WorkInfo storage info = workInfo[msg.sender];
+
+        // first bid
+        if (info.depositedETH == 0) {
+            require(msg.value >= minAllowedBid, "Bid must be at least minimum");
```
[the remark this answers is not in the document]
vzotova: Some bidders will see min bid and will try with that value which will result in an error. And it will be need to explain why need to put bid 10 ETH + 1 wei, so I'm for `>=`. But want to hear other opinions

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -89,94 +119,370 @@ contract WorkLock {
...
+
+        // all participants bid with the same minimum amount of eth
+        uint256 index = nextBidderToCheck;
+        if (bonusETHSupply == 0) {
+            require(tokenSupply / bidders.length <= maxAllowableLockedTokens, "Not enough bidders");
+            index = bidders.length;
+        }
+
+        uint256 maxBonusTokens = maxAllowableLockedTokens - minAllowableLockedTokens;
+        uint256 bonusTokenSupply = tokenSupply - bidders.length * minAllowableLockedTokens;
```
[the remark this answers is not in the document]
vzotova: Indeed interesting, this will make this method very cheap (for this case) but will slightly increase gas cost for other cases. I guess if it's realistic to achieve such number of bidders let's add this

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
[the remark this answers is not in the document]
arjunhassard: I agree with @vzotova , it's a weird UX to state the minimum as 10 ETH but then force refund bids of that size. It's very unlikely that not a single bidder exceeds the minAllowedBid anyway

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
arjunhassard: Yeh, I'm not thrilled about my suggestion, but can't think of anything better. It's important people understand that this variable equals the segment of the ETH deposit that can be reclaimed at a 'base' ETH/NU rate, not the rate itself.

nucypher/cli/painting.py
arjunhassard: I vote for `Claiming phase open` for the reasons @derekpierre stated. Even better: we consistently use '**Re**claiming/**re**claim' in all worklock files

vzotova: Changed to `Claiming phase open`

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
derekpierre: I'm not sure I see a "rate" either - we could make the variable more explanatory eg. `baseETHToReclaim` and `bonusETHToReclaim` since they are just respective proportions of the `_ethToReclaim`

KPrasch:
Manually tested via CLI - bid, cancelled, then bid again: <https://gist.github.com/KPrasch/a16c92554a8d09bfd1a5919b417a207e>

PR #1785, opened 2020-03-20
Manual shutdown for worklock

### LINKED ISSUE #1776: Provide for manual shutdown of WorkLock contract by allowing shutdown() to be owner-callable.

vzotova:
Fixes #1776
Opens #1787

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
@@ -4,6 +4,7 @@ pragma solidity ^0.5.3;
 import "zeppelin/math/SafeMath.sol";
 import "zeppelin/token/ERC20/SafeERC20.sol";
 import "zeppelin/utils/Address.sol";
+import "zeppelin/ownership/Ownable.sol";
 import "contracts/NuCypherToken.sol";
 import "contracts/StakingEscrow.sol";
 import "contracts/lib/AdditionalMath.sol";
@@ -12,7 +13,7 @@ import "contracts/lib/AdditionalMath.sol";
 /**
 * @notice The WorkLock distribution contract
 */
-contract WorkLock {
+contract WorkLock is Ownable {
     using SafeERC20 for NuCypherToken;
     using SafeMath for uint256;
     using AdditionalMath for uint256;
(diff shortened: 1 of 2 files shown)


nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -338,11 +338,22 @@ contract WorkLock {
...
+    /**
+    * @notice Cancels distribution, makes possible to retrieve all bids and owner gets all tokens
+    */
+    function internalShutdown() internal {
+        startBidDate = 0;
+        endBidDate = 0;
         endCancellationDate = uint256(0) - 1; // "infinite" cancellation window
-        token.safeTransfer(creator, tokenSupply);
+        token.safeTransfer(owner(), tokenSupply);
+        emit Shutdown(msg.sender);
```
derekpierre: 🎉

michwill:
Am I right that all bidders would get their ETH back as a result of shutdown / cancelling bids? (should it be in tests?)

vzotova:
> Am I right that all bidders would get their ETH back as a result of shutdown / cancelling bids? (should it be in tests?)

Shutdown makes cancellation phase infinite, so each bidder can call `cancelBid` anytime (and this is in tests)

michwill:
Ahh, right, and cancelBid is tested separately :+1:

nucypher/blockchain/eth/sol/source/contracts/WorkLock.sol
```diff
@@ -338,11 +338,22 @@ contract WorkLock {
     }
 
     /**
-    * @notice Cancels distribution, makes possible to retrieve all bids and creator gets all tokens
+    * @notice Cancels distribution, makes possible to retrieve all bids and owner gets all tokens
     */
-    function shutdown() internal {
+    function shutdown() external onlyOwner {
+        require(!isClaimingAvailable(), "Claiming has already been enabled");
```
KPrasch: s the any mitigation or follow up for this condition?

vzotova: Why need any mitigations?

PR #1788, opened 2020-03-23
Staking contracts updates

vzotova:
This PR is preparation for "pooling contract" PR
Attempt to make `PreallocationEscrow` just as an example of a staking contract

* `WorkLock` methods in `StakingInterface`
* Updates for `AbstractStakingContract`
* Refactoring for tests

Versions:
`StakingInterface` : `v1.3.1` -> `v1.4.1`

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1127,7 +1127,7 @@ class Staker(NucypherTokenActor):
         """Collect rewarded ETH."""
         if self.is_contract:
             if collector_address and collector_address != self.beneficiary_address:
-                raise ValueError("Allowed to withdraw policy reward only to the beneficiary address")
+                raise ValueError("Policy rewards must be withdrawn to the beneficiary address")
             self.preallocation_escrow_agent.collect_policy_reward()  # TODO save receipt
             receipt = self.preallocation_escrow_agent.withdraw_eth()
         else:
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1126,8 +1126,10 @@ class Staker(NucypherTokenActor):
     def collect_policy_reward(self, collector_address=None) -> dict:
         """Collect rewarded ETH."""
         if self.is_contract:
-            withdraw_address = collector_address or self.beneficiary_address
-            receipt = self.preallocation_escrow_agent.collect_policy_reward(collector_address=withdraw_address)
+            if collector_address and collector_address != self.beneficiary_address:
+                raise ValueError("Allowed to withdraw policy reward only to the beneficiary address")
+            self.preallocation_escrow_agent.collect_policy_reward()  # TODO save receipt
+            receipt = self.preallocation_escrow_agent.withdraw_eth()
         else:
             withdraw_address = collector_address or self.checksum_address
             receipt = self.policy_agent.collect_policy_reward(collector_address=withdraw_address,
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -1029,9 +1029,7 @@ class PreallocationEscrowDeployer(BaseContractDeployer, UpgradeableContractMixin
         self.check_deployment_readiness()
         router_contract = self.blockchain.get_contract_by_name(registry=self.registry,
                                                                contract_name=self._router_deployer.contract_name)
-        constructor_args = (router_contract.address,
-                            self.token_contract.address,
-                            self.staking_escrow_contract.address)
+        constructor_args = (router_contract.address,)
 
         deployer_address = self.sidekick_address if use_sidekick else self.deployer_address
         self._contract, deploy_receipt = self.blockchain.deploy_contract(deployer_address,
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -48,8 +48,8 @@ from nucypher.blockchain.eth.decorators import validate_secret, validate_checksu
 from nucypher.blockchain.eth.interfaces import (
     BlockchainDeployerInterface,
     BlockchainInterfaceFactory,
-    VersionedContract
-)
+    VersionedContract,
+    BlockchainInterface)
 from nucypher.blockchain.eth.registry import AllocationRegistry
 from nucypher.blockchain.eth.registry import BaseContractRegistry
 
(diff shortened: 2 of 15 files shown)


nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -542,6 +542,59 @@ contract StakingEscrow is Issuer {
...
+    /**
+    * @notice Batch deposit. Allowed only initial deposit for each staker
+    * @param _stakers Stakers
+    * @param _numberOfSubStakes Number of sub-stakes which belong to staker in _values and _periods arrays
+    * @param _values Amount of tokens to deposit for each staker
+    * @param _periods Amount of periods during which tokens will be locked for each staker
+    */
+    function batchDeposit(
+        address[] calldata _stakers,
+        uint256[] calldata _numberOfSubStakes,
```
derekpierre: Is the max number of sub-stakes for a staker 30. In which case do we need a `uint256[]` for each of these numbers? seems excessive...or is this some solidity/gas magic 🐇 🎩 ?

nucypher/blockchain/eth/sol/source/contracts/staking_contracts/StakingInterface.sol
```diff
@@ -5,62 +5,99 @@ import "contracts/staking_contracts/AbstractStakingContract.sol";
...
+        WorkLock _workLock
     )
         public
     {
         require(_token.totalSupply() > 0 &&
             _escrow.secondsPerPeriod() > 0 &&
-            _policyManager.secondsPerPeriod() > 0);
+            _policyManager.secondsPerPeriod() > 0 &&
+            // in case there is no worklock contract
+            (address(_workLock) == address(0) || _workLock.startBidDate() > 0));
```
derekpierre: Does `no worklock` in this case describe the situation where contracts are deployed (but not initialized) before WorkLock is deployed?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
michwill: I am not sure that this would do any significant optimization. In the code we have `j` / `endIndex` which can be potentially large, so if `_numberOfSubStakes` was `uint8`, we'd probably need to do a type conversion every time. Which also obstructs code readability.
And this is call parameters, not storage (where storing less could be very worthwhile)

nucypher/blockchain/eth/actors.py
```diff
@@ -1119,8 +1119,10 @@ def calculate_policy_reward(self) -> int:
     def collect_policy_reward(self, collector_address=None) -> dict:
         """Collect rewarded ETH."""
         if self.is_contract:
-            withdraw_address = collector_address or self.beneficiary_address
-            receipt = self.preallocation_escrow_agent.collect_policy_reward(collector_address=withdraw_address)
+            if collector_address and collector_address != self.beneficiary_address:
+                raise ValueError("Allowed to withdraw policy reward only to the beneficiary address")
```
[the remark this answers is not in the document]

nucypher/blockchain/eth/sol/source/contracts/staking_contracts/StakingInterface.sol
vzotova: It's more for situation when we deploy only main contracts (for example, some testnet)

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: @michwill said everything correct. Just want to add that type reducing helps only when we talk about contract state (and then not always)

derekpierre: 👍

nucypher/blockchain/eth/deployers.py
```diff
@@ -870,11 +870,20 @@ def __init__(self, *args, **kwargs):
                                                                     contract_name=policy_contract_name,
                                                                     proxy_name=policy_proxy_name)
 
+        worklock_name = WorklockDeployer.contract_name
+        try:
+            self.worklock_contract = self.blockchain.get_contract_by_name(registry=self.registry,
+                                                                          contract_name=worklock_name)
+        except BaseContractRegistry.UnknownContract:
+            self.worklock_contract = None
```
KPrasch: Does this need follow once deployed on mainnet?

vzotova: I thought to use this moment more for testnets without worklock, so depending on that

tests/blockchain/eth/contracts/contracts/StakingContractsTestSet.sol
```diff
@@ -94,10 +97,10 @@ contract PolicyManagerForStakingContractMock {
     uint32 public secondsPerPeriod = 1;
     uint256 public minRewardRate;
 
-    function withdraw(address payable _recipient) public returns (uint256) {
+    function withdraw() public returns (uint256) {
         uint256 value = address(this).balance;
         require(value > 0);
-        _recipient.transfer(value);
+        msg.sender.transfer(value);
```
KPrasch, line 103: Same as comment above about using separate collection addresses.

nucypher/blockchain/eth/sol/source/contracts/staking_contracts/StakingInterface.sol
```diff
@@ -144,17 +181,17 @@ contract StakingInterface {
...
 
     /**
     * @notice Withdraw available reward from the policy manager to the staking contract
     */
-    function withdrawPolicyReward(address payable _recipient) public {
-        uint256 value = getStateContract().policyManager().withdraw(_recipient);
-        emit PolicyRewardWithdrawn(_recipient, value);
+    function withdrawPolicyReward() public {
+        uint256 value = getStateContract().policyManager().withdraw();
+        emit PolicyRewardWithdrawn(msg.sender, value);
```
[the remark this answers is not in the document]
vzotova: To make possible to control eth flows, otherwise owner will withdraw everything. But with this changes - eth will go to staking contract and it means each staking control must decide who can withdraw and how much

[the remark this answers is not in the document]
cygnusv: Ah yes, it makes a lot of sense. Nothing to see here then :)

tests/blockchain/eth/contracts/integration/test_intercontract_integration.py
```diff
@@ -828,7 +828,9 @@ def test_all(testerchain,
     testerchain.wait_for_receipt(tx)
     assert staker2_balance < testerchain.client.get_balance(staker2)
     staker3_balance = testerchain.client.get_balance(staker3)
-    tx = preallocation_escrow_interface_1.functions.withdrawPolicyReward(staker3).transact({'from': staker3, 'gas_price': 0})
+    tx = preallocation_escrow_interface_1.functions.withdrawPolicyReward().transact({'from': staker3, 'gas_price': 0})
+    testerchain.wait_for_receipt(tx)
+    tx = preallocation_escrow_1.functions.withdrawETH().transact({'from': staker3, 'gas_price': 0})
```
cygnusv, line 896: I'm curious to know why do you need an explicit gas_price of 0 here and in other tests.

vzotova: to control balance, in that case balance can be changed only with funds from contract

PR #1811, opened 2020-03-27
First iteration of pooling contract

### LINKED ISSUE #1674: Implement a delegation/pooling smart contract
Many commercial staking operators take a pooling approach that conflicts with NuCypher's 1:1 mapping between stakers and workers. Fortunately, stakers can be smart contracts which means we can implement pooling via a smart contract staker.

It's also important that this contract includes some sort of fee logic, where the staking service can set a fee (likely a % of staking (NU inflation) and/or policy (ETH fees) rewards).

vzotova:
Refs #1674
Consider this PR as starting point for pooling contract
Opens #2202

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/staking_contracts/PoolingStakingContract.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/staking_contracts/PoolingStakingContract.sol
@@ -12,36 +12,36 @@ import "contracts/staking_contracts/AbstractStakingContract.sol";
 contract PoolingStakingContract is AbstractStakingContract, Ownable {
     using SafeMath for uint256;
 
-    event TokensDeposited(address indexed sender, uint256 value, uint256 coefficient);
+    event TokensDeposited(address indexed sender, uint256 value, uint256 depositedTokens);
     event TokensWithdrawn(address indexed sender, uint256 value);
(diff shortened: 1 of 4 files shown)


nucypher/blockchain/eth/sol/source/contracts/staking_contracts/PoolingStakingContract.sol
```diff
@@ -0,0 +1,260 @@
...
+
+    struct Delegator {
+        uint256 depositedTokens;
+        uint256 withdrawnReward;
+        uint256 withdrawnETH;
+    }
+
+    StakingEscrow public immutable escrow;
+
+    uint256 public totalDepositedTokens;
```
cygnusv, line 28: I think we should change this approach. Instead of assuming we're tracking here deposited tokens in absolute numbers, we can follow an approach where users get a share (i.e. percentage) of the pool. The cool thing about this is that we could consider tokenize participation in these pools, similarly to what DeFi projects are currently doing. That means that delegators receive an amount of the pool's own token (let's call it `pNU` for the sake of the argument), in exchange for their `NU`. This would drastically improve the viability of staking for smaller participants, as it will provide them of quick entry and exit (e.g, by buying/selling `pNU` in Uniswap).

cygnusv (approved):
Great addition! I agree with @KPrasch in adding a README or something

Issue #1698, opened 2020-02-15
Invalid worker Stamp In Tests


KPrasch:
```
Unhandled error during node learning: Traceback (most recent call last):
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/twisted/internet/base.py", line 1283, in run
    self.mainLoop()
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/twisted/internet/base.py", line 1292, in mainLoop
    self.runUntilCurrent()
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/twisted/internet/base.py", line 913, in runUntilCurrent
    call.func(*call.args, **call.kw)
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/twisted/internet/task.py", line 239, in __call__
    d = defer.maybeDeferred(self.f, *self.a, **self.kw)
--- <exception caught here> ---
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "/home/k/Git/nucypher/nucypher/network/nodes.py", line 626, in keep_learning_about_nodes
    self.learn_from_teacher_node(eager=False)
  File "/home/k/Git/nucypher/nucypher/network/nodes.py", line 808, in learn_from_teacher_node
    response = self.network_middleware.get_nodes_via_rest(node=current_teacher,
  File "/home/k/Git/nucypher/nucypher/network/middleware.py", line 260, in get_nodes_via_rest
    response = self.client.get(node_or_sprout=node,
  File "/home/k/Git/nucypher/nucypher/network/middleware.py", line 117, in method_wrapper
    host, node_certificate_filepath, http_client = self.verify_and_parse_node_or_host_and_port(node_or_sprout, host, port)
  File "/home/k/Git/nucypher/nucypher/network/middleware.py", line 66, in verify_and_parse_node_or_host_and_port
    node.verify_node(network_middleware_client=self)
  File "/home/k/Git/nucypher/nucypher/network/nodes.py", line 1214, in verify_node
    self.validate_metadata(registry=registry)
  File "/home/k/Git/nucypher/nucypher/network/nodes.py", line 1175, in validate_metadata
[...]

PR #1697, opened 2020-02-15
Idle Ursula on startup until funded and bonded

KPrasch:
Based on #1693 

- Full support for configuring and running a Worker *before* a staking, bonding, or funding.
- Allows `Ursula` in be instantiated without a staker address
- Creates a new blocking `staticmethod`: `Worker.block_until_ready` that waits for the worker address to be funded and bonded with an optional timeout (default).
- Do not collect staker address interactively via CLI when configuring a `Worker`
- Provide interactive worker configuration file selection on `nucypher ursula run`

Resolves #1656.

PR #1819, opened 2020-04-01
Bug fixes for Ursula: availability sensor, cert/stamp verification, nicknames

KPrasch:
Fixes several regressions introduced in #1462 and #1697 

Advances fixes for https://github.com/nucypher/nucypher/issues/1698, #1823, and #1820 

Tested in live production environment.

### DIFF
--- a/nucypher/cli/painting.py
+++ b/nucypher/cli/painting.py
@@ -138,7 +138,7 @@ def paint_node_status(emitter, ursula, start_time):
         stats.extend([current_period, worker_address])
 
     if ursula._availability_sensor:
-        score = 'Availability Score .. {}'.format(ursula._availability_sensor.score)
+        score = 'Availability Score .. {} ({} responders)'.format(ursula._availability_sensor.score, len(ursula._availability_sensor.responders))
         stats.append(score)
 
     emitter.echo('\n' + '\n'.join(stats) + '\n')
--- a/nucypher/cli/painting.py
+++ b/nucypher/cli/painting.py
@@ -138,7 +138,11 @@ def paint_node_status(emitter, ursula, start_time):
         stats.extend([current_period, worker_address])
 
     if ursula._availability_sensor:
-        score = 'Availability Score .. {} ({} responders)'.format(ursula._availability_sensor.score, len(ursula._availability_sensor.responders))
+        if ursula._availability_sensor.running:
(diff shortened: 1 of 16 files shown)


derekpierre:
@KPrasch a reminder of my comment in #1462: https://github.com/nucypher/nucypher/pull/1462#discussion_r401049059

Basically, if each failed check results in the reduction of the score by a factor of 0.9:
`new_score = 0 + 0.9 * <previous_score>`
then we would never get a score of 0 for the `shutdown_threshold`. Instead, we could set the `shutdown_threshold` to some low non-zero value. For example, setting the shutdown threshold to be `1` would require 22 consecutive failed checks (assuming ursulas start at a score of `10`) for the score to become <= 1 and force a shutdown.

nucypher/network/sensors.py
```diff
@@ -43,31 +43,36 @@ def __init__(self, ursula, enforce_loneliness: bool = True):
             9: self.mild_warning,
             7: self.medium_warning,
             2: self.severe_warning,
-            0: self.shutdown_everything
+            1: self.shutdown_everything  # 0 is unobtainable
```
KPrasch, line 46: @derekpierre :heavy_check_mark:

nucypher/cli/painting.py
```diff
@@ -517,7 +521,12 @@ def paint_stakes(emitter, stakeholder, paint_inactive: bool = False, staker_addr
         missing = staker.missing_confirmations
         min_reward_rate = prettify_eth_amount(staker.min_reward_rate)
 
-        staker_data = [f'Missing {missing} confirmation{"s" if missing > 1 else ""}' if missing else f'Confirmed #{last_confirmed}',
+        if missing == -1:
+            missing_info = "Never Confirmed (New Stake)"
+        else:
+            missing_info = f'Missing {missing} confirmation{"s" if missing > 1 else ""}' if missing else f'Confirmed #{last_confirmed}'
```
cygnusv, line 527: I think we can unroll these conditionals too.

derekpierre:
Some subsequent comments from @tuxxy regarding the scale of the scores.

```
I just mean that beginning at 100 makes the number more natural/fine.
...
At 10, with 7 missed checks, you're < 5. At 100, with 7 missed checks, you're < 50.
...
the distance between a score of 3 to 5 feels much smaller than it actually is.
compared to a score of 30 to 50.
or even 300 to 500.
```

PR #1822, opened 2020-04-01
Preparation for Gemini

cygnusv:


derekpierre:
@cygnusv @vepkenez : there are some ansible files that have the following text in them `"{{ nucypher_hosts | default('cassandra') }}"`, do we need to change the `cassandra` default? 

eg. `get_worker_status.yml`, `check_running_ursula.yml`, `init_worker.yml`, `run_geth.yml`...?

PR #1831, opened 2020-04-02
Relock Deps

KPrasch:
Possible fix for #1837

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -52,7 +52,7 @@ py-solc = {git = "https://github.com/nucypher/py-solc.git",ref = "v5.0.0-eol.0"}
 ansible = "*"
 bumpversion = "*"
 # Documentation
-sphinx = "*"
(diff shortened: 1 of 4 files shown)


derekpierre:
@KPrasch re: docs build failure.

We seem to be using a pre-release version of sphinx (`sphinx==3.0.0b1`). When I switched to version `2.4.4` locally and then the build the docs, everything was fine.

KPrasch:
Hmm - Looking into `/home/kieran/Git/nucypher/docs/source/guides/development/getting_started.md:124:Inline literal start-string without end-string.` when building docs on the new version of sphinx... perhaps we will roll it back like you mentioned above @derekpierre

derekpierre:
Perhaps for the moment - I could see find any workaround for the pre-released version - bigger fish to fry 🐟 😄

PR #1855, opened 2020-04-07
Add keystore signer

banteg:
This pull request adds an additional signer which works very much like in [pymaker](https://github.com/makerdao/pymaker/blob/master/pymaker/keys.py).

It uses `eth_account` and works with keystore files directly. The new signer can be used like this:
```
nucypher ursula run --signer key_file=keystore.json,pass_file=password.txt
```
`pass_file` is optional, if it's omitted, user will be prompted for a password.

One notable limitation is that only one account can be added this way.

### DIFF
--- a/nucypher/blockchain/eth/signers.py
+++ b/nucypher/blockchain/eth/signers.py
@@ -1,7 +1,10 @@
 from abc import ABC, abstractmethod
-from typing import List
+import getpass
+from typing import List, Optional
 from urllib.parse import urlparse
 
+from eth_account import Account
+from eth_account.messages import encode_defunct
 from eth_utils import to_checksum_address, to_normalized_address, apply_formatters_to_dict
 from hexbytes import HexBytes
 from twisted.logger import Logger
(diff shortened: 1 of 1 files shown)


nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +196,49 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+            read_pass = getpass.getpass(prompt=f"Password for {key_file}: ")
+        priv = Account.decrypt(read_key, read_pass)
+        account = Account.from_key(priv)
+        self.registred_keys[account.address] = account
+
+    def is_device(self, account: str) -> bool:
+        return True
+
+    def accounts(self) -> List[str]:
+        return list(self.registred_keys)
```
KPrasch, line 229: :+1:

KPrasch:
Great work here - Leaving this open as a base but moving efforts to PR #1858

KPrasch:
Moved to #1858 followed by #1922 - Thanks for your contribution @banteg - Feel free to review ongoing changes in this area :-)

PR #1858, opened 2020-04-08
Add keystore signer implemenation for ursula only

svenski123:
Adds a local signer implementation for use by ursula only.
Signer factory function is now dispatched by URL scheme (i.e. clef:..., key:..., ipc:..., etc.).
Local key store signer is unlocked same way as existing web3 signer so passing password via NUCYPHER_WORKER_ETH_PASSWORD environment variable works.
Infact existing keystore file and password from local geth node can be used as is.
Class level boolean flags added to Signer and CharacterConfiguration to allow only ursula to make use of local signing.

This builds on @banteg's work in PR #1855 and should be combined with that effort.


svenski123:
The address needs to be available before the private key is available (if ever) - callers may iterate over Signer.accounts and then choose one to unlock, etc.

I've added a sanity that when the account is unlocked, if the public key computed from the private key doesn't match the one read from the file, it throws.

Also replaced storing the private key directly with a eth_account.LocalAccount wrapper as in your PR - a bit of extra overhead but hides the keying material nicely.

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+        self._account = Web3.toChecksumAddress(address)
+        self._signer = None
+
+    def _is_my_account(self, account) -> bool:
+        return Web3.isAddress(account) and \
+            self.account == Web3.toChecksumAddress(account)
+
+    @classmethod
+    def from_signer_uri(cls, uri: str) -> 'Signer':
+        return cls(uri)
```
KPrasch, line 249: Hmm - Just a passthrough method here?  Just a suggestion: Parse the URI itself here, then passing the keyfile or password into the `__init__`.  This starts to suggest that this class is more aptly named `KeyfileSigner`, `LocalPrivateKeySigner`, etc.

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+    @property
+    def account(self) -> str:
+        return self._account
+
+    @validate_checksum_address
+    def is_device(self, account: str) -> bool:
+        return False
+
+    def accounts(self) -> List[str]:
+        return [self.account]
```
KPrasch, line 260: Interesting moment here - To live up to it;s name What do you think about the `KeystoreSigner` being able to manage the the keystore *directory*, reading the private key into memory just-in-time?  This will be a bit more consistent with other signing providers.

Alternately - IMO it's okay to disable all other account management if using a `KeystoreSigner`.

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+        if self._signer is None:
+            try:
+                self._signer = Account.from_key(
+                    Account.decrypt(self.key, password)
+                )
+            except ValueError:
+                return False
+
+            if self.account != self._signer.address:
+                raise ValueError("unexpected address '%s' recovered from private key; expected '%s'" % (
```
KPrasch, line 276: Can you update this to an f-string? :-)  

We require python 3.6+ for installation.

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+            transaction_dict = dissoc(transaction_dict, 'to')
+
+        return self._signer.sign_transaction(
+                transaction=transaction_dict,
+            ).raw
+
+    @validate_checksum_address
+    def sign_message(self, account: str, message: bytes, **kwargs) -> HexBytes:
+        if not self._is_my_account(account):
+            return None
```
KPrasch, line 304: This is a very odd scenario (current design expects multi-account usage)... Let's raise an exception instead.

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+
+        return self._signer.sign_message(
```
KPrasch, line 306: Just a style preference, but we prefer if return values are assigned to a variable to easier debugging.

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+
+        self._signer = None
+        return True
+
+    @validate_checksum_address
+    def sign_transaction(self, transaction_dict: dict) -> HexBytes:
+        # Do not include a 'to' field for contract creation.
+        if not transaction_dict['to']:
+            transaction_dict = dissoc(transaction_dict, 'to')
+
```
KPrasch, line 296: This method needs to have a sanity check to verify that the 'from' field matches the local account...  Is that included in `Account.sign_transaction`?

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+        with open(decoded_uri.path) as file_obj:
+            key = json.load(file_obj)
+
+        address = key['address']
+        if not Web3.isAddress(address):
+            raise ValueError(f"'{decoded_uri.path}' does not contain a valid address")
+
+        self.key = key
+        self._account = Web3.toChecksumAddress(address)
```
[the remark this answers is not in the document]
svenski123: The single underscore already indicates it's private, mangling the name with a second underscore shouldn't be necessary (and stil doesn't make it const either), but if that extra emphasis is desired, I have absolutely no objection.

The other members should also be private, particularly self.key which is an implementation detail.

web3.Web3 doesn't export to_checksum_address() just toChecksumAddress() - as per published api is here:
https://web3py.readthedocs.io/en/stable/overview.html#Web3.toChecksumAddress

Of course looking at the web3 code web3/main imports to_checksum_address from eth_utils, which Web3 then exposes as toChecksumAddress(), so you can call the same underlying function any of these ways:

* eth_utils.to_checsum_address
* web3.main.to_checksum_address
* web3.Web3.toChecksumAddress

Which of the two underscore_names is preferable?

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+        self._signer = None
+
+    def _is_my_account(self, account) -> bool:
+        return Web3.isAddress(account) and \
```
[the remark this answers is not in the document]
svenski123: It wasn't clear to me that parameter account was guaranteed to be a checksum address as opposed to just a hexadecimal address hence the isAddress check and toChecksumAddress and ultimately why I put it in a separate private method.

If the simple equality check suffices - or even an toChecksumAddress(account) == self.account - I would just inline it and drop the function.

No issue with doc string, if the method stays I'll add one.

nucypher/blockchain/eth/signers.py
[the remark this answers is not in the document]
KPrasch: > The single underscore already indicates it's private, mangling the name with a second underscore shouldn't be necessary (and stil doesn't make it const either), but if that extra emphasis is desired, I have absolutely no objection.

I'd prefer double underscore here to prevent setting the attribute explicitly; I also noticed this attribute is exposed as a `@property` which suggests the "read-only" nature of the namespace mangling.

> Which of the two underscore_names is preferable?

For consistency with the rest of the codebase, use `eth_utils.to_checsum_address`.

nucypher/blockchain/eth/signers.py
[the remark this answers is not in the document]
KPrasch: > It wasn't clear to me that parameter account was guaranteed to be a checksum address as opposed to just a hexadecimal address

This validation is encapsulated in the `@validate_checksum_address` decorator; However, IIRC, keystore filenames are in normalized format?

nucypher/blockchain/eth/signers.py
banteg: Yes, [here](https://github.com/ethereum/eth-account/blob/82d1205e5d1ba591734356cf65b7c0c16eade579/eth_account/account.py#L638)

KPrasch: Yes - good.  Do we want the `TypeError` to bubble up here?

nucypher/blockchain/eth/signers.py
svenski123: '%' formatting works in 3.6+ and hasn't been deprecated, it's also more readable (in this case) than having the expressions in line...

Of course if f-strings are what the code convention calls for, who am I to argue?

nucypher/blockchain/eth/signers.py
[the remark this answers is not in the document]
banteg: Do you mean address field in the keystore? Judging by the files I found on my machine, they are lowercase and no 0x prefix.

nucypher/blockchain/eth/signers.py
KPrasch: Heh - Trying to start back up the the f-string holy wars :joy:

nucypher/blockchain/eth/signers.py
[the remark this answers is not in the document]
KPrasch: > Do you mean address field in the keystore?

Yes.

> Judging by the files I found on my machine, they are lowercase and lack 0x prefix.

Sounds like they are in the normalized format.  We can just ensure that we cast to checksum address at the first possible moment.

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+        super().__init__()
+
+        decoded_uri = urlparse(uri)
+        if decoded_uri.scheme not in ('key', 'keystore') or decoded_uri.netloc:
+            raise cls.InvalidSignerURI(uri)
+
+        with open(decoded_uri.path) as file_obj:
+            key = json.load(file_obj)
+
+        address = key['address']
```
KPrasch, line 235: Is there any situation in which `KeyError` might be raised here?

nucypher/blockchain/eth/signers.py
```diff
@@ -191,3 +216,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+    def lock_account(self, account: str) -> str:
+        if not self._is_my_account(account):
+            return False
+
+        self._signer = None
+        return True
+
+    @validate_checksum_address
+    def sign_transaction(self, transaction_dict: dict) -> HexBytes:
+        # Do not include a 'to' field for contract creation.
```
KPrasch, line 293: What is the expected behavior in the case of invoking this method with a locked account?

nucypher/blockchain/eth/signers.py
svenski123: Actually supporting only a single file and requiring its name kind of sucks, especially as the keystore file names that geth uses are incredibly long and vary depending on address.

Multiple account support it is then (#1855 was half way there too).

banteg: Does supporting a keystore dir even make sense if we only allow this for Usula?

nucypher/blockchain/eth/signers.py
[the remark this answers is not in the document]
svenski123: Done - member names double underscored, camel case functions replaced with eth_utils.*.

nucypher/blockchain/eth/signers.py
svenski123: Done - URI decoded has been factored out as suggested.

nucypher/blockchain/eth/signers.py
svenski123: Done - the % formatting strings have been f'd up. :)

nucypher/blockchain/eth/signers.py
svenski123: Done - sign_message / sign_transaction now assign values to be returned to local variables prior to returning said variables.

nucypher/blockchain/eth/signers.py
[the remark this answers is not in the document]
svenski123: Ok, as @validate_checksum_address ensures passed in accounts are in checksum format, I've simply removed _is_my_account() and its to_checksum_address() calls.

The address from the key store is canonicalised once in the ctor.

nucypher/blockchain/eth/signers.py
svenski123: Yes but only if the key store file is corrupt or invalid.

nucypher/blockchain/eth/signers.py
svenski123: An AttributeError will be raised as self.__signer will be None if the account is locked.

Web3Signer/sign_* methods would raise a ValueError (assuming the RPC calls return an error result).

If something cleaner is needed, I'd suggest adding another exception type to Signer (i.e. `LockedAccount`) and having all the signers throw this when signing with a locked account.

nucypher/blockchain/eth/signers.py
svenski123: I should think so - I believe you would get an RPC error in this situation with Web3Signer.

nucypher/blockchain/eth/signers.py
svenski123: The sign_* methods will now throw in case of locked or unknown accounts and the exception message text should be helpful in these cases.

nucypher/blockchain/eth/signers.py
svenski123: Done - sign_message() now throws if the account is unexpected.

nucypher/blockchain/eth/signers.py
svenski123: Ok, thanks for the comments and issues raised - the current commit on this PR branch should hopefully address all of them adequately and be suitable as usable single-key-store-file implementation for the worker.

Adding multi-file support is not trivial but not a big stretch and could be made backward compatible - perhaps this is best left to another PR.

nucypher/blockchain/eth/signers.py
```diff
@@ -194,3 +219,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+        if not self.__signer:
+            raise ValueError('account is locked')
+
+        # Do not include a 'to' field for contract creation.
+        if not transaction_dict['to']:
+            transaction_dict = dissoc(transaction_dict, 'to')
+
+        raw_transaction = self.__signer.sign_transaction(
+                transaction=transaction_dict,
+            ).raw
```
banteg, line 295:
```suggestion
            ).rawTransaction
```

nucypher/blockchain/eth/signers.py
```diff
@@ -194,3 +219,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+    def sign_message(self, account: str, message: bytes, **kwargs) -> HexBytes:
+        if not self.__signer:
+            raise ValueError('account is locked')
+
+        if self.__account != account:
+            raise ValueError(f"account '{account}' is unknown")
+
+        signature = self.__signer.sign_message(
+                signable_message=encode_defunct(primitive=message),
+            )['signature']
```
banteg, line 309: for consistency sake
```suggestion
            ).signature
```

svenski123:
Thanks for the bug fix and edit @banteg - both applied.

nucypher/blockchain/eth/signers.py
```diff
@@ -194,3 +220,93 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
+        if not self.__signer:
+            try:
+                self.__signer = Account.from_key(
+                    Account.decrypt(self.__key, password)
+                )
+            except ValueError:
+                return False
+
+            if self.__account != self.__signer.address:
+                raise ValueError(f"unexpected address '{self.__signer.address}' recovered from private key; expected '{self.__account}'")
```
derekpierre, line 273: Since `unlock_account` failed, should we set `self.__signer` to `None` (i.e. what it was prior to the function call) before raising the error?

nucypher/blockchain/eth/signers.py
derekpierre: Should we be more explicit with that error instead of a `KeyError`? We could do a `get()` and if the returned address is `None`, throw a `ValueError` indicating file may be invalid or corrupt.

nucypher/blockchain/eth/signers.py
```diff
@@ -15,17 +19,38 @@ class Signer(ABC):
 
     class InvalidSignerURI(ValueError):
         """Raised when an invalid signer URI is detected"""
+        def __init__(self, uri=None, text=None):
```
derekpierre, line 22: It feels like we should just have the creator of the `InvalidSignerURI` provide the full text for the error, instead of having logic to incorporate the `uri` i.e. I don't think we should have the `uri` parameter for the function.

nucypher/blockchain/eth/signers.py
svenski123: Constructing a KeyStoreSigner can raise the following non-exclusive exceptions:

- `FileNotFoundError` or `PermissionError` in `open`
- `JSONDecodeError' in `json.load`
- `TypeError` in `key['address']` if the deserialised JSON was not an object (i.e. array, string, number, etc.)
- `KeyError` in `key['address']` if there's no address key in the object
- `ValueError` raised explicitly if `key['address']` isn't a valid Ethereum address

There will be other possible exceptions that could conceivably crop up here.
The question is whether there is anything intelligent or useful `KeyStoreSigner.__init__()` can do about it here or whether it is better placed to interpret the exception and add useful information for the caller.

It seems to me the answer is no here but I of course could be mistaken - the fact that any exception occurs here tells the caller construction of the signer object failed which is the key information. Unless there is a generalised exception hierarchy specifically intended to allow callers to effectively handle correctable errors programatically in a general way then I see no point in adding additional code to normalise python exceptions.

But that is just my humble opinion!

nucypher/blockchain/eth/signers.py
svenski123: That seems reasonable.

nucypher/blockchain/eth/signers.py
svenski123: As one would not raise an InvalidSignerURI unless there were an invalid URI to complain about, having the ctor take the uri makes sense.

Text formatting is not necessary and in fact the base exception class stuff all of the arguments into self.args anyways which appear when the exception is printed (i.e. __str__() returns them), so have added *args / **kwargs to __init__() and left it that.

derekpierre: My hesitation was that InvalidSignerURI constructor allowed `uri` to default `None` which then leads to extra logic within `__init__` for checking whether the `uri` is set or not. If that is your expectation then `uri` should not default to `None` but rather we should force a value to be provided.

If `uri` is indeed forced to be passed, then the logic in `__init__` can be boiled down to:
```python
    text = f"'{uri}': {text or 'is not a valid signer URI'}"
    args.append(text)
    super().__init__(*args)
```

nucypher/blockchain/eth/signers.py
derekpierre: The other exceptions are clearer to me. If we are trying to be clear about the source of this problem, in my mind, a `KeyError` is much less clear than a `ValueError` with a message that the file is invalid or corrupt. 

That being said, this is not a make or break for the PR - just a thought.

derekpierre:
@svenski123 this is a great addition to the codebase. Thanks!

nucypher/blockchain/eth/signers.py
svenski123: Latest commits takes a minimilist approach - any exceptions raised in Signer.from_signer_uri() are wrapped in an InvalidSignerURI (unless it's an InvalidSignerURI already).

As far as Ursula goes, all CryptoError and ValueError exceptions are caught in actions.make_cli_character() which just displays a red line with no information from the exception which is perhaps a bit too brief - however that's out of scope of this PR.

As for InvalidSignerURI, it possible to set attributes such as `uri` and `inner_exc` on the exception object so upstream handlers can inspect them but as none currently do, I've gone with minimal.

svenski123:
@derekpierre Thanks again for your reviews!

The PR is ready to go from my perspective, I've tested it with Ursula - both sign_message() and sign_transaction() were hit.

I've included two other unrelated small bug fixes that I came across in `nucypher/cli/actions.py` and `nucypher/blockchain/eth/token.py` - these are in distinct commits and can be cherry picked if desired.

nucypher/blockchain/eth/signers.py
```diff
@@ -231,3 +253,100 @@ def unlock_account(self, account: str, password: str, duration: int = None) -> b
...
     def lock_account(self, account: str):
         return True
+
+
+class KeyStoreSigner(Signer):
+    """Local Web3 signer implementation supporting a single account/keystore file"""
+
+    IS_SIGNER_LOCAL = True
+
+    __key: Dict
```
KPrasch, line 263: Woah - Is this the dataclass syntax :-)  

This means we'd be committing to not supporting 3.6 (only 3.7+), correct?

<https://www.python.org/dev/peps/pep-0557/>

michwill: Uhm... I'd still be supporting 3.6, tbh

svenski123: Those are instance variable type hints in the class body as per PEP 526, no \@dataclass decorator.

https://www.python.org/dev/peps/pep-0526/#class-and-instance-variable-annotations

For what it's worth I use the stock python on Centos 7 which is 3.6.8; I haven't noticed any issues with it:
```
Python 3.6.8 (default, Apr 25 2019, 21:02:35)
[GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux
Type "help", "copyright", "credits" or "license" for more information.
```

KPrasch: Ah, yeah I see - They look similar!

KPrasch:
I've been developing / running based on this PR already for a few days.  It's a solid thoughtful addition with high UX value!

Let's continue to iterate in #1922, move some of this PRs comments to docstrings to be picked up by `autodoc`, and open any follow up Issues.

:cowboy_hat_face:

KPrasch:
Needs a rebase over master to resolve a small conflict in `actions.py` then this PR is read for merge.

Issue #1774, opened 2020-03-16
Heartbeat demo passes on CI despite errors


KPrasch:
Original Report:

```
Traceback (most recent call last):
  File "doctor.py", line 13, in <module>
    from nucypher.datastore.keypairs import DecryptingKeypair, SigningKeypair
ModuleNotFoundError: No module named 'nucypher.datastore'
```

> How is it possible the Demo test didn't failed?

> Looking closer here the job appears  to pass but is actually rasing an (unrelated) error, The CircleCI Demos must be misconfigured <https://circleci.com/gh/nucypher/nucypher/75485>

```
TypeError: retrieve() got an unexpected keyword argument 'message_kit'
```

KPrasch:
@vepkenez - <https://circleci.com/gh/nucypher/nucypher/78008>

KPrasch:
Closed in #2581

PR #1872, opened 2020-04-10
fix doctor.retrieve param in heartbeat_demo example

obernardovieira:
When running *doctor.py* from the _heartbeat_demo_ example, I've got an error

```
  File "doctor.py", line 103, in <module>
    alice_verifying_key=alices_sig_pubkey
TypeError: retrieve() got an unexpected keyword argument 'message_kit'
```

Since I'm using **2.1.0b2**, I've decided to had a look at the function definition [here](https://github.com/nucypher/nucypher/blob/2da817ae1a958a6c53e8738b31c48bfbaad0c94f/nucypher/characters/lawful.py#L680). From that, I did a little change and it worked.

### DIFF
--- a/examples/heartbeat_demo/doctor.py
+++ b/examples/heartbeat_demo/doctor.py
@@ -97,8 +97,8 @@ for message_kit in message_kits:
     try:
         start = timer()
         retrieved_plaintexts = doctor.retrieve(
+            message_kit,
             label=label,
-            message_kit=message_kit,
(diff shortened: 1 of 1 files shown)


KPrasch:
Related to #1774

obernardovieira:
gosh, I looked for a PR or an issue related to this and didn't see that one :man_facepalming:

PR #1873, opened 2020-04-10
Autodetect POA

### LINKED ISSUE #1857: Autoconfigure POA
The value for the POA flag is fixed for each network (e.g., it's false for mainnet, true for goerli, etc). Maybe we should autodetect this and configure it to the default for each network.

cygnusv:
Closes #1857

### DIFF
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -46,7 +46,7 @@ from web3.exceptions import ValidationError
 from web3.gas_strategies import time_based
 from web3.middleware import geth_poa_middleware
 
-from nucypher.blockchain.eth.clients import EthereumClient
+from nucypher.blockchain.eth.clients import EthereumClient, POA_CHAINS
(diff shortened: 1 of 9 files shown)


nucypher/blockchain/eth/interfaces.py
```diff
@@ -230,14 +230,17 @@ def get_gas_strategy(cls, gas_strategy: Union[str, Callable]) -> Callable:
         return gas_strategy
 
     def attach_middleware(self):
+        if self.poa is None:  # If POA is not set explicitly, try to autodetect from chain id
+            chain_id = int(self.client.chain_id)
+            self.poa = chain_id in POA_CHAINS
```
KPrasch, line 235: :+1:

cygnusv:
Sure, I can work on that too :)

nucypher/blockchain/eth/interfaces.py
KPrasch: As long as the value of `poa is None` from CLI input or configuration files, this auto-detection will work. :+1:

Issue #281, opened 2018-06-01
Versioning Arrangements


jMyles:
We want room after launch to make changes to the Arrangement format once we make observations about network activity.  Let's add a version to Arrangements so that Alice can always choose an Ursula who supports her Arrangement version.

(And it's possible that other objects and their respective APIs will need this too.)

KPrasch:
Is this still open? It's like we need a versioning TODO list. :-)

jMyles:
Yeah, I think this is actually still reasonably open.

KPrasch:
Closed by #2832

PR #1878, opened 2020-04-11
Versioning and "Futureproofing" of bytes serialized Nucypher products

### LINKED ISSUE #343: Versioning TreasureMaps
For the same reasons that #281 makes sense, we probably need to version TreasureMaps.

vepkenez:
When we go to mainnet and people start making Treasuremaps, Arrangements and MessageKits and storing them in side channels, one day we may make a change that breaks compatibility with some of those existing bytestrings in the wild.  This is supposed to solve that.

Goals:
* nobody has to change anything
* everything will 'just' work
* adds seamless hooks to accommodate any future unforeseen changes 
closes #343 and #281 

- [x] figure out how to do this
- [x] implement it for message kits
- [x] exception handling
- [ ] do these versions factor into learning loop/availability/sampling/network? (discussion needed)
- [x] implementation and tests for treasure maps 
- [x] implementation and tests for whatever else

### DIFF
--- a/nucypher/crypto/kits.py
+++ b/nucypher/crypto/kits.py
@@ -67,7 +67,7 @@ class MessageKit(CryptoKit):
             as_bytes += bytes(self.sender_verifying_key)
 
         as_bytes += VariableLengthBytestring(self.ciphertext)
-        return super().add_version(as_bytes)
+        return super().prepend_version(as_bytes)
 
     @classmethod
     def splitter(cls, *args, **kwargs):
@@ -81,7 +81,7 @@ class MessageKit(CryptoKit):
         return self._signature
 
     def __bytes__(self):
-        return super().add_version(bytes(self.capsule) + VariableLengthBytestring(self.ciphertext))
+        return super().prepend_version(bytes(self.capsule) + VariableLengthBytestring(self.ciphertext))
 
 
 class MessageKitV1(MessageKit):
--- a/nucypher/crypto/kits.py
+++ b/nucypher/crypto/kits.py
@@ -17,9 +17,10 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 from constant_sorrow.constants import UNKNOWN_SENDER, NOT_SIGNED
 from bytestring_splitter import BytestringKwargifier, VariableLengthBytestring
 from nucypher.crypto.splitters import key_splitter, capsule_splitter
+from nucypher.primitives import VersionedBytes
 
 
-class CryptoKit:
+class CryptoKit(VersionedBytes):
     """
     A package of discrete items, meant to be sent over the wire or saved to disk (in either case, as bytes),
     capable of performing a distinct cryptological function.
@@ -35,7 +36,8 @@ class CryptoKit:
         return splitter(some_bytes)
 
     @classmethod
-    def from_bytes(cls, some_bytes):
+    def from_bytes(cls, input_bytes):
+        cls, some_bytes = super().parse_version(input_bytes)
         return cls.split_bytes(some_bytes)
 
 
@@ -65,7 +67,7 @@ class MessageKit(CryptoKit):
             as_bytes += bytes(self.sender_verifying_key)
 
         as_bytes += VariableLengthBytestring(self.ciphertext)
-        return as_bytes
+        return super().add_version(as_bytes)
 
     @classmethod
     def splitter(cls, *args, **kwargs):
@@ -79,13 +81,18 @@ class MessageKit(CryptoKit):
         return self._signature
 
     def __bytes__(self):
-        return bytes(self.capsule) + VariableLengthBytestring(self.ciphertext)
+        return super().add_version(bytes(self.capsule) + VariableLengthBytestring(self.ciphertext))
+
+
+class MessageKitV1(MessageKit):
+    version = 1
 
 
 class PolicyMessageKit(MessageKit):
     """
     A MessageKit which includes sufficient additional information to be retrieved on the NuCypher Network.
     """
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._sender = UNKNOWN_SENDER.bool_value(False)
@@ -102,6 +109,10 @@ class PolicyMessageKit(MessageKit):
 
     def __bytes__(self):
         return super().to_bytes(include_alice_pubkey=True)
+
+
+class PolicyMessageKitV1(PolicyMessageKit):
+    version = 1
 
 
 UmbralMessageKit = PolicyMessageKit  # Temporarily, until serialization w/ Enrico's
--- a/nucypher/crypto/kits.py
+++ b/nucypher/crypto/kits.py
@@ -17,10 +17,10 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 from constant_sorrow.constants import UNKNOWN_SENDER, NOT_SIGNED
 from bytestring_splitter import BytestringKwargifier, VariableLengthBytestring
 from nucypher.crypto.splitters import key_splitter, capsule_splitter
-from nucypher.primitives import VersionedBytes
+from nucypher.utilities.versioning import ByteVersioningMixin
 
 
-class CryptoKit(VersionedBytes):
+class CryptoKit(ByteVersioningMixin):
     """
     A package of discrete items, meant to be sent over the wire or saved to disk (in either case, as bytes),
     capable of performing a distinct cryptological function.
--- a/nucypher/primitives.py
+++ b/nucypher/primitives.py
@@ -21,13 +21,17 @@ class VersionedBytes:
 
     @classmethod
     def __get_class(cls, klass, version):
-        v = int.from_bytes(version, 'little')
+        v = int.from_bytes(version, 'big')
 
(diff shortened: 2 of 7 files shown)


nucypher/primitives.py
```diff
@@ -0,0 +1,66 @@
...
+            return klass
+
+        versioned_subclasses = VersionedBytes.__get_versioned_subclasses(klass)
+        try:
+            outclass = next(iter([c for c in versioned_subclasses if c.version == v]))
+        except StopIteration:
+
+            if len(versioned_subclasses) and 99 >= v > 1:
+                # We have received data that was clearly created by a newer version of Nucypher,
+                # TODO:  I don't know exactly what to do here.
```
KPrasch, line 39: Needs a formal Issue to follow up.

nucypher/primitives.py
```diff
@@ -0,0 +1,66 @@
...
+                # would a bob be receiving this?  Or an Alice?
+                # Who needs to know about this?
+                # can we notify the staker of a node that their Worker needs an update?
+
+                raise VersionedBytes.NucypherNeedsUpdateException("This node is running outdated NuCypher code")
```
KPrasch, line 44: I think - this may be too generic for all cases, at minimum can we include the expected entity version number here?

tests/crypto/test_bytestring_types.py
```diff
@@ -76,3 +81,101 @@ def test_message_kit_serialization_via_enrico(enacted_federated_policy, federate
...
+    mkit, signature = enrico.encrypt_message(message=plaintext_bytes)
+    # simulate an enrico from the future creating a message kit unsupported by this install
+
+    v99 = (99).to_bytes(2, 'big')
+    V99_mkit_bytes = (
+        v99 + bytes(mkit.capsule) + bytes(mkit.sender_verifying_key) + VariableLengthBytestring(mkit.ciphertext))
+
+    with pytest.raises(VersionedBytes.NucypherNeedsUpdateException):
+        # we should catch this NucypherNeedsUpdateException here
+        UmbralMessageKit.from_bytes(V99_mkit_bytes)
```
KPrasch, line 181: Nice!

nucypher/primitives.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if not hasattr(cls, 'version'):
+            versioned_subclasses = VersionedBytes.__get_versioned_subclasses(cls)
+            # take the highest
+            cls = sorted(versioned_subclasses, key=lambda x: x.version)[-1]
+        return super(VersionedBytes, cls).__new__(cls)
+
+    @classmethod
+    def __get_versioned_subclasses(cls, klass):
+        return [cls for cls in klass.__subclasses__() if hasattr(cls, 'version')]
```
KPrasch, line 23: If I understand correctly - this is an effort to allow for continued backward compatibility of versioned entities?

nucypher/crypto/kits.py
```diff
@@ -17,9 +17,10 @@
 from constant_sorrow.constants import UNKNOWN_SENDER, NOT_SIGNED
 from bytestring_splitter import BytestringKwargifier, VariableLengthBytestring
 from nucypher.crypto.splitters import key_splitter, capsule_splitter
+from nucypher.primitives import VersionedBytes
 
 
-class CryptoKit:
+class CryptoKit(VersionedBytes):
```
KPrasch, line 23: Phew - I can appreciate the semi-polymorphic action implemented via inheritance here, but I'm unsure I agree that a `CryptoKit` or `TreasureMap` *is a type of* `VersionedBytes` since they also exist in non-serialized form.

Curious: Is it instead possible for versioned serializable entities to *have* a `VersionedBytes` or similar class resolver *without* using inheritance here?

nucypher/primitives.py
```diff
@@ -0,0 +1,39 @@
...
+        versioned_subclasses = VersionedBytes.__get_versioned_subclasses(klass)
+        outclass = next(iter([c for c in versioned_subclasses if c.version == v]))
+        return outclass
+
+    @classmethod
+    def parse_version(cls, some_bytes):
+        version_bytes = some_bytes[:2]
+        return VersionedBytes.__get_class(cls, version_bytes), some_bytes[2:]
+
+    def add_version(self, some_bytes):
```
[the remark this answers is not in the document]
vepkenez: Yeah, I was planning on making this change.

nucypher/crypto/kits.py
derekpierre: Agreed. It feels like you could simply use a `VersionedBytes` helper for the `__bytes__` and `from_bytes` functions since the same class does the serialization and deserialization - I don't think anything cares whether some object/class is an instance of `VersionedBytes` - but I can accept that I may be missing nuanced aspects of this.

vepkenez: Yeah you guys are right, I'm giving that a try.

nucypher/primitives.py
vepkenez: Yep.

nucypher/crypto/kits.py
vepkenez: @derekpierre @KPrasch I'm having trouble working out how to do make this change because of this one  specific construct.

One major piece of the mechanics here which is a big part of making this work seamlessly is that we use this `__new__` method.

```
def __new__(cls, *args, **kwargs):
        """
        When instantiating a specific version of a versioned Bytestring,
        we want to automatically instantiate the latest version without
        requiring implementors to have to deal with any of this.

        If we are instantiating a specifically versioned class, allow for that also
        """

        if not hasattr(cls, 'version'):
            versioned_subclasses = ByteVersioner.__get_versioned_subclasses(cls)
            # take the highest
            cls = sorted(versioned_subclasses, key=lambda x: x.version)[-1]
        return super(VersionedBytes, cls).__new__(cls)
```

This makes it so when you do something like `messagekit = MessageKit(*args, **kwargs)` you get a MessageKitVersion1 or if you do `MessageKit.from_bytes(<v5>...<bytes>)` you get a MessageKitVersion5.

Can one of you suggest a structure to do this as a helper method without using `__new__` that will have the same benefits?

I mean... this would be ok I guess.
```
messagekit = MessageKit.version_util.latest(*args, **kwargs)
```
and then in `to_bytes`
```
self.version_util.prepend_version(output_bytes)
```

The only real difference would be that we'd have to always update the
`UmbralMessageKit` reference to make sure it's the latest version, but this itself to me is an inferior situation with only arguable stylistic benefits.

What do you guys think about all this?

vepkenez: Maybe it's more of a naming problem.
`ByteVersioningHelperMixin`?

KPrasch: No doubt this *seems* like a mixin, however, `VersionedBytes` performs an operation the subclass *requires*, which is not in the spirit of a mixin for some added functionality.

I think it's possible to keep the `VersionedBytes` class (perhaps renaming it to `VersionedSerializer`), while decoupling the versioning of *serialization* independently from the entity itself.  

What does this look like? Instead of having multiple `MessageKit` classes, we'd end up with multiple versioned `MessageKitSerializer`s that can be composed *on* a `MessageKit` 

IMO this separation of logic will end up being more testable, accurate and maintainable since it draws a clear separation between serialization vs native-object abstraction layers and only one `MessageKit` class needs to be maintained.  

In fact - you can already see evidence of this pattern emerging in your test wherein `MessageKitVersion19` *only* needs to override serialization methods `__bytes__` and `splitter`.

vepkenez: > In fact - you can already see evidence of this pattern emerging in your test wherein MessageKitVersion19 only needs to override serialization methods __bytes__ and splitter.

This is not true.   I specially wrote [this test](https://github.com/nucypher/nucypher/pull/1878/files#diff-77ab6f80bf0c9dcad115c140306fcac2R127) to show that this implementation supports arbitrary addition of methods or any other changes a class might need over the hopefully long future.

I understand what you're saying but I think that it's slightly presumptuous to say "the only thing that will ever change on our classes will be the byte splitter and the __bytes__ method".  I think that to sacrifice the flexibility we get from the way this is setup in the service of some arguable compositional style benefits is not the right decision.

That being said @KPrasch I defer to your view of the scope here and will gladly make this change if you specifically acknowledge these concerns I have laid out here.

KPrasch: > the only thing that will ever change on our classes will be the byte splitter and the bytes method"

For the scope of `VersionedBytes`, this is actually True - The implementation and methods of the entity being serialized is independent from the serializer's.

KPrasch: > I think that to sacrifice the flexibility we get from the way this is setup...

The design I proposed above does not sacrifice this functionality, rather, builds upon your excellent engineering.

>  ...in the service of some arguable compositional style benefits is not the right decision.

This is more than style - My concerns are a matter of API deigned in a abstraction layer that is both accurate and maintainable.

vepkenez: 👍 on it.

KPrasch: In any case - I think this consideration deserves the attention of @jMyles and @cygnusv :-)

vepkenez:
Ok, closing this for now and moving some of this logic to https://github.com/nucypher/bytestringSplitter

Issue #1867, opened 2020-04-09
Request source IP does mot match Ursulas metadata if Worker IP changes.


KPrasch:

KPrasch:
Closing as duplicate of #2478

Issue #1866, opened 2020-04-09
Store validity status in serialized Nodes


jMyles:
If we validate a Node, and that Node's metadata does not change, there is no reason to validate it again after deserializing it.

(I'll think more later.)

KPrasch:
From #1453 

> If a Learner has already validated metadata or verified a node, there's no reason to do it again at next launch. Let's keep a flag of verification status in node storage.

derekpierre:
Closed via #3307

PR #1881, opened 2020-04-13
Blue oyster mushrooms: Evolving node validity, optimization, domain separation, and migration.

### LINKED ISSUE #1580: Weirdness in network/domains mapping
Even after #1576, we still have to reconcile the fact that `domains` were designed to be a set of identifiers for different learning contexts, while `network` is intended to be an identifier for a single "smart contracts" context.

This problem is specially problematic for StakeHolder as its configuration inherits from `CharacterConfiguration`, which has `domains`, but StakeHolder should have a single `network` instead.

As a consequence, we end up with weird mappings like:
`network=list(self.domains)[0]`  (see e.g., nucypher/config/node.py after #1576)
or the other way around:
`domains={self.network}`  (see e.g., nucypher/cli/commands/stake.py after #1576)

### LINKED ISSUE #1623: Invalid node bytes encountered in node storage
```
sys.exit(nucypher_cli())   File "/home/ubuntu/.local/lib/python3.6/site-packages/click/core.py", line 764, in __call__     return self.main(*args, **kwargs)   File "/home/ubuntu/.local/lib/python3.6/site-packages/click/core.py", line 717, in main     rv = self.invoke(ctx)   File "/home/ubuntu/.local/lib/python3.6/site-packages/click/core.py", line 1137, in invoke     return _process_result(sub_ctx.command.invoke(sub_ctx))   File "/home/ubuntu/.local/lib/python3.6/site-packages/click/core.py", line 1137, in invoke     return _process_result(sub_ctx.command.invoke(sub_ctx))   File "/home/ubuntu/.local/lib/python3.6/site-packages/click/core.py", line 956, in invoke     return ctx.invoke(self.callback, **ctx.params)   File "/home/ubuntu/.local/lib/python3.6/site-packages/click/core.py", line 555, in invoke     return callback(*args, **kwargs)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/cli/options.py", line 134, in wrapper     return func(**kwargs)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/cli/options.py", line 134, in wrapper     return func(**kwargs)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/cli/options.py", line 134, in wrapper     return func(**kwargs)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/cli/commands/ursula.py", line 321, in run     json_ipc=general_config.json_ipc)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/cli/commands/ursula.py", line 225, in create_character     start_learning_now=load_seednodes)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/cli/actions.py", line 350, in make_cli_character     **config_args)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/config/node.py", line 225, in __call__     return self.produce(**character_kwargs)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/config/characters.py", line 104, in produce     ursula = self.CHARACTER_CLASS(**merged_parameters)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/characters/lawful.py", line 949, in __init__     self.start_learning_loop(now=True)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/network/nodes.py", line 547, in start_learning_loop     self.load_seednodes()   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/network/nodes.py", line 440, in load_seednodes     self.read_nodes_from_storage()   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/network/nodes.py", line 447, in read_nodes_from_storage     stored_nodes = self.node_storage.all(federated_only=self.federated_only)  # TODO: #466   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/config/storages.py", line 487, in all     node = self.__read_metadata(filepath=metadata_path, federated_only=federated_only)  # TODO: 466   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/config/storages.py", line 457, in __read_metadata     node = Ursula.from_bytes(node_bytes)   File "/home/ubuntu/.local/lib/python3.6/site-packages/nucypher/characters/lawful.py", line 1151, in from_bytes     version, payload = cls.version_splitter(ursula_as_bytes, return_remainder=True)   File "/home/ubuntu/.local/lib/python3.6/site-packages/bytestring_splitter/__init__.py", line 144, in __call__     raise BytestringSplittingError(message.format(self.nice_message_types(), len(self), len(splittable))) bytestring_splitter.BytestringSplittingError: Not enough bytes to constitute message types int:2 - need 2, got 0
```

### LINKED ISSUE #1820: Handle missing wallet address on TLS certificates
```
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/base.py", line 1283, in run
    self.mainLoop()
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/base.py", line 1292, in mainLoop
    self.runUntilCurrent()
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/base.py", line 913, in runUntilCurrent
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/task.py", line 239, in __call__
    d = defer.maybeDeferred(self.f, *self.a, **self.kw)
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/kieran/Git/nucypher/nucypher/network/sensors.py", line 85, in handle_measurement_errors
    failure.raiseException()
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/python/failure.py", line 488, in raiseException
    raise self.value.with_traceback(self.tb)
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "/home/kieran/Git/nucypher/nucypher/network/sensors.py", line 134, in maintain
    self.measure()
  File "/home/kieran/Git/nucypher/nucypher/network/sensors.py", line 183, in measure
    certificate_filepath = self._ursula.node_storage.store_node_certificate(certificate=certificate)
  File "/home/kieran/Git/nucypher/nucypher/config/storages.py", line 499, in store_node_certificate
    certificate_filepath = self._write_tls_certificate(certificate=certificate, force=force)
  File "/home/kieran/Git/nucypher/nucypher/config/storages.py", line 102, in _write_tls_certificate
    pseudonym = certificate.subject.get_attributes_for_oid(NameOID.PSEUDONYM)[0]
builtins.IndexError: list index out of range

```

### LINKED ISSUE #1843: Problem deserializing domains in node metadata
**Describe the Bug**
Seems that some nodes are experiencing this error, probably related with availability computations?

````
Apr  6 13:45:41 do-nucypher-work nucypher[1275]: Unhandled error in Deferred:
Apr  6 13:45:41 do-nucypher-work nucypher[1275]: Traceback (most recent call last):
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/usr/lib/python3.6/threading.py", line 916, in _bootstrap_inner
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     self.run()
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/usr/lib/python3.6/threading.py", line 864, in run
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     self._target(*self._args, **self._kwargs)
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/twisted/_threads/_threadworker.py", line 46, in work
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     task()
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/twisted/_threads/_team.py", line 190, in doWork
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     task()
Apr  6 13:45:41 do-nucypher-work nucypher[1275]: --- <exception caught here> ---
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/twisted/python/threadpool.py", line 250, in inContext
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     result = inContext.theWork()
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/twisted/python/threadpool.py", line 266, in <lambda>
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     inContext.theWork = lambda: context.call(ctx, func, *args, **kw)
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/twisted/python/context.py", line 122, in callWithContext
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     return self.currentContext().callWithContext(ctx, func, *args, **kw)
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/twisted/python/context.py", line 85, in callWithContext
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     return func(*args,**kw)
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/nucypher/network/server.py", line 203, in learn_about_announced_nodes
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     node.mature()
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/nucypher/network/nodes.py", line 268, in mature
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     mature_node = self.finish()
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/bytestring_splitter/__init__.py", line 53, in finish
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     return self._receiver(**finished_values, **self._additional_kwargs)
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/nucypher/characters/lawful.py", line 1409, in from_processed_bytes
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     domains_vbytes = VariableLengthBytestring.dispense(processed_objects.pop('domains'))
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:   File "/home/nucypher/nucypher-venv/lib/python3.6/site-packages/bytestring_splitter/__init__.py", line 422, in dispense
Apr  6 13:45:41 do-nucypher-work nucypher[1275]:     raise BytestringSplittingError("This does not appear to be a VariableLengthBytestring, or is not the correct length.")
Apr  6 13:45:41 do-nucypher-work nucypher[1275]: bytestring_splitter.BytestringSplittingError: This does not appear to be a VariableLengthBytestring, or is not the correct length.
```
 - OS Version:
 - Nucypher Version:
 - Ethereum Node Version:

### LINKED ISSUE #1869: Test with a hard-coded, versioned node metadata bytestring
Imagine this situation:
1. Let's assume we start with our current LearningLoop version, which is version 1
2. Now I introduce a change in the LearningLoop that affects the metadata and that's not backwards-compatible.
3. In principle, I should also increment the LL version to version 2, but I forget about that.
4. If necessary, I adapt current tests to work with my new changes and move on. All tests pass.
5. The changes are merged introduced in a new release.
6. Suddenly we have nodes in the network with the same LL version but different metadata format.

@jMyles @KPrasch Is this a plausible scenario? If so, I propose the following:

- Introduce a new test specific to version 1. This test will include a hard-coded node metadata bytestring, which will be validated by test ursulas. The purpose of this test is to fail if a change in the codebase affects node metadata and the programmer forgets to increment the version.
- When we need to upgrade  to version 2, we need to produce a new bytestring.

As an alternative, we can introduce a test that disassemblies node metadata bytestrings of current test ursulas, and tries to validate the format against the current specification (currently, version 1). Perhaps this is even easier. We'd have to be very clear that this test cannot be changed, and that it fails is an indication that the format has changed.

### LINKED ISSUE #1871: Increment LearningLoop version and see what happens
Just because.

### LINKED ISSUE #2236: Relocate `nucypher.datastore.keypairs.py` to `nucypher.crypto`
For some reason, keypairs' logic was left in the datastore module after the last reorganization. It should live on `nucypher.crypto`.

jMyles:
Contains the following PRs:
- #2231 
- #2265 
- #2280 

Closes issues:
- Closes #1871: Bumped Learning Loop version from 1 to 2.
- Closes #1869: See new tests in tests/integration/learning/test_learning_versions.py.
- Closes #2236: Relocate keypairs.py from datastore to crypto.
- Closes #1580: For the moment, networks == domains
- Closes #1843: Handles bad domains from stored node metadata
- Closes #1623: Handles bad node metadata from storage
- Closes #1820: Handle missing certificate checksum address pseudonym

Issues to address:
* #1866 
* #1867


KPrasch:
I'd like to suggest that we use the "scope hammer" for Issues regarding migration and include them in another PR.

nucypher/utilities/seednodes.py
```diff
@@ -23,54 +23,48 @@
...
-        if domain_static_nodes:
-            uris.extend(domain_static_nodes)
-
-        # 3 - Hardcoded teachers from module
-        hardcoded_uris = TEACHER_NODES.get(domain)
-        if hardcoded_uris:
-            uris.extend(hardcoded_uris)
-
-    return uris
+# TODO: This module seems unused
```
KPrasch, line 29: https://github.com/nucypher/nucypher/issues/2282

KPrasch (approved):
Turned into a very nice cleanup. :cowboy_hat_face: 

In it goes.

nucypher/blockchain/eth/actors.py
```diff
@@ -104,7 +104,7 @@ class ActorError(Exception):
         pass
 
     @validate_checksum_address
-    def __init__(self, registry: BaseContractRegistry, domains=None, checksum_address: ChecksumAddress = None):
+    def __init__(self, registry: BaseContractRegistry, domain: str = None, checksum_address: ChecksumAddress = None):
```
fjarri, line 107: `Optional` would be better here

nucypher/network/nodes.py
```diff
@@ -163,13 +163,13 @@ class Learner:
     __DEFAULT_MIDDLEWARE_CLASS = RestMiddleware
 
     LEARNER_VERSION = LEARNING_LOOP_VERSION
+    LOWEST_COMPATIBLE_VERSION = 2   # Disallow versions lower than this
```
fjarri, line 166: Feels like a semantic version tuple would be more intuitively understandable here.

cygnusv: YES! I though something like that but also had second thoughts regarding how to encode this efficiently.

tests/integration/learning/test_learning_versions.py
```diff
@@ -0,0 +1,99 @@
...
+
+import pytest
+
+from constant_sorrow.constants import UNKNOWN_VERSION
+
+from nucypher.characters.lawful import Ursula
+from nucypher.config.constants import TEMPORARY_DOMAIN
+from nucypher.network.nodes import Teacher
+
+ursulas_v1 = (
```
fjarri, line 26: Would be useful to have a comment describing how these values were generated.

cygnusv: What if I told you I just ran a debugger in some test that uses the `blockchain_ursulas` fixture, use an interactive console at some breakpoint, and just copy-pasted the output of `print(bytes(ursula).hex())`? yeah I know 🙈 🙈 🙈

fjarri: Well then I'd probably tell you can you don't need to involve a debugger for that, and it can just be made into a script and put in a comment :)

cygnusv: ✅ 
I introduced a "script" that simply uses the `blockchain_ursulas` fixture and print the proper hex-strings. I say "script" because it's actually a test, since I want to reuse some pytest fixtures and I don't want to duplicate any logic. Still a bit sketchy, but sufficient IMO, given how uncommon will be to bump the learning loop version.

nucypher/network/nodes.py
KPrasch: This raises two questions from me:

 - What does a minor vs major bump mean?
 - How and when does the version get updated?

KPrasch:
Needs rebase to resolve conflict

vepkenez:
This PR is deployed on 5 nodes and seems to be completely functional.

nucypher/network/nodes.py
cygnusv: Both good questions. The current design in `main` implicitly assumes all versions are going to be back-wards compatible, in the sense that if you detect a node from a past version, no special handling of the metadata is performed. There are cases, like our particular bump from version 1 to version 2, where this is not desirable (as this is a "major" change in semver parlance), but I certainly can imagine other cases where we bump the version and at the same time try to maintain compatibility with the old version (a "minor" change). The approach taken in this PR is very simplistic, and I think there's room for improvement, but at least ensures that no metadata from version 1 will linger on the network.

KPrasch: :thinking:  Interesting - I suggest we open an Issue to track and leave it for a future PR

fjarri: > What does a minor vs major bump mean?

Major bump == `LOWEST_COMPATIBLE_VERSION` bump
Minor bump == `LEARNER_VERSION` bump
That's essentially what these variables are. Arranging them in a tuple just makes it more obvious.

Now I must say that based on the logic of `Ursula.from_bytes()` it seems that the treatment of versions is more restricted in some ways and more relaxed in others compared to the semver default. What happens now:
- if the version of the serialized data is within a certain range, expect it to conform to the format *exactly* 
- if the version of the serialized data is too far in the past, or higher than the class version, try to decode parts of it anyway (since it's only used for the error message, I guess it's fine)

That raises the question, if `payload_splitter()` can decode something encoded by the code with a different `LEARNER_VERSION` without skipping anything, why was `LEARNER_VERSION` bumped in the first place? 

How I would expect it to work:
- If the major version differs, raise an error. Possibly still try to extract some info to show in the error message.
- If the minor version of the serialized data is higher, load all the data you are aware of and skip the rest
- If the minor version of the serialized data is lower, load all the available data and fill the missing data with defaults.

This assumes that when the minor version is bumped, a new field is appended to the serialized data, and nothing else is changed. This will also require keeping a "history" of added fields vs minor versions so that the deserialization of data from the past could be performed safely. Not sure if you want to go there.

vepkenez: The way I am planning on handling this sort of thing in the VersionedBytestrings PR is by using the BytestringKwargifier, which takes the class to deserialize as an input, we can return a different subclass for each version.

So, if Ursula has some changes, we just need to keep the old V1 Ursula accessible and then return that class for any Ursula bytestring that has V1 metadata.

Does this makes sense?

fjarri: How can we keep a whole other `Ursula` class accessible? What's the mechanism here?

vepkenez: Kind of like what's happening in this test:
https://github.com/nucypher/bytestringSplitter/blob/master/tests/test_kwargifiers.py#L127

fjarri: What I meant is that you'll have to have the V1 Ursula class defined somewhere. Which is huge, and has links to other parts of the codebase.

vepkenez: Well yeah... it depends on where the future takes us.  If the V1 can be the super class of the V2, then it's not so bad, we just change/add some methods.

If it's a total brand new distinct class, then it's less fun.

vepkenez: I just realized, something similar to this situation is already happening in the case of TreasureMaps vs SignedTreasureMaps.

Two distinct classes with different bytestring de/serializations, one a subclass of the other.

[here](https://github.com/nucypher/nucypher/blob/bc3cd71e3047a7b0b50b9a8d5a4c71c1a0d34107/nucypher/policy/collections.py)

nucypher/blockchain/eth/actors.py
cygnusv: ✅

nucypher/characters/lawful.py
```diff
@@ -1481,6 +1480,10 @@ def internal_splitter(cls, splittable, partial=False):
         result = splitter(splittable, partial=partial)
         return result
 
+    @classmethod
+    def is_compatible_version(cls, version: int) -> bool:
+        return cls.LOWEST_COMPATIBLE_VERSION <= version <= cls.LEARNER_VERSION
```
derekpierre, line 1485: 👍

nucypher/characters/lawful.py
```diff
@@ -1493,34 +1496,32 @@ def from_bytes(cls,
...
 
             try:
-                canonical_address, _ = BytestringSplitter(PUBLIC_ADDRESS_LENGTH)(payload, return_remainder=True)
+                canonical_address, _ = BytestringSplitter(ETH_ADDRESS_BYTE_LENGTH)(payload, return_remainder=True)
                 checksum_address = to_checksum_address(canonical_address)
                 nickname, _ = nickname_from_seed(checksum_address)
                 display_name = cls._display_name_template.format(cls.__name__, nickname, checksum_address)
                 message = cls.unknown_version_message.format(display_name, version, cls.LEARNER_VERSION)
+                if version > cls.LEARNER_VERSION:
+                    message += " Is there a new version of NuCypher?"
```
derekpierre, line 1513:
```suggestion
                    message += " Is there a newer version of NuCypher?"
```

nucypher/crypto/api.py
```diff
@@ -219,13 +223,14 @@ def generate_self_signed_certificate(*args, **kwargs):
...
+    """Return the checksum address written into a TLS certificates pseudonym field or raise an error."""
     try:
         pseudonym = certificate.subject.get_attributes_for_oid(NameOID.PSEUDONYM)[0]
     except IndexError:
-        raise RuntimeError("Invalid teacher certificate encountered: No checksum address present as pseudonym.")
+        raise InvalidNodeCertificate("Invalid teacher certificate encountered: No checksum address present as pseudonym.")
     checksum_address = pseudonym.value
     if not is_checksum_address(checksum_address):
-        raise RuntimeError("Invalid certificate checksum_address encountered")
+        raise InvalidNodeCertificate("Invalid certificate checksum_address encountered")
```
derekpierre, line 233:
```suggestion
        raise InvalidNodeCertificate("Invalid certificate checksum address encountered")
```

nucypher/network/nodes.py
```diff
@@ -818,12 +825,10 @@ def learn_from_teacher_node(self, eager=False, canceller=None):
...
 
-        if not set(self.learning_domains).intersection(set(current_teacher.serving_domains)):
-            teacher_domains = ",".join(current_teacher.serving_domains)
-            learner_domains = ",".join(self.learning_domains)
-            self.log.debug(
-                f"{current_teacher} is serving {teacher_domains}, but we are learning {learner_domains}")
-            return  # This node is not serving any of our domains.
+        if self.learning_domain != current_teacher.serving_domain:
+            self.log.debug(f"{current_teacher} is serving '{current_teacher.serving_domain}', "
+                           f"but we are learning '{self.learning_domain}'")
```
derekpierre, line 830:
```suggestion
                           f"ignore since we are learning about '{self.learning_domain}'")
```

nucypher/utilities/prometheus/collector.py
```diff
@@ -117,7 +117,7 @@ def _collect_internal(self) -> None:
         base_payload = {'app_version': nucypher.__version__,
                         'teacher_version': str(self.ursula.TEACHER_VERSION),
                         'host': str(self.ursula.rest_interface),
-                        'domains': str(', '.join(self.ursula.learning_domains)),
+                        'domain': str(self.ursula.learning_domain),
```
derekpierre, line 120: already a string no?

```suggestion
                        'domain': self.ursula.learning_domain,
```

tests/integration/learning/test_discovery_phases.py
```diff
@@ -189,7 +189,7 @@ def faster_receive_map(treasure_map_id, *args, **kwargs):
 
         # The number of nodes having the map is approximately the number you'd expect from full utilization of Alice's publication threadpool.
         # TODO: This line fails sometimes because the loop goes too fast.
-        assert len(nodes_that_have_the_map_when_we_unblock) == pytest.approx(policy.publishing_mutex._block_until_this_many_are_complete, .2)
+        # assert len(nodes_that_have_the_map_when_we_unblock) == pytest.approx(policy.publishing_mutex._block_until_this_many_are_complete, .2)
```
derekpierre, line 192: Do we want to do any follow-up here?

cygnusv: It's in issue #2016

derekpierre (approved):
🎸

vepkenez (approved):

PR #1887, opened 2020-04-14
Add list of node-as-a-service providers

### LINKED ISSUE #1886: Add opt-in list of staking-as-a-service providers to docs
* Make sure to note that these are not vetted providers or explicity endorsed by NuCypherCo
* Provide instructions to submit a PR to opt-in to the list

mswilkison:
Closes #1886 

Screen Shot 2020-04-14 at 2 53 53 PM
Screen Shot 2020-04-14 at 2 54 00 PM

### DIFF
--- a/docs/source/support/node_providers.rst
+++ b/docs/source/support/node_providers.rst
@@ -1,0 +1,19 @@
+.. _node-providers:
+
+===========================
+Node as a Service Providers
+===========================
+
+.. warning::
+  This is an opt-in list.
+  Providers have not been vetted or endorsed by the core development team.
+  Use your judgement in selecting a provider.
+
+.. note::
+  If you are a provider and wish to be included in this list,
+  please submit a pull request.
+
+The following staking pools and node infrastructure providers support the NuCypher Network (alphabetical order):
+
+* `Provider Name <https://example.com/>`_
+
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -181,6 +181,7 @@ Whitepapers
    :maxdepth: 1
    :caption: Support
 
+   support/node_providers
    support/community
    support/troubleshooting
    support/faq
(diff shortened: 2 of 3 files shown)


docs/source/support/node_providers.rst
```diff
@@ -0,0 +1,16 @@
...
+
+This is an opt-in list; providers have not been vetted or endorsed by the core development team. Use your judgement in selecting a provider.
+
+If you are a provider and wish to be included in this list, please submit a pull request.
+
+============= ===
+Provider Name URL
+============= ===
+Provider Name URL
+============= ===
```
cygnusv, line 15: Let's change this into bullets instead of a table. RST has this stupid requirement where all rows have to have the same width, including the separators (`=====`). This means that if suddenly we have a provider with a long name or URL, they will have to restructure the whole table in their submission.

mswilkison: ✅

derekpierre:
@mswilkison: we should link to this page from the following FAQ question: https://docs.nucypher.com/en/latest/support/faq.html#q-this-all-seems-too-complex-for-me-can-i-still-participate-in-some-way

docs/source/support/node_providers.rst
```diff
@@ -0,0 +1,17 @@
+===========================
+Node as a Service Providers
+===========================
+
+.. warning::
```
derekpierre, line 5: 👍

mswilkison: ✅

mswilkison:
> @mswilkison: we should link to this page from the following FAQ question: https://docs.nucypher.com/en/latest/support/faq.html#q-this-all-seems-too-complex-for-me-can-i-still-participate-in-some-way

✅

gaia:
I get this is dead simple stuff, but here are some repos I've come across that have accomplished this goal:

- https://github.com/terra-project/validator-profiles (same goal)
- https://github.com/trustwallet/assets (look into /blockchains. larger goal, but includes it)
- https://hackmd.io/DZyJU5HSThmGX8ER7tprXg#Baker-Registry-FAQ (tackles a different problem - on chain maintenance of validator information, beyond just exposing github content)

HTHs

mswilkison:
Good call @gaia! I think it will be easier to manage in a separate repo: https://github.com/nucypher/validator-profiles

Issue #1504, opened 2019-12-06
Follow-up failed Transactions with more details


KPrasch:

KPrasch:
Closed near 941043d98a87d38862752fa60a1fff14e0c7c060.  Still room for improvement here, particularly verifying that the requirements for contract transactions are met before execution.  For that we will use separate issues

Issue #1185, opened 2019-08-02
Better error message during _deploy_essential


jMyles:
If, while deploying contracts, there is insufficient gas, this is the error:

`ValueError: {'code': -32000, 'message': 'insufficient funds for gas * price + value'}`

It might make sense to add more details, such as how much as *is* available in the wallet and which contract failed to deploy.

PR #1891, opened 2020-04-16
Verify consistent ownership during deployment; Handle Insufficient funds TX failure.

### LINKED ISSUE #1777: PolicyManager and Adjudicator deployers should check that they're being deployed by owner of StakingEscrow

KPrasch:
Fixes #1777  #1185  

Advances #1504

### DIFF
--- a/nucypher/blockchain/eth/deployers.py
+++ b/nucypher/blockchain/eth/deployers.py
@@ -42,7 +42,7 @@ from nucypher.blockchain.eth.agents import (
     MultiSigAgent,
     ContractAgency
 )
-from nucypher.blockchain.eth.constants import DISPATCHER_CONTRACT_NAME, NULL_ADDRESS
+from nucypher.blockchain.eth.constants import DISPATCHER_CONTRACT_NAME, NULL_ADDRESS, STAKING_ESCROW_CONTRACT_NAME
 from nucypher.blockchain.eth.decorators import validate_checksum_address
 from nucypher.blockchain.eth.interfaces import (
     BlockchainDeployerInterface,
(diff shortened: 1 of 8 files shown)


nucypher/blockchain/eth/interfaces.py
```diff
@@ -108,6 +110,45 @@ class UnknownContract(InterfaceError):
...
+                     message: str,
+                     transaction_dict: dict,
+                     contract_function: Union[ContractFunction, ContractConstructor],
+                     *args):
+
+            self.base_message = message
+            self.name = get_transaction_name(contract_function=contract_function)
+            self.payload = transaction_dict
+            self.contract_function = contract_function
+            self.failures = {
```
derekpierre, line 131: Would this be better served as a class variable instead of being created per object?

nucypher/blockchain/eth/deployers.py
```diff
@@ -130,7 +130,8 @@ def is_deployed(self, contract_version: str = None) -> bool:
     def check_deployment_readiness(self,
                                    contract_version: str = None,
                                    ignore_deployed=False,
-                                   fail=True
+                                   fail=True,
+                                   additional_rules: tuple = None,
```
derekpierre, line 134: 👍 - nice!

nucypher/blockchain/eth/interfaces.py
derekpierre: ^ @KPrasch one outstanding suggestion from me.

KPrasch: Yeah - it's totally natural and I agree - but I cannot get the other signer classes in scope without breaking this module into a package. and juggling circular imports :-(  

Perhaps give it a shot? :-)

PR #1926, opened 2020-04-23
Extricate Documentation Requirements; Reconfigure RTD

KPrasch:


derekpierre:
@KPrasch Question: how would we update the `docs/requirements.txt` file later on?

KPrasch:
>Would we need a separate Pipfile and Pipfile.lock in the docs directory?

More pipfiles?!? I hope not!  :-) I'm trying to manage these requirements differences with pip extras.
 
I created that first set by creating a fresh environment, then installing packages one at a time until there was the dependencies were met - then exported `docs-requirements.txt` with `pip freeze > docs-requirements.txt`.

Since the docs specific decencies do not change with the same frequency as others, this is good enough IMO.

derekpierre:
Haha definitely don't want more Pipfiles 🤢- just wondering since we only have a `requirements.txt` file for docs. The frozen versions work for now - assuming we don't need to do any updating to make https://github.com/nucypher/nucypher/issues/1928 work...TBD.

PR #1927, opened 2020-04-23
Deprecate aafigure

KPrasch:
the aafigure package is abandoned and incompatible with modern Sphinx.  Perhaps we can restore this later, or pre-render images before including them in docs.   For now - Let's just remove it for less ops overhead.


derekpierre:
Sub-stakes graphs will not look as clean as before, because of the `+` signs in the x-axis, but I could live with that for now 😄 .

Screen Shot 2020-04-23 at 4 30 03 PM

PR #1934, opened 2020-04-28
Adapt `nucypher-deploy` to Clef+Infura. New IBEX testnet 🐐

cygnusv:
* Adapt `nucypher-deploy` to Clef+Infura
* New documentation on how to use Clef with automatic rules
* Allows Infura to be used on other Ethereum networks
* Generalizes StakingWallet as Wallet, which is now a generic account management abstraction
* Changes to `nucypher-deploy` were tested live with the deployment of a new internal testnet on Rinkeby network


nucypher/blockchain/eth/providers.py
```diff
@@ -34,16 +49,25 @@ def _get_websocket_provider(provider_uri):
...
     uri_breakdown = urlparse(provider_uri)
     infura_envvar = 'WEB3_INFURA_PROJECT_ID'
     os.environ[infura_envvar] = os.environ.get(infura_envvar, uri_breakdown.netloc)
 
     try:
-        # TODO: Only testnet for now - #1496
-        from web3.auto.infura.goerli import w3
+        # TODO: Is this the right approach? Looks a little bit shabby... Also #1496
+        if "mainnet.infura.io" in provider_uri:
+            from web3.auto.infura.mainnet import w3
```
KPrasch, line 62: Nit picking here - but perhaps make this into a dict.

nucypher/blockchain/eth/signers.py
```diff
@@ -188,8 +193,15 @@ def sign_transaction(self, transaction_dict: dict) -> HexBytes:
             'chainId': Web3.toHex,
             'from': to_checksum_address
         }
-        transaction_dict = apply_formatters_to_dict(formatters, transaction_dict)
-        signed = self.__ipc_request("account_signTransaction", transaction_dict)
+
+        # Workaround for contract creation TXs
```
KPrasch, line 197: This pattern appears several times around the codebase - Do you think this transaction dict mutation is appropriately handled at this abstraction layer or a higher one?

nucypher/cli/actions.py
```diff
@@ -405,42 +406,38 @@ def select_stake(stakeholder, emitter, divisible: bool = False, staker_address:
...
-
-    if not provider_uri:
-        provider_uri = wallet.blockchain.provider_uri
 
-    # Lazy connect the blockchain interface
-    if not BlockchainInterfaceFactory.is_interface_initialized(provider_uri=provider_uri):
-        BlockchainInterfaceFactory.initialize_interface(provider_uri=provider_uri, poa=poa, emitter=emitter)
-    blockchain = BlockchainInterfaceFactory.get_interface(provider_uri=provider_uri)
+    # We use Wallet internally as an account management abstraction
+    if not wallet:
```
KPrasch, line 425: :+1: - This is going in a good direction.

nucypher/cli/actions.py
```diff
@@ -450,8 +447,10 @@ def select_client_account(emitter,
     headers = ['Account']
     if show_staking:
         headers.append('Staking')
-    if show_balances:
-        headers.extend(('', ''))
```
KPrasch, line 454: How was this ever ('', '') anyways.... :monocle_face:

nucypher/cli/actions.py
```diff
@@ -405,42 +406,38 @@ def select_stake(stakeholder, emitter, divisible: bool = False, staker_address:
...
-        BlockchainInterfaceFactory.initialize_interface(provider_uri=provider_uri, poa=poa, emitter=emitter)
-    blockchain = BlockchainInterfaceFactory.get_interface(provider_uri=provider_uri)
+    # We use Wallet internally as an account management abstraction
+    if not wallet:
+        if not provider_uri and not signer_uri:
+            raise ValueError("At least a provider URI or signer URI is necessary to select an account")
+        # Lazy connect the blockchain interface
+        if not BlockchainInterfaceFactory.is_interface_initialized(provider_uri=provider_uri):
+            BlockchainInterfaceFactory.initialize_interface(provider_uri=provider_uri, poa=poa, emitter=emitter)
+        wallet = Wallet(provider_uri=provider_uri, signer=Signer.from_signer_uri(signer_uri))
```
derekpierre, line 431: What happens here if signer_uri is `None`, since we are calling `Signer.from_signer_uri(signer_uri)`? At this point `provider_uri` or `signer_uri` or both are set.

cygnusv: Excellent eye 👁️! I was literally fixing this now, since tests were failing for this reason. I'm adding a commit right now with the fix.

nucypher/blockchain/eth/actors.py
```diff
@@ -1493,76 +1493,75 @@ def was_this_evidence_evaluated(self, evidence) -> dict:
...
+    def __contains__(self, checksum_address: str) -> bool:
+        return bool(checksum_address in self.accounts)
 
-        def __init__(self,
-                     registry: BaseContractRegistry,
-                     client_addresses: set = None,
-                     signer=None):
+    @property
+    def active_account(self) -> str:
+        return self.blockchain.transacting_power.account
```
KPrasch, line 1525: Perhaps the codebase has finally gotten to a place where we can remove the notion of an "active account".  Is it fulfilling an important function from your view, @cygnusv ?

nucypher/blockchain/eth/actors.py
```diff
@@ -1493,76 +1493,75 @@ def was_this_evidence_evaluated(self, evidence) -> dict:
...
 
-            self.__local_accounts = dict()
-            self.__client_accounts = set()  # Note: Account index is meaningless here
-            self.__transacting_powers = dict()
+    def __get_accounts(self) -> None:
+        if self.__signer:
+            signer_accounts = self.__signer.accounts
+            self.__client_accounts.extend([a for a in signer_accounts if a not in self.__client_accounts])
```
KPrasch, line 1530: This filtration feels a bit clunky to me - is it possible to use a dict/set to track unique addressable accounts here?

cygnusv: It's clunky, but sets are definitely not an option here. Sets have an undefined order, which changes with every execution. See this commit of this very same PR (b18b9d3a2d843751280cd24f49f4b63057d259bd). It took me a lot to debug the problems caused by sets.

nucypher/blockchain/eth/providers.py
KPrasch: Maybe this is too weird, and *not* an actual RFC :thread: 

``` python
from importlib import import_module

import_module(f'web3.auto.infura.{network}.web3')
```

nucypher/blockchain/eth/actors.py
KPrasch: Despite your aversion to sets - How about a dictionary?

cygnusv: It's not aversion – I love sets; they're simply not the right collection here as we want to preserve order of insertion. Yes, we can use dicts, I thought about that too, but to be honest, I also find that weird: we'd be using a strange feature of dicts which only exists from python 3.7 and we wouldn't be using dict values. Current solution is clunky but it's declaratively precise.

KPrasch: Nothing strange about it - It's actually considered an implementation detail as of python 3.6.0 (and earlier), but can be completely considered to be ordered for implementation as of python 3.7.

See this StackOverflow Question: <https://stackoverflow.com/questions/39980323/are-dictionaries-ordered-in-python-3-6>

KPrasch: Aside from the dictionary implementation, IMO if you need a unique insertion-ordered collection and are forced to "manually" use filtration methods, this may indicate code inefficiency and make for some flaky features.

cygnusv: Let's add a TODO here. I didn't use a dictionary on purpose since I didn't find it elegant at that time, given that there was no value to store. However, I had an issue in my head with the behavior of Wallet, and perhaps in this case the values of the dict could be of use. Basically, we need to discern between accounts from the signer and accounts from a non-signing provider (e.g. Infura), as the latter won't be able to transact. This will simply improve error-handling, so it's not urgent.

KPrasch: :+1: Sounds good!

PR #1935, opened 2020-04-28
Canceled

zuyyina-eng:


mswilkison:
Hi @zuyyina-eng if this was meant to be a validator profile it should go in this repo: https://github.com/nucypher/validator-profiles

Issue #305, opened 2018-06-14
Ways to implement NuCypher slashing protocol


michwill:
When Bob asks Ursula to do a re-encryption for him, there are two ways of misehavior possible:

* Ursula simply doesn't re-encrypt;
* Ursula produces garbage in response to re-encryption request.

It's not clear if we can do the latter on-chain, but we need some sort of agreement between nodes if we want to prove the former.

Here are 2 ways possible to solve this.

1. Challenge protocol w/o any side-chain
------------------------------------------

Let's introduce a "generalized ping". Instead of a real ping, it's a challenge with a response respected from Ursula.
If no response comes, the ping was "unresponded".

In brief, we have nodes `ursulas[0..N-1]`, and many nodes ping one node.
The idea is that the target Ursula cannot distinguish if she's being challenged, or it is a legitimate request which she has to do.
Also she cannot see if any two pings come from the same node or not.

With these assumptions, here's the protocol:

It would be too expensive (in terms of gas) for every Ursula to produce evidence about every other Ursula, so we have to definie which Ursula votes on which.
So, we randomly select who votes on who in the following fashion.
We don't necessarily want other Ursulas to know who "pings" them.
At the beginning of the period, hash of the block when the period is started on Ethereum blockchain is `h`.
Then we have::

    r = hash(sign(eth_pubkey, h))
    ursulas_to_challenge = [
        closest_hamming_distance(ursulas, hash(r + i))
        for i in range(N_to_challenge)]

Importantly, a smart contract can test whether the pseudorandom number `r` indeed belongs to this Ursula by using precompiled `ecrecover` method.

So, now each Ursula has `N_to_challenge` Ursulas to ping during this period.
They have to ping them multiple times (let say, 100) during the current period randomly, while keeping these requests anonymous (so they shouldn't have same IP
addresses).
After having done that, each Ursula has statistics of how much time the challened Ursula was up.
After that, Ursulas vote (using hash-and-reveal protocol) on which Ursula was up for how much time.

If measurement by one of challenging Ursulas is different from the other swarm significantly enough (so that it's not a statistical artifact), this Ursula gets
slashed. So all Ursulas are incentivized to honestly measure uptime of each Ursula with a precision within the specified margins.

In our case, generalized ping is producing a random capsule to re-encrypt and ask to re-encrypt it. It doesn't have to be a decryptable capsule produced by
Alice: Ursula can "re-encrypt" and produce a proof about any capsule.

[...]

PR #1951, opened 2020-05-04
Proposal: set slashing coefficients to almost zero

### LINKED ISSUE #1916: Slashing penalty
We need to finalize the penalty calculations for a false re-encrytion.

vzotova:
Fixes #1916

### DIFF
--- a/nucypher/blockchain/economics.py
+++ b/nucypher/blockchain/economics.py
@@ -66,9 +66,9 @@ class BaseEconomics:
 
     # Adjudicator
     _default_hash_algorithm = HASH_ALGORITHM_SHA256
-    _default_base_penalty = 100
-    _default_penalty_history_coefficient = 10
-    _default_percentage_penalty_coefficient = 8
+    _default_base_penalty = 2
+    _default_penalty_history_coefficient = 0
+    _default_percentage_penalty_coefficient = 100000  # 0.001%
     _default_reward_coefficient = 2
 
     # Worklock


mswilkison:
Is detaching on slash already implemented?

vzotova:
> Is detaching on slash already implemented?

No and this PR is excuse not to do it. I'm still thinking that spending time on slashing mechanics before we are understand how it should work properly is wasting time

nucypher/blockchain/economics.py
```diff
@@ -66,9 +66,9 @@ class BaseEconomics:
 
     # Adjudicator
     _default_hash_algorithm = HASH_ALGORITHM_SHA256
-    _default_base_penalty = 100
-    _default_penalty_history_coefficient = 10
-    _default_percentage_penalty_coefficient = 8
+    _default_base_penalty = 2
+    _default_penalty_history_coefficient = 0
+    _default_percentage_penalty_coefficient = 100000  # 0.001%
```
michwill, line 71: Does this mean slashing 0.001% if a misbehavior is detected?

michwill: If so, I'd rather prefer something like 10 ** 18 here.

The reason is: an attacker who gained the access shouldn't be able to profit draining the pool (e.g. gas costs should outweigh his gains). With 0.001% it's not the case (if I understand correctly) if the stake is >~ $100k

nucypher/blockchain/economics.py
```diff
@@ -66,9 +66,9 @@ class BaseEconomics:
 
     # Adjudicator
     _default_hash_algorithm = HASH_ALGORITHM_SHA256
-    _default_base_penalty = 100
-    _default_penalty_history_coefficient = 10
-    _default_percentage_penalty_coefficient = 8
+    _default_base_penalty = 2
+    _default_penalty_history_coefficient = 0
```
arjunhassard, line 70: I think this should remain zero until there is a compelling behavioural and/or threat-based rationale to ramp penalties. Hypothetically, if a single incorrect re-encryption could occur innocently, but five from the same Ursula is no accident (and thus indicates an attack), then it makes sense to considerably weaken that staker with increasing slashes. But in the absence of a good reason, and with community resistance, this component will do more harm than good.

mswilkison: I think the argument for low but non-zero numbers is that it gives us a relatively low-risk way to evaluate the slashing logic in production.

arjunhassard: That argument is completely valid. But 'evaluating' the slashing mechanism involves observing real-world behaviour, not only testing the code. If in production, a group of stakers start repeatedly producing incorrect re-encryptions, then that certainly provides a compelling reason to increase the punishment and/or introduce ramping. As it stands, we can't concretely articulate the benefit of producing incorrect re-encryptions to a staker, nor the risk to the network. This makes it difficult to parametrize the difference in punishment between one incorrect re-encryption and five. However, this may change once the network is live!

nucypher/blockchain/economics.py
vzotova: penalty will be allways 2 * 10^-18, I changed `_default_percentage_penalty_coefficient` for just in case but with those values it'll be unused

mswilkison:
Another proposal: 1% slashing per violation (limited to 1 slash per period so if someone messes up a bunch they don't get wiped out).

1% is enough that people will want to avoid it but it's not a catastrophe if it happens.

vzotova:
> Another proposal: 1% slashing per violation (limited to 1 slash per period so if someone messes up a bunch they don't get wiped out).
> 
> 1% is enough that people will want to avoid it but it's not a catastrophe if it happens.

limitation one slash per period is also changes in code and it's what I'm trying to avoid here. But again if you're sure and this will not be temporary solution - I'll do that

arjunhassard:
1% seems like a good balance. The limit of 1 slash per period grants stakers the opportunity to fix the error, and addresses the unfairness of highly variable request cadences. Some details to iron out: 

- Say a staker re-encrypts incorrectly 10x times within a 24h period, then they shall lose 1% of their stake in total, or 10% spaced out over 10 periods? The former is more in the spirit of this rule, but does imply that 1 mistake and 1,000 mistakes could result in the same punishment. This isn't necessarily a problem. 
- A persistently disruptive staker – who does not fix the problem or take their workers down, and is pinged daily for re-encryptions – will still have quite a lot of their original stake after many months (see graph). Choosing a non-zero base penalty carefully will address this – remember that they'll get a smaller share of new work each day, but still serve the same number of policies that predate the first offence. 

image

KPrasch:
I want to emphasize that the proposition of this PR is to introduce precisely the minimum about of penalty to measure for continued Iteration.

arjunhassard:
@KPrasch I agree with the objective you mentioned - I would say that even the simplest slashing mechanism has design choices – e.g. should the offender's stake never reach zero, and if not, what should the base penalty be?

We can move the iteration towards a more sophisticated version to an issue, although #1916 is closed by this PR, so probably better to discuss things like detaching in nucypher/protocol/issues/12? Wdyt?

@vzotova Personally I think limiting slashes per period is unavoidable. It's way too risky for stakers with the prospect of attackers batching transactions etc. The best way to avoid implementing changes over and over is for us to think through all the details asap. We can start by agreeing that regardless of how many incorrect re-encryptions are committed, you will only ever see 1% + base penalty deleted from your stake per day. Yay or nay?

vzotova:
we have #305 to discuss this
Current numbers almost zero, so even all blocks in a day for whole ethereum will contain slashing txs then in the result slashing value will be ~10^-12 NU (5760 blocks per day, 10mlns gas each, 500k gas each tx)

PR #1953, opened 2020-05-04
Fix for Autodoc build on ReadTheDocs

derekpierre:
Fixes #1928

PR #1958, opened 2020-05-06
`nucypher` module must be present to build APIs docs section

derekpierre:
Ensure that `nucypher` module is installed when building docs on RTD.

Follow-up to #1953 .


KPrasch:
merge-at-will - Tests staged changes on RTD.

PR #1989, opened 2020-05-14
Advancing Mock Tools; > 90% Test coverage for CLI Actions

KPrasch:
Co-Author: @vzotova 

- Majority of the diff is in `tests` additions, aside from agent method decoration
- Introduces a complete automatic agent/contract mocking API integration with pytest
- Sets test coverage for CLI actions > 90%
- Numerous small preventative bug fixes
- Functional test suite runs in ~2 seconds (`tests.cli.functional`)

#### New Entities

- `MockContractAgent`
- `MockContractAgency`
- `ContractInterfaces(Enum)`
- `@contract_api` decorator

#### New Type Aliases (Thanks @fjarri!)

```python
Agent = TypeVar('Agent', bound='EthereumContractAgent')

```

```python
ContractInterfaces = Union[
    CONTRACT_CALL,
    TRANSACTION,
    CONTRACT_ATTRIBUTE,
    UNKNOWN_CONTRACT_INTERFACE
]
```

#### Fixtures (Autouse)

- `mock_contract_registry`
- `mock_token_agent`
- `mock_staking_agent`
- `mock_staking_agent`
- `mock_policy_manager_agent`
- `mock_adjudicator_agent`


tests/cli/functional/conftest.py
```diff
@@ -16,20 +16,107 @@
...
+    # Customize the mock agent
+    spec = {'boosting_refund': economics.worklock_boosting_refund_rate,
+            'slowing_refund': 1,  # TODO: another way to get this value?
+            'start_bidding_date': economics.bidding_start_date,
+            'end_bidding_date': economics.bidding_end_date,
+            'end_cancellation_date': economics.cancellation_end_date,
+            'minimum_allowed_bid': economics.worklock_min_allowed_bid,
+            'lot_value': economics.worklock_supply}
+    for method_name, return_value in spec.items():
+        setattr(mock_agent.__class__, method_name, return_value)
```
vzotova, line 107: can we remove strings and use `mock_agent.boosting_refund.return_value = economics.worklock_boosting_refund_rate` or equivalent from `PropertyMock`?

tests/mock/agents.py
```diff
@@ -1,221 +1,153 @@
...
+    def get_unexpected_transactions(self, allowed: Union[Iterable[Callable], None]) -> List[Callable]:
+        if allowed:
+            predicate = lambda tx: tx not in allowed and tx.called
+        else:
+            predicate = lambda tx: tx.called
+        unexpected_transactions = list(filter(predicate, self.all_transactions))
+        return unexpected_transactions
 
-        super().__init__()
+    def assert_no_unexpected_transactions(self, allowed: Iterable[Callable]) -> None:
```
vzotova, line 115: double negation in name reads difficult, at least for me

nucypher/blockchain/eth/agents.py
```diff
@@ -724,18 +741,22 @@ def get_missing_commitments(self, checksum_address: str) -> int:
         return result
 
     @property
+    @contract_api(ContractInterfaces.CALL)
     def is_test_contract(self):
         return self.contract.functions.isTestContract().call()
 
     @property
+    @contract_api(ContractInterfaces.CALL)
```
vzotova, line 749: Maybe we should use here `ATTRIBUTE` instead of call? (because it's property)

tests/mock/agents.py
```diff
@@ -1,221 +1,153 @@
...
-    TRANSACTIONS = tuple()
+            # Get default effect
+            interface = getattr(agent_method, cls.__COLLECTION_MARKER)
+            default_return = cls.__DEFAULTS.get(interface)
 
-    # Spy
-    _SPY_TRANSACTIONS = defaultdict(list)
-    _SPY_CALLS = defaultdict(list)
+            # Setup Mock - Carry over the decorator marker to the mock
+            mock = Mock(return_value=default_return)
```
vzotova, line 60: I think some of them should be `PropertyMock`

nucypher/blockchain/eth/actors.py
```diff
@@ -858,10 +862,35 @@ def locked_tokens(self, periods: int = 0) -> NU:
 
     @property
     def current_stake(self) -> NU:
+        """The total number of staked tokens, i.e., tokens locked in the current period."""
+        return self.locked_tokens(periods=0)
+
+    @property
+    def active_stakes(self) -> Iterable[Stake]:
+        """Returns a list of active stakes for this staker."""
```
derekpierre, line 870: Since not actually returning a List:

```suggestion
        """Returns active stakes for this staker."""
```

nucypher/blockchain/eth/actors.py
```diff
@@ -858,10 +862,35 @@ def locked_tokens(self, periods: int = 0) -> NU:
...
+
+    def __filter_divisible_stakes(self, stake: Stake) -> bool:
         """
-        The total number of staked tokens, i.e., tokens locked in the current period.
+        Helper function for use as a filtration predicate when determining
+        what constitutes a divisible stake amongst a given iterable of stakes.
         """
-        return self.locked_tokens(periods=0)
+        min_stake_value = self.economics.minimum_allowed_locked * 2
+        result = bool(stake.value >= min_stake_value)
```
derekpierre, line 886: Minor nitpick:

```suggestion
        min_divisible_stake_value = self.economics.minimum_allowed_locked * 2
        result = bool(stake.value >= min_divisible_stake_value)
```

nucypher/cli/actions/confirm.py
```diff
@@ -19,10 +19,12 @@
 import click
 from constant_sorrow.constants import UNKNOWN_DEVELOPMENT_CHAIN_ID
 
+from nucypher.blockchain.eth.interfaces import BlockchainDeployerInterface
 from nucypher.blockchain.eth.token import NU
+from nucypher.characters.control.emitters import StdoutEmitter
 from nucypher.cli.literature import (
     ABORT_DEPLOYMENT,
-    CONFIRM_ENABLE_RESTAKING,
+    CHARACTER_DESTRUCTION, CONFIRM_ENABLE_RESTAKING,
```
derekpierre, line 27:
```suggestion
    CHARACTER_DESTRUCTION, 
    CONFIRM_ENABLE_RESTAKING,
```

nucypher/cli/actions/confirm.py
```diff
@@ -32,45 +34,77 @@
...
+from nucypher.config.node import CharacterConfiguration
 
 
-def confirm_deployment(emitter, deployer_interface) -> bool:
+def confirm_deployment(emitter: StdoutEmitter, deployer_interface: BlockchainDeployerInterface) -> bool:
+    """
+    Interactively confirm deployment by asking the user to type the ALL CAPS name of
+    the network they are deploying to or 'DEPLOY' if the network if not a known public chain.
+
+    Aborts if the confirmation work is incorrect.
```
derekpierre, line 45:
```suggestion
    Aborts if the confirmation fails.
```

nucypher/cli/actions/confirm.py
```diff
@@ -32,45 +34,77 @@
...
+    click.confirm(CONFIRM_STAGED_STAKE.format(nunits=str(value.to_nunits()),
+                                              tokens=str(value.to_tokens()),
                                               staker_address=staker_address,
                                               lock_periods=lock_periods), abort=True)
+    return True
 
 
-def confirm_large_stake(value: NU = None, lock_periods: int = None) -> None:
+def confirm_large_stake(value: NU = None, lock_periods: int = None) -> bool:
+    """Interactively confirm a large or a long stake"""
```
derekpierre, line 88:
```suggestion
    """Interactively confirm a large stake and/or a long stake duration"""
```

nucypher/cli/actions/select.py
```diff
@@ -88,24 +99,38 @@ def select_client_account(emitter,
                           poa: bool = None
                           ) -> str:
     """
+    Interactively select an ethereum wallet account from a table of nucypher account metadata.
+
     Note: Showing ETH and/or NU balances, causes an eager blockchain connection.
     """
 
+    if wallet and (provider_uri or signer_uri or signer):
+        raise ValueError("If pass a wallet, don't pass a signer, provider URI, or signer URI also.")
```
derekpierre, line 108:
```suggestion
        raise ValueError("If a wallet is provided, don't provide a signer, provider URI, or signer URI.")
```

nucypher/cli/actions/select.py
```diff
@@ -196,10 +222,27 @@ def select_network(emitter: StdoutEmitter) -> str:
...
+    Behaviour
+    ~~~~~~~~~
+
+    - If checksum address is supplied by parameter or worker address env var - confirm there is a corresponding
+      file on the disk or raise ValueError.
+
+    - If there is only one configuration file for the character, automatically return its filepath.
+
+    - If there are multiple character configurations on the disk in the same configuration root,
+      interactively selection commences.
```
derekpierre, line 241:
```suggestion
      use interactive selection.
```

nucypher/cli/literature.py
```diff
@@ -55,7 +55,7 @@
 * Ursula Node Operator Notice *
 -------------------------------
 
-By agreeing to stake {str(value)} ({str(value.to_nunits())} NuNits):
+By agreeing to stake {tokens} ({nunits} NuNits):
```
derekpierre, line 58: I think I may have preferred the previous use of `value`. I like having the one value passed, and the conversion to NuNits being automatically performed. I'm interested in your reasoning for the change...?

nucypher/cli/painting/staking.py
```diff
@@ -22,13 +22,18 @@
 from nucypher.blockchain.eth.constants import STAKING_ESCROW_CONTRACT_NAME
 from nucypher.blockchain.eth.token import NU
 from nucypher.blockchain.eth.utils import datetime_at_period, prettify_eth_amount
-from nucypher.cli.literature import NO_ACTIVE_STAKES, NO_STAKES_AT_ALL, NO_STAKING_ACCOUNTS, POST_STAKING_ADVICE
+from nucypher.characters.control.emitters import StdoutEmitter
+from nucypher.cli.literature import POST_STAKING_ADVICE
 from nucypher.cli.painting.transactions import paint_receipt_summary
 
+STAKE_TABLE_COLUMNS = ('Idx', 'Value', 'Remaining', 'Enactment', 'Termination')
+STAKER_TABLE_COLUMNS = ('Status', 'Restaking', 'Winding Down', 'Unclaimed Fees', 'Min fee rate')
```
derekpierre, line 30: Should we call these `_HEADERS` instead of `_COLUMNS`?

tests/cli/functional/actions/test_select_client_account.py
```diff
@@ -0,0 +1,191 @@
...
+
+    error_message = "Pass either signer or signer_uri but not both."
+    with pytest.raises(ValueError, match=error_message):
+        select_client_account(emitter=test_emitter, signer=Mock(), signer_uri=MOCK_SIGNER_URI)
+
+    #
+    # Explicit wallet
+    #
+
+    error_message = "If pass a wallet, don't pass a signer, provider URI, or signer URI also."
```
derekpierre, line 59: Based on my previous suggestion:

```suggestion
    error_message = "If a wallet is provided, don't provide a signer, provider URI, or signer URI."
```

tests/cli/functional/actions/test_select_client_account_for_staking.py
```diff
@@ -0,0 +1,22 @@
+import pytest
+
+from nucypher.blockchain.eth.actors import StakeHolder
+from nucypher.blockchain.eth.registry import IndividualAllocationRegistry
+from nucypher.cli.actions.select import select_client_account_for_staking
+from nucypher.config.constants import TEMPORARY_DOMAIN
+
+
+@pytest.mark.skip(reason="David, please send help.")
```
derekpierre, line 9: 😂

tests/cli/functional/actions/test_select_config_file.py
```diff
@@ -0,0 +1,115 @@
...
+                                config_root=tmpdir)
+
+    output = stdout_trap.getvalue()
+    for filename, account in accounts:
+        assert account.address in output
+
+    table_data = output.split('\n')
+    table_addresses = [row.split()[1] for row in table_data[2:-2]]
+
+    # TODO: Finish this test
```
derekpierre, line 107: ???

tests/cli/functional/actions/test_select_network.py
```diff
@@ -0,0 +1,18 @@
+import pytest
+
+from nucypher.blockchain.eth.networks import NetworksInventory
+from nucypher.cli.actions.select import select_network
+
+
+__NETWORKS = NetworksInventory.NETWORKS
+
+
+@pytest.mark.parametrize('user_input', range(0, len(__NETWORKS)-1))
```
derekpierre, line 10: 👍

tests/cli/functional/test_ursula_local_keystore_cli_functionality.py
```diff
@@ -16,58 +16,26 @@
...
 from nucypher.blockchain.eth.signers import KeystoreSigner
 from nucypher.blockchain.eth.token import StakeList
 from nucypher.cli.main import nucypher_cli
 from nucypher.config.characters import UrsulaConfiguration
-from nucypher.config.constants import NUCYPHER_ENVVAR_KEYRING_PASSWORD, NUCYPHER_ENVVAR_WORKER_ETH_PASSWORD, \
-    TEMPORARY_DOMAIN
+from nucypher.config.constants import TEMPORARY_DOMAIN
 from tests.constants import (
-    INSECURE_DEVELOPMENT_PASSWORD,
+    CLI_ENV, INSECURE_DEVELOPMENT_PASSWORD,
```
derekpierre, line 28:
```suggestion
    CLI_ENV, 
    INSECURE_DEVELOPMENT_PASSWORD,
```

tests/cli/functional/test_worklock_cli_functionality.py
```diff
@@ -364,16 +353,17 @@ def test_participant_status(click_runner,
...
+    expected_calls = (mock_worklock_agent.check_claim,
+                      mock_worklock_agent.eth_to_tokens,
+                      mock_worklock_agent.get_deposited_eth,
+                      mock_worklock_agent.get_eth_supply,
+                      mock_worklock_agent.get_base_deposit_rate,
+                      mock_worklock_agent.get_bonus_lot_value,
+                      mock_worklock_agent.get_bonus_deposit_rate,
+                      mock_worklock_agent.get_bonus_refund_rate,
+                      mock_worklock_agent.get_base_refund_rate,
+                      # 'get_completed_work',  # TODO Yes or no?
```
derekpierre, line 365: ??? - what's going on here? 👀

nucypher/utilities/networking.py
```diff
@@ -0,0 +1,57 @@
...
+            raise
+    else:
+        # Interactive
+        if not force:
+            if not click.confirm(CONFIRM_URSULA_IPV4_ADDRESS.format(rest_host=rest_host)):
+                rest_host = click.prompt(COLLECT_URSULA_IPV4_ADDRESS, type=IPV4_ADDRESS)
+        else:
+            emitter.message(FORCE_DETECT_URSULA_IP_WARNING.format(rest_host=rest_host), color='yellow')
+
+        return rest_host
```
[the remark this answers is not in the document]
derekpierre: Haha soon this will be a thing of the past 😈 - #2001

tests/cli/functional/actions/test_select_client_account.py
```diff
@@ -0,0 +1,191 @@
...
+
+    # From uninitialized Provider
+    mocker.patch.object(BlockchainInterfaceFactory, 'is_interface_initialized', return_value=False)
+    mocker.patch.object(BlockchainInterfaceFactory, '_interfaces', return_value={})
+    mocker.patch.object(BlockchainInterfaceFactory, 'get_interface', return_value=mock_testerchain)
+    selected_account = select_client_account(emitter=test_emitter, provider_uri=MOCK_PROVIDER_URI)
+    assert selected_account == expected_account
+
+
+@pytest.mark.parametrize('selection,show_staking,show_eth,show_tokens,stake_info', (
```
derekpierre, line 114: 🤠

tests/cli/ursula/test_local_keystore_integration.py
```diff
@@ -26,18 +28,23 @@
 from nucypher.blockchain.eth.token import StakeList
 from nucypher.cli.main import nucypher_cli
 from nucypher.config.characters import UrsulaConfiguration
-from tests.cli.functional.test_ursula_local_keystore_cli_functionality import (CLI_ENV, KEYFILE_NAME_TEMPLATE,
-                                                                               MOCK_KEYSTORE_PATH, MOCK_SIGNER_URI,
-                                                                               NUMBER_OF_MOCK_ACCOUNTS)
-from tests.constants import (INSECURE_DEVELOPMENT_PASSWORD, MOCK_IP_ADDRESS, TEST_PROVIDER_URI)
+from tests.constants import (
+    CLI_ENV, INSECURE_DEVELOPMENT_PASSWORD,
```
derekpierre, line 32:
```suggestion
    CLI_ENV, 
    INSECURE_DEVELOPMENT_PASSWORD,
```

tests/mock/agents.py
KPrasch: Yeah naming is tricky here.  I want to assert that there are no unintentional side effects in the form of a transaction. ...Suggestions?

tests/mock/agents.py
KPrasch: Let me take a closer look. Two thoughts here:

1. We are overriding the name namespce of the property with a `Mock` explicitly so I am surprised we need a PropertyMock at all.

2.  We need a clear convention for when to use a property to expose a contract interface: Perhaps we use properties for *all contract attributes?*

tests/cli/functional/test_worklock_cli_functionality.py
KPrasch: Erhm, Needs a close eye to determine if this is an expected call or not - I was seeing some flaky test behavior with this line uncommented.

tests/cli/functional/actions/test_select_client_account.py
KPrasch: No more Mr. NIce Guy!

nucypher/utilities/seednodes.py
```diff
@@ -135,33 +126,4 @@ def load_seednodes(emitter,
...
-    else:
-        # Interactive
-        if not force:
-            if not click.confirm(CONFIRM_URSULA_IPV4_ADDRESS.format(rest_host=rest_host)):
-                rest_host = click.prompt(COLLECT_URSULA_IPV4_ADDRESS, type=IPV4_ADDRESS)
-        else:
-            emitter.message(FORCE_DETECT_URSULA_IP_WARNING.format(rest_host=rest_host), color='yellow')
-
-        return rest_host
+    return teacher_nodes
```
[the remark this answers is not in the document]
KPrasch: Hopefully this will be picked up by a pre-commit hook with a rebase since #2001 is merged.

tests/cli/functional/conftest.py
KPrasch: Yes.  Great feedback here.  We want to eliminate the bulk of mock strings for better maintainability.

nucypher/blockchain/eth/actors.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/actors.py

nucypher/cli/actions/confirm.py
KPrasch: :heavy_check_mark:

nucypher/cli/actions/confirm.py
KPrasch: :heavy_check_mark:

nucypher/cli/actions/confirm.py
KPrasch: :heavy_check_mark:

nucypher/cli/actions/select.py
KPrasch: :heavy_check_mark:

nucypher/cli/actions/select.py
KPrasch: :heavy_check_mark:

nucypher/cli/literature.py
KPrasch: @derekpierre - Raising a good question here... We need a solid guideline for inclusion in `literature.py` (see #2010).  

Primary motivator here for me is to keep python code from appearing in strings and to keep them as pure string templates. 

I think leaving the string-expressed-logic is actually a potential vector for future regression.

nucypher/cli/painting/staking.py
KPrasch: Personally, I find column to be a more intuitive concept here.

tests/cli/functional/actions/test_select_client_account.py
KPrasch: :heavy_check_mark:

tests/cli/functional/actions/test_select_config_file.py
KPrasch: I hear ya - When is this test done?  

Perhaps I can make a follow up Issue.  Unsure how granular we need to be around the GUI test at this time.

tests/cli/functional/conftest.py
KPrasch: :heavy_check_mark: - Let me know if this is enough.

tests/cli/functional/test_ursula_local_keystore_cli_functionality.py
KPrasch: :heavy_check_mark:

tests/cli/ursula/test_local_keystore_integration.py
KPrasch: :heavy_check_mark:

tests/config/test_character_configuration.py
```diff
@@ -146,3 +153,42 @@ def test_ursula_development_configuration(federated_only=True):
         ursula = config()
         assert ursula not in ursulas
         ursulas.append(ursula)
+
```
KPrasch, line 156: Needs Follow up Issue

nucypher/cli/literature.py
derekpierre: Ah - fair. Maybe to explicitly keep the relation you can rename `nunits` in the string template to `token_nunits`?

tests/mock/agents.py
KPrasch: Turns out be nontrival.  I pushed an update that I think will work without an explicit `PropertyMock`.

KPrasch:
Rebased over master @ a8bb61766ef9a3e194e1c897ba08ef2bd8a6e2de

tests/mock/agents.py
vzotova: `assert_only_expected_transactions`? I like when something starts `assert_only`

nucypher/blockchain/eth/actors.py
```diff
@@ -858,10 +862,35 @@ def locked_tokens(self, periods: int = 0) -> NU:
...
+
+    def __filter_divisible_stakes(self, stake: Stake) -> bool:
         """
-        The total number of staked tokens, i.e., tokens locked in the current period.
+        Helper function for use as a filtration predicate when determining
+        what constitutes a divisible stake amongst a given iterable of stakes.
         """
-        return self.locked_tokens(periods=0)
+        min_divisible_stake_value = self.economics.minimum_allowed_locked * 2
+        result = bool(stake.value >= min_divisible_stake_value)
```
fjarri, line 886: Seems a little redundant, comparison is already `bool`.

nucypher/blockchain/eth/agents.py
```diff
@@ -38,25 +38,29 @@
...
+    TRANSACTION_REGISTRY = dict()
 
     @classmethod
     def get_agent(cls,
-                  agent_class,
+                  agent_class: Type['EthereumContractAgent'],
                   registry: BaseContractRegistry = None,
-                  provider_uri: str = None,
-                  ) -> 'EthereumContractAgent':
+                  provider_uri: str = None):
```
fjarri, line 63: If the returned object is of type `agent_class` (that is, there's no funny business in the `agent_class` constructor), the return annotation can be preserved here by defining something like `AgentClass = TypeVar('AgentClass', EthereumContractAgent)` and using it for `agent_class` and return annotation.

nucypher/blockchain/eth/actors.py
KPrasch: Heh - yeah I have this style habit for bool functions.

nucypher/blockchain/eth/agents.py
KPrasch: Oh that's lovely :+1:

tests/cli/functional/conftest.py
```diff
@@ -16,18 +16,99 @@
...
+    mock_agent.minimum_allowed_bid = economics.worklock_min_allowed_bid
+    mock_agent.lot_value = economics.worklock_supply
+
+    yield mock_agent
+    mock_agent.reset()
+
+
+@pytest.fixture(scope='function')
+def mock_click_prompt(mocker):
+    return mocker.patch.object(click, 'prompt')
```
fjarri, line 106: Just curious, why mock separate `click` methods instead of just mocking `stdin`? Is there something that can't be done this way?

.bumpversion.cfg
```diff
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 2.1.0-beta.6
+current_version = 2.1.0-beta.7
```
cygnusv, line 2: We should leave the bump version to a separate PR.

nucypher/blockchain/eth/actors.py
```diff
@@ -858,10 +862,35 @@ def locked_tokens(self, periods: int = 0) -> NU:
 
     @property
     def current_stake(self) -> NU:
+        """The total number of staked tokens, i.e., tokens locked in the current period."""
+        return self.locked_tokens(periods=0)
+
+    @property
+    def active_stakes(self) -> Generator[Stake, None, None]:
```
cygnusv, line 869: IMO, it's clearer to annotate this as `Iterable[Stake]`, as discussed here (https://docs.python.org/3/library/typing.html#typing.Generator).

nucypher/blockchain/eth/actors.py
```diff
@@ -858,10 +862,35 @@ def locked_tokens(self, periods: int = 0) -> NU:
...
         """
-        return self.locked_tokens(periods=0)
+        min_divisible_stake_value = self.economics.minimum_allowed_locked * 2
+        result = bool(stake.value >= min_divisible_stake_value)
+        return result
+
+    @property
+    def divisible_stakes(self) -> List[Stake]:
+        """chop chop"""
+        stakes = list(filter(self.__filter_divisible_stakes, self.stakes))
```
cygnusv, line 892: If the result is a list, this could be simpler with just list comprehension, without needing a separate function for the filter (which is also not used anywhere else). Something like this:
```python
stakes = [stake for stake in self.stakes if stake.value >= min_divisible_stake_value]
```

.bumpversion.cfg
KPrasch: Absolutely - This a rebase artifact.

nucypher/blockchain/eth/actors.py
KPrasch: "simpler" is in the eye of the beholder here - I've never really preferred comprehensions to higher order functions or vice-versa so either way is OK IMO.

tests/cli/functional/conftest.py
KPrasch: I just felt is was not necessary to test any deeper than `click.confirm` in a effort trying to precisely target layers StdIn -> Click.  This way we dont test `click` itself handling input.

nucypher/blockchain/eth/agents.py
KPrasch: On second look - the return type is actually an instance of a subclass of `EthereumContractAgent`.  Does that mean we need to use a `Union` of all agent subclasses?

.bumpversion.cfg
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/actors.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/actors.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/agents.py
fjarri: Yep, that's what the typevar is for. I just checked it, seems to work as expected, except I forgot that you need to use `bound`:

```
from typing import TypeVar, Type

class A: pass
class B(A): pass
class C(A): pass

T_A = TypeVar('T_A', bound=A)

def test(tp:Type[T_A]) -> T_A:
    return tp()

a = test(A)
reveal_type(a) # should be A
b = test(B)
reveal_type(b) # should be B
c = test(C)
reveal_type(c) # should be C
```

tests/cli/functional/conftest.py
fjarri: Yes, pretty much. I think it will be simpler and close to mimicking an actual user input, unless I'm missing some prompts where it wouldn't work.

nucypher/blockchain/eth/agents.py
KPrasch: :heart_eyes: :heavy_check_mark:

tests/cli/functional/conftest.py
KPrasch: Yeah - this sounds like an excellent Idea to me, further isolating our application code from click at this test layer.  Let's address this in a follow up PR.  Hopefully this can also fix some of the flakiness of some tests that catch too much stdout currently.

nucypher/blockchain/eth/agents.py
```diff
@@ -38,25 +38,30 @@
...
 class ContractAgency:
     # TODO: Enforce singleton - #1506 - Okay, actually, make this into a module
 
     __agents = dict()
+    TRANSACTION_REGISTRY = dict()
 
     @classmethod
     def get_agent(cls,
-                  agent_class,
+                  agent_class: Type['EthereumContractAgent'],
```
fjarri, line 62: Should be `Type[Agent]` here.

KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/token.py
```diff
@@ -212,9 +212,33 @@ def __init__(self,
...
+        try:
+            that_stake = (other.index,
+                          other.value,
+                          other.first_locked_period,
+                          other.final_locked_period,
+                          other.staker_address,
+                          other.staking_agent.contract_address)
+        except AttributeError:
+            return False
+        return bool(sorted(map(str, this_stake)) == sorted(map(str, that_stake)))
```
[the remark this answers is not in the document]
KPrasch: Yeah - What was I thinking here... mapping to a `str`? - I will review and cherry-pick your commit.  Thank you!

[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/decorators.py
```diff
@@ -84,8 +89,37 @@ def wrapped(actor=None, *args, **kwargs):
...
+
+# TODO: Disable collection in prod (detect test package)
+COLLECT_CONTRACT_API = True
+
+
+class ContractInterfaces(Enum):
+    CALL = 0
+    TRANSACTION = 1
+    ATTRIBUTE = 2
+    UNKNOWN = 3
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark: Looks beautiful with constant sorrow! :-)

nucypher/blockchain/eth/decorators.py
```diff
@@ -84,8 +89,37 @@ def wrapped(actor=None, *args, **kwargs):
...
+
+
+def contract_api(interface: ContractInterfaces = None) -> Callable:
+    """Decorator factory for contract API markers"""
+    if not interface:
+        interface = ContractInterfaces.UNKNOWN
+
+    def decorator(agent_method: Callable) -> Callable:
+        """Marks an agent method as containing transactions"""
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/decorators.py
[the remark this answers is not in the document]
KPrasch: As a follow up - I wonder if this is a good example for documentation in the constant sorrow readme.

nucypher/cli/actions/confirm.py
```diff
@@ -32,45 +35,77 @@
...
     click.confirm(CONFIRM_ENABLE_WINDING_DOWN.format(staking_address=staking_address), abort=True)
     return True
 
 
-def confirm_staged_stake(staker_address, value, lock_periods) -> None:
-    click.confirm(CONFIRM_STAGED_STAKE.format(value=value,
+def confirm_staged_stake(staker_address: str, value: NU, lock_periods: int) -> bool:
+    """Interactively confirm a new stake reviewing all staged stake details."""
+    click.confirm(CONFIRM_STAGED_STAKE.format(nunits=str(value.to_nunits()),
+                                              tokens=str(value.to_tokens()),
```
cygnusv, line 82:
```suggestion
                                              tokens=value,
```
Otherwise, it would print the token amount without denomination. Also, `str()` is superfluous.

nucypher/blockchain/eth/decorators.py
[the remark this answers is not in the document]
fjarri: If would argue that an enum has its advantages when it comes to type checking/namespace pollution. Would be nice if there was a `ConstantSorrow` enum that would allow one to just write

```
class ContractInterfaces(CS.Enum):
    CALL
    TRANSACTION
    ATTRIBUTE
    UNKNOWN
```

and the field got assigned CS constants as values. But I don't want to ask Kieran to redo it once again.

nucypher/cli/actions/confirm.py
```diff
@@ -32,45 +35,77 @@
...
 
 
-def confirm_large_stake(value: NU = None, lock_periods: int = None) -> None:
+def confirm_large_stake(value: NU = None, lock_periods: int = None) -> bool:
+    """Interactively confirm a large stake and/or a long stake duration."""
     if value and (value > NU.from_tokens(150000)):
         click.confirm(CONFIRM_LARGE_STAKE_VALUE.format(value=value), abort=True)
     if lock_periods and (lock_periods > 365):
         click.confirm(CONFIRM_LARGE_STAKE_DURATION.format(lock_periods=lock_periods), abort=True)
+    return True
```
cygnusv, line 94: Was it necessary to make this function return a `bool`? Its returned value is never used, specially since the call to `click.confirm()` uses `abort=True`. Same comment to all the confirmation functions in here.

nucypher/cli/actions/select.py
```diff
@@ -88,24 +99,38 @@ def select_client_account(emitter,
...
-            BlockchainInterfaceFactory.initialize_interface(provider_uri=provider_uri, poa=poa, emitter=emitter)
-        signer = Signer.from_signer_uri(signer_uri) if signer_uri else None
+
+        if provider_uri:
+            # Lazy connect the blockchain interface
+            if not BlockchainInterfaceFactory.is_interface_initialized(provider_uri=provider_uri):
+                BlockchainInterfaceFactory.initialize_interface(provider_uri=provider_uri, poa=poa, emitter=emitter)
+
+        if signer_uri:
+            signer = Signer.from_signer_uri(signer_uri) if signer_uri else None
```
cygnusv, line 125:
```suggestion
            signer = Signer.from_signer_uri(signer_uri)
```

nucypher/cli/literature.py
cygnusv: > Primary motivator here for me is to keep python code from appearing in strings and to keep them as pure string templates.

Why is this a goal?

> I think leaving the string-expressed-logic is actually a potential vector for future regression.

Also curious to hear your reasoning here. In fact, I think current approach in this particular case is more prone to errors.

nucypher/utilities/seednodes.py
```diff
@@ -16,37 +16,28 @@
```
cygnusv, line 16: Good relocation 👍

KPrasch:
What do you think of this modified implementation of `@contract_api` in order to validate all outbound checksum addresses?

```python
def contract_api(interface: Optional[ContractInterfaces] = UNKNOWN_CONTRACT_INTERFACE) -> Callable:
    """Decorator factory for contract API markers"""

    def decorator(agent_method: Callable) -> Callable:
        """
        Marks an agent method as containing transactions
        and validates outbound checksum addresses.

        If `COLLECT_CONTRACT_API` is True when running tests,
        all marked methods will be collected for automatic mocking
        and integration with pytest fixtures.
        """
        if COLLECT_CONTRACT_API:
            agent_method.contract_api = interface

        @validate_checksum_address
        @functools.wraps(agent_method)
        def wrapped(*args, **kwargs):
            return agent_method(*args, **kwargs)
        return agent_method

    return decorator

```

nucypher/blockchain/eth/decorators.py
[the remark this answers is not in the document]
KPrasch: Interesting... @jMyles might be interested in this idea.

nucypher/cli/actions/confirm.py
KPrasch: Two motivations here: Personally, I like when a confirmation function returns `True`, secondly this allows `assert confirm_function()` in-test, which is a strong indication of function completion.

nucypher/utilities/seednodes.py
KPrasch: I mean - maintaining executable python code in string format is obviously less maintainable: it is not parsed by linters and is overtly implicit.

nucypher/cli/actions/confirm.py
KPrasch: :heavy_check_mark:  - Personally, I like the explicit `str`.

nucypher/cli/actions/select.py
KPrasch: :heavy_check_mark:

tests/cli/functional/conftest.py
```diff
@@ -58,9 +149,12 @@ def test_registry():
...
 def test_registry_source_manager(mock_testerchain, test_registry):
-    return make_mock_registry_source_manager(blockchain=mock_testerchain,
-                                             test_registry=test_registry,
-                                             mock_backend=True)
+    real_inventory = make_mock_registry_source_manager(blockchain=mock_testerchain,
+                                                       test_registry=test_registry,
+                                                       mock_backend=True)
+    yield
+    # restore the original state
+    NetworksInventory.NETWORKS = real_inventory
```
cygnusv, line 157: Perhaps both the `yield` and the restoration should be done in `make_mock_registry_source_manager`.

tests/cli/functional/test_worklock_cli_functionality.py
cygnusv: Does this need an issue?

tests/config/test_character_configuration.py
```diff
@@ -146,3 +153,42 @@ def test_ursula_development_configuration(federated_only=True):
...
+#
+# TODO: To be implemented
```
cygnusv, line 158:
```suggestion
# TODO: To be implemented  - #2016
```

tests/fixtures.py
```diff
@@ -585,7 +554,10 @@ def _make_agency(testerchain,
 
 @pytest.fixture(scope='module')
 def test_registry_source_manager(testerchain, test_registry):
-    return make_mock_registry_source_manager(blockchain=testerchain, test_registry=test_registry)
+    real_inventory = make_mock_registry_source_manager(blockchain=testerchain, test_registry=test_registry)
+    yield
+    # restore the original state
+    NetworksInventory.NETWORKS = real_inventory
```
cygnusv, line 560: See comment above re: `make_mock_registry_source_manager`

nucypher/blockchain/eth/decorators.py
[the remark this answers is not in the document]
fjarri: After some thought, I don't think CS is the best place for it. The attributes are not created ad-hoc in each module which is CS's modus operandi, but declared once. And they need to have `ContractInterfaces` type, for consistency. The only CS constant's property that is needed here is that printing an attribute gives you its full name (plus the enum's name, preferably). And the snippet above is actually syntactically incorrect, attributes must have either values or annotations, so it'll look like:
```
class ContractInterfaces(MyEnum):
    CALL: 'ContractInterfaces'
    TRANSACTION: 'ContractInterfaces'
    ATTRIBUTE: 'ContractInterfaces'
    UNKNOWN: 'ContractInterfaces'
```
(which will also help `mypy`). 

All in all, it seems easier to just use the standard library enum and assign CS constants to its attributes...

tests/cli/functional/conftest.py
KPrasch: :heavy_check_mark:

tests/config/test_character_configuration.py
KPrasch: :heavy_check_mark:

tests/fixtures.py
KPrasch: :heavy_check_mark:

PR #1959, opened 2020-05-06
Some renamings

vzotova:
Fixes #1717 
Refs #1933 
Fixes #1530
Fixes #1834 

This PR could be done without hard fork but upgrading will take two steps

* policy reward -> policy fee
* mine -> mint
* lastMinedPeriod -> previousFeePeriod
* burn -> donate
* workerToStaker -> stakerFromWorker
* confirmActivity -> commitToNextPeriod
* lastActivePeriod -> lastCommittedPeriod
* setWorker -> bondWorker
* detach-worker-> unbond-worker

PR #1990, opened 2020-05-14
Removal of 'min' from fee range & more informative/consistent comments 

arjunhassard:
Follow on from #1959 

`MinFeeRateRange` -> `FeeRateRange`
`min_fee_rate_range` -> `fee_rate_range`
etc 

Use of _min_ is a bit misleading when the variable comprises a fixed, universal maximum, minimum and default rate (also referred to as the "global fee range"). Furthermore, stakers must choose their own _minimum_ acceptable fee rate, which must fall within the global fee range, and so unnecessary name similarities may cause unnecessary confusion.

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1203,7 +1203,7 @@ class Staker(NucypherTokenActor):
 
     @property
     def min_fee_rate(self) -> int:
-        """Minimum fee rate that staker earns"""
+        """Minimum fee rate that staker accepts"""
         staker_address = self.checksum_address
         min_fee = self.policy_agent.get_min_fee_rate(staker_address)
         return min_fee
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -511,9 +511,9 @@ class ContractAdministrator(NucypherTokenActor):
                                                         deployer_address=self.deployer_address,
                                                         economics=self.economics)
         receipt = policy_manager_deployer.set_fee_rate_range(minimum=minimum,
-                                                                 default=default,
-                                                                 maximum=maximum,
-                                                                 gas_limit=transaction_gas_limit)
+                                                             default=default,
+                                                             maximum=maximum,
+                                                             gas_limit=transaction_gas_limit)
         return receipt
 
 
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -502,10 +502,10 @@ class ContractAdministrator(NucypherTokenActor):
         return filepath
 
     def set_fee_rate_range(self,
-                               minimum: int,
-                               default: int,
-                               maximum: int,
-                               transaction_gas_limit: int = None) -> dict:
+                           minimum: int,
+                           default: int,
+                           maximum: int,
+                           transaction_gas_limit: int = None) -> dict:
 
         policy_manager_deployer = PolicyManagerDeployer(registry=self.registry,
                                                         deployer_address=self.deployer_address,
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1191,7 +1191,7 @@ class Staker(NucypherTokenActor):
     @only_me
     @save_receipt
     def set_min_fee_rate(self, min_rate: int) -> Tuple[str, str]:
-        """Public facing method for staker to set their minimum acceptable fee rate"""
+        """Public facing method for staker to set the minimum acceptable fee rate for their associated worker"""
         minimum, _default, maximum = self.policy_agent.get_fee_rate_range()
         if min_rate < minimum or min_rate > maximum:
             raise ValueError(f"Minimum fee rate {min_rate} must fall within global fee range of [{minimum}, {maximum}]")
@@ -1210,7 +1210,7 @@ class Staker(NucypherTokenActor):
 
     @property
     def raw_min_fee_rate(self) -> int:
-        """Minimum acceptable fee rate set by staker.
+        """Minimum acceptable fee rate set by staker for their associated worker.
         This fee rate is only used if it falls within the global fee range.
         If it doesn't a default fee rate is used instead of the raw value (see `min_fee_rate`)"""
         staker_address = self.checksum_address
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -501,7 +501,7 @@ class ContractAdministrator(NucypherTokenActor):
             file.write(data)
         return filepath
 
-    def set_min_fee_rate_range(self,
+    def set_fee_rate_range(self,
                                minimum: int,
                                default: int,
                                maximum: int,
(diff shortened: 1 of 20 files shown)


nucypher/blockchain/eth/actors.py
```diff
@@ -501,7 +501,7 @@ def save_deployment_receipts(self, receipts: dict, filename_prefix: str = 'deplo
             file.write(data)
         return filepath
 
-    def set_min_fee_rate_range(self,
+    def set_fee_rate_range(self,
                                minimum: int,
```
vzotova, line 505: lines are not aligned after renaming

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -13,7 +13,7 @@ import "contracts/proxy/Upgradeable.sol";
 
 
 /**
-* @notice Contract holds policy data and locks fees
+* @notice Contract holds policy data and locks accrued policy fees until collection
 * @dev |v5.1.1|
```
vzotova, line 17: because some signatures changed - need to change major version, `v6.1.1`

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -13,7 +13,7 @@ import "contracts/proxy/Upgradeable.sol";
 
 
 /**
-* @notice Contract holds policy data and locks fees
+* @notice Contract holds policy data and locks accrued policy fees until collection
```
vzotova, line 16: not only until collection, staker must fulfill a commitment before fee collection

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -103,7 +103,7 @@ contract PolicyManager is Upgradeable {
 
     mapping (bytes16 => Policy) public policies;
     mapping (address => NodeInfo) public nodes;
-    Range public minFeeRateRange;
+    Range public FeeRateRange;
```
vzotova, line 106: this shoud start from lower case

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -145,24 +145,24 @@ contract PolicyManager is Upgradeable {
     }
 
     /**
-    * @notice Set range for the minimum fee rate for all nodes
+    * @notice Set maximum, minimum and and default fee rate for all stakers and all policies ('global fee range')
```
vzotova, line 148: two "and"

nucypher/blockchain/eth/actors.py
arjunhassard: ✔️

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
arjunhassard: Can definitely remove 'until collection'. I put it there to preempt stakers reading the contract description and thinking their fees were locked for a similar duration as their stakes. Also, I don't think that 'until collection' implies that they _don't_ need to fulfil some obligations to get their fees, it's more neutral

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
arjunhassard: ✔️

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
arjunhassard: changed all to lowercase 'f' except:  `setFeeRateRange` and `delegateGetFeeRateRange`

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
arjunhassard: ✔️

nucypher/blockchain/eth/actors.py
vzotova: also there are few more places where is the same

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
vzotova: But events start from uppercase :) so for lines 54 and 155 better to return back to `FeeRateRangeSet`

arjunhassard:
@vzotova Thank you very much for reviewing, I have fixed the problems you spotted!

Regarding Alice's ability to choose a higher rate than the range max, that's a very fair point and I understand why the 'min' was retained. In practice, network users can discover all stakers' minimum acceptable fees, so there's no straightforward economic reason* to pay more than the cheapest rate that gets you the requisite `n` stakers. So perhaps we can add a comment in PolicyManager that explains that technically Alice can choose to pay more, but it's not very relevant to the marketplace/price dynamics in practice. I believe stating this characteristic in the name itself is overkill. 

Regarding global fee range, if _global_ sounds too strict given the overpaying Alice caveat, perhaps 'universal' is better? I would love there to be a consistent name we refer to (that I can use in a pricing protocol paper), and that it tells you this rule applies to all stakers equally.  

*Alices who wish to favour specific stakers by paying them more than the range max is not a big problem in isolated cases but is not preferable behaviour for the health of the marketplace (it's generally better when fees are more uniform and broadly distributed). Moreover it could be an attack on a future mechanism that links subsidies to fees OR some kind of performance logging or reputation system.

vzotova:
> @vzotova Thank you very much for reviewing, I have fixed the problems you spotted!
> 
> Regarding Alice's ability to choose a higher rate than the range max, that's a very fair point and I understand why the 'min' was retained. In practice, network users can discover all stakers' minimum acceptable fees, so there's no straightforward economic reason* to pay more than the cheapest rate that gets you the requisite `n` stakers. So perhaps we can add a comment in PolicyManager that explains that technically Alice can choose to pay more, but it's not very relevant to the marketplace/price dynamics in practice. I believe stating this characteristic in the name itself is overkill.
> 
> Regarding global fee range, if _global_ sounds too strict given the overpaying Alice caveat, perhaps 'universal' is better? I would love there to be a consistent name we refer to (that I can use in a pricing protocol paper), and that it tells you this rule applies to all stakers equally.
> 
> *Alices who wish to favour specific stakers by paying them more than the range max is not a big problem in isolated cases but is not preferable behaviour for the health of the marketplace (it's generally better when fees are more uniform and broadly distributed). Moreover it could be an attack on a future mechanism that links subsidies to fees OR some kind of performance logging or reputation system.

ok, sold :) left just few things that I mentioned and I'll approve this

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
arjunhassard: haha i have so much to learn

nucypher/blockchain/eth/actors.py
```diff
@@ -1203,17 +1203,16 @@ def set_min_fee_rate(self, min_rate: int) -> Tuple[str, str]:
 
     @property
     def min_fee_rate(self) -> int:
-        """Minimum acceptable fee rate"""
+        """Minimum fee rate that staker earns"""
```
cygnusv, line 1206:
```suggestion
        """Minimum fee rate that staker accepts"""
```

nucypher/blockchain/eth/agents.py
```diff
@@ -820,17 +820,20 @@ def get_fee_amount(self, staker_address: str) -> int:
         fee_amount = self.contract.functions.nodes(staker_address).call()[0]
         return fee_amount
 
-    def get_min_fee_rate_range(self) -> Tuple[int, int, int]:
-        minimum, default, maximum = self.contract.functions.minFeeRateRange().call()
+    def get_fee_rate_range(self) -> Tuple[int, int, int]:
+        """Check maximum, minimum & default fee rate for all stakers and all policies ('global fee range')"""
```
cygnusv, line 824: Follow the same order of variables that's returned
```suggestion
        """Check minimum, default & maximum fee rate for all stakers and all policies ('global fee range')"""
```

PR #1991, opened 2020-05-14
Documentation of running ETH node and using clef

### LINKED ISSUE #1952: Improve guide regarding running local providers
1. With hosted node solutions, providing a guide regarding running local nodes (especially with respect to mainnet) has lost some relevance, either remove and/or we need to add documentation for non-local ethereum provider.

2. The Clef guide has ballooned and can probably be split between a guide and a reference

derekpierre:
Fixes #1952 .

### DIFF
--- a/docs/source/guides/ethereum_node.rst
+++ b/docs/source/guides/ethereum_node.rst
@@ -1,0 +1,231 @@
+.. _using-eth-node:
+
+======================
+Using an Ethereum Node
+======================
+
+For operations that require an ethereum provider for transaction signing and broadcasting, either a local or remote
+ethereum node may be used. For general background information about choosing a node technology and operation,
+see https://web3py.readthedocs.io/en/stable/node.html.
+
+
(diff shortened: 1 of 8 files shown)


docs/source/guides/ethereum_node.rst
```diff
@@ -0,0 +1,231 @@
...
+For operations that require an ethereum provider for transaction signing and broadcasting, either a local or remote
+ethereum node may be used. For general background information about choosing a node technology and operation,
+see https://web3py.readthedocs.io/en/stable/node.html.
+
+
+Local Ethereum Node
+~~~~~~~~~~~~~~~~~~~
+
+This is the typical configuration for a locally operated trusted ethereum node. For detailed information on using the
+geth CLI and Javascript console, see https://geth.ethereum.org/interface/Command-Line-Options.
```
KPrasch, line 16: :+1: - I am happy to see these resources being used here instead of maintaining them ourselves.

PR #1997, opened 2020-05-18
Fix some small signer bugs. Update IBEX registry 🐐

### LINKED ISSUE #1941: CLI: Don't collect password when using a signer
**Describe the Bug**
If a CLI command uses a signer, then there's no need to collect the password, as it's something that the signer must handle independently

**To Reproduce**
Try to stake with a signer

 - OS Version:
 - Nucypher Version:
 - Ethereum Node Version:

cygnusv:
- Closes #1941


nucypher/blockchain/eth/interfaces.py
```diff
@@ -442,6 +442,10 @@ def __log_transaction(self, transaction_dict: dict, contract_function: ContractF
         # Format
         if tx.get('to'):
             tx['to'] = to_checksum_address(contract_function.address)
+        try:
+            tx['selector'] = contract_function.selector
+        except AttributeError:
+            pass
```
KPrasch, line 448: try / except / pass is always an eyebrow-raiser for me - Here's an alternate approach:

- Relocate `BlockchainInterface.__log_transaction` to `EthereumClient` and `Signer` methods .
- Override a base abstract method to handle any local attributes or functionality

KPrasch:
:cowboy_hat_face:

Issue #2002, opened 2020-05-19
Nonce too low


cygnusv:
**Describe the Bug**
I was trying to run an Ursula for the first time and, as expected, the first thing it should do is confirm activity. It did, but immediately it tried to confirm again, reusing the same nonce.
It looks like a chain synchronization issue. 

**To Reproduce**
I'm using a brand new staker and ursula, with Infura as provider and the Keystore signer.

**Traceback or Screenshots (Optional) **
```
 ,ggg,         gg                                                     
dP""Y8a        88                                   ,dPYb,            
Yb, `88        88                                   IP'`Yb            
 `"  88        88                                   I8  8I            
     88        88                                   I8  8'            
     88        88   ,gggggg,    ,g,     gg      gg  I8 dP    ,gggg,gg 
     88        88   dP""""8I   ,8'8,    I8      8I  I8dP    dP"  "Y8I 
     88        88  ,8'    8I  ,8'  Yb   I8,    ,8I  I8P    i8'    ,8I 
     Y8b,____,d88,,dP     Y8,,8'_   8) ,d8b,  ,d8b,,d8b,_ ,d8,   ,d8b,
      "Y888888P"Y88P      `Y8P' "YY8P8P8P'"Y88P"`Y88P'"Y88P"Y8888P"`Y8

the Untrusted Re-Encryption Proxy.
Crimson King Ivory Alembic

Starting services...
Starting services...
✓ Database pruning
✓ Database pruning
Starting Learning Loop NOW.
Finished learning about all seednodes.
Found 0 known node metadata files at /home/plavalaguna/.local/share/nucypher/known_nodes/metadata
No seednodes were available after 3 attempts
Can't learn right now: Need some nodes to start learning from.
Can't learn right now: Need some nodes to start learning from.
✓ Node Discovery (ibex)
✓ Node Discovery (ibex)
✓ Availability Checks
✓ Availability Checks
Checking for new period. Current period is 18401
MISSED COMMITMENTS - 18401 missed staking commitments detected.
Made a commitment to period 18401
[TX-COMMITTONEXTPERIOD] | chainId: 4, nonce: 0, from: ..., gasPrice: 1 gwei, selector: 0x41fe00a0
Broadcasting COMMITTONEXTPERIOD Transaction (154427 gwei @ 1000000000)...
Broadcasting COMMITTONEXTPERIOD Transaction (154427 gwei @ 1000000000)...
[...]

PR #2003, opened 2020-05-19
Test reproducing double CA bug and Fix 1/2 of (#2002)

KPrasch:
This PR proves that CA is called only once on start up then once period.

Advances #2002 

Co-Author: @cygnusv

### DIFF
--- a/tests/blockchain/eth/entities/actors/test_worker.py
+++ b/tests/blockchain/eth/entities/actors/test_worker.py
@@ -46,24 +46,27 @@ def test_worker_auto_commitments(mocker,
 
     commit_spy.assert_not_called()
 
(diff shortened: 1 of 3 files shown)


tests/blockchain/eth/entities/actors/test_worker.py
```diff
@@ -54,9 +62,11 @@ def verify(_):
         last_committed_period = staker.staking_agent.get_last_committed_period(staker_address=staker.checksum_address)
         current_period = staker.staking_agent.get_current_period()
         assert (last_committed_period - current_period) == 1
+        assert commit_spy.call_count == current_period - initial_period + 1
```
KPrasch, line 65: Heh - this gives me a chuckle.

Issue #2005, opened 2020-05-20
[Draft] When are there enough tests?


KPrasch:
#### Problem 

The codebase is not resistant against regression and lacks sufficient coverage to verify correctness. Current test suites are inaccurate, fragile, slow, course, and do not cover enough of the state space of nucypher. This leads to lower code quality, undefined code behavior, and more importantly fear of change, wasted development hours, and difficulty on-boarding.

#### Solution

Stop living in fear. Develop an incremental but comprehensive testing strategy using "test pyramid theory" and layer isolation via mock tooling.

![](https://www.oreilly.com/library/view/continuous-delivery-with/9781787125230/assets/27ed025e-e2e6-46f6-b1ce-32678d75428c.png)

#### Proposed Testing Roadmap

1) ~~Performance: Accept PRs #1949, #1998, and #1989 as immediate help around worklock.  Leave current E2E Tests in-place and develop an enclave functional test suite.  All new incoming tests must implement mocking and precisely target *only some* of nucypher's abstraction layers.~~ Complete

2) ~~Reorganize: Create a new top-level ACCEPTANCE test directory and relocate tests that execute all layers (CLI through to the EVM).  Understand these tests to be slow end-to-end integration tests against PyEVM. Over time we can introduce other clients into the suite (trinity, geth, parity).~~ Complete

3) Test Agent Integrates with Contracts.  Verify the correctness and compatibility of agents with public contract APIs.  This helps cover the untested space leftover from mocking agents in FT, and is the highest priority after performance.  Fuzz / paramaterize contract API inputs via agency to search for unhandled code paths and explore the state space of contracts from python.

4) Test Actor Integrates Agent. Verify correct usage and implementation of contract logic via agents.

5) Copy / Adjust existing CLI tests (now ACCEPTANCE) to test from CLI to Actor layer *only*

6) Interoperability: Follow-up on step 2 with system tests with Trinity / Geth / Parity / Linux / Mac / Windows, etc.  Use a tox configuration on CI/CD for python versions 3.6 - 3.8.

#### Considerations

- Deprecation of `--run-slow` in favor of multiple test entry points
- Do *Contract Tests* stand alone, implementing it's own discrete testing strategy (@vzotova) ?
- How do the federated (offchain) vs. decentralized test domains fit into the plan (@jMyles)?
- What is the appropriate number of entry points / suites? (Contracts, Acceptance, Functional, Federated, Systems, Metrics, etc.)

#### Resources

[...]

PR #2017, opened 2020-05-20
Reorganize tests in accordance with #2005, #1708

### LINKED ISSUE #1708: Organize tests and optimization test run time to under 120s.
As we approach mainnet, we need to ensure that we have a short enough cycle of emergency bugfixes to be able to extinguish fires with a minimum of consequences.  A short build time is a big part of this.

KPrasch:
### What this does

Reorganizes tests and modifies the CI configuration.  

There are *zero logic changes* (198 line diff from imports and new modules - 98 in CI config).

Provides an initial foundation for new unit and integration test authorship. Take the first steps described in #2005, relocating existing slow end-to-end tests to an ACCEPTANCE test suite.

####  Fixes #1708  (Depending on your interpretation of this Issue.)

I assert this PR provides a fix for #1708 by introducing a test suite of ~200 test points that run in less than 120 seconds (more than 100 points in 500ms), excluding acceptance, metrics, and contracts.  

Additionally, a small number of tests take up a large portion of test run time.  These tests are easy to pick out in CI builds ( for example: `test_hendrix_handles_content_length_validation` takes up nearly half of the runtime of integration suite). We can further profile an optimize for performance in follow up PRs.

#### Multiple Test Entry Points

Modifies the codebase to adhere to multiple test entry points (as directories) as discussed in #2005:
  * contracts
  * unit
  * integration 
  * acceptance
  * metrics

#### Reflow CircleCI Config

Reflects the new test directories as jobs.  Parallelism tuning for fast individual job runtimes and more useful feedback. Try to save expensive jobs for other workflows where they are more likely to pass by depending on fast running test suite coverage and accuracy.  Deliver feedback about less expensive tests first - keeping the test runtime of unit and functional tests under 120s.

Screenshot from 2020-05-22 18-12-39

#### Future Development

- Addition of a "meta" entry point with markers for *federated*
- Character based entry point markers `pytest --alice` `pytest --ursula`
- Deprecation of `--run-slow`
- Pre-Commit hook to run unit tests

#### How to Review

- Checkout this PR locally and browse through `tests`
- Run individual test suites
- Inspect CiecleCI workflow and config
- Analyze the diff and it is as minimal as possible and preserves all code.


KPrasch:
Going to make a second attempt - I think I organize a bit more accurately.

KPrasch:
Okay done - This version now based over master @ 8b7ba918b588e2ae51684d06ae8e24f7a3b28956

KPrasch:
> What's up with the reports directories in each test subdirectory

Off - certainly a mistake when running `git add`.  

I will remove these - thanks for catching this!

tests/integration/blockchain/test_currency.py
```diff
@@ -0,0 +1,117 @@
...
+    expected = NU(min_NU_locked, 'NU')
+    assert min_allowed_locked == expected
+
+    # Starting Big
+    min_allowed_locked = NU(min_NU_locked, 'NU')
+    assert token_economics.minimum_allowed_locked == int(min_allowed_locked)
+    assert token_economics.minimum_allowed_locked == int(min_allowed_locked.to_nunits())
+    assert str(min_allowed_locked) == '15000 NU'
+
+    # Alternate construction
```
derekpierre, line 40: Not to get too much into the weeds as we mentioned on the call - but it feels like from this line onward is a `unit` test 😊

tests/integration/network/test_failure_modes.py
```diff
@@ -172,6 +172,8 @@ def test_huge_treasure_maps_are_rejected(federated_alice, federated_ursulas):
         )
     """
 
+
+@pytest.mark.skip("Hangs forever")
```
derekpierre, line 176: 😱

KPrasch: @vepkenez

tests/integration/blockchain/test_currency.py
KPrasch: I agree 100% - I think I just missed this one in the great sweep :-)

KPrasch:
@jMyles - It's hard to find a sufficiently generic set of entry point names to prevent controversy!

How specifically do you think we can improve the test organization here in a way that expresses both run-time and accuracy?  

I dont want to commit us all to another large renaming PR, but also dont want to let the perfect become the enemy of the good.

Thanks for your review! :-)

tests/fixtures.py
```diff
@@ -985,7 +985,9 @@ def highperf_mocked_bob(fleet_of_highperf_mocked_ursulas):
 def stdout_trap(mocker):
     trap = StringIO()
     mocker.patch('sys.stdout', new=trap)
-    return trap
+    trap.truncate(0)
+    yield trap
+    trap.truncate(0)
```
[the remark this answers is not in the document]
fjarri: `stdout_trap()` is obsoleted by PR #2033

jMyles:
There are at least three problems with `(unit, integration, acceptance)` as packages:

1) The boundaries between these types of test are unclear enough that many of the best tests will fall "on the line", with authors experiencing confusion or controversy about where to place new tests.
2) It is possible that best organizational practices will involve tests of varying weight and breadth *in the same module* to tell a story correctly.  If an integration test uses a module-level fixture, it might make great sense for a subsequent acceptance test to use that same fixture instance (blasphemy, I know).
3) This set of testing types, or even the broader understanding of "Testing levels" as described on the wikipedia "software testing" article, are not static, and are subject to dramatic change both in definition and in number of the years.  Here is a list of 100 types of software testing: https://www.guru99.com/types-of-software-testing.html

derekpierre:
@jMyles perfection in this regard is difficult since there are differing views on testing. In terms of boundary cases for where cases belong, that's less of an issue in my mind - it improves over time, doing reviews will help and we can comment on where it may belong, and it's not the end of the world if a small number are in the so-called "wrong" spot.

While I agree that there are different types of testing that can be done, I think on benefit of this categorization is that as we encounter bugs, we may have better indications of the types of tests that are lacking, and improve that over time i.e. are we encountering bugs in individual classes (`unit` tests), or in how they get used (`integration`) or are the bugs more system-wide (`acceptance`) - that is a strong benefit in my view.

PR #1968, opened 2020-05-07
Correct path handling in keystore signer / support individual files

svenski123:
Simple bugfix / feature enhancement to keystore signer (related to issue #1962).

Path given to keystore signer is stat'd and:
- if it is a directory, it is scanned and regular directory entries are processed
- if it is a regular file, it is processed
- if something else (i.e. device node, etc.), an exception is raised

Additionally during directory scanning, OSError exceptions (such as permission error, etc.) are appended to InvalidSignerURI exception text and raised.

PR #2018, opened 2020-05-21
Fix up 3rd local keystore test case

svenski123:
Clean up of 3rd local keystore related testcase missed out in #1968.


svenski123:
@KPrasch This should take care of all three local keystore test cases.

KPrasch:
:cowboy_hat_face:

PR #1890, opened 2020-04-15
Introduce Pre-Commit Framework

KPrasch:
##### Give it a try: 

`pip install nucypher -e . -r dev-requirements`
`pre-commit install`
`pre-commit run --all-files`

Pre-commit hooks will execute on commits automatically preventing commit on failure.  This can be disabled in a few ways, to name a few: 

- `PRE_COMMIT_ALLOW_NO_CONFIG=1 git commit ...`
- `git commit ... --no-verify`

##### Example commit stage output:
```
(nucypher) kieran@pegasus:~/Git/nucypher$ git commit -m "Pre-hook configuration; Includes bandit setup and custom circleCI checker."
Don't commit to branch...................................................Passed
Forbid new submodules....................................................Passed
Check for byte-order marker..............................................Passed
Check that executables have shebangs.....................................Passed
Check for added large files..............................................Passed
Check for broken symlinks............................(no files to check)Skipped
Check Yaml...............................................................Passed
Check Toml...........................................(no files to check)Skipped
Check python ast.....................................(no files to check)Skipped
Check for merge conflicts................................................Passed
Debug Statements (Python)............................(no files to check)Skipped
Detect Private Key.......................................................Passed
Check docstring is first.............................(no files to check)Skipped
Check vcs permalinks.....................................................Passed
bandit...............................................(no files to check)Skipped
Validate CircleCI Config.................................................Passed
[milestoning db23059d0] Pre-hook configuration; Includes bandit setup and custom circleCI checker.
 3 files changed, 52 insertions(+)
 create mode 100644 .bandit
 create mode 100644 .pre-commit-config.yaml
 mode change 100644 => 100755 scripts/hooks/validate_circleci_config.sh

```

PR #2019, opened 2020-05-21
Add pre-commit hook to ensure that the nucypher license is included in python files

### LINKED ISSUE #2012: Pre-commit hook to ensure AGPL license inclusion
Related to #1890

derekpierre:
The following files were created/edited, everything else was updated by the pre-commit hook:
- `.pre-commit-config.yaml`
- `.nucypher-license`

`insert-license` hook - see the official list here: https://pre-commit.com/hooks.html

Closes #2012 .
Related to #1890 .

**One item of note:**
This hook adds a leading space at the beginning of every line of the license i.e.
```python
"""
 This file is part of nucypher.

 nucypher is free software: you can redistribute it and/or modify
 it under the terms of the GNU Affero General Public License as published by
 the Free Software Foundation, either version 3 of the License, or
 (at your option) any later version.

 nucypher is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU Affero General Public License for more details.

"""
```

This is done via https://github.com/Lucas-C/pre-commit-hooks/blob/master/pre_commit_hooks/insert_license.py#L74. 
I suspect that this is because of typical comment characters that are used at the beginning of comment lines eg. "*" or "//" for Java, or "#" for Python, in which case a space is needed to separate the character from the text. However, we use a docstring for the license. 

In files that already include the license, we don't have a space at the beginning of each line, however, the tool is smart enough to recognize that as a valid license and doesn't modify the file eg.

```
"""
This file is part of nucypher.

nucypher is free software: you can redistribute it and/or modify
it under the terms of the GNU Affero General Public License as published by
the Free Software Foundation, either version 3 of the License, or
(at your option) any later version.

nucypher is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
GNU Affero General Public License for more details.

"""
```

Ultimately all this means is that existing files with licenses in them remain the same, and when licenses are added to the top of a file by the hook, the license includes a leading space for each line. Some files will have licenses with spaces at the beginning of each line, and some files won't. IMHO this is fine - if you feel strongly about this discrepancy, feel free to speak up.

### DIFF
--- a/.nucypher-license
+++ b/.nucypher-license
@@ -1,0 +1,14 @@
+This file is part of nucypher.
+
+nucypher is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+nucypher is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -37,7 +37,7 @@ repos:
       - id: check-vcs-permalinks
 
   - repo: https://github.com/Lucas-C/pre-commit-hooks
-    rev: v1.1.7
+    rev: ffbd448645bad2e7ca13f96fca5830058d27ccd5
     hooks:
       # License
       - id: insert-license
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,6 +36,15 @@ repos:
       # Docs
       - id: check-vcs-permalinks
 
+  - repo: https://github.com/Lucas-C/pre-commit-hooks
+    rev: v1.1.7
+    hooks:
+      # License
+      - id: insert-license
+        files: \.py$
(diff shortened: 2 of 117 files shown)


fjarri:
While having initial spaces in the license is not a big deal, couldn't we just copy that hook and modify it? Plus we won't have a script from some third-party repo to just act automatically on the whole codebase while having write permissions.

KPrasch:
@fjarri Raises a good point about how intrusive we want to allow file re-writers to be.  Here are some options:

- Only use re-writers in the local repo
- Turn off automatic updates for pre-commit hooks
- other ideas?

derekpierre:
@KPrasch are the pre-commit hooks fixed based on the `rev:` version specified in the yaml file? The only way to update is to manually change the rev file.

KPrasch:
> are the pre-commit hooks fixed based on the rev: version specified in the yaml file? 

Yes - I think so <https://pre-commit.com/#updating-hooks-automatically>.

fjarri:
Technically, in the config the pulled version of the hook is tied to a tag, not to a revision number, so it is possible for a malicious third party to "update it".

KPrasch:
Perhaps this is a compelling enough reason to only use self-maintained file re-writers.

derekpierre:
You can use a commit hash for the value of `rev:`:

(From https://pre-commit.com/)
> pre-commit assumes that the value of rev is an immutable ref (such as a tag or SHA) and will cache based on that

I tried it locally and that worked.
`rev: ffbd448645bad2e7ca13f96fca5830058d27ccd5`.

KPrasch:
Ah - that's a good solution IMO.

Issue #2016, opened 2020-05-20
Flaky Tests List


KPrasch:
We have a small pile of sometimes-failing tests that need investigation: 

- Finnegan's Wake Demo
- Heartbeat Demo
- `test_destroy_configurations`
- Functional suite fails when run all together with CLI tests due to background-running Ursula from previous fixtures:

```
        output = stdout_trap.getvalue()
>       assert not output
E       assert not "Can't learn right now: Need some nodes to start learning from.\n"

```

KPrasch:
Here's an example: <https://app.circleci.com/pipelines/github/nucypher/nucypher/4721/workflows/7d166e90-6397-492b-a09f-f5f2bb98170c/jobs/86605>

KPrasch:
Related to #2021 #2006

cygnusv:
I guess you want to add `test_contract_upgradeability` to that list too?

KPrasch:
> I guess you want to add test_contract_upgradeability to that list too?

Not quite sure Id call that one "flaky" since re-running it does not cause it to pass, unless I've misunderstood.

fjarri:
Heartbeat demo (`doctor.py`) fails because it can't find `policy-metadata.json` (e.g. https://app.circleci.com/pipelines/github/nucypher/nucypher/4830/workflows/eb0cc07b-c546-44e7-bf14-f684313de218/jobs/88063/steps) - where should it appear from during a normal run?

cygnusv:
The original problem is this:
```
nucypher.policy.policies.Rejected: Selected Ursulas rejected too many arrangements - only 2 of 3 accepted.
 ```
There's no policy metadata since the policy wasn't created ("only 2 of 3 accepted")

KPrasch:
Here's a one-off unexpected CI failure that was resolved by re-running:

 https://app.circleci.com/pipelines/github/nucypher/nucypher/4884/workflows/36b39084-cc3c-4d09-baf6-0523f5f3bae2/jobs/88737/steps

KPrasch:
Unsure if this is flaky but certainly unexpected failure from side effects: <https://app.circleci.com/pipelines/github/nucypher/nucypher/4954/workflows/fb426163-2747-4546-9158-f9809b65ad3e/jobs/89720/steps>

cygnusv:
> Here's a one-off unexpected CI failure that was resolved by re-running:
> 
[...]

PR #2032, opened 2020-05-22
Check the actual boolean values instead of asserting a non-empty tuple

fjarri:
Seems to be a typo in tests. The assertions as they are are always true, and generate pytest warnings.


KPrasch:
Haha - Thank you! I have noticed these as well..

derekpierre:
😱

KPrasch:
@vzotova

fjarri:
Uh oh, I seem to have forgotten myself how `all()` works :)

KPrasch:
Two approvals with that bug!

fjarri:
That heartbeat demo fail doesn't seem to be related to the PR

KPrasch:
Yep - its flaky - See #2016

PR #2049, opened 2020-05-27
Bugfixes to WorkLock command

cygnusv:


nucypher/cli/commands/worklock.py
```diff
@@ -194,17 +202,22 @@ def bid(general_config, worklock_options, force, hw_wallet, value):
     value = int(Web3.toWei(Decimal(value), 'ether'))
 
     if not force:
-        paint_bidding_notice(emitter=emitter, bidder=bidder)
-        click.confirm(f"Place WorkLock bid of {prettify_eth_amount(value)}?", abort=True)
+        if not existing_bid_amount:
+            paint_bidding_notice(emitter=emitter, bidder=bidder)
+            click.confirm(f"Place WorkLock bid of {prettify_eth_amount(value)}?", abort=True)
```
derekpierre, line 207: Should the text here and in other places in the file belong in `nucypher.cli.literature`?

cygnusv: Yeah, I think that's OK, but I tried to keep the scope of this reduced. There are some other strings in this module anyway, so it needs a revision some time.

KPrasch:
Let's ensure there is sufficient follow up planned for missing test coverage so this is the last time we encounter a simple API regression with worklock commands.

PR #2056, opened 2020-05-28
Simplify staker sampling

fjarri:
More logical weighted sampling for stakers, with the sampler extracted to a separate class for the ease of unit testing. `additional_ursulas` and `attempts` parameters removed from `sample()`.

The second commit makes exposes the staker sampler as an iterator-like object. `BlockchainPolicy.sample_essential()` now works as follows:

- `quantity` stakers is needed
- Get the stakers reservoir based on the list of stakers returned by the contract (excluding `handpicked_ursulas` right away)
- Draw `quantity` of stakers, see which ones are already known, and send the rest to the learner
- Wait a little, check for newly known nodes. If it's still not enough (some nodes are being checked), draw more stakers from the reservoir and send them to the learner.
- Loop until time is expired or we have enough nodes

The nodes that were drawn first have the priority. The returned result is shuffled (or should it be shuffled in `sample()` instead, that is shuffle the known stakers as well?)

Advantages over the previous sampling method:
- no need to "pre-sample" 1.5x of Ursulas; we draw samples as we need them. 
- simpler algorithm structure
- we give addresses to the learner in a batch instead of one by one
- do not run the loop again and again; `sleep()` and let the learner do its thing

There are more advanced sampling methods, e.g. ["Weighted random sampling with a reservoir" by Efraimidis and Spirakis](https://www.sciencedirect.com/science/article/abs/pii/S002001900500298X), but they require floats. I'll stick to integers for now, in case we ever want it back in a contract.

### DIFF
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -727,15 +727,6 @@ class StakingEscrowAgent(EthereumContractAgent):
             staker_address: ChecksumAddress = self.contract.functions.stakers(index).call()
             yield staker_address
 
-    def sample(self,
-               quantity: int,
-               duration: int,
-               pagination_size: Optional[int] = None
-               ) -> List[ChecksumAddress]:
-        reservoir = self.get_stakers_reservoir(duration=duration, pagination_size=pagination_size)
-        return reservoir.draw(quantity)
-
-    @contract_api(CONTRACT_CALL)
     def get_stakers_reservoir(self,
                               duration: int,
                               without: Iterable[ChecksumAddress] = [],
@@ -1622,8 +1613,8 @@ class WeightedSampler:
     Samples random elements with probabilities proportional to given weights.
     """
 
-    def __init__(self, elements: Iterable, weights: Iterable[int]):
-        assert len(elements) == len(weights)
+    def __init__(self, weighted_elements: Dict[Any, int]):
+        elements, weights = zip(*weighted_elements.items())
         self.totals = list(accumulate(weights))
         self.elements = elements
 
@@ -1666,10 +1657,8 @@ class WeightedSampler:
 
 class StakersReservoir:
 
-    def __init__(self, stakers_map):
-        addresses = list(stakers_map.keys())
-        tokens = list(stakers_map.values())
-        self._sampler = WeightedSampler(addresses, tokens)
+    def __init__(self, stakers_map: Dict[ChecksumAddress, int]):
+        self._sampler = WeightedSampler(stakers_map)
         self._rng = random.SystemRandom()
 
     def __len__(self):
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -740,17 +740,22 @@ class StakingEscrowAgent(EthereumContractAgent):
                               duration: int,
                               without: Iterable[ChecksumAddress] = [],
                               pagination_size: Optional[int] = None) -> 'StakersReservoir':
+
         n_tokens, stakers_map = self.get_all_active_stakers(periods=duration,
-                                                     pagination_size=pagination_size)
+                                                            pagination_size=pagination_size)
 
-        self.log.debug(f"Got {len(stakers_map)} stakers with {n_tokens} total tokens")
-
+        filtered_out = 0
         for address in without:
-            del stakers_map[address]
+            if address in stakers_map:
+                n_tokens -= stakers_map[address]
+                del stakers_map[address]
+                filtered_out += 1
 
-        # TODO: or is it enough to just make sure the number of remaining stakers is non-zero?
-        if sum(stakers_map.values()) == 0:
-            raise self.NotEnoughStakers('There are no locked tokens for duration {}.'.format(duration))
+        self.log.debug(f"Got {len(stakers_map)} stakers with {n_tokens} total tokens "
+                       f"({filtered_out} filtered out)")
+
+        if n_tokens == 0:
+            raise self.NotEnoughStakers(f'There are no locked tokens for duration {duration}.')
 
         return StakersReservoir(stakers_map)
 
@@ -1614,7 +1619,7 @@ class ContractAgency:
 
 class WeightedSampler:
     """
-    Samples random elements with probabilities proportioinal to given weights.
+    Samples random elements with probabilities proportional to given weights.
     """
 
     def __init__(self, elements: Iterable, weights: Iterable[int]):
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -15,8 +15,9 @@ You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
+from bisect import bisect_right
+from itertools import accumulate
 import random
-
 import math
 import sys
 from constant_sorrow.constants import (  # type: ignore
@@ -730,15 +731,11 @@ class StakingEscrowAgent(EthereumContractAgent):
     def sample(self,
                quantity: int,
                duration: int,
-               additional_ursulas: float = 1.5,
-               attempts: int = 5,
                pagination_size: Optional[int] = None
                ) -> List[ChecksumAddress]:
         """
         Select n random Stakers, according to their stake distribution.
-
-        The returned addresses are shuffled, so one can request more than needed and
-        throw away those which do not respond.
+        The returned addresses are shuffled.
 
         See full diagram here: https://github.com/nucypher/kms-whitepaper/blob/master/pdf/miners-ruler.pdf
 
@@ -757,42 +754,30 @@ class StakingEscrowAgent(EthereumContractAgent):
         Only stakers which made a commitment to the current period (in the previous period) are used.
         """
 
-        system_random = random.SystemRandom()
         n_tokens, stakers_map = self.get_all_active_stakers(periods=duration, pagination_size=pagination_size)
+
+        # TODO: can be implemented as an iterator if necessary, where the user can
+        # sample addresses one by one without calling get_all_active_stakers() repeatedly.
+
         if n_tokens == 0:
             raise self.NotEnoughStakers('There are no locked tokens for duration {}.'.format(duration))
 
-        sample_size = quantity
-        for _ in range(attempts):
-            sample_size = math.ceil(sample_size * additional_ursulas)
-            points = sorted(system_random.randrange(n_tokens) for _ in range(sample_size))
-            self.log.debug(f"Sampling {sample_size} stakers with random points: {points}")
+        if quantity > len(stakers_map):
+            raise self.NotEnoughStakers(f'Cannot sample {quantity} out of {len(stakers)} total stakers')
 
-            addresses = set()
-            stakers = list(stakers_map.items())
+        addresses = list(stakers_map.keys())
+        tokens = list(stakers_map.values())
+        sampler = WeightedSampler(addresses, tokens)
 
-            point_index = 0
-            sum_of_locked_tokens = 0
-            staker_index = 0
-            stakers_len = len(stakers)
-            while staker_index < stakers_len and point_index < sample_size:
-                current_staker = stakers[staker_index][0]
-                staker_tokens = stakers[staker_index][1]
-                next_sum_value = sum_of_locked_tokens + staker_tokens
+        system_random = random.SystemRandom()
+        sampled_addresses = sampler.sample_no_replacement(system_random, quantity)
 
-                point = points[point_index]
-                if sum_of_locked_tokens <= point < next_sum_value:
-                    addresses.add(to_checksum_address(current_staker))
-                    point_index += 1
-                else:
-                    staker_index += 1
-                    sum_of_locked_tokens = next_sum_value
+        # Randomize the output to avoid the largest stakers always being the first in the list
+        system_random.shuffle(sampled_addresses) # inplace
 
-            self.log.debug(f"Sampled {len(addresses)} stakers: {list(addresses)}")
-            if len(addresses) >= quantity:
-                return system_random.sample(addresses, quantity)
+        self.log.debug(f"Sampled {len(addresses)} stakers: {list(sampled_addresses)}")
 
-        raise self.NotEnoughStakers('Selection failed after {} attempts'.format(attempts))
+        return sampled_addresses
 
     @contract_api(CONTRACT_CALL)
     def get_completed_work(self, bidder_address: ChecksumAddress) -> Work:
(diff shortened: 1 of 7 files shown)


tests/acceptance/blockchain/agents/test_sampling_distribution.py
```diff
@@ -49,7 +51,7 @@ def token(token_economics, deploy_contract):
     return contract
 
 
-@pytest.mark.nightly
+#@pytest.mark.nightly
```
fjarri, line 54: woops, staged accidentally

nucypher/blockchain/eth/agents.py
```diff
@@ -688,41 +684,23 @@ def sample(self,
         Only stakers which made a commitment to the current period (in the previous period) are used.
         """
 
-        system_random = random.SystemRandom()
+        # TODO: can be implemented as an iterator if necessary, where the user can
```
KPrasch, line 687: We need to prioritize bottlenecks of course, but I think this is an excellent idea... not that were ever sampling a ton of addresses at once... or are we?

nucypher/blockchain/eth/agents.py
```diff
@@ -1525,3 +1503,44 @@ def execute(self,
...
+
+    def sample(self, rng, quantity: int, no_replacement: bool=False) -> list:
+        """
+        If ``no_replacement`` is ``False``, samples ``quantity`` of elements from the internal array.
+        Each sample is independent, and the probablity of an element to appear is proportional
+        to the weight provided to the constructor.
+
+        If ``no_replacement`` is ``True``, the elements will not repeat;
+        every time an element is sampled its weight is set to 0.
+        (does not mutate the object and only applies to the current invocation of ``sample()``).
```
cygnusv, line 1526: `no_replacement=False` goes against some core design principles of the network, like that no Ursula should have 2 kfrags for the same policy. So, in order to make our code as misusage-resistant as possible, I think you should simply remove the possibility of sampling with replacement. 
Also, there's the discussion about the use of "negative" names, which sometimes make APIs more complicated.

nucypher/blockchain/eth/agents.py
```diff
@@ -688,41 +684,23 @@ def sample(self,
...
-                    sum_of_locked_tokens = next_sum_value
+        addresses, tokens = zip(*stakers)
+        sampler = WeightedSampler(addresses, tokens)
+        system_random = random.SystemRandom()
 
-            self.log.debug(f"Sampled {len(addresses)} stakers: {list(addresses)}")
-            if len(addresses) >= quantity:
-                return system_random.sample(addresses, quantity)
+        # TODO: shuffle the result? Otherwise the addresses with more weights
+        # are more probable to appear first.
```
cygnusv, line 699: Fantastic idea.

nucypher/blockchain/eth/agents.py
fjarri: Personally, I am a proponent of the guideline that any optional boolean parameter must be designed in a way that its default value is `False`. But I agree that since we don't need sampling with replacement, we shouldn't expose it. Removed the option.

nucypher/blockchain/eth/agents.py
fjarri: Added shuffling.

nucypher/blockchain/eth/agents.py
fjarri: It's not so much about the bottlenecks, but rather about the typical usage scenarios. If the user of `sample()` later discards several of the results for some reason and needs the next best candidates, it is better to make an iterator-like API for it.

KPrasch: IIRC, We're sample a buffered amount of nodes at the moment for higher-level fault handling so this sounds like a good iteration.

This ends up being as a finite generator producing ordered sample results.?

KPrasch:
@fjarri - Can the integration test failure be fixed via rebase over master?

fjarri:
Evidently, yes :)

nucypher/blockchain/eth/agents.py
fjarri: We might as well include it in this PR.

@KPrasch , where does the list of stakers get filtered? Is it just `BlockchainPolicy.__find_ursulas()`?

nucypher/blockchain/eth/agents.py
```diff
@@ -1577,3 +1537,71 @@ def get_agent_by_contract_name(cls,
         agent_class: Type[EthereumContractAgent] = getattr(agents_module, agent_name)
         agent: EthereumContractAgent = cls.get_agent(agent_class=agent_class, registry=registry, provider_uri=provider_uri)
         return agent
+
+
+class WeightedSampler:
+    """
+    Samples random elements with probabilities proportioinal to given weights.
```
mswilkison, line 1544:
```suggestion
    Samples random elements with probabilities proportional to given weights.
```

nucypher/blockchain/eth/agents.py
```diff
@@ -1577,3 +1537,71 @@ def get_agent_by_contract_name(cls,
...
+            samples.append(self.elements[idx])
+
+            # Adjust the totals so that they correspond
+            # to the weight of the element `idx` being set to 0.
+            prev_total = totals[idx - 1] if idx > 0 else 0
+            weight = totals[idx] - prev_total
+            for j in range(idx, len(totals)):
+                totals[j] -= weight
+
+        return samples
```
derekpierre, line 1583: Pretty cool algo.

nucypher/policy/policies.py
```diff
@@ -576,58 +574,69 @@ def generate_policy_parameters(n: int,
...
-            raise self.NotEnoughBlockchainUrsulas(error)
+            # TODO: would be nice to wait for the learner to finish an iteration here,
+            # because if it hasn't, we really have nothing to do.
+            time.sleep(learner_timeout)
+
+            delta = maya.now() - start_time
+            if delta.total_seconds() >= timeout:
+                still_checking = ', '.join(to_check)
+                raise RuntimeError(f"Timed out after {timeout} seconds; "
+                                   f"need {quantity} more, still checking {still_checking}.")
```
derekpierre, line 631: Should this be `quantity_remaining`?

```suggestion
                                   f"need {quantity_remaining} more, still checking {still_checking}.")
```

nucypher/blockchain/eth/agents.py
fjarri: Fixed.

nucypher/blockchain/eth/agents.py
```diff
@@ -657,73 +658,32 @@ def swarm(self) -> Iterable[ChecksumAddress]:
...
+            del stakers_map[address]
 
-        Only stakers which made a commitment to the current period (in the previous period) are used.
-        """
-
-        system_random = random.SystemRandom()
-        n_tokens, stakers_map = self.get_all_active_stakers(periods=duration, pagination_size=pagination_size)
-        if n_tokens == 0:
+        # TODO: or is it enough to just make sure the number of remaining stakers is non-zero?
+        if sum(stakers_map.values()) == 0:
```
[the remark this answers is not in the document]
fjarri: Not sure how much it will affect performance, but it doesn't affect readability, so fixed.

nucypher/blockchain/eth/agents.py
```diff
@@ -657,73 +658,32 @@ def swarm(self) -> Iterable[ChecksumAddress]:
...
+                              without: Iterable[ChecksumAddress] = [],
+                              pagination_size: Optional[int] = None) -> 'StakersReservoir':
+        n_tokens, stakers_map = self.get_all_active_stakers(periods=duration,
+                                                     pagination_size=pagination_size)
 
-        ```
-        Stakes: |----- S0 ----|--------- S1 ---------|-- S2 --|---- S3 ---|-S4-|----- S5 -----|
-        Points: ....R0.......................R1..................R2...............R3...........
-        ```
+        self.log.debug(f"Got {len(stakers_map)} stakers with {n_tokens} total tokens")
```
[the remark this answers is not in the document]
fjarri: That's true, fixed. Also added a check that an address in `without` is actually present in `stakers_map` - I don't think it's guaranteed.

nucypher/policy/policies.py
fjarri: Thanks, fixed.

nucypher/policy/policies.py
```diff
@@ -576,58 +574,69 @@ def generate_policy_parameters(n: int,
...
+        new_to_check = to_check
+
+        while True:
+
+            # Check if the sampled addresses are already known.
+            # If we're lucky, we won't have to wait for the learner iteration to finish.
+            known = list(filter(lambda x: x in self.alice.known_nodes, to_check))
+            to_check = list(filter(lambda x: x not in self.alice.known_nodes, to_check))
 
+            known = known[:min(len(known), quantity_remaining)] # we only need so many
```
cygnusv, line 610: I think this step would produce a bias towards whatever order `known` has, which comes from `alice.known_nodes` and goes all the way down to `FleetStateTracker`, where they are ordered by insertion. Maybe it'd be better to randomly sample the quantity you need.

nucypher/policy/policies.py
```diff
@@ -576,58 +574,69 @@ def generate_policy_parameters(n: int,
...
+
+        # Randomize the output to avoid the largest stakers always being the first in the list
+        system_random = random.SystemRandom()
+        system_random.shuffle(found_ursulas) # inplace
 
-        # Capture the selection and search the network for those Ursulas
-        selected_addresses.update(sampled_addresses)
-        found_ursulas = self.__find_ursulas(sampled_addresses, quantity)
-        return found_ursulas
+        return set(found_ursulas)
```
cygnusv, line 639: If the output is a set, I don't know if the shuffling is necessary as the iteration order over sets is arbitrary.

nucypher/blockchain/eth/agents.py
```diff
@@ -684,73 +685,37 @@ def swarm(self) -> Iterable[ChecksumAddress]:
...
-
-        See full diagram here: https://github.com/nucypher/kms-whitepaper/blob/master/pdf/miners-ruler.pdf
+        reservoir = self.get_stakers_reservoir(duration=duration, pagination_size=pagination_size)
+        return reservoir.draw(quantity)
 
-        This method implements the Probability Proportional to Size (PPS) sampling algorithm.
-        In few words, the algorithm places in a line all active stakes that have locked tokens for
-        at least `duration` periods; a staker is selected if an input point is within its stake.
-        For example:
+    @contract_api(CONTRACT_CALL)
```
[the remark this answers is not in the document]
fjarri: Yes, I think it can be safely removed.

nucypher/blockchain/eth/agents.py
```diff
@@ -1608,3 +1573,71 @@ def get_agent_by_contract_name(cls,
...
         return agent
+
+
+class WeightedSampler:
+    """
+    Samples random elements with probabilities proportional to given weights.
+    """
+
+    def __init__(self, elements: Iterable, weights: Iterable[int]):
+        assert len(elements) == len(weights)
```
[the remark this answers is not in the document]
fjarri: I changed the `WeightedSampler.__init__()` to just take a `dict` instead, that's what `StakersReservoir` uses, anyway.

nucypher/blockchain/eth/agents.py
```diff
@@ -1608,3 +1573,71 @@ def get_agent_by_contract_name(cls,
         agent_class: Type[EthereumContractAgent] = getattr(agents_module, agent_name)
         agent: EthereumContractAgent = cls.get_agent(agent_class=agent_class, registry=registry, provider_uri=provider_uri)
         return agent
+
+
+class WeightedSampler:
```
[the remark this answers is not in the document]
fjarri: Perhaps if it was used in several different places... I'm not opposed to it, but where would it go there? The existing submodules don't suit it.

nucypher/policy/policies.py
```diff
@@ -411,11 +409,12 @@ def make_arrangement(self, ursula: Ursula, *args, **kwargs):
...
 
     @abstractmethod
-    def sample_essential(self, quantity: int, handpicked_ursulas: Set[Ursula] = None) -> Set[Ursula]:
+    def sample_essential(self, quantity: int, handpicked_ursulas: Set[Ursula]) -> Set[Ursula]:
         raise NotImplementedError
 
-    def sample(self, handpicked_ursulas: Set[Ursula] = None) -> Set[Ursula]:
-        selected_ursulas = set(handpicked_ursulas) if handpicked_ursulas else set()
+    def sample(self, handpicked_ursulas: Optional[Set[Ursula]] = None) -> Set[Ursula]:
+        handpicked_ursulas = handpicked_ursulas if handpicked_ursulas else set()
```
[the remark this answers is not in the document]
fjarri: Personally, I prefer to just have `set()` as the default value, but `or` is a no-disadvantage improvement over the current state. Fixed.

nucypher/policy/policies.py
```diff
@@ -576,58 +574,69 @@ def generate_policy_parameters(n: int,
...
 
-    def __find_ursulas(self,
-                       ether_addresses: List[str],
-                       target_quantity: int,
-                       timeout: int = 10) -> set:  # TODO #843: Make timeout configurable
+    def sample_essential(self,
+                         quantity: int,
+                         handpicked_ursulas: Set[Ursula],
+                         learner_timeout: int = 1,
+                         timeout: int = 10) -> Set[Ursula]:
```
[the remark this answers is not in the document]
fjarri: Fixed

nucypher/blockchain/eth/agents.py
```diff
@@ -684,73 +685,37 @@ def swarm(self) -> Iterable[ChecksumAddress]:
             staker_address: ChecksumAddress = self.contract.functions.stakers(index).call()
             yield staker_address
 
-    @contract_api(CONTRACT_CALL)
     def sample(self,
```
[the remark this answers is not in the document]
fjarri: It's used by `AvailabilityTracker.sample()` and `Policy.sample()`, isn't it?

nucypher/policy/policies.py
```diff
@@ -576,58 +574,69 @@ def generate_policy_parameters(n: int,
...
-                ether_address = ether_addresses.pop()
-            else:
-                ether_address = unknown_addresses.popleft()
+        to_check = reservoir.draw(quantity_remaining)
 
-            try:
-                # Check if this is a known node.
-                selected_ursula = self.alice.known_nodes[ether_address]
+        # Sample stakers in a loop and feed them to the learner to check
+        # until we have enough in selected_ursulas`.
```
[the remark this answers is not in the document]
fjarri: Fixed

nucypher/policy/policies.py
```diff
@@ -576,58 +574,69 @@ def generate_policy_parameters(n: int,
...
-                continue
+        start_time = maya.now()
+        new_to_check = to_check
+
+        while True:
+
+            # Check if the sampled addresses are already known.
+            # If we're lucky, we won't have to wait for the learner iteration to finish.
+            known = list(filter(lambda x: x in self.alice.known_nodes, to_check))
+            to_check = list(filter(lambda x: x not in self.alice.known_nodes, to_check))
```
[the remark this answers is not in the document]
fjarri: Actually, we might as well just use the sets throughout. Fixed.

nucypher/policy/policies.py
fjarri: Fixed

nucypher/policy/policies.py
fjarri: Hm, don't sets share the ordered behaviour with dicts since Py3.7?

nucypher/blockchain/eth/agents.py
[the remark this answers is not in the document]
cygnusv: Sounds fair.

nucypher/policy/policies.py
```diff
@@ -411,11 +409,12 @@ def make_arrangement(self, ursula: Ursula, *args, **kwargs):
...
     @abstractmethod
-    def sample_essential(self, quantity: int, handpicked_ursulas: Set[Ursula] = None) -> Set[Ursula]:
+    def sample_essential(self, quantity: int, handpicked_ursulas: Set[Ursula]) -> Set[Ursula]:
         raise NotImplementedError
 
-    def sample(self, handpicked_ursulas: Set[Ursula] = None) -> Set[Ursula]:
-        selected_ursulas = set(handpicked_ursulas) if handpicked_ursulas else set()
+    def sample(self, handpicked_ursulas: Optional[Set[Ursula]] = None) -> Set[Ursula]:
+        handpicked_ursulas = handpicked_ursulas or set()
+        selected_ursulas = set(handpicked_ursulas)
```
derekpierre, line 417: Feels like the previous one-liner was simpler

fjarri: Fair enough, not sure how it ended up this way. Fixed.

PR #2072, opened 2020-06-04
Hide 'commit-next' from help messages

### LINKED ISSUE #1970: Make manual `confirma-activity` available only for debug
Or remove support for `nucypher ursula confirm-activity`

vzotova:
Fixes #1970

### DIFF
--- a/tests/acceptance/cli/ursula/test_ursula_command.py
+++ b/tests/acceptance/cli/ursula/test_ursula_command.py
@@ -74,10 +74,28 @@ def test_ursula_command_help(protocol, ursula):
     with capture_output() as (out, err):
         protocol.lineReceived(line=b'bananas')
 
+    commands = protocol.commands
+    hidden_commands = ['?', 'commit_next']
+    commands = list(set(commands) - set(hidden_commands))
+
     # Ensure all commands are in the help text
     result = out.getvalue()
-    for command in protocol.commands:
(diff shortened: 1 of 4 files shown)


KPrasch:
This raises a larger question: What role does the Ursula REPL play in nucypher?  

It was *once* an integral part of my workflow, but now I dont think we want to encourage running a worker as a non-service.  

I can imagine keeping this feature around for debugging situations (think get, IPFS console, etc.) however presenting it preeminently has been a source of UX confusion.

tests/acceptance/cli/ursula/test_ursula_command.py
```diff
@@ -74,10 +74,28 @@ def write(data: bytes):
     with capture_output() as (out, err):
         protocol.lineReceived(line=b'bananas')
 
+    commands = protocol.commands
+    hidden_commands = ['?', 'commit_next']
```
fjarri, line 78: Shouldn't we rather check for `commit-next` (with a hyphen)?

vzotova: there were underscores in interactive console for commands, so I just did the same (in CLI still hyphen)

fjarri: Ah, I see, I thought it was CLI.

PR #2073, opened 2020-06-05
(Optionally) Run Unit Tests as pre-push hook

KPrasch:
#### What this does

- Fully deprecates --runslow marker
- Uses two separate pytest.ini configurations: one for circleCI and one for more typical CLI usage.
- Relocates the slowest unit test into integrations tests, since it also hits the disk
- Run unit tests on pre-push with additional setup (`pre-commit install && pre-commit install -t pre-push`)
- Holdover for sudden rough economics model very small error
- Adjust `make install` to also remove any existing pipenv


nucypher/crypto/powers.py
```diff
@@ -108,7 +108,7 @@ class AccountLocked(PowerUpError):
         """Raised when signing cannot be performed due to a locked account"""
         pass
 
-    @validate_checksum_address
+    # @validate_checksum_address  # TODO: Workaround when `account is NO_WORKER_BONDED` raising invalid checksum address.
```
derekpierre, line 111: Interesting 🤔  - what's cas when a TransactionPower is attempted to be instantiated without a valid account address?

KPrasch: When `account is NO_WORKER_BONDED` which is somewhat of a corner case.

derekpierre: I was wondering more along the lines of, why not prevent TransactionPower from being initialized in that case? - initializing it with the account not being set to an actual address does not make sense to me.

KPrasch: Unfortunately, I do not have a solid answer for you at this time.  This test suddenly started failing for this seemingly unrelated reason, so commenting out this line and leaving a comment is in an effort to make the smallest change possible while for PR precision.

To address your suggestion - I agree - There can be some higher-level handling preventing the `TransactingPower` from being set if there is  no address.  One issue that needs to faced is that `NO_WORKER_ADDRESS` may be returned from `Ursula.worker_address` which results in mixed signaling and property logic.

KPrasch: There are still two failing tests caused by this - open to suggestions on how to get CI to accept the pre-push changes despite this diversion.

nucypher/blockchain/eth/actors.py
```diff
@@ -1112,7 +1112,7 @@ def worker_address(self) -> str:
             self.__worker_address = worker_address
 
         if self.__worker_address == NULL_ADDRESS:
-            return NO_WORKER_BONDED.bool_value(False)
+            return NO_WORKER_BONDED(NULL_ADDRESS)
```
cygnusv, line 1115: What's the benefit of having a CS constant with an explicit string value, over using that same string as a constant (even with a different name)? Taking into account that CS constants are not IDE-friendly, which is a UX friction point for readers of the codebase and potential external collaborators.

KPrasch: I must point you to the above dialog.  I am not targeting this part of the code - just hacking around the failures here.

tests/unit/test_blockchain_economics_model.py
```diff
@@ -47,22 +47,26 @@ def test_rough_economics():
...
 
     # Sanity check that total and reward supply calculated correctly
     assert int(LOG2 / (e.token_halving * 365) * (e.erc20_total_supply - int(e.first_phase_total_supply))) == int(initial_rate)
     assert int(e.reward_supply) == int(e.erc20_total_supply - Decimal(int(1e9)))
 
-    # Sanity check for lock_duration_coefficient_1 (k1), issuance_decay_coefficient (d) and lock_duration_coefficient_2 (k2)
-    assert e.lock_duration_coefficient_1 * e.token_halving == \
-           e.issuance_decay_coefficient * e.lock_duration_coefficient_2 * LOG2 * e.small_stake_multiplier / 365
-
+    with localcontext() as ctx:  # TODO: Needs follow up - why the sudden failure (python 3.8.0)?
```
KPrasch, line 64: Preemptively - I'd like to indicate this addition also... 

This is a badly behaving PR!  No Idea why this started to fail suddenly.  (very small error)

nucypher/crypto/powers.py
derekpierre: Are the two failing tests the following?
1. tests/acceptance/cli/ursula/test_stakeholder_and_ursula.py::test_stake_unbond_worker
2. tests/acceptance/blockchain/actors/test_staker.py::test_staker_manages_winding_down 

For 1) it may just be a "Falsey" issue
```
E       AssertionError: assert not NO_WORKER_BONDED (0x0000000000000000000000000000000000000000)
 
E        +  where NO_WORKER_BONDED (0x0000000000000000000000000000000000000000) = Staker(address='0xaaa23A5c74aBA6ca5E7c09337d5317A7C4563075').worker_address
```

 I added the following  to the test:
```
print(f">>>> Derek {staker.worker_address}")
```
and obtained `>>>> Derek 0x0000000000000000000000000000000000000000`

For 2) it seems that the test assumes that the staker already has a bonded worker, but I don't think it does. You may have to add something like the following before using `staker.worker_address`
```
    worker_address = testerchain.unassigned_accounts[-1] <--- I saw this in other tests
    staker.bond_worker(worker_address=worker_address)
```

KPrasch: For an accurate representation of this draft - do not consider the last commit (which is me hacking and slashing the other day looking for a solution.)

I dropped 2e72d373592708b666326b3f634aa5e33e39ebc3. Take a peek at CI afterwards to get the same initial impression I had after finished the pre-push changes.

KPrasch: Okay - this is a better starting place for examination: <https://app.circleci.com/pipelines/github/nucypher/nucypher/5057/workflows/0108b710-9cf0-4587-afc8-27136231029d/jobs/91189>

nucypher/blockchain/eth/actors.py
KPrasch: :heavy_check_mark: - removed

nucypher/crypto/powers.py
KPrasch: :heavy_check_mark:

KPrasch:
Rebased @ 5ed97e6d44c125840977b69d9e78ad64484b0345

KPrasch:
Rebased @ 76eea579a3b94e1ea2a8019601e0b454f81e5e2c

KPrasch:
Rebased @ ae371611ed7e456f69e4ce7b5a8a68e111076251

KPrasch:
rebased @ 8d092d90df62e6c0dc3a8ca34d223f474a923507

derekpierre:
Re-iterating @mswilkison 's comment regarding whether should we add pre-commit installation and set up to the contributing guide as part of this PR?

KPrasch:
Responded to RFCs - added a new section to `CONTRIBUTING.rst`

docs/source/guides/contribution_guide.rst
```diff
@@ -80,6 +80,36 @@ Optionally, to run the full, slow, verbose test suite run:
 
   (nucypher)$ pytest
 
+Setup Commit & Push Hooks
+--------------------------
+
+Pre-commit and pre-push are used for quality control by preventing a commit that will fail
+if passed along to CI servers or by making small formatting changes directly to source code files.
```
derekpierre, line 87:
```suggestion
`Pre-commit <https://pre-commit.com/>`_ and pre-push are used for quality control to identify and prevent the inclusion of problematic code changes. It may prevent a commit that will fail
if passed along to CI servers or make small formatting changes directly to source code files.
```

derekpierre: ^ Updated "It" to "They"

KPrasch:
Rebased over main @ 7afa1d5

pytest.ini
```diff
@@ -1,5 +1,2 @@
 [pytest]
-addopts = -s -v --runslow --junitxml=./reports/pytest-results.xml --strict-markers --durations=0 --cov=nucypher --cov-report xml:reports/coverage.xml
-markers =
-    slow: marks tests as slow (skipped by default, use '--runslow' to include these tests)
-junit_family = xunit1
+addopts = --capture fd --maxfail 1
```
cygnusv, line 2: What does these options mean?

requirements.txt
```diff
@@ -1,23 +1,23 @@
...
-cffi==1.14.0
+cffi==1.14.1
 chardet==3.0.4
 click==7.1.2
 coincurve==13.0.0
 colorama==0.4.3
 constant-sorrow==0.1.0a9
 constantly==15.1.0
-cryptography==2.9.2
+cryptography==3.0
```
cygnusv, line 20: Uhh, maybe we should check what's going on in this new version (pinging @tuxxy )

cygnusv (approved):

requirements.txt
KPrasch: Ohhh - good eye!

tuxxy (approved):

requirements.txt
tuxxy: Looks fine to me.

pytest.ini
KPrasch: Feel free to alter these on your system to tweak the way pytest runs:

 `--capture fd` means *take stdout and 1 and 2 and redirect to a temp file (<https://docs.pytest.org/en/stable/capture.html#setting-capturing-methods-or-disabling-capturing>).  This is important if you want to see the console output of CLI tests while still being able to hit a debugging breakpoint.  The shorthand is `-s`.

`--maxfail 1` means - *Fail and exit pytest after 1 failure.*

Issue #2068, opened 2020-06-03
Improve "Transaction Finality" logic


cygnusv:
As we approach a new level of maturity in our codebase, we can focus on certain areas that need more care. One of them is our Ethereum Clients logic, and more specifically, our logic for "Transaction Finality" (i.e., when we can consider to a reasonable degree, that a TX we sent is included in a block). The problem of TX finality can manifest in several ways, but usually it comes as a "`Nonce too low`" error.

A holistic solution to this problem must consider too the requirement of a configurable number of block confirmations, and a correct and optimized integration with web3's middlewares. 
At the same time, we must accept there's no perfect solution here that doesn't involve waiting an unreasonable amount of time. 

I've been thinking that a simple strategy that starts by waiting for some time and then performing additional and potentially optional checks for block confirmations and chain re-orgs, may be a good start, and that we can iterate from there to achieve a more resilient design. Related to this, this comment from the Infura docs is very interesting:
> Cache Ethereum data locally. Barring exceptionally rare deep reorganizations of the chain, **blocks more than a couple blocks below the head of the chain can be cached indefinitely**. Ask for the data once then keep it locally.

Following this advice mean that simply waiting for, let's say, 30 seconds should be more than enough. This can be particularly useful during deployment, since with our current logic, recovering from a TX failure can be very difficult. For other use cases in our codebase, like Alice or Ursula TXs, which are not that critical in case of failure, waiting for 30 seconds seem excessive, so we could reduce it to some seconds. There's room for a lot of discussion yet.

Other related issues, that should be fixed by the solution to this issue:
* #2002 
* #2067 
* #2030
* #1433

cygnusv:
#2043 improved the situation a lot. Currently, the best way to "ensure" finality of transactions is to use the block confirmations logic, which deals with the problems described in the issue.

cygnusv:
Not as relevant now that there are much less TXs to track.

PR #2074, opened 2020-06-05
[Concept] Nonce stability testing

KPrasch:


KPrasch:
Related to #2068

KPrasch:
Unsure what to do with this PR for now.  It demonstrates a potential issue with nonce reuse in extreme conditions but a solution is not included.  I forsee one of two outcomes here: Modify the test so that it's parameters are just under the failure threshold, or mark as *wontfix.*

KPrasch:
I opened this PR to demonstrate nonce stability issues in tests - but there is no real solution... Test creates relatively extreme transaction frequency.  It's unclear how it can be implemented to give meaningful test output at this time.

PR #2078, opened 2020-06-07
Fixes #1909 - Use a decorator to skip tests on CI

### LINKED ISSUE #1909: Decorator to skip circleCI
Current we do the following to skip a test on CI:

```
    if 'CIRCLECI' in os.environ:
        pytest.skip("Do not run Geth nodes in CI")
```

This can be implemented as a decorator instead.

KPrasch:


tests/markers.py
```diff
@@ -19,5 +19,5 @@
 import pytest
 
 
-circleci_only = pytest.mark.skipif(condition=('CIRCLECI' not in os.environ),
-                                   reason='Only run on CircleCI')
+circleci_only = pytest.mark.skipif(condition=('CIRCLECI' not in os.environ), reason='Only run on CircleCI')
+skip_circleci = pytest.mark.skipif(condition=('CIRCLECI' in os.environ), reason='Only run on CircleCI')
```
fjarri, line 23: `skip_on_circleci` would sound a little better to me, but I don't insist on it.

KPrasch: :heavy_check_mark: - Yes, this is more clear.

PR #2086, opened 2020-06-09
Removes TestPyPI as part of CD workflow

KPrasch:


.circleci/config.yml
```diff
@@ -741,40 +722,6 @@ jobs:
...
-      - run:
-          name: Build Python Wheel (Test Deploy)
-          command: make dist
-      - run:
-          name: upload to TestPyPI
-          command: python3 -m twine upload --repository testpypi dist/* --verbose
-      - run:
-          name: pip install from TestPyPI
-          command: |
-            pip3 install --user --index-url https://test.pypi.org/simple/ --extra-index-url https://pypi.org/simple nucypher --verbose
```
KPrasch, line 775: As it turns out - this has been doing the wrong thing for a long time - installing an unspecified version from test PyPI results in the *first ever upload*.  From a high-level this test is superfluous since we can smoke-test the distribution without uploading to a remote index server.

cygnusv: What a coincidence. I've never seen this job fail and I just experienced now doing a new release:
https://app.circleci.com/pipelines/github/nucypher/nucypher/5062/workflows/824f2879-1250-4945-980a-9caea482425d/jobs/91277

PR #2098, opened 2020-06-16
Simple Enrico file encryption

KPrasch:
Bare minimum support of enrico encrypting files via CLI and sending them to an IPFS gateway in time for an active hackathon.

- Encryption of files via `nucypher enrico encrypt --file <PATH> ...`
- Simple upload of message kits to ipfs with `nucypher enrico --ipfs <GATEWAY>`
- Simple retrieval of message kits from IPFS ipfs via CID `nucypher bob retireve --ipfs <GATEWAY> ... --message-kit <CID>`

### DIFF
--- a/nucypher/characters/control/controllers.py
+++ b/nucypher/characters/control/controllers.py
@@ -141,7 +141,8 @@ class CLIController(CharacterControlServer):
     def handle_request(self, method_name, request):
         start = maya.now()
         response = self._perform_action(action=method_name, request=request)
-        return self.emitter.ipc(response=response, request_id=start.epoch, duration=maya.now() - start)
+        self.emitter.ipc(response=response, request_id=start.epoch, duration=maya.now() - start)
+        return response
 
 
 class JSONRPCController(CharacterControlServer):
--- a/nucypher/characters/control/controllers.py
+++ b/nucypher/characters/control/controllers.py
@@ -60,7 +60,6 @@ class CharacterControllerBase(ABC):
         method = getattr(self.interface, action, None)
         serializer = method._schema
         params = serializer.load(request) # input validation will occur here.
-
         response = method(**params)  # < ---- INLET
 
         response_data = serializer.dump(response)
--- a/nucypher/characters/control/controllers.py
+++ b/nucypher/characters/control/controllers.py
@@ -138,7 +138,7 @@ class CLIController(CharacterControlServer):
     def test_client(self):
         return
 
-    def handle_request(self, method_name, request):
+    def handle_request(self, method_name, request) -> dict:
         start = maya.now()
         response = self._perform_action(action=method_name, request=request)
         self.emitter.ipc(response=response, request_id=start.epoch, duration=maya.now() - start)
--- a/nucypher/characters/control/interfaces.py
+++ b/nucypher/characters/control/interfaces.py
@@ -138,7 +138,7 @@ class AliceInterface(CharacterPublicInterface):
         if len(failed_revocations) > 0:
             for node_id, attempt in failed_revocations.items():
                 revocation, fail_reason = attempt
-                if fail_reason == NotFound:
+                if fail_reason == NotFound:  # FIXME
                     del (failed_revocations[node_id])
         if len(failed_revocations) <= (policy.n - policy.treasure_map.m + 1):
             del (self.character.active_policies[policy_id])
@@ -241,11 +241,12 @@ class BobInterface(CharacterPublicInterface):
 class EnricoInterface(CharacterPublicInterface):
 
     @attach_schema(enrico.EncryptMessage)
-    def encrypt_message(self, message: str):
+    def encrypt_message(self, plaintext: Union[str, bytes]):
         """
         Character control endpoint for encrypting data for a policy and
         receiving the messagekit (and signature) to give to Bob.
         """
-        message_kit, signature = self.character.encrypt_message(bytes(message, encoding='utf-8'))
+        plaintext = bytes(plaintext, encoding='utf-8')
+        message_kit, signature = self.character.encrypt_message(plaintext=plaintext)
         response_data = {'message_kit': message_kit, 'signature': signature}
         return response_data
--- a/nucypher/characters/control/interfaces.py
+++ b/nucypher/characters/control/interfaces.py
@@ -24,6 +24,7 @@ from nucypher.characters.control.specifications import alice, bob, enrico
 from nucypher.crypto.kits import UmbralMessageKit
 from nucypher.crypto.powers import DecryptingPower, SigningPower
 from nucypher.crypto.utils import construct_policy_id
+from nucypher.datastore.datastore import NotFound
 
 
 def attach_schema(schema):
@@ -138,7 +139,7 @@ class AliceInterface(CharacterPublicInterface):
         if len(failed_revocations) > 0:
             for node_id, attempt in failed_revocations.items():
                 revocation, fail_reason = attempt
-                if fail_reason == NotFound:  # FIXME
+                if fail_reason == NotFound:
                     del (failed_revocations[node_id])
         if len(failed_revocations) <= (policy.n - policy.treasure_map.m + 1):
             del (self.character.active_policies[policy_id])
--- a/nucypher/characters/control/interfaces.py
+++ b/nucypher/characters/control/interfaces.py
@@ -246,7 +246,6 @@ class EnricoInterface(CharacterPublicInterface):
         Character control endpoint for encrypting data for a policy and
         receiving the messagekit (and signature) to give to Bob.
         """
-        plaintext = bytes(plaintext, encoding='utf-8')
         message_kit, signature = self.character.encrypt_message(plaintext=plaintext)
         response_data = {'message_kit': message_kit, 'signature': signature}
         return response_data
--- a/nucypher/characters/control/specifications/enrico.py
+++ b/nucypher/characters/control/specifications/enrico.py
@@ -64,4 +64,4 @@ class EncryptMessage(BaseSchema):
 
     # output
     message_kit = fields.UmbralMessageKit(dump_only=True)
-    signature = fields.UmbralSignature(dump_only=True)  # maybe we need a signature field?
+    signature = fields.UmbralSignature(dump_only=True)
--- a/nucypher/characters/control/specifications/enrico.py
+++ b/nucypher/characters/control/specifications/enrico.py
@@ -62,7 +62,6 @@ class EncryptMessage(BaseSchema):
 
         return {"plaintext": data}
 
-
     # output
     message_kit = fields.UmbralMessageKit(dump_only=True)
-    signature = fields.String(dump_only=True)  # maybe we need a signature field?
+    signature = fields.UmbralSignature(dump_only=True)  # maybe we need a signature field?
--- a/nucypher/characters/control/specifications/enrico.py
+++ b/nucypher/characters/control/specifications/enrico.py
@@ -28,7 +28,7 @@ class EncryptMessage(BaseSchema):
 
     # input
     message = fields.Cleartext(
-        load_only=False,
+        load_only=True,
         allow_none=True,
         click=click.option('--message', help="A unicode message to encrypt for a policy")
     )
--- a/nucypher/characters/control/specifications/enrico.py
+++ b/nucypher/characters/control/specifications/enrico.py
@@ -52,13 +52,13 @@ class EncryptMessage(BaseSchema):
         we output one of them as the "plaintext" arg to enrico.encrypt_message
         """
 
-        if data['message'] and data['filepath']:
+        if data.get('message') and data.get('filepath'):
             raise exceptions.InvalidArgumentCombo("choose either a message or a filepath but not both.")
 
         if data.get('message'):
-            data = bytes(data.get('message'), encoding='utf-8')
+            data = bytes(data['message'], encoding='utf-8')
         else:
-            data = data.get('filepath')
+            data = data['filepath']
 
         return {"plaintext": data}
 
--- a/nucypher/characters/control/specifications/enrico.py
+++ b/nucypher/characters/control/specifications/enrico.py
@@ -18,23 +18,32 @@
 import click
 
 from nucypher.characters.control.specifications import fields
-from nucypher.characters.control.specifications.base import BaseSchema
 from nucypher.cli import options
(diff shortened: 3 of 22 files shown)


vepkenez:
@KPrasch seems like the remaining failing test is unrelated to our changes...

nucypher/characters/control/controllers.py
```diff
@@ -142,7 +141,8 @@ def test_client(self):
     def handle_request(self, method_name, request):
```
derekpierre, line 141:
```suggestion
    def handle_request(self, method_name, request) -> dict:
```

nucypher/cli/commands/bob.py
```diff
@@ -286,13 +288,15 @@ def public_keys(general_config, character_options, config_file):
 @option_config_file
 @BobInterface.connect_cli('retrieve')
 @group_general_config
+@click.option('--ipfs', help="Download an encrypted message from IPFS at the specified gateway URI")
```
derekpierre, line 291: Should we generalize this option to some storage gateway uri (i.e. allow for extension to other storage technologies eg. filecoin, swarm etc.) with `ipfs://` as a specific scheme that is implemented?

Similar comment for other cases of `--ipfs` option.

nucypher/utilities/prometheus/metrics.py
```diff
@@ -36,9 +36,12 @@
...
 
-from prometheus_client.core import Timestamp
-from prometheus_client.registry import CollectorRegistry, REGISTRY
-from prometheus_client.utils import floatToGoString
+try:
+    from prometheus_client.core import Timestamp
+    from prometheus_client.registry import CollectorRegistry, REGISTRY
+    from prometheus_client.utils import floatToGoString
+except ImportError:
+    raise DevelopmentInstallationRequired(importable_name='prometheus_client')
```
derekpierre, line 44: @KPrasch this was modified in #2100 based on our discussion.

nucypher/characters/control/specifications/enrico.py
```diff
@@ -16,25 +16,52 @@
...
+            data = bytes(data['message'], encoding='utf-8')
+        else:
+            data = data['filepath']
+
+        return {"plaintext": data}
 
     # output
     message_kit = fields.UmbralMessageKit(dump_only=True)
-    signature = fields.String(dump_only=True) # maybe we need a signature field?
+    signature = fields.UmbralSignature(dump_only=True)  # maybe we need a signature field?
```
derekpierre, line 67: You wanted it - you got it!

```suggestion
    signature = fields.UmbralSignature(dump_only=True)
```

KPrasch, line 67: This TODO can be removed

nucypher/characters/control/specifications/fields/signature.py
```diff
@@ -0,0 +1,44 @@
...
+class UmbralSignature(BaseField, fields.Field):
+
+    def _serialize(self, value: Signature, attr, obj, **kwargs):
+        return b64encode(bytes(value)).decode()
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, bytes):
+            return value
+        try:
+            return b64decode(value)
```
derekpierre, line 36: are we returning bytes instead of `Signature`?

nucypher/characters/control/specifications/fields/file.py
```diff
@@ -0,0 +1,33 @@
...
+
+from nucypher.characters.control.specifications.exceptions import InvalidInputData, InvalidNativeDataTypes
+from nucypher.characters.control.specifications.fields.base import BaseField
+
+
+class FileField(BaseField, fields.String):
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        with open(value, 'rb') as plaintext_file:
+            plaintext = plaintext_file.read()
```
derekpierre, line 29: do we need to be concerned about file size here i.e. what about a really large file? Would something buffered be more prudent?

nucypher/characters/control/specifications/enrico.py
```diff
@@ -16,25 +16,52 @@
...
+        we output one of them as the "plaintext" arg to enrico.encrypt_message
+        """
+
+        if data.get('message') and data.get('filepath'):
+            raise exceptions.InvalidArgumentCombo("choose either a message or a filepath but not both.")
+
+        if data.get('message'):
+            data = bytes(data['message'], encoding='utf-8')
+        else:
+            data = data['filepath']
```
derekpierre, line 61: I assume that this is the file data and not the string path of the file...?

nucypher/utilities/prometheus/metrics.py
KPrasch: :+1: - Divergence needs a rebase here

nucypher/characters/control/specifications/enrico.py
vepkenez: yes at that point, the filepath field's serialization method has read the data from the file.

nucypher/characters/control/specifications/fields/signature.py
vepkenez: yeah, I think we should be returning a signature here.

nucypher/characters/control/specifications/fields/file.py
vepkenez: It's going to have to be in memory eventually... not sure.

vepkenez: oh, unless we want to put each buffer in its own message kit and output them individually?

vepkenez: but yeah, we could put a size check into the validation of the FileField

nucypher/cli/commands/bob.py
KPrasch: Interesting idea, perhaps this can be combined with the `--message-kit`?

derekpierre: Interested to better understand what you mean, but I would caution that to remain storage agnostic, it may be best not to mix the message kit with the underlying storage medium.

nucypher/characters/control/specifications/fields/file.py
derekpierre: > oh, unless we want to put each buffer in its own message kit and output them individually?

The underlying pyumbral `pre.encrypt(...)` takes bytes, so something like chunking the file data and each chunk producing its own message kit would be a solution. We did chat some time ago about Enrico being able to handle a stream of data.

nucypher/characters/control/specifications/enrico.py
derekpierre: I figured, it's just that the key `filepath` through me off.

nucypher/cli/commands/bob.py
KPrasch: You are close to my general sentiment on this PR...  Mostly advancing the requirements placed by hackathon participants.

nucypher/characters/control/specifications/enrico.py
KPrasch: oh - it's not really the filepath?

nucypher/characters/control/specifications/fields/file.py
KPrasch: Hmm, time to rename this PR ;-).  Le't kick out the large file handling to a follow up PR.

nucypher/utilities/prometheus/metrics.py
KPrasch: :heavy_check_mark:

nucypher/characters/control/specifications/fields/signature.py
KPrasch: :heavy_check_mark:

nucypher/characters/control/specifications/enrico.py
vepkenez: That method basically allows you to customize the output of the serializer.  It is run after all of the fields process the input data in their own ways.  The filepath field deserializes its input by reading the file data so that’s what is present in that key right there.

derekpierre: Minor nit pick but is it possible to just change the key to be `file` instead of `filepath`? Just to make it clearer.

nucypher/characters/control/specifications/fields/signature.py
```diff
@@ -0,0 +1,44 @@
...
+class UmbralSignature(BaseField, fields.Field):
+
+    def _serialize(self, value: Signature, attr, obj, **kwargs):
+        return b64encode(bytes(value)).decode()
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if isinstance(value, bytes):
+            return value
+        try:
+            return Signature.from_bytes(b64decode(value))
```
derekpierre, line 36: I noticed that from bytes has other parameters:
```
    def from_bytes(cls,
                   signature_as_bytes: bytes,
                   der_encoded: bool = False,
                   curve: Optional[Curve] = None) -> 'Signature':
```

Are none default values for the other parameters something we have to worry about? (@cygnusv )

nucypher/cli/commands/bob.py
derekpierre: What do you think is the resolution here?

nucypher/characters/control/specifications/fields/signature.py
cygnusv: No, NuCypher for the moment always uses the same curve.

nucypher/cli/commands/bob.py
KPrasch: Phew - that's asking for alot of vision!

derekpierre: Haha - I just mean, are we keeping `--ipfs` and then doing a follow-up PR later for something quick, or should we just make something general (eg. `--getway-uri`) instead and have ipfs as the only thing supported for now...

KPrasch: This is so minimal it even deserves the proverbial `--yolo` flag :sweat_smile:

nucypher/characters/control/specifications/enrico.py
vepkenez: @KPrasch you cool with that?  I am.  Lets do it.

nucypher/characters/lawful.py
```diff
@@ -1550,11 +1550,9 @@ def __init__(self, policy_encrypting_key=None, controller: bool = True, *args, *
         self.log = Logger(f'{self.__class__.__name__}-{bytes(self.public_keys(SigningPower)).hex()[:6]}')
         self.log.info(self.banner.format(policy_encrypting_key))
 
-    def encrypt_message(self,
-                        message: bytes
-                        ) -> Tuple[UmbralMessageKit, Signature]:
+    def encrypt_message(self, plaintext: bytes) -> Tuple[UmbralMessageKit, Signature]:
```
cygnusv, line 1553: Maybe this method should be renamed to simply `encrypt`. Not necessarily in this PR, but I think it makes sense.

nucypher/characters/control/controllers.py
```diff
@@ -139,10 +138,11 @@ def make_control_transport(self):
     def test_client(self):
         return
 
-    def handle_request(self, method_name, request):
+    def handle_request(self, method_name, request) -> dict:
         start = maya.now()
         response = self._perform_action(action=method_name, request=request)
-        return self.emitter.ipc(response=response, request_id=start.epoch, duration=maya.now() - start)
+        self.emitter.ipc(response=response, request_id=start.epoch, duration=maya.now() - start)
+        return response
```
cygnusv, line 145: Good catch!

nucypher/characters/control/interfaces.py
```diff
@@ -138,7 +138,7 @@ def revoke(self, label: bytes, bob_verifying_key: bytes) -> dict:
         if len(failed_revocations) > 0:
             for node_id, attempt in failed_revocations.items():
                 revocation, fail_reason = attempt
-                if fail_reason == NotFound:
+                if fail_reason == NotFound:  # FIXME
```
cygnusv, line 141: What does need a fix here? Should we open an issue or add a TODO?

nucypher/characters/control/specifications/enrico.py
cygnusv: I agree. `filepath` is confusing.

nucypher/characters/control/specifications/fields/file.py
cygnusv: Needs tracking this with a TODO/issue

KPrasch: #2106

nucypher/characters/control/specifications/enrico.py
KPrasch: :heavy_check_mark:

nucypher/characters/control/interfaces.py
KPrasch: :heavy_check_mark:

nucypher/characters/lawful.py
KPrasch: #2107

nucypher/cli/commands/bob.py
KPrasch: #2108

nucypher/characters/control/specifications/fields/file.py
```diff
@@ -0,0 +1,33 @@
...
+
+from nucypher.characters.control.specifications.exceptions import InvalidInputData, InvalidNativeDataTypes
+from nucypher.characters.control.specifications.fields.base import BaseField
+
+
+class FileField(BaseField, fields.String):
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        with open(value, 'rb') as plaintext_file:
+            plaintext = plaintext_file.read()  # TODO: #2106 Handle large files
```
KPrasch, line 29: :heavy_check_mark:

fjarri, line 29: So we deserialize a file by reading it from disk? I'm not sure which contexts it will be used in, but it just seems a little... error-prone? Even if it's the same machine, what if the file is missing?

KPrasch: For some context read the `specifications` package, This is part of the EnricoEncrypt specification used in the CLI, and in web controllers, and is mapped from `--file`.  Types are validated and enforced prior to this step for missing or invalid files.

Issue #2088, opened 2020-06-11
KFrag bytes are not persisted in ursula.db (sqlite)


cygnusv:
**Describe the Bug**

Our latest live tests show the following error:
https://gist.github.com/KPrasch/5914392dd1eeac25d48ea6f7cb062859

After manual inspection, we detected that KFrag bytes are `null` in the `policyarrangements` table of the database, which means that there's a bug in our storage logic. Note that other columns of the table are correctly stored (e.g, `id`, `expiration`), but `kfrag` and `alice_signature` are `null`.

There's a recent changeset (0a2dd3a2090a223cbc4411cff903f9db081efe95)  on that part of the codebase, but at first sight I don't see where's the error. 

**To Reproduce**
Just trying to retrieve.

**Screenshot**
https://gist.github.com/KPrasch/5914392dd1eeac25d48ea6f7cb062859

![image](https://user-images.githubusercontent.com/2564234/84423992-b3e03380-ac1f-11ea-9373-3a7131be6cab.png)

**System (please complete the following information):**
 - OS Version:
 - Nucypher Version:
 - Ethereum Node Version:

vepkenez:
issue must be between this line:
> https://github.com/nucypher/nucypher/blob/master/nucypher/network/server.py#L242
and this line:
> https://github.com/nucypher/nucypher/blob/master/nucypher/network/server.py#L298

KPrasch:
Closed in #2099

PR #2099, opened 2020-06-17
Datastore Roundup

tuxxy:
### What this does:
1. Abolishes SQLite as a datastore for Ursula.
2. Implements a datastore on top of lmdb.
3. Lots of changes to the `datastore` module.
    - Adds a `DatastoreRecord` base class that enables for ORM-like typing (but not all the way)
    - Builds a `Datastore` class that has a `describe` method for performing CRUD operations, and a `query_by` method for performing queries.
    - `query_by` accepts a `filter_field` and `filter_func` param for an elegant and minimalist method of performing powerful queries.
5. Extensive tests-as-documentation.
6. Uses the new `Datastore.describe` and `Datastore.query_by` methods as required for Ursula.

### For reviewers
I strongly suggest that you read through the tests _before_ (or at the same time) looking at `Datastore` or `DatastoreRecord` logic. It will immediately clarify design decisions and you'll see why I made the decisions I did when building this. The tests also serve as great documentation for the `Datastore`, as well.

I assure you this isn't a difficult PR to review. The actual datastore changes are relatively small (despite what the line changes report). Most of the changes are Pipfile/reqs and test additions.


nucypher/datastore/datastore.py
```diff
@@ -180,7 +125,7 @@ def attach_kfrag_to_saved_arrangement(self, alice, id_as_hex, kfrag, session=Non
         if policy_arrangement is None:
             raise NotFound("Can't attach a kfrag to non-existent Arrangement {}".format(id_as_hex))
 
-        if policy_arrangement.alice_verifying_key.key_data != alice.stamp:
+        if policy_arrangement.alice_verifying_key != alice.stamp:
             raise alice.SuspiciousActivity
 
         policy_arrangement.kfrag = bytes(kfrag)
```
KPrasch, line 131: This record is not persisted correctly, see #2088.

nucypher/datastore/db/models.py
```diff
@@ -14,18 +14,21 @@
...
 from nucypher.crypto.utils import fingerprint_from_key
 from nucypher.datastore.db import Base
 
 
-FieldEncoder = NamedTuple('FieldEncoder', [('encode', Callable),
-                                           ('decode', Callable)])
+class RecordField(NamedTuple):
+    field_type: Any
+    encode: Optional[Callable] = lambda field: field
+    decode: Optional[Callable] = lambda field: field
```
KPrasch: Why set a default at all if it's an *empty callable*?

nucypher/datastore/db/models.py
```diff
@@ -14,18 +14,21 @@
...
-from typing import Callable, NamedTuple
+from typing import Any, Callable, NamedTuple, Optional, Union
 
 from nucypher.crypto.utils import fingerprint_from_key
 from nucypher.datastore.db import Base
 
 
-FieldEncoder = NamedTuple('FieldEncoder', [('encode', Callable),
-                                           ('decode', Callable)])
+class RecordField(NamedTuple):
```
KPrasch: :+1:

nucypher/datastore/db/models.py
```diff
@@ -34,71 +37,84 @@ class DatastoreRecord(Mapping):
...
-    def __getattr__(self, attr):
-        return self.__data[attr]
+    def __setattr__(self, attr, value):
+        try:
+            field = getattr(self, attr)
+            if not type(value) == field.field_type:
+                raise TypeError(f'The {attr} attr can only be set to a {field.field_type}.')
+        except AttributeError:
+            # TODO: Is there a better way to do this? :/
+            if not attr == '_DatastoreRecord__encoded_data':
```
KPrasch: :grimacing:

tests/integration/datastore/test_datastore.py
```diff
@@ -14,10 +14,52 @@
...
 from nucypher.datastore import datastore, keypairs
+from nucypher.datastore.db.models import DatastoreRecord, RecordField
+
+
+def test_datastore_record():
+
+    class TestRecord(DatastoreRecord):
+        test = RecordField(bytes)
+        test_date = RecordField(datetime,
+                encode=lambda val: datetime.isoformat(val).encode(),
```
KPrasch, line 32: Why a lambda?

nucypher/datastore/db/models.py
```diff
@@ -34,71 +37,84 @@ class DatastoreRecord(Mapping):
...
+                raise TypeError(f"Either there's no existing RecordField attr for {attr} or you're trying to mutate the attribute, which is immutable.")
```
KPrasch: Understanding the design - What is "the attribute" referring to here?

tuxxy: You’re not supposed to be able to mutate any attributes on the Record. If I already have a record instance, I shouldn’t be able to mutate the attributes or create new ones.

tests/integration/datastore/test_datastore.py
tuxxy: The `DatastoreRecord`, when encoding or decoding, passes the value as an arg into the callable given. In this case, I need to make it a lambda so that it can encode the string into a utf-8 bytestring.

nucypher/datastore/db/models.py
tuxxy: Yeah, would love a better way to do this.

nucypher/datastore/db/models.py
tuxxy: This just simply allows us to call `encode` or `decode` regardless if the type needs to be encoded or decoded. It makes logic in `DatastoreRecord` much cleaner and avoids unnecessary “look before you leap” patterns.

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,189 @@
...
-
-    def del_policy_arrangement(self, arrangement_id: bytes, session=None) -> int:
-        """
-        Deletes a PolicyArrangement from the Keystore.
+        If the record is used outside the scope of the context manager, any
+        writes or reads will error.
         """
-        session = session or self._session_on_init_thread
-        deleted_records = session.query(PolicyArrangement).filter_by(id=arrangement_id).delete()
+        with suppress(ValueError):
```
KPrasch, line 107: why suppress here?

tuxxy: If we are given a `record_id` it can be either an `int` or a `str`. Since keys must be encoded as `bytes`, it becomes really difficult to manage both without making a mistake somewhere. So we cast all IDs to be an `int` if they are capable.

KPrasch: In what situation do you expect a ValueError to be raised?

tuxxy:
```python
int('my_id_is_a_string')
# vs
int('123123')
```

nucypher/config/storages.py
```diff
@@ -290,84 +290,6 @@ def initialize(self) -> bool:
         return not bool(self.__metadata or self.__certificates)
 
 
-class SQLiteForgetfulNodeStorage(ForgetfulNodeStorage):
```
derekpierre, line 293: Any reason for this deletion? The NuCypher Status Monitor (`nucypher/nucypher-monitor`) uses this storage mechanism - https://github.com/nucypher/nucypher-monitor/blob/master/monitor/crawler.py

tuxxy: What part of "Rip out as much SQLite testing stuff as I can" do you not understand?!??!

I'm kidding, but I was really ripping everything out that said SQLalchemy or SQLite as fast as I could haha. It can probably go back in, but perhaps we can replace a lot of it with the new datastore? Could it potentially get simplified?

KPrasch: Perhaps, just be aware this is part of a different API around node storage, unrelated to the datastore.

tuxxy: I'm not exactly sure I understand how it's being used. What is the reason we have a "forgetful" **SQLite** storage mechanism? Why not use a dict instead?

derekpierre: It is forgetful because it is associated with data from the learning loop - which restarts from scratch every time the monitor is restarted.

In terms of why not use a `dict` - our queries are a little more involved than just (key, value) pairs and sqlite queries make that easier - not that it can't be done, but it is easier; remember as well that the Status Monitor is expected to run for a long time. For more insight into the queries used look in the `CrawlerStorageClient` class, see https://github.com/nucypher/nucypher-monitor/blob/master/monitor/db.py#L15.

tuxxy: I think this class could be sensibly moved to the `nucypher-monitor` repo and maintained there. I don't see a reason to continue maintaining SQLite for complex queries we won't be performing in this repo.

Perhaps `nucypher-monitor` can grow to be a library for performing those more complex queries?

derekpierre: We can leave SQLiteForgetfulNodeStorage for now. Filed #2118 as a follow-up.

derekpierre: > "Rip out as much SQLite testing stuff as I can"
ditto except for this piece of old wallpaper ... 😅

KPrasch:
Heh - this PR name is... let's say "minimal" compared to the complete rework contained within :joy: .

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,207 @@
...
+            # If the ID can be an int, we convert it
+            key_parts[-1] = int(key_parts[-1])
+        return cls(*key_parts)
 
-            if not workorders:
-                raise NotFound
+    def compare_key(self, key_bytestring: bytes) -> bool:
+        other_key = DatastoreKey.from_bytestring(key_bytestring)
+        return self.record_type == (other_key.record_type or self.record_type) and \
```
derekpierre, line 61: What happens if I have a `DatastoreKey`with actual values, and run `compare_key` on a `key_bytestring` and `cls` that returns `None` values for `record_type`, `record_field`, and `record_id`? Because of the `or self.<member>` in the `self.<member> == (other.<member> or self.<member>` comparison statements, this could still return `True` no?

tuxxy: If I'm understanding you correctly, in practice, the `Datastore` cannot return a key with `None` types in those fields, so this case is not possible.

This is used for quickly iterating over the keys in lmdb for performing queries since the Python library can't do queries based on prefixes, so this is a relatively efficient implementation of that.

If you query for a partial key (such as `PolicyArrangement`), you're either going to find a key that matches it (meaning you have `PolicyArrangement` records) or you won't find a key that matches (meaning you have _no_ `PolicyArrangement` records). In the latter case, lmdb returns `None` and we always check, when setting the iterator to a lmdb key, that it _does not_ return `None`. If it does, we break our query loop.

So again, if I'm understanding you correctly, that's why we can effectively ignore that case. We never encounter keys that have a `None` and we effectively know that we will always have, at least, a `record_type`.

tuxxy: Oh, I guess it would help to know that lmdb organizes keys lexicographically. Because of this, if I query for a key of `a` when I only have a key of `aardvark`, then my iterator will be placed at `aardvark`. Accordingly, if I query for a `PolicyArrangement:foobar_id`, but I only have a `PolicyArrangement:not_foobar_id`, then the cursor will return `None`. Hopefully that answers your question.

derekpierre: It makes a bit more sense now why you do the `or self.<member>` check - it's to place the iterator at a relevant spot.

So is the reason you return `True` when keys aren't actually equal, is to have a pointer to a particular spot when so that you can potentially do a subsequent insertion of a new record? However, I guess this means whenever just querying for a record (not caring about insertion if the data does not exist) you have to ensure that the returned data is indeed what you wanted and not just the iterator pointing to a potential insertion point.

Have I understood that correctly?

tuxxy: So the key is irrelevant to insertion since lmdb handles all that logic for us. However, lmdb does _not_ handle query logic, so that's predominantly what this class is for.

When performing a query, we need to match keys that fit a certain structure and that comes with a sense of precedence. Consider the following query keys (as `<RecordType>:<RecordField>:<RecordID>`):

1. `TestRecord`
2. `TestRecord:test_field`
3. `TestRecord:test_field:1234`

These keys are enumerated in descending order based on specificity; 1 is least specific and 3 is most specific.
We need to be able to match keys that get returned whilst iterating over the key-value store and know if they are _within_ a query or not. The 1st query key will return _all_ `TestRecord`s. The 2nd query key will return `TestRecord`s that have a `test_field` field. The 3rd query key will return a `TestRecord` that has a `test_field` and an id of `1234`.

To perform queries like the first and the second, we need some precedence in the query key structure. Wherever the "key slot" (meaning the relevant attr in `DatastoreKey`) is _not_ set, the key slot will be `None` by default. This acts as a precedence wildcard.

In Python, the logic `primary or secondary` will always return `primary` when it's `True`, otherwise it will return `secondary`. The `(other.<member> or self.<member>)` logic acts as a method of setting precedence in the key structure; we _always_ compare to (and prefer) the _given key_ if stated, otherwise we'll take the precedence of the _comparing key_ (the key calling the `compare_key` method).

In the `Datastore` query logic, our _given key_ is the query key (like those above) and the _comparing key_ is the key retrieved from lmdb.

tuxxy: I added some comments for clarification in the `compare_key` method.

It's best to think of this process as _comparing specific keys to partial keys_ where we need to know if a specific key matches a given partial key.

derekpierre: 👍

nucypher/datastore/base.py
```diff
@@ -0,0 +1,212 @@
...
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
+along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+"""
+import msgpack
+from typing import Any, Callable, Iterable, NamedTuple, Optional, Union
+
+
+class DBWriteError(Exception):
```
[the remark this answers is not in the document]
tuxxy: No, this is basically at a lower-level than the `Datastore` where the record can't be written to the underlying DB. There's no `Datastore` yet here, the error you really mean is `DatastoreTransactionError`.

KPrasch:
Whas is the relation between this PR and #2132 ?

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,238 @@
...
+                        field = getattr(record(writeable=False), filter_field)
+                    except (TypeError, AttributeError):
+                        continue
+
+                    record = partial(record_type, datastore_tx, curr_key.record_id)
+                    # We pass the field to the filter_func if `filter_field` and
+                    # `filter_func` are both provided. In the event that the
+                    # given `filter_field` doesn't exist for the record or the
+                    # `filter_func` returns `False`, we call `continue`.
+                    if filter_field and filter_func:
```
derekpierre, line 210: isn't this check redundant based on L199?

nucypher/network/server.py
```diff
@@ -254,12 +240,10 @@ def set_policy(id_as_hex):
...
 
-        with ThreadedSession(db_engine) as session:
-            datastore.attach_kfrag_to_saved_arrangement(
-                alice,
-                id_as_hex,
-                kfrag,
-                session=session)
+        with datastore.describe(PolicyArrangement, id_as_hex, writeable=True) as policy_arrangement:
+            if not policy_arrangement.alice_verifying_key == alice.stamp.as_umbral_pubkey():
+                raise alice.SuspiciousActivity
```
derekpierre, line 245: Perhaps a message for context eg. `Mismatched verifying key`?

nucypher/network/server.py
```diff
@@ -273,22 +257,17 @@ def revoke_arrangement(id_as_hex):
 
         revocation = Revocation.from_bytes(request.data)
         log.info("Received revocation: {} -- for arrangement {}".format(bytes(revocation).hex(), id_as_hex))
+
+        # Check that the request is the same for the provided revocation
+        if not id_as_hex == revocation.arrangement_id.hex():
+            log.debug("Couldn't identify an arrangement with id {}".format(id_as_hex))
+            return Response(status_code=400)
```
derekpierre, line 264: Response message for context perhaps...`Mismatched arrangement ID`?

tests/acceptance/cli/test_worklock_cli.py
```diff
@@ -230,7 +231,8 @@ def test_refund(click_runner, testerchain, agency_local_registry, token_economic
                     checksum_address=bidder,
                     worker_address=worker_address,
                     rest_host=MOCK_IP_ADDRESS,
-                    rest_port=select_test_port())
+                    rest_port=select_test_port(),
+                    db_filepath=tempfile.mkdtemp())
```
derekpierre, line 235: Should this be an temp file, rather than a temp directory? It doesn't seem that we use `subdir=True` (https://lmdb.readthedocs.io/en/release/#lmdb.lmdb.open) when using `lmdb.open`.

tests/acceptance/cli/ursula/test_run_ursula.py
```diff
@@ -147,7 +147,7 @@ def run_ursula(teacher_uri):
 
     yield d
 
-
+@pytest.mark.skip("Let's put this on ice until we get 2099 and Treasure Island working together.")
```
derekpierre, line 150: TODO / follow-up PR?

tests/acceptance/cli/test_worklock_cli.py
tuxxy: The `subdir` param is `True` by default. It could be turned off, but it would require that we create the directories first.

tests/acceptance/cli/ursula/test_run_ursula.py
tuxxy: This is in #2132

nucypher/datastore/datastore.py
tuxxy: Oh good catch! The check isn't just redundant, the whole block is!

I must've not deleted this block during the rebase.

tuxxy: Resolved in 078b5b1

tuxxy:
> Whas is the relation between this PR and #2132 ?

#2132 should be considered stale until it gets rebased and corrected over this PR. There's three commits there that are useful, but the PR itself has nothing to do with this one.

nucypher/datastore/datastore.py
tuxxy: Can we get a formal "production" test for this before merging this changeset?

nucypher/characters/lawful.py
```diff
@@ -1085,10 +1085,18 @@ def __init__(self,
 
     def __prune_arrangements(self) -> None:
         """Deletes all expired arrangements and kfrags in the datastore."""
-        now = datetime.fromtimestamp(self._arrangement_pruning_task.clock.seconds())
+        now = maya.MayaDT.from_datetime(datetime.fromtimestamp(self._arrangement_pruning_task.clock.seconds()))
         try:
-            result = self.datastore.del_expired_policy_arrangements(now=now)
-        except OperationalError:
+            with self.datastore.query_by(PolicyArrangement,
```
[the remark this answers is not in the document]
tuxxy: I was thinking of making a `datastore.queries` module and moving a lot of these into that.

Since the datastore is accessed through a context manager and returns the `DatastoreRecord` objects, I wanted to have the module simply just be a bunch of `contextmanager` decorated functions that take a `datastore` as an argument. I did want to do this in a separate PR, though.

tests/integration/datastore/test_datastore.py
```diff
@@ -14,64 +14,392 @@
...
+        with storage.describe(TestRecord, 'test_id', writeable=True) as test_record:
+            # Valid write
+            test_record.test = b'this will not persist'
+            # Erroneous write causing an abort
+            test_record.nonexistent = b'causes an error and aborts the write'
 
-    # Test add workorder
-    new_workorder1 = test_datastore.save_workorder(bob_keypair_sig1.pubkey, b'test0', arrangement_id)
-    new_workorder2 = test_datastore.save_workorder(bob_keypair_sig2.pubkey, b'test1', arrangement_id)
+    # Check that nothing was written from the aborted transaction above.
```
derekpierre, line 98: 👍

nucypher/datastore/base.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def __hash__(self):
+        """
+        WARNING: Records are only considered unique per their record IDs in this method.
+        In some cases (Workorder vs PolicyArrangement), these records may have
+        the same IDs. This comparison is useful when iterating over the 
+        datastore _as a subset of record type_ (as we do in queries), but not
+        useful when comparing two records of different type.
+        """
+        return hash(self._record_id)
```
KPrasch, line 201: I'm a little worried about this becoming the source of a bug.  why is `__eq__` not enough? ( @jMyles please advise )

nucypher/utilities/prometheus/collector.py
```diff
@@ -117,7 +119,11 @@ def _collect_internal(self) -> None:
 
         self.metrics["learning_status"].state('running' if self.ursula._learning_task.running else 'stopped')
         self.metrics["known_nodes_gauge"].set(len(self.ursula.known_nodes))
-        self.metrics["work_orders_gauge"].set(len(self.ursula.work_orders()))
+        try:
```
KPrasch, line 122: IMO, this belongs in a method on Ursula.

tests/constants.py
```diff
@@ -121,7 +121,7 @@
 MOCK_IP_ADDRESS = '192.0.2.100'
 MOCK_IP_ADDRESS_2 = '203.0.113.20'
 
-MOCK_URSULA_DB_FILEPATH = ':memory:'
+MOCK_URSULA_DB_FILEPATH = tempfile.mkdtemp()
```
KPrasch, line 124: When does this get cleaned up?

tests/integration/config/test_character_configuration.py
```diff
@@ -83,7 +83,7 @@ def test_federated_development_character_configurations(character, configuration
     # Node Storage
     assert configuration.TEMP_CONFIGURATION_DIR_PREFIX in thing_one.keyring_root
     assert isinstance(thing_one.node_storage, ForgetfulNodeStorage)
-    assert thing_one.node_storage._name == ":memory:"
+    assert ':memory:' in thing_one.node_storage._name
```
KPrasch, line 86: :+1:

tests/integration/learning/test_discovery_phases.py
```diff
@@ -105,12 +114,14 @@ def mock_receive_treasure_map(treasure_map_id):
...
-                                highperf_mocked_bob, b"any label", m=20, n=30,
-                                expiration=maya.when('next week'),
-                                publish_treasure_map=False)
+                    with patch("nucypher.datastore.models.PolicyArrangement._alice_verifying_key",
+                               new=not_public_key_record_field):
+                        with mock_cert_loading, mock_metadata_validation, mock_message_verification:
+                            with mock_secret_source():
+                                policy = highperf_mocked_alice.grant(
+                                    highperf_mocked_bob, b"any label", m=20, n=30,
+                                    expiration=maya.when('next week'),
```
KPrasch, line 123: :laughing:

tests/mock/performance_mocks.py
```diff
@@ -499,7 +500,7 @@ def __getitem__(self, route_name):
 
     def actual_rest_app(self):
         if self._actual_rest_app is None:
-            self._actual_rest_app, _datastore = make_rest_app(db_filepath="no datastore",
```
KPrasch, line 502: Good find - this was leaving an artifact in the filesystem after tests; Nonetheless, let's make sure we clean this up too and not just sweep it under the rug.

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
+        An optional `filter_field` can be provided as a `str` to perform a
+        query on a specific field for a `record_type`. This will cause the
+        `filter_func` to receive the decoded `filter_field` per the `record_type`.
+        Additionally, providing a `filter_field` will limit the query to
+        iterating over only the subset of records specific to that field.
+
+        If records can't be found, this method will raise `RecordNotFound`.
+        """
+        valid_records = set()
+        with self.__db_env.begin(write=writeable) as datastore_tx:
```
KPrasch, line 163: This `with` block is massive!

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
+
+                record = partial(record_type, datastore_tx, curr_key.record_id)
+
+                # We pass the field to the filter_func if `filter_field` and
+                # `filter_func` are both provided. In the event that the
+                # given `filter_field` doesn't exist for the record or the
+                # `filter_func` returns `False`, we call `continue`.
+                if filter_field and filter_func:
+                    try:
+                        field = getattr(record(writeable=False), filter_field)
```
KPrasch, line 203: `class` > `def` > `with` > `for` > `if` > `try` - is too nested IMO.

nucypher/datastore/datastore.py
tuxxy: Yes, it's a bit unfortunate. I'd say it's mostly comments. :P
I'd try to make it more condense, but this is the API that `lmdb` offers.

KPrasch: The size of this with block is not a limitation of lmbd. You can break up this function into smaller ones.

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
+                # At last, we yield the queried records
+                yield list(valid_records)
+            except (AttributeError, TypeError, DBWriteError) as tx_err:
+                # Handle `RecordNotFound` cases when `writeable` is `False`.
+                if not writeable and isinstance(tx_err, AttributeError):
+                    raise RecordNotFound(tx_err)
+                raise DatastoreTransactionError(f'An error was encountered during the transaction (no data was written): {tx_err}')
+            finally:
+                for record in valid_records:
+                    record.__dict__['_DatastoreRecord__writeable'] = False
```
KPrasch, line 235: You have self in scope here - why do you need to directly mutate the object dict?

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
+                    raise RecordNotFound(tx_err)
+                raise DatastoreTransactionError(f'An error was encountered during the transaction (no data was written): {tx_err}')
+            finally:
+                # Now we ensure that the record is not writeable
+                record.__dict__['_DatastoreRecord__writeable'] = False
+
+    @contextmanager
+    def query_by(self,
+              record_type: 'DatastoreRecord',
+              filter_func: Callable[[Union[Any, 'DatastoreRecord']], bool]=None,
```
KPrasch, line 141: Spaces around the `=` are conventional here.

nucypher/datastore/datastore.py
tuxxy: I'm confused -- I'm not mutating `self` here, I'm mutating `record`.
This ensures that modifying attrs on the record, outside of the context manager, raises a `TypeError` and not fail silently.
Due to how `DatastoreRecord` attrs are used to access data, this avoids calling `__setattr__`.

KPrasch: Ah - I must've misread it.  In any case this is a major code smell - whats going on here?

KPrasch: > This ensures that modifying attrs on the record, outside of the context manager, raises a TypeError and not fail silently.
Due to how DatastoreRecord attrs are used to access data, this avoids calling __setattr__.

Sure - there sound design in there :+1:  - but mechanically this is awkward - I'm not entirely sure what the code is telling us here, perhaps that the scope is not correct :thinking:

tuxxy: So if we have a query:
```
with datastore.query_by(FooRecord, writeable=True) as my_foo_records:
    my_record = my_foo_records[0]
    .... # Do something with my_record

# I later try to do something with `my_record` when I'm not supposed to:
my_record.foobar = b'uh oh'
```

I expect the `my_record.foobar` to be an error -- it doesn't have access to a transaction to even set the `foobar` field, but I don't want it to even get that far. This should be an error caught in `__setattr__`.

KPrasch: To be totally clear - you are trying to access a protected attribute in another class, which is intentional obscured.  

A fix here might be as simple as loosening up and just removing a single underscore :-)

KPrasch:
```suggestion
                    record._writeable(False)
```

nucypher/datastore/base.py
jMyles: I... guess I agree that this is a very plausible source of a bug.  And the benefit is not immediately obvious to me.  Why not add one or more objects to the hash that make it decidedly unique?

nucypher/datastore/datastore.py
tuxxy: I tried that previously. Imho, it makes the method as a whole much less readable.

nucypher/datastore/datastore.py
KPrasch: > I tried that previously. Imho, it makes the method as a whole much less readable.

@tuxxy - So you then also think this is acceptable?

nucypher/datastore/datastore.py
tuxxy: I guess it could be written as `_writeable`, but that also has implications on `DatastoreRecord` as well. Presently, field types are expressed as `_<name>`, so this would make `_writeable` share the same "namespace" as those types, which was my reasoning for making this `__writeable` since I know nothing outside of the `Datastore` would (or should) be accessing it.

nucypher/datastore/datastore.py
tuxxy: Are you suggesting that the `with` indentation makes the method less readable? I'm not sure I agree.

KPrasch: Not merely the `with` - but the function is ~72 lines (unnecessarily long), and has an extremely high cognitive complexity that include many conditionals making it hard to understand and maintain.

nucypher/network/server.py
tuxxy: I'll defer judgement here to @jMyles since these responses aren't in the scope of the PR, but more than happy to make them more specific.

nucypher/datastore/datastore.py
KPrasch: Yes - let's do an experiment on testnet sooner vs. later!

KPrasch:
@tuxxy - Looks like there is a filed docs build, perhaps from introducing a new module and nod listing in in the TOC?

tests/integration/datastore/test_datastore.py
```diff
@@ -14,64 +14,392 @@
...
+        assert records[0].foo == b'one record'
+
+        # This record isn't writeable
+        with pytest.raises(TypeError):
+            records[0].foo = b'this will error'
+
+    # Make a writeable query on BarRecord
+    with storage.query_by(BarRecord, writeable=True) as records:
+        records[0].bar = b'this writes'
+        records[1].bar = b'this writes'
```
derekpierre, line 245: These values aren't actually written to the datastore until the context manager is exited..correct? If so then the assertions below that check the written value should instead be done outside of this context manager, in a different one - I guess the records would have to be queried for again, and then the values checked

tuxxy: Technically, none of the statements in a context manager are executed until it's exited and all exceptions are handled, I believe.

So you can do either/or: check records in the context manager, or check records outside of the context manager -- both work fine. When I do it inside/outside, I just do it to make the action being performed obvious, e.g. a query for a record with a field that didn't get written is a `RecordNotFound`, etc.

tests/constants.py
tuxxy: In short, the answer is complicated...
It heavily depends on the operating system and the distribution. My all-encompassing-answer guess would be either at boot time or at least once per-day.

In some cases, they're never written to disk and only written to a RAM-disk (if you have a `tmpfs`). I think on OS X, there's some policy about deleting tempfiles if they aren't accessed in a certain amount of time.

tests/integration/datastore/test_datastore.py
tuxxy: I reread what I wrote here, and I'm not very keen on my answer.

You shouldn't confuse the write `records[0].bar = b'this writes'` with something like a buffer. When that statement is executed, `b'this writes'` is written to lmdb. What lmdb does behind the scenes is ensure that all errors are handled before being written to disk, I think.

Technically, yes, it's not written until the context manager is exited. When the database transaction is finalized (I think on that context manager exit), it goes from being written to memory to being written to the disk. So, while you're in the context manager, you can make these read statements and they will read from memory with what lmdb intends to write to disk.

You point out that the check should be outside the context manager to truly assert a read -- and you're right. If you really want to confirm a write, it should be outside the context manager, but for the purposes of some of these tests it doesn't seem very important to assert that. What I'm really trying to test is the `Datastore` and not lmdb.

nucypher/utilities/prometheus/collector.py
tuxxy: I'd like to make this a query in `datastore.queries` instead. I'll do that in a follow-up PR. :+1:

nucypher/datastore/base.py
tuxxy: As we discussed in Discord, this is decidedly unique as long as record IDs are unique. I think this is a reasonable assumption when designing models and is incredibly obvious when not being done.

nucypher/cli/painting/nodes.py
```diff
@@ -54,6 +56,12 @@ def paint_node_status(emitter, ursula, start_time):
     # Build FleetState status line
     fleet_state = build_fleet_state_status(ursula=ursula)
 
+    try:
```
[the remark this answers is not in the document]
tuxxy: Agreed, would prefer to move this to `datastore.queries` :+1:

nucypher/characters/lawful.py
```diff
@@ -1493,17 +1501,9 @@ def interface_info_with_metadata(self):
         return constants.BYTESTRING_IS_URSULA_IFACE_INFO + bytes(self)
 
     #
-    # Work Orders & Re-Encryption
+    # Re-Encryption
     #
 
-    def work_orders(self, bob=None) -> List['WorkOrder']:
```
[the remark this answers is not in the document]
tuxxy: Tracking in #2153

nucypher/datastore/datastore.py
tuxxy: Before this is merged, preferably. If there's an issue, I need to know _now_.

tests/constants.py
derekpierre: The docs indicate the following (https://docs.python.org/3.4/library/tempfile.html#tempfile.mkdtemp):

```
The user of mkdtemp() is responsible for deleting the temporary directory and its contents when done with it.
```

If you don't want to handle deletion, an alternative seems to be https://docs.python.org/3.4/library/tempfile.html#tempfile.TemporaryFile

tests/integration/datastore/test_datastore.py
derekpierre: 👍

KPrasch:
Outstanding issue with the docs build, several RFCs, possibly new Issues can be opened if they not are addressed.

tests/integration/datastore/test_datastore.py
```diff
@@ -14,64 +14,392 @@
...
+    #
+    # Tests for `Datastore.describe`
+    #
 
-    # Test del pubkey
-    test_datastore.del_key(federated_bob.stamp.fingerprint())
-    with pytest.raises(datastore.NotFound):
-        del_key = test_datastore.get_key(federated_bob.stamp.fingerprint())
+    # Getting writeable access to a record can be done by not setting `writeable` to `True`.
+    # `writeable` is, by default, `False`.
```
fjarri, line 47: Doesn't look like "not" should be there.

tests/integration/datastore/test_datastore.py
```diff
@@ -14,64 +14,392 @@
...
-    assert {new_workorder1, new_workorder2}.issubset(query_workorders)
+    # However, a handled error will not cause an abort.
+    with storage.describe(TestRecord, 'test_id', writeable=True) as test_record:
+        # Valid operation
+        test_record.test = b'this will persist'
+        try:
+            # Maybe we don't know that this field exists or not?
+            # Erroneous, but handled, operation -- doesn't cause an abort
+            should_error = test_record.bad_read
+        except TypeError:
```
fjarri, line 110: Would it be more logical to raise `AttributeError` if the field is not found?

fjarri: Hm, actually just below a similar construction does raise an `AttributeError`...

tests/integration/datastore/test_datastore.py
tuxxy: haha good looking out! Thanks!

tests/integration/datastore/test_datastore.py
tuxxy: I have very uncertain opinions about how I use these two errors, and it's one of the most obvious sore spots in this code (for me, at least).

The difference between these errors is mostly a Pythonic notion. A `TypeError` is raised when you try to do an operation on an invalid type. For example, `[print(i) for i in 1234]` is a `TypeError` because you can't iterate over an `int`. You can think of it like trying to do an unsupported operation.
In contrast, an `AttributeError` is for errors with assignment or references. The datastore API, unfortunately, has the ill effect of blurring these lines sometimes.

This is a `TypeError` because you're trying to read from a record with a field that doesn't exist in its model definition. It goes beyond an "unsupported operation" and into this just not being possible at all.

The `AttributeError` below is very similar, but it's also very different: you're trying to read from a record that does have this field in its model definition, but has no data. This is indeed a supported operation, meaning you're still able to write to it. In comparison to the `TypeError`, you're not able to write to that field at all (or ever) until the model is updated. Does that make sense?

KPrasch: Consider a custom exception class.

fjarri: I actually find Python inconsistent here. `[print(i) for i in 1234]` checks for `__iter()__`, doesn't find it and raises `TypeError`. But if you try `(1234).__iter__()`, you'll get `AttributeError`. Would be nice to have some kind of `InterfaceError`, but alas...

So for me a more expected behavior would be 
- an `AttributeError` if the field doesn't exist in the model
- a `ValueError` if you try to read from a field that exists but has no data (why even raise anything, by the way? Would just returning `None` work?). Perhaps a custom exception would be even better than `ValueError` here.

fjarri: Adding to that, if reading from an existing field with no data is an error, why is it possible to create a record with such a field in the first place?

tuxxy: > Adding to that, if reading from an existing field with no data is an error, why is it possible to create a record with such a field in the first place?

Why would you not be able to? This is the equivalent of a "nullable" field in any sort of relational DB. An example we use of this is when we handle arrangements for the first time from Alice; we create an arrangement without a KFrag and set it when it's received. This logic would prevent us from doing that.

> Would just returning `None` work?

Not really. There's a very subtle problem with this logic, and a more practical usability issue. If I return `None` from a field, am I returning `None` or does the field have a `NoneType` in it?
For the usability issue, the argument is much better. Reading an empty field in a record _should not be silent_. It should raise an error at a minimum. This ensures that datastore readers handle null logic, and prevents a `None` from bubbling up through the code where it gets difficult to debug. You're right, though: a custom error type might make more sense for this part.

> Consider a custom exception class.

I intend to, but I need to figure out a more broad direction of where the datastore's API is going to go, and what cases I may figure out. I'd like to simplify this along with some `__setattr__` and `__getattr__` logic with better errors, but it needs iterative work.

tuxxy: Additionally, current deletion logic is handled by setting a field to `None`, which tells the datastore to delete the field.

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
+            # We begin by comparing the current key to the query key.
+            # If the key doesn't match the query key, we know that there are
+            # no records for the query because lmdb orders the keys lexicographically.
+            # Ergo, if the current key doesn't match the query key, we know
+            # we have gone beyond the relevant keys and can `break` the loop.
+            # Additionally, if the record is already in the `valid_records`
+            # set (identified by the `record_id`, we call `continue`.
+            for db_key in db_cursor.iternext(keys=True, values=False):
+                curr_key = DatastoreKey.from_bytestring(db_key)
+                if not curr_key.compare_key(query_key):
```
cygnusv, line 190: After our discord discussion, I still think that  the name of this method should be related to `matches`, `conforms_to` or similar. I mean, even in the comment above you say "match".

nucypher/datastore/base.py
```diff
@@ -43,6 +43,7 @@ def __init__(self,
                  record_id: Union[int, str],
                  writeable: bool = False) -> None:
         self._record_id = record_id
+        self._fields = [field[1:] for field in type(self).__dict__ if type(type(self).__dict__[field]) == RecordField]
```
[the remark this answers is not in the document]
tuxxy: That's outdated. :)

tests/acceptance/cli/ursula/test_stake_via_allocation_contract.py
```diff
@@ -521,7 +522,8 @@ def test_collect_rewards_integration(click_runner,
                     rest_host='127.0.0.1',
                     rest_port=ursula_port,
                     start_working_now=False,
-                    network_middleware=MockRestMiddleware())
+                    network_middleware=MockRestMiddleware(),
+                    db_filepath=tempfile.mkdtemp())
```
vzotova, line 526: Is it possible to use here and in other places some test constant? like `MOCK_URSULA_DB_FILEPATH `. Otherwise I guess many directories will be created when running tests

tests/integration/datastore/test_datastore.py
```diff
@@ -14,64 +14,392 @@
...
+    _test_date = RecordField(datetime,
+            encode=lambda val: datetime.isoformat(val).encode(),
+            decode=lambda val: datetime.fromisoformat(val.decode()))
 
-@pytest.mark.usefixtures('testerchain')
-def test_key_sqlite_datastore(test_datastore, federated_bob):
 
-    # Test add pubkey
-    test_datastore.add_key(federated_bob.stamp, is_signing=True)
+def test_datastore_describe():
```
vzotova, line 35: as mentioned in discord - this test file looks more like unit test

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
+            db_cursor = datastore_tx.cursor()
+
+            # Set the cursor to the closest key (if it exists) by the query params.
+            #
+            # By providing a `filter_field`, the query will immediately be
+            # limited to the subset of keys for the `filter_field`.
+            query_key = f'{record_type.__name__}:{filter_field}'.encode()
+            if not db_cursor.set_range(query_key):
+                # The cursor couldn't identify any records by the key
+                raise RecordNotFound(f"No records exist for the key from the specified query parameters: '{query_key}'")
```
vzotova, line 173: There are several places with this error text , maybe extract to constant/error?

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
 
-        if policy_arrangement.alice_verifying_key.key_data != alice.stamp:
-            raise alice.SuspiciousActivity
+class DatastoreKey(NamedTuple):
+    """
+    Used for managing keys when querying the datastore.
+    """
+    record_type: Optional[str] = None
+    record_field: Optional[str] = None
+    record_id: Optional[Union[bytes, int]] = None
```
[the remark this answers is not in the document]
tuxxy: Because `lmdb` returns keys as `bytes` or `int`s.

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
 
-        policy_arrangement.kfrag = bytes(kfrag)
-        self.__commit(session=session)
+    @classmethod
+    def from_bytestring(cls, key_bytestring: bytes) -> 'DatastoreKey':
+        key_parts = key_bytestring.decode().split(':')
```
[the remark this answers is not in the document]
tuxxy: You can't have `:` in a class name or an attribute, so that's not possible there. It's only possible in the ID. There's currently no checks in `DatastoreRecord` for invalid characters in the record ID, so this would make it such that any call to `query_by` that hits this ID will error resulting in an impossible to finish query. Good catch.

I don't think the solution is to prevent `:` from being in the ID (that may actually be useful at some point for encoding further information in the key string). Additionally, I much prefer to avoid writing branching logic for things like bad characters.

What do you think about simply putting the return of this method in a `try/except`, and when we get a `TypeError` for passing too many pieces we simply provide the first three? This allows us to ensure that all key strings are evaluated equitably, and additionally lets us avoid having to concern ourselves with the minutiae of what a developer names their IDs all while keeping an extremely flexible key structure.

nucypher/datastore/datastore.py
```diff
@@ -93,182 +124,94 @@ def describe(self, record_type: 'DatastoreRecord', record_id: Union[int, str], w
...
-#        """
-#        session = session or self._session_on_init_thread
-#
-#        workorders = session.query(Workorder).filter_by(arrangement_id=arrangement_id)
-#        deleted = workorders.delete()
-#        self.__commit(session=session)
-#        return deleted
+    @contextmanager
+    def query_by(self,
+              record_type: 'DatastoreRecord',
```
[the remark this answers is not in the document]
tuxxy: Good catch, thanks!

nucypher/datastore/datastore.py
tuxxy: After the recent tests, I'm going to consider this resolved now.

requirements.txt
```diff
@@ -18,80 +18,82 @@ colorama==0.4.3
...
+lmdb==0.98
 lru-dict==1.1.6
-markupsafe==2.0.0a1
-marshmallow==3.6.1
+markupsafe==1.1.1; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
+marshmallow==3.7.0
 maya==0.6.1
 msgpack-python==0.5.6
-multiaddr==0.0.9
+msgpack==1.0.0
```
derekpierre, line 51: @tuxxy I guess this is what @cygnusv was referring to regarding different versions of msgpack - seems msgpack-python is now deprecated - https://pypi.org/project/msgpack-python.

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
-                raise NotFound
 
-        return list(workorders)
-
-    def del_workorders(self, arrangement_id: bytes, session=None) -> int:
-        """
-        Deletes a Workorder from the Keystore.
-        """
-        session = session or self._session_on_init_thread
+class Datastore:
```
KPrasch, line 80: This full path has become `nucypher.datastore.datastore.Datastore` - which is a bit repetitive.

nucypher/datastore/datastore.py
```diff
@@ -14,269 +14,222 @@
...
+    A persistent storage layer for arbitrary data for use by NuCypher characters.
+    """
 
-        workorders = session.query(Workorder).filter_by(arrangement_id=arrangement_id)
-        deleted = workorders.delete()
-        self.__commit(session=session)
-        return deleted
+    # LMDB has a `map_size` arg that caps the total size of the database.
+    # We can set this arbitrarily high (1TB) to prevent any run-time crashes.
+    LMDB_MAP_SIZE = 1_000_000_000_000
```
KPrasch, line 87: We've recently observed this value leading to the inability to allocate enough vram to run the tests that include concurrent ursulas and multiple lmdb instances.  

Let's set a different value for tests vs. production and go from there.

tests/integration/conftest.py
```diff
@@ -54,6 +57,38 @@
...
 
+class TestLMDBEnv:
+    """
+    This class is used to have LMDB environments open just-in-time as they're
+    needed.
+
+    This is necessary in testing environments because there may be too many
+    LMDB environments open at once.
+    """
+
```
KPrasch, line 68:
```suggestion
    __test__ = False  # prohibit pytest from collecting this object as a test
```

requirements.txt
derekpierre: Do we need to remove the `msgpack-python` dependency?

tuxxy: I don't really know. I'd like to avoid making more dependency changes in this PR, though because of the testing we've performed with it. Perhaps we start another dependency update PR into `main` for this?

derekpierre: A follow-up issue makes sense. Filed #2171

tuxxy: Sounds good, thanks!

nucypher/datastore/datastore.py
tuxxy: I think I resolved this with a monkeypatch. Will you take a peek in the latest commit?

nucypher/datastore/datastore.py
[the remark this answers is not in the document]
cygnusv: That would mean that only the first part of such hypothetical record ID is considered. That's fine as long as it's documented somewhere. I also had an alternative idea: what if you simply use ByteStringSplitter here?

PR #2069, opened 2020-06-03
Initial Refactor of ursula prometheus metrics work from #1923

derekpierre:
Related to #1923 

Closes #2058.

Follow-up #2091.

PR #2100, opened 2020-06-17
Ensure that prometheus imports are locally scoped

derekpierre:
Related to #2069 .

### DIFF
--- a/nucypher/cli/commands/ursula.py
+++ b/nucypher/cli/commands/ursula.py
@@ -73,7 +73,6 @@ from nucypher.config.constants import (
 )
 from nucypher.config.keyring import NucypherKeyring
 from nucypher.utilities.networking import determine_external_ip_address
-from nucypher.utilities.prometheus.metrics import PrometheusMetricsConfig
 
 
 class UrsulaConfigOptions:
@@ -382,8 +381,10 @@ def run(general_config, character_options, config_file, interactive, dry_run, me
                                                                config_file=config_file,
                                                                json_ipc=general_config.json_ipc)
 
-    prometheus_config = None
+    prometheus_config: 'PrometheusMetricsConfig' = None
     if prometheus:
+        # Locally scoped to prevent import without prometheus explicitly installed
+        from nucypher.utilities.prometheus.metrics import PrometheusMetricsConfig
         prometheus_config = PrometheusMetricsConfig(port=metrics_port,
                                                     metrics_prefix=metrics_prefix,
                                                     listen_address=metrics_listen_address)
--- a/nucypher/utilities/prometheus/collector.py
+++ b/nucypher/utilities/prometheus/collector.py
@@ -18,7 +18,7 @@
 try:
     from prometheus_client import Gauge, Enum, Counter, Info, Histogram, Summary
 except ImportError:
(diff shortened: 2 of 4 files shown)


derekpierre:
@KPrasch any objection to adding `prometheus_client` to dev-requirements?

KPrasch:
It's already in the ursula extra `pip install nucypher[ursula]` - I suppose we can tack it into the `dev` extra.

derekpierre:
After chatting with @KPrasch - since `prometheus_client` is installed via `pip install nucyher[dev]` there is no need to add it to the dev-packages in the Pipfile.

Taking this issue off of draft mode.

nucypher/cli/commands/ursula.py
```diff
@@ -382,8 +381,10 @@ def run(general_config, character_options, config_file, interactive, dry_run, me
                                                                config_file=config_file,
                                                                json_ipc=general_config.json_ipc)
 
-    prometheus_config = None
+    prometheus_config: 'PrometheusMetricsConfig' = None
     if prometheus:
+        # Local scoped to help prevent import without prometheus installed
```
KPrasch, line 386: :+1:

nucypher/utilities/prometheus/collector.py
```diff
@@ -30,6 +29,7 @@
 from nucypher.blockchain.eth.interfaces import BlockchainInterfaceFactory
 from nucypher.blockchain.eth.registry import BaseContractRegistry
 
+from prometheus_client.metrics import MetricWrapperBase
```
KPrasch, line 32: Do you want to go as far as wrapping this for `ImportError` and re-raising `DevelopmentInstallationRequired`?

derekpierre: Do you mean on L21 - the check is performed there, this is just an additional import.

We could throw a specific exception or perhaps just provide a more specific message. However, `DevelopmentInstallationRequired` might not tell the whole story since it is not solely a dev installation problem because `pip install nucypher[ursula]` would install `prometheus_client`.

KPrasch: True, and since it's only a single dependency it's not worth introducing specific guidance?

derekpierre: I guess the real goal for `pip install nucypher[ursula]` is to install optional ursula packages - currently only `prometheus`? I only ask because the ursula directive only installs prometheus so we could just add that to the message - but will the list of what get installed for that directive be added to over time?

An alternative is to point to prometheus docs (yet to be written 😅). 

Perhaps, for now, we just add the `pip install nucypher[ursula]` to skip the extra bloat of `dev` - wdyt? Although docs should still be written haha.

KPrasch: > I guess the real goal for pip install nucypher[ursula] is to install optional ursula packages
:+1: - Yep - tying to scope the package as thin and purpose-built as possible, this being the first step.

> Perhaps, for now, we just add the pip install nucypher[ursula] to skip the extra bloat of dev - wdyt? Although docs should still be written haha.

Yes, and yes. :-)

derekpierre: ✔️

PR #2101, opened 2020-06-17
Correct usage of TxReceipt.data during grant

vepkenez:
Some upstream library/protocol change changed the structure of the transaction AttributeDict and broke our  `fetch_arrangement_addresses_from_policy_txid` method.

This commit fixes that.

### DIFF
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -237,7 +237,7 @@ class StakingEscrowAgent(EthereumContractAgent):
     def get_staker_population(self) -> int:
         """Returns the number of stakers on the blockchain"""
         return self.contract.functions.getStakersLength().call()
-    
+
     @contract_api(CONTRACT_CALL)
     def get_current_period(self) -> Period:
         """Returns the current period"""
@@ -321,7 +321,7 @@ class StakingEscrowAgent(EthereumContractAgent):
     #
     # StakingEscrow Contract API
     #
-    
+
     @contract_api(CONTRACT_CALL)
     def get_global_locked_tokens(self, at_period: Optional[Period] = None) -> NuNits:
         """
@@ -435,7 +435,7 @@ class StakingEscrowAgent(EthereumContractAgent):
                       dry_run: bool = False,
                       gas_limit: Optional[Wei] = None
                       ) -> Union[TxReceipt, Wei]:
-        
+
         min_gas_batch_deposit: Wei = Wei(250_000)  # TODO: move elsewhere?
         if gas_limit and gas_limit < min_gas_batch_deposit:
             raise ValueError(f"{gas_limit} is not enough gas for any batch deposit")
@@ -618,7 +618,7 @@ class StakingEscrowAgent(EthereumContractAgent):
     @contract_api(CONTRACT_CALL)
     def staking_parameters(self) -> StakingEscrowParameters:
         parameter_signatures = (
-            
+
             # Period
             'secondsPerPeriod',  # Seconds in single period
 
@@ -806,7 +806,10 @@ class PolicyManagerAgent(EthereumContractAgent):
         # TODO: Won't it be great when this is impossible?  #1274
         _receipt = self.blockchain.client.wait_for_receipt(txhash, timeout=timeout)
         transaction = self.blockchain.client.w3.eth.getTransaction(txhash)
-        _signature, parameters = self.contract.decode_function_input(transaction.data)
+        try:
+            _signature, parameters = self.contract.decode_function_input(self.blockchain.client.parse_transaction_data(transaction))
+        except AttributeError:
+            raise RuntimeError(f"Eth Client incompatibility issue: {self.blockchain.client} could not extract data from {transaction}")
         return parameters['_nodes']
 
     @contract_api(TRANSACTION)
@@ -1229,7 +1232,7 @@ class WorkLockAgent(EthereumContractAgent):
     def check_claim(self, checksum_address: ChecksumAddress) -> bool:
         has_claimed: bool = bool(self.contract.functions.workInfo(checksum_address).call()[2])
         return has_claimed
-    
+
     #
     # Internal
     #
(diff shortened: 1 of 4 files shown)


nucypher/datastore/datastore.py
```diff
@@ -183,7 +183,8 @@ def attach_kfrag_to_saved_arrangement(self, alice, id_as_hex, kfrag, session=Non
         if policy_arrangement.alice_verifying_key.key_data != alice.stamp:
             raise alice.SuspiciousActivity
 
-        policy_arrangement.kfrag = bytes(kfrag)
```
vepkenez, line 186: this is consistently and reproducibly not actually happening in live contexts.

nucypher/network/server.py
```diff
@@ -294,12 +294,10 @@ def set_policy(id_as_hex):
         if not kfrag.verify(signing_pubkey=alices_verifying_key):
             raise InvalidSignature("{} is invalid".format(kfrag))
```
vepkenez, line 296: this is an experiment and didn't work...

nucypher/blockchain/eth/agents.py
```diff
@@ -806,7 +806,7 @@ def fetch_arrangement_addresses_from_policy_txid(self, txhash: Union[str, bytes]
         # TODO: Won't it be great when this is impossible?  #1274
         _receipt = self.blockchain.client.wait_for_receipt(txhash, timeout=timeout)
         transaction = self.blockchain.client.w3.eth.getTransaction(txhash)
-        _signature, parameters = self.contract.decode_function_input(transaction.data)
+        _signature, parameters = self.contract.decode_function_input(transaction.input)
```
KPrasch, line 809: Im trying to isolate the upstream API version change to reference.

nucypher/datastore/datastore.py
KPrasch: This appears to be the syntax of FlaskSQLAlchemy update, however, FlaskSQLAlchemy is not implemented here :thinking:

nucypher/datastore/datastore.py
```diff
@@ -183,7 +183,8 @@ def attach_kfrag_to_saved_arrangement(self, alice, id_as_hex, kfrag, session=Non
...
+        # TODO:  is this ok?
+        session.query(PolicyArrangement).filter_by(id=id_as_hex.encode()).update({"kfrag": bytes(kfrag)})
```
KPrasch, line 187: Despite this change, the fragment is still not being persisted.  Are we missing a call to session.add ?

KPrasch: Let's remove these part of the chagseset for this PR to focus on the other the API change, leaving the DB handling for follow-up WRT #2088

nucypher/datastore/datastore.py
KPrasch: related to #2088

nucypher/blockchain/eth/agents.py
KPrasch: Ok here is a small lead: <https://web3py.readthedocs.io/en/latest/web3.eth.html#web3.eth.Eth.getTransaction>

This call executes an RPC call `eth_getTransaction`, which may need to be composed into the client layer for cross-compatibility.

vepkenez: hmm well, these docs show no "data" key, which... seems consistent.

KPrasch: As we've seen before, different ethereum client can have slightly different return values.  As a matter of design, all of our RPC calls are routed through clients, but this one bypasses that compatibility layer resulting in the CI failure.

cygnusv: I'd say that this is something that the Web3 layer should handle. Not that we shouldn't do nothing, but I think it's an inconsistency on their side. If we can confirm this is the case, we should open an issue there.

KPrasch: I'd argue that we already accommodate eth-tester in this regard, perhaps the issue is directed there?

vepkenez: once I get the tests passing, I was planning on building this into our Clients which should be less janky than what I'm doing.

KPrasch: :+1: - Seems like those two are actually the same effort!

vepkenez:
I made a big merge mistake here... gonna force push again :|

nucypher/blockchain/eth/clients.py
```diff
@@ -529,6 +532,9 @@ def unlock_account(self, *args, **kwargs) -> bool:
     def sync(self, *args, **kwargs) -> bool:
         return True
```
vepkenez, line 534: Maybe it was just the infura client that has this different data format...

nucypher/blockchain/eth/clients.py
```diff
@@ -428,6 +428,9 @@ def check_for_timeout(t):
 
         return True
 
+    def parse_transaction_data(self, transaction):
+        return transaction.data
```
cygnusv, line 432: I think the correct field is `data`. It's what web3py, web3js and infura mention in their docs.
EthereumTester, however, has an open issue for this very same problem: https://github.com/ethereum/eth-tester/issues/173

I think we should use `input` here, and `data` in `EthereumTesterClient`
```suggestion
        return transaction.input
```

cygnusv: Maybe we should consider fixing other places where this is incorrectly used:

* `eth/multisig.py, L46, L99?, L109?`
* somewhere else?

vepkenez: yeah... I added a TODO.  The way Proposals are created needs to be refactored a bit maybe, to do this correctly.

nucypher/blockchain/eth/clients.py
```diff
@@ -585,6 +588,9 @@ def sign_message(self, account: str, message: bytes) -> str:
         signature_and_stuff = Account.sign_message(signable_message=signable_message, private_key=signing_key)
         return signature_and_stuff['signature']
 
+    def parse_transaction_data(self, transaction):
+        return transaction.data
```
cygnusv, line 592:
```suggestion
        return transaction.data  # TODO: See https://github.com/ethereum/eth-tester/issues/173
```

PR #2131, opened 2020-07-13
General Purpose Static File service

vepkenez:
In this PR, Ursula and Felix are configured to serve static files out of the /statics directory at /nucypher.
This could be used for serving data files, javascript, css... whatever

### DIFF
--- a/nucypher/characters/chaotic.py
+++ b/nucypher/characters/chaotic.py
@@ -44,9 +44,10 @@ from nucypher.characters.base import Character
 from nucypher.config.constants import MAX_UPLOAD_CONTENT_LENGTH, TEMPLATES_DIR
 from nucypher.crypto.powers import SigningPower, TransactingPower
 from nucypher.datastore.threading import ThreadedSession
+from nucypher.network.nodes import FleetStateTracker
+from nucypher.network.server import NonTLSHost
 
-
-class Felix(Character, NucypherTokenActor):
+class Felix(Character, NucypherTokenActor, NonTLSHost):
     """
     A NuCypher ERC20 faucet / Airdrop scheduler.
 
@@ -281,9 +282,7 @@ class Felix(Character, NucypherTokenActor):
 
         self.start_time = maya.now()
         payload = {"wsgi": self.rest_app, "http_port": port}
-        deployer = HendrixDeployWithStatics(action="start", options=payload)
-        deployer.resources = get_statics()
-        click.secho(f"Running {self.__class__.__name__} on {host}:{port}")
+        deployer = self.get_deployer(host, port, options=payload)
 
         if distribution is True:
             self.start_distribution()
--- a/nucypher/characters/chaotic.py
+++ b/nucypher/characters/chaotic.py
@@ -281,7 +281,9 @@ class Felix(Character, NucypherTokenActor):
 
         self.start_time = maya.now()
         payload = {"wsgi": self.rest_app, "http_port": port}
-        deployer = HendrixDeploy(action="start", options=payload)
+        deployer = HendrixDeployWithStatics(action="start", options=payload)
+        deployer.resources = get_statics()
+        click.secho(f"Running {self.__class__.__name__} on {host}:{port}")
 
         if distribution is True:
             self.start_distribution()
--- a/nucypher/characters/chaotic.py
+++ b/nucypher/characters/chaotic.py
@@ -44,7 +44,6 @@ from nucypher.characters.base import Character
 from nucypher.config.constants import MAX_UPLOAD_CONTENT_LENGTH, TEMPLATES_DIR
 from nucypher.crypto.powers import SigningPower, TransactingPower
 from nucypher.datastore.threading import ThreadedSession
-from nucypher.network.nodes import FleetStateTracker
 from nucypher.network.server import NonTLSHost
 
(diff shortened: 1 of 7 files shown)


nucypher/network/protocols.py
```diff
@@ -96,4 +116,128 @@ def __radd__(self, other):
...
+        ThreadPoolService(self.threadpool).setServiceParent(self)
+
+        # create the base resource and add any additional static resources
+        resource = HendrixResourceWithStatics(reactor, self.threadpool, application, loud=loud)
+        if resources:
+            resources = sorted(resources, key=lambda r: r.namespace)
+            for res in resources:
+                if hasattr(res, 'get_resources'):
+                    for sub_res in res.get_resources():
+                        resource.putNamedChild(sub_res)
```
KPrasch, line 194: woah nested!

vepkenez: recurse this!

nucypher/config/constants.py
```diff
@@ -62,8 +62,9 @@
 # Web
 CLI_ROOT = NUCYPHER_PACKAGE / 'network' / 'templates'
 TEMPLATES_DIR = CLI_ROOT / 'templates'
+STATICS_DIR = os.getenv('NUCYPHER_STATICS_DIR') or os.path.join(
```
derekpierre, line 65: You can try your hand at `pathlib.Path` like the other constants above 😄 - https://docs.python.org/3/library/pathlib.html. Example,
```python
pathlib.Path(__file__) / 'statics'
```

nucypher/characters/chaotic.py
```diff
@@ -44,9 +44,9 @@
 from nucypher.config.constants import MAX_UPLOAD_CONTENT_LENGTH, TEMPLATES_DIR
 from nucypher.crypto.powers import SigningPower, TransactingPower
 from nucypher.datastore.threading import ThreadedSession
+from nucypher.network.server import NonTLSHost
 
-
-class Felix(Character, NucypherTokenActor):
+class Felix(Character, NucypherTokenActor, NonTLSHost):
```
KPrasch, line 49: Can `NonTLSHost` simply be instantiated inside the `__init__`?  I'm unsure I agree that this entity is accurately describe through inheritance, especially when considering the other Members being inherited.  

To be fair, with a bit of work, both `Character`  and `NucypherTokenActor` can be removed from `Felix` - then I don't mind the `NonTLSHost` parent so much.

nucypher/datastore/keypairs.py
```diff
@@ -190,7 +190,7 @@ def __init__(self,
         self.certificate_filepath = certificate_filepath
 
     def get_deployer(self, rest_app, port):
-        return HendrixDeployTLS("start",
+        return DeployTLSWithStatics("start",
```
KPrasch, line 193: What made you decide to create a new object, instead of a feature switch kwarg on `HendrixDeployTLS`?

nucypher/network/protocols.py
KPrasch: Yes - please extract to another function to help break up this nesting!

nucypher/network/protocols.py
```diff
@@ -96,4 +116,128 @@ def __radd__(self, other):
         return bytes(other) + bytes(self)
 
     def __repr__(self):
-        return self.uri
+	    return self.uri
+
+
+class HendrixResourceWithStatics(HendrixResource):
```
KPrasch, line 122: Much of this code feels out of scope to the nucypher codebase and more aligned with hendrix itself.

nucypher/network/protocols.py
```diff
@@ -96,4 +116,128 @@ def __radd__(self, other):
...
+    def getChild(self, name, request):
+
+        path = name.decode('utf-8')
+        if path in self.children:
+            return self.children[path]
+
+        return super().getChild(name, request)
+
+
+class HendrixTCPServiceWithTLSAndStatics(HendrixTCPServiceWithTLS):
```
KPrasch, line 133: This class name reads like a conditional to me!

nucypher/characters/chaotic.py
vepkenez: I put that there just to be clear that Felix was non TLS.

nucypher/datastore/keypairs.py
vepkenez: I don't remember but I did this about a year ago and probably wasn't as confident in making large changes to Nucypher or Hendrix at the time...

nucypher/network/protocols.py
vepkenez: Yeah.  If you want I can add it to Hendrix and we can take it out of here.  I had comments that alluded to doing that someday.  Or we can merge it and use it and then do that as a separate issue.

KPrasch:
Considering we are essentially enabling Ursula to execute JavaScript - how much process isolation is required for safety? 

Do we need to run Ursula's monitoring services on another Process entirely?

vepkenez:
> Considering we are essentially enabling Ursula to execute JavaScript - how much process isolation is required for safety?
> 
> Do we need to run Ursula's monitoring services on another Process entirely?

It's not really going to execute anything, just serve files that allow browsers to do more interesting things.  At the end of the day, the same HTTP endpoints will be exposed.

PR #2135, opened 2020-07-14
Ensure Prometheus metric for CommitmentMade is correctly updated on startup

### LINKED ISSUE #2129: Prometheus metrics for CommitmentMade/ActivityConfirmed can sometimes be innaccurate
Related to #1923, #2069 

**Describe the Bug**
I think the metrics for "Activity Confirmed" metric may be off - it relies on the `CommitmentMade` event, but if I made a commitment before running with prometheus/or after restarting it won't get updated until a subsequent `CommitmentMade` event is generated for the subsequent period occurs - I'll file a bug for that

**To Reproduce**
Restart a node that made a commitment for the next period. On the dashboard the node will be categorized as not confirmed until it makes a commitment again for the subsequent period.

 - OS Version: -
 - Nucypher Version: 2.1.0b12
 - Ethereum Node Version:  -

derekpierre:
Fixes #2129 .

### DIFF
--- a/nucypher/utilities/prometheus/collector.py
+++ b/nucypher/utilities/prometheus/collector.py
@@ -325,7 +325,8 @@ class CommitmentMadeEventMetricsCollector(EventMetricsCollector):
             # however, 'staker' and 'period' filters should help reduce load
             initial_event_filter = self.contract_agent.contract.events[self.event_name].createFilter(
                 fromBlock=0, argument_filters=arg_filters)
-            events = initial_event_filter.get_new_entries()
+            events = initial_event_filter.get_all_entries()
+            # we expect only one event to be returned due to period filter
             for event in events:
                 self._event_occurred(event)
 
--- a/nucypher/utilities/prometheus/collector.py
+++ b/nucypher/utilities/prometheus/collector.py
@@ -305,10 +305,35 @@ class EventMetricsCollector(BaseMetricsCollector):
         return f'{self.event_name}_{arg_name}'
 
 
+class CommitmentMadeEventMetricsCollector(EventMetricsCollector):
+    """Collector for CommitmentMade event."""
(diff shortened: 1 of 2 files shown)


nucypher/utilities/prometheus/collector.py
```diff
@@ -305,10 +305,36 @@ def _get_arg_metric_key(self, arg_name: str):
...
+            last_committed_period = self.contract_agent.get_last_committed_period(staker_address=self.staker_address)
+            arg_filters = {'staker': self.staker_address, 'period': last_committed_period}
+            # check from the beginning - difficult to determine blocknumber for previous period
+            # however, 'staker' and 'period' filters should help reduce load
+            initial_event_filter = self.contract_agent.contract.events[self.event_name].createFilter(
+                fromBlock=0, argument_filters=arg_filters)
+            events = initial_event_filter.get_all_entries()
+            # we expect only one event to be returned due to period filter
+            for event in events:
+                self._event_occurred(event)
```
vzotova, line 331: Method looks good. But it possible to have multiple events for one period. Not sure is it an issue or not

derekpierre: ✔️

nucypher/utilities/prometheus/metrics.py
```diff
@@ -214,16 +218,17 @@ def create_staking_events_metric_collectors(ursula: 'Ursula', metrics_prefix: st
     collectors: List[MetricsCollector] = []
     staking_agent = ContractAgency.get_agent(StakingEscrowAgent, registry=ursula.registry)
 
+    staker_address = ursula.checksum_address
```
KPrasch, line 221: This is subject to change, and will likely cause this collector to break when it does.

Issue #2124, opened 2020-07-07
Viral spread for incidental payloads


jMyles:
Presently, only qualifying nodes (per `matching_nodes_among(some_nodes, ...)`) *at grant time* receive a Treasure Map.  It makes sense to also enlist nodes to spread incidental payloads to other qualifying nodes later in their runtime.

jMyles:
This issue is totally unimportant until suddenly it's very important.  I'm inclined to put this in the "Before Activating" milestone.

cygnusv:
As a follow-up of my comment here (https://github.com/nucypher/nucypher/pull/1741#discussion_r482238973), I have this idea for an alternative method for `Bob.matching_nodes`:

```python
def matching_nodes(self, nodes: FleetSensor):

    bob_bytes = bytes(self.public_keys(DecryptingPower))

    def node_index(canonical_address: bytes) -> int:
        digest = SHA256(bob_bytes + canonical_address)  # potentially some metadata too (e.g. label?)
        index = int.from_bytes(digest, 'big')
        return index

    indexed_nodes = sorted([(node_index(node.canonical_public_address), node) for node in nodes])
    target_nodes = list(zip(*indexed_nodes))[1]
    return target_nodes
```
Instead of giving you a selection of nodes, this would give you a _deterministic ordering_ of nodes, where you're guaranteed that the TMap is in some of the first nodes of that list (assuming there's still one of these original Ursulas online, obviously). This would work even as the network grows, at the expense of decreasing the success probability for Bob per Ursula (i.e, if the network is 10 times bigger, then Bob will get a 10x longer list, with ~9 false positives in that list per correct Ursula).

jMyles:
1) I don't like the idea of adding policy-dependent metadata (such as the label), for reasons I describe here: https://hackmd.io/GEpM-Sr8SR6QzVtXYOQj3g

2) I don't like the idea of Bob needing to perform an additional SHA256 for every node on the network at retrieve time.

3) This method gives a list of *all the nodes*.  To how many nodes does Alice deliver in this scenario?

cygnusv:
Random failure related to current search-boundary approach:
https://app.circleci.com/pipelines/github/nucypher/nucypher/6086/workflows/0eb2e11a-0504-4610-83a0-dc6b09188156/jobs/105465

KPrasch:
I do not think this issue is oily rags.

cygnusv:
[...]

PR #2137, opened 2020-07-14
Three small bugfixes.

jMyles:


nucypher/characters/base.py
```diff
@@ -340,6 +337,14 @@ def from_public_keys(cls,
 
         return cls(is_me=False, crypto_power=crypto_power, *args, **kwargs)
 
+    def _set_known_node_class(self, known_node_class, federated_only):
+        if not known_node_class:
+            # Once in a while, in tests or demos, we init a plain Character who doesn't already know about its node class.
+            from nucypher.characters.lawful import Ursula
+            known_node_class = Ursula
```
tuxxy, line 344: Refresh me -- in what cases can we presume there may exist known nodes who _are not Ursulas_?

tuxxy:
Are these bugs/issues enumerated somewhere, or are these local to Treasure Island?

examples/run_demo_ursula_fleet.py
```diff
@@ -0,0 +1,73 @@
...
+        reactor.run()  # GO!
+    finally:
+        with suppress(FileNotFoundError):
+            os.remove("sage.db")
+        for u in ursulas[1:]:
+            with suppress(FileNotFoundError):
+                os.remove(f"{u.rest_interface.port}.db")
+
+if __name__ == "__main__":
+    spin_up_federated_ursulas()
```
derekpierre, line 73: If the intention of this script is just for demos, 12 seems like overkill given that the demos do 2-of-3 thresholds. Perhaps we could reduce the default and allow a command-line argument to be specified to override the default quantity (whatever it ends up being). Thoughts?

nucypher/characters/base.py
jMyles: None, but building ourselves into a box (by not having a class-level namespace used as an interface for the known node class) is an awful idea, right?

examples/run_demo_ursula_fleet.py
jMyles: Yeah, I don't disagree.  The thing is: Alice and Bob now bail on trying to publish / find the `TreasureMap` if there are less than 8 nodes (I chose 8 arbitrarily).  If we *don't* do this, then it's possible for Alice to publish to too few nodes (as there may be too few which are `matching_nodes_among(nodes_she_knows_about_at_that_moment)`, leading to a broken Policy.

It also seems to me that nodes, in federated mode anyway, are super cheap; this script spins up 12 of them no problem, but can also do 100 without issue on my laptop.

jMyles: ...but I do basically agree with you.  Part of this is that #2124 needs to be closed.  Then I'll have less anxiety about publishing to so few nodes.

nucypher/characters/lawful.py
```diff
@@ -1624,7 +1624,7 @@ def __init__(self, policy_encrypting_key=None, controller: bool = True, *args, *
 
         # Encrico never uses the blockchain, hence federated_only)
         kwargs['federated_only'] = True
-        kwargs['known_node_class'] = Ursula
+        kwargs['known_node_class'] = None
```
KPrasch, line 1627: `None`?

nucypher/network/nodes.py
```diff
@@ -759,30 +776,31 @@ def learn_from_teacher_node(self, eager=False):
...
-                              f'{sprout} is not bonded to a Staker.')
-
-            except sprout.Invalidsprout:
-                self.log.warn(sprout.invalid_metadata_message.format(sprout))
-
-            except sprout.SuspiciousActivity:
-                message = f"Suspicious Activity: Discovered sprout with bad signature: {sprout}." \
-                          f"Propagated by: {current_teacher}"
-                self.log.warn(message)
+            # TODO: This whole section is weird; sprouts down have any of these things.
```
KPrasch, line 779: What do you think needs to happen with the exception handling here - is it duplicated?

tests/integration/learning/test_discovery_phases.py
```diff
@@ -193,3 +202,7 @@ def find_successful_responses(map_publication_responses):
         # But with debuggers and other processes running on laptops, we give a little leeway.
         assert initial_blocking_duration.total_seconds() < 3
         assert complete_distribution_time.total_seconds() < 10
+
+        # Takes about .8 on jMyles' laptop; still the bulk of init time.
+        # Assuming this is 10 percent of the fleet, that's another 8 second savings for first policy enactment.
+        assert maturation_time.total_seconds() < 2
```
KPrasch, line 208: pre-commit hooks not running?

nucypher/characters/lawful.py
jMyles: Doesn't that seem about right now `Enrico`?

tests/integration/learning/test_discovery_phases.py
jMyles: Yeah, It think I missed a memo - what do I do?

nucypher/network/nodes.py
jMyles: I honestly don't know.  I've intentionally kinda left it out of my mind for the moment in an effort to stay focused on this PR; it needs to be addressed in a later one, along with a generalized high-level effort toward optimal logging and error handing.

examples/run_demo_ursula_fleet.py
```diff
@@ -0,0 +1,73 @@
...
+from twisted.internet import reactor
+
+from nucypher.characters.lawful import Ursula
+
+FLEET_POPULATION = 12
+DEMO_NODE_STARTING_PORT = 11500
+
+ursula_maker = partial(Ursula, rest_host='127.0.0.1',
+                       federated_only=True,
+                       domains=["finnegans/heartbeat/whatever"]
```
KPrasch, line 32: Of course this implies also changing the Ursula test fleet's domain depending on which test is being run, which is a less generic configuration that before - Are there CI or ops follow ups associated with supporting a local fleet on another domain?

examples/run_demo_ursula_fleet.py
```diff
@@ -0,0 +1,73 @@
...
+                       )
+
+
+def spin_up_federated_ursulas(quantity: int = FLEET_POPULATION):
```
KPrasch, line 36: Not using docker - so a full development installation is needed, correct?

jMyles: :thinking: is that so?  I can't immediately think of a requirement that's only satisfied in the dev installation.  

It doesn't use pytest, doesn't require appdirs, doesn't use prometheus... 

But it's really only meant for development anyway, so it won't bother me if it does.

nucypher/characters/lawful.py
KPrasch: Yeah, I guess so.

examples/run_demo_ursula_fleet.py
KPrasch: Sounds like this is working with the new script - but we also need to maintain compatibility with the docker fleets - Have you been able to run them?  

There may be additional footwork needed to accommodate this properly (surely out of scope for "Three small bugfixes").

nucypher/characters/lawful.py
```diff
@@ -1624,7 +1624,7 @@ def __init__(self, policy_encrypting_key=None, controller: bool = True, *args, *
 
         # Encrico never uses the blockchain, hence federated_only)
```
cygnusv, line 1635:
```suggestion
        # Enrico never uses the blockchain, hence federated_only
```

nucypher/blockchain/eth/actors.py
```diff
@@ -1299,7 +1299,7 @@ def __init__(self,
                 self.stakes = StakeList(registry=self.registry, checksum_address=self.checksum_address)
                 self.stakes.refresh()
                 self.work_tracker = work_tracker or WorkTracker(worker=self)
-                self.work_tracker.start(act_now=False)
+                self.work_tracker.start(act_now=start_working_now)
```
cygnusv, line 1302: There's another invocation of `work_tracker.start()` as part of `Ursula.run()`. How does this interplay with the call in this line?
https://github.com/nucypher/nucypher/blob/523ea089f43581dda28e6b6a337a019d63a13e10/nucypher/characters/lawful.py#L1133-L1136

scripts/circle/docker-compose.yml
```diff
@@ -43,7 +43,7 @@ services:
     image: circle:nucypher
     depends_on:
       - circleursula2
```
cygnusv, line 45:
```suggestion
      - circleursula1
```
For consistency

tests/integration/cli/actions/test_select_client_account_for_staking.py
```diff
@@ -20,8 +20,10 @@
 from nucypher.cli.literature import PREALLOCATION_STAKE_ADVISORY
 from nucypher.config.constants import TEMPORARY_DOMAIN
 from tests.constants import YES
+import pytest
 
 
+@pytest.mark.usefixtures('mock_contract_agency')
```
cygnusv, line 26: A reminder on my comment yesterday on Discord about `usefixtures`:
> Well, sorry for nitpicking, but I also don't have usefixtures in very high estime, for two reasons: 1. Since the fixture name is in a string, it makes refactoring more difficult, and 2. It doesn't work on top of  another fixture
> As a bonus, for some time we had a typo in some of them and we had usefixture instead and pytest didn't complain. This is of course not a problem of usefixtures but it does activate my spidersense
Basically, it makes me think there's no benefit of using it instead of simply passing the fixture name as an argument

PR #2154, opened 2020-07-22
Improved gas price logic. Introduction of gas oracles

cygnusv:
- Gas strategies were being ignored when building the transaction. This PR fixes that.
- Sets web3's `fast` as the default gas strategy.
- Web3 gas strategies are overly expensive to compute when using Infura/Alchemy. 1 or 2 minutes to get a gas price. This PR introduces gas price oracles to solve that.
- Fixes to tests as a consequence of the gas price problems (since gas strategies were being ignored), as well as to other problems in the BlockchainInterface and EthereumClient layers.


nucypher/utilities/oracles.py
```diff
@@ -0,0 +1,120 @@
...
+
+    @classmethod
+    def construct_gas_strategy(cls):
+        def oracle_based_gas_price_strategy(web3: Web3, transaction_params: TxParams = None) -> Wei:
+            oracle = cls()
+            gas_price = oracle.get_gas_price()
+            return gas_price
+        return oracle_based_gas_price_strategy
+
+    def __repr__(self):
```
derekpierre, line 72: Any reason we didn't put this method in the `Oracle` base class?

tests/unit/test_oracles.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def test_etherchain():
+    oracle = EtherchainGasPriceOracle()
+    with patch.object(oracle, '_probe_oracle'):
+        oracle._raw_data = etherchain_json
+        assert oracle.get_gas_price('safeLow') == Web3.toWei(99.0, 'gwei')
+        assert oracle.get_gas_price('standard') == Web3.toWei(105.0, 'gwei')
+        assert oracle.get_gas_price('fast') == Web3.toWei(108.0, 'gwei')
+        assert oracle.get_gas_price('fastest') == Web3.toWei(119.9, 'gwei')
```
derekpierre, line 82: Should we use `EtherchainGasPriceOracle._speed_names` here for testing? and `._default_speed` in the line below.

Same for UpvestOracle.

nucypher/utilities/oracles.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+def oracle_fallback_gas_price_strategy(web3: Web3, transaction_params: TxParams = None) -> Wei:
+    oracles = (EtherchainGasPriceOracle, UpvestGasPriceOracle)
+
+    for gas_price_oracle_class in oracles:
+        try:
+            gas_strategy = gas_price_oracle_class.construct_gas_strategy()
+            gas_price = gas_strategy(web3, transaction_params)
+        except EtherchainGasPriceOracle.OracleError:
```
derekpierre, line 109: Did you mean: 
```suggestion
        except Oracle.OracleError:
```

nucypher/utilities/oracles.py
cygnusv: No reason... Bad refactoring :P ... I'll change it

cygnusv: ✔️

tests/unit/test_oracles.py
cygnusv: Added an assertion of `_speed_names` being a subset of the expected speed names. Wrt the default speed, it's implicitly tested when `get_gas_price` is called without arguments, but I added an assertion of being also part of the expected speed names.

nucypher/blockchain/eth/interfaces.py
```diff
@@ -267,18 +269,28 @@ def attach_middleware(self):
...
+        # Gas Price Strategy:
+        # Bundled web3 strategies are too expensive for Infura (it takes ~1 minute to get a price),
+        # so we use external gas price oracles, instead (see #2139)
+        if isinstance(self.client, InfuraClient):
+            gas_strategy = oracle_fallback_gas_price_strategy
+        else:
+            gas_strategy = self.gas_strategy
+        self.client.set_gas_strategy(gas_strategy=gas_strategy)
+        gwei_gas_price = Web3.fromWei(self.client.generate_gas_price(), 'gwei')
+        self.log.debug(f"Currently, our gas strategy returns a gas price of {gwei_gas_price} gwei")
```
vzotova, line 289: Is it possible to add gas strategy name here? Because overwriting for Infura clients could confuse for the first time

nucypher/utilities/oracles.py
```diff
@@ -0,0 +1,120 @@
...
+        gas_price_wei = Wei(self.gas_prices[speed])
+        return gas_price_wei
+
+    @classmethod
+    def construct_gas_strategy(cls):
+        def oracle_based_gas_price_strategy(web3: Web3, transaction_params: TxParams = None) -> Wei:
+            oracle = cls()
+            gas_price = oracle.get_gas_price()
+            return gas_price
+        return oracle_based_gas_price_strategy
```
vzotova, line 73: Creation instance each call looks too much, but not sure if possible to make only one instance for all calls

nucypher/utilities/oracles.py
```diff
@@ -0,0 +1,120 @@
...
+        self._raw_data = response.json()
+
+    def __repr__(self):
+        return f"{self.name} ({self.api_url})"
+
+
+class EthereumGasPriceOracle(Oracle):
+    """Base class for Ethereum gas price oracles"""
+
+    _speed_names = NotImplemented
```
vzotova, line 54: `_speed_names` is unused in oracle ierarchy (or I missed it), how about add checking for names in `get_gas_price`?

nucypher/blockchain/eth/interfaces.py
cygnusv: The problem is that gas strategies themselves don't have names, as they're plain functions.

nucypher/utilities/oracles.py
cygnusv: It will only be called once. This is similar to what web3 does to create their bundled gas strategies.

nucypher/utilities/oracles.py
```diff
@@ -0,0 +1,120 @@
...
+    api_url = "https://www.etherchain.org/api/gasPriceOracle"
+    _speed_names = ('safeLow', 'standard', 'fast', 'fastest')
+    _default_speed = 'fast'
+
+    def _parse_gas_prices(self):
+        self._probe_oracle()
+        self.gas_prices = {k: int(Web3.toWei(v, 'gwei')) for k, v in self._raw_data.items()}
+
+
+class UpvestGasPriceOracle(EthereumGasPriceOracle):
```
tuxxy, line 89: I don't really see a reason for this design pattern to be separated per API.

It makes more sense to me that you would have a single gas price oracle which fetches from multiple sources. Why not just combine these two classes into one gas price oracle?

tuxxy:
@cygnusv As we briefly chatted about, I think we could refer to these classes as `Datafeed`s rather than Oracles. Do you have any opinions on that?

cygnusv:
> @cygnusv As we briefly chatted about, I think we could refer to these classes as `Datafeed`s rather than Oracles. Do you have any opinions on that?

After giving it more thought, I think the term "Gas price oracle" is fairly common:

https://github.com/ethereum/go-ethereum/wiki/Gas-Price-Oracle
https://www.etherchain.org/tools/gasPriceOracle
https://medium.com/@ethgasstation/gas-price-oracles-and-the-future-of-ethgasstation-859a55c57ad5

I agree, however, that the base class `Oracle` should be renamed to `Datafeed` or something similar.

nucypher/blockchain/eth/interfaces.py
fjarri: I guess you could use `.__name__`. Or functors instead of functions if names are really important.

vzotova (approved):

nucypher/utilities/oracles.py
arjunhassard: Seems like upvest has a more sophisticated feed that handles price spikes and source degradation. @cygnusv is one of these the default oracle to get a price more cheaply? Are there compatibility differences with Infura vs Alchemy?

arjunhassard (approved):
Very important tool! 👌

nucypher/utilities/oracles.py
cygnusv: This PR creates the possibility of using this data feed, but it requires more work to make it more customizable. Right now, what it does is:
- When using a normal provider, the default is to use the "fast" web3py strategy. You can still configure it manually to use a different web3py strategy (but not the ones from the data feed)
- When using Infura, the default is to use a chain of data feeds, starting from Etherchain, then Upvest, and then the default of the provider (which relies on their own ETH nodes). In this case, you can't change the strategy manually.

So, this PR is somewhat very opinionated on the matter, but basically because I didn't want to spend too much time building a truly generic and customizable mechanism. We can do that in a follow-up PR.

PR #2158, opened 2020-07-27
`create`/`increase`: flag to use unlocked tokens, `merge` command in `stake`

### LINKED ISSUE #1766: Add way to increase locked amount with a value less than min
Current idea:
Split target sub-stake into old one and another one. Old one is only for honest reward calculation. New amount will be add to new sub-stake.

### LINKED ISSUE #2015: StakingEscrow mergeStake function
This is a feature request.

Essentially I'm requesting a function to revert effects of `divideStake`. Suppose that I divided some sub-stake and prolonged one of the parts. Then, situation changed and I want to prolong the another part too (resume staking in full). I can prolong it via `prolongStake`, although the number of sub-stakes will still be increased. I risk hitting the `MAX_SUB_STAKES` limit eventually.
I'd be helpful to be able to merge two sub-stakes with the same `periods` into one.

Obviously, a regular user is unlikely to experience this kind of trouble. The intended users of the feature are some automated smart-contract based stacking solutions.

### LINKED ISSUE #2065: Weird UX for `nucypher stake create` when no NU is available in account
**Describe the Bug**

I tried to create a stake without having any NU i.e. 0 NU, and was still prompted to enter a stake value (weird), and then I was allowed to enter a value of 15000 NU even though I had 0 NU, also weird. Eventually, after enterring other values and accepting the terms and conditions a `nucypher.blockchain.eth.actors.InsufficientTokens: Insufficient token balance` exception was raised.

This exception could probably be raised earlier, or we can abort earlier in the CLI.  We should probably check the token balance before prompting the user to enter a stake value, and we should check our prompting logic to ensure that the lower limit (15000 NU in this case) is less than the upper limit (0 NU in this case).

``` bash
derek @ Dereks-MacBook-Pro-2 ~/Documents/Github/repos/forks/derek/nucypher
 [186] → nucypher stake create
    Account
--  ------------------------------------------
 0  0x...
Select index of staking account [0]: 0
Selected 0: 0x...
Enter stake value in NU (15000 NU - 0 NU) [0]: 15000     <----- this line
Enter stake duration (30 - 47119) [365]: 30
```

**To Reproduce**
Try to run `nucypher stake create` without any NU in your account.

If applicable, add the full error message / Traceback or upload screenshots to help explain your problem.

 - OS Version: Mac 10.15.4
 - Nucypher Version: `master` - https://github.com/nucypher/nucypher/commit/f94bed6a220f22815ea213b01741f59cff6ff7f1
 - Ethereum Node Version:

vzotova:
Fixes #1766
Fixes #2015 
Fixes #2065

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -883,17 +883,22 @@ class Staker(NucypherTokenActor):
                         parent_status: Stake.Status = None,
                         filter_function: Callable[[Stake], bool] = None
                         ) -> Iterable[Stake]:
-        """Returns stakes for this staker which filtered by specified or child status or by provided function."""
+        """Returns stakes for this staker which filtered by status or by a provided function."""
         if not parent_status and not filter_function:
-            raise ValueError(f"Pass parent status or filter function or both.")
+            raise ValueError("Pass parent status or filter function or both.")
 
         # Read once from chain and reuse these values
         staker_info = self.staking_agent.get_staker_info(self.checksum_address)  # TODO related to #1514
         current_period = self.staking_agent.get_current_period()                 # TODO #1514 this is online only.
 
-        stakes = (stake for stake in self.stakes
-                  if (parent_status is None or stake.status(staker_info, current_period).is_child(parent_status)) and
-                  (filter_function is None or filter_function(stake)))
+        stakes = list()
+        for stake in self.stakes:
+            if parent_status and not stake.status(staker_info, current_period).is_child(parent_status):
+                continue
+            if filter_function and not filter_function(stake):
+                continue
+            stakes.append(stake)
+
         return stakes
 
     def sorted_stakes(self,
@@ -901,8 +906,11 @@ class Staker(NucypherTokenActor):
                       filter_function: Callable[[Stake], bool] = None
                       ) -> List[Stake]:
         """Returns a list of filtered stakes sorted by account wallet index."""
-        filtered_stakes = self.filtered_stakes(parent_status, filter_function) \
-            if parent_status is not None or filter_function is not None else self.stakes
+        if parent_status is not None or filter_function is not None:
+            filtered_stakes = self.filtered_stakes(parent_status, filter_function)
+        else:
+            filtered_stakes = self.stakes
+
         stakes = sorted(filtered_stakes, key=lambda s: s.address_index_ordering_key)
         return stakes
 
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -68,7 +68,7 @@ from nucypher.blockchain.eth.multisig import Authorization, Proposal
 from nucypher.blockchain.eth.registry import BaseContractRegistry, IndividualAllocationRegistry
 from nucypher.blockchain.eth.signers import KeystoreSigner, Signer, Web3Signer
 from nucypher.blockchain.eth.token import NU, Stake, StakeList, WorkTracker, validate_prolong, validate_increase, \
-    validate_divide
+    validate_divide, validate_merge
 from nucypher.blockchain.eth.utils import (
     calculate_period_duration,
     datetime_at_period,
@@ -84,6 +84,7 @@ from nucypher.crypto.powers import TransactingPower
 from nucypher.network.nicknames import nickname_from_seed
 from nucypher.types import NuNits, Period
 from nucypher.utilities.logging import Logger
+from typing import Callable
 
 
 class BaseActor:
@@ -878,19 +879,30 @@ class Staker(NucypherTokenActor):
         """The total number of staked tokens, i.e., tokens locked in the current period."""
         return self.locked_tokens(periods=0)
 
-    def stakes_filtered_by_status(self, parent_status: Stake.Status) -> Iterable[Stake]:
-        """Returns stakes for this staker which have specified or child status."""
+    def filtered_stakes(self,
+                        parent_status: Stake.Status = None,
+                        filter_function: Callable[[Stake], bool] = None
+                        ) -> Iterable[Stake]:
+        """Returns stakes for this staker which filtered by specified or child status or by provided function."""
+        if not parent_status and not filter_function:
+            raise ValueError(f"Pass parent status or filter function or both.")
 
         # Read once from chain and reuse these values
         staker_info = self.staking_agent.get_staker_info(self.checksum_address)  # TODO related to #1514
         current_period = self.staking_agent.get_current_period()                 # TODO #1514 this is online only.
 
-        stakes = (stake for stake in self.stakes if stake.status(staker_info, current_period).is_child(parent_status))
+        stakes = (stake for stake in self.stakes
+                  if (parent_status is None or stake.status(staker_info, current_period).is_child(parent_status)) and
+                  (filter_function is None or filter_function(stake)))
         return stakes
 
-    def sorted_stakes(self, parent_status: Stake.Status = None) -> List[Stake]:
+    def sorted_stakes(self,
+                      parent_status: Stake.Status = None,
+                      filter_function: Callable[[Stake], bool] = None
+                      ) -> List[Stake]:
         """Returns a list of filtered stakes sorted by account wallet index."""
-        filtered_stakes = self.stakes_filtered_by_status(parent_status) if parent_status is not None else self.stakes
+        filtered_stakes = self.filtered_stakes(parent_status, filter_function) \
+            if parent_status is not None or filter_function is not None else self.stakes
         stakes = sorted(filtered_stakes, key=lambda s: s.address_index_ordering_key)
         return stakes
 
@@ -1026,7 +1038,8 @@ class Staker(NucypherTokenActor):
     def prolong_stake(self,
                       stake: Stake,
                       additional_periods: int = None,
-                      expiration: maya.MayaDT = None) -> TxReceipt:
+                      expiration: maya.MayaDT = None
+                      ) -> TxReceipt:
         self._ensure_stake_exists(stake)
 
         if not (bool(additional_periods) ^ bool(expiration)):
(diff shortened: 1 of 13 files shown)


nucypher/blockchain/eth/actors.py
```diff
@@ -878,19 +879,30 @@ def current_stake(self) -> NU:
...
+        if not parent_status and not filter_function:
+            raise ValueError(f"Pass parent status or filter function or both.")
 
         # Read once from chain and reuse these values
         staker_info = self.staking_agent.get_staker_info(self.checksum_address)  # TODO related to #1514
         current_period = self.staking_agent.get_current_period()                 # TODO #1514 this is online only.
 
-        stakes = (stake for stake in self.stakes if stake.status(staker_info, current_period).is_child(parent_status))
+        stakes = (stake for stake in self.stakes
+                  if (parent_status is None or stake.status(staker_info, current_period).is_child(parent_status)) and
```
KPrasch, line 895: woah, wordy!

nucypher/blockchain/eth/actors.py
```diff
@@ -917,7 +930,7 @@ def initialize_stake(self,
         elif not entire_balance and not amount:
             raise ValueError("Specify an amount or entire balance, got neither")
 
-        token_balance = self.token_balance
+        token_balance = self.token_balance if not only_lock else self.calculate_staking_reward()
```
KPrasch, line 933: Is this logic local to this block - or can it be part of `self.token_balance`?

nucypher/blockchain/eth/actors.py
```diff
@@ -932,7 +945,10 @@ def initialize_stake(self,
                                            lock_periods=lock_periods)
 
         # Create stake on-chain
-        receipt = self._deposit(amount=new_stake.value.to_nunits(), lock_periods=new_stake.duration)
+        if not only_lock:
```
KPrasch, line 948: What is `only_lock` here?

nucypher/blockchain/eth/actors.py
```diff
@@ -1042,6 +1063,25 @@ def prolong_stake(self,
         self.refresh_stakes()
         return receipt
 
+    @only_me
+    def merge_stakes(self,
+                     stake_1: Stake,
+                     stake_2: Stake
```
KPrasch, line 1069: These stakes are truly equal, or can one of them be considered the "base"?

nucypher/blockchain/eth/actors.py
```diff
@@ -1064,6 +1104,17 @@ def _deposit(self, amount: int, lock_periods: int) -> TxReceipt:
                                                         call_data=Web3.toBytes(lock_periods))
         return receipt
 
+    def _lock_and_create(self, amount: int, lock_periods: int) -> TxReceipt:
+        """Public facing method for token locking without depositing."""
+        # TODO #1497 #1358
```
KPrasch, line 1109: :thinking:  Do we know if this is still an outstanding issue?

nucypher/blockchain/eth/agents.py
```diff
@@ -519,6 +519,12 @@ def prolong_stake(self, staker_address: ChecksumAddress, stake_index: int, perio
         receipt = self.blockchain.send_transaction(contract_function=contract_function, sender_address=staker_address)
         return receipt
 
+    @contract_api(TRANSACTION)
+    def merge_stakes(self, staker_address: ChecksumAddress, stake_index_1: int, stake_index_2: int) -> TxReceipt:
+        contract_function: ContractFunction = self.contract.functions.mergeStake(stake_index_1, stake_index_2)
+        receipt = self.blockchain.send_transaction(contract_function=contract_function, sender_address=staker_address)
+        return receipt
```
KPrasch, line 526: :+1: - Cool!

nucypher/cli/commands/stake.py
```diff
@@ -106,6 +107,7 @@
 option_lock_periods = click.option('--lock-periods', help="Duration of stake in periods.", type=click.INT)
 option_worker_address = click.option('--worker-address', help="Address to bond as an Ursula-Worker", type=EIP55_CHECKSUM_ADDRESS)
 option_index = click.option('--index', help="The staker-specific stake index to edit", type=click.INT)
+option_only_lock = click.option('--only-lock', help="Locking tokens without deposit", is_flag=True)
```
KPrasch, line 110: I find "only lock" a bit awkward.  How about "Deposit Only" (This is a common phrase for this purpose in the US at least)?

nucypher/cli/commands/stake.py
```diff
@@ -570,7 +577,10 @@ def increase(general_config, transacting_staker_options, config_file, force, val
     #
 
     if not value:
-        token_balance = STAKEHOLDER.token_balance
+        if not only_lock:
```
KPrasch, line 580: This conditional can be combined with the one above using `and`.

nucypher/cli/commands/stake.py
```diff
@@ -874,6 +884,73 @@ def prolong(general_config, transacting_staker_options, config_file, force, lock
...
+
+    client_account, staking_address = select_client_account_for_staking(
+        emitter=emitter,
+        stakeholder=STAKEHOLDER,
+        staking_address=transacting_staker_options.staker_options.staking_address,
+        individual_allocation=STAKEHOLDER.individual_allocation,
+        force=force)
+
+    # Handle stakes selection
+    stake_1 = None
```
KPrasch, line 911:
```suggestion
    stake1, stake2 = None, None

File: nucypher/blockchain/eth/actors.py (hunk 1 above)
Comment (vzotova) [reply nucypher/blockchain/eth/actors.py:895]: yeah, is there way to make it prettier?

File: nucypher/blockchain/eth/actors.py (hunk 2 above)
Comment (vzotova) [reply nucypher/blockchain/eth/actors.py:933]: it's local logic

File: nucypher/blockchain/eth/actors.py (hunk 4 above)
Comment (vzotova) [reply nucypher/blockchain/eth/actors.py:1069]: truly equal, order doesn't matter

File: nucypher/blockchain/eth/actors.py (hunk 5 above)
Comment (vzotova) [reply nucypher/blockchain/eth/actors.py:1109]: I think after we will see first pooling contracts and demand for CLI for them then will be better understanding what to do we those issues

File: nucypher/blockchain/eth/actors.py (hunk 1 above)
Comment (KPrasch) [reply nucypher/blockchain/eth/actors.py:895]: Somehow, by giving it a name by using a named variable at minimum, or possibly a function/method composed elsewhere?

File: nucypher/blockchain/eth/actors.py (hunk 3 above)
Comment (vzotova) [reply nucypher/blockchain/eth/actors.py:948]: I know that name is not good,
so this flag is for switching source of new lock between staker balance and unlocked balance inside escrow

File: nucypher/blockchain/eth/actors.py (hunk 1 above)
Comment (vzotova) [reply nucypher/blockchain/eth/actors.py:895]: how about extract this to for loop?

File: nucypher/cli/commands/stake.py (hunk 7 above)
Comment (vzotova) [reply nucypher/cli/commands/stake.py:110]: "only lock" is more "without deposit", not sure how to name it

File: nucypher/cli/commands/stake.py (hunk 8 above)
Comment (vzotova) [reply nucypher/cli/commands/stake.py:580]: not sure how to do this, or you mean do separate check before `if not value and not only_lock:`?

Comment (KPrasch) [reply nucypher/cli/commands/stake.py:580]: Yes - unless there is the need for both levels, perhaps I am mis-reading the diff.

Comment (vzotova) [reply nucypher/cli/commands/stake.py:580]: not sure about that, `if not value:` has 3 inner `if` conditions, which represents interactive way to choose `value`, so extracting only one of them looks not clear

File: nucypher/cli/commands/stake.py [hunk 10]
```diff
@@ -874,6 +884,73 @@ def prolong(general_config, transacting_staker_options, config_file, force, lock
...
+    stake_2 = None
+
+    if index_1 is not None and index_2 is not None:
+        stake_1 = STAKEHOLDER.stakes[index_1]
+        stake_2 = STAKEHOLDER.stakes[index_2]
+    elif index_1 is not None:  # 0 is valid.
+        stake_1 = STAKEHOLDER.stakes[index_1]
+    elif index_2 is not None:
+        stake_2 = STAKEHOLDER.stakes[index_1]
```
derekpierre, line 920: Can this block be condensed into the following?

```suggestion
    if index_1 is not None:  # 0 is valid.
        stake_1 = STAKEHOLDER.stakes[index_1]
    if index_2 is not None:
        stake_2 = STAKEHOLDER.stakes[index_2]
```

Since you check whether the relevant stake is `None` below and act accordingly.

derekpierre: If rejecting this suggestion, note that in the original code used `index_1` for `stake_2` instead of `index_2`. We may need to fortify with a test.

tests/integration/cli/actions/test_select_stake.py
```diff
@@ -281,3 +281,89 @@ def test_select_divisible_stake(test_emitter,
...
+
+    # FAILURE: no stakes with specified final period
+    with pytest.raises(click.Abort):
+        select_stake(emitter=test_emitter,
+                     staker=stakeholder,
+                     stakes_status=Stake.Status.LOCKED,
+                     filter_function=lambda stake: stake.final_locked_period == current_period)
+
+    # Divisible warning was displayed, but having
+    # no divisible stakes cases an expected failure
```
derekpierre, line 365:
```suggestion
    # no divisible stakes causes an expected failure
```

nucypher/blockchain/eth/actors.py
derekpierre: Is it like "only use locked tokens"?

nucypher/blockchain/eth/token.py
```diff
@@ -493,11 +493,26 @@ def validate_prolong(stake: Stake, additional_periods: int) -> None:
...
                                  f'of {stake.economics.minimum_locked_periods} periods.')
 
 
+def validate_merge(stake_1: Stake, stake_2: Stake) -> None:
+    if stake_1.index == stake_2.index:
+        raise Stake.StakingError(f'Stakes must be different. '
+                                 f'Selected stakes have the same index {stake_1.index}.')
+
+    if stake_1.final_locked_period != stake_2.final_locked_period:
+        raise Stake.StakingError(f'Both stakes must have the same final locked period. '
```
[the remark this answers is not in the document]
vzotova: duration is not necessary to be the same, *unlock date* is good

nucypher/cli/commands/stake.py
vzotova: this was done on purpose - first stake can be chosen from all editable stakes but second one - editable + same final period so it's why there index_1 fro stake_2

nucypher/blockchain/eth/actors.py
vzotova: close, but not exactly
depositing includes transfer token and locking them in one tx
so *only lock* for case when there is no transferring, and only locking of unlocked tokens which belong to staker in `StakingEscrow`

nucypher/cli/commands/stake.py
vzotova: ohh, I was wrong, should be `stake_1 = STAKEHOLDER.stakes[index_2]`, I'll add test for that

nucypher/blockchain/eth/actors.py
derekpierre: 👍 A for loop would be cleaner

PR #2159, opened 2020-07-28
Signer support for Felix.

KPrasch:


nucypher/blockchain/eth/interfaces.py
```diff
@@ -465,16 +465,11 @@ def __log_transaction(self, transaction_dict: dict, contract_function: ContractF
 
     @validate_checksum_address
     def build_transaction(self,
```
cygnusv, line 467: The name doesn't seem right. This is more a function for "payloads" (following the language we use in the codebase). What do you think if we change it to `build_payload` or something along those lines?

KPrasch: Yes - you are correct is not really a complete transaction.  `build_payload` is fine. :+1:

nucypher/config/characters.py
```diff
@@ -229,10 +229,13 @@ def __init__(self,
         self.db_filepath = db_filepath or os.path.join(self.config_root, self.DEFAULT_DB_NAME)
 
     def static_payload(self) -> dict:
+        if not self.signer_uri:
+            self.signer_uri = self.provider_uri
         payload = dict(
          rest_host=self.rest_host,
          rest_port=self.rest_port,
          db_filepath=self.db_filepath,
+        signer_uri=self.signer_uri
```
derekpierre, line 238: weird spacing here.

KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/interfaces.py
KPrasch: :heavy_check_mark:

PR #2162, opened 2020-08-04
Restore default return values in mock agents

cygnusv:
Mock.reset_mock() doesn't do that by default so return values remained between test modules

### DIFF
--- a/tests/mock/agents.py
+++ b/tests/mock/agents.py
@@ -160,11 +160,12 @@ class MockContractAgent:
         unexpected_transactions = self.get_unexpected_transactions(allowed=None)
         assert not bool(unexpected_transactions)
 
-    def reset(self, clear_side_effects: bool = True) -> None:
+    def reset(self, clear_side_effects: bool = True, clear_return_values: bool = True) -> None:
         for mock in self._MOCK_METHODS:
-            mock.reset_mock()
-            if clear_side_effects:
-                mock.side_effect = None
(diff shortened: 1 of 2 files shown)


tests/mock/agents.py
```diff
@@ -160,11 +160,12 @@ def assert_no_transactions(self) -> None:
         unexpected_transactions = self.get_unexpected_transactions(allowed=None)
         assert not bool(unexpected_transactions)
 
-    def reset(self, clear_side_effects: bool = True) -> None:
+    def reset(self, clear_side_effects: bool = True, clear_return_values: bool = True) -> None:
         for mock in self._MOCK_METHODS:
-            mock.reset_mock()
-            if clear_side_effects:
-                mock.side_effect = None
+            mock.reset_mock(return_value=clear_return_values, side_effect=clear_side_effects)
```
vzotova, line 165: Just to make sure: `return_value` will be same as set in `MockContractAgent`?

cygnusv: No, that line alone it's not enough. That's why the next lines are needed to restore the default return values.

cygnusv: Oh should I do that conditional to `clear_return_values`?

derekpierre: Good point - I think so

cygnusv: ✔️

PR #2166, opened 2020-08-06
Solidity -> 0.7.0

vzotova:
Test for upgrading contracts is failing because we compile two sets of contracts with the same compiler version


derekpierre:
@vzotova , since we are using v0.7.0 do we need to make any changes regarding our use of shift operators? https://solidity.readthedocs.io/en/v0.7.0/070-breaking-changes.html#how-to-update-your-code

vzotova:
> @vzotova , since we are using v0.7.0 do we need to make any changes regarding our use of shift operators? https://solidity.readthedocs.io/en/v0.7.0/070-breaking-changes.html#how-to-update-your-code

I think no, because all right operands are unsigned already and tests are green

PR #2169, opened 2020-08-08
New Logger with automatic escaping of curly braces 

### LINKED ISSUE #724: Twisted's Logger doesn't like curly braces
Once in a while, when revoking, Ursula has trouble logging the event because the string representation of the bytes representation of the revocation ends up having a "{", which the logger doesn't like.

  File "/home/jmyles/.virtualenvs/nucypher-EzeC4Bqx/lib/python3.7/site-packages/twisted/logger/_observer.py", line 131, in __call__
    observer(event)
  File "/home/jmyles/.virtualenvs/nucypher-EzeC4Bqx/lib/python3.7/site-packages/twisted/logger/_file.py", line 50, in __call__
    text = self.formatEvent(event)
  File "/home/jmyles/.virtualenvs/nucypher-EzeC4Bqx/lib/python3.7/site-packages/twisted/logger/_json.py", line 252, in <lambda>
    lambda event: u"{0}{1}\n".format(recordSeparator, eventAsJSON(event))
  File "/home/jmyles/.virtualenvs/nucypher-EzeC4Bqx/lib/python3.7/site-packages/twisted/logger/_json.py", line 203, in eventAsJSON
    flattenEvent(event)
  File "/home/jmyles/.virtualenvs/nucypher-EzeC4Bqx/lib/python3.7/site-packages/twisted/logger/_flatten.py", line 87, in flattenEvent
    aFormatter.parse(event["log_format"])
builtins.ValueError: Single '}' encountered in format string
bytes(revocation)
b'REVOKE-\x8fX_Z\xa1\x7f_\x1e\xd9BO]\xcdx\x0c\xfa\x87\x83\x1c\xd8\xd7\xaf\xe9S[\xd3\xd1\x01D[\x0f\x02(\xf51N\x05\xfc]\x95\xf7\xea\xd7x\xb4\xee\xb4<\xad\x9f\xbf&\xf9\x87\x13\x15\xbe\xfa}\x0b\xdb\xab\x8b=\n\x7fG\xfd\x8c\xd6\xdd\xb8\xff\xad\r\x0bj\x0b\xb9\xdf\x80lG.\x87g\xdc\x04\xb9~\x99s\xa9\x8ef\xa4'

cygnusv:
Drop-in replacement of Twisted's Logger, patching the emit() method to tolerate inputs with curly braces, i.e., not compliant with PEP 3101's format strings.

See Issue #724 and, particularly, https://github.com/nucypher/nucypher/issues/724#issuecomment-600190455

Fixes #724

### DIFF
--- a/nucypher/config/constants.py
+++ b/nucypher/config/constants.py
@@ -46,6 +46,8 @@ CONTRACT_ROOT = SOL_PACKAGE / 'source' / 'contracts'
 APP_DIR = AppDirs(nucypher.__title__, nucypher.__author__)
 DEFAULT_CONFIG_ROOT = os.getenv('NUCYPHER_CONFIG_ROOT', default=APP_DIR.user_data_dir)
 USER_LOG_DIR = os.getenv('NUCYPHER_USER_LOG_DIR', default=APP_DIR.user_log_dir)
+DEFAULT_LOG_FILENAME = "nucypher.log"
+DEFAULT_JSON_LOG_FILENAME = "nucypher.json"
 
 
 # Static Seednodes
--- a/nucypher/config/constants.py
+++ b/nucypher/config/constants.py
@@ -23,7 +23,6 @@ from appdirs import AppDirs
 from pathlib import Path
 
 import nucypher
-from nucypher.blockchain.eth import sol
 
 # Environment variables
 NUCYPHER_ENVVAR_KEYRING_PASSWORD = "NUCYPHER_KEYRING_PASSWORD"
@@ -38,8 +37,6 @@ NUCYPHER_ENVVAR_WORKER_IP_ADDRESS = 'NUCYPHER_WORKER_IP_ADDRESS'
 NUCYPHER_PACKAGE = Path(nucypher.__file__).parent.resolve()
 BASE_DIR = NUCYPHER_PACKAGE.parent.resolve()
 DEPLOY_DIR = BASE_DIR / 'deploy'
-SOL_PACKAGE = Path(sol.__file__).parent.resolve()
-CONTRACT_ROOT = SOL_PACKAGE / 'source' / 'contracts'
 
 
 # User Application Filepaths
--- a/nucypher/utilities/logging.py
+++ b/nucypher/utilities/logging.py
@@ -21,17 +21,22 @@ from contextlib import contextmanager
 import pathlib
 from twisted.logger import (
     FileLogObserver,
-    LogLevel,
     formatEvent,
     formatEventAsClassicLogText,
     globalLogPublisher,
-    jsonFileLogObserver
+    jsonFileLogObserver,
+    LogLevel,
 )
 from twisted.logger import Logger as TwistedLogger
 from twisted.python.logfile import LogFile
 
 import nucypher
-from nucypher.config.constants import NUCYPHER_SENTRY_ENDPOINT, USER_LOG_DIR
+from nucypher.config.constants import (
+    DEFAULT_JSON_LOG_FILENAME,
+    DEFAULT_LOG_FILENAME,
+    NUCYPHER_SENTRY_ENDPOINT,
+    USER_LOG_DIR,
+)
 
 ONE_MEGABYTE = 1_048_576
 MAXIMUM_LOG_SIZE = ONE_MEGABYTE * 10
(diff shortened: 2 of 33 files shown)


nucypher/config/constants.py
```diff
@@ -23,7 +23,7 @@
 from pathlib import Path
 
 import nucypher
-from nucypher.blockchain.eth import sol
+# from nucypher.blockchain.eth import sol
```
fjarri, line 26: Why not just remove it?

nucypher/config/constants.py
```diff
@@ -38,14 +38,15 @@
 NUCYPHER_PACKAGE = Path(nucypher.__file__).parent.resolve()
 BASE_DIR = NUCYPHER_PACKAGE.parent.resolve()
 DEPLOY_DIR = BASE_DIR / 'deploy'
-SOL_PACKAGE = Path(sol.__file__).parent.resolve()
-CONTRACT_ROOT = SOL_PACKAGE / 'source' / 'contracts'
+# SOL_PACKAGE = Path(sol.__file__).parent.resolve()
+# CONTRACT_ROOT = SOL_PACKAGE / 'source' / 'contracts'
```
fjarri, line 42: Same question here, why not delete these?

nucypher/config/constants.py
cygnusv: You're right. Fixed.

nucypher/utilities/logging.py
```diff
@@ -163,15 +176,43 @@ def _ensure_dir_exists(path):
...
+class Logger(TwistedLogger):
+    """Drop-in replacement of Twisted's Logger, patching the emit() method to tolerate inputs with curly braces,
+    i.e., not compliant with PEP 3101.
+
+    See Issue #724 and, particularly, https://github.com/nucypher/nucypher/issues/724#issuecomment-600190455"""
+
+    CURLY_BRACES_REGEX = re.compile('{+|}+')
+
+    @classmethod
+    def escape_format_string(cls, string):
```
fjarri, line 202: Does it really have to be a classmethod? Since it doesn't actually create an instance of `cls`, perhaps a standalone function would do.

nucypher/utilities/logging.py
```diff
@@ -163,15 +176,43 @@ def _ensure_dir_exists(path):
...
+        def escape_group_of_curly_braces(match):
+            curlies = match.group()
+            if len(curlies) % 2 == 1:
+                curlies += curlies
+            return curlies
+
+        escaped_string = cls.CURLY_BRACES_REGEX.sub(escape_group_of_curly_braces, string)
+        return escaped_string
+
+    def emit(self, level, format=None, **kwargs):
```
fjarri, line 216: So does this essentially mean that we won't ever use the formatting functionality of `twisted`'s `Logger`? All the curly braces will be escaped?

nucypher/utilities/logging.py
```diff
@@ -163,15 +176,43 @@ def _ensure_dir_exists(path):
...
+        Escapes curly braces from a PEP-3101's format string when there's a sequence of odd length
+        """
+
+        def escape_group_of_curly_braces(match):
+            curlies = match.group()
+            if len(curlies) % 2 == 1:
+                curlies += curlies
+            return curlies
+
+        escaped_string = cls.CURLY_BRACES_REGEX.sub(escape_group_of_curly_braces, string)
```
fjarri, line 213: Would `.replace('{', '{{').replace('}', '}}')` work?

nucypher/utilities/logging.py
cygnusv: Yes, pretty much. If someone is interested in that they can always use the original Logger. The problem is that sometimes we directly log messages coming from 3rd party libraries, like exceptions or responses, which may not be adequate for this use, and you never know in advance.

nucypher/utilities/logging.py
cygnusv: We already do that as a workaround, and that works too, but it's a bit too much since if the string was already acceptable if will double the number of braces unnecessarily. The approach here only modifies those that are problematic.

I have to investigate a bit more, since there's an old commit somewhere which says that escaping is not enough, and the workaround was to remove them altogether. I'll look for it later.

nucypher/utilities/logging.py
```diff
@@ -163,15 +176,43 @@ def _ensure_dir_exists(path):
...
+    def escape_format_string(cls, string):
+        """
+        Escapes curly braces from a PEP-3101's format string when there's a sequence of odd length
+        """
+
+        def escape_group_of_curly_braces(match):
+            curlies = match.group()
+            if len(curlies) % 2 == 1:
+                curlies += curlies
+            return curlies
```
vzotova, line 211: if we try to avoid unnecessary escaping - maybe add only one symbol and not double all of them? just suggestion

PR #2170, opened 2020-08-09
Few more fixes for `staker`

### LINKED ISSUE #1436: Rethink `nucypher stake list --all`
`biznaga` introduces a new option `--all` to the `nucypher stake list` command that shows also "inactive" (sub)stakes. 

@michwill suggests a new name for this option like `--show-inactive` or `--with-inactive`

But I also remember that @vzotova told me something regarding how these are not actually "inactive", but I don't remember the details. 

Let's rethink how this command works, and then how to name it.

### LINKED ISSUE #1545: Potential problems with registry mutability, computation of registry ID and ContractAgency's agents cache
Initial discussion:
https://github.com/nucypher/nucypher/pull/1407/files#r347151652
```
michwill on 17 Nov Member
Oh wow... Was it supposed to be in all the time?

vzotova on 18 Nov Author Member
I think in production it's impossible to have two different registries classes with the same content, but in test it was all the time.
Also interesting moment - locally I still have old behavior with ordering which "good" for tests, but one commit changed it for CircleCI

cygnusv 11 days ago Member
Instead of the class name we should use the registry type:

Suggested change 
        blake.update(self.__class__.__name__.encode())
        blake.update(self.REGISTRY_TYPE.encode())
We don't want the ID to be different depending if it's a LocalContractRegistry, a TemporaryContractRegistry or an InMemoryContractRegistry, even though all of them are contract registries.

vzotova 10 days ago Author Member
while tests we create two same registries Local and Memory and they have same ID, it leads to bugs depending on ordering in cache dictionary. I see this ID not for content but for instance

cygnusv 6 days ago Member
Well, but if this is a problem for tests, then we should monkey-patch it in tests then. I don't see impossible that in production we could have two registries with the same content but different classes (e.g., an InMemory registry persisted to a LocalRegistry)
@KPrasch what do you think?
```

Last discussion on Discord: https://discordapp.com/channels/411401661714792449/411401661714792451/658609071976349726
```
vicky.zToday at 11:01 AM
I'm still thinking that comparing only content is design issue, because registry cache has id as key and instance as value, and will be two same keys for different instances,
I see two solutions - change cache structure somehow or use id with class
 
dnunezToday at 11:02 AM
ok, can you open an issue?
 
vicky.zToday at 11:03 AM
hmmm, I'm not sure what it should be, I mean id with class is simplest solution and it's already there
 
dnunezToday at 11:05 AM
well, I still don't see why an in-memory registry should be mapped differently than a file-based registry if they have the same data, which is what happens with this solution
 
vicky.zToday at 11:06 AM
I can provide example - if we have two same registries but in different instances, it's possible to change only one instance and they will be different
 
dnunezToday at 11:07 AM
ok....? what's wrong with that?
 
vicky.zToday at 11:09 AM
will be totally uncontrolled situation when we have two different registries, but agents that were created will think that only one registry, my point that if we want id -> content, then we should store content not instance
 
dnunezToday at 11:10 AM
ah, the issue is more related to the registry being mutable?
 
vicky.zToday at 11:12 AM
kind of, but not only this, we have cache as id -> agent_class -> agent, so issue when you try to get agent using instance1 but will return agent with instance2 because ids were equal
ContractAgency.get_agent(instance1) will return Agent(instance2), I see that as an issue, so we just want ContractAgency.get_agent(instance1) will return Agent(instance1)
 
dnunezToday at 11:15 AM
well, if content is the same, what's the problem? Agents are stateless anyway
 
vicky.zToday at 11:15 AM
problem from code perspective, it's bad practice when you get not what you want
 
vicky.zToday at 11:17 AM
It was really difficult to find out where is problem, because locally I can't reproduce it
so what if it will affect somehow production, I'm for transparent layers and logic
other possible solution: change cache to something like id -> registry_class -> agent_class -> agent
```

### LINKED ISSUE #1739: 'stake status stakers': show locked NU in the next period
Right now 'staked' value is shown as locked tokens in the current period, it could confuse because when sub-stake created - it starts form the next period. 
Show two values (lock(current) + lock(next)) or max of  them

### LINKED ISSUE #2066: Link to staking guide after running `nucypher stake create` is incorrect
**Describe the Bug**
After running `nucypher stake create` successfully, the blurb at the end provides the wrong link to our staking guide.

```bash
$ nucypher stake create

...

View your stakes by running 'nucypher stake list'
or set your Ursula worker node address by running 'nucypher stake set-worker'.

See https://docs.nucypher.com/en/latest/guides/staking_guide.html
```

The staking guide link should be "https://docs.nucypher.com/en/latest/guides/network_node/staking_guide.html".

**To Reproduce**
Successfully run `nucypher stake create`

See above

 - OS Version: Mac 10.15.4
 - Nucypher Version: `master` - https://github.com/nucypher/nucypher/commit/d41cc8fe80af56dca38ac6703fa8b5182731674b
 - Ethereum Node Version:

### LINKED ISSUE #2114: Calling `staker.worker_address` twice when no worker is bonded returns two different values
**Describe the Bug**
Calling the property `staker.worker_address` twice on the same staker object when there is no bonded worker returns `NO_WORKER_BONDED` first, and then returns `NULL_ADDRESS` for subsequent calls.

**To Reproduce**
I was attempting to highlight `Worker <worker_address>` as red in the output of `nucypher stake list` when no worker is bonded to the staker. To do so I decided to modify the line:

`emitter.echo(f"Worker {staker.worker_address} ════")`

to instead be

`emitter.echo(f"Worker {staker.worker_address} ════", color='red' if (staker.worker_address == NO_WORKER_BONDED) else None)`

which did not work. The reason being that the first call to `staker.worker_address` returns NO_WORKER_BONDED, but a second (or subsequent) call(s) returns `NULL_ADDRESS`.

Looking at the `worker_address` function:
```python
    @property
    def worker_address(self) -> str:
        if self.__worker_address:
            # TODO: This is broken for StakeHolder with different stakers - See #1358
            return self.__worker_address
        else:
            worker_address = self.staking_agent.get_worker_from_staker(staker_address=self.checksum_address)
            self.__worker_address = worker_address

        if self.__worker_address == NULL_ADDRESS:
            return NO_WORKER_BONDED.bool_value(False)
        return self.__worker_address
```

On the second call to this function, `self.__worker_address` is already set and simply returns its value which is `NULL_ADDRESS` instead of getting modified to be `NO_WORKER_BONDED` again.

It should probably be modified to be

```python
    @property
    def worker_address(self) -> str:
        if not self.__worker_address:
            worker_address = self.staking_agent.get_worker_from_staker(staker_address=self.checksum_address)
            self.__worker_address = worker_address

        if self.__worker_address == NULL_ADDRESS:
            return NO_WORKER_BONDED.bool_value(False)
        return self.__worker_address
```

I didn't want to make the change myself because I am unsure of the repercussions for such a change.

N/A

 - OS Version: Linux Ubuntu 18.04.4
 - Nucypher Version: master @ https://github.com/nucypher/nucypher/commit/202e502ad4239227b613c1346cd161723d892765
 - Ethereum Node Version:  Geth 1.9.15-stable-0f77f34b

vzotova:
Fixes #2114
Fixes #1739
Fixes #1436
Fixes #2066
Fixes #1545

### DIFF
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -100,13 +100,13 @@ class EthereumContractAgent:
                  transaction_gas: Optional[Wei] = None):
 
         self.log = Logger(self.__class__.__name__)
-        self.registry = registry
+        self.registry_repr = str(registry)
 
         self.blockchain = BlockchainInterfaceFactory.get_or_create_interface(provider_uri=provider_uri)
 
         if not contract:  # Fetch the contract
             contract = self.blockchain.get_contract_by_name(
-                registry=self.registry,
+                registry=registry,
                 contract_name=self.contract_name,
                 proxy_name=self._proxy_name,
                 use_proxy_address=self._forward_address
@@ -122,12 +122,12 @@ class EthereumContractAgent:
         self.log.info("Initialized new {} for {} with {} and {}".format(self.__class__.__name__,
                                                                         self.contract.address,
                                                                         self.blockchain.provider_uri,
-                                                                        self.registry))
+                                                                        self.registry_repr))
 
     def __repr__(self) -> str:
         class_name = self.__class__.__name__
         r = "{}(registry={}, contract={})"
-        return r.format(class_name, self.registry, self.contract_name)
+        return r.format(class_name, self.registry_repr, self.contract_name)
 
     def __eq__(self, other: Any) -> bool:
         return bool(self.contract.address == other.contract.address)
@@ -945,6 +945,9 @@ class PreallocationEscrowAgent(EthereumContractAgent):
         self.__read_principal()
         self.__read_interface(registry)
 
+        self.token_agent: NucypherTokenAgent = ContractAgency.get_agent(NucypherTokenAgent, registry)
+        self.staking_agent: StakingEscrowAgent = ContractAgency.get_agent(StakingEscrowAgent, registry)
+
         super().__init__(contract=self.principal_contract, registry=registry, *args, **kwargs)
 
     def __read_interface(self, registry: BaseContractRegistry) -> None:
@@ -1003,18 +1006,15 @@ class PreallocationEscrowAgent(EthereumContractAgent):
     @property  # type: ignore
     @contract_api(CONTRACT_ATTRIBUTE)
     def available_balance(self) -> int:
-        token_agent: NucypherTokenAgent = ContractAgency.get_agent(NucypherTokenAgent, self.registry)
-        staking_agent: StakingEscrowAgent = ContractAgency.get_agent(StakingEscrowAgent, self.registry)
-
-        overall_balance = token_agent.get_balance(self.principal_contract.address)
-        seconds_per_period = staking_agent.contract.functions.secondsPerPeriod().call()
-        current_period = staking_agent.get_current_period()
+        overall_balance = self.token_agent.get_balance(self.principal_contract.address)
+        seconds_per_period = self.staking_agent.contract.functions.secondsPerPeriod().call()
+        current_period = self.staking_agent.get_current_period()
         end_lock_period = epoch_to_period(self.end_timestamp, seconds_per_period=seconds_per_period)
 
         available_balance = overall_balance
         if current_period <= end_lock_period:
-            staked_tokens = staking_agent.get_locked_tokens(staker_address=self.principal_contract.address,
-                                                            periods=end_lock_period - current_period)
+            staked_tokens = self.staking_agent.get_locked_tokens(staker_address=self.principal_contract.address,
+                                                                 periods=end_lock_period - current_period)
             if self.unvested_tokens > staked_tokens:
                 # The staked amount is deducted from the locked amount
                 available_balance -= self.unvested_tokens - staked_tokens
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -100,7 +100,7 @@ class EthereumContractAgent:
                  transaction_gas: Optional[Wei] = None):
 
         self.log = Logger(self.__class__.__name__)
-        self.registry_repr = str(registry)
+        self.registry_str = str(registry)
 
         self.blockchain = BlockchainInterfaceFactory.get_or_create_interface(provider_uri=provider_uri)
 
@@ -122,12 +122,12 @@ class EthereumContractAgent:
         self.log.info("Initialized new {} for {} with {} and {}".format(self.__class__.__name__,
                                                                         self.contract.address,
                                                                         self.blockchain.provider_uri,
-                                                                        self.registry_repr))
+                                                                        self.registry_str))
 
     def __repr__(self) -> str:
         class_name = self.__class__.__name__
         r = "{}(registry={}, contract={})"
-        return r.format(class_name, self.registry_repr, self.contract_name)
+        return r.format(class_name, self.registry_str, self.contract_name)
 
     def __eq__(self, other: Any) -> bool:
         return bool(self.contract.address == other.contract.address)
--- a/nucypher/cli/commands/stake.py
+++ b/nucypher/cli/commands/stake.py
@@ -320,13 +320,13 @@ def config(general_config, config_file, config_options):
 @stake.command('list')
 @group_staker_options
 @option_config_file
-@click.option('--all', help="List all stakes, including unlocked", is_flag=True)
+@click.option('--show-all', help="List all stakes, including unlocked and inactive", is_flag=True)
 @group_general_config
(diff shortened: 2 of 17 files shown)


nucypher/blockchain/eth/agents.py
```diff
@@ -100,13 +100,13 @@ def __init__(self,
                  transaction_gas: Optional[Wei] = None):
 
         self.log = Logger(self.__class__.__name__)
-        self.registry = registry
+        self.registry_repr = str(registry)
```
KPrasch, line 103: Heh - slightly confusion name here - since `str` calls `__str__` and not `_repr__`.

nucypher/cli/commands/stake.py
```diff
@@ -320,13 +320,13 @@ def config(general_config, config_file, config_options):
 @stake.command('list')
 @group_staker_options
 @option_config_file
-@click.option('--all', help="List all stakes, including unlocked", is_flag=True)
+@click.option('--show-all', help="List all stakes, including unlocked and inactive", is_flag=True)
```
KPrasch, line 323: I find "show" to be a bit pedantic here and prefer simply `--all`.  There is an additional kwarg for click options to map to an alias to avoid name-shadowing the all function here.

tests/acceptance/cli/ursula/test_stakeholder_and_ursula.py
```diff
@@ -696,7 +697,7 @@ def test_stake_unbond_worker(click_runner,
                     checksum_address=manual_staker,
                     registry=agency_local_registry)
 
-    assert not staker.worker_address
+    assert staker.worker_address == NULL_ADDRESS
```
KPrasch, line 700: :+1:

tests/unit/test_registry_basics.py
```diff
@@ -75,7 +81,13 @@ def test_contract_registry(tempfile_path):
...
     current_dataset.append([test_name, test_addr, test_abi])
     test_registry.write(current_dataset)
+    assert test_registry.id != registry_id
 
     # Check that searching for an unknown contract raises
     with pytest.raises(BaseContractRegistry.InvalidRegistry):
         test_registry.search(contract_address=test_addr)
+
+    # Check id of new registry with the same content
+    new_registry = InMemoryContractRegistry()  # TODO finish this
```
KPrasch, line 91: This TODO Needs an Issue to track?

nucypher/blockchain/eth/agents.py
vzotova: you're right
changed 
✔️

tests/unit/test_registry_basics.py
vzotova: forgot to remove this
✔️

nucypher/cli/commands/stake.py
vzotova: ✔️

tests/integration/cli/test_stake_cli_functionality.py
```diff
@@ -1154,15 +1211,108 @@ def test_merge_non_interactive(click_runner,
...
+            remaining = sub_stake.last_period - current_period + 1
+            start_datetime = datetime_at_period(period=sub_stake.first_period,
+                                                seconds_per_period=token_economics.seconds_per_period,
+                                                start_of_period=True)
+            unlock_datetime = datetime_at_period(period=sub_stake.last_period + 1,
+                                                 seconds_per_period=token_economics.seconds_per_period,
+                                                 start_of_period=True)
+            enactment = start_datetime.local_datetime().strftime("%b %d %Y")
+            termination = unlock_datetime.local_datetime().strftime("%b %d %Y")
+            search = f"{index}\\s+│\\s+" \
```
KPrasch, line 1263: :+1: Sufficiently generic to remain flexible

nucypher/cli/painting/staking.py
```diff
@@ -82,7 +82,8 @@ def paint_stakes(emitter: StdoutEmitter,
         snippet_with_line = network_snippet + '═'*(line_width-len(network_snippet)+1)
         emitter.echo(snippet_with_line, bold=True)
     emitter.echo(f"Staker {staker.checksum_address} ════", bold=True, color='red' if missing else 'green')
-    emitter.echo(f"Worker {staker.worker_address} ════")
+    worker = staker.worker_address if staker.worker_address != NULL_ADDRESS else 'NO_WORKER_BONDED'
```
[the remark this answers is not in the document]
vzotova: I left `NO_WORKER_BONDED` only to not change all documentation, but if you think it worth it - then I'll do that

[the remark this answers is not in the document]
cygnusv: Don't worry to change the documentation, the change is too small to justify perfect consistency. We cam also leave as it is and merge. As you prefer.

PR #2173, opened 2020-08-10
"TMI-rebased" - ping-pong

jMyles:

PR #2175, opened 2020-08-11
[WIP] TMI-rebased test fixes - Intermediate CI run for tests

derekpierre:
Related to #2173 
- [x] Ensure created ursula is stopped; fixes tests/acceptance/cli/ursula/test_local_keystore_integration.py failure.
- [x] Ensure dry-run Ursula is stopped; fixes tests/acceptance/cli/ursula/test_stakeholder_and_ursula.py failure.

### DIFF
--- a/nucypher/cli/commands/ursula.py
+++ b/nucypher/cli/commands/ursula.py
@@ -392,10 +392,15 @@ def run(general_config, character_options, config_file, interactive, dry_run, me
                                                     metrics_prefix=metrics_prefix,
                                                     listen_address=metrics_listen_address)
 
-    return URSULA.run(emitter=emitter,
-                      start_reactor=not dry_run,
-                      interactive=interactive,
-                      prometheus_config=prometheus_config)
+    # TODO should we just not call run at all for "dry_run"
+    try:
+        URSULA.run(emitter=emitter,
+                   start_reactor=not dry_run,
+                   interactive=interactive,
+                   prometheus_config=prometheus_config)
+    finally:
+        if dry_run:
+            URSULA.stop()
 
 
 @ursula.command(name='save-metadata')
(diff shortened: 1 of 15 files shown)


tests/acceptance/cli/ursula/test_local_keystore_integration.py
```diff
@@ -147,15 +147,18 @@ def test_ursula_and_local_keystore_signer_integration(click_runner,
...
+
+        # Show that we can produce the exact same signer as pre-config...
+        assert pre_config_signer.path == ursula.signer.path
+
+        # ...and that transactions are signed by the keytore signer
+        receipt = ursula.commit_to_next_period()
+        transaction_data = testerchain.client.w3.eth.getTransaction(receipt['transactionHash'])
+        assert transaction_data['from'] == worker_account.address
+    finally:
+        ursula.stop()
```
KPrasch, line 164: :+1: ( although still somewhat confused about disenchant vs. stop @jMyles )

nucypher/cli/commands/ursula.py
```diff
@@ -392,10 +392,15 @@ def run(general_config, character_options, config_file, interactive, dry_run, me
...
-                      interactive=interactive,
-                      prometheus_config=prometheus_config)
+    # TODO should we just not call run at all for "dry_run"
+    try:
+        URSULA.run(emitter=emitter,
+                   start_reactor=not dry_run,
+                   interactive=interactive,
+                   prometheus_config=prometheus_config)
+    finally:
+        if dry_run:
```
KPrasch, line 402: Why only in the case of a dry run?

KPrasch:
Included in #2173

nucypher/cli/commands/ursula.py
derekpierre: If Ursula is starting from the command line, then we don't want to stop it other than for dry-run.

KPrasch: Perhaps I am missing something - but why not?  At the end of this test it seems like we can stop the ursula no mater what the situation is.

derekpierre: If I'm understanding you correctly - you are correct, but the code change is not to the test, but rather the ursula command module itself.

PR #2182, opened 2020-08-18
Help pytest handle the same test file name in this folder and in acceptance/blockchain/interfaces

fjarri:
Identical file names (in this case, `test_token_and_stake.py`) in directories without `__init__.py` lead to test discovery errors.

*Edit:* instead of adding `__init__.py`, renamed the offending files. I feel like it's the temporary solution though - with tests separated by type (acceptance/integration/etc) it is quite possible that this problem will resurface for some other part of the API.


KPrasch:
Adding `__init__.py` indicates to python that the directory is a "regular" importable python package instead of a namespaced one.  (<https://docs.python.org/3/reference/import.html#regular-packages>) 

By default packaging tools will recognize these as part of distribution.   This can create a somewhat unexpected situation when someone installs a package and their local `tests` namespace is overridden by the dependency.  Currently we mitigate this by explicitly excluding `tests` in `setup.py` and implement a `Manifest`.  Since tests are not typically *regular python packages* I'm somewhat against reintroducing them ( I removed them all in a PR a year ago or so).


> How does everyone else run tests without encountering this error?

My guess based on the information I have is that this is typically avoided by running a subset of tests that do not include dupicate names, like the unit, integration, or acceptance suites.

Because of the way pytest collects tests and test packages, all test names must be unique in a given session.  If duplicate test names are causing an issue the fix is simple: rename the test.  This usually results in a better namespace anyways and is perhaps even a code smell.

> How does CI run tests without encountering this error?

CircleCI runs subsets of tests from a suite based on their prior timing data.

fjarri:
> By default packaging tools will recognize these as part of distribution.

So, could we just not include them in the distribution? 

> My guess based on the information I have is that this is typically avoided by running a subset of tests that do not include dupicate names, like the unit, integration, or acceptance suites.

For me `pytest` fails at the test discovery stage, regardless of what subset of tests I run.

cygnusv (approved):

PR #2193, opened 2020-08-25
Further reinforce hypothetical nature of WorkLock example scenarios

derekpierre:
The values used in the examples can be modified at a later date, since it takes time to determine good values for the scenarios. As a stop-gap, I will overly highlight the `hypothetical` nature of the examples even more, until I can work through some examples with actual values.

Updated page:

Screen Shot 2020-08-25 at 2 39 10 PM

### DIFF
--- a/docs/source/architecture/worklock.rst
+++ b/docs/source/architecture/worklock.rst
@@ -37,20 +37,25 @@ Participants that fail to successfully run a NuCypher network node for six month
 WorkLock contract.
 
 
-Hypothetical Escrow Scenarios
-------------------------------
+Hypothetical WorkLock Scenarios
+-------------------------------
 
-.. note::
+.. important::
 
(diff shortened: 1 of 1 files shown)


cygnusv (approved):
Something unrelated to this PR: in Scenario 3 section, the formatting of `whale_1` and `whale_2` doesn't show correctly because it's already in a paragraph surrounded by `** ... **`. Maybe there's a way to correct that?

derekpierre:
@cygnusv `rst` is weird about nesting styles. I tried before but to no avail.

mswilkison:
Unfortunately, I don't know if the disclaimer about the hypothetical natures is getting through. WDYT about changing to the actual numbers?

derekpierre:
👍 I'll try to put together some new examples and prioritize it for tomorrow. This was only meant as a stop gap.

PR #2213, opened 2020-08-31
Relock Dependencies

KPrasch:


fjarri:
Is all the deleted stuff in `dev-requirements` duplicates of `requirements`?

KPrasch:
> Is all the deleted stuff in dev-requirements duplicates of requirements?

Hmm - @derekpierre - did you do the last lock by chance?  Might be an OS difference.  This output is generated by pipenv.

derekpierre:
@KPrasch I think the last relock deps was done by you (?) - https://github.com/nucypher/nucypher/commit/f74838c1c6636702ed0778640c910eb14102dfe9

derekpierre:
It is interesting that so many lines were deleted - perhaps some sub-depenencies in various packages changed.

KPrasch:
@fjarri - Im not sure exactly what happened last time to cause the red lines - but it looks fixed now.

nucypher/datastore/datastore.py
```diff
@@ -84,7 +84,7 @@ class Datastore:
 
     # LMDB has a `map_size` arg that caps the total size of the database.
     # We can set this arbitrarily high (1TB) to prevent any run-time crashes.
-    LMDB_MAP_SIZE = 1_000_000_000_000
+    LMDB_MAP_SIZE = 1_000_000
```
fjarri, line 87: Is it the size in bytes? So it's only 1Mb now? Why such a massive change? 

Also if it's intentional, the comment should be changed too.

derekpierre, line 87: Unsure about using the smaller - I think the original large value is needed...see https://discordapp.com/channels/411401661714792449/411401661714792451/750728568509562931. @tuxxy ?

derekpierre (approved):
🎸

nucypher/datastore/datastore.py
KPrasch: I hear ya - to be fair this is in a commit all by itself with the message "Use small LMDB_MAP_SIZE to get CI to run".

If I do not change this value, CI will fail.  What is the best way to handle this considering the very well defined scope of this PR?

fjarri: Considering the scope of this PR, this commit perhaps doesn't belong here at all :) Can we set the value based on an environment variable?

fjarri:
Seems like `dev-requirements` has everything from `requirements` removed again...

PR #2220, opened 2020-09-02
Trezorlib Transaction Signing

KPrasch:
Integrates `trezorlib` into NuCypher directly as an implementation of a `Signer` subclass and implements transaction signing directly.

Extends and integrates original authorship by @tuxxy from last year.


derekpierre:
@KPrasch I thought the failure was due to the updated version of py-solc-x (worked locally) but that didn't seem to do the trick.

I'll take more of a look later, but the `character` tests pass now 🎉 .

derekpierre:
The problematic test that times out seems to be `tests/acceptance/blockchain/interfaces/test_solidity_compiler.py::test_multi_source_compilation`.

derekpierre:
There is something fishy with py-solc-x v1.0.0. I used circleci's docker image (https://hub.docker.com/r/circleci/python v3.7) to run the `tests/acceptance/blockchain/interfaces/test_solidity_compiler.py::test_multi_source_compilation` test with v1.0.0 and it hangs. When I change the version to 0.10.1 (`pip install py-solc-x==0.10.1`) the test runs within ~20 seconds.

I suspect changing the Pipfile to indicate py-solc-x = "==0.10.1" and rebuilding the full requirement files may solve the problem on CI. In my commit (https://github.com/nucypher/nucypher/pull/2220/commits/4f964b82f9db36856ef3517d1546c1efeabf4327), I only changed the py-solc-x version in `dev-requirements.txt` but it would be better to relock the dependencies. I tried relocking myself but some dependencies eg. `rlp` went missing in dev-requirements.txt - unsure why that is.
 
@vzotova is there anything that jumps out at you that would cause this given the changelog notes for v1.0.0 - https://github.com/iamdefinitelyahuman/py-solc-x/blob/master/CHANGELOG.md.

derekpierre:
Build works 🎉

KPrasch:
Well done @derekpierre

vepkenez:
Looks like I forgot to add the new hendrix version to the Pipfile.
Fixing that.

KPrasch:
I may have a few more changes incoming but I am ready to start the review process on this PR.

nucypher/blockchain/eth/signers/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Affero General Public License for more details.
+
+ You should have received a copy of the GNU Affero General Public License
+ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+"""
+
+from nucypher.blockchain.eth.signers.base import Signer
+from nucypher.blockchain.eth.signers.software import ClefSigner, KeystoreSigner
+from nucypher.blockchain.eth.signers.hardware import TrezorSigner
```
fjarri, line 20: Does this mean that the user will be unable to import any of the signer classes if `trezorlib` is not installed (which, as far as I understand, is planned to be an extra, but I don't see it anywhere in the `setup.py` or `requirements` at the moment)?

KPrasch: At the moment I'm leaning towards leaving it included in the standard installation - Nothing wrong with supporting the trezor "out of the box" IMO.

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,250 @@
...
+            hd_path = parse_path(f"{self.DERIVATION_ROOT}/{index}")
+        else:
+            try:
+                hd_path = self.__addresses[checksum_address]
+            except KeyError:
+                raise RuntimeError(f"{checksum_address} was not loaded into the device address cache.")
+        return hd_path
+
+    @handle_trezor_call
+    def __get_address(self, index: int = None, hd_path: Address = None, show_display: bool = True) -> str:
```
fjarri, line 120: `index` and `hd_path` should be `Optional`

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,250 @@
...
+            raise self.NoDeviceDetected("Could not find a TREZOR device to connect to. Have you unlocked it?")
+        self._device_id = self.__client.get_device_id()
+        self.__addresses = dict()  # track derived addresses
+        self.__load_addresses()
+
+    #
+    # Internal
+    #
+
+    def __get_address_path(self, index: int = None, checksum_address: str = None) -> List[H_]:
```
fjarri, line 106: `index` and `checksum_address` should be `Optional`

nucypher/blockchain/eth/signers/__init__.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+SIGNERS = {
+    ClefSigner.URI_SCHEME: ClefSigner,
+    KeystoreSigner.URI_SCHEME: KeystoreSigner,
+    TrezorSigner.URI_SCHEME: TrezorSigner
+}
+
+Signer.SIGNERS = SIGNERS
```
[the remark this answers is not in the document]
KPrasch: Good eye - The only reason I ended up with this here is to avoid importing Issues.  Where can can it be defined with all of the subclass names still in scope?

nucypher/blockchain/eth/signers/base.py
```diff
@@ -0,0 +1,94 @@
...
+from urllib.parse import urlparse
+
+from hexbytes.main import HexBytes
+
+from nucypher.utilities.logging import Logger
+
+
+class Signer(ABC):
+
+    URI_SCHEME = NotImplemented
```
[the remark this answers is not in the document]
KPrasch: I rather like the overriding pattern of - Is there an advantage of moving it to a method?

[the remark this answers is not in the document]
fjarri: The same advantage you get from all the other methods marked as abstract - its existence will be enforced. It just seems strange to me to have direct redefinition mixed with `ABC` usage in the same class.

Of course, the problem here is that it's a static method, and `abc` can only enforce definition of abstract methods on class instantiation (and it doesn't check that the defined method is actually static). Still, it would provide some protection from mistakes.

nucypher/blockchain/eth/signers/__init__.py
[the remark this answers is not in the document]
fjarri: You're right, circular import will be a problem. I suppose the only alternative would be to import the derived classes dynamically, and have a method of `Signer` like 
```
    @staticmethod
    def signers():
        from nucypher.blockchain.eth.signers.software import ClefSigner, KeystoreSigner
        from nucypher.blockchain.eth.signers.hardware import TrezorSigner
        return {
            ClefSigner.URI_SCHEME: ClefSigner,
            KeystoreSigner.URI_SCHEME: KeystoreSigner,
            TrezorSigner.URI_SCHEME: TrezorSigner,
        }
```

tests/unit/test_web3_signers.py
```diff
@@ -211,11 +217,76 @@ def test_sign_message(mocker, good_signer, mock_account, mock_key):
...
+
+    trezor_transaction = TrezorSigner._format_transaction(web3_transaction_dict)
+
+    assert trezor_transaction['chain_id'] == web3_transaction_dict['chainId']
+    assert trezor_transaction['nonce'] == web3_transaction_dict['nonce']
+    assert trezor_transaction['gas_price'] == web3_transaction_dict['gasPrice']
+    assert trezor_transaction['gas_limit'] == web3_transaction_dict['gas']
+    assert trezor_transaction['to'] == web3_transaction_dict['to']
+    assert trezor_transaction['value'] == web3_transaction_dict['value']
+    assert trezor_transaction['data'] == Web3.toBytes(HexBytes(web3_transaction_dict['data']))
```
derekpierre, line 249: Do we still need this line? We don't modify the 'data' value in `__format_transation` anymore

Pipfile
```diff
@@ -55,7 +57,7 @@ bandit = "*"
 mypy = "*"
 coverage = "*"
 # Develop & Deploy
-py-solc-x = "*"
+py-solc-x = "==0.10.1"
```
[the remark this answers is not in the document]
KPrasch: Yea, I think so :thinking: - not sure why this is reporting as a green line.

nucypher/blockchain/eth/signers/__init__.py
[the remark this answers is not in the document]
KPrasch: Yeah - It's just a matter of namespace - might be able to do some kind of dynamic class creation, or subclass_init hook, but that might overkill.

tests/unit/test_web3_signers.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/signers/base.py
```diff
@@ -0,0 +1,94 @@
...
+            signer_class = cls.SIGNERS[scheme]
+        except KeyError:
+            # This block can be considered the "pass-through"
+            # for providers to be used as external signers.
+            try:
+                from nucypher.blockchain.eth.signers.software import Web3Signer
+                signer = Web3Signer.from_signer_uri(uri=uri)
+            except cls.InvalidSignerURI:
+                message = f'{uri} is not a valid signer URI.  Available schemes: {", ".join(cls.SIGNERS)}'
+                raise cls.InvalidSignerURI(message)
```
tuxxy, line 64: This is a huge anti-pattern, imho.
Implicitly falling back to a Web3 signer seems like something that would occur _outside_ this class and be done explicitly instead.
I'll point out as well that the classes that implement this ABC also do checking to ensure that the URI is valid for the signer (e.g. `TrezorSigner`).

IMHO, when it comes to signing APIs, everything should be explicit and not make assumptions on behavior like this.

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,243 @@
...
+from nucypher.blockchain.eth.signers.base import Signer
+
+
+def handle_trezor_call(device_func):
+    """
+    Decorator for calls to trezorlib that require physical device interactions.
+    """
+    @wraps(device_func)
+    def wrapped(trezor, *args, **kwargs):
+        import usb1  # may not be installable on some systems (consider CI)
```
tuxxy, line 46: I definitely wrote some mocks and things for this specifically to prevent CI errors. Can you not find them?

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,243 @@
...
+        """
+        Signs a message via the TREZOR ethereum sign_message API and returns
+        a named tuple containing the signature and the address used to sign it.
+        This method requires interaction between the TREZOR and the user.
+        """
+        hd_path = self.__get_address_path(checksum_address=checksum_address)
+        signed_message = trezorlib.ethereum.sign_message(self.__client, hd_path, message)
+        return self.SignedMessage(signed_message.signature, signed_message.address)
+
+    def sign_transaction(self,
```
tuxxy, line 190: Has anything changed from my version in this? Can't immediately tell.

tuxxy (changes requested):
Looks great, see my comments about the implicit fallback to the Web3 signer.

nucypher/blockchain/eth/signers/hardware.py
KPrasch: Yeah - I am having a hard time recreating that exact construction - but I am certainly interested in reviving it - perhaps you can lend a hand?

nucypher/blockchain/eth/signers/hardware.py
KPrasch: I suggest doing a diff against `kprasch/trezor` for a detailed view.

nucypher/blockchain/eth/signers/base.py
```diff
@@ -0,0 +1,94 @@
...
+            self.message = f'{account} is locked.'
+            super().__init__(self.message)
+
+    class UnknownAccount(SignerError):
+        def __init__(self, account: str):
+            self.message = f'Unknown account {account}.'
+            super().__init__(self.message)
+
+    @classmethod
+    def from_signer_uri(cls, uri: str) -> 'Signer':
```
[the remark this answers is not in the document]
KPrasch: What advantages do you see to using a class instead of a string here for `'trezor://'` ?

nucypher/blockchain/eth/signers/base.py
KPrasch: It's not abundantly obvious - but this is preserving the pre-exsting behavior wherein `--provider` can be used as a signer.  

I don't really characterize this situation as a *fallback* since `signer` can simply be `None` - rather it's a scheme detection *pass-through* -- since there are many ways to specify node comm ( `ipc://`, `file://`, `http://`... see https://github.com/nucypher/nucypher/blob/main/nucypher/blockchain/eth/interfaces.py#L396 ) using the passthrough here allows a simple and DRY usage of existing scheme detection and web3 connection logic.

nucypher/blockchain/eth/signers/__init__.py
KPrasch: I don't think it adds much to catch the import error of a package included in the standard install - if it did - why not wrap *any* import statement in a try/except?  It will already raise an `ImportError`.

derekpierre (approved):
🎸

KPrasch:
This PR needs an additional reviewer to complete a full end-to-end walk-through with a physical trezor on testnet for acceptance.

requirements.txt
```diff
@@ -42,11 +44,13 @@ ipfshttpclient==0.6.1
 itsdangerous==2.0.0a1
 jinja2==3.0.0a1
 jsonschema==3.2.0
+libusb1==1.8
```
cygnusv, line 47: Can this requirement be problematic?

nucypher/blockchain/eth/__init__.py
```diff
@@ -14,12 +14,3 @@
...
 """
-
-from nucypher.blockchain.eth.signers import Signer, ClefSigner, KeystoreSigner
-
-SIGNERS = {
-    ClefSigner.URI_SCHEME: ClefSigner,
-    KeystoreSigner.URI_SCHEME: KeystoreSigner,
-}
-
-Signer.SIGNERS = SIGNERS
```
cygnusv, line 25: Great 👍

nucypher/blockchain/eth/signers/base.py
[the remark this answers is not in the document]
cygnusv: > The same advantage you get from all the other methods marked as abstract - its existence will be enforced.

That, to me, is a great motivation to do it.

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,243 @@
...
+        return v, r, s
+
+    #
+    # Trezor Signer API
+    #
+
+    @classmethod
+    def from_signer_uri(cls, uri: str) -> 'TrezorSigner':
+        """Return a trezor signer from URI string i.e. trezor:///my/trezor/path """
+        if uri != cls.URI_SCHEME:
```
cygnusv, line 156: I guess you mean something like this?
```
        uri_breakdown = urlparse(uri)
        if uri_breakdown.scheme != cls.URI_SCHEME:
```

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,243 @@
...
+            raise cls.InvalidSignerURI(f'{uri} is not a valid trezor URI scheme')
```
cygnusv, line 157: Also, we can be more helpful here indicating what's the expected trezor URI scheme

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,243 @@
...
+    @handle_trezor_call
+    def sign_message(self, message: bytes, checksum_address: str) -> SignedMessage:
+        """
+        Signs a message via the TREZOR ethereum sign_message API and returns
+        a named tuple containing the signature and the address used to sign it.
+        This method requires interaction between the TREZOR and the user.
+        """
+        hd_path = self.__get_address_path(checksum_address=checksum_address)
+        signed_message = trezorlib.ethereum.sign_message(self.__client, hd_path, message)
+        return self.SignedMessage(signed_message.signature, signed_message.address)
```
cygnusv, line 188: This seems to be breaking the expected return type in the abstract class, as well as wrt the behavior of other signers:
```
@abstractmethod
    def sign_message(self, account: str, message: bytes, **kwargs) -> HexBytes:
        return NotImplemented
```

cygnusv (changes requested):
Some initial comments, looking good in general but there are a few places where I'd like to see some changes.

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,243 @@
...
+
+        # Format transaction fields for Trezor, Lookup HD path, and Sign Transaction
+        # If `chain_id` is included, an EIP-155 transaction signature will be applied
+        # https://github.com/trezor/trezor-core/pull/311
+        trezor_transaction = self._format_transaction(transaction_dict=transaction_dict)
+        n = self.__get_address_path(checksum_address=checksum_address)
+        v, r, s = self.__sign_transaction(n=n, trezor_transaction=trezor_transaction)
+
+        # Format the transaction for eth_account Transaction consumption
+        # v = (v + 2) * (chain_id + 35)
```
fjarri, line 226: This comment is confusing to me. This formula is nowhere to be seen in the code. Should it be here?

requirements.txt
KPrasch: Potentially - on CI for example.  I've taken a few preliminary steps to prevent it from being imported.  This dependency will certainly change the requirements a bit which can be reflected in the docs.  Perhaps we can also consider using a pip extra for this situation as we did before? `pip install nucypher[llama]`

nucypher/blockchain/eth/signers/hardware.py
KPrasch: Agreed.  I pushed an update, let's make this more clear. :+1:

nucypher/blockchain/eth/signers/base.py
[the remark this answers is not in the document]
KPrasch: While I understand the intention here I cannot see the implementation:  In order to make this a property, it needs to be an instance method, however, I actually need a class-scoped attribute for use before instantiation.

How shall I proceed?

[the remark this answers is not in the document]
KPrasch: Additionally - in terms of preventing human error - leaving it as a `NotImplemented` class attribute will still result in a loud and early failure.

nucypher/blockchain/eth/signers/__init__.py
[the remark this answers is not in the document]
fjarri: Wouldn't the code from my comment be enough?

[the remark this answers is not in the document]
KPrasch: Yes - it will work - but the namespace will be `signers.Signer.signers()` which I find less useful than simply `signers.SIGNERS`.

nucypher/blockchain/eth/__init__.py
KPrasch: Heh - it's still there but moved into a new package.

nucypher/blockchain/eth/signers/hardware.py
KPrasch: It's a new situation - I am trying to support both URIs that include a complex scheme, and also simply just a keyword, for example:

`--signer keystore://<PATH>` & `--signer trezor`.

What might a trezor URI be ( I suppose it can contain the device ID - bu that seems unnecessary)?

nucypher/blockchain/eth/signers/hardware.py
KPrasch: I agree - we need to decide on what a Trezor URI even *is*.

nucypher/blockchain/eth/signers/hardware.py
KPrasch: Opened new Issues #2262 and #2261

nucypher/blockchain/eth/signers/base.py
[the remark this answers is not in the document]
fjarri: > In order to make this a property, it needs to be an instance method, however, I actually need a class-scoped attribute for use before instantiation.

You're right, you can't enforce a "class property" in Python, unfortunately. I personally would choose restriction over syntactic sugar here and just use a static classmethod, but I do not insist on it.

> in terms of preventing human error - leaving it as a NotImplemented class attribute will still result in a loud and early failure.

My point was that all the other attributes of this base class already use `abc` machinery, and only this specific one relies on `NotImplemented` - so it seems that we agree that `abc` is a superior approach.

nucypher/blockchain/eth/signers/__init__.py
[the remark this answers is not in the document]
fjarri: Does it ever get used like that? Currently all the use cases are `cls.SIGNERS`.

(In fact, if it stays that way, the constant's name in `__init__.py` should perhaps be changed to `_SIGNERS`)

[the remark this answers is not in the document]
KPrasch: :heavy_check_mark: - Internalized.

nucypher/blockchain/eth/signers/base.py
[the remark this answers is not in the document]
KPrasch: Pushed an update as an abstract classmethod - let me know what you think.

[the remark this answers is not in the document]
fjarri: Looks great to me.

nucypher/blockchain/eth/signers/hardware.py
cygnusv: But the URI can be a trezor with a derivation path, no?

nucypher/blockchain/eth/signers/__init__.py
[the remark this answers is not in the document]
fjarri: Cool, removing the module constant completely is a logical next step.

So, why are you reluctant to make it a method in the `Signer` class? Is it a performance issue, dislike of dynamic imports, or do you want its usage to look as a constant?

nucypher/blockchain/eth/signers/hardware.py
KPrasch: Great idea! :bulb:  that had not occurred to me.  How might this look?

nucypher/blockchain/eth/signers/__init__.py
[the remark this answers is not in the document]
KPrasch: I'm *almost* sold on moving it. :smile: 

Yeah - I think mostly the last two: dislike of dynamic imports, and usage as a constant.

nucypher/blockchain/eth/signers/base.py
```diff
@@ -90,5 +95,5 @@ def sign_transaction(self, transaction_dict: dict) -> HexBytes:
         return NotImplemented
 
     @abstractmethod
-    def sign_message(self, account: str, message: bytes, **kwargs) -> HexBytes:
+    def sign_message(self, account: str, message: bytes, **kwargs) -> SignedMessage:
```
cygnusv, line 102: I don't agree much with this change. As I understand it, `SignedMessage.signer == account` will always hold, making it totally redundant. Returning `HexBytes` seems to me like the cleanest return type.

nucypher/blockchain/eth/signers/hardware.py
cygnusv: (For some reason the related discussion we had on L152 is lost, continuing here). 
See what geth does as the output of `personal_listWallets`:
```
> personal.listWallets
INFO [09-18|19:48:22.776] USB wallet discovered new account        url=trezor://1209:53c1:01 address=0xfoobar path=m/44'/60'/0'/0/0 balance=0 nonce=0
[{
    accounts: [{
        address: "0xfoobar",
        url: "trezor://1209:53c1:01/m/44'/60'/0'/0/0"
    }],
    status: "Trezor v2.3.2 'te_da_cuen' online",
    url: "trezor://1209:53c1:01"
}]
```

We can potentially allow URIs of format similar to `"trezor://1209:53c1:01/m/44'/60'/0'/0/0"`, where the indices are variable depending on the account that is derived. In any case, I think we can do more than assuming the only valid URI is `trezor`.

cygnusv (commented):
Additional review suggesting some changes/TODOs

KPrasch:
Excellent suggestions @cygnusv - Will update.

nucypher/blockchain/eth/signers/base.py
KPrasch: :heavy_check_mark:

cygnusv (approved):
Great job! Thanks @KPrasch and @tuxxy !

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -0,0 +1,241 @@
...
+        # Format contract data field for both trezor and eth_account's Transaction
+        if transaction_dict.get('data') is not None:  # empty string is valid
+            transaction_dict['data'] = Web3.toBytes(HexBytes(transaction_dict['data']))
+
+        # Format transaction fields for Trezor, Lookup HD path, and Sign Transaction
+        # Leave the chain ID in tact for the trezor signing request:
+        # If `chain_id` is included, an EIP-155 transaction signature will be applied
+        # https://github.com/trezor/trezor-core/pull/311
+        if 'chainId' not in transaction_dict:
+            raise self.SignerError('Invalid EIP-155 transaction - "chain_id" field is missing in trezor signing request.')
```
derekpierre, line 219:
```suggestion
            raise self.SignerError('Invalid EIP-155 transaction - "chainId" field is missing in trezor signing request.')
```
```suggestion
            raise self.SignerError('Invalid EIP-155 transaction - "chain_id" field is missing in trezor signing request.')
```

tuxxy (approved):

nucypher/blockchain/eth/signers/hardware.py
cygnusv: I think this is the only RFC pending before merging. Waiting for @KPrasch to confirm if the suggestion is correct, but I think it is.

KPrasch: Trezor actually expects `chain_id` - despite this not being standard with `chainId`

KPrasch (commented):
Confirmed successful staking with physical trezor model T on ibex: https://dpaste.com/3YNCFPFLJ

nucypher/blockchain/eth/signers/hardware.py
derekpierre: Oh right, we convert "chainId" -> "chain_id" for the trezor transaction format. I know we are trying to be proactive here by checking early, perhaps we should check this after the `trezor_transaction = self._format_transaction(transaction_dict=transaction_dict)` to make this clearer...?

PR #2231, opened 2020-09-08
[BlueOyster] Single-domain learning

### LINKED ISSUE #1580: Weirdness in network/domains mapping

### LINKED ISSUE #1869: Test with a hard-coded, versioned node metadata bytestring

### LINKED ISSUE #1871: Increment LearningLoop version and see what happens

### LINKED ISSUE #2236: Relocate `nucypher.datastore.keypairs.py` to `nucypher.crypto`

cygnusv:
As agreed, for the moment we're done with multiple domains. This PR makes the necessary changes to use a single domain everywhere. 

**Closes:** 
* Closes #1871: Bumped Learning Loop version from 1 to 2.
* Closes #1869: See new tests in `tests/integration/learning/test_learning_versions.py`.
* Closes #2236: Relocate `keypairs.py` from `datastore` to `crypto`.
* Closes #1580: For the moment, networks == domains


nucypher/config/node.py
```diff
@@ -55,7 +55,7 @@ class CharacterConfiguration(BaseConfiguration):
     'Sideways Engagement' of Character classes; a reflection of input parameters.
     """
 
-    VERSION = 1  # bump when static payload scheme changes
+    VERSION = 2  # bump when static payload scheme changes
```
derekpierre, line 58: 👍

nucypher/network/nodes.py
```diff
@@ -985,9 +979,19 @@ class UnbondedWorker(InvalidNode):
...
+        """Raised when deserializing a Character from a unexpected and incompatible version."""
+
+    class IsFromTheFuture(UnexpectedVersion):
         """Raised when deserializing a Character from a future version."""
 
+    class AreYouFromThePast(UnexpectedVersion):
+        """Raised when deserializing a Character from a previous, now unsupported version."""
+
+    unknown_version_message = "{} purported to be of version {}, but we're version {}."
+    really_unknown_version_message = "Unable to glean address from node that perhaps purported to be version {}. " \
```
derekpierre, line 992: "perhaps purported" feels redundant.

```suggestion
    really_unknown_version_message = "Unable to glean address from node that purported to be version {}. " \
```

derekpierre (approved):
🎸 - just one suggestion.

examples/finnegans_wake_demo/finnegans-wake-demo.py
```diff
@@ -75,7 +75,7 @@
 ######################################
 
 ALICE = Alice(network_middleware=RestMiddleware(),
-              domains={TEMPORARY_DOMAIN},
+              domain=TEMPORARY_DOMAIN,
```
KPrasch, line 78: Ahhh - so refreshing.

nucypher/blockchain/eth/actors.py
```diff
@@ -107,8 +107,8 @@ def __init__(self, registry: BaseContractRegistry, domains=None, checksum_addres
             self.checksum_address = checksum_address  # type: str
 
         self.registry = registry
-        if domains:  # StakeHolder config inherits from character config, which has 'domains' - #1580
-            self.network = list(domains)[0]
+        if domain:  # StakeHolder config inherits from character config, which has 'domains' - See #1580
```
KPrasch, line 110: Just curious - why is `domain` optional here?

nucypher/characters/lawful.py
```diff
@@ -1209,7 +1210,7 @@ def run(self,
         # if learning:  # TODO: Include learning startup here with the rest of the services?
```
KPrasch, line 1210: @jMyles - May I draw your attention here :point_up_2:

KPrasch (approved):
A comment and a question but no RFCs from me.  Looks good!

nucypher/characters/lawful.py
jMyles: So drawn.  How can I help?

KPrasch: The TODO says it all

nucypher/characters/lawful.py
```diff
@@ -1304,18 +1305,16 @@ def __bytes__(self):
...
 
-        domains = {domain.encode('utf-8') for domain in self.serving_domains}
         as_bytes = bytes().join((version,
                                  self.canonical_public_address,
-                                 bytes(VariableLengthBytestring.bundle(domains)),
+                                 bytes(VariableLengthBytestring(self.serving_domain.encode('utf-8'))),
                                  self.timestamp_bytes(),
                                  bytes(self._interface_signature),
-                                 bytes(decentralized_identity_evidence),
+                                 bytes(VariableLengthBytestring(self.decentralized_identity_evidence)),  # FIXME: Fixed length doesn't work with federated
```
jMyles, line 1317: ...why?  This isn't fixed length?

cygnusv: It is but not for federated, where it's a constant sorrow constant.

nucypher/network/nodes.py
cygnusv: ✅

PR #2240, opened 2020-09-10
Worklock pooling contract

### LINKED ISSUE #1674: Implement a delegation/pooling smart contract

ronmnm:
Fixes #1674
The contract allows users to delegate liquid tokens and participate in worklock.
Key features:
- Contract owner is multisig wallet
- Each delegator can withdraw ETH or NU at any time

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/staking_contracts/WorkLockPoolingContract.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/staking_contracts/WorkLockPoolingContract.sol
@@ -27,6 +27,7 @@ contract WorkLockPoolingContract is InitializableStakingContract, Ownable {
     );
     event ETHWithdrawn(address indexed sender, uint256 value);
     event DepositSet(address indexed sender, bool value);
+    event Bid(address indexed sender, uint256 depositedETH);
     event Claimed(address indexed sender, uint256 claimedTokens);
     event Refund(address indexed sender, uint256 refundETH);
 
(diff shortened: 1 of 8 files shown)


nucypher/blockchain/eth/sol/source/contracts/staking_contracts/WorkLockPoolingContract.sol
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    /**
+     * @notice Get available reward for all delegators and owner
+     */
+    function getAvailableReward() public view returns (uint256) {
+        uint256 stakedTokens = escrow.getAllTokens(address(this));
+        uint256 freeTokens = token.balanceOf(address(this));
+        uint256 reward = stakedTokens.add(freeTokens).sub(totalDepositedTokens);
+        if (reward > freeTokens) {
```
cygnusv, line 206:
```suggestion
        if (stakedTokens > totalDepositedTokens) {
```
This should be equivalent, but I find it more clear, since it's very easy to understand that this covers the case when you have more tokens in StakingEscrow that you deposited in the first place.

nucypher/blockchain/eth/sol/source/contracts/staking_contracts/WorkLockPoolingContract.sol
```diff
@@ -0,0 +1,439 @@
...
+            return freeTokens;
+        }
+        return reward;
```
cygnusv, line 209: I'm having a hard time understanding the logic here.

We have `reward = staked + free - total`
Then we have two cases: 
1) if `reward > free` (which is equivalent to `staked + free - total > free`, and hence to `staked > total`, see comment above), we return `free`
2) otherwise, we return `reward`.

So essentially, when we have more tokens staked than deposited, the available rewards are coming exclusively from the pool balance (?); otherwise, it's the pool balance + rewards in StakingEscrow.  I'm afraid I don't understand.

cygnusv (commented):
Some initial comments

nucypher/blockchain/eth/sol/source/contracts/staking_contracts/WorkLockPoolingContract.sol
vzotova: Yeah, that's equivalent, for me current option is more clear because I see that any reward can't be more than contract has. What others will say?

nucypher/blockchain/eth/sol/source/contracts/staking_contracts/WorkLockPoolingContract.sol
vzotova: Looks like you said everything correct. Reward is everything that more than deposited, so we get info about all tokens which belonging to pool and then subtract totalDeposited. And we can't give reward more than contract has at contract address

KPrasch (approved):
The pre-worklock time crunch pushed this PR to completion and it represents an entire project team's work output built over several weeks all at once so it's hard to know when it is sufficiently reviewed without dedicating several full engineering days to reviewing it.

Despite that - this PR's code has been established in (third-party) production and backed by NU stakes for more than a month which does bring some merit - I have read through this diff, attended a walk-through from the authors - I approve on that basis.

cygnusv (approved):
Thanks @ronmnm and @vzotova ! This is a great piece of work, and hopefully the basis of other staking pool contracts. Let's merge this now and continue iterating if necessary.

PR #2241, opened 2020-09-10
Allow learners to explicity opt out of verifying stranger bonding

KPrasch:
This PR is motivated by reducing the number of calls nucypher/nucypher-monitor needs to make to ethereum clients when operating as a `Crawler(Learner)`.

### DIFF
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -203,7 +203,8 @@ class Learner:
                  node_storage=None,
                  save_metadata: bool = False,
                  abort_on_learning_error: bool = False,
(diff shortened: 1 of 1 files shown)


KPrasch:
This PR still needs a test to show that a learner can successfully opt out of the on-chain staking check.

cygnusv (approved):

derekpierre (approved):
🎸 - if we can a unit test would be 👌

KPrasch:
Follow up Issue ( low priority ) https://github.com/nucypher/nucypher/issues/2252

PR #2242, opened 2020-09-11
Add thread lock to MockLMDB

### LINKED ISSUE #2176: lmdb: Invalid reuse of reader locktable slot
```
    @contextmanager
    def describe(self,
                 record_type: Type['DatastoreRecord'],
                 record_id: Union[int, str],
                 writeable: bool = False) -> Type['DatastoreRecord']:
        """
        This method is used to perform CRUD operations on the datastore within
        the safety of a context manager by returning an instance of the
        `record_type` identified by the `record_id` provided.
    
        When `writeable` is `False`, the record returned by this method
        cannot be used for any operations that write to the datastore. If an
        attempt is made to retrieve a non-existent record whilst `writeable`
        is `False`, this method raises a `RecordNotFound` error.
    
        When `writeable` is `True`, the record can be used to perform writes
        on the datastore. In the event an error occurs during the write, the
        transaction will be aborted and no data will be written, and a
        `DatastoreTransactionError` will be raised.
    
        If the record is used outside the scope of the context manager, any
        writes or reads will error.
        """
        with suppress(ValueError):
            # If the ID can be converted to an int, we do it.
            record_id = int(record_id)
    
>       with self.__db_env.begin(write=writeable) as datastore_tx:
E       lmdb.BadRslotError: mdb_txn_begin: MDB_BAD_RSLOT: Invalid reuse of reader locktable slot

nucypher/datastore/datastore.py:125: BadRslotError
```

See full traceback in CI Build: 

https://app.circleci.com/pipelines/github/nucypher/nucypher/5509/workflows/1994c9bf-611f-4e3b-8732-39ecc67418c0/jobs/97113

tuxxy:
Attempts to resolve #2176

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -76,12 +76,12 @@ class TestLMDBEnv:
     @contextmanager
     def begin(self, *args, **kwargs):
         try:
-            self.thread_lock.acquire()
-            with self.open() as lmdb_env:
-                with lmdb_env.begin(*args, **kwargs) as lmdb_tx:
-                    yield lmdb_tx
+            with self.thread_lock:
+                with self.open() as lmdb_env:
+                    with lmdb_env.begin(*args, **kwargs) as lmdb_tx:
+                        yield lmdb_tx
         finally:
-            self.thread_lock.release()
+            pass
 
     def path(self):
         return self.db_path
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -67,16 +67,16 @@ class TestLMDBEnv:
     __test__ = False    # Prohibit pytest from collecting this
 
     LMDB_OPEN_FUNC = lmdb.open
+    THREAD_LOCK = threading.Lock()
 
     def __init__(self, *args, **kwargs):
         self.db_path = args[0]
         self.open = partial(self.LMDB_OPEN_FUNC, *args, **kwargs)
-        self.thread_lock = threading.Lock()
 
     @contextmanager
     def begin(self, *args, **kwargs):
         try:
-            with self.thread_lock:
+            with self.THREAD_LOCK:
                 with self.open() as lmdb_env:
                     with lmdb_env.begin(*args, **kwargs) as lmdb_tx:
                         yield lmdb_tx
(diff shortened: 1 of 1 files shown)


tests/integration/conftest.py
```diff
@@ -70,15 +71,17 @@ class TestLMDBEnv:
     def __init__(self, *args, **kwargs):
         self.db_path = args[0]
         self.open = partial(self.LMDB_OPEN_FUNC, *args, **kwargs)
+        self.thread_lock = threading.Lock()
 
     @contextmanager
     def begin(self, *args, **kwargs):
         try:
+            self.thread_lock.acquire()
```
derekpierre, line 79: Should the acquisition of the lock be outside of the try-finally block? i.e. we don't want to call `release()` if there was a problem acquiring the lock...

```suggestion
        self.thread_lock.acquire()
        try:
```

cygnusv: https://docs.python.org/3/library/threading.html#using-locks-conditions-and-semaphores-in-the-with-statement describe exactly what Derek propose. But perhaps it'd be even better to use the lock as a context manager, as the docs show.

derekpierre: ah, even better - 👍

derekpierre:
Seems like the issue still occurs - https://app.circleci.com/pipelines/github/nucypher/nucypher/5848/workflows/8559564e-d78f-452c-b0a1-484aac0df215/jobs/101963

KPrasch:
Looks like the problem persists on CI.

derekpierre (commented):
@tuxxy Well that seemed to have done the trick. 🎉 

Were there multiple instances of `TestLMDBEnv` being created, so a per-instance lock was useless?

cygnusv (approved):
This seems reasonable to me, but I really can't say. 
The failing test is one of our flaky list, so the fact that it's working now doesn't necessarily means it's fixed, time will say, but looking good :)
Thanks @tuxxy !

tuxxy:
@derekpierre Yeah, that's what was going on. It just wasn't using the same `Lock`.

Keep a look out for a failure in the event this didn't work, but I think this will do the trick.

derekpierre (approved):
🎸

PR #2063, opened 2020-06-02
[WIP] Parallel `Bob.retreive()`

fjarri:
Attempting to fix issue #1176 (also the working changes should fix #1359)

This is a proof of concept, I tried to simplify `retreive()` and isolate the parallelizable parts.  The non-parallel version works, but when futures are added (the last commit), there's some strange SQL error...
```
    def do_execute(self, cursor, statement, parameters, context=None):
>       cursor.execute(statement, parameters)
E       sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) no such table: policyarrangements
E       [SQL: SELECT policyarrangements.id AS policyarrangements_id, policyarrangements.expiration AS policyarrangements_expiration, policyarrangements.kfrag AS policyarrangements_kfrag, policyarrangements.alice_verifying_key_id AS policyarrangements_alice_verifying_key_id, policyarrangements.alice_signature AS policyarrangements_alice_signature, policyarrangements.created_at AS policyarrangements_created_at, keys_1.id AS keys_1_id, keys_1.fingerprint AS keys_1_fingerprint, keys_1.key_data AS keys_1_key_data, keys_1.is_signing AS keys_1_is_signing, keys_1.created_at AS keys_1_created_at 
E       FROM policyarrangements LEFT OUTER JOIN keys AS keys_1 ON keys_1.id = policyarrangements.alice_verifying_key_id 
E       WHERE policyarrangements.id = ?
E        LIMIT ? OFFSET ?]
E       [parameters: (<memory at 0x107f37340>, 1, 0)]
E       (Background on this error at: http://sqlalche.me/e/e3q8)

../../../.local/share/virtualenvs/nucypher-3pdGgPxB/lib/python3.8/site-packages/sqlalchemy/engine/default.py:590: OperationalError
```
Some race condition?

PR #2244, opened 2020-09-13
Simplify Bob.retrieve()

fjarri:
Since PR #2063 (parallel `retrieve()`) is currently blocked by LMDB issues, this PR is created to at least make some progress on the parts of #2063 that simplify the logic flow in `retrieve()` and prepare it to be split into independent worker processes.

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -480,6 +480,26 @@ class Alice(Character, BlockchainPolicyAuthor):
         return controller
 
 
+def ensure_correct_sender(message_kit: UmbralMessageKit,
+                          enrico: Optional["Enrico"] = None,
+                          policy_encrypting_key: Optional[UmbralPublicKey] = None):
+    """
+    Make sure that the sender of the message kit is set and corresponds to
+    the given ``enrico``, or create it from the given ``policy_encrypting_key``.
+    """
+    if message_kit.sender:
+        if enrico and message_kit.sender != enrico:
+            raise ValueError
+    elif enrico:
+        message_kit.sender = enrico
+    elif message_kit.sender_verifying_key and policy_encrypting_key:
+        # Well, after all, this is all we *really* need.
+        message_kit.sender = Enrico.from_public_keys(verifying_key=message_kit.sender_verifying_key,
+                                                     policy_encrypting_key=policy_encrypting_key)
+    else:
+        raise TypeError
+
+
 class Bob(Character):
     banner = BOB_BANNER
     _interface_class = BobInterface
@@ -779,35 +799,25 @@ class Bob(Character):
         # Part I: Assembling the WorkOrders.
         capsules_to_activate = set(mk.capsule for mk in message_kits)
 
+        # Normalization
         for message in message_kits:
+            ensure_correct_sender(message,
+                                  enrico=enrico,
+                                  policy_encrypting_key=policy_encrypting_key)
 
-            # Two sanity checks before we get into network activity.
-            # First sanity check: We have some representation of the sender, so that we can later check the signature.
+        # Sanity check: If we're not using attached cfrags, we don't want a Capsule which has them.
+        if not use_attached_cfrags and any(len(message.capsule) > 0 for message in message_kits):
+            raise TypeError(
+                "Not using cached retrievals, but the MessageKit's capsule has attached CFrags. "
+                "In order to retrieve this message, you must set cache=True. "
+                "To use Bob in 'KMS mode', use cache=False the first time you retrieve a message.")
 
-            if message.sender:
-                if enrico and message.sender != enrico:
-                    raise ValueError
-            elif enrico:
-                message.sender = enrico
-            elif message.sender_verifying_key and policy_encrypting_key:
-                # Well, after all, this is all we *really* need.
-                message.sender = Enrico.from_public_keys(verifying_key=message.sender_verifying_key,
-                                                         policy_encrypting_key=policy_encrypting_key)
-            else:
-                raise TypeError
+        # OK, with the sanity checks behind us, we'll proceed to the WorkOrder assembly.
+        # We'll start by following the treasure map, setting the correctness keys, and attaching cfrags from
+        # WorkOrders that we have already completed in the past.
 
-            # Second sanity check: If we're not using attached cfrags, we don't want a Capsule which has them.
-
+        for message in message_kits:
             capsule = message.capsule
-
-            if len(capsule) > 0:
-                if not use_attached_cfrags:
-                    raise TypeError(
-                        "Not using cached retrievals, but the MessageKit's capsule has attached CFrags.  In order to retrieve this message, you must set cache=True.  To use Bob in 'KMS mode', use cache=False the first time you retrieve a message.")
-
-            # OK, with the sanity checks behind us, we'll proceed to the WorkOrder assembly.
-            # We'll start by following the treasure map, setting the correctness keys, and attaching cfrags from
-            # WorkOrders that we have already completed in the past.
 
             capsule.set_correctness_keys(receiving=self.public_keys(DecryptingPower))
             capsule.set_correctness_keys(verifying=alice_verifying_key)
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -590,7 +590,7 @@ class Bob(Character):
         return unknown_ursulas, known_ursulas, treasure_map.m
 
     def get_treasure_map(self, alice_verifying_key, label):
-        _hrac, map_id = self.construct_hrac_and_map_id(verifying_key=alice_verifying_key, label=label)
+        map_identifier = self.construct_map_id(verifying_key=alice_verifying_key, label=label)
 
         if not self.known_nodes and not self._learning_task.running:
             # Quick sanity check - if we don't know of *any* Ursulas, and we have no
@@ -602,11 +602,6 @@ class Bob(Character):
             if not self.known_nodes:
                 raise self.NotEnoughTeachers("Can't retrieve without knowing about any nodes at all.  Pass a teacher or seed node.")
 
-        # Ugh stupid federated only mode....
-        if not self.federated_only:
-            map_identifier = _hrac.hex()
-        else:
-            map_identifier = map_id
         treasure_map = self.get_treasure_map_from_known_ursulas(self.network_middleware,
                                                                 map_identifier)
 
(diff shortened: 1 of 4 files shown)


KPrasch:
Nice progress here - this demonstrates which tasks can be better encapsulated and optimized.  I surmise this PR will take a while for acceptance amidst the network launch.

KPrasch:
> Since PR #2063 (parallel retrieve()) is currently blocked by LMDB issues

@fjarri - Is this still the case?

fjarri:
I haven't checked since then, but I imagine it would be. Let me see how hard it is to rebase #2063

nucypher/characters/lawful.py
```diff
@@ -678,28 +663,104 @@ def work_orders_for_capsules(self,
...
-
-        return cfrags
-
     def join_policy(self, label, alice_verifying_key, node_list=None, block=False):
         if node_list:
             self._node_ids_to_learn_about_immediately.update(node_list)
         treasure_map = self.get_treasure_map(alice_verifying_key, label)
         self.follow_treasure_map(treasure_map=treasure_map, block=block)
 
+    def _fix_message_kits(self, message_kits, enrico, policy_encrypting_key, use_attached_cfrags):
```
KPrasch, line 678: ...fix?

fjarri: The logic that I put here checks if the `sender` attribute is correct and changes it under some conditions. So, I think, "fix" is justified, but it can be made more specific, "ensure_correct_sender" or something?

It may work better as a method of `UmbralMessageKit`, but all the classes in `characters/control/specifications` are very barebones, so I'm not sure it will fit there.

Of course, ideally I'd get rid of mutation completely. But this PR just tries to extract existing pieces of logic into functions.

KPrasch:
```suggestion
    def _ensure_correct_sender(self, message_kits, enrico, policy_encrypting_key, use_attached_cfrags):
```

vepkenez: The actual UmbralMessagekit class is here:
https://github.com/nucypher/nucypher/blob/main/nucypher/crypto/kits.py#L88
Maybe it would be better as a method on this class.

KPrasch:
Needs a rebase

nucypher/characters/lawful.py
fjarri: Making it a method there would require importing `Enrico` from `lawful.py`, which will lead to a circular dependency... For now just made it a standalone function with a more specific name.

nucypher/policy/collections.py
```diff
@@ -186,6 +186,20 @@ def public_id(self) -> str:
         """
         return self._id
 
+    @classmethod
+    def from_obj_or_bytes(cls, obj):
+
+        if isinstance(obj, cls):
+            return obj
+
+        # TODO: This LBYL is ugly and fraught with danger.  NRN
```
KPrasch, line 195: Let's track this with an Issue.

nucypher/characters/lawful.py
```diff
@@ -461,6 +462,20 @@ def revoke():
         return controller
 
 
+def ensure_correct_sender(message_kit: UmbralMessageKit, enrico=None, policy_encrypting_key=None):
```
KPrasch, line 465: Annotations and docstring please! unless you have near-term plans to make additional changes here.

nucypher/characters/lawful.py
```diff
@@ -670,12 +672,12 @@ def work_orders_for_capsules(self,
                 raise KeyError(f"Bob doesn't have the TreasureMap {map_id}; can't generate work orders.")
 
         incomplete_work_orders = OrderedDict()
-        complete_work_orders = OrderedDict()
+        complete_work_orders = defaultdict(lambda: [])
```
KPrasch, line 675: why the `lambda` here?

KPrasch:
```suggestion
        complete_work_orders = defaultdict(list)
```

nucypher/characters/lawful.py
```diff
@@ -706,28 +708,83 @@ def work_orders_for_capsules(self,
                 # TODO: Presently, the order here is haphazard .  Do we want to do the complete or incomplete specifically first? NRN
                 break
 
-        if incomplete_work_orders == OrderedDict():
-            self.log.warn("No new WorkOrders created. Try calling this with different parameters.")  # TODO: Clearer instructions.  NRN
+        if len(incomplete_work_orders) == 0:
```
KPrasch, line 711:
```suggestion
        if not incomplete_work_orders:
```

KPrasch (changes requested):
Just a few comments, good code reorg :+1:

nucypher/policy/collections.py
fjarri: Funny thing, I actually can't remember now why I needed this method. It's not used anywhere. Seems to have been partially based on the bit in the beginning of `retrieve()`, and that's where the TODO comes from, too. Don't even know what LBYL means. 

Deleted it completely.

nucypher/characters/lawful.py
fjarri: Done.

nucypher/characters/lawful.py
fjarri: Yeah, must have been brain freeze. Fixed.

nucypher/characters/lawful.py
fjarri: Fixed

nucypher/characters/lawful.py
```diff
@@ -565,8 +552,16 @@ def follow_treasure_map(self,
...
+        alice = Alice.from_public_keys(verifying_key=alice_verifying_key)
+        compass = self.make_compass_for_alice(alice)
+        try:
+            treasure_map.orient(compass)
+        except treasure_map.InvalidSignature:
+            raise  # TODO: Maybe do something here?  NRN
+
     def get_treasure_map(self, alice_verifying_key, label):
-        _hrac, map_id = self.construct_hrac_and_map_id(verifying_key=alice_verifying_key, label=label)
+        map_identifier = self.construct_map_id(verifying_key=alice_verifying_key, label=label)
```
KPrasch, line 564: :+1:

KPrasch (approved):

nucypher/characters/lawful.py
```diff
@@ -720,23 +720,77 @@ def work_orders_for_capsules(self,
...
+                # TODO: WARNING - This block is untested.
+                from nucypher.policy.collections import IndisputableEvidence
+                evidence = IndisputableEvidence(task=pre_task, work_order=work_order)
+                # I got a lot of problems with you people ...
+                the_airing_of_grievances.append(evidence)
+
+        if the_airing_of_grievances:
+            return False, the_airing_of_grievances
+        else:
+            return True, cfrags
```
cygnusv, line 772: What do you think about adding a type hint to this method for `Tuple[bool, list]`?

nucypher/characters/lawful.py
```diff
@@ -833,9 +813,8 @@ def retrieve(self,
 
         # Normalization
         for message in message_kits:
-            ensure_correct_sender(message,
-                                  enrico=enrico,
-                                  policy_encrypting_key=policy_encrypting_key)
+            message.ensure_correct_sender(enrico=enrico,
+                                          policy_encrypting_key=policy_encrypting_key)
```
cygnusv, line 817: ahh much better!

cygnusv (approved):
Great work, @fjarri ! This part of the code is a bit ugly, I must admit, but this PR is improving it a lot 👍

nucypher/characters/lawful.py
fjarri: I actually think that throwing an exception would be better, but this function was intended to be executed in parallel, and it all depends on whether the parallel runner can properly track exceptions. Added the signature for now.

nucypher/characters/lawful.py
```diff
@@ -730,28 +718,83 @@ def work_orders_for_capsules(self,
...
-
-        return cfrags
-
     def join_policy(self, label, alice_verifying_key, node_list=None, block=False):
         if node_list:
             self._node_ids_to_learn_about_immediately.update(node_list)
         treasure_map = self.get_treasure_map(alice_verifying_key, label)
         self.follow_treasure_map(treasure_map=treasure_map, block=block)
 
+    def _filter_work_orders_and_capsules(self, work_orders, capsules, m):
```
derekpierre, line 733: Should we provide a type hint for our return values here `-> Tuple [List, Set]`?

nucypher/crypto/kits.py
```diff
@@ -106,6 +108,26 @@ def sender(self, enrico):
...
+    def ensure_correct_sender(self,
+                              enrico: Optional["Enrico"] = None,
+                              policy_encrypting_key: Optional[UmbralPublicKey] = None):
+        """
+        Make sure that the sender of the message kit is set and corresponds to
+        the given ``enrico``, or create it from the given ``policy_encrypting_key``.
+        """
+        if self.sender:
+            if enrico and self.sender != enrico:
+                raise ValueError
```
derekpierre, line 120: Perhaps a error message here:
```suggestion
                raise ValueError("Unexpected sender {enrico}")
```

or something of the sort...?

derekpierre (commented):
Any reason for the empty newsfragment?

nucypher/characters/lawful.py
fjarri: Done. Added more typing here and in `_reencrypt()` as well.

nucypher/crypto/kits.py
fjarri: Added a message here and in another branch as well.

fjarri:
> Any reason for the empty newsfragment?

There don't seem to be any changes affecting the user (or even most of the codebase).

derekpierre:
> There don't seem to be any changes affecting the user (or even most of the codebase).
Should we just not have a release note entry?

fjarri:
With an empty newsfragment we won't, will we? But the CI requires there to be a newsfragment file present.

derekpierre:
> With an empty newsfragment we won't, will we? But the CI requires there to be a newsfragment file present.

The empty file still creates a link (the text for the link being the PR number) in the release notes - albeit without any associated description. You can run `towncrier --draft` locally to see what the rst output would be.

If the PR doesn't have anything meaningful to the user (the audience of the release notes), then my view is why even provide the newsfragment and the link. The CI check is not smart enough to understand the context of the PR but instead ensures that the dev has at least thought about the newsfragment and whether it should be present. The CI check not passing is irrelevant in my mind i.e. PR context dictates the presence of newsfragments and not we should include one to make a primitive CI check happy.

@KPrasch and I had a discussion about it in PR 2474 - https://github.com/nucypher/nucypher/pull/2474. That being said, he feels a bit differently than I do 😅.

KPrasch:
Rebase over main will fix the failing unit test.

KPrasch:
Thanks @fjarri :boom:

PR #2239, opened 2020-09-09
[WIP] [Do not merge] We need to batch/throttle when obtaining events - some providers will limit us. More Limited scope PR #2245 to be addressed first

derekpierre:
Some node providers limit the number of blocks for which events can be checked in one call. The smallest I've heard of so far is alchemy with 1000. Without limiting the number of blocks and iterating the calls, the one filter query will fail.

Example
```
  File "/Users/derek/.local/share/virtualenvs/nucypher-z-zdnWw9/lib/python3.7/site-packages/web3/eth.py", line 498, in getFilterLogs
    RPC.eth_getFilterLogs, [filter_id],
  File "/Users/derek/.local/share/virtualenvs/nucypher-z-zdnWw9/lib/python3.7/site-packages/web3/manager.py", line 153, in request_blocking
    raise ValueError(response["error"])
ValueError: {'code': -32600, 'message': 'Requested block range for eth_getLogs is greater than the limit of 1000 blocks. Please reduce range and retry.'}
```

Places to check
- `nucypher stake events...`
- `nucypher status events...`
(related to #1670)
- `CommitmentMadeEventMetricsCollector` for prometheus metrics collection (related to #2135 )

For the CLI, we may want to add the `--throttle`/`--batch` option...TBD?

First iteration for prometheus in #2245.

PR #2245, opened 2020-09-14
Batch/Throttle events filter results 1000 (default) blocks at a time for CommitmentMade prometheus metrics collector

derekpierre:
Some node providers limit the number of blocks for which events can be checked in one call. The smallest I've heard of so far is alchemy with 1000. Without limiting the number of blocks and iterating the calls, the one filter query will fail.

Example
```
  File "/Users/derek/.local/share/virtualenvs/nucypher-z-zdnWw9/lib/python3.7/site-packages/web3/eth.py", line 498, in getFilterLogs
    RPC.eth_getFilterLogs, [filter_id],
  File "/Users/derek/.local/share/virtualenvs/nucypher-z-zdnWw9/lib/python3.7/site-packages/web3/manager.py", line 153, in request_blocking
    raise ValueError(response["error"])
ValueError: {'code': -32600, 'message': 'Requested block range for eth_getLogs is greater than the limit of 1000 blocks. Please reduce range and retry.'}
```

This PR has limited scope relative to #2239. Instead of trying to tackle all instances of event filters, only tackle the prometheus metrics collection for now, since the scope of the query for this purposes is limited - a max 2 periods of ethereum blocks is searched. The CLI commands (`nucypher status events`, and `nucypher stake events`) can search many more blocks and throttling would cause efficiency issues.

The problem with modifying CLI commands is that their scope cannot be enforced. `status events` can specify the `--from-block` option to increase the number of blocks to search, and `stake events` searches from block 0 by default. We could improve these by limiting the from-block to be from contract creation, since searching earlier than the contract is useless, but that is only sufficient in the short-term, and will be eventually still be an issue later on and thereby needs more thought.

For this PR we settle for fixing prometheus event data collection now, and we will address CLI commands at a later time.

### DIFF
--- a/nucypher/blockchain/eth/events.py
+++ b/nucypher/blockchain/eth/events.py
@@ -14,9 +14,12 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
+import os
+
 from web3.contract import Contract
 
 from nucypher.blockchain.eth.interfaces import BlockchainInterfaceFactory
+from nucypher.config.constants import NUCYPHER_EVENTS_THROTTLE_MAX_BLOCKS
 
 
 class EventRecord:
@@ -83,12 +86,15 @@ class ContractEventsThrottler:
     """
     Enables Contract events to be retrieved in batches.
     """
+    # default to 1000 - smallest default heard about so far (alchemy)
+    DEFAULT_MAX_BLOCKS_PER_CALL = int(os.environ.get(NUCYPHER_EVENTS_THROTTLE_MAX_BLOCKS, 1000))
+
     def __init__(self,
                  agent: 'EthereumContractAgent',
                  event_name: str,
                  from_block: int,
                  to_block: int = None,  # defaults to latest block
-                 max_events_per_call: int = 1000,  # smallest default heard about so far (alchemy)
+                 max_blocks_per_call: int = DEFAULT_MAX_BLOCKS_PER_CALL,
                  **argument_filters):
         if not agent:
             raise ValueError(f"Contract agent must be provided")
@@ -102,16 +108,16 @@ class ContractEventsThrottler:
         if to_block <= from_block:
             raise ValueError(f"Invalid block range provided ({from_block} - {to_block})")
 
-        self.max_events_per_call = max_events_per_call
+        self.max_blocks_per_call = max_blocks_per_call
         self.argument_filters = argument_filters
 
     def __iter__(self):
         current_from_block = self.from_block
-        current_to_block = min(self.from_block + self.max_events_per_call, self.to_block)
+        current_to_block = min(self.from_block + self.max_blocks_per_call, self.to_block)
         while current_from_block < current_to_block:
             for event_record in self.event_filter(from_block=current_from_block,
                                                   to_block=current_to_block,
                                                   **self.argument_filters):
                 yield event_record
             current_from_block = current_to_block
-            current_to_block = min(current_from_block + self.max_events_per_call, self.to_block)
+            current_to_block = min(current_from_block + self.max_blocks_per_call, self.to_block)
--- a/nucypher/blockchain/eth/events.py
+++ b/nucypher/blockchain/eth/events.py
@@ -77,3 +77,41 @@ class ContractEvents:
     def __iter__(self):
         for event_name in self.names:
             yield self[event_name]
+
+
+class ContractEventsThrottler:
+    """
+    Enables Contract events to be retrieved in batches.
+    """
+    def __init__(self,
+                 agent: 'EthereumContractAgent',
+                 event_name: str,
+                 from_block: int,
+                 to_block: int = None,  # defaults to latest block
+                 max_events_per_call: int = 1000,  # smallest default heard about so far (alchemy)
+                 **argument_filters):
+        if not agent:
+            raise ValueError(f"Contract agent must be provided")
+        if not event_name:
+            raise ValueError(f"Event name must be provided")
+
+        self.event_filter = agent.events[event_name]
+        self.from_block = from_block
+        self.to_block = to_block if to_block else agent.blockchain.client.block_number
+        # validity check of block range
+        if to_block <= from_block:
+            raise ValueError(f"Invalid block range provided ({from_block} - {to_block})")
+
+        self.max_events_per_call = max_events_per_call
+        self.argument_filters = argument_filters
+
+    def __iter__(self):
+        current_from_block = self.from_block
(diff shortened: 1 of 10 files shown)


nucypher/blockchain/eth/events.py
```diff
@@ -77,3 +80,44 @@ def __getattr__(self, event_name: str):
...
+            raise ValueError(f"Contract agent must be provided")
+        if not event_name:
+            raise ValueError(f"Event name must be provided")
+
+        self.event_filter = agent.events[event_name]
+        self.from_block = from_block
+        self.to_block = to_block if to_block else agent.blockchain.client.block_number
+        # validity check of block range
+        if to_block <= from_block:
+            raise ValueError(f"Invalid block range provided ({from_block} - {to_block})")
```
cygnusv, line 109: What do you think about allowing `to_block='latest'`?

nucypher/utilities/prometheus/collector.py
```diff
@@ -330,13 +332,23 @@ def initialize(self, metrics_prefix: str, registry: CollectorRegistry) -> None:
...
             arg_filters = {'staker': self.staker_address, 'period': last_committed_period}
-            # check from the beginning - difficult to determine block number for previous period
-            # however, 'staker' and 'period' filters should help reduce load
-            initial_event_filter = self.contract_agent.contract.events[self.event_name].createFilter(
-                fromBlock=0, argument_filters=arg_filters)
-            events = initial_event_filter.get_all_entries()
-            for event in events:
-                self._event_occurred(event)
+            latest_block = self.contract_agent.blockchain.client.block_number
+            previous_period = self.contract_agent.get_current_period() - 1  # just in case
```
cygnusv, line 336: I know it's not the scope of this PR, but I find weird how are we assuming here that `self.contract_agent` is a `StakingEscrowAgent`. Perhaps we should use a specific instance for it?

tests/unit/test_blockchain_utils.py
```diff
@@ -0,0 +1,66 @@
...
+
+    current_period = epoch_to_period(epoch=timestamp, seconds_per_period=SECONDS_PER_PERIOD)
+    assert current_period == (timestamp // SECONDS_PER_PERIOD)
+
+
+def test_period_to_epoch():
+    current_period = 12345678
+    epoch = period_to_epoch(period=current_period, seconds_per_period=SECONDS_PER_PERIOD)
+    assert epoch == (current_period * SECONDS_PER_PERIOD)
+
```
cygnusv, line 38: Thanks for introducing these tests!

nucypher/blockchain/eth/events.py
```diff
@@ -77,3 +80,44 @@ def __getattr__(self, event_name: str):
...
+    def __iter__(self):
+        current_from_block = self.from_block
+        current_to_block = min(self.from_block + self.max_blocks_per_call, self.to_block)
+        while current_from_block < current_to_block:
+            for event_record in self.event_filter(from_block=current_from_block,
+                                                  to_block=current_to_block,
+                                                  **self.argument_filters):
+                yield event_record
+            current_from_block = current_to_block
+            current_to_block = min(current_from_block + self.max_blocks_per_call, self.to_block)
```
cygnusv, line 123: This is a very clever piece of code 😎

cygnusv (approved):

vzotova (approved):

nucypher/blockchain/eth/events.py
derekpierre: `to_block` is currently provided as an int in the constructor, but if it is None (default) then the latest block is used - see L106. I didn't want `to_block` to be a string since it just causes more issues with validating the string.

nucypher/utilities/prometheus/collector.py
derekpierre: `self.contract_agent` is obtained from the parent `EventMetricsCollector` class, which is a base class for all event collection including for other agents eg. WorkLock, PolicyManager.

Definitely open to improving the `contract_agent` to be more explicit if you have insight there - see https://github.com/derekpierre/nucypher/blob/prometheus-events-throttle/nucypher/utilities/prometheus/metrics.py#L204  for more context.

tests/unit/test_blockchain_utils.py
derekpierre: 👍

nucypher/blockchain/eth/events.py
```diff
@@ -77,3 +80,44 @@ def __getattr__(self, event_name: str):
...
         for event_name in self.names:
             yield self[event_name]
+
+
+class ContractEventsThrottler:
+    """
+    Enables Contract events to be retrieved in batches.
+    """
+    # default to 1000 - smallest default heard about so far (alchemy)
+    DEFAULT_MAX_BLOCKS_PER_CALL = int(os.environ.get(NUCYPHER_EVENTS_THROTTLE_MAX_BLOCKS, 1000))
```
KPrasch, line 90: They are piling up!  

Here's a follow-up <https://github.com/nucypher/nucypher/issues/2255>

nucypher/blockchain/eth/events.py
```diff
@@ -77,3 +80,44 @@ def __getattr__(self, event_name: str):
...
+        self.from_block = from_block
+        self.to_block = to_block if to_block else agent.blockchain.client.block_number
+        # validity check of block range
+        if to_block <= from_block:
+            raise ValueError(f"Invalid block range provided ({from_block} - {to_block})")
+
+        self.max_blocks_per_call = max_blocks_per_call
+        self.argument_filters = argument_filters
+
+    def __iter__(self):
```
KPrasch, line 114: tasty

nucypher/blockchain/eth/events.py
KPrasch: Some code authors think "clever" is actually an insult. ;-)

Definitely a solid line of code however, core of the concept - perhaps add a comment here :+1:

KPrasch (approved):
:cowboy_hat_face: - Do we have the mainnet NU contract block number as a constant in the codebase somewhere?

nucypher/blockchain/eth/events.py
derekpierre: 👍

nucypher/blockchain/eth/events.py
derekpierre: 👍  - I've definitely used "clever" with others regarding their code and meant it as a positive. Hopefully, no one was insulted by my use of it 😅

cygnusv: Oh my, everything can sound insulting in English

derekpierre:
>  Do we have the mainnet NU contract block number as a constant in the codebase somewhere?

@KPrasch I don't believe we do. @cygnusv and I were talking about that. It will likely be part of #2239 for throttling the CLI operations

PR #2259, opened 2020-09-17
Fix broken docs build due to dependencies

derekpierre:
```
ERROR: Cannot install nucypher[docs] 3.0.0b0 and nucypher 3.0.0b0 because these package versions have conflicting dependencies.

The conflict is caused by:
    nucypher[docs] 3.0.0b0 depends on certifi==2020.4.5.1; extra == "docs"
    nucypher 3.0.0b0 depends on certifi==2020.6.20

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/user_guide/#fixing-conflicting-dependencies
```


tuxxy (approved):

KPrasch (approved):
:cowboy_hat_face: - Let it rip!

PR #2260, opened 2020-09-17
Equalize dependency versions with nucypher requirements.

derekpierre:
And another one.


KPrasch (approved):
In it goes :cowboy_hat_face:

PR #2264, opened 2020-09-17
Export ursula_availability_score prometheus metric

AlexMapley:
## Background
The prometheus exporter is great, it's seriously so rare to have an exporter built into a protocol.

I've been keeping track of an ursula's availability score by parsing logs, but it would be easier if this were just exported directly instead.

## Notes
I'm a little rusty at python at this point, and I couldn't get the [pre-commit hooks](https://github.com/nucypher/nucypher/blob/main/docs/source/guides/contribution_guide.rst#setup-commit--push-hooks) working so I skipped them with a `git push --no-verify`. I'm hoping the ci tests will cover everything.

@derekpierre guessing you know this code best as the previous author, so wondering your thoughts. 
@KPrasch too as we always talk about prometheus stuff


nucypher/utilities/prometheus/collector.py
```diff
@@ -121,6 +124,7 @@ def _collect_internal(self) -> None:
 
         self.metrics["learning_status"].state('running' if self.ursula._learning_task.running else 'stopped')
         self.metrics["known_nodes_gauge"].set(len(self.ursula.known_nodes))
+        self.metrics["availability_score_gauge"].set(len(self.ursula._availability_tracker.score))
```
AlexMapley, line 127: I'm pretty sure this is the correct reference according to https://github.com/nucypher/nucypher/blob/main/tests/acceptance/network/test_availability.py#L41:

```
assert ursula._availability_tracker.score == 10
```

Hoping a reviewer can verify that this will export the correct availability score.

nucypher/utilities/prometheus/collector.py
```diff
@@ -107,6 +107,9 @@ def initialize(self, metrics_prefix: str, registry: CollectorRegistry) -> None:
             "policies_held_gauge": Gauge(f'{metrics_prefix}_policies_held',
                                          'Policies held',
                                          registry=registry),
+           "availability_score_gauge": Gauge(f'{metrics_prefix}_availability_score',
```
AlexMapley, line 110: this should export as `usrula_availability_score`, just like how [3 lines above^](https://github.com/nucypher/nucypher/pull/2264/files#diff-1c60e77304346ca5724df601af97db7bL107) ends up getting exported as `ursula_policies_held` according to the [p2p-org project](https://github.com/p2p-org/nucypher-monitoring/blob/f1aed84231a4052808c492a6ff89de4f97732f94/grafana/provisioning/dashboards/nucypher.json#L1834)

KPrasch:
@AlexMapley - Thanks for your first time contribution - We'll take a look at this soon! :-)

nucypher/utilities/prometheus/collector.py
```diff
@@ -121,6 +124,7 @@ def _collect_internal(self) -> None:
 
         self.metrics["learning_status"].state('running' if self.ursula._learning_task.running else 'stopped')
         self.metrics["known_nodes_gauge"].set(len(self.ursula.known_nodes))
+        self.metrics["availability_score_gauge"].set(self.ursula._availability_tracker.score)
```
AlexMapley, line 127: I'm pretty sure this is the correct reference according to https://github.com/nucypher/nucypher/blob/main/tests/acceptance/network/test_availability.py#L41:

```
assert ursula._availability_tracker.score == 10
```

Hoping a reviewer can verify that this will export the correct availability score.

derekpierre:
@AlexMapley, nice idea. Thanks for the contribution! I'll try to take a look at it as soon as I can.

nucypher/utilities/prometheus/collector.py
```diff
@@ -107,6 +107,9 @@ def initialize(self, metrics_prefix: str, registry: CollectorRegistry) -> None:
             "policies_held_gauge": Gauge(f'{metrics_prefix}_policies_held',
                                          'Policies held',
                                          registry=registry),
+           "availability_score_gauge": Gauge(f'{metrics_prefix}_availability_score',
+                                         'Availability score',
+                                         registry=registry),
```
derekpierre, line 112: Text alignment could be improved here:

```suggestion
            "availability_score_gauge": Gauge(f'{metrics_prefix}_availability_score',
                                                                      'Availability score',
                                                                      registry=registry),
```

nucypher/utilities/prometheus/collector.py
derekpierre: Interesting situation here - availability tracker can be disabled (see https://github.com/nucypher/nucypher/blob/main/nucypher/characters/lawful.py#L1214) in which case it is never started and it simply reports the initial base score of `10`. In this case this code will report `10` as opposed to something indicating that it is disabled - which can lead to confusion.

It may be better to store something like `-1` if the `Gauge` allows for it, or store some other type of metric for the score that allows for text eg. `NA`, or `disabled` to indicate when the tracker is not being used.

KPrasch:
Looking good - but I second what @derekpierre mentioned about the availability tracker being disabled, which is now the default in `main`.

nucypher/utilities/prometheus/collector.py
AlexMapley: Good point, I was wondering about this. Currently running my ursulas with `--disable-availability-check`, so this scenario would apply for me.

Biased from my own use case, but I'd prefer `-1` as the output if the tracker is disabled - if only to work well with prometheus queries downstream that are expecting numbers.

AlexMapley: Updated with https://github.com/nucypher/nucypher/pull/2264/commits/894b69678276a0dbb78e8b0876edd0dcd86ce4dc

nucypher/utilities/prometheus/collector.py
```diff
@@ -121,6 +124,10 @@ def _collect_internal(self) -> None:
 
         self.metrics["learning_status"].state('running' if self.ursula._learning_task.running else 'stopped')
         self.metrics["known_nodes_gauge"].set(len(self.ursula.known_nodes))
+        if self.ursula._availability_check:
```
derekpierre, line 127: I don't believe that we can only use this check here. There are two determinants of whether the availability check is run:
1. The ursula configuration file (self.ursula._availability_check)
2. The `ursula run` CLI option (--availability-check / --disable-availability-check)

Therefore `self.ursula._availability_check` could be true, but the CLI option `--disable-availability-check` is specified, in which case the availability tracker is not run. These are the two values checked here - https://github.com/nucypher/nucypher/blob/main/nucypher/characters/lawful.py#L1214

I think you can instead use

```suggestion
        if self.ursula._availability_tracker.running:
```

@KPrasch interested in your thoughts here.

nucypher/utilities/prometheus/collector.py
```diff
@@ -107,6 +107,9 @@ def initialize(self, metrics_prefix: str, registry: CollectorRegistry) -> None:
             "policies_held_gauge": Gauge(f'{metrics_prefix}_policies_held',
                                          'Policies held',
                                          registry=registry),
+            "availability_score_gauge": Gauge(f'{metrics_prefix}_availability_score',
+                                                                      'Availability score',
```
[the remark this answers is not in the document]
AlexMapley: Sure thing will do, apologies for the back and forth

nucypher/utilities/prometheus/collector.py
```diff
@@ -99,14 +99,17 @@ def initialize(self, metrics_prefix: str, registry: CollectorRegistry) -> None:
...
+                                        'Number of accepted work orders',
+                                        registry=registry),
             "policies_held_gauge": Gauge(f'{metrics_prefix}_policies_held',
-                                         'Policies held',
-                                         registry=registry),
+                                        'Policies held',
+                                        registry=registry),
+            "availability_score_gauge": Gauge(f'{metrics_prefix}_availability_score',
+                                        'Availability score',
+                                        registry=registry),
```
AlexMapley, line 112: See how this block[ looks now](https://github.com/nucypher/nucypher/blob/b52e0a0d83d64e14cf9007df1e10f252b1951eec/nucypher/utilities/prometheus/collector.py#L101-L112)- there were a mix of tabs and spaces in here before

nucypher/utilities/prometheus/collector.py
AlexMapley: `self.ursula._availability_tracker.running:` does look like it would be the better source of truth here, at least looking at [this block](https://github.com/nucypher/nucypher/blob/76eea579a3b94e1ea2a8019601e0b454f81e5e2c/nucypher/cli/painting/nodes.py#L84-L88):

```
    if ursula._availability_tracker:
        if ursula._availability_tracker.running:
            score = 'Availability Score .. {} ({} responders)'.format(ursula._availability_tracker.score, len(ursula._availability_tracker.responders))
        else:
            score = 'Availability Score .. Disabled'
```

We may also want to use that same nested conditional:
```
if ursula._availability_tracker:
    if ursula._availability_tracker.running:
```
to prevent any TypeErrors

AlexMapley: Tried this out with https://github.com/nucypher/nucypher/pull/2264/commits/3fd1f8b267c4043e4c8c7d9ce015dffb1aca43ae

nucypher/utilities/prometheus/collector.py
```diff
@@ -121,6 +124,13 @@ def _collect_internal(self) -> None:
 
         self.metrics["learning_status"].state('running' if self.ursula._learning_task.running else 'stopped')
         self.metrics["known_nodes_gauge"].set(len(self.ursula.known_nodes))
+        if self.ursula._availability_tracker:
+            if self.ursula._availability_tracker.running:
```
AlexMapley, line 128: Used the same 
```
    if ursula._availability_tracker:
        if ursula._availability_tracker.running:
```
check from https://github.com/nucypher/nucypher/blob/76eea579a3b94e1ea2a8019601e0b454f81e5e2c/nucypher/cli/painting/nodes.py#L84-L85

nucypher/utilities/prometheus/collector.py
derekpierre: 💯

nucypher/utilities/prometheus/collector.py
```diff
@@ -121,6 +124,13 @@ def _collect_internal(self) -> None:
...
+                self.metrics["availability_score_gauge"].set(self.ursula._availability_tracker.score)
+            else:
+                self.metrics["availability_score_gauge"].set(-1)
+        else:
+            self.metrics["availability_score_gauge"].set(-1)
```
derekpierre, line 133: You can simplify this block.

```suggestion
        if self.ursula._availability_tracker and self.ursula._availability_tracker.running:
            self.metrics["availability_score_gauge"].set(self.ursula._availability_tracker.score)
        else:
            self.metrics["availability_score_gauge"].set(-1)
```

derekpierre (commented):
A couple minor comments:
- seems like `get-pip.py` was not supposed to be committed
- Just one minor RFC with the `if` statement

@AlexMapley - thanks for all of the hard work on this. Once the above issues are addressed I'll approve.

AlexMapley:
> seems like get-pip.py was not supposed to be committed
Oi sorry about that, I thought I removed that file this morning.

Appreciate all the commit suggestions by the way, super helpful!

derekpierre (approved):
🎸

nucypher/utilities/prometheus/collector.py
```diff
@@ -121,6 +124,10 @@ def _collect_internal(self) -> None:
 
         self.metrics["learning_status"].state('running' if self.ursula._learning_task.running else 'stopped')
         self.metrics["known_nodes_gauge"].set(len(self.ursula.known_nodes))
+        if self.ursula._availability_tracker and self.ursula._availability_tracker.running:
```
KPrasch, line 127: :+1:

KPrasch (approved):
:cowboy_hat_face:

AlexMapley:
First PR woooooooooooo

AlexMapley:
First PR in! Thanks for all the help :)

PR #2276, opened 2020-09-21
Make sure datastores are not retained in tests

### LINKED ISSUE #2172: Use lower lmdb memory allocation in tests to avoid failure

fjarri:
Fixes #2172

The problem is not caused by the DB size (see the comments in the issue), but persistent `Datastore` objects. Fixing that is the main purpose of the PR. Reduced DB size is not relevant - LMDB files are dynamic anyway (although one could argue that the reduced DB size still helps catch some errors).

What it does:
- changes `make_rest_app()` to not retain a reference to `Ursula` or the created datastore in the returned REST app object, **now it is the caller's responsibility;**
- adds a cleanup method to `Ursula`, to be used in fixtures. Weakrefs will make sure the datastore is cleaned up if all references to `Ursula` is released, but that may not be the case in tests, so an explicit method is needed.
- removes the DB size reduction fixture. It doesn't affect file sizes, and doesn't really test anything. 

I was not sure where to place the cleanup method. Currently it's separate; other possibilities are (overridden) `disenchant()`, or `stop()`. If either of these is allowed to invalidate the `Ursula`, we can put the cleanup there.

**TODO:**
* Any other places where Ursulas are created in tests and need to be cleaned up?

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -41,7 +41,12 @@ import nucypher
 from bytestring_splitter import BytestringKwargifier, BytestringSplitter, BytestringSplittingError, \
     VariableLengthBytestring
 from constant_sorrow import constants
-from constant_sorrow.constants import INCLUDED_IN_BYTESTRING, PUBLIC_ONLY, STRANGER_ALICE, UNKNOWN_VERSION, READY
+from constant_sorrow.constants import (INCLUDED_IN_BYTESTRING,
+                                       PUBLIC_ONLY,
+                                       STRANGER_ALICE,
+                                       UNKNOWN_VERSION,
+                                       READY,
+                                       INVALIDATED)
 from nucypher.acumen.nicknames import nickname_from_seed
 from nucypher.acumen.perception import FleetSensor
 from nucypher.blockchain.eth.actors import BlockchainPolicyAuthor, Worker
@@ -1278,6 +1283,20 @@ class Ursula(Teacher, Character, Worker):
                 self._arrangement_pruning_task.stop()
         if halt_reactor:
             reactor.stop()
+
+    def _finalize(self):
+        """
+        Cleans up Ursula from objects that may eat up system resources.
+        Useful for testing purposes, where many Ursulas are created and destroyed,
+        and references to them may persist for too long.
+        This method is not needed if all references to the Ursula are released.
+
+        **Warning:** invalidates the Ursula.
+        """
+
+        # `rest_server` holds references to the datastore (directly and via `rest_app`).
+        # An open datastore hogs up file descriptors.
+        self.rest_server = INVALIDATED
 
     def rest_information(self):
         hosting_power = self._crypto_power.power_ups(TLSHostingPower)
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -18,13 +18,14 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 import binascii
 import os
 import uuid
+import weakref
 from bytestring_splitter import BytestringSplitter
 from constant_sorrow import constants
 from constant_sorrow.constants import FLEET_STATES_MATCH, NO_BLOCKCHAIN_CONNECTION, NO_KNOWN_NODES
 from flask import Flask, Response, jsonify, request
 from hendrix.experience import crosstown_traffic
 from jinja2 import Template, TemplateError
-from typing import Tuple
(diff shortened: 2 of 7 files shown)


nucypher/network/server.py
```diff
@@ -84,13 +84,27 @@ def make_rest_app(
         log=Logger("http-application-layer")
         ) -> Tuple:
 
-    forgetful_node_storage = ForgetfulNodeStorage(federated_only=this_node.federated_only)
+    # A trampoline function for the real REST app,
+    # to ensure that a reference to the datastore object is not held by the closure.
 
     from nucypher.datastore import datastore
+    import weakref
```
fjarri, line 91: Note to self: move it to module imports when making other changes.

nucypher/network/server.py
```diff
@@ -84,12 +85,24 @@ def make_rest_app(
...
+    # to ensure that a reference to the node and the datastore object is not held by the app closure.
+    # One would think that it's enough to only remove a reference to the node,
+    # but `rest_app` somehow holds a reference to itself, Uroboros-like,
+    # and will hold the datastore reference if it is created there.
 
     from nucypher.datastore import datastore
 
     log.info("Starting datastore {}".format(db_filepath))
     datastore = datastore.Datastore(db_filepath)
+    rest_app = _make_rest_app(weakref.proxy(datastore), weakref.proxy(this_node), serving_domains, log)
```
cygnusv, line 98: To be fair, @jMyles complained when I introduced the change that passed a reference to `this_node` to the rest app. I didn't like how bloated was getting this method call, with >10 attributes from this node and growing, so I thought that maybe it was simpler to pass the node directly. In hindsight, this was contributing to the circular references problem.

cygnusv (approved):
I'm not well-versed on Python internals to completely understand this PR, but it seems very reasonable. I'm glad that @fjarri is looking at this with fresh eyes :)

nucypher/characters/lawful.py
```diff
@@ -1279,6 +1284,20 @@ def stop(self, halt_reactor: bool = False) -> None:
...
+        and references to them may persist for too long.
+        This method is not needed if all references to the Ursula are released.
+
+        **Warning:** invalidates the Ursula.
+        """
+
+        # `rest_server` holds references to the datastore (directly and via `rest_app`).
+        # An open datastore hogs up file descriptors.
+        self.rest_server = INVALIDATED
+
```
KPrasch, line 1300: :zap: Excellent - I might need a bit of clarification on the character's lifecycle at this point considering, `start`, `stop`, `enchant`, `disenchant`, and `finalize`.

nucypher/network/server.py
```diff
@@ -84,12 +85,24 @@ def make_rest_app(
         log=Logger("http-application-layer")
         ) -> Tuple:
 
-    forgetful_node_storage = ForgetfulNodeStorage(federated_only=this_node.federated_only)
+    # A trampoline function for the real REST app,
```
KPrasch, line 88: This is meant as a docstring? Perhaps include `"""`

nucypher/characters/lawful.py
fjarri: I would love that myself. There's even a note in `gwme` that for some reason `Ursula` has `stop` and all the others have `disenchant`. Ideally, I think there should be a non-invalidating `stop` (which could be followed by `start`), and an invalidating `finalize` (used primarily in tests to speed up the cleanup), the same for all characters. Or whichever other names as long as they're consistent.

KPrasch (approved):
Thank you very much for the deep dive here - I'm sure this will pay off for a long time to come.  Hopefully we can chase out all the circular references - @jMyles will surely be interested to see this.

nucypher/network/server.py
fjarri: Not really, it's not something a user of it should see. The only important part is that the returned `rest_app` doesn't hold a reference to `datastore`, but I would think that's the default. The docstring would be whatever the original `make_rest_app()` had, and it didn't have any. I can add one though.

fjarri: Added a docstring and some type annotations.

nucypher/characters/lawful.py
derekpierre: ^ @jMyles

KPrasch: Opened: https://github.com/nucypher/nucypher/issues/2293

KPrasch:
:boom:

PR #2304, opened 2020-09-28
Some contract updates

vzotova:
* deposits  from worklock allowed only after initialization
* `setSnapshots` method for `StakingInterface`
* check snapshots in `batchDeposit`
* allow `setWindDown` without deposit

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
@@ -556,7 +556,7 @@ contract StakingEscrow is Issuer, IERC900History {
     * If true then stake's duration will be decreasing in each period with `commitToNextPeriod()`
     * @param _windDown Value for parameter
     */
-    function setWindDown(bool _windDown) external onlyStaker {
+    function setWindDown(bool _windDown) external {
         StakerInfo storage info = stakerInfo[msg.sender];
         if (info.flags.bitSet(WIND_DOWN_INDEX) == _windDown) {
             return;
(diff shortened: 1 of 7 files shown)


nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -540,7 +540,7 @@ contract StakingEscrow is Issuer, IERC900History {
         uint256 _value,
         uint16 _periods
     )
-        external
+        external isInitialized
```
cygnusv, line 543: Simple and elegant :)

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -660,7 +660,7 @@ contract StakingEscrow is Issuer, IERC900History {
             uint256 endIndex = j + numberOfSubStakes;
             require(numberOfSubStakes > 0 && subStakesLength >= endIndex);
             StakerInfo storage info = stakerInfo[staker];
-            require(info.subStakes.length == 0);
+            require(info.subStakes.length == 0 && !info.flags.bitSet(SNAPSHOTS_DISABLED_INDEX));
```
cygnusv, line 663: What's the rationale for this change?

vzotova: Because we already have code that assumes that snapshots were enabled (in `batchDeposit`) and `setSnapsots` works for everybode (not only staker), I see that we need to protect one or another method

derekpierre (approved):
🎸

cygnusv (approved):

KPrasch (approved):

PR #2315, opened 2020-09-29
Confirmations and pre-flight for upgrade CLI

KPrasch:
Co-Author: @derekpierre 

- Support block confirmations for upgrade, re-targeting, rollback, etc.
- Display current contract version numbers with `inspect` command
- Eager "Pre-flight" checks for upgrade CLI 
- Deprecated `nucypher deploy transfer tokens`
- Deprecates automated deployment series test (needs further actor-level deprecation)

### DIFF
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -630,9 +630,8 @@ class BlockchainInterface:
                          fire_and_forget: bool = False  # do not wait for receipt.
                          ) -> dict:
 
-        # TODO: Are confirmations and fire_and_forget conflicting options?
-        # if fire_and_forget and confirmations > 0:
-        #     raise ValueError
+        if fire_and_forget and confirmations > 0:
+            raise ValueError('Transaction Prevented: Cannot use confirmations and fire_and_forget options together.')
 
         transaction = self.build_contract_transaction(contract_function=contract_function,
                                                       sender_address=sender_address,
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -627,8 +627,12 @@ class BlockchainInterface:
                          payload: dict = None,
                          transaction_gas_limit: int = None,
                          confirmations: int = 0,
-                         fire_and_forget: bool = False
+                         fire_and_forget: bool = False  # do not wait for receipt.
                          ) -> dict:
+
+        # TODO: Are confirmations and fire_and_forget conflicting options?
+        # if fire_and_forget and confirmations > 0:
+        #     raise ValueError
 
         transaction = self.build_contract_transaction(contract_function=contract_function,
                                                       sender_address=sender_address,
--- a/nucypher/blockchain/eth/registry.py
+++ b/nucypher/blockchain/eth/registry.py
@@ -152,7 +152,7 @@ class RegistrySourceManager:
     def get_primary_sources(cls):
         return [source for source in cls._FALLBACK_CHAIN if source.is_primary]
 
-    def fetch_latest_publication(self, registry_class, network: str = NetworksInventory.DEFAULT):  # TODO: see #1496
+    def fetch_latest_publication(self, registry_class, network: str):  # TODO: see #1496
         """
         Get the latest contract registry data available from a registry source chain.
         """
--- a/nucypher/blockchain/eth/registry.py
+++ b/nucypher/blockchain/eth/registry.py
@@ -152,7 +152,7 @@ class RegistrySourceManager:
     def get_primary_sources(cls):
         return [source for source in cls._FALLBACK_CHAIN if source.is_primary]
 
-    def fetch_latest_publication(self, registry_class, network: str):  # TODO: see #1496
+    def fetch_latest_publication(self, registry_class, network: str):
         """
         Get the latest contract registry data available from a registry source chain.
         """
--- a/nucypher/cli/commands/deploy.py
+++ b/nucypher/cli/commands/deploy.py
@@ -364,7 +364,8 @@ def upgrade(general_config, actor_options, retarget, target_address, ignore_depl
                                                                     target_address=target_address), abort=True)
         transaction = ADMINISTRATOR.retarget_proxy(contract_name=contract_name,
                                                    target_address=target_address,
-                                                   just_build_transaction=True)
+                                                   just_build_transaction=True,
+                                                   confirmations=confirmations)
 
         trustee_address = select_client_account(emitter=emitter,
                                                 prompt="Select trustee address",
--- a/nucypher/cli/commands/deploy.py
+++ b/nucypher/cli/commands/deploy.py
@@ -319,19 +319,30 @@ def inspect(general_config, provider_uri, config_root, registry_infile, deployer
 def upgrade(general_config, actor_options, retarget, target_address, ignore_deployed, multisig, confirmations):
     """Upgrade NuCypher existing proxy contract deployments."""
 
+    #
     # Setup
+    #
+
     emitter = general_config.emitter
     ADMINISTRATOR, deployer_address, blockchain, local_registry = actor_options.create_actor(emitter, is_multisig=bool(multisig))  # FIXME: Workaround for building MultiSig TXs | NRN
+
+    contract_name = actor_options.contract_name
+    if not contract_name:
+        raise click.BadArgumentUsage(message="--contract-name is required when using --upgrade")
+
+    github_registry = establish_deployer_registry(emitter=emitter, download_registry=True, network=actor_options.network)
+    try:
+        Deployer = ADMINISTRATOR.deployers[contract_name]
+    except KeyError:
+        emitter.echo(f'No such contract "{contract_name}"', color='red')
+        raise click.Abort()
+    deployer = Deployer(registry=local_registry)
 
     #
     # Pre-flight
     #
 
-    contract_name = actor_options.contract_name
-
     # Check deployer address is owner
-    Deployer = ADMINISTRATOR.deployers[contract_name]
-    deployer = Deployer(registry=local_registry)
     if Deployer._ownable and deployer_address != deployer.owner:  # blockchain read
         emitter.echo(DEPLOYER_IS_NOT_OWNER.format(deployer_address=deployer_address,
                                                   contract_name=contract_name,
@@ -341,14 +352,12 @@ def upgrade(general_config, actor_options, retarget, target_address, ignore_depl
         emitter.echo('✓ Verified deployer address as contract owner', color='green')
 
     # Check registry ID has changed locally compared to remote source
-    github_registry = establish_deployer_registry(emitter=emitter, download_registry=True, network=actor_options.network)
     if (github_registry.id == local_registry.id) and not actor_options.force:
         emitter.echo(IDENTICAL_REGISTRY_WARNING.format(github_registry=github_registry,
                                                        local_registry=local_registry), color='red')
         raise click.Abort()
     else:
         emitter.echo('✓ Verified local registry contains updates', color='green')
-
 
     #
     # Business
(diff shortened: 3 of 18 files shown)


nucypher/blockchain/eth/interfaces.py
```diff
@@ -627,9 +627,13 @@ def send_transaction(self,
                          payload: dict = None,
                          transaction_gas_limit: int = None,
                          confirmations: int = 0,
-                         fire_and_forget: bool = False
+                         fire_and_forget: bool = False  # do not wait for receipt.
                          ) -> dict:
 
+        # TODO: Are confirmations and fire_and_forget conflicting options?
```
fjarri, line 633: Or even, should we treat `confirmations == 0` as `fire_and_forget = True`?

KPrasch: Good though - `fire_and_forget` is meant for receipts, however.  Suppose you want to wait for a receipt, but not confirmation... or *is* a receipt an implicit single confirmation?

nucypher/cli/commands/deploy.py
```diff
@@ -364,14 +391,17 @@ def upgrade(general_config, actor_options, retarget, target_address, ignore_depl
             raise click.BadArgumentUsage(message="--target-address is required when using --retarget")
         if not actor_options.force:
             click.confirm(CONFIRM_RETARGET.format(contract_name=contract_name, target_address=target_address), abort=True)
-        receipt = ADMINISTRATOR.retarget_proxy(contract_name=contract_name,target_address=target_address)
+        receipt = ADMINISTRATOR.retarget_proxy(contract_name=contract_name,target_address=target_address, confirmations=0)
```
fjarri, line 394: Just checking - is `confirmations=0` intended here? Doesn't look like it was the default logic before.

KPrasch: Good eye - that one is an over-zealous paste, meant to be `confirmations`.

tests/acceptance/cli/test_deploy_commands.py
```diff
@@ -30,6 +34,14 @@
 ALTERNATE_REGISTRY_FILEPATH_2 = '/tmp/nucypher-test-registry-alternate-2.json'
 
 
+@pytest.fixture(autouse=True)
+def monkeypatch_confirmations(testerchain, monkeypatch):
+    def monkey_block_until_enough_confirmations(client, transaction_hash, timeout, confirmations):
```
fjarri, line 39: Does it actually block until enough confirmations? `confirmations` is unused

fjarri: Ah, my bad, it's supposed to ignore the confirmations

KPrasch: :+1: - Good to indicate for future reviewers though - there is another test `test_block_confirmations` that checks for that behavior, so @derekpierre and I built this as a mock.

nucypher/cli/commands/deploy.py
KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/interfaces.py
fjarri: Hm, I suppose there is a difference. Even with `confirmations == 0` the call is still waits until the transaction is included into a block. The question is, is there a point in waiting for that but not waiting for confirmations?

cygnusv: I think "confirmations" is a bit ambiguous; in our codebase it means how many blocks are on top of the block where you TX is included. So `fire_and_forget` and `confirmations > 0` are indeed incompatible, but that doesn't mean that `(confirmations == 0) == fire_and_forget`: you can wait for the receipt of your TX and as soon as you have one, continue, which is in fact the default behaviour in `nucypher` (i.e., `fire_and_forget=False and confirmations==0`.

To put it logically, `fire_and_forget` implies `confirmations == 0`.

nucypher/blockchain/eth/registry.py
```diff
@@ -152,7 +152,7 @@ def __getitem__(self, index):
     def get_primary_sources(cls):
         return [source for source in cls._FALLBACK_CHAIN if source.is_primary]
 
-    def fetch_latest_publication(self, registry_class, network: str = NetworksInventory.DEFAULT):  # TODO: see #1496
+    def fetch_latest_publication(self, registry_class, network: str):  # TODO: see #1496
```
cygnusv, line 155:
```suggestion
    def fetch_latest_publication(self, registry_class, network: str):
```

nucypher/cli/commands/deploy.py
```diff
@@ -313,21 +313,56 @@ def inspect(general_config, provider_uri, config_root, registry_infile, deployer
...
+    #
+
+    emitter = general_config.emitter
+    ADMINISTRATOR, deployer_address, blockchain, local_registry = actor_options.create_actor(emitter, is_multisig=bool(multisig))  # FIXME: Workaround for building MultiSig TXs | NRN
 
     contract_name = actor_options.contract_name
     if not contract_name:
         raise click.BadArgumentUsage(message="--contract-name is required when using --upgrade")
 
+    github_registry = establish_deployer_registry(emitter=emitter, download_registry=True, network=actor_options.network)
```
cygnusv, line 333: This can be problematic, as it overwrites the default registry, which we don't know if can have local changes or not. Maybe we can prompt and ask for confirmation before doing this?

nucypher/cli/commands/deploy.py
```diff
@@ -336,7 +371,8 @@ def upgrade(general_config, actor_options, retarget, target_address, ignore_depl
                                                                     target_address=target_address), abort=True)
         transaction = ADMINISTRATOR.retarget_proxy(contract_name=contract_name,
                                                    target_address=target_address,
-                                                   just_build_transaction=True)
+                                                   just_build_transaction=True,
+                                                   confirmations=confirmations)
```
cygnusv, line 375: The multisig workflow (which is not being used, to be fair) doesn't use confirmations, as it only builds a TX call data, so the changes in these 2 lines should be reverted.

nucypher/cli/commands/deploy.py
```diff
@@ -364,14 +400,25 @@ def upgrade(general_config, actor_options, retarget, target_address, ignore_depl
...
         paint_receipt_summary(emitter=emitter, receipt=receipt)
+
     else:
         if not actor_options.force:
+            # Check for human verification of versioned upgrade details
             click.confirm(CONFIRM_BEGIN_UPGRADE.format(contract_name=contract_name), abort=True)
-        receipts = ADMINISTRATOR.upgrade_contract(contract_name=contract_name, ignore_deployed=ignore_deployed)
+            if deployer._ownable:  # Only ownable + upgradeable contracts apply
+                old_agent = ContractAgency.get_agent(agent_class=deployer.agency, registry=github_registry)
+                new_agent = ContractAgency.get_agent(agent_class=deployer.agency, registry=local_registry)
```
cygnusv, line 414: Nice use of two agents with different registries! I think it's the first time I see this used that explicitly and I'm liking it a lot :)

nucypher/cli/commands/deploy.py
```diff
@@ -527,47 +574,6 @@ def allocations(general_config, actor_options, allocation_infile, gas):
...
-    click.confirm(confirmation, abort=True)
-
-    if allowance:
-        receipt = token_agent.approve_transfer(amount=value,
-                                               sender_address=deployer_address,
-                                               spender_address=target_address)
-    else:
-        receipt = token_agent.transfer(amount=value, sender_address=deployer_address, target_address=target_address)
-    paint_receipt_summary(emitter=emitter, receipt=receipt)
-
```
cygnusv, line 569: Why are we getting rid of this? There's very useful logic here, like for example creating approvals. I strongly think we should keep it.

nucypher/cli/painting/deployment.py
```diff
@@ -153,6 +153,7 @@ def paint_deployer_contract_inspection(emitter, registry, deployer_address) -> N
 
             proxy_payload = f"""
 {agent.contract_name} .... {bare_contract.address}
+    ~ Version ............ {bare_contract.version}
```
cygnusv, line 156: 💯

cygnusv (changes requested):

nucypher/cli/commands/deploy.py
vzotova: Looks good, but I think it's over complicated considering that we only want to get version of contracts, maybe change it to `registry.search`

nucypher/cli/commands/deploy.py
KPrasch: I see - I added the `confirmations` kwargs here because it's now required.  In order to exclude it here, Ill need to make it optional.

cygnusv: To be fair, it's just going to be ignored, so it's ok to leave it as is, then.

nucypher/cli/commands/deploy.py
KPrasch: I removed it to deal with a failing test after making `--contract-name` required, thinking that this was completely unused.  I ended up reverting that change and leaving it as optional for now.  

Happy to add it back if you think it's useful after all.

nucypher/cli/commands/deploy.py
KPrasch: I don't think it overrides the on-disk registry:

```python
    if download_registry:
        registry = InMemoryContractRegistry.from_latest_publication(network=network)
        emitter.message(PRODUCTION_REGISTRY_ADVISORY.format(source=registry.source))
        return registry
```

cygnusv: I see. Thanks fine then!

nucypher/cli/commands/deploy.py
KPrasch: @vzotova - Here's how that ends up looking:

```python
old_contract = github_registry.search(contract_name=contract_name)[-1]  # latest GH version
new_contract = local_registry.search(contract_name=contract_name)[-1]   # latest local version
```

KPrasch: :heavy_check_mark:

nucypher/blockchain/eth/interfaces.py
KPrasch: :heavy_check_mark: - Let me know if you think the added `ValueError` is enough.

nucypher/blockchain/eth/registry.py
KPrasch: :heavy_check_mark:

nucypher/cli/commands/deploy.py
KPrasch: :heavy_check_mark:

KPrasch:
This PR is passing locally, but failing on CI due to a timeout (most likely waiting for some human input).

I will keep adding tests and looking for the missing interaction, but can be merged at will with approvals.

nucypher/cli/commands/deploy.py
```diff
@@ -461,14 +462,14 @@ def contracts(general_config, actor_options, mode, activate, gas, ignore_deploye
     deployment_parameters = {}
     if parameters:
         with open(parameters) as json_file:
-            deployment_parameters = json.load(json_file)
+            deployment_parameters = json.load(json_file)  # TODO: Seems like this is bypassing existing flow in economics.py
```
cygnusv, line 465: This is meant to be used by contracts that have non-default constructor parameters, like the MultiSig (currently not used).

cygnusv (approved):
Awesome, let's test this manually to have greater assurance

tests/acceptance/cli/test_deploy_cli.py
```diff
@@ -45,74 +54,6 @@ def test_echo_solidity_version(click_runner):
     assert str(SOLIDITY_COMPILER_VERSION) in result.output, 'Solidity version text was not produced.'
 
 
-def test_nucypher_deploy_contracts(click_runner,
```
KPrasch, line 48: Note for reviewers: Here we are beginning the retirement + deprecation of the automated series deployment for *all* contracts.  Single contract deployment is the new recommended way.

nucypher/cli/commands/deploy.py
KPrasch: Ah - I see.

vzotova (approved):

PR #2319, opened 2020-09-30
[WIP] Bug fixes for upgrade CLI

KPrasch:


nucypher/cli/actions/confirm.py
```diff
@@ -115,3 +121,31 @@ def confirm_destroy_configuration(config: CharacterConfiguration) -> bool:
...
+                           local_registry: LocalContractRegistry,
+                           source_registry: InMemoryContractRegistry,
+                           deployer: Type[BaseContractDeployer],
+                           ) -> None:
+    """
+    Compares the versions of two 'implementation' contracts using a local and source registry.
+    """
+    old_contract: VersionedContract = blockchain.get_contract_by_name(
+        registry=source_registry,
+        contract_name=deployer.contract_name,
```
derekpierre, line 137: You pass `contract_name` as a parameter to this function - should we just remove it since it is obtained from the deployer?

derekpierre (approved):
🎸 - one outstanding RFC - looks good otherwise!

nucypher/cli/actions/confirm.py
KPrasch: Hmm - not sure I understand your question.

derekpierre: `contract_name` is a parameter, and so is `deployer`.

Within this function we use both `deployer.contract_name` (except on L149, which I assume can use `deployer.contract_name` as well). Can we simply use `deployer.contract_name` and no longer have `contract_name` as a parameter?

KPrasch: Oh, I see.  That implies a deeper change to an existing intermediate API. It's not always the case that a `deployer` is in context when calling `blockchain.get_contract_by_name.`

KPrasch: Oh - you meant for `verify_upgrade_details` - Got it.  Yes I'll make that change.

derekpierre: 👍

nucypher/blockchain/eth/utils.py
```diff
@@ -85,18 +85,19 @@ def estimate_block_number_for_period(period: int, seconds_per_period: int,  late
 def etherscan_url(item, network: str, is_token=False) -> str:
     if network is None or network is UNKNOWN_DEVELOPMENT_CHAIN_ID:
         raise ValueError("A network must be provided")
-    elif network == 'mainnet':
+
+    network = network.lower()
```
fjarri, line 89: Seems like a red flag that we have to do it for something coming from inside our own code and not from the user. `blockchain/eth/clients.py` already has a `# TODO: Consider creating a ChainInventory class and/or moving this to a separate module`, which would certainly help.

nucypher/cli/commands/deploy.py
```diff
@@ -397,6 +394,7 @@ def upgrade(general_config, actor_options, retarget, target_address, ignore_depl
         filepath = f'proposal-{trustee.multisig_agent.contract_address[:8]}-TX-{transaction_proposal.nonce}.json'
         transaction_proposal.write(filepath=filepath)
         emitter.echo(SUCCESSFUL_SAVE_MULTISIG_TX_PROPOSAL.format(filepath=filepath), color='blue', bold=True)
+        return  # Exit
```
fjarri, line 397: Is it just to emphasize the end of execution here? Doesn't seem like it's necessary.

derekpierre: Haha we had a conversation about this - it was @KPrasch personal preference for a clear end of execution.

fjarri: It may be a sign that (in Justin's voice) these branches want to be functions

cygnusv: That's a great @jMyles' impersonation.

Wrt to your comment, there's a tendency in the codebase to add some syntactic redundancy for the sake of clarity (e.g., redundant `returns`, wrapping boolean expressions with `bool(...)`). I personally don't think they hurt, so I'm completely OK with that.

nucypher/blockchain/eth/utils.py
cygnusv: You're totally right, let's do this :)

cygnusv: I included an additional commit that checks wrt to the chain names dictionary. We are still not addressing the TODO you referenced above, as we should try to keep the scope of this PR focused to the upgrades' logic.

cygnusv (approved):

Issue #2316, opened 2020-09-30
Manual deployment of idle network test timeout on CI


KPrasch:
https://app.circleci.com/pipelines/github/nucypher/nucypher/6048/workflows/0c616900-0c39-45dc-8c3e-460d745609d2/jobs/104915/parallel-runs/5?filterBy=ALL

introduced by PR #2315 - ~~I think some kind of human input is expected, but never received.~~

KPrasch:
I tracked down the issue on CI using pytest-timeout (see #2326) -- Resolved by creating a new test subdirectory for all the deploy CLI commands, with a common mocking fixture for confirmations.

PR #2326, opened 2020-10-01
Pytest-timeout configuration; Dependency relock

KPrasch:
https://github.com/pytest-dev/pytest-timeout

### DIFF
--- a/.circleci/pytest.ini
+++ b/.circleci/pytest.ini
@@ -1,3 +1,3 @@
 [pytest]
-addopts = -s -v --junitxml=./reports/pytest-results.xml --strict-markers --durations=0 --cov=nucypher --cov-report xml:reports/coverage.xml
+addopts = -s -v --junitxml=./reports/pytest-results.xml --strict-markers --durations=0 --cov=nucypher --cov-report xml:reports/coverage.xml --timeout 580
 junit_family = xunit1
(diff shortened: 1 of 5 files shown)


.circleci/pytest.ini
```diff
@@ -1,3 +1,3 @@
 [pytest]
-addopts = -s -v --junitxml=./reports/pytest-results.xml --strict-markers --durations=0 --cov=nucypher --cov-report xml:reports/coverage.xml
+addopts = -s -v --junitxml=./reports/pytest-results.xml --strict-markers --durations=0 --cov=nucypher --cov-report xml:reports/coverage.xml --timeout 580
```
fjarri, line 2: > `--timeout 580`

That's... very specific

KPrasch: Heh - I chose this number because it's 20 seconds less than the forced timeout on circleCI.

fjarri: Why do we need one more timeout then?

KPrasch: If we can catch the timeout with pytest-timout - it will cause the test to fail via `pytest.fail`, and dump the stack for inspection.   When circleCI issues a timeout, it's halts the docker container running the job leaving no clues as to why the timeout happened.  This is ultimately how I was able to discover the cause of issue #2316

fjarri: Ah, I see, thanks for the explanation.

fjarri (approved):

Issue #1754, opened 2020-02-27
Reconsider Embedded Emitters


KPrasch:
Currently, emitters are instantiated at the CLI level however sometimes context from deeper in the stack is needed.  For example: interactively confirming transactions just in time, before signing and broadcasting. 

Question: "Is it okay to use an embedded emitter?"
At the moment we just init emitters on-the-fly, However this can sometimes case issues where JSON RPC output is expected. (see <https://github.com/nucypher/nucypher/blob/master/nucypher/blockchain/eth/actors.py#L1276>).

Alternately, we can re-implement emitters as process local singletons.

fjarri:
I don't think at the moment having emitter singletons is justified. There are only two non-CLI emitters in the codebase:

1. In `Worker.block_until_ready()` - it reports the staking address and the ETH balance on creation.

Proposed solution: replace the emitter with logging, make the address and the balance accessible from the Ursula object, and emit them from CLI where it is created.

2. In `BlockchainInterface.sign_and_broadcast_transaction()` - it asks the user to confirm the transaction in case of a hardware wallet.

This is a little more tricky. It cannot really be moved outside the method, but using an emitter there (even if it is some kind of a singleton one) is not great either - imagine someone building a GUI around `nucypher`, would they have to capture stdout to intercept this message? I think a better way would be to have some kind of a `confirmation_callback` parameter which takes a callable with a transaction info parameter(s).

How often is this confirmation needed? Does it happen for any transaction, or just for some types? In the latter case we can just require a callback in the corresponding agent methods.

KPrasch:
After deliberation, usage, and reconsideration embedded emitters have become the norm.  If there is still emitter injection into the python API, it can be removed.

KPrasch:
Yes - I think so. We see 'em working everyday in worker consoles.

fjarri:
I mean, they didn't really become the norm. It would be easy to get rid of them, and that's what I would prefer.

KPrasch:
> Proposed solution: replace the emitter with logging, make the address and the balance accessible from the Ursula object, and emit them from CLI where it is created.

This isn't really possible for running ursulas without a significant reworking the abstraction layers.  Many of the messages there are the result of a task, consider the worker startup:

```
Authenticating Ursula
Awaiting worker qualification
Startup will resume when 0x... is funded and bonded to a staking account.
✓ Worker is bonded to (0x...)!
✓ Worker is funded with 0.05 ETH!
[...]

PR #2327, opened 2020-10-01
Log encoded constructor params

cygnusv:
Log encoded constructor params when deploying/upgrading contract, to facilitate source code validation of deployed contracts.


KPrasch (commented):
What do you think about displaying this (and all required etherscan verification info) information at the end of deploy/upgrade commands?

cygnusv:
I think that's possible, but the problem is that this happen in the BlockchainInterface layer, and outputting the message in the CLI is several layers up. So we will probably have to do again something like `BlockchainInterface.sign_and_broadcast_transaction()` does and obtain the emitter in a very ugly way (#1754). 

BTW, I think this can potentially be part of #2319. It's within the scope

KPrasch:
Yep - agreed.  I wonder if it's instead possible to simply pass this data up the call stack, instead of embedding an emitter.

nucypher/blockchain/eth/utils.py
```diff
@@ -85,18 +85,19 @@ def estimate_block_number_for_period(period: int, seconds_per_period: int,  late
 def etherscan_url(item, network: str, is_token=False) -> str:
     if network is None or network is UNKNOWN_DEVELOPMENT_CHAIN_ID:
         raise ValueError("A network must be provided")
-    elif network == 'mainnet':
+
+    network = network.lower()
+    if network == 'mainnet':
```
derekpierre, line 90: Should we use the `NetworksInventory.MAINNET` constant?

nucypher/blockchain/eth/interfaces.py
```diff
@@ -825,13 +826,18 @@ def deploy_contract(self,
                       f"deployer address {deployer_address} "
                       f"and parameters {pprint_args}")
 
-        transaction_function = contract_factory.constructor(*constructor_args, **constructor_kwargs)
+        constructor_function = contract_factory.constructor(*constructor_args, **constructor_kwargs)
+        constructor_calldata = encode_constructor_arguments(self.client.w3,
+                                                            constructor_function,
+                                                            *constructor_args,
+                                                            **constructor_kwargs)
+        self.log.info(f"Constructor calldata: {constructor_calldata}")
```
derekpierre, line 834: Should this string just replace the `pprint_args` one used on L827 (i.e. move that one here and replace `pprint_args` with `constructor_calldata`, or do we need both log entries?

nucypher/blockchain/eth/utils.py
```diff
@@ -85,18 +85,19 @@ def estimate_block_number_for_period(period: int, seconds_per_period: int,  late
...
-        network = network.lower()
         testnets_supported_by_etherscan = ('ropsten', 'goerli', 'rinkeby', 'kovan')
         if network in testnets_supported_by_etherscan:
             domain = f"https://{network}.etherscan.io"
         else:
             raise ValueError(f"'{network}' network not supported by Etherscan")
 
     if is_address(item):
-        item_type = 'address' if not is_token else 'token'
+        item_type = 'token' if is_token else 'address'
```
derekpierre, line 100: 👍

nucypher/blockchain/eth/utils.py
cygnusv: `NetworksInventory` refers to NuCypher networks, and it's just a coincidence that Ethereum and NuCypher's mainnets are both called "mainnet".

nucypher/blockchain/eth/interfaces.py
cygnusv: We need both entries because they're quite different:
```
2020-10-01T19:00:48+0200 [test-blockchain#info] Deploying contract StakingInterface:v1.6.1 with deployer address 0x7E5F4552091A69125d5DfCb7b8C2659029395Bdf and parameters 0x2946259E0334f33A064106302415aD3391BeD384, 0x51a240271AB8AB9f9a21C82d9a85396b704E164d, 0xA3183498b579bd228aa2B62101C40CC1da978F24, 0x7758F98C1c487E5653795470eEab6C4698bE541b
2020-10-01T19:00:48+0200 [test-blockchain#info] Constructor calldata: 0x0000000000000000000000002946259e0334f33a064106302415ad3391bed38400000000000000000000000051a240271ab8ab9f9a21c82d9a85396b704e164d000000000000000000000000a3183498b579bd228aa2b62101c40cc1da978f240000000000000000000000007758f98c1c487e5653795470eeab6c4698be541b
```

nucypher/blockchain/eth/utils.py
derekpierre: Ah! 😞

derekpierre (approved):
🎸

nucypher/utilities/ethereum.py
```diff
@@ -38,3 +41,20 @@ def get_array_data_location(array_location: int) -> int:
     # See https://solidity.readthedocs.io/en/latest/internals/layout_in_storage.html#mappings-and-dynamic-arrays
     data_location = Web3.toInt(Web3.keccak(to_bytes32(array_location)))
     return data_location
+
+
+def encode_constructor_arguments(web3: Web3,
+                                 constructor_function: ContractConstructor,
+                                 *constructor_args, **constructor_kwargs) -> HexStr:
```
KPrasch, line 48: Type hint here can Include `None`-type

cygnusv:
Included as part of #2319

Issue #2309, opened 2020-09-29
Incremental gas price strategy 


cygnusv:
Although we've discussed this in the past, @banteg eloquently put it in Discord yesterday:
> have you considered adopting pseudo escalator strategy as seen in pymaker to fight stuck txs and get overall cheaper txs?
> barebones example https://gist.github.com/banteg/b9a62c0f66ec9f37bed3a88c2025456a
> this is mostly ripped from here
> https://github.com/makerdao/pymaker/blob/master/pymaker/gas.py#L129
> https://github.com/makerdao/pymaker/blob/master/pymaker/__init__.py#L574
> 
> basically it allows supplying a gas_price(seconds_elapsed) func which e.g. could go linearly from slow to fast in 2 minutes and beyond. and the tx func broadcasts replacement txs every 10% of the bump, and returns when either any of the sent txs is mined or account nonce increases externally
> this method is very robust, i've sent tens of thousands txs this way
> the gas price func could also factor in an external gas oracle so it can react to rapid spikes in gas market
> if the things are the same as they were in cassandra, the nodes basically have the whole day to send these txs
> so it's possible to make a gas strategy that lowballs the gas price for like 20 hours and only then starts to escalate. but i think that's a bit of an overkill in fighting for cheapness
>
> another downside of using web3 gas strategies is that they use retroactive data
> if you recommend users to use geth, you can leverage its graphql mempool api. i've been using this as an oracle lately. obviously you need to tune the parameter to match your lowball glacier use case.
> https://gist.github.com/banteg/6413cdd71f6a6d6f1b63f7c1bb364b79
> i think it can go through few fallbacks 1. geth mempool 2. geth blocks (web3 strategies) 3. external oracle

cygnusv:
wontfix

Issue #2243, opened 2020-09-13
web3.exceptions.ValidationError: Constrained sample size is 0


cygnusv:
**Describe the Bug**
This error is from the web3 gas price strategies. These strategies try to get information from the latest blocks, but if there are no blocks (e.g., because it's not synced), they throw this very uninformative error: 
```
web3.exceptions.ValidationError: Constrained sample size is 0
```

**To Reproduce**
Run any nucypher command with a provider that's not synced. 

**Traceback or Screenshots (Optional)**
If applicable, add the full error message / Traceback or upload screenshots to help explain your problem.

**System (please complete the following information):**
 - OS Version:
 - Nucypher Version:
 - Ethereum Node Version:

cygnusv:
Closed in 32a8d92320579e7086d54b1c943ab9a8693dbd22

PR #2331, opened 2020-10-05
[WIP] Improvement and reorganization of gas strategies logic

cygnusv:
Current gas strategies logic is still very primitive and inflexible. This PR aims to improve this by:
* Promoting gas strategies logic to a separate module as a starting point
* Introducing a simple string representation for gas strategies to facilitate finer-grained configuration of the possible gas strategies. For example, we currently only allow to select the web3 gas strategies, but our codebase supports data feeds, which cannot be used since there's no way to select them from the configuration. The idea is to allow the operator to configure the gas strategy with a string like `<strategy_provider>/<strategy_option>`, e.g., `web3/fast`. This is still WIP and any ideas are welcome!
* Introduce more complex strategies, such as the "increasing" strategy outlined in #2309, or others involving fallback chains.

TODO:
* #2243 
* #2309

### DIFF
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -15,7 +15,6 @@ You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
-
 import os
 import pprint
 import time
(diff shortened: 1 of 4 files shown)


cygnusv:
Closed as stale.

PR #2339, opened 2020-10-07
Ensure that testnet flag is set correctly for deployer

### LINKED ISSUE #2338: Deployer uses case sensitive check when determining if running on testnet or not
Description copied from https://github.com/nucypher/nucypher/issues/2296#issuecomment-698434999:

For the determination of `testnet` here, https://github.com/nucypher/nucypher/blob/main/nucypher/cli/commands/deploy.py#L208, 
default network name conversion with `mainnet` causes a mismatch based on capitalization.

We compare the chain name obtained from the chain id 
<blockchain/eth/clients.py>
```python
PUBLIC_CHAINS = {
                                 1: "Mainnet",
```

With the network name 
<blockchain/eth/networks.py>
```python
MAINNET = 'mainnet'
```

Related to changes from #2220 .

derekpierre:
Fixes #2338. Ensure proper testnet check is performed and added a test.

### DIFF
--- a/nucypher/cli/commands/deploy.py
+++ b/nucypher/cli/commands/deploy.py
@@ -16,15 +16,16 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
 import json
+import os
 from typing import Tuple
 
 import click
-import os
 from constant_sorrow import constants
 from constant_sorrow.constants import FULL
 
 from nucypher.blockchain.eth.actors import ContractAdministrator, Trustee
 from nucypher.blockchain.eth.agents import ContractAgency, MultiSigAgent
+from nucypher.blockchain.eth.clients import PUBLIC_CHAINS
 from nucypher.blockchain.eth.constants import STAKING_ESCROW_CONTRACT_NAME
 from nucypher.blockchain.eth.interfaces import BlockchainInterface
 from nucypher.blockchain.eth.networks import NetworksInventory
@@ -207,7 +208,7 @@ class ActorOptions:
                 password = get_client_password(checksum_address=deployer_address)
 
         # Produce Actor
-        testnet = deployer_interface.client.chain_name != NetworksInventory.MAINNET
+        testnet = deployer_interface.client.chain_name != PUBLIC_CHAINS[1]  # Mainnet
         signer = Signer.from_signer_uri(self.signer_uri, testnet=testnet) if self.signer_uri else None
         ADMINISTRATOR = ContractAdministrator(registry=local_registry,
                                               client_password=password,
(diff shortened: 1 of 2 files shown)


nucypher/cli/commands/deploy.py
```diff
@@ -207,7 +208,7 @@ def create_actor(self,
                 password = get_client_password(checksum_address=deployer_address)
 
         # Produce Actor
-        testnet = deployer_interface.client.chain_name != NetworksInventory.MAINNET
+        testnet = deployer_interface.client.chain_name != PUBLIC_CHAINS[1]  # Mainnet
```
KPrasch, line 211: :+1: - Yes I think this will do the trick.

tests/acceptance/cli/deploy/test_deploy_cli.py
```diff
@@ -61,6 +63,32 @@ def test_deploy_single_contract(click_runner, tempfile_path):
...
+        # fail trying to deploy contract to testnet since ETH blanace is 0, signer will already have been initialized
+        result = click_runner.invoke(deploy, command, input=user_input, catch_exceptions=False)
+        assert result.exit_code != 0, result.output  # expected failure given eth balance is 0
+        spy_from_signer_uri.assert_called_with(TEST_PROVIDER_URI, testnet=True)
+
+        # fail trying to deploy contract to mainnet (:-o) since ETH balance is 0, signer will already have been initialized
+        with patch('nucypher.blockchain.eth.clients.EthereumTesterClient.chain_name', PropertyMock(return_value='Mainnet')):
+            result = click_runner.invoke(deploy, command, input=user_input, catch_exceptions=False)
+            assert result.exit_code != 0, result.output  # expected failure given invalid contract name
+            spy_from_signer_uri.assert_called_with(TEST_PROVIDER_URI, testnet=False)  # the "real" deal
```
KPrasch, line 89: :joy:

nucypher/cli/commands/deploy.py
KPrasch: Retracing your thinking here - this makes the `PUBLIC_CHAINS` list case sensitive, right?

derekpierre: Instead of caring about case sensitivity, now you are comparing apples to apples - the chain_name returns a value from the PUBLIC_CHAINS dict, so we just compare it to the mainnet value in the same dict.

KPrasch: Ah, right :100:

KPrasch (approved):

nucypher/cli/commands/deploy.py
cygnusv: 🍎2🍎

cygnusv (approved):

PR #2343, opened 2020-10-08
Allow --from-unlocked flag to specify when only uncollected staking rewards and unlocked substakes are used for creating/increasing a stake.

### LINKED ISSUE #2273: Clarify staking liquid vs. escrow-unlocked NU.
Current interactive prompt when staking via CLI:

```
Transfer tokens from the staker address? Otherwise, unlocked tokens in the escrow will be used [True]
```

derekpierre:
Fixes #2273 .

### DIFF
--- a/nucypher/cli/commands/stake.py
+++ b/nucypher/cli/commands/stake.py
@@ -24,7 +24,7 @@ from nucypher.blockchain.eth.events import EventRecord
 from nucypher.blockchain.eth.interfaces import BlockchainInterfaceFactory
 from nucypher.blockchain.eth.registry import IndividualAllocationRegistry
 from nucypher.blockchain.eth.signers.software import ClefSigner
-from nucypher.blockchain.eth.token import NU, StakeList, Stake
+from nucypher.blockchain.eth.token import NU, Stake
 from nucypher.blockchain.eth.utils import datetime_at_period
 from nucypher.cli.actions.auth import get_client_password
 from nucypher.cli.actions.configure import get_or_update_configuration, handle_missing_configuration_file
@@ -36,7 +36,6 @@ from nucypher.cli.actions.confirm import (
     confirm_staged_stake, confirm_disable_snapshots
 )
 from nucypher.cli.actions.select import select_client_account_for_staking, select_stake
-from nucypher.cli.utils import setup_emitter
 from nucypher.cli.config import group_general_config
 from nucypher.cli.literature import (
     BONDING_DETAILS,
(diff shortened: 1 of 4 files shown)


nucypher/cli/commands/stake.py
```diff
@@ -451,8 +454,8 @@ def unbond_worker(general_config, transacting_staker_options, config_file, force
 @option_value
 @option_lock_periods
 @group_general_config
-@option_only_lock
-def create(general_config, transacting_staker_options, config_file, force, value, lock_periods, only_lock):
+@option_use_uncollected_rewards
```
KPrasch, line 457: :+1:  How about "unclaimed" throughout; `--from-unclaimed-rewards`

tests/integration/cli/test_stake_cli_functionality.py
```diff
@@ -977,10 +978,11 @@ def test_create_lock_interactive(click_runner,
 
     command = ('create',
                '--provider', MOCK_PROVIDER_URI,
-               '--network', TEMPORARY_DOMAIN)
+               '--network', TEMPORARY_DOMAIN,
+               '--use-rewards')
 
     user_input = '\n'.join((str(selected_index),
-                            NO,
+                            YES,
```
KPrasch, line 985: Haha - this is a funny--to-read diff line :joy:

nucypher/cli/commands/stake.py
derekpierre: 🤔 Interestingly, I had `unclaimed` before, but I realized that we only use "claim" for worklock in our code - "collect" was used in the context of rewards.

I'm fine to change it if desired, but was trying to keep the parlance the same.

derekpierre: Using "from" instead of "use" is good

derekpierre: ✔️

nucypher/cli/commands/stake.py
```diff
@@ -102,13 +101,17 @@
...
+from nucypher.cli.utils import setup_emitter
 from nucypher.config.characters import StakeHolderConfiguration
 
 option_value = click.option('--value', help="Token value of stake", type=click.INT)
 option_lock_periods = click.option('--lock-periods', help="Duration of stake in periods.", type=click.INT)
 option_worker_address = click.option('--worker-address', help="Address to bond as an Ursula-Worker", type=EIP55_CHECKSUM_ADDRESS)
 option_index = click.option('--index', help="The staker-specific stake index to edit", type=click.INT)
-option_only_lock = click.option('--only-lock', help="Locking tokens without deposit", is_flag=True)
+option_from_uncollected_rewards = click.option('--from-uncollected-rewards',
+                                               help="Only use uncollected staking rewards, and not tokens from staker address",
```
vzotova, line 112: I guess this is fine, but if to be precisely - unlocked tokens could be not only reward but also unlocked sub-stake

KPrasch: Yes, this is truly just "unlocked and escrowed" NU.  It's hard to describe both clearly and tersely at the same time... :thinking:

derekpierre: Something like `--from-alternative-source` (or if you like a slightly more descriptive name, `--from-unlocked-or-rewards`) and then use the help text to clarify? I think it'll be hard to cover the full picture in the option name.

derekpierre: Used `from-unlocked` - ✔️

vzotova (approved):

KPrasch (approved):
:cowboy_hat_face: Thank You!

PR #2347, opened 2020-10-10
Nickname refactoring & changing nickname symbols to more common ones

fjarri:
*Edit:* the scope of this PR has been limited to get done with public-visible changes first. Can't change the branch name, but it still kinda works.

* use a `Nickname` class to encapsulate nickname data instead of passing around tuples of tuples
* use NATO phonetic alphabet instead of Unicode glyphs for nicknames (to avoid problems with glyph representation in different fonts)

To get a few more bits in the nickname, we can add Æ and Þ (since they have convenient words to match with in English). 

**For reviewers:** please pay specific attention to correct serialization/deserialization. I'm not 100% sure I looked through all the instances where it happens.

### DIFF
--- a/nucypher/acumen/nicknames.py
+++ b/nucypher/acumen/nicknames.py
@@ -17,24 +17,22 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 
 
 import json
+from typing import List
 import random
 from os.path import abspath, dirname, join
 
 import unicodedata
 
-HERE = BASE_DIR = abspath(dirname(__file__))
-with open(join(HERE, 'web_colors.json')) as f:
-    colors = json.load(f)
+_HERE = abspath(dirname(__file__))
+with open(join(_HERE, 'web_colors.json')) as f:
+    _COLORS = json.load(f)['colors']
 
-colors = colors['colors']
-pairs = []
-
-symbols_tuple = ("♈", "♉", "♊", "♋", "♌", "♍", "♎", "♏", "♐", "♑", "♒", "♓",
-                 "♚", "♛", "♜", "♝", "♞", "♟", "⚓", "⚔", "⚖", "⚗", "⚑", "⚘",
-                 "⚪", "⚵", "⚿", "⛇", "⛈", "⛰", "⛸", "⛴", "⛨", "✈", "☤",
-                 "⏚", "☠", "☸", "☿", "☾", "♁", "♃", "♄", "☄", "☘", "⚜", "⚚",
-                 "⏲", "☣", "☥", "♣", "♥", "♦", "♠", "♫", "🟒", "⚛", "⚙", "⎈",
-                 "☮", "☕", "☈", "♯", "♭")
+_SYMBOLS = ("♈", "♉", "♊", "♋", "♌", "♍", "♎", "♏", "♐", "♑", "♒", "♓",
+           "♚", "♛", "♜", "♝", "♞", "♟", "⚓", "⚔", "⚖", "⚗", "⚑", "⚘",
+           "⚪", "⚵", "⚿", "⛇", "⛈", "⛰", "⛸", "⛴", "⛨", "✈", "☤",
+           "⏚", "☠", "☸", "☿", "☾", "♁", "♃", "♄", "☄", "☘", "⚜", "⚚",
+           "⏲", "☣", "☥", "♣", "♥", "♦", "♠", "♫", "🟒", "⚛", "⚙", "⎈",
+           "☮", "☕", "☈", "♯", "♭")
 
 
 def nicename(symbol):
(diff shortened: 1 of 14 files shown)


nucypher/acumen/nicknames.py
```diff
@@ -45,18 +44,45 @@ def nicename(symbol):
...
+class Nickname:
+
+    @classmethod
+    def from_seed(cls, seed, length=2):
+        # TODO: #1823 - Workaround for new nickname every restart
+        # if not seed:
+        #     raise ValueError("No checksum provided to derive nickname.")
+        rng = random.Random(seed)
+        nickname_symbols = rng.sample(symbols, length)
+        nickname_colors = rng.sample(colors, length)
```
fjarri, line 73: I don't think there's actually any harm in doing choice with replacement, but currently just keeping the old behavior.

nucypher/acumen/nicknames.py
```diff
@@ -22,41 +22,91 @@
...
-    pairs = []
-    for pair in range(number_of_pairs):
-        color = random.choice(colors)
-        symbol = random.choice(symbols)
-        symbols.remove(symbol)
-        pairs.append((color, symbol))
-    nickname = " ".join(("{} {}".format(c['color'], nicename(s)) for c, s in pairs))
-    return nickname, pairs
+
+symbols = {
```
cygnusv, line 31: Shouldn't this be `SYMBOLS`?

cygnusv (commented):
At first sight, this looks great. The first commit alone is a drastic improvement on how nicknames are handled in the codebase. I still have to look at the PR a bit more.

For the record, this is how status page would look like with the change:
image

nucypher/acumen/nicknames.py
jMyles: Will replacement make cross-language implementations simpler?  If so, let's do it.

nucypher/acumen/nicknames.py
jMyles: Or `_SYMBOLS`, even.

jMyles (approved):
Lovely.

nucypher/acumen/nicknames.py
fjarri: Replacement here is in the sense that the drawn element is replaced in the list. That is, the symbols/colors can repeat.

nucypher/acumen/nicknames.py
fjarri: I suppose so, and the other constants as well.

fjarri: Fixed.

tests/acceptance/characters/test_ursula_web_status.py
```diff
@@ -41,9 +41,11 @@ def test_ursula_html_renders(ursula, client):
     assert response.status_code == 404
     response = client.get('/status/')
     assert response.status_code == 200
+    with open("status.html", "wb") as f:
+        f.write(response.data)
```
fjarri, line 45: My bad, testing stuff. Will remove that.

cygnusv: BTW, I checked that using the `federated_ursulas` fixture instead of `blockchain_ursulas` works equally well for this test, but it's waaaay faster.

cygnusv (approved):
Apart from my comment regarding the fixture, this is good to go. Great work @fjarri !

tests/acceptance/characters/test_ursula_web_status.py
fjarri: Good idea! Still requires an explicit call to `learn_from_teacher_node()` to propagate fleet states, but yeah, works much faster. I don't want to over-complicate this PR, so I'll put this change into the one I'm building on top of it.

fjarri: Fixed.

fjarri:
@tuxxy , were you actually serious about F for Flag and S for Star? If you were, now's the time to put in this change.

cygnusv:
Add `Kilo -> King` to that list :P

nucypher/acumen/nicknames.py
fjarri: BTW, this is also a kind of choice we'd better make in this PR. Are repeating symbols ok? I don't see any problem myself, but it wasn't the case in the original code; so maybe there was some reason.

fjarri:
For the record, I'm really -0 on any deviations from NPA. All new words would actually be better than replacing one or two - it's like we decided to write "A" upside down because we like it more this way. Also makes us marginally less culpable if someone manages to compose something offensive out of these words.

But I'm not going to argue too much.

nucypher/acumen/nicknames.py
jMyles: I don't recall the reason I opted to prohibit repetition.  There may not be a good one.

I also think that, if we are going to use this, it makes sense to move away from `random.Random(seed)` and to a more cross-language-friendly method of determinism.

fjarri: Good point. Implementing a simple linear congruential generator would work, we don't really need much from randomness here.

derekpierre (approved):
🎸

nucypher/acumen/nicknames.py
```diff
@@ -17,46 +17,94 @@
...
+
+    def __str__(self):
+        return self._text
+
+
+class Nickname:
+
+    @classmethod
+    def from_seed(cls, seed, length: int = 2):
+        # TODO: #1823 - Workaround for new nickname every restart
```
KPrasch, line 90: As in - federated mode?

nucypher/acumen/nicknames.py
```diff
@@ -17,46 +17,94 @@
...
+        # if not seed:
+        #     raise ValueError("No checksum provided to derive nickname.")
+        rng = random.Random(seed)
+        nickname_symbols = rng.sample(list(_SYMBOLS), length)
+        nickname_colors = rng.sample(_COLORS, length)
+        characters = [
+            NicknameCharacter(symbol, color['color'], color['hex'])
```
KPrasch, line 97:
```suggestion
            cls(symbol, color_name=color['color'], color_hexcolor=['hex'])
```

KPrasch (approved):

PR #2351, opened 2020-10-13
Even nicer status page

fjarri:
* [x] use Mako instead of Jinja as a template engine. Advantages:
  * uses Python syntax
  * easier template composition (can define several template functions in the same file)
  * better error reporting 

  Disadvantages:
  * doesn't allow inline conditionals (e.g. in `{%if something %}text{%endif}` in Jinja)
* [x] update status page design

Current design state:
Screen Shot 2020-10-13 at 15 57 33


vepkenez:
In my opinion switching from Jinja to Mako is a mistake and the limited logic capability of Jinja templates is a feature.
Now we have Python utility methods in a file that ends in `.mako`.  I don't feel like this is an improvement.

Also, this change results in a status page that is functionally identical to the old one.
What are we actually trying to solve here?

vepkenez:
I should also say... looks much better!

fjarri:
> In my opinion switching from Jinja to Mako is a mistake and the limited logic capability of Jinja templates is a feature.
Now we have Python utility methods in a file that ends in .mako. I don't feel like this is an improvement.

I understand that argument. Yes, there is a compulsion to put complicated logic into a template if it's allowed. But most of the logic in the current `basic_status.mako` consists of template functions, which you cannot really move out of a template. In Jinja the only difference would be that you'd have 5 files instead of 1, and I'm not even sure how it would work out - Jinja uses template inheritance instead of function calls. You can argue that I could move `hex_to_rgb()` and `contrast_color()` out of the template, but everything else belongs there.

> This is adding on yet another syntax for html rendering.

Jinja is being removed, isn't it? And it's exactly Jinja that adds its own syntax. Mako just uses Python's.

> Also, this change results in a status page that is functionally identical to the old one.

This page is functionally identical to that of Stackoverflow:
Screen Shot 2020-10-13 at 16 58 04
Would you use it?

> What are we actually trying to solve here?

User experience. The current page looks like the node crashed halfway through rendering it. If I were a user, I could run that for fun, but I wouldn't trust it with my money.

fjarri:
To add to that: if you make a typo in the template in this PR, that's what you'll get:

Screen Shot 2020-10-13 at 17 16 18

Current version doesn't even check for errors, but even with `StrictUndefined`, all you get is a generic error. I couldn't find a way to format it in Jinja.

nucypher/utilities/templates/cloud_deploy_ansible_inventory.mako
```diff
@@ -0,0 +1,51 @@
+all:
```
KPrasch, line 1: Heh - Mako all day

KPrasch (approved):
:cowboy_hat_face:

vepkenez:
The CSS changes in this PR would accomplish 99% of the goals.
I think we have other things to do that are important right now rather than change out our HTML templating language.

vepkenez (approved):

jMyles:
Although I green-flagged the transition to mako in conversation with @fjarri, I also echo @vepkenez's concerns about business logic creeping into templates.

jMyles (approved):

fjarri:
I'm glad I at least sold @vepkenez on the redesign. 

> I think we have other things to do that are important right now rather than change out our HTML templating language.

It's not a massive change, and it only affects one page. 

> but I don't really agree that it's good idea, at least right now, and I think as a matter of process, it would be nice if we could discuss these sorts of decisions.

Well, isn't that what we're doing now? I can roll back Mako if it causes such animosity, but I honestly think using it makes things much easier.

> Also, while yes this PR does improve the status page, does it improve enough to justify the effort?

Yes, I believe 3 hours I spent on one of the main UI elements of our codebase that will be looked at by hundreds of people, are more than justifiable. Or do you mean the effort of replacing Jinja with Mako? If anything, it saved me time.

> I also echo @vepkenez's concerns about business logic creeping into templates.

`basic_status.mako` has exactly two pieces of it that can be moved out: `hex_to_rgb()` (4 lines) and `contrast_color()` (6 lines), both very specific to this template (and `rgb_to_hex()` can actually be removed). I can move them out, that's no big deal. The rest are template functions that belong there.

As a change of viewpoint, consider a `mako` template not as a helper object, but simply an "inverted" Python module where strings are first-class entities that don't require escaping, and everything else does. All the rules normal for Python modules apply in `mako` templates as well. The only disadvantage now is that `coverage`, `mypy` or other tools do not cover it.

TLDR: you want Mako out? Just say the word.

KPrasch:
Although imperfect, this is still a step in the right direction IMO.  Let's not show too much logic into the template, and consider how to best serve a real-time dashboard in the future,

Issue #1995, opened 2020-05-15
Use "canonical address" instead of checksum address to key known_nodes


jMyles:
Even though we no longer fully instantiate nodes before using them (allowing them to remain a `NodeSprout` until we are about to verify), we still key all known nodes using their checksum address.

Thus, any time many more nodes are added to `known_nodes` (which happens essentially on every run at some point) is thus a substantially expensive CPU operation compared to everything that happens around it.  We can decrease this cost by keying using the raw bytes (known in web3 terms as "canonical address") instead.

jMyles:
Profiling info from `block_until_number_of_known_nodes_is` when run in `test_alice_can_learn_about_a_whole_bunch_of_ursulas`.  Shows that this is more than a fifth of the profile time of node learning (and is certainly exacerbated in a debugger).

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    2.775    2.775 nodes.py:577(block_until_number_of_known_nodes_is)
        1    0.008    0.008    2.674    2.674 nodes.py:741(learn_from_teacher_node)
        1    0.006    0.006    1.830    1.830 lawful.py:1531(batch_from_bytes)
    15009    0.024    0.000    1.428    0.000 __init__.py:143(__call__)
    15009    1.077    0.000    1.125    0.000 __init__.py:178(actually_split)
        1    0.013    0.013    1.062    1.062 __init__.py:337(repeat)
     5003    0.008    0.000    0.730    0.000 lawful.py:1464(from_bytes)
     5003    0.025    0.000    0.722    0.000 lawful.py:1446(internal_splitter)
     5003    0.039    0.000    0.709    0.000 nodes.py:360(remember_node)
    30000    0.019    0.000    0.634    0.000 nodes.py:105(checksum_address)
     5003    0.023    0.000    0.581    0.000 address.py:110(to_checksum_address)
```

KPrasch:
@fjarri @jMyles - What's the latest understanding on the optimization targets here?  Shall we move this Issue to the next milestone?

PR #2352, opened 2020-10-13
Fleet sensor refactor

fjarri:
The goal of this PR is to make interactions with `FleetSensor` consistent and avoid exposing implementation details, or carrying around ad-hoc states (`fleet_state_*` attributes). More specifically:

* `FleetState` is made into a real class. It encapsulates incremental updates and collection-like interface to iterate over the state's nodes. It is exposed from `FleetSensor` as `current_state`. 
* `FleetSensor` itself manages the buffers with new/deleted nodes and a list of previous states.
* Previous states are kept as simple `ArchivedFleetState` instances that do not retain references to actual nodes (since we don't use them anyway), only checksum/nickname/population. 
* The user (meaning the higher levels of the codebase) must call `record_fleet_state()` in order for the state to get updated. Currently there are multiple cases of accessing the node list without an explicit update.
* Fleet states for remote nodes are kept in `FleetSensor.remote_states` instead of in `fleet_state_*` attributes.
* `abridged_*` methods are renamed to better represent their purpose. The former `abridged_node_details()` (now `status_info()` is now used for both JSON and HTML `/status` endpoints outputs.

Rough edges and possible improvements:

* This PR may include a fix for #1995, if we decide on the course of actions.
* It is possible to avoid calculating the new state's checksum in some cases. Since remote nodes are immutable, if we retain the local node's metadata and remote nodes' addresses in archived states, we can compare those with the current ones, and use the archived state's checksum if there's a match. Not sure if there will be a noticeable benefit.
* The usage of the current state can be made more explicit if we don't forward most of the calls from `FleetSensor` to `FleetSensor.current_state`. 
* An alternative to the explicit `record_fleet_state()` call is to automatically update the state (if there are new nodes) whenever `current_state` is accessed. 
* In case of matching fleet states, the teacher sends `FLEET_STATES_MATCH` *along with* the fleet state. That seems excessive, but changing that without a proper protocol versioning will be a problem.

**Note:** if this PR is merged, https://github.com/nucypher/nucypher-monitor/blob/master/monitor/crawler.py will need to be updated.

### DIFF
--- a/nucypher/acumen/nicknames.py
+++ b/nucypher/acumen/nicknames.py
@@ -74,7 +74,7 @@ class NicknameCharacter:
         self.color_hex = color_hex
         self._text = color_name + " " + _SYMBOLS[symbol]
 
-    def payload(self):
+    def to_json(self):
         return dict(symbol=self.symbol,
                     color_name=self.color_name,
                     color_hex=self.color_hex)
@@ -103,8 +103,9 @@ class Nickname:
         self.icon = "[" + "".join(character.symbol for character in characters) + "]"
         self.characters = characters
 
-    def payload(self):
-        return [character.payload() for character in self.characters]
+    def to_json(self):
+        return dict(text=self._text,
+                    characters=[character.to_json() for character in self.characters])
 
     def __str__(self):
         return self._text
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -16,7 +16,9 @@
 """
 
 import binascii
+import itertools
 import random
+import weakref
 
 import maya
 
@@ -30,45 +32,137 @@ from nucypher.crypto.api import keccak_digest
 from nucypher.utilities.logging import Logger
 
 
-class FleetSensor:
-    """
-    A representation of a fleet of NuCypher nodes.
-    """
-    _checksum = NO_KNOWN_NODES.bool_value(False)
-    _nickname = NO_KNOWN_NODES
-    _tracking = False
-    most_recent_node_change = NO_KNOWN_NODES
-    snapshot_splitter = BytestringSplitter(32, 4)
-    log = Logger("Learning")
-    FleetState = namedtuple("FleetState", ("nickname", "icon", "nodes", "updated", "checksum"))
+class BaseFleetState:
 
-    def __init__(self, domain: str):
-        self.domain = domain
-        self.additional_nodes_to_track = []
-        self.updated = maya.now()
-        self._nodes = OrderedDict()
-        self._marked = defaultdict(list)  # Beginning of bucketing.
-        self.states = OrderedDict()
-
-    def __setitem__(self, checksum_address, node_or_sprout):
-        if node_or_sprout.domain == self.domain:
-            self._nodes[checksum_address] = node_or_sprout
-
-            if self._tracking:
-                self.log.info("Updating fleet state after saving node {}".format(node_or_sprout))
-                self.record_fleet_state()
+    def __str__(self):
+        if len(self) != 0:
+            # TODO: draw the icon in color, similarly to the web version?
+            return '{checksum} ⇀{nickname}↽ {icon}'.format(icon=self.nickname.icon,
+                                                           nickname=self.nickname,
+                                                           checksum=self.checksum[:7])
         else:
-            msg = f"Rejected node {node_or_sprout} because its domain is '{node_or_sprout.domain}' but we're only tracking '{self.domain}'"
-            self.log.warn(msg)
+            return 'No Known Nodes'
+
+
+class ArchivedFleetState(BaseFleetState):
+
+    def __init__(self, checksum, nickname, timestamp, population):
+        self.checksum = checksum
+        self.nickname = nickname
+        self.timestamp = timestamp
+        self.population = population
+
+    def to_json(self):
+        return dict(checksum=self.checksum,
+                    nickname=self.nickname.to_json() if self.nickname is not None else 'NO_KNOWN_NODES',
+                    timestamp=self.timestamp.rfc2822(),
+                    population=self.population)
+
+
+# Assumptions we're based on:
+# - Every supplied node object, after its constructor has finished,
+#   has a ``.checksum_address`` and ``bytes()`` (metadata)
+# - checksum address or metadata does not change for the same Python object
+# - ``this_node`` (the owner of FleetSensor) may not have a checksum address initially
+#   (when the constructor is first called), but will have one at the time of the first
+#   `record_fleet_state()` call. This applies to its metadata as well.
+# - The metadata of ``this_node`` **can** change.
+# - For the purposes of the fleet state, nodes with different metadata are considered different,
+#   even if they have the same checksum address.
+
+class FleetState(BaseFleetState):
+
+    @classmethod
+    def new(cls, this_node=None):
+        this_node_ref = weakref.ref(this_node) if this_node is not None else None
+        # Using empty checksum so that JSON library is not confused.
+        # Plus, we do need some checksum anyway. It's a legitimate state after all.
+        return cls(checksum=keccak_digest(b"").hex(),
+                   nodes={},
+                   this_node_ref=this_node_ref,
+                   this_node_metadata=None)
+
+    def __init__(self, checksum, nodes, this_node_ref, this_node_metadata):
+        self.checksum = checksum
+        self.nickname = None if checksum is None else Nickname.from_seed(checksum, length=1)
+        self._nodes = nodes
+        self.timestamp = maya.now()
+        self._this_node_ref = this_node_ref
+        self._this_node_metadata = this_node_metadata
+
+    def archived(self):
+        return ArchivedFleetState(checksum=self.checksum,
+                                  nickname=self.nickname,
+                                  timestamp=self.timestamp,
+                                  population=self.population)
+
+    def with_updated_nodes(self, new_nodes, marked_nodes):
+
+        # Checking if the node already has a checksum address
+        # (it may be created later during the constructor)
+        # or if it mutated since the last check.
+        if self._this_node_ref is not None and getattr(self._this_node_ref(), 'finished_initializing', False):
+            this_node = self._this_node_ref()
+            this_node_metadata = bytes(this_node)
+            this_node_changed = self._this_node_metadata != this_node_metadata
+            this_node_list = [this_node]
+        else:
+            this_node_metadata = self._this_node_metadata
+            this_node_changed = False
+            this_node_list = []
+
+        new_nodes = {checksum_address: node for checksum_address, node in new_nodes.items()
+                     if checksum_address not in marked_nodes}
+
+        remote_nodes_updated = any(
+            (checksum_address not in self._nodes or bytes(self._nodes[checksum_address]) != bytes(node))
+                and checksum_address not in marked_nodes
+            for checksum_address, node in new_nodes.items())
+
+        remote_nodes_slashed = any(checksum_address in self._nodes for checksum_address in marked_nodes)
+
+        if this_node_changed or remote_nodes_updated or remote_nodes_slashed:
+            # TODO: if nodes were kept in a Merkle tree,
+            # we'd have to only recalculate log(N) checksums.
+            # Is it worth it?
+            nodes = dict(self._nodes)
+            nodes.update(new_nodes)
+            for checksum_address in marked_nodes:
+                if checksum_address in nodes:
+                    del nodes[checksum_address]
+
+            all_nodes_sorted = sorted(itertools.chain(this_node_list, nodes.values()),
+                                      key=lambda node: node.checksum_address)
+            joined_metadata = b"".join(bytes(node) for node in all_nodes_sorted)
+            checksum = keccak_digest(joined_metadata).hex()
+        else:
+            nodes = self._nodes
+            checksum = self.checksum
+
+        return FleetState(checksum=checksum,
+                          nodes=nodes,
+                          this_node_ref=self._this_node_ref,
+                          this_node_metadata=this_node_metadata)
+
+    @property
+    def population(self):
+        """Returns the number of all known nodes, including itself, if applicable."""
+        return len(self) + int(self._this_node_metadata is not None)
 
     def __getitem__(self, checksum_address):
         return self._nodes[checksum_address]
 
+    def addresses(self):
+        return self._nodes.keys()
+
     def __bool__(self):
-        return bool(self._nodes)
+        return len(self) != 0
 
     def __contains__(self, item):
-        return item in self._nodes.keys() or item in self._nodes.values()
+        if isinstance(item, str):
+            return item in self._nodes
+        else:
+            return item.checksum_address in self._nodes
 
     def __iter__(self):
         yield from self._nodes.values()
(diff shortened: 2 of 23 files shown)


nucypher/acumen/perception.py
```diff
@@ -30,141 +33,324 @@
...
+                                  nickname=self.nickname,
+                                  timestamp=self.timestamp,
+                                  population=self.population)
+
+    def with_updated_nodes(self, new_nodes, marked_nodes):
+
+        # Checking if the node already has a checksum address
+        # (it may be created later during the constructor)
+        # or if it mutated since the last check.
+        if self._this_node_ref is not None and getattr(self._this_node_ref(), 'finished_initializing', False):
```
jMyles, line 103: While I like the way this logic is evolving, these weakref gymnastics aren't really doing it for me.  Is this all just to avoid having to lug around `additional_nodes_to_track`?

fjarri: Not really. There are several issues being solved here.

1. Circular references. The current code with `additional_nodes_to_track` has them too: `Ursula` -> `known_nodes` -> `additional_nodes_to_track` -> `Ursula`. Hence the weakrefs.

2. This code can be called at `Ursula` construction time, when its full metadata is still not available. A check for the local node to be available (currently a very awkward `getattr` of `finished_initializing`) removes the need in an additional mutating call that you have to remember to invoke sometime after you create the `Ursula`.

3. The local node can be mutated anytime. So we have to request the new metadata every time the state is updated.

jMyles: None of these are strong enough to justify the enormous hit in readability, IMO.

`weakref` is something best saved for when it's truly needed, like when it's the only way out of a hairy performance bottleneck with an object whose representation is too large (in memory) to justify doing some other way.

fjarri: > None of these are strong enough to justify the enormous hit in readability, IMO.

I wouldn't call two dereferences of a weakref an "enormous hit in readability". And the weakref is there only for dealing with the first point; if you get rid of it, the code will remain pretty much the same because there are still points 2 and 3 to worry about. Now if we prohibit node mutation, *that* will improve the readability.

> weakref is something best saved for when it's truly needed, like when it's the only way out of a hairy performance bottleneck with an object whose representation is too large (in memory) to justify doing some other way.

`Ursula` is the main object being held by the cycle here, and it's pretty large. Of course, it mainly matters for testing, but we've already hit a similar problem not so long ago. 

Personally, I believe that it is better to take a slight readability hit and use a weakref than debug a problem caused by a reference cycle half a year later.

nucypher/acumen/perception.py
```diff
@@ -30,141 +33,324 @@
...
+#   (when the constructor is first called), but will have one at the time of the first
+#   `record_fleet_state()` call. This applies to its metadata as well.
+# - The metadata of ``this_node`` **can** change.
+# - For the purposes of the fleet state, nodes with different metadata are considered different,
+#   even if they have the same checksum address.
+
+class FleetState(BaseFleetState):
+
+    @classmethod
+    def new(cls, this_node=None, nodes=[]):
```
KPrasch, line 75: mutable default

nucypher/network/nodes.py
```diff
@@ -763,7 +764,7 @@ def learn_from_teacher_node(self, eager=False, canceller=None):
 
         current_teacher = self.current_teacher_node()  # Will raise if there's no available teacher.
 
-        if Teacher in self.__class__.__bases__:
+        if Teacher in self.__class__.__bases__: # TODO: why not `isinstance`?
```
KPrasch, line 767: :+1:

nucypher/network/nodes.py
```diff
@@ -1347,30 +1322,40 @@ def node_details(node):
...
                    "staker_address": node.checksum_address,
                    "timestamp": node.timestamp.iso8601(),
                    "last_seen": last_seen,
-                   "fleet_state": node.fleet_state_checksum or 'unknown',
-                   "fleet_state_icon": fleet_icon,
                    "domain": node.serving_domain,
                    'version': nucypher.__version__}
         return payload
 
+    def _known_nodes_details(self) -> dict:
```
KPrasch, line 1336: General commentary - Let;s document where these metadata / description methods are being used, for an example: If a method is used for HTTP rendering, CLI, etc.  Consider reducing the number of these descriptor methods where possible.

nucypher/network/server.py
```diff
@@ -171,7 +171,7 @@ def all_known_nodes():
             # TODO: Is this every something we don't want to do?
             this_node.start_learning_loop()
 
-        if this_node.known_nodes.checksum is NO_KNOWN_NODES:
+        if len(this_node.known_nodes) == 0:
```
KPrasch, line 174:
``` suggestion
        if not this_node.known_nodes):
```

nucypher/network/server.py
```diff
@@ -427,14 +427,13 @@ def status():
 
         else:
             headers = {"Content-Type": "text/html", "charset": "utf-8"}
-            previous_states = list(reversed(this_node.known_nodes.states.values()))[:5]
+            previous_states = reversed(this_node.known_nodes.latest_states(5))
```
KPrasch, line 430: :clap:

KPrasch (commented):
Excellent readability improvements throughout - +1 on `ArchivedStates` as part of the fleet state lifecycle.

nucypher/acumen/perception.py
fjarri: I don't understand why it gets such a bed rep. Arguments with default values should never be mutated anyway, and arguments in general are rarely mutated (ideally), so an incorrect usage is easy to spot. And you get a more readable signature (and the type).

But in this case this parameter is actually unused, so I'll just remove it.

nucypher/network/nodes.py
fjarri: Fixed to use `isinstance()`

nucypher/acumen/perception.py
KPrasch: Consider this example which I think demonstrates the implicit nature of unintentional mutation:

```python
def mutable_default(nodes=[]):
    nodes.append(1)
    return nodes

x = mutable_default()
print(x)
[1]

y = mutable_default()
print(y)
[1, 1]

```

fjarri: I understand how it can backfire. But `nodes.append()` when `nodes` is an argument already calls for increased attention, and when you see that `nodes` has a default (whatever that default is), it's an instant RFC. 

Adding attributes to classes outside `__init__` causes problems much more often, and they are much harder to spot, yet nobody adds `__slots__` to everything.

KPrasch: > nodes.append() when nodes is an argument already calls for increased attention

Passing a mutable default suggests that the original intent is the create a new object, not reuse one.  Otherwise - why pass it in al all?

> Adding attributes to classes outside __init__ causes problems much more often, and they are much harder to spot, yet nobody adds __slots__ to everything.

This has an explicit failure mode, `AttributeError` which is easier to spot than implicit mutation.

Plus, I'm far from the only one who thinks it's a more safe, intuitive, natural convention.

fjarri: > Passing a mutable default suggests that the original intent is the create a new object, not reuse one. Otherwise - why pass it in at all?

Everything in Python is mutable, except for some simple types. `constant_sorrow` constants, for example, which we use as defaults in many places. That's really no different than using `[]`.

As for why pass it - basically 1) to simplify the signature (the `[]` default needs no explanation as opposed to `None`) 2) to simplify the type signature, 3) to avoid the `if x is None: x = []` constructs.

> This has an explicit failure mode, AttributeError which is easier to spot than implicit mutation.

Only if you have a custom `__setattr__`, or `__slots__`. 

> Plus, I'm far from the only one who thinks it's a more safe, intuitive, natural convention.

As I said, I never really understood it. Doesn't seem intuitive or natural to me. There is some degree of unsafety, but I think it does not outweigh the simpler code, especially in a codebase that's careful about mutation. Although I do wish Python had an immutable dict-like and set-like literals; `[]` can be replaced by `()` at least.

nucypher/network/server.py
fjarri: Fixed.

nucypher/network/nodes.py
fjarri: Now this info is used both for JSON and HTML status responses.

nucypher/acumen/perception.py
```diff
@@ -30,146 +33,323 @@
...
-                # FIXME: generalize in case we want to extend the number of symbols in the state nickname
-                "symbol": state.nickname.characters[0].symbol,
-                "color_hex": state.nickname.characters[0].color_hex,
-                "color_name": state.nickname.characters[0].color_name,
-                "updated": state.updated.rfc2822(),
-                }
+    def unpack_snapshot(data):
+        return FleetState.unpack_snapshot(data)
+
+    def record_fleet_state(self):
```
derekpierre, line 330: There will be some consequences to the status monitor code which will need to be updated - https://github.com/nucypher/nucypher-monitor/blob/master/monitor/crawler.py#L251. Not a huge deal, just something to note (@KPrasch )

nucypher/acumen/nicknames.py
```diff
@@ -74,7 +74,7 @@ def __init__(self, symbol: str, color_name: str, color_hex: str):
         self.color_hex = color_hex
         self._text = color_name + " " + _SYMBOLS[symbol]
 
-    def payload(self):
+    def to_json(self):
```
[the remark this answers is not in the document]
fjarri: I admit, I do not like this entirely. But there's already `Card.to_json()` that does the same, and it's not just a dict - it's a dict that can be serialized into JSON, which is a small subset of all dicts. I'm open to suggestions. `to_json_serializable`? `json_payload`?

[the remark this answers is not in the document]
derekpierre: Perhaps something like `to_serializable_dict()`; more general than just JSON serialization, or `serializable_payload()`...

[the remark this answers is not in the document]
KPrasch: Changes to the Card API seem out if scope in this PR, especially considering there are open issues that can be solved along with these in a separate changeset.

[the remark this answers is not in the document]
derekpierre: Yep, my recommendation was for this code change, and not for the `Card.to_json` - it can be renamed in a follow-up. I think Bogdan only brought it up as an analogous example somewhere else in the codebase.

KPrasch:
How many reviewers would you like on this PR?

nucypher/acumen/nicknames.py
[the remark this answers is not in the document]
fjarri: > I think Bogdan only brought it up as an analogous example somewhere else in the codebase.

Rather, I wanted to reduce the amount of differently-named serialization methods, and looked for an existing one that did what I needed.

I would personally prefer to either keep `to_json()`, and possibly rename it everywhere in another PR, or rename all currently existing occurrences of `to_json()` (including the `Card` one). Just to keep the codebase consistent at each point. 

As for the name, I don't particularly like `to_serializable_dict()`, because it's not necessarily a dict - can be anything serializable to JSON (a list, a string etc).

nucypher/acumen/perception.py
```diff
@@ -30,146 +33,323 @@
...
+        yield from self._current_state
 
     def __len__(self):
-        return len(self._nodes)
+        return len(self._current_state)
 
+    # FIXME: is it ever used?
     def __eq__(self, other):
-        return self._nodes == other._nodes
+        raise Exception
```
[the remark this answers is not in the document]
fjarri: Yes, just wanted to see if it's triggered in tests anywhere. Hard to find usages of `__eq__` directly.

nucypher/acumen/perception.py
```diff
@@ -30,146 +33,323 @@
 from nucypher.utilities.logging import Logger
 
 
+class BaseFleetState:
+
```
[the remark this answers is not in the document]
fjarri: There's a somewhat different behavior for `nickname` in `FleetState` and `ArchivedFleetState`, and increasing complexity just because of `checksum` didn't seem to worth it. But I agree, currently `BaseFleetState` is more of a mixin adding `__str__` behavior.

nucypher/acumen/perception.py
fjarri: Thanks for noticing that, it'll have to be updated.

nucypher/acumen/nicknames.py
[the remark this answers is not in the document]
derekpierre: > because it's not necessarily a dict 

I suspect that this was the reason for the original use of `payload()`. You could use what you mentioned before `to_json_serializable()`, it is just that if it is JSON serializable, isn't it just python serializable anyway eg. with `pickle`? Or am I mistaken? In which case, it might as well just be `to_serializable()`.

In any case, I'm good with making all the changes as part of a single and separate PR.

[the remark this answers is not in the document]
fjarri: > it is just that if it is JSON serializable, isn't it just python serializable anyway eg. with pickle? 

It is. Pretty much anything in Python is pickle-serializable (so it's not even worth an explicit specification), but JSON is much more restrictive. It would be nice to have some term like "simple hierarchical data" that would encompass, say, JSON and YAML, but I'm not aware of any such thing.

nucypher/acumen/perception.py
```diff
@@ -30,146 +33,297 @@
...
+# Assumptions we're based on:
+# - Every supplied node object, after its constructor has finished,
+#   has a ``.checksum_address`` and ``bytes()`` (metadata)
+# - checksum address or metadata does not change for the same Python object
+# - ``this_node`` (the owner of FleetSensor) may not have a checksum address initially
+#   (when the constructor is first called), but will have one at the time of the first
+#   `record_fleet_state()` call. This applies to its metadata as well.
+# - The metadata of ``this_node`` **can** change.
+# - For the purposes of the fleet state, nodes with different metadata are considered different,
+#   even if they have the same checksum address.
```
cygnusv, line 72: Thanks for this paragraph 🙏

nucypher/network/nodes.py
```diff
@@ -749,7 +751,7 @@ def learn_from_teacher_node(self, eager=False, canceller=None):
 
         if not self.done_seeding:
             try:
-                remembered_seednodes = self.load_seednodes(record_fleet_state=False)
+                remembered_seednodes = self.load_seednodes(record_fleet_state=True)
```
cygnusv, line 753: What's the implication of changing here this from False to True?

nucypher/network/nodes.py
```diff
@@ -1310,75 +1291,64 @@ def timestamp_bytes(self):
...
+
+        info['domain'] = self.domain
+        info['version'] = nucypher.__version__
+
+        latest_states = self.known_nodes.latest_states(5)
+
+        info['fleet_state'] = latest_states[-1].to_json()
+        info['previous_fleet_states'] = [state.to_json() for state in latest_states[:-1]]
+        info['known_nodes'] = [node._status_info_remote(self.known_nodes, raise_invalid=raise_invalid)
+                               for node in self.known_nodes.values()]
```
cygnusv, line 1352: I know this wasn't introduced by this PR, but how about adding a parameter here so that you can get the status info with/without the known_nodes data? That would be particularly useful if we also enable such parameters in the `/status` endpoint. @vepkenez and I were working on some potential crawling logic that checks nodes status information and the known_nodes is just too much data, which in most cases it's not needed.

cygnusv (approved):
Great work @fjarri, as usual!

nucypher/network/nodes.py
fjarri: Previously nodes remembered in `known_nodes` were accessible right away after they were recorded. This PR makes it more strict - only those nodes that make up the current fleet state are accessible, and the newly recorded nodes are in temporary storage. So in order for the seednodes to be visible we must record the fleet state.

nucypher/network/nodes.py
fjarri: Added an `omit_known_nodes` optional parameter - not sure about the name though.

nucypher/acumen/nicknames.py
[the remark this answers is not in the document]
fjarri: Filed #2540

nucypher/acumen/perception.py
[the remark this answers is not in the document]
fjarri: Fixed.

tests/integration/learning/test_domains.py
```diff
@@ -114,7 +114,7 @@ def test_learner_ignores_stored_nodes_from_other_domains(lonely_ursula_maker, tm
 
     # Once pest made its way into learner, learner taught passed it to other mainnet nodes.
 
-    learner.known_nodes._nodes[pest.checksum_address] = pest  # This used to happen anyway.
+    learner.known_nodes.record_node(pest)  # This used to happen anyway.
```
KPrasch, line 117: Excellent - something similar is unfolding in #2513

nucypher/acumen/perception.py
jMyles: Related discusson: https://ptb.discord.com/channels/411401661714792449/411401661714792451/772182575237431306

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
-        return nodes_we_know_about
+    def values(self):
+        return self._current_state.values()
+
+    def latest_states(self, quantity):
+        """
+        Returns at most ``quantity`` latest archived states (including the current one),
+        in chronological order.
+        """
+        return self._archived_states[-min(len(self._archived_states), quantity):]
```
KPrasch, line 314: Consider using a named variable here so that I have a place to trigger my debugger 😄

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
+        # TODO: for now we're not using `label` in any way, so we're just ignoring it
+        self._marked.add(node.checksum_address)
+
+    def record_remote_fleet_state(self, checksum_address, state_checksum, timestamp, population):
+        # TODO: really we can just create the timestamp here
 
-        if self._nodes.get(node):
-            del self._nodes[node]
+        nickname = Nickname.from_seed(state_checksum, length=1) # TODO: create in a single place
+        self.remote_states[checksum_address] = ArchivedFleetState(state_checksum, nickname, timestamp, population)
```
KPrasch, line 351: Use keyword arguments here?

nucypher/network/server.py
```diff
@@ -422,30 +422,28 @@ def receive_treasure_map():
 
     @rest_app.route('/status/', methods=['GET'])
     def status():
-        if request.args.get('json'):
-            payload = this_node.abridged_node_details(raise_invalid=False)
-            response = jsonify(payload)
-            return response
+
+        return_json = request.args.get('json') == 'true'
```
KPrasch, line 426: Why check for `true` explicitly here - to avoid the case of evaluating `false` as True?  I find this to be a bit awkward, perhaps we need to consider using an alternate endpoint?

tests/integration/learning/test_fleet_state.py
```diff
@@ -87,41 +98,51 @@ def test_state_is_recorded_after_learning(federated_ursulas, lonely_ursula_maker
...
 
+    states = some_ursula_in_the_fleet.known_nodes._archived_states
 
-    teacher_states_before = list(some_ursula_in_the_fleet.known_nodes.states.values())
+    states_before = len(states)
     lonely_learner.learn_from_teacher_node()
-    teacher_states_after = list(some_ursula_in_the_fleet.known_nodes.states.values())
+    states_after = len(states)
+
+    # FIXME: some kind of a timeout is required here to wait for the learning to end
```
KPrasch, line 138: I'm sensing that `FIXME` is overused in the code base.  Perhaps change this to a `TODO` and open an Issue to track.

tests/unit/test_external_ip_utilities.py
```diff
@@ -32,8 +32,15 @@
 from tests.constants import MOCK_IP_ADDRESS
 
 
+MOCK_NETWORK = 'holodeck'
```
KPrasch, line 35: 🖖🏻

KPrasch (commented):
Thanks @fjarri, left a few comments for ya!

nucypher/acumen/perception.py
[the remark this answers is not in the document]
fjarri: Removed this completely. `ArchivedFleetState` doesn't seem to need `__str__` anyway.

nucypher/acumen/perception.py
fjarri: Fixed.

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
+            return '{checksum} ⇀{nickname}↽ {icon}'.format(icon=self.nickname.icon,
+                                                           nickname=self.nickname,
+                                                           checksum=self.checksum[:7])
+        else:
+            return 'No Known Nodes'
+
+
+class ArchivedFleetState(BaseFleetState):
+
+    def __init__(self, checksum, nickname, timestamp, population):
```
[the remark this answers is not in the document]
fjarri: Fixed.

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
+    def new(cls, this_node=None):
+        this_node_ref = weakref.ref(this_node) if this_node is not None else None
+        # Using empty checksum so that JSON library is not confused.
+        # Plus, we do need some checksum anyway. It's a legitimate state after all.
+        return cls(checksum=keccak_digest(b"").hex(),
+                   nodes={},
+                   this_node_ref=this_node_ref,
+                   this_node_metadata=None)
+
+    def __init__(self, checksum, nodes, this_node_ref, this_node_metadata):
```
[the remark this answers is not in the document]
fjarri: Fixed.

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
+        self.checksum = checksum
+        self.nickname = None if checksum is None else Nickname.from_seed(checksum, length=1)
```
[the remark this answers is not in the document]
fjarri: I opted to remove this condition completely - `nickname` is actually never `None` here.

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
+    # But it's hard to change now because older nodes will be looking for it.
+    def snapshot(self):
+        checksum_bytes = binascii.unhexlify(self.checksum)
+        timestamp_bytes = self.timestamp.epoch.to_bytes(4, byteorder="big")
+        return checksum_bytes + timestamp_bytes
+
+    snapshot_splitter = BytestringSplitter(32, 4)
+
+    @staticmethod
+    def unpack_snapshot(data):
```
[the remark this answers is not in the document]
fjarri: Fixed.

nucypher/acumen/perception.py
fjarri: Fixed, but can't you set a breakpoint on a `return`?

nucypher/network/server.py
fjarri: Isn't that the correct way to do it? As far as I understand, URL arguments are given to the request object as strings, casting them is the user's responsibility.

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
+
+        new_nodes = {checksum_address: node for checksum_address, node in new_nodes.items()
+                     if checksum_address not in marked_nodes}
+
+        remote_nodes_updated = any(
+            (checksum_address not in self._nodes or bytes(self._nodes[checksum_address]) != bytes(node))
+                and checksum_address not in marked_nodes
+            for checksum_address, node in new_nodes.items())
+
+        remote_nodes_slashed = any(checksum_address in self._nodes for checksum_address in marked_nodes)
```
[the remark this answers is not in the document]
fjarri: Changed to `nodes_to_remove`, let's have higher levels worry about slashing.

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
+            this_node_list = [this_node]
+        else:
+            this_node_metadata = self._this_node_metadata
+            this_node_changed = False
+            this_node_list = []
+
+        new_nodes = {checksum_address: node for checksum_address, node in new_nodes.items()
+                     if checksum_address not in marked_nodes}
+
+        remote_nodes_updated = any(
```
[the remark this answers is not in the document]
fjarri: Moved to `_remote_nodes_updated()`

nucypher/acumen/perception.py
```diff
@@ -30,146 +32,320 @@
...
+            (checksum_address not in self._nodes or bytes(self._nodes[checksum_address]) != bytes(node))
+                and checksum_address not in marked_nodes
+            for checksum_address, node in new_nodes.items())
+
+        remote_nodes_slashed = any(checksum_address in self._nodes for checksum_address in marked_nodes)
+
+        if this_node_changed or remote_nodes_updated or remote_nodes_slashed:
+            # TODO: if nodes were kept in a Merkle tree,
+            # we'd have to only recalculate log(N) checksums.
+            # Is it worth it?
```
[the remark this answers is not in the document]
fjarri: I was thinking something along the lines of a balanced binary tree, where 

    checksum(tree_node) = checksum(
        checksum(tree_node.ursula_metadata), 
        checksum(tree_node.branch1), 
        checksum(tree_node.branch2))

So each time a node's metadata changes we only need to recalculate checksums along the path in the tree leading to that node. Of course when nodes are added/removed we'll need to rebalance the tree and recalculate some of the checksums...

nucypher/acumen/perception.py
fjarri: Fixed.

tests/integration/learning/test_fleet_state.py
fjarri: Filed #2568

nucypher/network/server.py
KPrasch: Yeah, alright, fair enough.

nucypher/acumen/perception.py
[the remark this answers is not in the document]
KPrasch: This is a very interesting suggestion, perhaps worth moving off this PR for further discussion.

nucypher/acumen/perception.py
```diff
@@ -16,160 +16,349 @@
...
+            self._nodes_to_add.add(node)
+
+            if self._auto_update_state:
+                self.log.info(f"Updating fleet state after saving node {node}")
+                self.record_fleet_state()
+        else:
+            msg = f"Rejected node {node} because its domain is '{node.domain}' but we're only tracking '{self._domain}'"
+            self.log.warn(msg)
+
+    def __getitem__(self, item):
```
KPrasch, line 261: Just a note: I've deprecated this method in #2513

nucypher/acumen/perception.py
```diff
@@ -16,160 +16,349 @@
...
+        # TODO: for now we're not using `label` in any way, so we're just ignoring it
+        self._nodes_to_remove.add(node.checksum_address)
+
+    def record_remote_fleet_state(self,
+                                  checksum_address: ChecksumAddress,
+                                  state_checksum: str,
+                                  timestamp: maya.MayaDT,
+                                  population: int):
+        nickname = Nickname.from_seed(state_checksum, length=1)
+        self.remote_states[checksum_address] = ArchivedFleetState(checksum=state_checksum,
```
KPrasch, line 361: 👍🏻 - Great.

KPrasch (approved):
Well done @fjarri - Thanks for taking to time to think through the nuances of this PR.

PR #2358, opened 2020-10-14
Introduce gas limit overestimation. Fixes to `nucypher stake` password handling with Trezor. 

### LINKED ISSUE #2337: Reverted CommitToNextPeriod TXs by PolicyManager
**Describe the Bug**
Some commitToNextPeriod TXs are being reverted, not by StakingEscrow but by PolicyManager.
What's worrying is:
* we don't know why it reverts
* it reverts near the end, so it spends practically all the gas in the TX (~ 200,000)
* today it took 3 attempts to commit, so I spent roughly 600K gas (thank goodness is rinkeby eth)

We know it's PolicyManager since the internal TXs show that the problem is in `0x72e5d40bc4ec4391f6c36ba68b33d8840af0941d`, which is current PolicyManager on IBEX.

**To Reproduce**
Not clear yet, still investigating

See one of these failing TXs: https://rinkeby.etherscan.io/tx/0xd5d0d52b6c0d5ddc7411f000980e0f51b299cbb3276a5fac12d2f4b65c5dd41b
If you go to my workers' TX history (https://rinkeby.etherscan.io/address/0xbb0207f53328b6556D13Fd5065Fd9Cb7477eC675), you'll see that today (2020-10-07), there were 2 failed transactions before the successful one.
Still have to gather logs from my machine but I doubt there will be any useful there.

 - OS Version:
 - Nucypher Version:
 - Ethereum Node Version:

cygnusv:
- Attempt to fix #2337 by using an overestimation of 50% for the gas limit when committing to next period.
- Fixes wrong password handling in `nucypher stake` when using Trezor as a signer

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1749,7 +1749,7 @@ class Wallet:
 
         # Blockchain
         self.blockchain = BlockchainInterfaceFactory.get_interface(provider_uri)
-        self.__signer = signer
+        self.signer = signer
 
         self.__get_accounts()
         if client_addresses:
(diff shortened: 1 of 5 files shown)


vzotova (approved):

KPrasch (approved):
:cowboy_hat_face:

derekpierre (approved):
🎸

PR #2359, opened 2020-10-14
Cloud Deploy UX improvements

vepkenez:
* custom per-host SSH hints,
* infer provider if all hosts are the same provider
* bugfix in digital ocean deploy


KPrasch (approved):
:cowboy_hat_face:

Issue #2184, opened 2020-08-19
Global fee range parameters


arjunhassard:
The global fee range is a means of establishing quasi-universal pricing for the NuCypher service. It is enforced via the function ``feeRateRange`` in ``PolicyManager.sol``, which specifies per sharing policy and per 24h period constraints expressed in **WEI**. 

We can discuss these parameters using units of GWEI and approximate $ value. 

At v least we need to be on the same page wrt: 
- Operational cost scenarios
- Value perception & budgets of prospective users & use cases
- Degree of pricing freedom (i.e. range width)
- Pricing decomposition

These are introduced and explained [here](https://docs.google.com/document/d/1531eB1_byKOA6sOMlBrgWDD9KNoIpCx-Fafdzgiq9wI/edit?usp=sharing) 

See #2179 for related documentation PR

KPrasch:
With the introduction of parameters in PR #2363 - This issue is closed.

PR #2363, opened 2020-10-14
Genesis global fee range -> docs

arjunhassard:
addresses #2184 
for parameter origins see https://github.com/nucypher/whitepaper/blob/master/economics/pricing_protocol/NuCypher_Network__Pricing_Protocol_Economics.pdf
- general pricing skew favours demand-side with min fee aiming to be <$0.50 per user per year with 'average' behavior 
- default = min so if stakers are passive it favours demand-side
- assumes ETHUSD of $318 (100-day RA), so slight benefit to service-side if current price holds or rises


vzotova (approved):

docs/source/architecture/service_fees.rst
```diff
@@ -3,89 +3,67 @@
...
+The global fee range is a means of establishing quasi-universal pricing for the NuCypher service. It is enforced via the function ``feeRateRange`` in ``PolicyManager.sol``, which specifies per sharing policy and per 24h period constraints expressed in **WEI**. Note that elsewhere, fee rates are discussed in **GWEI** and fiat (USD).
 
 The minimum fee rate chosen by stakers must fall within the global fee range. The network will launch with the following parameters:
 
 **Minimum fee rate**
 
-XX GWEI (~$X.XX x10^-X) *per period*
-
-XX,XXX GWEI ($X.XXX) *per year*
+50 GWEI (~$1.59 x10^-5) *per period*
```
KPrasch, line 31: As gas prices change, so will this number - it might be better to display only the *gwei* denomination here, and not USD.

docs/source/architecture/service_fees.rst
```diff
@@ -3,89 +3,67 @@
...
-Note that the global fee range is only applicable to stakers and Ursulas. Alices are free to pay as high a rate as they like.
+The fee range must be adhered to in identical fashion by all NuCypher stakers, regardless of their stake size or capacity. The fee range applies to all sharing policies, irrespective of the volume of re-encryption requests or other distinguishing attributes. It also applies equally to all periods in the future, until the moment that the global fee range’s parameters are adjusted or the range is removed, via official governance processes (see below). If an update of this sort occurs, sharing policies that were previously established, but have not yet expired, should not have the per-period fee rate retroactively modified. Note that the global fee range is only applicable to stakers and Ursulas. Alices are free to pay as high a rate as they like.
```
cygnusv, line 45: This is an extraordinary important paragraph, that adds tons of very important details that I actually never considered. Great work!

cygnusv (approved):

docs/source/architecture/service_fees.rst
arjunhassard: I do think they provide a useful reference (with the static exchange rate stated below as a caveat), but happy to remove. I remember @derekpierre had a similar view. Although these conversions change with the ETHUSD rate, not with fluctuations to viable gas prices – so not nearly as volatile!

KPrasch (approved):

derekpierre (approved):
🎸 - great stuff @arjunhassard !

PR #2364, opened 2020-10-14
The return of the Jinja

fjarri:
Since I don't feel we reached a consensus in #2351 , here's a chance to roll back Mako without rolling back the design.

Two things I want to point out:
- look at all the functions at the beginning of `server.py`. This is the kind of thing Mako allows you to avoid. Also all these little files per each function, reminiscent of Matlab.
- if you make a typo in a template, Mako just gives you "'NNN' is undefined" (note that I added the check for that in this PR; the original actually could never fail). Compare that to [this nice traceback](https://github.com/nucypher/nucypher/pull/2351#issuecomment-708079587).

So, don't like Mako? Just press the button.


vepkenez:
Omg no... it was fine!

jMyles:
> Two things I want to point out:

https://getyarn.io/yarn-clip/bf874312-d735-4eeb-9316-1658c7498b29

KPrasch (changes requested):
Let's stick with mako! :-)

cygnusv:
I vote for sticking with mako.

fjarri:
Awesome, I think all the interested parties expressed their opinion here. Closing then.

PR #2366, opened 2020-10-14
Since web3=5.12.2 its methods need positional arguments

fjarri:
`web3` was bumped from `5.12.1` to `5.12.2` in commit ad7935c725df. Evidently, they changed their method calls to some generalized `Method` class, so it is very strict about positional arguments being passed as positional arguments.

Update:

I'm using https://github.com/ethereum/web3.py/blob/a7ed9070e37c34f0e411a7441d4b79d8e5932a36/web3/eth.py to find all methods that are now `Method` instances. For some reason, not all of these changes were caught in tests. The list of changed methods:
* `getBlock()`
* `sendRawTransaction()`
* `getTransactionReceipt()`
* `getTransaction()`
* `sendTransaction()`
* `signTransaction()`

I think I got them all now.

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -629,7 +629,7 @@ class Allocator:
 
         batch_size = 1
         if not gas_limit:
-            block_limit = self.staking_agent.blockchain.client.w3.eth.getBlock(block_identifier='latest').gasLimit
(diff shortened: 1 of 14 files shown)


KPrasch (approved):
:cowboy_hat_face:

fjarri:
Actually there seems to be more of those calls. I wonder why they're not causing trouble - either tests don't cover them, or not all calls are converted to `Method` yet.

PR #2371, opened 2020-10-15
Set teacher min_stake to 15000 NU.

tuxxy:


nucypher/cli/options.py
```diff
@@ -47,7 +47,7 @@
 option_light = click.option('--light', help="Indicate that node is light", is_flag=True, default=None)
 option_lonely = click.option('--lonely', help="Do not connect to seednodes", is_flag=True)
 option_m = click.option('--m', help="M-Threshold KFrags", type=click.INT)
-option_min_stake = click.option('--min-stake', help="The minimum stake the teacher must have to be a teacher", type=click.INT, default=0)
+option_min_stake = click.option('--min-stake', help="The minimum stake the teacher must have to be a teacher", type=click.INT, default=15000)
```
derekpierre, line 50: I think @vepkenez floated the idea of a seed node not even staking - or am I mistaken?

KPrasch: Nodes are not considered valid teachers if they are not staking.

KPrasch: @tuxxy - probably better to use the constant instead of hardcoding the value.

KPrasch:
@tuxxy - What is the status of this PR?

nucypher/cli/options.py
KPrasch: @tuxxy - Im unsure that the denomination is correct here - are you sure it's not nunits?

cygnusv: In other CLI commands, we've used NU, but with the `DecimalRange` type. In this case, it would be `DecimalRange(min=15000)`.

KPrasch:
Picking up this PR so it does not grow stale.

cygnusv (approved):

cygnusv:
@KPrasch The only weird thing is the 2 intermediate commits, which I guess should be squashed together.

KPrasch:
Squashed the intermediate commits.

KPrasch (approved):

Issue #2373, opened 2020-10-15
Way too many blockchain reads happening.


jMyles:
Nodes seem to be performing many, many more blockchain `.call`s than we expected. 

Possible explanations:

* We are evaluating `_staker_is_actually_staking` outside of verification
* We are verifying nodes after they've already been verified (even in the same runtime)

derekpierre:
Additional factor which is the use of the`--prometheus` flag. When enabled, `prometheus` runs a looping call every 10s (too aggressive :scream:)- https://github.com/nucypher/nucypher/blob/main/nucypher/utilities/prometheus/metrics.py#L145

The collectors used collect a variety of information about the worker - https://github.com/nucypher/nucypher/blob/main/nucypher/utilities/prometheus/collector.py

Mostly, token balance information, missing commitments, num known nodes, num workorders, blocknumber. We can probably reduce the interval by a some factor. I suspect part of the original goal for such a quick interval was to be able to detect any weird occurrences quickly and take execute mitigating steps, but perhaps still too aggressive.

BisonAl:
Just a note from our own experience, we've seen a pretty consistent `~200` rpc requests per minute with the `--prometheus` flag enabled.

This has been the case for both `ibex` and `mainnet`.

KPrasch:
Many stakers are successfully using infura's free tier ethereum service with and without Prometheus.  Further optimization is possible, but the issue as described here is resolved.

PR #2376, opened 2020-10-16
Be cool prometheus...just be cool

derekpierre:
Related to #2373 

Reduce the greediness of prometheus metrics collection.

On average prometheus metric collection performs ~40 requests/10s 😱 

- [X] Reduce frequency
- [X] Remove WorkLock event metrics collections.


docs/source/guides/network_node/ursula_configuration_guide.rst
```diff
@@ -218,6 +218,10 @@ parameters to the ``nucypher ursula run`` command:
 The corresponding endpoint, ``http://<node_ip>:<METRICS PORT>/metrics``, can be used as a Prometheus data source for
 monitoring including the creation of alert criteria.
 
+By default metrics will be collected every 90 seconds but this can be modified using the ``--metrics-interval`` option.
```
KPrasch, line 221: :boom:

nucypher/utilities/prometheus/collector.py
```diff
@@ -432,65 +432,3 @@ def _event_occurred(self, event) -> None:
         super()._event_occurred(event)
         self.metrics["current_worker_is_me_gauge"].set(
             self.contract_agent.get_worker_from_staker(self.staker_address) == self.worker_address)
-
-
-class BidRefundCompositeEventMetricsCollector(MetricsCollector):
-    """
-    Collector for both Bid and Refund WorkLock events.
-
-    Both Bid and Refund events additionally update the same metric of how much eth is deposited by the staker
```
cygnusv, line 441: I agree Bid is useless now, but Refund will be useful for a while

jMyles:
One thing I'm not clear on - are we thinking that this is a substantial part of #2373?  How many nodes are actually using the prometheus integration?

nucypher/utilities/prometheus/collector.py
derekpierre: ✔️

derekpierre:
> How many nodes are actually using the prometheus integration?

@jMyles I'm unsure of that proportion of the network, but even if it is a small portion of actual nodes, prometheus is definitely one factor in higher request counts for those who use it. Note that this issue doesn't claim to fix #2373, just related to it.

nucypher/utilities/prometheus/collector.py
```diff
@@ -434,63 +433,19 @@ def _event_occurred(self, event) -> None:
...
+        self.metrics["worklock_deposited_eth_gauge"] = Gauge(f'{metrics_prefix}_worklock_current_deposited_eth',
+                                                             'Worklock deposited ETH',
+                                                             registry=registry)
 
-    def collect(self) -> None:
-        for collector in self.collectors:
-            collector.collect()
+    def _event_occurred(self, event) -> None:
+        super()._event_occurred(event)
+        self.metrics["worklock_deposited_eth_gauge"].set(self.contract_agent.get_deposited_eth(self.staker_address))
```
vzotova, line 451: Sorry for dumb question: I see name of class includes "refund" and here "deposited", could you explain a little bit?

derekpierre: Not dumb at all. Original code from P2O - both Bid and Refund events were tracked and when either event occurred, in addition to event information, they also updated the common metric `worklock_deposited_eth_gauge` - the thought was that bid increased the deposited ETH and refund decreased it. The naming could be improved - but that would involve a change to the dashboard as well.

vzotova: Still do not understand everything, I mean bid is removed, and name contains refund but metrics for deposit. Anyway if it's on purpose then ok

KPrasch: I agree this reads a bit confusing - why set `contract_agent.get_deposited_eth` for the *Refund* event?

derekpierre: When a refund occurs for WorkLock the following also happens:
```python
info.depositedETH = info.depositedETH.sub(refundETH);
```
And I think this is what is attempted to be captured.

vzotova (approved):

KPrasch (approved):
:cowboy_hat_face:

PR #2377, opened 2020-10-16
Actually allow minimum stake of 15,000 NU

derekpierre:
Discovered during testing - ensure that minimum bid of 15,000 NU is actually allowed.

### DIFF
--- a/tests/integration/cli/test_stake_cli_functionality.py
+++ b/tests/integration/cli/test_stake_cli_functionality.py
@@ -838,8 +838,6 @@ def test_create_interactive(click_runner,
     lock_periods = 366
     value = NU.from_nunits(token_economics.minimum_allowed_locked * 11)
 
-    mock_token_agent.get_balance.return_value = token_economics.minimum_allowed_locked - 1
-
     command = ('create',
                '--provider', MOCK_PROVIDER_URI,
                '--network', TEMPORARY_DOMAIN)
@@ -854,12 +852,15 @@ def test_create_interactive(click_runner,
                             YES,
                             INSECURE_DEVELOPMENT_PASSWORD))
 
+    # insufficient existing balance
+    mock_token_agent.get_balance.return_value = token_economics.minimum_allowed_locked - 1
     result = click_runner.invoke(stake, command, input=user_input, catch_exceptions=False)
     assert result.exit_code == 1
     assert INSUFFICIENT_BALANCE_TO_CREATE in result.output
     assert MAXIMUM_STAKE_REACHED not in result.output
     assert SUCCESSFUL_STAKE_INCREASE not in result.output
 
+    # already at max stake
     mock_staking_agent.get_locked_tokens.return_value = token_economics.maximum_allowed_locked
     balance = token_economics.minimum_allowed_locked * 12
     mock_token_agent.get_balance.return_value = balance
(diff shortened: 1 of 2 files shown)


cygnusv (commented):
This is a great find, but it also points to the fact that there may be missing some test somewhere?

vzotova (approved):
looks good
if you have time - would be great to add test for second changed clause, when `upper_limit==lower_limit`

derekpierre:
@vzotova the added test should do that already - that way both the lower limit and upper limit modified checks are tested 😄  

```python
 # successfully stake minimum allowed which equals available balance
 mock_staking_agent.get_locked_tokens.return_value = token_economics.minimum_allowed_locked
 mock_token_agent.get_balance.return_value = token_economics.minimum_allowed_locked
```

vzotova:
@derekpierre 
I mean - If you return only second clause back (`<=`) then test will still work but should fail

derekpierre:
Yep - I tested it, and if either of those conditions is reverted to `<=` the test will fail. This is because lower limit (economics.min_allowed = 15000 NU) does equal upper limit (token_balance = 15000 NU).

derekpierre:
@vzotova try it out - and if I missed something let me know.

vzotova:
@derekpierre I see, you're right, I was confused that we have one test for two statements, but we have two previous tests for failed cases so it's fine

tests/integration/cli/test_stake_cli_functionality.py
```diff
@@ -870,7 +871,32 @@ def test_create_interactive(click_runner,
     assert MAXIMUM_STAKE_REACHED in result.output
     assert SUCCESSFUL_STAKE_INCREASE not in result.output
 
+    # successfully stake minimum allowed which equals available balance
+    mock_staking_agent.get_locked_tokens.return_value = token_economics.minimum_allowed_locked
+    mock_token_agent.get_balance.return_value = token_economics.minimum_allowed_locked
+    min_stake_value = NU.from_nunits(token_economics.minimum_allowed_locked)
+    min_amount_user_input = '\n'.join((str(selected_index),
```
KPrasch, line 878: might be helpful to include some comments here as to what each input is meant for.

KPrasch (approved):
Good catch and test :+1:

tests/integration/cli/test_stake_cli_functionality.py
```diff
@@ -870,7 +871,32 @@ def test_create_interactive(click_runner,
...
+    result = click_runner.invoke(stake, command, input=min_amount_user_input, catch_exceptions=False)
+    assert result.exit_code == 0
+    assert CONFIRM_STAKE_USE_UNLOCKED not in result.output  # default is to use staker address
+    assert INSUFFICIENT_BALANCE_TO_CREATE not in result.output
+    assert PROMPT_STAKE_CREATE_VALUE.format(lower_limit=min_stake_value, upper_limit=min_stake_value) in result.output
+    assert CONFIRM_STAGED_STAKE.format(nunits=str(min_stake_value.to_nunits()),
+                                       tokens=min_stake_value,
+                                       staker_address=surrogate_stakers[selected_index],
+                                       lock_periods=lock_periods) in result.output
+    assert CONFIRM_BROADCAST_CREATE_STAKE in result.output
```
KPrasch, line 895: :+1: - Excellent use of those constants

PR #2378, opened 2020-10-16
Status page fixes

### LINKED ISSUE #2368: Fleet States not... really... working on Ursula status page.
Here's what I see:

Nodes are properly recording fleet states, but they are:

* Not displaying the current fleet state
* Displaying the last known fleet state as a "previous state" (rather than current)
* Not displaying the correct number of nodes from previous states.

@kevinvitale [pointed this out in Discord](https://ptb.discord.com/channels/411401661714792449/411401661714792451/766110465846935602).

From the status page accompanying @kevinvitale's node, the problem is easily observed:

Screenshot from 2020-10-14 18-36-39

In this shot, we expect the Fleet State to read `Bisque Five` instead of `NO FLEET STATE AVAILABLE`, and we expect the states listed underneath to begin with `Coral Tango` (the second one listed), and for the number of nodes listed for each to match the number of nodes in that fleet state, rather than "5 nodes".

fjarri:
* Fixes #2368 (NO_FLEET_STATE_AVAILABLE and related issues). I was taking the current state info from the attributes set by `update_snapshot()`, while the most recent one is actually contained in `.known_nodes`. 
* Displays the correct state population.
* Displays the total number of known nodes (as @cygnusv suggested).
* Spaces out nickname symbols a little (thanks, @jMyles).
    Before:
    Screen Shot 2020-10-16 at 16 23 16
    After:
    Screen Shot 2020-10-16 at 16 22 34

### DIFF
--- a/nucypher/network/templates/basic_status.mako
+++ b/nucypher/network/templates/basic_status.mako
@@ -127,7 +127,11 @@ ${fleet_state_icon(state.checksum, state.nickname, state.population())}
     }
 
     .this-node-info {
-        margin-bottom: 1em;
+        margin-bottom: 2em;
+    }
+
+    h3 {
+        margin-bottom: 0em;
     }
 
     .this-node {
--- a/nucypher/network/templates/basic_status.mako
+++ b/nucypher/network/templates/basic_status.mako
@@ -192,7 +192,7 @@ ${fleet_state_icon(state.checksum, state.nickname, state.population())}
         </tr>
     </table>
 
-    <h3>Known nodes</h3>
+    <h3>${len(known_nodes)} ${"known node" if len(known_nodes) == 1 else "known nodes"}:</h3>
 
     <table class="known-nodes">
         <thead>
--- a/nucypher/network/templates/basic_status.mako
+++ b/nucypher/network/templates/basic_status.mako
@@ -45,7 +45,12 @@ NO FLEET STATE AVAILABLE
 
 
 <%def name="fleet_state_icon_from_state(state)">
-${fleet_state_icon(state.checksum, state.nickname, len(state))}
+${fleet_state_icon(state.checksum, state.nickname, len(state.nodes))}
+</%def>
+
+
+<%def name="fleet_state_icon_from_known_nodes(state)">
+${fleet_state_icon(state.checksum, state.nickname, state.population())}
 </%def>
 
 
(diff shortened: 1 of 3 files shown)


nucypher/network/templates/basic_status.mako
```diff
@@ -184,7 +196,7 @@ ${fleet_state_icon(state.checksum, state.nickname, len(state))}
         </tr>
     </table>
 
-    <h3>Known nodes</h3>
+    <h3>${len(known_nodes)} ${"known node" if len(known_nodes) == 1 else "known nodes"}:</h3>
```
KPrasch, line 199: The logic is creeping in!

fjarri: Would look the same in Jinja, really. It's not business logic, it's display logic.

KPrasch: Fair point - in either case why not per-compute this and pass it as rendering context?

fjarri: That's exactly why - it's not a context, it's display-specific. I could format it as 
```
%if len(known_nodes)==1:
known node
%else:
known nodes
%endif
```
to avoid having Python logic in an evaluated expression, but it's only a formal change.

In general, if we start passing all the little things that only make sense inside the template from outside, we will end up with dozens of parameters in the call. For example, why not precalculate `contrast_colors()` outside, or `checksum[0:8]`, or `character_span()`?

fjarri: To add to that: in my understanding, a template takes data that is convenient to work with in the code (in this case, a node object) and turns it into something that a human can read. Sometimes it means repeating information, working with grammar, adding colors and so on. All that is not the caller's business. 

If we start passing everything to the template, we don't really need a template engine at all - f-strings would suffice. Then, since you'll need to precompute a lot of values, you'll extract that code into a function, or maybe even its own module, and call it by passing only the node object to it. Which means you'll again end up with a Mako template. Except Mako adds one small thing - it inverts the Python syntax in this module so the strings do not need to be escaped anymore.

KPrasch (approved):
:cowboy_hat_face: - Compelling points. Glad for the fixes too!

nucypher/network/templates/basic_status.mako
jMyles: I'm OK with this reasoning.

jMyles (approved):

PR #2410, opened 2020-10-28
restart worker task on fail

### LINKED ISSUE #2405: Ursula does not re-attempt failed replacement transaction
Failure to "retry the retry"... if you will.  Related to https://github.com/nucypher/nucypher/issues/2398

Despite not tracking the gas price from the previous retry, Ursula also does not continue period commitments if this exception is raised even though it continues to run.

### LINKED ISSUE #2411: Restart on unhandled errors in the work tracking loop
Presently, our work tracking loop errback handler, `handle_working_errors`, has no way to call `start` on its `LoopingCall` (`self._tracking_task.start`).

Let's to do this automatically `if not self._abort_on_error`.  This (along with actually changing `self._abort_on_error` to `False`) will prevent one path by which nodes fall into the odd limbo state where a loop has crashed but the node marches on.

vepkenez:
closes #2411 
closes #2405

### DIFF
--- a/nucypher/blockchain/eth/token.py
+++ b/nucypher/blockchain/eth/token.py
@@ -612,6 +612,7 @@ class WorkTracker:
         if self._abort_on_error:
             self.log.critical(f'Unhandled error during node work tracking. {failure!r}',
                               failure=failure)
+            self.stop()
             reactor.callFromThread(self._crash_gracefully, failure=failure)
         else:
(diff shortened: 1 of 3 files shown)


tuxxy (approved):

derekpierre (approved):
🎸

jMyles (changes requested):
Consider https://github.com/vepkenez/nucypher/pull/1

jMyles (approved):

PR #2396, opened 2020-10-21
Simple GitHub action to ensure release note entry included in PRs into main.

derekpierre:
The check that this PR adds, is expected to fail for this PR hehehe 😅 .

Release note entry format readme can be viewed - https://github.com/derekpierre/nucypher/blob/release-notes-action/newsfragments/README.md.

PR #2415, opened 2020-10-30
Update newsfragment documentation

derekpierre:


vepkenez (approved):
Thanks, I needed this 🥇

fjarri (approved):

KPrasch (commented):
Not sure I understand the point of this PR - is it not encouraged to use issue numbers for news fragments?

derekpierre:
Our current GitHub action ensures that the PR has a release note entry file with the PR issue number in the file name. The use of `<ISSUE>` in the document made it seem that the issue number could also be used as @cygnusv pointed out. This PR just makes it explicit that the PR number should be used.

KPrasch:
> the PR number should be used

:thinking:  Perhaps I am out of the loop here, but why not use an issue number?

derekpierre:
> but why not use an issue number?
IMHO for a release we make code changes, and PRs signify that best - not just what the problem was, but also how it was fixed. Issues focus on the manifestation of a problem whereas PRs address the underlying cause.

If we want to allow issue numbers, I'll need to see if linked issues can be used for the automatic check, and update the GitHub Action merged in #2396 .

cygnusv:
If it's of any help, you can have multiple newsfragments files, so you can have the PR and all the issues closed by it. I used this for 2395.

derekpierre:
Thanks @cygnusv, I thought I saw that somewhere. It's possible to do both, a fragment for PR and another for Issue number, but the action currently only verifies for the PR one.

So you can have, just the PR, or PR and Issue, but never the Issue by itself.

KPrasch:
> So you can have, just the PR, or PR and Issue (s), but never the Issue by itself.

:+1: - This seems good enough

derekpierre:
> So you can have, just the PR, or PR and Issue (s), but never the Issue by itself.

I'll add an extra blurb indicating as much.

fjarri:
@KPrasch , a newsfragment is necessarily a part of a PR, so there's a 1-to-1 relationship. On the other hand, a PR may fix some issues, or no issues at all, or just make some progress in fixing an issue. So the arrangement makes sense to me. But the phrasing was indeed confusing, as I discovered myself just a few days ago.

cygnusv (approved):

PR #2423, opened 2020-11-03
Moving domain check into FleetSensor.

### LINKED ISSUE #2356: Sprouts and mature nodes need a single attribute for the value of "domain"
The following occurs on the network status monitor which uses forgetful node storage:

```
File "/Users/derek/Documents/Github/repos/monitor/nucypher-monitor/monitor/crawler.py", line 328, in learn_from_teacher_node
    new_nodes = super().learn_from_teacher_node(*args, **kwargs)
  File "/Users/derek/.local/share/virtualenvs/nucypher-monitor-3hnjUmmD/lib/python3.7/site-packages/nucypher/network/nodes.py", line 764, in learn_from_teacher_node
    remembered_seednodes = self.load_seednodes(record_fleet_state=False)
  File "/Users/derek/.local/share/virtualenvs/nucypher-monitor-3hnjUmmD/lib/python3.7/site-packages/nucypher/network/nodes.py", line 339, in load_seednodes
    nodes_restored_from_storage = self.read_nodes_from_storage() if read_storage else []
  File "/Users/derek/.local/share/virtualenvs/nucypher-monitor-3hnjUmmD/lib/python3.7/site-packages/nucypher/network/nodes.py", line 353, in read_nodes_from_storage
    node_domain = node.domain.decode('utf-8')
builtins.AttributeError: 'Ursula' object has no attribute 'domain'
```

Discord conversation:

```
derek09/29/2020
@dnunez it's forgetful. It's the monitor's SQLiteForgetfulStorage

dnunez09/29/2020
Right.... So that must be it
The type of nodes that storages fetch are different. That's actually not a good API...

...

dnunez09/30/2020
@jMyles @derek the thing is that sprouts have domain but matured nodes don't, as they use it to construct the serving_domain attribute. Maybe we should simplify that and just have domain ?

jMyles09/30/2020
Absolutely.
For a primordial variant of a polymorhpic class to have an attribute that is not either an attribute or a method on a commensurate evolved variant is an antipattern.
In my opinion.

derek09/30/2020
:thumbsup: Agreed!
```

### LINKED ISSUE #2417: Lower-level domain check
Domain checks (ie, ensuring that a given Node B to whom Node A is connecting, for any purpose, or whose metadata Node A is remembering upon discovery from some Node C) are sensible at two layers:

1) A high-level layer, to prevent lower-level logic from needing to be executed inside a large loop of node metadata
2) A low-level layer, to decisively prevent any code path from causing nodes from one domain to spread metadata about nodes from another.

Presently, we have solution 1 in place, but we never implemented solution 2.

I propose that we do the following to achieve this:

* Add `domain` to `FleetSensor`
* Add domain gatekeeping to `FleetSensor.__setitem__`

jMyles:
Domain "leakage", or nodes saving metadata about nodes from other domains (but never being able to verify them) was still possible because domain-checking only occurred in the high-level APIs (and not, for example, when checking metadata POSTed to the node_metadata_exchange endpoint).  This fixes that (fixes #2417).

**As such, the mechanism by which ibex nodes are leaking into some mainnet fleet states is now disabled.**

Additionally, domains are no longer separated into "serving" or "learning".  Each Learner instance now has exactly one domain, and it is called domain.  Fixes #2356.

*Note: the behavior of `handpicked ursulas` is unchanged.  In other words, it is still possible to explicitly specify an off-domain Ursula during an action that takes `handpicked_ursulas` and use it despite it being on a different domain.*

### DIFF
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -50,15 +50,19 @@ class FleetSensor:
         self._marked = defaultdict(list)  # Beginning of bucketing.
         self.states = OrderedDict()
 
-    def __setitem__(self, key, value):
-        self._nodes[key] = value
+    def __setitem__(self, checksum_address, node_or_sprout):
+        if node_or_sprout.domain == self.domain:
+            self._nodes[checksum_address] = node_or_sprout
 
-        if self._tracking:
-            self.log.info("Updating fleet state after saving node {}".format(value))
-            self.record_fleet_state()
+            if self._tracking:
+                self.log.info("Updating fleet state after saving node {}".format(node_or_sprout))
+                self.record_fleet_state()
+        else:
+            msg = f"Rejected node {node_or_sprout} because its domain is '{node_or_sprout.domain}' but we're only tracking '{self.domain}'"
+            self.log.warn(msg)
 
-    def __getitem__(self, item):
-        return self._nodes[item]
+    def __getitem__(self, checksum_address):
+        return self._nodes[checksum_address]
 
     def __bool__(self):
         return bool(self._nodes)
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -42,7 +42,8 @@ class FleetSensor:
     log = Logger("Learning")
     FleetState = namedtuple("FleetState", ("nickname", "icon", "nodes", "updated", "checksum"))
 
-    def __init__(self):
+    def __init__(self, domain: str):
+        self.domain = domain
         self.additional_nodes_to_track = []
         self.updated = maya.now()
         self._nodes = OrderedDict()
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -224,7 +224,7 @@ class Learner:
         self._learning_listeners = defaultdict(list)
         self._node_ids_to_learn_about_immediately = set()
 
-        self.__known_nodes = self.tracker_class()
+        self.__known_nodes = self.tracker_class(domain=domain)
         self._verify_node_bonding = verify_node_bonding
 
         self.lonely = lonely
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -210,7 +210,7 @@ class Learner:
         self.log = Logger("learning-loop")  # type: Logger
 
         self.learning_deferred = Deferred()
-        self.learning_domain = domain
+        self.domain = domain
         if not self.federated_only:
             default_middleware = self.__DEFAULT_MIDDLEWARE_CLASS(registry=self.registry)
         else:
@@ -293,8 +293,8 @@ class Learner:
 
         discovered = []
 
-        if self.learning_domain:
-            canonical_sage_uris = self.network_middleware.TEACHER_NODES.get(self.learning_domain, ())
+        if self.domain:
+            canonical_sage_uris = self.network_middleware.TEACHER_NODES.get(self.domain, ())
 
             for uri in canonical_sage_uris:
                 try:
@@ -352,16 +352,16 @@ class Learner:
         for node in stored_nodes:
             try:  # Workaround until #2356 is fixed
                 node_domain = node.domain.decode('utf-8')
-            except:
-                node_domain = node.serving_domain
-            if node_domain != self.learning_domain:
+            except:  # TODO: The heck is happening here?
+                node_domain = node.domain
+            if node_domain != self.domain:
                 invalid_nodes[node_domain].append(node)
                 continue
             restored_node = self.remember_node(node, record_fleet_state=False)  # TODO: Validity status 1866
             restored_from_disk.append(restored_node)
 
         if invalid_nodes:
-            self.log.warn(f"We're learning about domain '{self.learning_domain}', but found nodes from other domains; "
+            self.log.warn(f"We're learning about domain '{self.domain}', but found nodes from other domains; "
                           f"let's ignore them. These domains and nodes are: {dict(invalid_nodes)}")
 
         return restored_from_disk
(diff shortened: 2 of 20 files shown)


nucypher/network/nodes.py
```diff
@@ -352,16 +352,16 @@ def read_nodes_from_storage(self) -> List:
         for node in stored_nodes:
             try:  # Workaround until #2356 is fixed
                 node_domain = node.domain.decode('utf-8')
-            except:
-                node_domain = node.serving_domain
-            if node_domain != self.learning_domain:
+            except:  # TODO: The heck is happening here?
+                node_domain = node.domain
+            if node_domain != self.domain:
```
cygnusv, line 362: Does this PR fixes #2356? It seems so. If that's the case, then the try/except can be removed, as it was only a workaround until we had a unified attribute for domains.

nucypher/network/nodes.py
```diff
@@ -102,6 +102,11 @@ def __bytes__(self):
     def stamp(self) -> bytes:
         return self.processed_objects['verifying_key'][0]
 
+    @property
+    def domain(self) -> str:
+        domain_bytes = PartiallyKwargifiedBytes.__getattr__(self, "domain")
+        return domain_bytes.decode("utf-8")
```
cygnusv, line 108: Wonderful 💯

tests/integration/learning/test_domains.py
```diff
@@ -89,8 +89,35 @@ def test_learner_restores_metadata_from_storage(lonely_ursula_maker, tmpdir):
...
+
+def test_learner_ignores_stored_nodes_from_other_domains(lonely_ursula_maker, tmpdir):
+
+    learner, other_staker = lonely_ursula_maker(domain="call-it-mainnet",
+                                        know_each_other=True,
+                                        quantity=2)
+
+    pest, *other_ursulas_from_the_wrong_side_of_the_tracks = lonely_ursula_maker(domain="i-dunno-testt-maybe",
+                                       quantity=5,
+                                       know_each_other=True)
```
cygnusv, line 107: Fix indentation of these lines.

cygnusv (approved):
Great work @jMyles ! I only have one comment in my review.
BTW, after this is merged and released, is it necessary for nodes to forget?

nucypher/acumen/perception.py
```diff
@@ -42,22 +42,27 @@ class FleetSensor:
...
-        self._nodes[key] = value
+    def __setitem__(self, checksum_address, node_or_sprout):
+        if node_or_sprout.domain == self.domain:
+            self._nodes[checksum_address] = node_or_sprout
 
-        if self._tracking:
-            self.log.info("Updating fleet state after saving node {}".format(value))
-            self.record_fleet_state()
+            if self._tracking:
+                self.log.info("Updating fleet state after saving node {}".format(node_or_sprout))
```
derekpierre, line 58: Should we move this to `debug` instead of `info`?

nucypher/network/nodes.py
```diff
@@ -102,6 +102,11 @@ def __bytes__(self):
     def stamp(self) -> bytes:
         return self.processed_objects['verifying_key'][0]
 
+    @property
+    def domain(self) -> str:
+        domain_bytes = PartiallyKwargifiedBytes.__getattr__(self, "domain")
```
derekpierre, line 107: Does it make sense to store the result of the first run in a member variable (eg. `self.domain`) so that you don't have to get the bytes and then decode it every time?

nucypher/network/nodes.py
derekpierre: Agreed! 👍

jMyles: Oh uhh, right!  I almost forgot about that convo.  Yep, it fixes it.

So I'll nix this block.

nucypher/network/nodes.py
jMyles: As an optimization, you mean?

Since we will almost always do this once per node, I think it might actually be a counter-optimization.  The timing here is tiny to begin with: it takes about 1ms to decode 5000 strings on my laptop.  However, it takes 3ms to decode 5000 strings and use the result to set an attribute on an instance.  See this gist: https://gist.github.com/jMyles/e6458de8a557274883e1f2c6cf2e5862

So I think that setting an attribute will actually be slower.  Either way, I think this optimization is premature.  I am definitely interested in ensuring, in the coming weeks, that the total runtime of a grant or retrieve is as slim as possible, but there are places where we can probably shave off hundreds of milliseconds, so I'd prefer to leave single milliseconds alone for the moment until we consider it more holistically.

Note: `PartiallyKwargifiedBytes` already caches its result (which is substantially more onerous) internally.

nucypher/acumen/perception.py
jMyles: Oh, uhh... I guess so?  Is this one really noisy?  I don't think we really roll with `_tracking` on, and I presume that if someone / something does (like a test) that it's for a reason, and worth logging as info.  But... I dunno?  You like debug better for this?

vepkenez (approved):
❤️

nucypher/acumen/perception.py
derekpierre: I was uncertain about its noisiness, but was thinking about it in the same spirit given the network has some ~1800 nodes. I didn't know the default value for `_tracking` (incorrectly assumed True), but if by default it is `False` then info log level is probably fine.

derekpierre (approved):
🎸

jMyles:
> Great work @jMyles ! I only have one comment in my review.
> BTW, after this is merged and released, is it necessary for nodes to forget?

I already answered in Discord, but I'll also leave the same comment here for posterity:

My expectation is that off-domain nodes will be ignored upon loading from storage, so that a simple restart will purge them from known_nodes.

PR #2348, opened 2020-10-12
[Concept] Client layer middleware, CLI Signer handling, Stale Deprecations

KPrasch:
Based on #2335, Includes #2331, working to advance #2141 

###### Clients & Interfaces
- Rework of web3 middleware management on the client layer
- Introduces stalecheck middleware (dual speed, longer syncing delay permitted in tests)
- Implement client-layer default gas strategies via inhert-and-override
- Auto-detection of POA  middleware requirement
- Avoid nonce reusage in tests with block cache disable/reconfigure
- Simplifies interface connection logic

###### Signer handling
- Remove implicit default signer from configuration flow and internally
- Understand `--signer` as a required option where possible

###### Ursula 
- `fire_and_forget ` -> `wait_for_receipt`
- `WorkTracker` tracks pending worker commitment transactions by hash (needs follow-up)

###### Tests & Code Quality
- Completes test dummy mocking integration for Interfaces, Clients, Providers, and Web3
- Organize and optimize imports

###### CLI
- Numerous language and formatting improvements found while manually testing `stake` and `ursula` command.

##### Deprecation

- `--poa` flag (in favor of auto-detection)
- `--geth` Internal geth process management (stale, unused)
- `NuCypherGethDevProcess` and friends (unused, stale)
- `ContractAdministrator.deploy_network_contracts` (stale, too optimistic, use `Testerchain.bootstrap_network` in tests)
- `nucypher-deploy contracts` Automated deployment series (stale, save actions and utils)
- `EthereumClient.sync` - RPC sync polling (unused)
- `actors.Wallet` (replaces with signers and dependency injection)
- `BlockchainInterface.to_dict` and `BlockchainInterface.from_dict` (unused)

PR #2425, opened 2020-11-04
More work tracker improvements. Adds --gas-price option to nucypher stake

### LINKED ISSUE #2406: Ursula submitted commitToNextPeriod() transaction with less than 50k gas limit causing it to fail
Ursula mainnet worker running v4.1.0 submitted a commitToNextPeriod() transaction that failed due to an Out of Gas condition.

The gas limit on the failing transaction was 49,691 which seems unusually low.

Seventeen minutes later the worker submitted a new commitToNextPeriod() transaction that succeeded - that one had a gas limit of 280,248 of which 169,534 was actually used.

Unclear why the first transaction had such a low gas limit, perhaps there is a transient issue with gas estimation.

Failed transaction: https://etherscan.io/tx/0x7f418a89fb200e1dafa9c15b4e8d3265fd25ccfd1595130660075497e3752728
Successful transaction: https://etherscan.io/tx/0xb040b25d00d72b2325b367f43f5e0e887e0a517df1a4779096a642c0ae17ab34

Console log output in Ursula was not terribly enlightening:
```
Broadcasting COMMITTONEXTPERIOD Transaction (0.001017942110745368 ETH @ 20.485442248 gwei)...
Broadcasting COMMITTONEXTPERIOD Transaction (0.00574356834034656 ETH @ 20.49459172 gwei)...
```

### LINKED ISSUE #2421: Missing `KeyError` is preventing worker from committing to next period
After four successful periods of commitment, I got this missing `KeyError`...this is something that's happened to me in the past, as well.

Nucypher, Version `4.1.0`

<pre>
<code>
Decrypting ursula keyring...
Checking worker settings: waiting for bonding and funding ...
    ✓ Worker is bonded to (<REDACTED>)!
    ✓ Worker is funded with <REDACTED> ETH!
✓ Worker settings confirmed
Starting services...
✓ Database pruning
✓ Work Tracking
✓ Prometheus Exporter
Starting Ursula on <REDACTED>:9151
Working ~ Keep Ursula Online!
Broadcasting COMMITTONEXTPERIOD Transaction (<REACTED>)...
Broadcasting COMMITTONEXTPERIOD Transaction (<REACTED>)...
Broadcasting COMMITTONEXTPERIOD Transaction (<REACTED>)...
Broadcasting COMMITTONEXTPERIOD Transaction (<REACTED>)...
Unhandled error during node work tracking. <twisted.python.failure.Failure builtins.KeyError: None>
Unhandled Error
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/base.py", line 1283, in run
    self.mainLoop()
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/base.py", line 1292, in mainLoop
    self.runUntilCurrent()
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/base.py", line 913, in runUntilCurrent
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/task.py", line 239, in __call__
    d = defer.maybeDeferred(self.f, *self.a, **self.kw)
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/base.py", line 886, in runUntilCurrent
    f(*a, **kw)
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 608, in _crash_gracefully
    failure.raiseException()
  File "/usr/local/lib/python3.7/site-packages/twisted/python/failure.py", line 488, in raiseException
    raise self.value.with_traceback(self.tb)
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(*args, **kw)
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 715, in _do_work
    self.__handle_replacement_commitment(current_block_number=current_block_number)
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 692, in __handle_replacement_commitment
    if wait_time_in_seconds > self.max_confirmation_time():
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 572, in max_confirmation_time
    expected_time = EXPECTED_CONFIRMATION_TIME_IN_SECONDS[self.gas_strategy]
builtins.KeyError: None
</code>
</pre>

cygnusv:
- In the Worker's commitment tracker, the logic to check for what's the current period should happen before the logic to track pending transactions. Otherwise, it's possible to try to fire a replacement transaction in the context of a different period. Closes #2406
- Now ``BlockchainInterface.gas_strategy`` always has a value; previously it was possible to pass ``None`` via the constructor (e.g. if the config file had an explicit ``"null"`` value). Closes #2421 
- Added support for a user-provided gas price to the ``nucypher stake`` command, using ``--gas-price GWEI``.

### DIFF
--- a/newsfragments/2406.bugfix.rst
+++ b/newsfragments/2406.bugfix.rst
@@ -1,2 +1,1 @@
-In the Worker's commitment tracker, the logic to check for what's the current period should happen before the logic to track pending transactions.
-Otherwise, it's possible to try to fire a replacement transaction in the context of a different period.
+Fixes logical bug in ``WorkTracker`` to ensure commitment transactions can only be issued once per period.
--- a/newsfragments/2406.bugfix.rst
+++ b/newsfragments/2406.bugfix.rst
@@ -1,0 +1,2 @@
+In the Worker's commitment tracker, the logic to check for what's the current period should happen before the logic to track pending transactions.
+Otherwise, it's possible to try to fire a replacement transaction in the context of a different period.
--- a/newsfragments/2425.feature.rst
+++ b/newsfragments/2425.feature.rst
@@ -1,0 +1,1 @@
+Added support for a user-provided gas price to the ``nucypher stake`` command, using ``--gas-price GWEI``.
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -22,7 +22,7 @@ import sys
 import time
 import traceback
 from decimal import Decimal
-from typing import Callable
+from typing import Callable, Union
 from typing import Dict, Iterable, List, Optional, Tuple
 
 import click
@@ -31,6 +31,7 @@ from constant_sorrow.constants import FULL, WORKER_NOT_RUNNING
 from eth_tester.exceptions import TransactionFailed as TestTransactionFailed
 from eth_typing import ChecksumAddress
 from eth_utils import to_canonical_address, to_checksum_address
+from hexbytes import HexBytes
 from web3 import Web3
 from web3.exceptions import ValidationError
 from web3.types import TxReceipt
@@ -1632,7 +1633,7 @@ class Worker(NucypherTokenActor):
 
     @only_me
     @save_receipt  # saves txhash instead of receipt if `fire_and_forget` is True
-    def commit_to_next_period(self, fire_and_forget: bool = True) -> TxReceipt:
+    def commit_to_next_period(self, fire_and_forget: bool = True) -> Union[TxReceipt, HexBytes]:
         """For each period that the worker makes a commitment, the staker is rewarded"""
         txhash_or_receipt = self.staking_agent.commit_to_next_period(worker_address=self.__worker_address,
                                                                      fire_and_forget=fire_and_forget)
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -280,7 +280,7 @@ class BlockchainInterface:
             self.client.inject_middleware(geth_poa_middleware, layer=0)
 
         self.client.add_middleware(middleware.time_based_cache_middleware)
-        # self.client.add_middleware(middleware.latest_block_based_cache_middleware)
+        # self.client.add_middleware(middleware.latest_block_based_cache_middleware)  # TODO: This line causes failed tests and nonce reuse in tests. See #2348.
         self.client.add_middleware(middleware.simple_cache_middleware)
 
         self.set_gas_strategy()
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -39,7 +39,6 @@ from hexbytes.main import HexBytes
 from web3 import Web3, middleware
 from web3.contract import Contract, ContractConstructor, ContractFunction
 from web3.exceptions import TimeExhausted, ValidationError
-from web3.gas_strategies import time_based
 from web3.middleware import geth_poa_middleware
 from web3.providers import BaseProvider
 from web3.types import TxReceipt
@@ -157,7 +156,7 @@ class BlockchainInterface:
                  provider_process=NO_PROVIDER_PROCESS,
                  provider_uri: str = NO_BLOCKCHAIN_CONNECTION,
                  provider: BaseProvider = NO_BLOCKCHAIN_CONNECTION,
-                 gas_strategy: Union[str, Callable] = DEFAULT_GAS_STRATEGY):
+                 gas_strategy: Union[str, Callable] = None):
 
         """
         TODO: #1502 - Move to API docs.
@@ -231,7 +230,7 @@ class BlockchainInterface:
         self.client = NO_BLOCKCHAIN_CONNECTION         # type: EthereumClient
         self.transacting_power = READ_ONLY_INTERFACE
         self.is_light = light
-        self.gas_strategy = gas_strategy
+        self.gas_strategy = gas_strategy or self.DEFAULT_GAS_STRATEGY
 
     def __repr__(self):
         r = '{name}({uri})'.format(name=self.__class__.__name__, uri=self.provider_uri)
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -280,21 +280,26 @@ class BlockchainInterface:
             self.log.debug('Injecting POA middleware at layer 0')
             self.client.inject_middleware(geth_poa_middleware, layer=0)
 
-        # Gas Price Strategy:
-        # Bundled web3 strategies are too expensive for Infura (it takes ~1 minute to get a price),
-        # so we use external gas price oracles, instead (see #2139)
-        if isinstance(self.client, InfuraClient):
-            gas_strategy = datafeed_fallback_gas_price_strategy
-            self.gas_strategy = 'fast'  # FIXME
-        else:
-            gas_strategy = self.get_gas_strategy(self.gas_strategy)
-        self.client.set_gas_strategy(gas_strategy=gas_strategy)
-        gwei_gas_price = Web3.fromWei(self.client.gas_price_for_transaction(), 'gwei')
-        self.log.debug(f"Currently, our gas strategy returns a gas price of {gwei_gas_price} gwei")
-
         self.client.add_middleware(middleware.time_based_cache_middleware)
         # self.client.add_middleware(middleware.latest_block_based_cache_middleware)
         self.client.add_middleware(middleware.simple_cache_middleware)
+
+        self.set_gas_strategy()
+
+    def set_gas_strategy(self, gas_strategy: Optional[Callable] = None):
+        if gas_strategy:
+            reported_gas_strategy = f"fixed/{gas_strategy.name}"
+        elif isinstance(self.client, InfuraClient):
+            gas_strategy = datafeed_fallback_gas_price_strategy
+            self.gas_strategy = 'fast'
+            reported_gas_strategy = "datafeed/fast"
+        else:
+            reported_gas_strategy = f"web3/{self.gas_strategy}"
+            gas_strategy = self.get_gas_strategy(self.gas_strategy)
+        self.client.set_gas_strategy(gas_strategy=gas_strategy)
+        gwei_gas_price = Web3.fromWei(self.client.gas_price_for_transaction(), 'gwei')
+        self.log.debug(f"Using gas strategy '{reported_gas_strategy}'. "
+                       f"Currently, it returns a gas price of {gwei_gas_price} gwei")
 
     def connect(self):
 
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -156,7 +156,7 @@ class BlockchainInterface:
                  provider_process=NO_PROVIDER_PROCESS,
                  provider_uri: str = NO_BLOCKCHAIN_CONNECTION,
                  provider: BaseProvider = NO_BLOCKCHAIN_CONNECTION,
-                 gas_strategy: Union[str, Callable] = None):
+                 gas_strategy: Optional[Union[str, Callable]] = None):
 
         """
         TODO: #1502 - Move to API docs.
@@ -659,7 +659,7 @@ class BlockchainInterface:
                          gas_estimation_multiplier: Optional[float] = None,
                          confirmations: int = 0,
                          fire_and_forget: bool = False,  # do not wait for receipt.  See #2385
-                         ) -> dict:
+                         ) -> Union[TxReceipt, HexBytes]:
 
         if fire_and_forget:
             if confirmations > 0:
--- a/nucypher/blockchain/eth/token.py
+++ b/nucypher/blockchain/eth/token.py
@@ -14,14 +14,14 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
+
 import random
 from _pydecimal import Decimal
 from collections import UserList
 from enum import Enum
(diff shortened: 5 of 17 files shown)


nucypher/blockchain/eth/interfaces.py
```diff
@@ -231,7 +230,7 @@ def __init__(self,
         self.client = NO_BLOCKCHAIN_CONNECTION         # type: EthereumClient
         self.transacting_power = READ_ONLY_INTERFACE
         self.is_light = light
-        self.gas_strategy = gas_strategy
+        self.gas_strategy = gas_strategy or self.DEFAULT_GAS_STRATEGY
```
fjarri, line 233: So we're expecting the user of this class to ignore the type attribute and pass `None` anyway?

nucypher/blockchain/eth/interfaces.py
```diff
@@ -280,21 +279,26 @@ def attach_middleware(self):
             self.log.debug('Injecting POA middleware at layer 0')
             self.client.inject_middleware(geth_poa_middleware, layer=0)
 
-        # Gas Price Strategy:
-        # Bundled web3 strategies are too expensive for Infura (it takes ~1 minute to get a price),
-        # so we use external gas price oracles, instead (see #2139)
-        if isinstance(self.client, InfuraClient):
+        self.client.add_middleware(middleware.time_based_cache_middleware)
+        # self.client.add_middleware(middleware.latest_block_based_cache_middleware)
```
fjarri, line 283: What's with the commented line?

nucypher/blockchain/eth/actors.py
```diff
@@ -1632,7 +1633,7 @@ def last_committed_period(self) -> int:
 
     @only_me
     @save_receipt  # saves txhash instead of receipt if `fire_and_forget` is True
-    def commit_to_next_period(self, fire_and_forget: bool = True) -> TxReceipt:
+    def commit_to_next_period(self, fire_and_forget: bool = True) -> Union[TxReceipt, HexBytes]:
```
fjarri, line 1636: I remember bumping into this pattern before, and currently in our code the propagation of these types is... quite disordered. Like, `staking_agent.commit_to_next_period()` claims to return `Union[TxReceipt, HexBytes]`, but the return value in the function body is explicitly marked as `TxReceipt`, which is in turn obtained from a function that is marked as returning `dict`, in which the return value is obtained from a function returning `Union[TxReceipt, HexBytes]`. So... it probably deserves a PR by itself.

nucypher/blockchain/eth/interfaces.py
cygnusv: The commented line was already there before, I just moved this block. The commented line was something that seemed to cause problems in some tests, but I haven't checked it personally.

nucypher/blockchain/eth/token.py
```diff
@@ -641,11 +641,8 @@ def __tracking_consistency_check(self) -> bool:
             return True  # OK!
         if txs_in_mempool > len(self.__pending):  # We're missing some pending TXs
             return False
-
-        # TODO: Not sure what to do in this case, but let's do this for the moment
-        #       Note that the block my have changed since the previous query
-        # elif txs_in_mempool < len(self.__pending):  # Our tracking is somehow outdated
-        #     return False
+        else:  # TODO: What to do when txs_in_mempool < len(self.__pending)? What does this imply?
+            return True
```
fjarri, line 645: If I understand the logic correctly, it could happen if somehow some transactions were canceled externally. 

Feels like it would be better to at least return `False`, or maybe even raise an exception/write something in log in this case. 

An interesting side effect here is that if `txs_in_mempool == len(self.__pending) - 1`, and we return `False`, `__track_pending_commitments()` adds an element in `self.__pending`, and the subsequent calls to `__tracking_consistency_check()` return `True` again. Maybe it's worth using a different variable to store that state.

nucypher/blockchain/eth/interfaces.py
KPrasch: No.  The type attribute annotation is inaccurate.

nucypher/blockchain/eth/interfaces.py
KPrasch: This line causes failed tests and nonce reuse in tests - to be addressed in another PR.

nucypher/blockchain/eth/actors.py
KPrasch: When the `fire_and_forget` feature was added, a receipt of transaction hash can be returned.  Not surprised if there is *hint* inaccuracy.

nucypher/blockchain/eth/token.py
KPrasch: > If I understand the logic correctly, it could happen if somehow some transactions were canceled externally.

Yes - this might be from a transaction unassociated with nucypher, broadcasted before startup.

nucypher/blockchain/eth/interfaces.py
cygnusv: I'm unsure what's the issue here. There are many situations where we don't explicitly initialize the `BlockchainInterface` with a gas strategy (e.g., a staker has a configuration file with `null` as gas strategy because none was set with `nucypher ursula init`, see #2421), and the default should be used. In current `main`, the default is set as an argument default, but we sometimes need to change it in subclasses of `BlockchainInterface`, and we have already discussed in the past how this pattern doesn't work and we've been using incorrectly (see description of 4a263940ba037d89302213b0a8679b0fdaff1b65).

If the issue is the lack of an `Optional` type hint, I've just added.

nucypher/blockchain/eth/actors.py
cygnusv: > Like, `staking_agent.commit_to_next_period()` claims to return `Union[TxReceipt, HexBytes]`, but the return value in the function body is explicitly marked as `TxReceipt`, which is in turn obtained from a function that is marked as returning `dict`, in which the return value is obtained from a function returning `Union[TxReceipt, HexBytes]`

This was addressed in this PR, at least partially (see 35af6a3a07ee73eaa530198df1a05342a1de438c). I just included some other type hints there as well.

newsfragments/2406.bugfix.rst
```diff
@@ -0,0 +1,2 @@
+In the Worker's commitment tracker, the logic to check for what's the current period should happen before the logic to track pending transactions.
```
KPrasch, line 1: At minimum - I'd change should -> must...

```suggestion
 In the Worker's commitment tracker, the logic to check for what's the current period must happen before the logic to track pending transactions.
```

... However this sounds more like a detailed description than it does a news headline.  Here is a more terse example:

```suggestion
Ensure replacement transactions can only be issued once per period.
```

newsfragments/2425.feature.rst
```diff
@@ -0,0 +1 @@
+Added support for a user-provided gas price to the ``nucypher stake`` command, using ``--gas-price GWEI``.
```
KPrasch, line 1: :+1: - Well formed newsfragement!

nucypher/utilities/gas_strategies.py
```diff
@@ -64,23 +63,43 @@ def datafeed_fallback_gas_price_strategy(web3: Web3, transaction_params: TxParam
...
 }
+
+
+#
+# Fixed-price gas strategy
+#
+
+
+def construct_fixed_price_gas_strategy(gas_price, denomination: str = "wei") -> Callable:
+    gas_price_in_wei = Web3.toWei(gas_price, denomination)
```
KPrasch, line 98: Well done - perhaps a variant can be built on top of this using values calculated in gas estimation metrics.

nucypher/utilities/gas_strategies.py
```diff
@@ -64,23 +63,43 @@ def datafeed_fallback_gas_price_strategy(web3: Web3, transaction_params: TxParam
...
     Enriches the web3 exceptions thrown by gas strategies
     """
-    @functools.wraps(web3_gas_strategy)
+    web3_gas_strategy = __RAW_WEB3_GAS_STRATEGIES[speed]
+
     def _wrapper(*args, **kwargs):
         try:
             return web3_gas_strategy(*args, **kwargs)
         except ValidationError as e:
             raise GasStrategyError("Calling the web3 gas strategy failed, probably due to an unsynced chain.") from e
```
KPrasch, line 76: Helpful - but hopefully never encountered if stalecheck middleware is injected.

nucypher/cli/options.py
```diff
@@ -42,6 +42,7 @@
 option_event_name = click.option('--event-name', help="Specify an event by name", type=click.STRING)
 option_federated_only = click.option('--federated-only/--decentralized', '-F', help="Connect only to federated nodes", is_flag=True, default=None)
 option_force = click.option('--force', help="Don't ask for confirmation", is_flag=True)
+option_gas_price = click.option('--gas-price', help="Use this gas price (in GWEI)", type=DecimalRange(min=0))
```
KPrasch, line 45: Perhaps start of with a new click type here for gwei,

nucypher/cli/commands/stake.py
```diff
@@ -269,7 +271,11 @@ def create_character(self, emitter, config_file):
         )
 
     def get_blockchain(self):
-        return self.staker_options.get_blockchain()
+        blockchain = self.staker_options.get_blockchain()
+        if self.gas_price:
+            fixed_price_strategy = construct_fixed_price_gas_strategy(gas_price=self.gas_price, denomination="gwei")
```
KPrasch, line 276: This seems like very high-level, ad-hoc place to set this gas strategy - Perhaps this is beyond the scope of this PR but I think it will be more efficient and dry to compose this step into the init of `EthereumClient.` (see never-to-be-merged PR #2348)

nucypher/blockchain/eth/interfaces.py
KPrasch: A TODO will be helpful here, perhaps indirectly linked to #2348 etc.

nucypher/blockchain/eth/token.py
KPrasch: This block also may execute for a pending period commitment transaction, right after a node manually (re)starts.

nucypher/utilities/gas_strategies.py
cygnusv: What do you envision here? I don't understand

nucypher/utilities/gas_strategies.py
cygnusv: 🤞

nucypher/cli/commands/stake.py
cygnusv: Yes, definitely. As done other times, this is more a PoC approach that definitely can be extended and iterated in other PRs; for the moment it's only included in `nucypher stake`.

nucypher/utilities/gas_strategies.py
KPrasch: Not to repeat myself here but:

> using values calculated in gas estimation metrics


Something like `construct_estimated_price_gas_strategy`

KPrasch: > What do you envision here? 

A set of known defaults instead of a parameter for `gas_price`.

KPrasch (approved):
Approved with comments.  Thanks @cygnusv !

nucypher/cli/commands/stake.py
KPrasch: It is worth including a TODO for the future author here

nucypher/utilities/gas_strategies.py
KPrasch: It's also very presumptuous, and contingent on web3 middleware configuration to say *" probably due to an unsynced chain."*.

"Probably" is pretty deflating in terms of helping the end-user (try top offer direction help).  How about:

"Verify your ethereum provider connection and syncing status"

nucypher/utilities/gas_strategies.py
cygnusv: I'm still not following you. What's the difference with, say, web3 gas strategies? Or are you saying something like `gas_strategy_50_gwei`?

KPrasch: Hmm - yes...? If it does not make sense perhaps not a well formed idea after all.

derekpierre (approved):
👍 - thanks @cygnusv. Pre-emptively approving, but I agree with the RFCs from @KPrasch .

newsfragments/2406.bugfix.rst
cygnusv: ✅

nucypher/utilities/gas_strategies.py
cygnusv: ✅

nucypher/cli/commands/stake.py
cygnusv: ✅

nucypher/blockchain/eth/interfaces.py
cygnusv: ✅

nucypher/cli/options.py
cygnusv: ✅

nucypher/blockchain/eth/token.py
```diff
@@ -641,11 +641,8 @@ def __tracking_consistency_check(self) -> bool:
             return True  # OK!
         if txs_in_mempool > len(self.__pending):  # We're missing some pending TXs
             return False
-
-        # TODO: Not sure what to do in this case, but let's do this for the moment
-        #       Note that the block my have changed since the previous query
-        # elif txs_in_mempool < len(self.__pending):  # Our tracking is somehow outdated
-        #     return False
+        else:  # TODO: What to do when txs_in_mempool < len(self.__pending)? What does this imply?
```
fjarri, line 644:
```suggestion
        else:  # TODO #2429: What to do when txs_in_mempool < len(self.__pending)? What does this imply?
```

PR #2440, opened 2020-11-17
Relock dependencies

cygnusv:
* Relocks dependencies, using latest pipenv. Note how this time pipenv is much more opinionated in the version specifiers.
    * In particular, cryptography >= 3.2 (due to security issues, see https://github.com/nucypher/nucypher/pull/2409) and web3py <= 5.12.3 (for some reason, 5.13 fails for us)
* Updates the relocking script to remove automatic headers to requirements files.
* Forces usage of pip==20.2.4 in CI to avoid problem with pip resolver (https://github.com/pypa/pip/issues/9180). Waiting for pip 20.3.4 (see #2493)
* Changes Python version of docker images to 3.8

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -14,7 +14,7 @@ cached-property==1.5.2
 certifi==2020.12.5
 cffi==1.14.4
 cfgv==3.2.0; python_full_version >= '3.6.1'
-chardet==3.0.4
+chardet==4.0.0; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'
 click==7.1.2
 coincurve==13.0.0
 colorama==0.4.4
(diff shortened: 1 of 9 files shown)


KPrasch:
Related to https://github.com/nucypher/constantSorrow/issues/14

dev-requirements.txt
```diff
@@ -1,143 +1,50 @@
 -i https://pypi.python.org/simple
-apipkg==1.5
+apipkg==1.5; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
```
vepkenez, line 2: In the past there has been a lot of resistance to having these version qualifiers in here.  I personally am not objecting it but maybe @KPrasch and @derekpierre want to weigh in.

vepkenez (approved):

dev-requirements.txt
cygnusv: Yeah, those were automatically added by pipenv. I don't know if there's a flag we can use for this...

KPrasch: I find it to be strange that this is the default behavior for the latest pipenv, but who am I to blow against the wind - there must be a good reason, right? :laughing:

KPrasch (approved):

dev-requirements.txt
vepkenez: I have spent some time trying a lot of flags and have been unable to eliminate them.

PR #2441, opened 2020-11-17
Allow arbitrary decimal precision as input to nucypher CLI

### LINKED ISSUE #2438: Unable to increase stake with a very high precision
**Describe the Bug**
The CLI refused to allow me to increase my stake using the full amount available because it rounded _up_ the amount causing it to be outside the valid range.

**To Reproduce**
Attempt to increase an existing stake with an amount having more than 13 decimals.

**Traceback or Screenshots (Optional)**
Here's the approximate experience (amounts changed for privacy):
```
Select Stake (0): 0
Enter stake value in NU (up to 123.000000000007891234 NU): 123.00000000000789123
Error: 123.0000000000079 is not in the valid range of 0 to 123.000000000007891234.
Enter stake value in NU (up to 123.000000000007891234 NU): 123.00000000000789
Error: 123.0000000000079 is not in the valid range of 0 to 123.000000000007891234.
Enter stake value in NU (up to 123.000000000007891234 NU): 123.0000000000078

(continues)
...
```

 - OS Version: Ubuntu 20.04
 - Nucypher Version: 4.1.2
 - Ethereum Node Version:  Geth/1.9.24

cygnusv:
Closes #2438

### DIFF
--- a/nucypher/blockchain/eth/token.py
+++ b/nucypher/blockchain/eth/token.py
@@ -51,7 +51,7 @@ class NU:
     The easiest way to use NU, is to pass an int, Decimal, or str, and denomination string:
 
     Int:    nu = NU(100, 'NU')
-    Int:    nu_wei = NU(15000000000000000000000, 'NuNit')
+    Int:    nu = NU(15000000000000000000000, 'NuNit')
 
     Decimal:  nu = NU(Decimal('15042.445'), 'NU')
     String: nu = NU('10002.302', 'NU')
@@ -59,7 +59,7 @@ class NU:
     ...or alternately...
 
     Decimal: nu = NU.from_tokens(Decimal('100.50'))
-    Int: nu_wei = NU.from_nu_wei(15000000000000000000000)
+    Int: nu = NU.from_nunits(15000000000000000000000)
 
     Token quantity is stored internally as an int in the smallest denomination,
     and all arithmetic operations use this value.
--- a/nucypher/blockchain/eth/token.py
+++ b/nucypher/blockchain/eth/token.py
@@ -48,21 +48,24 @@ class NU:
     An amount of NuCypher tokens that doesn't hurt your eyes.
     Wraps the eth_utils currency conversion methods.
 
-    The easiest way to use NU, is to pass an int, float, or str, and denomination string:
+    The easiest way to use NU, is to pass an int, Decimal, or str, and denomination string:
 
     Int:    nu = NU(100, 'NU')
     Int:    nu_wei = NU(15000000000000000000000, 'NuNit')
 
-    Float:  nu = NU(15042.445, 'NU')
+    Decimal:  nu = NU(Decimal('15042.445'), 'NU')
     String: nu = NU('10002.302', 'NU')
 
     ...or alternately...
 
-    Float: nu = NU.from_tokens(100.50)
+    Decimal: nu = NU.from_tokens(Decimal('100.50'))
     Int: nu_wei = NU.from_nu_wei(15000000000000000000000)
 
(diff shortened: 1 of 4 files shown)


nucypher/cli/commands/stake.py
```diff
@@ -625,7 +625,7 @@ def increase(general_config: GroupGeneralConfig,
             emitter.echo(MAXIMUM_STAKE_REACHED, color='red')
             raise click.Abort
 
-        stake_value_range = click.FloatRange(min=0, max=upper_limit.to_tokens(), clamp=False)
+        stake_value_range = DecimalRange(min=0, max=upper_limit.to_tokens(), clamp=False)
         value = click.prompt(PROMPT_STAKE_INCREASE_VALUE.format(upper_limit=upper_limit),
                              type=stake_value_range)
     value = NU.from_tokens(value)
```
derekpierre, line 631: Any concern here since `from_tokens(...)` takes a parameter with a type hint of `Union[int, float, str]`, which does not include Decimal?

Stepping back, it seems odd that `NU.to_tokens()` returns a `Decimal` but `NU.from_tokens(...)` takes a `Union[int, float, str]`. Does `Decimal` need to be added to that `Union`? It seems this is what currency.to_wei (which is eventually called within `NU.__init__()`) ended up already doing:
```python
def to_wei(number: Union[int, float, str, decimal.Decimal], unit: str) -> int:
```

cygnusv: Yes, you're right. I'm going to add Decimal to those type hints. Actually, now that you mention, we shouldn't encourage using `float` precisely because of the issues this PR is trying to solve.

nucypher/blockchain/eth/token.py
```diff
@@ -48,21 +48,24 @@ class NU:
...
 
-    Float:  nu = NU(15042.445, 'NU')
+    Decimal:  nu = NU(Decimal('15042.445'), 'NU')
     String: nu = NU('10002.302', 'NU')
 
     ...or alternately...
 
-    Float: nu = NU.from_tokens(100.50)
+    Decimal: nu = NU.from_tokens(Decimal('100.50'))
     Int: nu_wei = NU.from_nu_wei(15000000000000000000000)
```
derekpierre, line 62:
```suggestion
    Int: nunits = NU.from_nunits(15000000000000000000000)
```

nucypher/blockchain/eth/token.py
```diff
@@ -48,21 +48,24 @@ class NU:
     An amount of NuCypher tokens that doesn't hurt your eyes.
     Wraps the eth_utils currency conversion methods.
 
-    The easiest way to use NU, is to pass an int, float, or str, and denomination string:
+    The easiest way to use NU, is to pass an int, Decimal, or str, and denomination string:
 
     Int:    nu = NU(100, 'NU')
     Int:    nu_wei = NU(15000000000000000000000, 'NuNit')
```
derekpierre, line 54: Remove reference to nu_wei; use nunits instead

```suggestion
    Int:    nunits = NU(15000000000000000000000, 'NuNit')
```

derekpierre (approved):
🎸  - a couple of suggestions regarding our current docstring.

KPrasch (approved):
Looks good, consider a CLI unit test showing this behavior to prevent a regression.

cygnusv:
> Looks good, consider a CLI unit test showing this behavior to prevent a regression.

Thanks for the reminder, @KPrasch !! I added test with decimal values and it made me caught an additional bug in the CLI with respect to decimal inputs 👍

vepkenez (approved):

Issue #2435, opened 2020-11-13
Replacement TX uses a wrong gas limit


cygnusv:
**Describe the Bug**
We've seen many examples recently of commitment TXs that are replaced with a new one but with a much smaller gas limit, which is bound to fail due to "out of gas". This usually resolves itself with a 3rd TX with the correct gas limit, but it's clear that this is not good. 

The expected behavior is that the second TX (i.e., the first replacement) is created with a correct gas price, which would lead to it being accepted normally.

**To Reproduce**
Unknown

**Traceback or Screenshots (Optional)**
See history of this staker, for example, Nov 13th: https://etherscan.io/address/0xa09fdd0551e6f93d5b3a0c6872613c26f658e8ef

There are 2 TXs listed in there for that day, the first one is reported as failed due to out of gas (https://etherscan.io/tx/0x5f8828c1a398a067447f2f385bd5421c721a53422908603ea61df71447d18381). If you click to see more details of the TX, you can see that this is example is even worse, as there are 3 previous TXs with the same nonce. The first one has a correct gas limit, but it's replaced with a second with incorrect limit. This one in turn is replaced by a 3rd TX with correct limit, and this one is again replaced by the TX linked above, with an incorrect limit, and which is the one that ends up being included in a block (and failing due to out of gas, obviously). 

**System (please complete the following information):**
 - OS Version:
 - Nucypher Version: 4.1.2
 - Ethereum Node Version: Infura

cygnusv:
Thanks to @vzotova, we realized that `eth_estimateGas` uses the state for the pending block as default. The implication of this is that if you already have a pending commitment TX and create a new one to replace it, any calls to `eth_estimateGas` will assume that the pending TX is final and, therefore, the new commitment TX will have a reduced gas limit (as expected).

The solution is to take a better approach to replace TXs: rather than firing a new one, we should "modify" the original one with a new price (using `eth.modifyTransaction`). Note that we should also try to be careful with the gas price and ensure that these replacements are always overpriced with respect to the original (#2398). I personally lean towards implementing a new endpoint in the client and blockchain interface layer for this.

cygnusv:
Fixed by #2486

Issue #2398, opened 2020-10-23
Replacement transaction underpriced


cygnusv:
**Describe the Bug**
It seems we should increase the gas price for replacement TXs

**To Reproduce**
TBD

**Traceback or Screenshots (Optional) **
```
Unhandled error during node work tracking. <twisted.python.failure.Failure builtins.ValueError: {'code': -32000, 'message': 'replacement transaction underpriced'}>
Unhandled Error
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/base.py", line 1283, in run
    self.mainLoop()
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/base.py", line 1292, in mainLoop
    self.runUntilCurrent()
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/base.py", line 913, in runUntilCurrent
    call.func(*call.args, call.kw)
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/task.py", line 239, in call
    d = defer.maybeDeferred(self.f, *self.a, self.kw)
--- <exception caught here> ---
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/base.py", line 886, in runUntilCurrent
    f(a, **kw)
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 608, in _crash_gracefully
    failure.raiseException()
  File "/usr/local/lib/python3.7/site-packages/twisted/python/failure.py", line 488, in raiseException
    raise self.value.with_traceback(self.tb)
  File "/usr/local/lib/python3.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred
    result = f(args, kw)
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 715, in _do_work
    self.handle_replacement_commitment(current_block_number=current_block_number)
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 701, in handle_replacement_commitment
    tx_firing_block_number=tx_firing_block_number)
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 680, in fire_replacement_commitment
    replacement_txhash = self.fire_commitment()  # replace
  File "/usr/local/lib/python3.7/site-packages/nucypher/blockchain/eth/token.py", line 756, in __fire_commitment
[...]

PR #2445, opened 2020-11-18
Not so fast... (More stability for the commitment tracker, gas price configuration improvements)

### LINKED ISSUE #2344: Allow Ursula to suspend commitment transactions
In the event of a serious gas spike, it's possible that Ursula will simply decide that it's not feasible to make the commitment transactions.

In such an event, some rational and otherwise honest nodes may opt to simply `halt` or otherwise kill their `nucypher` process.  But, given the choice, and with a solution of equal or greater ease, it's possible that they will instead choose to forego commitment but still continue to run.  Let's make it easy to do that.

### LINKED ISSUE #2420: More robust logic for gas price data feeds
Current default gas price from data feeds is hardcoded to EtherChain, which seems stuck for several hours (it reports block  11165953, while we're already on 11167843). At this moment (Nov 1st 00:08 UTC), the price given by this data feed was 60 gwei, while according to Etherscan it should have been ~20 gwei. 

We need a more complex and robust logic that can deal with these type of problems, considering several data sources.

### LINKED ISSUE #2428: WorkTracker can fall into an infinite loop of failure and doom
Current `main` activates the logic that restarts the work tracker in case of a failure. That's fine for some failures (e.g. some temporary problem with the ETH provider), but not for others, like a previous bug or a bad configuration that can repeatedly lead to failure.

I've just seen this behaviour in a test that fails in CI (https://app.circleci.com/pipelines/github/nucypher/nucypher/6397/workflows/45d0806d-e0cd-4f86-86ec-d2b87c236b03/jobs/110542), although it doens't fail locally to me. In any case, see the artifacts of that job, in particular logs for container 3, and you'll see at the end of the log fail how it fails forever:

```
2020-11-06T11:49:22+0000 [stake-tracker#info] START WORK TRACKING
2020-11-06T11:49:22+0000 [stake-tracker#info] Checking for new period. Current period is 18572
2020-11-06T11:49:22+0000 [stake-tracker#warn] Unhandled error during work tracking: 'Traceback (most recent call last):\n  File "/home/circleci/nucypher/nucypher/blockchain/eth/token.py", line 620, in handle_working_errors\n    self.start()\n  File "/home/circleci/nucypher/nucypher/blockchain/eth/token.py", line 599, in start\n    d = self._tracking_task.start(interval=self.random_interval(), now=act_now)\n  File "/home/circleci/.local/lib/python3.7/site-packages/twisted/internet/task.py", line 194, in start\n    self()\n  File "/home/circleci/.local/lib/python3.7/site-packages/twisted/internet/task.py", line 239, in __call__\n    d = defer.maybeDeferred(self.f, *self.a, **self.kw)\n--- <exception caught here> ---\n  File "/home/circleci/.local/lib/python3.7/site-packages/twisted/internet/defer.py", line 151, in maybeDeferred\n    result = f(*args, **kw)\n  File "/home/circleci/nucypher/nucypher/blockchain/eth/token.py", line 737, in _do_work\n    interval = onchain_period - self.worker.last_committed_period\n  File "/home/circleci/nucypher/nucypher/blockchain/eth/actors.py", line 1630, in last_committed_period\n    period = self.staking_agent.get_last_committed_period(staker_address=self.checksum_address)\n  File "/home/circleci/nucypher/nucypher/blockchain/eth/decorators.py", line 105, in wrapped\n    return func(*args, **kwargs)\n  File "/home/circleci/nucypher/nucypher/blockchain/eth/agents.py", line 553, in get_last_committed_period\n    period: int = self.contract.functions.getLastCommittedPeriod(staker_address).call()\n  File "/home/circleci/.local/lib/python3.7/site-packages/web3/contract.py", line 875, in __call__\n    clone._set_function_info()\n  File "/home/circleci/.local/lib/python3.7/site-packages/web3/contract.py", line 885, in _set_function_info\n    self.kwargs\n  File "/home/circleci/.local/lib/python3.7/site-packages/web3/_utils/contracts.py", line 163, in find_matching_fn_abi\n    raise ValidationError(message)\nweb3.exceptions.ValidationError: \nCould not identify the intended function with name `getLastCommittedPeriod`, positional argument(s) of type `(<class \'constant_sorrow.constants.NO_BLOCKCHAIN_CONNECTION\'>,)` and keyword argument(s) of type `{}`.\nFound 1 function(s) with the name `getLastCommittedPeriod`: [\'getLastCommittedPeriod(address)\']\nFunction invocation failed due to no matching argument types.\n'
```

The underlying failure in this traceback is probably a minor bug in that test, btw.

### LINKED ISSUE #2444: Ursula unable to use any gas price lower than instant when serviced by infura endpoint
**Describe the Bug**

Ursula unable to use any gas price lower than instant when serviced by infura endpoint

**To Reproduce**

send transaction with Infura endpoint and gas setting to SLOW

 - OS Version:
 - Nucypher Version: 4.1.2
 - Ethereum Node Version:  Geth 1.9.23-stable-8c2f2715

cygnusv:
**Type of PR:**
- [X] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
The main goal of this PR is to make the commitment tracker more stable and allow some gas price reductions. In particular:
* Fixes the problem with gas prices for clients serviced by Infura, which had a fixed fast gas strategy (#2444)
* Introduces a new composite gas price strategy that takes the __median__ price between 3 different gas price oracles, replacing the previous approach which was a fallback chain. The new approach is much more resilient against sporadic problems in one of the oracles, like a spike or getting stuck (#2420).
* Introduces the possibility of setting a max gas price limit via the `NUCYPHER_MAX_GAS_PRICE_GWEI` environment variable. This makes possible to wait indefinitely for the commitment TX to clear (within current period), and in the worst case, to skip the commitment on that period if the prices are too high (#2344). Note however that this makes the tracker more brittle. Also, there's disagreement wrt the use of an envvar (#2485).
* Several improvements to the commitment tracker, in particular, if there's an internal error, it will not retry immediately but after 15 minutes (#2428). 
* Takes a flexible approach with regards to naming of gas prices, as different oracles or providers use different names for similar prices (e.g., "slow" vs "low", "medium" vs "standard", etc). This functionality is, however, disabled for the moment as there are some rough spots in the commitment tracker which will be better addressed during the refactor (See #2447).

**Issues fixed/closed:**
- Closes #2444 
- Closes #2420
- Closes #2428 
- Closes #2344 

**Why it's needed:**
This PR is mainly an improvement for node operators, as it's very stable. It also surfaced many underlying problems and provided tons of information for an upcoming refactor of the commitment tracker.
This PR doesn't address any product-dev issue.

**Notes for reviewers:**
This PR is tricky. On the one hand, it has many improvements and several fixes, and has been tested in mainnet with several nodes to ensure it's fairly stable. I think that, from the node operator perspective, it's definitely an improvement. On the other hand, this PR still doesn't solve some outstanding issues with the commitment tracker (#2398, #2435), and we've learned along the way that solving them needs heavy refactoring of the tracker. IMHO, let's try to move on for the moment with this approach, but acknowledging at the same time that the commitment tracker will change soon.


tests/unit/test_datafeeds.py
```diff
@@ -120,29 +192,122 @@ def test_upvest():
...
+    mock_zoltu_gas_price = 4000
+    mock_rpc_gas_price = 42
+
+    def construct_mock_gas_strategy(gas_price) -> Callable:
+        def _mock_gas_strategy(web3, tx=None):
+            return gas_price
+        return _mock_gas_strategy
+
+    # In normal circumstances, all datafeeds in the strategy work, and the median is returned
+    with patch('nucypher.utilities.datafeeds.UpvestGasPriceDatafeed.construct_gas_strategy',
```
KPrasch, line 230: I think - this can be flattened by using the `mocker` fixture.

tests/unit/test_gas_strategies.py
```diff
@@ -33,3 +39,36 @@ def test_fixed_price_gas_strategy():
...
+
+    # Let's start assuming there's no limit
+    monkeypatch.delenv(NUCYPHER_ENVVAR_MAX_GAS_PRICE_GWEI, raising=False)
+    wrapped_strategy = max_price_gas_strategy_wrapper(mock_gas_strategy)
+
+    for price in gas_prices_wei:
+        assert wrapped_strategy("web3", "tx") == price
+
+    # If the corresponding envvar has a valid price in gwei, use this as the max gas price
+    monkeypatch.setenv(NUCYPHER_ENVVAR_MAX_GAS_PRICE_GWEI, max_gas_price_gwei)
```
KPrasch, line 60: Perhaps we can accept it for this PR for the sake of a near-term, but this option seems much more like a character configuration value than an environment variable - At minimum I suggest adding a TODO tied to a follow-up Issue for future follow-up.

```
nucypher ursula config --max-gas-price 150 
```

cygnusv: Opened #2485

vepkenez:
Now that this is all working as we want, should we replace/augment/append the `NUCYPHER_MAX_GAS_PRICE_GWEI` envar with a --cli-arg which updates the config file?

@KPrasch has successfully argued that we should do this for consistency.

KPrasch:
Yeah that sounds like a very sensible adjustment, I'll get right on it and have a few commits by the end of day. Can be included in this PR, or a follow-up

cygnusv:
> Yeah that sounds like a very sensible adjustment, I'll get right on it and have a few commits by the end of day. Can be included in this PR, or a follow-up

@KPrasch @vepkenez I opened issue #2485 to track this. I'd prefer not do this in this PR, basically to reduce time to merge. The problem I see with using a configuration value is that, currently, this would imply restarting the node for changing the limit, and I don't see why we'd want to do this when the envvar allows to adjust the price at runtime.

KPrasch:
Follow-up: After further discussion in #general (discord) - I am not going to make commits on this effort tonight.

docs/source/support/faq.rst
```diff
@@ -373,3 +373,51 @@ You do not have the ability to change staking configs with CoinList.
 There is one setting only – 6 month duration, restake on, winddown on – for all CoinList participants.
 
 See NuCypher coinlist for more details: `CoinList <https://coinlist.co/asset/nucypher>`_
+
+
+Q: How does my worker node chooses what price to use? Can I control this?
```
derekpierre, line 378:
```suggestion
Q: How does my worker node choose what price to use for transactions? Can I control this?
```

docs/source/support/faq.rst
```diff
@@ -373,3 +373,51 @@ You do not have the ability to change staking configs with CoinList.
...
+based on the approximate confirmation time:
+
+- ``slow``: Confirmation expected within **one hour**.
+- ``medium``: Confirmation expected within **five minutes**.
+- ``fast``: Confirmation expected within **one minute**.
+
+Note that the times are an approximation, and the confirmation time may vary
+when gas prices experience more volatility.
+In such situations, transactions may not be mined on the expected time;
+your node keeps track of the transactions and is capable of automatically
```
derekpierre, line 394:
```suggestion
In such situations, transactions may not be mined within the expected time.
However, your node keeps track of the transactions and is capable of automatically
```

docs/source/support/faq.rst
```diff
@@ -373,3 +373,51 @@ You do not have the ability to change staking configs with CoinList.
...
+
+.. code:: bash
+
+    (nucypher)$ nucypher ursula config --gas-strategy medium
+
+
+There's an additional, advanced control mechanism for limiting the maximum
+gas price that your node can spend.
+The ``NUCYPHER_MAX_GAS_PRICE_GWEI`` environment variable can be used to express,
+in `gwei`, such limit.
```
derekpierre, line 410:
```suggestion
The ``NUCYPHER_MAX_GAS_PRICE_GWEI`` environment variable, expressed
in `gwei`, can be used to specify this limit.
```

docs/source/support/faq.rst
```diff
@@ -373,3 +373,51 @@ You do not have the ability to change staking configs with CoinList.
...
+This is complementary to the gas strategy that you may have configured.
+
+.. warning::
+
+    If you set a maximum limit and the gas prices are high or very close
+    to that limit, it's possible that your node doesn't get
+    the commitment transaction included in the blockchain.
```
derekpierre, line 417: Unsure what you mean by "very close"...

```suggestion
    If you set a maximum limit and gas prices remain higher
    than that limit, it's possible that your node will not get
    the commitment transaction included in the blockchain.
```

docs/source/support/faq.rst
```diff
@@ -373,3 +373,51 @@ You do not have the ability to change staking configs with CoinList.
...
+    This implies that you can miss some rewards.
+
+.. warning::
+
+    The maximum gas price limit is an experimental feature and may change or
+    be removed in the future.
```
derekpierre, line 423:
```suggestion
    The maximum gas price limit is an experimental feature and may be changed
    removed in the future.
```

newsfragments/2445.feature.rst
```diff
@@ -0,0 +1 @@
+Improve gas strategy selection: Infura users now can choose between ``slow``, ``medium`` and ``fast``.
```
derekpierre, line 1: Should we also mention the env variable? Or is it that it is likely to change and mentioning it would be confusing?

If we mention it this time around, later on for any subsequent change that modifies/changes its usage, we can reference the env variable and explain the modification.

Just something to think about.

nucypher/blockchain/eth/interfaces.py
```diff
@@ -232,6 +232,10 @@ def __init__(self,
         self.client = NO_BLOCKCHAIN_CONNECTION
         self.transacting_power = READ_ONLY_INTERFACE
         self.is_light = light
+
+        # FIXME: Not ready to give users total flexibility. Let's stick for the moment to known values. See #2447
+        if gas_strategy not in ('slow', 'medium', 'fast', 'free', None):
```
derekpierre, line 237: Can we use key values from `WEB3_GAS_STRATEGIES.keys()` to prevent us from having to hardcode `'slow', 'medium', 'fast'`?

tests/unit/test_datafeeds.py
```diff
@@ -120,29 +192,81 @@ def test_upvest():
...
+                   return_value=construct_mock_gas_strategy(mock_etherchain_gas_price)):
+            with patch('nucypher.utilities.datafeeds.ZoltuGasPriceDatafeed.construct_gas_strategy',
+                       return_value=construct_mock_gas_strategy(mock_zoltu_gas_price)):
+                datafeed_median_gas_price_strategy = construct_datafeed_median_strategy()
+                assert datafeed_median_gas_price_strategy("web3", "tx") == median([mock_etherchain_gas_price,
+                                                                                   mock_upvest_gas_price,
+                                                                                   mock_zoltu_gas_price])
+
+    # If, for example, Upvest fails, the median is computed using the other two feeds
+    with patch('nucypher.utilities.datafeeds.UpvestGasPriceDatafeed._probe_feed',
```
KPrasch, line 242: These mocks and assertions are very repetitive and can be reduced in a helperfunction, or possibly fixturization.  Additionally, the nested `with` blocks can be avoided entirely if you use the `mocker`, `module_mocker`, or `session_mocker` fixtures, since they have their own setup and tear down methods.

vzotova: Agreed, hard to read

vzotova (approved):

nucypher/blockchain/eth/interfaces.py
```diff
@@ -232,7 +234,12 @@ def __init__(self,
         self.client = NO_BLOCKCHAIN_CONNECTION
         self.transacting_power = READ_ONLY_INTERFACE
         self.is_light = light
+
+        # FIXME: Not ready to give users total flexibility. Let's stick for the moment to known values. See #2447
+        if gas_strategy not in ('slow', 'medium', 'fast', 'free', None):  # FIXME: What is 'None' doing here?
```
derekpierre, line 239: Can we use key values from `WEB3_GAS_STRATEGIES.keys()` to prevent us from having to hardcode 'slow', 'medium', 'fast'?

I guess technically this can be part of #2447

nucypher/cli/options.py
```diff
@@ -46,7 +46,9 @@
 option_event_name = click.option('--event-name', help="Specify an event by name", type=click.STRING)
 option_federated_only = click.option('--federated-only/--decentralized', '-F', help="Connect only to federated nodes", is_flag=True, default=None)
 option_force = click.option('--force', help="Don't ask for confirmation", is_flag=True)
-option_gas_price = click.option('--gas-price', help="Use this gas price (in GWEI)", type=GWEI)
+option_gas_price = click.option('--gas-price', help="Set a static gas price (in GWEI)", type=GWEI)
+option_gas_strategy = click.option('--gas-strategy', help="Operate with a specified gas price strategy", type=click.STRING)  # TODO: GAS_STRATEGY_CHOICES
+option_max_gas_price = click.option('--max-gas-price', help="Maximum acceptable gas price in Gwei", type=GWEI)
```
derekpierre, line 51: Matching format of `option_gas_price`

```suggestion
option_max_gas_price = click.option('--max-gas-price', help="Maximum acceptable gas price (in GWEI)", type=GWEI)
```

derekpierre (commented):
Also have some prior text-based RFCs that are now hidden on the PR main page, take a look at those.

docs/source/support/faq.rst
KPrasch: :heavy_check_mark:

newsfragments/2445.feature.rst
KPrasch: :heavy_check_mark:

KPrasch:
Here's how this looks in action:

Screenshot from 2020-12-24 10-54-55

KPrasch (approved):

vepkenez (approved):

PR #2446, opened 2020-11-19
Help command displaying default config and log paths 

KPrasch:
Provides a quick and easy way to get the default configuration root and log directory paths for reconfiguration and troubleshooting.

`nucypher --config` example:

```bash
(nucypher) k@nova:~/Git/nucypher$ nucypher --config
/home/k/.local/share/nucypher
/home/k/.cache/nucypher/log

```


cygnusv:
This is wonderful. I've thought so many times in having something like this 🤩

nucypher/cli/painting/help.py
```diff
@@ -35,6 +36,12 @@ def echo_solidity_version(ctx, param, value):
     ctx.exit()
 
 
+def echo_config_location(ctx, param, value):
+    if not value or ctx.resilient_parsing:
+        return
+    click.secho(f"{DEFAULT_CONFIG_ROOT}\n{USER_LOG_DIR}")
```
derekpierre, line 42: Should we label the output? eg.
'Configuration Root: <>'
'Logging Directory: <>'

derekpierre (commented):
This is pretty cool 😎

nucypher/cli/painting/help.py
KPrasch: Is that a comment or a question? 😊

There is a part of me that thinks this command will be used with bash standard output redirection.

derekpierre: Both... 🤔  😅 

> There is a part of me that thinks this command will be used with bash standard output redirection
Interesting - what's the use case there?

cygnusv: Imagine something like `tail -f $(nucypher --log-path)`.
That requires however some modifications to current approach:
* Separate flags for the two current paths (e.g., `--config-root` for config root, `--log-path` for logs).
* Point to files, not directories
* Remove labels

cygnusv (approved):
This is a simple but great PR. I've wished many times to have this as I often forget what's the default location of the log folder, and this changes for different OSs. @KPrasch I wonder what's your opinion on https://github.com/nucypher/nucypher/pull/2446#discussion_r528672025, but I'm more than happy to have this PR as is. 

Can you consider to add a reference to this somewhere in the docs?

nucypher/cli/painting/help.py
derekpierre: So the use case is potentially usage as part of a script - that makes sense

> Point to files, not directories
Is pointing to files too specific? - if the goal is to use the config output information for some other more involved processing task

For example: 
* get the config root, and the do some processing on `ursula.json` or `stakeholder.json` to verify some setting, or to backup nucypher data eg. `ursula.db`
OR
* get the log-path, and search for some log entry in any of the log files including older rolled over files

cygnusv:
> Can you consider to add a reference to this somewhere in the docs?

https://docs.nucypher.com/en/latest/support/faq.html#q-where-is-my-ursula-config-path

^ This is a good place to reference it in the docs

derekpierre (approved):
🎸

KPrasch:
Responded to RFCs, applied a little of each suggestion.   

These bash commands are now possible as an example: 
- `tail -f $(nucypher --logging-path)/nucypher.log`
- `cat ($nucypher --config-path)/alice.json`

cygnusv (approved):

PR #2452, opened 2020-12-04
Remove imported 'tests' reference from packaged constants.

KPrasch:


vzotova (approved):

cygnusv:
We should try to come up with a test that demonstrates the failure being fixed here.

KPrasch:
Opened follow-up https://github.com/nucypher/nucypher/issues/2454

PR #2467, opened 2020-12-10
Include offending ursula list on nucypher.policy.policies.Rejected

KPrasch:


cygnusv (approved):
I'm fine with merging this now, but I noticed the empty newsfragment? Is this intentional? If so, I'm fine with that too, but would it cause any weird rendering in the release notes?

PR #2474, opened 2020-12-11
Pretty worker & staker console messages

KPrasch:
Extricated from #2348

### DIFF
--- a/newsfragments/2474.misc.rst
+++ b/newsfragments/2474.misc.rst
@@ -1,0 +1,1 @@
+Improves console messages for stakeholder CLI initialization and worker startup.
--- a/nucypher/cli/literature.py
+++ b/nucypher/cli/literature.py
@@ -375,7 +375,7 @@ DECRYPTING_CHARACTER_KEYRING = 'Authenticating {name}'
 #
 
 
-CONFIRM_URSULA_IPV4_ADDRESS = "Detected IPv4 address ({rest_host}) - Is this is the public-facing address of Ursula?"
+CONFIRM_URSULA_IPV4_ADDRESS = "Detected IPv4 address ({rest_host}) - Is this the public-facing address of Ursula?"
 
 COLLECT_URSULA_IPV4_ADDRESS = "Enter Ursula's public-facing IPv4 address:"
 
--- a/nucypher/cli/literature.py
+++ b/nucypher/cli/literature.py
@@ -92,7 +92,7 @@ CONFIRM_INCREASING_STAKE = "Confirm increase stake (index: {stake_index}) by {va
 
 INSUFFICIENT_BALANCE_TO_INCREASE = "There are no tokens to increase stake"
 
-INSUFFICIENT_BALANCE_TO_CREATE = "There are not enough tokens to create stake"
+INSUFFICIENT_BALANCE_TO_CREATE = "Insufficient NU for stake creation."
 
 MAXIMUM_STAKE_REACHED = "Maximum stake reached, can't lock more"
 
@@ -216,11 +216,10 @@ PROMPT_WORKER_ADDRESS = "Enter worker address"
 CONFIRM_WORKER_AND_STAKER_ADDRESSES_ARE_EQUAL = """
 
 {address}
-The worker address provided is the same as the staking account.
+The worker address provided is the same as the staker.
 It is *highly recommended* to use a different accounts for staker and worker roles.
 
-Continue?
-"""
+Continue using the same account for worker and staker?"""
 
 SUCCESSFUL_WORKER_BONDING = "\nWorker {worker_address} successfully bonded to staker {staking_address}"
 
@@ -300,8 +299,11 @@ NO_FEE_TO_WITHDRAW = "No policy fee can be withdrawn."
 # Configuration
 #
 
-MISSING_CONFIGURATION_FILE = """No {name} configuration file found. 
-To create a new persistent {name} run 'nucypher {init_command}'
+MISSING_CONFIGURATION_FILE = """
+
+No {name} configuration file found.  To create a new {name} configuration run:
+
+nucypher {init_command}
 """
 
 
@@ -330,7 +332,7 @@ Delete all {name} character files including:
 
 Are you sure?"""
 
-SUCCESSFUL_DESTRUCTION = "Successfully destroyed NuCypher configuration"
+SUCCESSFUL_DESTRUCTION = "Successfully destroyed nucypher configuration"
 
 CONFIRM_FORGET_NODES = "Permanently delete all known node data?"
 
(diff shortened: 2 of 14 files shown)


nucypher/cli/literature.py
```diff
@@ -342,16 +344,26 @@
...
+SUCCESSFUL_NEW_STAKEHOLDER_CONFIG = """
+Configured new stakeholder!
+Wrote JSON configuration to {filepath}
+
+* Review configuration     -> nucypher stake config
+* View connected accounts  -> nucypher stake accounts
+* Create a new stake       -> nucypher stake create
+* Bond a worker            -> nucypher stake bond-worker
+* List active stakes       -> nucypher stake list
+
```
cygnusv, line 356: 💯

cygnusv (approved):

KPrasch:
I'd like to run through this PR manually a bit more before merging - feel free to continue reviewing.

nucypher/cli/literature.py
derekpierre: 🔥

nucypher/cli/literature.py
```diff
@@ -363,7 +375,7 @@
 #
 
 
-CONFIRM_URSULA_IPV4_ADDRESS = "Is this the public-facing IPv4 address ({rest_host}) you want to use for Ursula?"
+CONFIRM_URSULA_IPV4_ADDRESS = "Detected IPv4 address ({rest_host}) - Is this is the public-facing address of Ursula?"
```
derekpierre, line 378:
```suggestion
CONFIRM_URSULA_IPV4_ADDRESS = "Detected IPv4 address ({rest_host}) - Is this the public-facing address of Ursula?"
```

tests/acceptance/cli/ursula/test_federated_ursula.py
```diff
@@ -46,10 +46,10 @@ def test_initialize_ursula_defaults(click_runner, mocker):
     assert result.exit_code == 0
 
     # REST Host
-    assert 'Is this the public-facing IPv4 address' in result.output
+    assert "Is this is the public-facing address of Ursula? " in result.output
```
derekpierre, line 49: Related change made in `literature.py`
```suggestion
    assert "Is this the public-facing address of Ursula? " in result.output
```

derekpierre (commented):
Any reason for the empty newsfragment?

KPrasch (commented):
> Any reason for the empty newsfragment?

Any reason *to* have a newsfragment?  Blank ones are acceptable IMO.  See web3.py as an example https://web3py.readthedocs.io/en/stable/releases.html#misc.  Are you suggesting that we establish a convention of *always* completing the newsfragment's content?

derekpierre:
> Are you suggesting that we establish a convention of always completing the newsfragment's content?
I guess I always assumed that if we were adding a fragment (link to the release page) i.e. the change is worthy of mention, then we might as well add a one-sentence explanation. Seeing just a link seems a bit too open-ended.

KPrasch:
Putting this PD to the side for a moment - Not *every* PR needs a newsfragment to have content, some are too simple and small and good judgement can be used, drawing attention to what is the most essential parts to communicate.  Spreading the news is only one function of the towncrier system the other half being tracking which PRs were included in a release, which is valuable.


TLDR - If I make a spelling correction, newsfragment text takes away from meaningful news and adds overhead, nonetheless it's still valuable to track PRs into releases.

KPrasch:
Slipped onto my return key and closed this PR - reopening it.

derekpierre:
> TLDR - If I make a spelling correction, newsfragment text takes away from meaningful news and adds overhead, nonetheless it's still valuable to track PRs into releases.

IMHO a spelling correction shouldn't even have an entry in the release notes - as you said doesn't have any real release note value. I somewhat disagree with what web3py did - for example PR 1796 has a link in their release notes, and its PR title is "Remove unneeded pytestmark warnings" - probably not needed to be in release notes since it isn't really "newsworthy" for the audience. This is what Bogdan was getting at with the subtle difference between release notes and a chnagelog.

KPrasch:
> the difference between release notes and a chnagelog

Point well taken, but since we don't have a changelog mechanism or practice in place- I'd still prefer to track every PR that goes into a release, which newsfragments + GH action ensures seamlessly.

> the newsfragment action can fail and we'll merge anyway.

I cannot agree with you there - Green check marks have a positive psycological effect, represent the landmark of an acceptable PR and bring information density to history viewers.

KPrasch:
Updated to Include newsfragment content.

nucypher/cli/literature.py
derekpierre: Unresolving since the text hasn't been updated...

tests/acceptance/cli/ursula/test_federated_ursula.py
derekpierre: Unresolving, the text hasn't been updated...

newsfragments/2474.misc.rst
```diff
@@ -0,0 +1 @@
+Improves console messages for stakeholder CLI initialization and worker startup.
```
derekpierre, line 1: 👍

derekpierre (changes requested):
> I'd still prefer to track every PR that goes into a release, which newsfragments + GH action ensures seamlessly.

I hope so as well 😄 , as long as we still cater to the audience of the release notes - the audience is not ourselves.

> I cannot agree with you there - Green check marks have a positive psycological effect, represent the landmark of an acceptable PR and add information density to history viewers.

I hear you that the lack of a green check mark is not ideal. Unfortunately, the check is primitive and cannot determine context, so it is more of something akin to an opt-out related to newsfragments so that the dev can make the decision on enforcement. It informs the dev, *"BTW you don't have a newsfragment as part of this PR, have you thought about whether there should be one or not? If there should be one, add it, and If not, then ignore me"*. 

We do something similar with PR review approvals - the rule is 3 but depending on the PR 2 or even 1 approval is sufficient and we subsequently merge and ignore the requirement of 3.

KPrasch:
> We do something similar with PR review approvals - the rule is 3 but depending on the PR 2 or even 1 approval is sufficient and we subsequently merge and ignore the requirement of 3.

It's not really a fair or equal comparison - not having enough reviewers doesn't put a big red X on information viewers.  This changes the meaning of GH checks to a mere suggestion versus a requirement, effectively removing the finishline needed for a merge and increases the mental effort required to understand if a PR is acceptable.  I'm not willing to compromise on those, as they are essential parts of the development workflow and PR lifecycle.

KPrasch:
Regarding the newsfragment, I must have made a commit error. In any case, I still purport this PR as ready to merge without further changes.

derekpierre:
To clarify, my request for changes was only due to this text
```
CONFIRM_URSULA_IPV4_ADDRESS = "Detected IPv4 address ({rest_host}) - Is this is the public-facing address of Ursula?"
```
specifically "Is this is the...", - the second "is" should be removed which is in my original suggestion.

KPrasch:
> To clarify, my request for changes was only due to this text

Ah, yeah - that makes sense. I'll get right on it.  Thanks for the thorough review as usual ✊

derekpierre:
> This changes the meaning of GH checks to a mere suggestion versus a requirement...

I wasn't trying to indicate that this applies to all GH checks, just that while that is ideal, and it should be, in this specific newsfragment case, given that to make any change to the codebase we use PRs like your spelling example from earlier, viewing the newsfragment check as a "requirement" may be a bit too strong.

KPrasch:
*Requiring* newsfragments have been a major improvement from my view.  Perhaps we'll integrate another system that is more perfectly suited for granular audiences (changelog) but towncrier is doing an excellent job at tracking and documenting which PRs are included in a release, and I have zero qualms about using it in a slightly more changelog-ish way.  

What's the drawback - imperfect abstractions?

nucypher/cli/literature.py
KPrasch: Whoops! I accidentally forced pushed over this branch after merging suggestions without pulling.

derekpierre (approved):
🎸

derekpierre:
> towncrier is doing an excellent job at tracking and documenting which PRs are included in a release

Oh I've been loving it - towncrier is great; thanks for adding it.

> I have zero qualms about using it in a slightly more changelog-ish way. What's the drawback - imperfect abstractions?

The tricky part in my mind is catering to the audience. It's possible to flood a release note with entries that aren't of value to the audience (node providers / users), and drown out the entries that they do value. For example, how does it answer the question "*Why should I bother upgrading the nucypher version my node is running to this version?"*. We (NuCypher) are not the target audience. 

There was some prior discussion about additional release note categorization, perhaps that may be a compromise.

Issue #2473, opened 2020-12-11
cloudworkers: unexpected kwarg 'config_file'


KPrasch:
**Describe the Bug**

Unhandled exception encountered when adding a cloudworker (home office example)

**To Reproduce**
```
$ nucypher cloudworkers add_for_stake --staker-address 0xdeadbeef --host-address 192.168.1.xxx --login-name x --key-path ~/.ssh/id_rsa
```

**Traceback**
```
Traceback (most recent call last):
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/bin/nucypher", line 33, in <module>
    sys.exit(load_entry_point('nucypher', 'console_scripts', 'nucypher')())
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/k/.local/share/virtualenvs/nucypher-iNjwDRsw/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/k/Git/nucypher/nucypher/cli/options.py", line 167, in wrapper
    return func(**kwargs)
  File "/home/k/Git/nucypher/nucypher/cli/options.py", line 167, in wrapper
    return func(**kwargs)
  File "/home/k/Git/nucypher/nucypher/cli/options.py", line 167, in wrapper
    return func(**kwargs)
TypeError: add_for_stake() got an unexpected keyword argument 'config_file'

```

**System (please complete the following information):**
 - OS Version: Ubuntu 20.04
[...]

Issue #2472, opened 2020-12-11
Cloudworkers ENVAR support


vepkenez:
currently the following envvars are hardcoded:
```
          NUCYPHER_KEYRING_PASSWORD: "{{NUCYPHER_KEYRING_PASSWORD}}"
          NUCYPHER_WORKER_ETH_PASSWORD: "{{NUCYPHER_WORKER_ETH_PASSWORD}}"
          NUCYPHER_SENTRY_DSN: "{{SENTRY_DSN | default('')}}"
          NUCYPHER_SENTRY_LOGS: "{{SENTRY_LOGS | default('no')}}"
```
But with the advent of https://github.com/nucypher/nucypher/pull/2445
we may now need the operator to have the power to add arbitrary envvars such as `NUCYPHER_MAX_GAS_PRICE_GWEI`

So... I propose these envvars should be put in a dynamic list and accessed into the ansible config via `with_items`

PR #2475, opened 2020-12-12
Cloudworkers: bugfixes, better cleaner envvar support

vepkenez:
fixes 
* #2473
* #2472

### DIFF
--- a/nucypher/cli/commands/cloudworkers.py
+++ b/nucypher/cli/commands/cloudworkers.py
@@ -139,7 +139,7 @@ def add(general_config, host_address, login_name, key_path, ssh_port, host_nickn
     """Adds an existing node to the local config for future management."""
 
     emitter = setup_emitter(general_config)
-    name = f'{namespace}-{network}-{host_nickname}'
+    name = host_nickname
 
     deployer = CloudDeployers.get_deployer('generic')(emitter, None, None, namespace=namespace, network=network, action='add')
     config = deployer.create_nodes([name], host_address, login_name, key_path, ssh_port)
--- a/nucypher/cli/commands/cloudworkers.py
+++ b/nucypher/cli/commands/cloudworkers.py
@@ -157,7 +157,7 @@ def add(general_config, host_address, login_name, key_path, ssh_port, host_nickn
 @click.option('--ssh-port', help="The port this host's ssh daemon is listening on", default=22)
 @click.option('--namespace', help="Namespace for these operations.  Used to address hosts and data locally and name hosts on cloud platforms.", type=click.STRING, default='local-stakeholders')
 @group_general_config
-def add_for_stake(general_config, staker_address, host_address, login_name, key_path, ssh_port, namespace):
+def add_for_stake(general_config, staker_options, config_file, staker_address, host_address, login_name, key_path, ssh_port, namespace):
     """Sets an existing node as the host for the given staker address."""
 
     emitter = setup_emitter(general_config)
--- a/nucypher/cli/commands/cloudworkers.py
+++ b/nucypher/cli/commands/cloudworkers.py
@@ -53,13 +53,13 @@ def cloudworkers():
 @click.option('--remote-provider', help="The blockchain provider for the remote node, if not provided, nodes will run geth.", default=None)
 @click.option('--nucypher-image', help="The docker image containing the nucypher code to run on the remote nodes. (default is nucypher/nucypher:latest)", default=None)
 @click.option('--seed-network', help="Do you want the 1st node to be --lonely and act as a seed node for this network", default=False, is_flag=True)
-@click.option('--sentry-dsn', help="a sentry dsn for these workers (https://sentry.io/)", default=None)
 @click.option('--include-stakeholder', 'stakes', help="limit worker to specified stakeholder addresses", multiple=True)
 @click.option('--wipe', help="Clear nucypher configs on existing nodes and start a fresh node with new keys.", default=False, is_flag=True)
 @click.option('--prometheus', help="Run Prometheus on workers.", default=False, is_flag=True)
 @click.option('--namespace', help="Namespace for these operations.  Used to address hosts and data locally and name hosts on cloud platforms.", type=click.STRING, default='local-stakeholders')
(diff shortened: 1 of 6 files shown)


nucypher/cli/commands/cloudworkers.py
```diff
@@ -185,15 +184,15 @@ def add_for_stake(general_config, staker_options, config_file, staker_address, h
...
 @click.option('--seed-network', help="Do you want the 1st node to be --lonely and act as a seed node for this network", default=False, is_flag=True)
-@click.option('--sentry-dsn', help="a sentry dsn for these workers (https://sentry.io/)", default=None)
 @click.option('--wipe', help="Clear your nucypher config and start a fresh node with new keys", default=False, is_flag=True)
 @click.option('--prometheus', help="Run Prometheus on workers.", default=False, is_flag=True)
 @click.option('--namespace', help="Namespace for these operations.  Used to address hosts and data locally and name hosts on cloud platforms.", type=click.STRING, default='local-stakeholders')
 @click.option('--network', help="The Nucypher network name these hosts will run on.", type=click.STRING, default='mainnet')
 @click.option('--gas-strategy', help="Which gas strategy?  (glacial, slow, medium, fast)", type=click.STRING)
 @click.option('--include-host', 'include_hosts', help="specify hosts to update", multiple=True, type=click.STRING)
+@click.option('--env', '-e', 'envvars', help="environment variables (ENVVAR=VALUE)", multiple=True, type=click.STRING, default=[])
```
cygnusv, line 193: How multiple works here? What kind of separator does it expect?

cygnusv (approved):

nucypher/cli/commands/cloudworkers.py
vepkenez: It’s like

`—e var1=something —e var2=somethingelse`

vepkenez: Maybe I’ll add a doc for that

nucypher/utilities/clouddeploy.py
```diff
@@ -289,16 +292,50 @@ def chain_name(self):
...
-    def generate_ansible_inventory(self, node_names, **kwargs):
+    def update_generate_inventory(self, node_names, **kwargs):
+
+        # filter out the nodes we will not be dealing with
+        nodes = {key: value for key, value in self.config['instances'].items() if key in node_names}
+        if not nodes:
+            raise KeyError(f"No hosts matched the supplied names: {node_names}.  Try `nucypher cloudworkers list-hosts`")
+
+        default_envvars = [
+            ('NUCYPHER_KEYRING_PASSWORD',  self.config['keyringpassword']),
```
KPrasch, line 303: I think... there is an importable constant for these envvars

nucypher/utilities/clouddeploy.py
```diff
@@ -289,16 +292,50 @@ def chain_name(self):
...
+        # but overridden on a per node basis if previously specified
+        for key, node in nodes.items():
+            for k, v in default_envvars:
+                if not k in nodes[key]['runtime_envvars']:
+                    nodes[key]['runtime_envvars'][k] = v
 
         inventory_content = self._inventory_template.render(
             deployer=self,
-            nodes=[value for key, value in self.config['instances'].items() if key in node_names],
+            nodes=nodes.values(),
```
KPrasch, line 331: :+1:

nucypher/utilities/clouddeploy.py
```diff
@@ -338,7 +375,7 @@ def _inventory_template(self):
 
     def deploy_nucypher_on_existing_nodes(self, node_names, wipe_nucypher=False):
 
-        playbook = 'deploy/ansible/worker/setup_remote_workers.yml'
+        playbook = os.path.join(DEPLOY_DIR, 'ansible/worker/setup_remote_workers.yml')
```
KPrasch, line 378: use Pathlib for better system support :-)

nucypher/utilities/clouddeploy.py
```diff
@@ -289,16 +292,50 @@ def chain_name(self):
...
+                node_vars.update({k: v})
+            nodes[key]['runtime_envvars'] = node_vars
+
+            # we want to update the config with the specified envvars
+            # so they will persist in future invocations
+            self.config['instances'][key] = copy.deepcopy(nodes[key])
+
+        # we don't want to save the default_envvars to the config file
+        # but we do want them to be specified to the inventory template
+        # but overridden on a per node basis if previously specified
```
arjunhassard, line 323: A staker who adds their own envars might expect all others/defaults to be in the config file, especially if they are being used for something, so this should be really spelled out in documentation

arjunhassard (approved):
🪕🪕🪕  looking forward to being the production guinea node

nucypher/utilities/clouddeploy.py
```diff
@@ -381,10 +418,9 @@ def deploy_nucypher_on_existing_nodes(self, node_names, wipe_nucypher=False):
         self.update_captured_instance_data(self.output_capture)
         self.give_helpful_hints(node_names, backup=True, playbook=playbook)
 
-
     def update_nucypher_on_existing_nodes(self, node_names):
 
-        playbook = 'deploy/ansible/worker/update_remote_workers.yml'
+        playbook = os.path.join(DEPLOY_DIR, 'ansible/worker/update_remote_workers.yml')
```
derekpierre, line 423: Could use `pathlib.Path` here as well, and other spots with `os.path.join`.

nucypher/utilities/clouddeploy.py
vepkenez: this refers to the *worker* config file, not the NuCypher ursula.json

There are two envars that are _required_, specifically the keychain and worker passwords.
Since these are generally handled by default for all nodes in a namespace, we don't want to write them on a per node basis in the worker_config.json, since it will generally be the same thing for every node and also, the less times we write passwords into files the better imo.

But this bit of code does allow them to be overridden on a per host basis and _only then_ would they be written to the config file.

KPrasch (approved):

Issue #408, opened 2018-08-31
Handle Invalid Policy Enacment Response


KPrasch:

KPrasch:
@tuxxy - @jMyles - Is this related to #2391, or can this be understood as post-enactment.

KPrasch:
Closing in favor of #2391

PR #2482, opened 2020-12-18
Parallel arrangement proposals

### LINKED ISSUE #2479: More than Rejected - Granular exceptions for grant failure
After Alice samples Ursulas - `Rejected` is raised for *any reason* if `Policy._accepted_arrangements` empty.  This can be be the result of a more broad set of possible issues, making "Rejected" inaccurate.  Debugging and maintaining the network will be more sustainable with more granular exceptions for Down nodes (#2478), HTTP failures, Timeouts, and *actual* rejections during arrangement proposal, etc.

fjarri:
**Type of PR:**
- Bugfix
- Feature

**Required reviews:** 
- 3

**What this does:**

- It merges sampling and proposing arrangements in `Policy` logic, allowing for better control on problems (to fix #2479) and error recovery. In particular, we do not end up in a situation where we sampled `n` nodes, and one of them later rejected the proposal.
- It refactors the policy classes, streamlining logic and removing unused code
- It makes proposals and enactment parallel, and uses the same underlying worker pool class for that and treasure map distribution

- Changes the logic in `WeightedSampler`: now it does not reset the weights after each `draw()`. It did not matter before since we've only ever been doing one drawing per `WeightedSampler` object.
- `Policy.make_arrangements()` and `Policy.enact()` are merged into just `enact()`
- `Policy.enact()` returns an `EnactedPolicy` object, containing the treasure map, the treasure map publisher, and the revocation kit; the original `Policy` object is unchanged.
- `Policy.NotEnoughUrsulas`/`BlockchainPolicy.NotEnoughBlockchainUrsulas` are thrown if enactment fails at proposal stage (instead of `Rejected` as before); `Policy.EnactmentError` is thrown if the failure is at actual enactment stage. 
- `PolicyCredential` deleted completely. It was not used anywhere.
- adds a `test/acceptance/utilities/test_concurrency.py` (not put into unit tests because it takes ~16s)

TODO:

**Issues fixed/closed:**
- Fixes #2479

**Why it's needed:**
Part of https://github.com/nucypher/nucypher/projects/13

**Notes for reviewers:**

Specific things to think about:
- Is there sufficient logging in case of errors? 
- Try to recover if the enactment stage fails? Easy to do for federated policies, but for the blockchain ones we've already posted a transaction - should we attempt another one?
- Do we really need to leave `TreasureMapPublisher` (former `NodeEngagementMutex`) dangling after we called `block_until_success_is_reasonably_likely()`? If we're Alice, we'll have to wait for its completion anyway, so why not just do that explicitly?

There are a lot of files changed, but the main ones are:
- `nucypher/blockchain/eth/agents.py` for  `WeightedSampler` changes
- `nucypher/policy/policies.py` has the bulk of the changes
- `nucypher/utilities/concurrency.py` has the new `WorkerPool` class; tests are in `test/acceptance/utilities/test_concurrency.py`
- `nucypher/*` contains changes in places where `Policy`'s public API was used, including the change in some mocks in `unlawful.py`
- `test/*` - a lot of small changes, but basically just fixes for the changed API

`WorkerPool` can be used for parallel `Bob.retrieve()` as well. Although this should be a temporary measure (yes, there is nothing more constant than temporary, but we need to try); we really need a proper top-to-bottom async solution.


vepkenez:
This seems like a great approach.
I'm trying to figure out how many parallel requests will be made, is there some argument to trio to give it a limit?

Overall though, this should be fast and I guess in theory does it exhaust the reservoir before failing?

fjarri:
Actually 10 is the timeout, `self.n` is the required number of successes. Currently it stops when it either reaches `self.n` successes, or exhausts the reservoir, or when the timeout expires.

Although I'm thinking that it might be a good idea to include enactment in the worker as well, so things will become a little more complicated - because we (presumably) don't want to send the same kfrag to several Ursulas, so we won't be able to have more than `n` workers running. As soon as one fails and releases its kfrag, another one can be started.

Of course, we can delay the kfrag assignment before it's time to call `enact_policy()`. This way we can have more than `self.n` workers at the same time, but the workers will have to synchronize midway, and it will make the code more complicated. But if a significant number of nodes is expected to fail before `enact_policy()`, it will improve reliability.

vepkenez:
Ah, so the number of workers is always the same as n?
So... what is a likely upper bound for highest # of n?
And what is the system load for say... 40 concurrent workers?
The actual requests are blocking aren't they?

fjarri:
> Ah, so the number of workers is always the same as n?

Nope, currently it can be more than `n`. For example, the loop in `run_workers()` can go as:
- `value_factory()` returns a batch of `n` addresses (as determined by `PrefetchStrategy`)
- after `stagger_timeout`, in the next iteration, let's say only 1 succeeded
- `value_factory(1)` produces `n - 1` different addresses and launches another batch of workers
- now we have `2n - 1` workers

> So... what is a likely upper bound for highest # of n?

If you mean the bound on the number of workers, currently none (besides the size of the reservoir). Although I agree, it might be useful to introduce one.

> The actual requests are blocking aren't they?

That's a good point. By default `trio` tasks are run in a single thread, and it can switch tasks only at "checkpoints", which are, in the first approximation, any calls of `trio` functions (e.g. `trio.sleep()`). So for now it can only happen in the loop in `_try_propose_arrangement()` where we wait until the node is known. If we add a call to `enact()` there, a `trio.sleep()` can be put between `propose_arrangement()` and `enact()` to let `trio` take control and abort if necessary. 

If that's not enough, we can launch tasks in separate threads, or use async requests library. 

> And what is the system load for say... 40 concurrent workers?

So, to answer that, if we're running tasks in the default mode (same thread), then 40 concurrent workers won't really be any different than just running stuff sequentially. If we put them in threads, that may be slow (at least on the weak cloud VMs).

vepkenez:
It might be very high performance to build this using an asynchronous callback style pattern, like with this library:  https://github.com/twisted/treq

So all of the actual requests could be concurrent and non blocking... 
I think this would speed up the grant significantly.

fjarri:
Oh no, the whole point of `trio` is avoiding callback hell. So the async requests library I had in mind would be something like https://github.com/theelous3/asks

fjarri:
Damn, after adding enactment to the workers I realized that for a blockchain policy Ursulas need to know that the policy has been published, so the list of target Ursulas needs to be established before that.

derekpierre (commented):
Any reason for the empty newsfragment? Parallelization, and better error handling/recovery seem pertinent to users.

nucypher/policy/policies.py
```diff
@@ -729,120 +614,72 @@ def generate_policy_parameters(n: int,
...
+                                           arrangements=arrangements,
+                                           publish_treasure_map=publish_treasure_map,
+                                           publication_transaction=transaction)
 
-            # Not in love with this block here, but I want 121 closed.
-            for arrangement in self._accepted_arrangements:
-                arrangement.publish_transaction = self.publish_transaction
 
-        publisher = super().enact(network_middleware, publish_treasure_map=False)
+class EnactedPolicy:
```
KPrasch, line 661: Trying to familiarize with the new object introduced in this PR.  Enacted policies can be federated or decentralized?

fjarri: They can be both, both `FederatedPolicy.enact()` and `BlockchainPolicy.enact()` return an `EnactedPolicy` (well, it's `Policy.enact()` for both). The contents of `EnactedPolicy` are essentially based on how the policy object was used after `enact()`. It is possible they can be grouped differently.

tests/integration/network/test_failure_modes.py
```diff
@@ -69,7 +69,7 @@ def test_alice_can_grant_even_when_the_first_nodes_she_tries_are_down(federated_
 
     for node in more_nodes:
         federated_alice.remember_node(node)
-    with pytest.raises(Policy.Rejected):
+    with pytest.raises(Policy.NotEnoughUrsulas):
```
KPrasch, line 72: :+1:

nucypher/policy/policies.py
```diff
@@ -407,288 +241,339 @@ def hrac(self) -> bytes:
...
+
+        value_factory = AllAtOnceFactory(list(zip(arrangements, self.kfrags)))
+        worker_pool = WorkerPool(worker=worker,
+                                 value_factory=value_factory,
+                                 target_successes=self.n,
+                                 timeout=timeout,
+                                 threadpool_size=self.n)
+
+        worker_pool.start()
+        worker_pool.join()
```
KPrasch, line 385: Perhaps a comment here indicating a blocking call.

nucypher/policy/policies.py
```diff
@@ -407,288 +241,339 @@ def hrac(self) -> bytes:
...
-            if publish_treasure_map is True:
-                return self.publish_treasure_map(network_middleware=network_middleware)  # TODO: blockchain_signer?
+    def _make_publisher(self,
+                        treasure_map: 'TreasureMap',
+                        network_middleware: RestMiddleware,
+                        ) -> TreasureMapPublisher:
 
-    def propose_arrangement(self, network_middleware, arrangement) -> bool:
-        negotiation_response = network_middleware.propose_arrangement(arrangement=arrangement)  # Wow, we aren't even passing node here.
+        # TODO: remove hardcoding of 8 nodes
```
KPrasch, line 439: Needs a follow-up Issue to trackk

KPrasch (commented):
Still reading - nice rework!  I think this PR warrants some content in it's newsfragment(s).

KPrasch (commented):
Considering enactment was modified in this PR - is this (low number!) issue in scope enough be linked here?

https://github.com/nucypher/nucypher/issues/408

examples/finnegans_wake_demo/finnegans-wake-demo.py
```diff
@@ -102,7 +102,7 @@
                      expiration=policy_end_datetime)
 
 assert policy.public_key == policy_pubkey
-policy.publishing_mutex.block_until_complete()
+policy.treasure_map_publisher.block_until_complete()
```
vepkenez, line 105: This name change from "publishing_mutex" to "treasure_map_publisher" is such a huge improvement by itself.  I already love this PR.
💯

KPrasch (approved):
Excellent work here @fjarri - I have been using this PR to grant live policies and the improvement in performance and reliability is abundantly clear.  If there isn't already, let's make sure there is a follow-up issue pertaining to the ongoing discussion about async libraries and long-term solutions.

nucypher/policy/policies.py
```diff
@@ -46,356 +47,189 @@
...
-            except Empty:
-                raise RuntimeError(f"Timed out.  Nodes completed: {self.completed}")
-            self.log.debug(f"{len(self.completed)} nodes were contacted while blocking for a little while.")
-            return completed_for_reasonable_likelihood_of_success
-        else:
-            return self.completed
-
+        self._worker_pool.block_until_target_successes()
+        completed = self.completed
+        self.log.debug(f"{len(completed)} nodes were contacted while blocking for a little while.")
```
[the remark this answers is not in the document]
fjarri: Fixed. This was a remainder from `NodeEngagementMutex`. Regrettably, I'll have to remove the youtube link too.

nucypher/policy/policies.py
```diff
@@ -407,288 +241,339 @@ def hrac(self) -> bytes:
...
-                                   *args, **kwargs)
+        return enacted_policy
 
-        if len(self._accepted_arrangements) < self.n:
-            formatted_offenders = '\n'.join(f'{u.checksum_address}@{u.rest_url()}' for u in sampled_ursulas)
-            raise self.Rejected(f'Selected Ursulas rejected too many arrangements'
-                                f'- only {len(self._accepted_arrangements)} of {self.n} accepted.\n'
-                                f'Offending nodes: \n{formatted_offenders}\n')
+    @abstractmethod
+    def _blockchain_signer(self) -> Optional[Callable[[bytes], bytes]]:
```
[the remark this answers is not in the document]
fjarri: `FederatedPolicy` doesn't have any signers, so this function either returns an actual blockchain signer, or nothing.

nucypher/utilities/concurrency.py
```diff
@@ -0,0 +1,301 @@
...
+            except BaseException as e:
+                self._unexpected_error.set(e)
+                self.cancel()
+                break
+
+        self._result_queue.put(PRODUCER_STOPPED)
+
+
+class AllAtOnceFactory:
+
```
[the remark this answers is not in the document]
fjarri: Fixed.

nucypher/policy/policies.py
fjarri: Fixed.

nucypher/policy/policies.py
fjarri: Filed #2516, added to TODO

fjarri:
@KPrasch , regarding #408 - I don't think it is addressed by this PR, the exceptions thrown remained essentially the same. I think that issue needs to be clarified, or closed.

nucypher/policy/policies.py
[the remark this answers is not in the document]
derekpierre: I understand the need and logic for the abstract signer method - but I was referring to the name of the method. Having `blockchain` in the method name, causes `FederatedPolicy._blockchain_signer()` to be defined, which is an oxymoron.

Something other than `blockchain` which is the direct antithesis of `Federated`, would be better, like `_signer()`, or `_transaction_signer()` (since you use `transaction` as a term in other areas of `Policy` eg. `_max_enactment_payload()` and `_enact_arrangements()`).

vepkenez (approved):

nucypher/policy/policies.py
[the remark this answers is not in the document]
fjarri: After some thought, I removed `_blockchain_signer()` completely and instead add the signature in the overridden `BlockchainPolicy._make_treasure_map()`. This seems more logical.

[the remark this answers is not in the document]
derekpierre: 👍 - even better

derekpierre (approved):
🎸 - Approved pending resolution of failing CI build.

fjarri:
Should be fixed by #2521

PR #2486, opened 2020-12-19
Estimate gas with block identifier 'latest'

svenski123:
Running `nucypher worklock refund` fails with "no refund availble" if a pending refund transaction is present in the mempool.
During gas estimation, the pending refund transaction is effectively assumed to have already completed thus causing the attempted refund transaction to fail when estimating gas.

This PR explicitly estimates gas with block identifier set to "latest" which ignores the pending transactions for purposes of gas estimation.

The PR also fixes failed transaction exception handling, where the resulting ValueError had a third unexpected "data" member causing the unpack and assign to two variables statement to fail and throw.

This PR is a bug fix and fairly small in scope - review should hopefully be easy and straightforward.

### DIFF
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -404,30 +404,33 @@ class BlockchainInterface:
         """
 
         try:
-            # Assume this error is formatted as an IPC response
             response = exception.args[0]
-            code = response['code']
-            message = response['message']
+        except (AttributeError, TypeError):
+            # Python exceptions must have the 'args' attribute which must be a sequence (i.e. indexable)
+            raise ValueError(f'{exception} is not a valid Exception instance')
 
-        except (ValueError, IndexError, AttributeError, KeyError, TypeError):
+        # Assume this error is formatted as an RPC response
+        try:
+            code = int(response['code'])
+            message = response['message']
+        except (KeyError, ValueError):
             # TODO: #1504 - Try even harder to determine if this is insufficient funds causing the issue,
             #               This may be best handled at the agent or actor layer for registry and token interactions.
             # Worst case scenario - raise the exception held in context implicitly
             raise exception
 
-        else:
-            if int(code) != cls.TransactionFailed.IPC_CODE:
-                # Only handle client-specific exceptions
-                # https://www.jsonrpc.org/specification Section 5.1
-                raise exception
+        if code != cls.TransactionFailed.IPC_CODE:
+            # Only handle client-specific exceptions
+            # https://www.jsonrpc.org/specification Section 5.1
+            raise exception
 
-            if logger:
-                logger.critical(message)  # simple context
+        if logger:
+            logger.critical(message)  # simple context
 
-            transaction_failed = cls.TransactionFailed(message=message,  # rich error (best case)
-                                                       contract_function=contract_function,
-                                                       transaction_dict=transaction_dict)
-            raise transaction_failed from exception
+        transaction_failed = cls.TransactionFailed(message=message,  # rich error (best case)
+                                                   contract_function=contract_function,
+                                                   transaction_dict=transaction_dict)
+        raise transaction_failed from exception
 
     def __log_transaction(self, transaction_dict: dict, contract_function: ContractFunction):
         """
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -405,7 +405,9 @@ class BlockchainInterface:
 
         try:
             # Assume this error is formatted as an IPC response
-            code, message = exception.args[0].values()
+            response = exception.args[0]
+            code = response['code']
+            message = response['message']
 
         except (ValueError, IndexError, AttributeError):
             # TODO: #1504 - Try even harder to determine if this is insufficient funds causing the issue,
@@ -496,7 +498,12 @@ class BlockchainInterface:
                                      use_pending_nonce=use_pending_nonce)
         self.__log_transaction(transaction_dict=payload, contract_function=contract_function)
         try:
-            transaction_dict = contract_function.buildTransaction(payload)  # Gas estimation occurs here
+            if 'gas' not in payload:
+                # As web3 buildTransaction() will estimate gas with block identifier "pending" by default,
+                # explicitly estimate gas here with block identifier 'latest' if not otherwise specified
+                # as a pending transaction can cause gas estimation to fail, notably in case of worklock refunds.
+                payload['gas'] = contract_function.estimateGas(payload, 'latest')
+            transaction_dict = contract_function.buildTransaction(payload)
         except (TestTransactionFailed, ValidationError, ValueError) as error:
             # Note: Geth raises ValueError in the same condition that pyevm raises ValidationError here.
             # Treat this condition as "Transaction Failed" during gas estimation.
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -409,7 +409,7 @@ class BlockchainInterface:
             code = response['code']
             message = response['message']
 
-        except (ValueError, IndexError, AttributeError):
+        except (ValueError, IndexError, AttributeError, KeyError, TypeError):
             # TODO: #1504 - Try even harder to determine if this is insufficient funds causing the issue,
             #               This may be best handled at the agent or actor layer for registry and token interactions.
             # Worst case scenario - raise the exception held in context implicitly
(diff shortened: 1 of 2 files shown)


KPrasch:
Thanks for the PR @svenski123 - We'll take a look soon :-)

svenski123:
This fixed the "worklock refund when refund transaction already pending" problem for me.

I think it also causes the <50k out of gas commit to next period transaction failures - when it ups the gas and resubmits the commit transaction, one commit transaction is already pending thus the gas estimate is for a no-op effectively (i.e. running commitToNextPeriod() when it has already run for a period and thus there's nothing to do - so less gas needed).

nucypher/blockchain/eth/interfaces.py
```diff
@@ -478,7 +480,12 @@ def build_contract_transaction(self,
                                      use_pending_nonce=use_pending_nonce)
         self.__log_transaction(transaction_dict=payload, contract_function=contract_function)
         try:
-            transaction_dict = contract_function.buildTransaction(payload)  # Gas estimation occurs here
+            if 'gas' not in payload:
+                # As web3 buildTransaction() will estimate gas with block identifier "pending" by default,
+                # explicitly estimate gas here with block identifier 'latest' if not otherwise specified
+                # as a pending transaction can cause gas estimation to fail, notably in case of worklock refunds.
```
KPrasch, line 486: More notably from worker commitment transactions in mempool.

nucypher/blockchain/eth/interfaces.py
```diff
@@ -478,7 +480,12 @@ def build_contract_transaction(self,
...
+                payload['gas'] = contract_function.estimateGas(payload, 'latest')
```
KPrasch, line 487: :+1:

nucypher/blockchain/eth/interfaces.py
```diff
@@ -387,7 +387,9 @@ def _handle_failed_transaction(cls,
 
         try:
             # Assume this error is formatted as an IPC response
-            code, message = exception.args[0].values()
+            response = exception.args[0]
+            code = response['code']
+            message = response['message']
 
         except (ValueError, IndexError, AttributeError):
```
KPrasch, line 394: This exception handling will need to be updated to consider the case of a `KeyError` (or reworked to exclude) and remove the `ValueError` since this block will no longer do tuple unpacking. If not in this PR than in a follow-up is acceptable, but with a TODO and/or Issue.

svenski123: Motivation for this was exception-during-exception-handler observed when gas estimation failed with geth.

Previous code clearly assumed passed in exception was constructed with a non-empty parameter list of which the first item is a dict or dict-like containing two items the values of which, when iterated over, make sense as 'code' and 'message' in that order.

_handle_failed_transaction() is only called from one site, in build_contract_transaction() in an exception handler surrounding a call to buildTransaction() where it is intended to return an exception that can be raised (i.e. raise _handle_failed_transaction(...)), however _handle_failed_transaction() does not return a value - it always raises an exception itself.

If the intention is for all exceptions during a transaction building to appear as TransactionFailed then the original exception and traceback should be captured and embedded in the TransactionFailed object rather than throwing from the exception handler, along with whatever ad-hoc efforts to extract an error message, etc. from the original exception.

svenski123: For this PR, I've added KeyError and TypeError to the except clause in _handle_failed_transaction().
Arguably this should be a blanket exception clause, as it should just rethrow the original error if it can't decode it.

nucypher/blockchain/eth/interfaces.py
svenski123: This works with recent geth versions but it's worth checking whether estimateGas() with a block identifier parameter works when using another node (old geth, parity, infura) or if it throws (as suggested by @cygnusv I believe).

nucypher/blockchain/eth/interfaces.py
KPrasch: I actually think we're lost meaningful context here by expanding number of handled exceptions.  

- Consider that `Exception.args` is a specification of python exceptions.  If we receive such an object via `AttributeError`, it's an invalid exception.
- A `ValueError` or `TypeError` can no longer caused be explicitly raised in this block and can be removed.  If an implicit exception *does* occur here, it represents an unhanded error can bubble up and raise. 
- `KeyError` indicates that the exception being passed is not formatted as an eth RPC response, and is missing fields.

In order to preserve this context, and eliminate superfluous exception handling I recommend using this approach:

```python
# Extract the exception's message or data.
try:
    response = exception.args[0]
except AttributeError:
    # Python exceptions must have the 'args attribute'.
    raise ValueError(f'{exception} is not a valid Exception instance.')
except IndexError:
    # No exception data or message available.
    raise exception

# Assume this error is formatted as a valid eth RPC response.
try:
    code = response['code']
    message = response['message']
except KeyError:
    # Response data does not contain valid RPC fields.
    raise exception
```

KPrasch: Here's how this handler looks in action:  

```
  File "/home/kieran/Git/nucypher/nucypher/cli/commands/stake.py", line 410, in bond_worker
    receipt = STAKEHOLDER.bond_worker(worker_address=worker_address)
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/decorators.py", line 116, in wrapped
    return func(actor, *args, **kwargs)
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/decorators.py", line 124, in wrapped
    receipt_or_txhash = actor_method(self, *args, **kwargs)
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/decorators.py", line 105, in wrapped
    return func(*args, **kwargs)
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/actors.py", line 1341, in bond_worker
    receipt = self.staking_agent.bond_worker(staker_address=self.checksum_address,
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/decorators.py", line 105, in wrapped
    return func(*args, **kwargs)
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/agents.py", line 569, in bond_worker
    receipt: TxReceipt = self.blockchain.send_transaction(contract_function=contract_function,
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/decorators.py", line 105, in wrapped
    return func(*args, **kwargs)
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/interfaces.py", line 630, in send_transaction
    transaction = self.build_contract_transaction(contract_function=contract_function,
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/decorators.py", line 105, in wrapped
    return func(*args, **kwargs)
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/interfaces.py", line 503, in build_contract_transaction
    raise self._handle_failed_transaction(exception=error,
  File "/home/kieran/Git/nucypher/nucypher/blockchain/eth/interfaces.py", line 428, in _handle_failed_transaction
    raise transaction_failed from exception
nucypher.blockchain.eth.interfaces.TransactionFailed: BONDWORKER from 0x37f3... 
Sender balance: 0 ETH 
Reason: execution reverted 
Transaction: {'chainId': 4, 'nonce': 0, 'from': '0x37f360567b6C4dF121302DaED8A9B7039FE09Deb'}
```

nucypher/blockchain/eth/interfaces.py
KPrasch: Indeed it does appear that Infura supports the `eth_estimateGas` endpoint (@cygnusv) - https://infura.io/docs/ethereum#operation/eth_estimateGas

nucypher/blockchain/eth/interfaces.py
KPrasch: bump @svenski123 (Happy New Year!) - Let's either integrate the changes mentioned above, or replace the PR with another one I can open based on your changes: <https://github.com/nucypher/nucypher/pull/2486#discussion_r548106237>

svenski123: Hi @KPrasch!
Happy belated new year to you as well!
I've rebased and force pushed the branch against main as well as refactored the exception decoding slightly.
In the spirit of paranoia, exception.args[0] can throw a TypeError if .args cannot be indexed, and int(code) where code = exception.args[0]['code'] can throw a ValueError if ...['code'] can't be cast to an int.

KPrasch (approved):
:cowboy_hat_face: - successfully staged mainnet nodes.

Thanks for your contribution!

PR #2483, opened 2020-12-19
Prevent Vladimir from creating a ":memory:" directory

fjarri:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

**What this does:**
Prevents `Vladimir` from littering in the test folder with its temporary database.

Issue #2330, opened 2020-10-02
:memory: lmdb directory left on disk after tests


KPrasch:

fjarri:
Fixed by PR #2483

PR #2491, opened 2021-01-03
In-memory mock LMDB for tests

### LINKED ISSUE #2490: Test Ursulas sharing LMDB files leads to problems
At the moment federated Ursula fleets created for testing (by `make_federated_ursulas()`) receive the same path for their datastore (and, in fact, it is reused between tests as well). 

As [`lmdb` docs](https://lmdb.readthedocs.io/en/release/#environment-class) state,

> It is a serious error to have open the same LMDB file in the same process at the same time. Failure to heed this may lead to data corruption and interpreter crash.

So far it's been working for us, but only because datastore was accessed from one thread. With recent efforts to parallelize various requests, it may no longer be the case, and results in various errors - for example, the record simply not being saved into the database, or an interpreter crash.

Also, sharing the database between Ursulas *and* tests may lead to subtle bugs (when records created in different Ursulas or tests interfere), or to missing some bugs (e.g. when we erroneously create a record only in one Ursula instead of many, but it is "replicated" to all of them). 

In view of all this, it may be beneficial to:
- (short term solution) use separate DB paths for Ursulas in `make_federated_ursulas()` (an exception will have to be made for `fleet_of_highperf_mocked_ursulas()`, since we can't have 5000 handles open at the same time). It will lead to more `/tmp` cluttering and some test performance degradation.
- (long term solution) mock LMDB in all places where it is not directly tested by replacing it with an in-memory store.

fjarri:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Replaces LMDB storage with an in-memory mock in most tests.

**Issues fixed/closed:**
Fixes #2490

**Notes for reviewers:**
- Which tests should still use the real LMDB? For now it's just `tests/unit/datastore` and `tests/integration/config/test_storages`
- Should we try to emulate correct behavior for the cases of DB being modified with multiple active cursors? Current mock version does not support DB writes while cursors are active, but that's rather hard to enforce.


tests/unit/datastore/test_datastore.py
```diff
@@ -32,13 +32,18 @@ class TestRecord(DatastoreRecord):
...
-def test_datastore_describe():
+def test_datastore_create():
     temp_path = tempfile.mkdtemp()
     storage = datastore.Datastore(temp_path)
     assert storage.LMDB_MAP_SIZE == 1_000_000_000_000
     assert storage.db_path == temp_path
     assert storage._Datastore__db_env.path() == temp_path
 
+
+def test_datastore_describe(mock_or_real_datastore):
```
cygnusv, line 43: What's the rationale for running a datastore unit test over a mocked datastore?

fjarri: To verify that all our assumptions about how LMDB work are valid for the mock as well. This is, in fact, how I debugged the mock in the first place.

cygnusv: Ah I see...makes sense!

cygnusv (approved):

tests/conftest.py
```diff
@@ -199,3 +201,9 @@ def patch_card_directory(session_mocker):
                                 new_callable=session_mocker.PropertyMock)
     yield
     tmpdir.cleanup()
+
+
+@pytest.fixture(scope='session', autouse=True)
+def mock_datastore(monkeysession):
```
KPrasch, line 207: Strangely like *Hammer Time*... :thinking:

tests/integration/conftest.py
```diff
@@ -55,42 +53,6 @@
 from tests.utils.ursula import MOCK_URSULA_STARTING_PORT
 
 
-class TestLMDBEnv:
```
KPrasch, line 58: :+1:

tests/mock/datastore.py
```diff
@@ -0,0 +1,128 @@
...
+        self._lock = Lock()
+
+    @contextmanager
+    def begin(self, write=False):
+        with self._lock:
+            with MockTransaction(self, write=write) as tx:
+                yield tx
+
+
+class MockTransaction:
```
KPrasch, line 46: Consider:

```suggestion
class MockDatabaseTransaction
```

tests/unit/datastore/test_datastore.py
KPrasch: > To verify that all our assumptions about how LMDB works are valid for the mock as well.

A good principal for unit tests, generally.

cygnusv:
@fjarri Do you think https://github.com/nucypher/nucypher/issues/2330 can be added to the scope of this PR (or a subsequent one)? Now that you're getting along with LMDB...

tuxxy (approved):

derekpierre (approved):
🎸 Feels like a "Misc" instead of a "Bugfix" to me - but good to go either way.

tests/mock/datastore.py
fjarri: I named the classes based on the original ones from `lmdb` - `Environment`, `Transaction` and `Cursor`. I think there's no harm in leaving it like this.

fjarri:
@cygnusv , I think I actually fixed it in PR #2483 . Didn't notice the open issue, sorry.

fjarri:
@derekpierre , there is an open issue that this PR closes, but it's a test bug, not a real bug, so perhaps it doesn't count.

fjarri:
So I take it that everyone is fine with basically only running real database tests in a few places and replace everything else with mocks?

I'm also worried about the shortcut I'm currently taking with regards to cursors (assuming that the DB is not modified while a cursor is used). I would like to add enforce it somehow, in case we ever use this functionality, but I can't think of a way.

PR #2507, opened 2021-01-07
Deprecates manual worker commitment via CLI

KPrasch:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other (Deprecation)

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

Deprecates CLI commands for manually committing to next period.

**Why it's needed:**

Configurable maximum gas price for workers provides a mitigation that supersedes the ability to manually commit.  While it has served it's purpose, Exposing `commit_to_next_period` as a CLI command is a proverbial "foot gun" that allows users to damage the health of the network. It's removal is part of guiding the codebase and node operators to use other tactics to manage gas costs while still providing a high level of grant availability.

Tangentially related to project https://github.com/nucypher/nucypher/projects/13

### DIFF
--- a/nucypher/cli/processes.py
+++ b/nucypher/cli/processes.py
@@ -47,7 +47,7 @@ class UrsulaCommandProtocol(LineReceiver):
         # Expose Ursula functional entry points
         self.__commands = {
 
-             # Help
+            # Help
             '?': self.paintHelp,
             'help': self.paintHelp,
 
@@ -55,9 +55,6 @@ class UrsulaCommandProtocol(LineReceiver):
             'status': self.paintStatus,
             'known_nodes': self.paintKnownNodes,
             'fleet_state': self.paintFleetState,
-
-            # Blockchain Control
-            'commit_next': self.commit_to_next_period,  # hidden
 
             # Learning Control
             'cycle_teacher': self.cycle_teacher,
@@ -69,6 +66,8 @@ class UrsulaCommandProtocol(LineReceiver):
 
         }
 
+        self._hidden_commands = ('?',)
+
     @property
     def commands(self):
         return self.__commands.keys()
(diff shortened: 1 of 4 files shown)


vzotova:
I suppose we want to remove it from `UrsulaCommandProtocol` too

newsfragments/2507.misc.rst
```diff
@@ -0,0 +1 @@
+Deprecated manual worker commitments using the CLI.
```
derekpierre, line 1: This could be a good candidate for the "removal" category for newsfragments that applies to deprecations and removals - see "pyproject.toml" i.e. `2507.removal.rst`

KPrasch: :heavy_check_mark:

KPrasch:
> I suppose we want to remove it from UrsulaCommandProtocol too

Done.

fjarri:
Was there also a way to turn off automatic commits, or was this option present to force the automatic commit to happen at a specific time? I'm asking because in the former case we need to remove that option as well.

nucypher/cli/processes.py
```diff
@@ -56,9 +56,6 @@ def __init__(self, ursula, emitter):
             'known_nodes': self.paintKnownNodes,
             'fleet_state': self.paintFleetState,
 
-            # Blockchain Control
-            'commit_next': self.commit_to_next_period,  # hidden
-
```
vzotova, line 61: Another mention of this command on line 82

KPrasch: Ah - missed it! :heavy_check_mark:

KPrasch:
> Was there also a way to turn off automatic commits, or was this option present to force the automatic commit to happen at a   specific time? I'm asking because in the former case we need to remove that option as well.

I believe these are the only two places that exposed this functionality in a public API (you can still use the contract / agents internal APIs however).  The original motive here is certainly fading away, but it this command was to make an instantaneous one-time commit without startup.

vzotova (approved):

cygnusv (approved):

Issue #2349, opened 2020-10-12
Another unhandled background error.


jMyles:
```
Unhandled error during node learning.  Attempting graceful crash.
Unhandled Error
Traceback (most recent call last):
  File "/home/jmyles/.local/share/virtualenvs/nucypher-oH13o4TM/lib/python3.8/site-packages/twisted/internet/base.py", line 1283, in run
    self.mainLoop()
  File "/home/jmyles/.local/share/virtualenvs/nucypher-oH13o4TM/lib/python3.8/site-packages/twisted/internet/base.py", line 1292, in mainLoop
    self.runUntilCurrent()
--- <exception caught here> ---
  File "/home/jmyles/.local/share/virtualenvs/nucypher-oH13o4TM/lib/python3.8/site-packages/twisted/internet/base.py", line 886, in runUntilCurrent
    f(*a, **kw)
  File "/home/jmyles/projects/nucypher/nucypher/nucypher/network/nodes.py", line 502, in _crash_gracefully
    failure.raiseException()
  File "/home/jmyles/.local/share/virtualenvs/nucypher-oH13o4TM/lib/python3.8/site-packages/twisted/python/failure.py", line 488, in raiseException
    raise self.value.with_traceback(self.tb)
  File "/home/jmyles/.local/share/virtualenvs/nucypher-oH13o4TM/lib/python3.8/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
    current.result = callback(current.result, *args, **kw)
  File "/home/jmyles/projects/nucypher/nucypher/nucypher/network/nodes.py", line 562, in _discover_or_abort
    result = self.learn_from_teacher_node(eager=False, canceller=self._discovery_canceller)
  File "/home/jmyles/projects/nucypher/nucypher/nucypher/network/nodes.py", line 805, in learn_from_teacher_node
    except current_teacher.InvalidNode as e:
  File "/home/jmyles/projects/nucypher/bytestringSplitter/bytestring_splitter/__init__.py", line 77, in __getattr__
    raise AttributeError(
builtins.AttributeError: <class 'nucypher.network.nodes.NodeSprout'> doesn't have a InvalidNode, and it's not a partially split object either; those are ['domain', 'interface_signature', 'decentralized_identity_evidence', 'verifying_key', 'encrypting_key', 'certificate', 'rest_interface']
```

KPrasch:
This issue has not been seen or reproduced recently.  Closing.

PR #2513, opened 2021-01-11
[WIP] Buckets

### LINKED ISSUE #567: Buckets.  We need Buckets.

### LINKED ISSUE #2478: Limit sampling resivoir by node labels
Related to #2311 

Compare the HTTP request's origin IP address when verifying ursula interfaces. Currently node validity only considers the serialized representation of ursula matches the signatures recovered address. This leaves "ghost ursulas" on the network: metadata belonging to a node whose IP address changed and host was relocated.

This reduces the availability of grant since sampled nodes must be available to receive key fragments, and have an up-to-date IP addresses to receive grant requests.

KPrasch:
**Type of PR:**

- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 

- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

*This PR is a WIP and these API changes are subject to changing further*

- Implements sorting mechanism for known nodes and exposes queries for labeled nodes over HTTP endpoints.
  - Exposes node buckets via GET request argument for node label on `/node_information` route
  - Exposes node buckets/labels on node HTTP status endpoints
  - `Alice` only samples nodes marked as verified
- Introduces public node labeling API (`FleetSensor.label` & `FleetSensor.unlabel`)
  - Introduces node buckets: Unverified, Verified, Unavailable, Suspicious, Unstaked, Invalid
- Introduces configurable node pruning strategies `acumen.pruning` (can use a better name)
  - `accept_node`
  - `reject_node`
  - `construct_node_stalecheck`
  - `construct_node_max_attempts`
- Encapsulates node verification and automated labeling logic in `Ursula.verify_and_sort`
- Uses a new looping call `Ursula._node_pruning_task` for periodic automated node pruning at a the character level.

**Issues fixed/closed:**

Closes #567 
Closes #2478 
Closes https://github.com/nucypher/nucypher/issues/2349
Possibly related to #2352 
Follow-up for #2515 

**Why it's needed:**

- Advances https://github.com/nucypher/nucypher/projects/13
- Provides a foundation for more granular control of node discovery
- Allows invalid or otherwise useless nodes to eventually become forgotten
- Mitigates a potential DOS attack surface
- Increases grant availability:

(Alice) Here is an example of granting without node labels/buckets:

```
2021-01-04T22:09:02-0800 [nucypher.policy.policies.BlockchainPolicy#debug] Could not find enough Ursulas to accept proposals.
Accepted: 0xf00f, oxdeadbeef
Rejected:
0xf00f: HTTPSConnectionPool(host='1.2.3.4', port=9151): Max retries exceeded with url: /public_information (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: self signed certificate (_ssl.c:1123)')))
0xf00f: HTTPSConnectionPool(host='1.2.3.4', port=9151): Max retries exceeded with url: /public_information (Caused by ConnectTimeoutError(<urllib3.connection.HTTPSConnection object at 0x7f0f99194df0>, 'Connection to 97.115.107.84 timed out. (connect timeout=2)'))
```

**Notes for reviewers:**

Co-Authored-By: @vepkenez, @derekpierre


nucypher/acumen/perception.py
```diff
@@ -218,41 +218,91 @@ def abridged_state_details(state):
...
+        # Remove one label
+        if label:
+            if node in self.__marked[label]:
+                self.__marked[label].remove(node)
+                return
+
+        # Remove all labels
+        for _label in self.BUCKETS:
+            if node in self.__marked[_label]:
+                self.__marked[_label].remove(node)
```
derekpierre, line 281: A node can't be labelled more than once, so can we short-circuit here and return after removing the node?

KPrasch: We can... although it might be nice to have this escape hatch in case a node does somehow end up with two labels?  Am I too paranoid :sweat_smile: ?

nucypher/network/server.py
```diff
@@ -189,14 +192,12 @@ def node_metadata_exchange():
...
-        for node in sprouts:
-            this_node.remember_node(node)
-
-        # TODO: What's the right status code here?  202?  Different if we already knew about the node(s)?
-        return all_known_nodes()
+        else:
+            sprouts = _node_class.batch_from_bytes(request.data)
+            for node in sprouts:
+                this_node.remember_node(node)
+            # TODO: What's the right status code here?  202?  Different if we already knew about the node(s)?
```
fjarri, line 199: This TODO looks completely out of place. Did it somehow migrate from `server.py`?

nucypher/network/nodes.py
```diff
@@ -390,51 +501,21 @@ def remember_node(self,
...
 
-        self.known_nodes[node.checksum_address] = node
-
-        if self.save_metadata:
-            self.node_storage.store_node_metadata(node=node)
+        self._remember_essential(node=node)
+        if not node.verified_node:
+            # set UNVERIFED bucket as default, but don't relabel
+            # already verified nodes as unverified.
+            self.known_nodes.label(node=node, label=UNVERIFIED)
```
KPrasch, line 508: Note to self: this logic is incorrect, instead check for the existence of any label, don't force `UNVERIFED` implicitly.

nucypher/acumen/perception.py
derekpierre: Additional tests may help reduce to that level of paranoia 😏, but if we keep the "escape hatch" then we should log a warning so that we are at least aware that multi-labelling is occurring, and can subsequently investigate why that is.

nucypher/network/nodes.py
derekpierre: ✔️

nucypher/acumen/perception.py
derekpierre: Logged a warning for now.

nucypher/characters/lawful.py
```diff
@@ -1106,10 +1113,13 @@ def __init__(self,
             self._availability_tracker = AvailabilityTracker(ursula=self)
 
             # Datastore Pruning
-            self.__pruning_task = None
             self._prune_datastore = prune_datastore
             self._datastore_pruning_task = LoopingCall(f=self.__prune_datastore)
 
+            # Known nodes pruning
+            self._node_garbage_collection = node_garbage_collection
+            self._node_pruning_task = LoopingCall(f=self.__prune_nodes)
```
fjarri, line 1121: So this means that pruning is for Ursulas only, right? Alice could benefit from it too.

nucypher/acumen/perception.py
```diff
@@ -168,8 +218,91 @@ def abridged_state_details(state):
...
+                    # TODO: Trash can label?
+            else:
+                # Reinstate this node to good standing by un/relabeling
+                if label in self.__pruning_strategies:
+                    self.unlabel(node=node, label=label)
+                    self.__reset_node_label_tracking(node=node)
+
+    def prune_nodes(self) -> None:
+        """Apply node pruning strategies to all marked nodes once"""
+        self._pruning_strategies: Dict[type, List[Callable]]
```
fjarri, line 320: Was it supposed to be `__pruning_strategies`?

nucypher/acumen/perception.py
```diff
@@ -168,8 +218,91 @@ def abridged_state_details(state):
...
+        for strategy in self.__pruning_strategies:
+            strategy(node=node, reset=True)
+
+    def register_node_pruning_strategy(self,
+                                       label,
+                                       strategy: Callable,
+                                       layer: Optional[int] = None
+                                       ) -> None:
+        if label not in self.BUCKETS:
+            raise self.UnknownLabel
```
fjarri, line 231: This will benefit from a message as is done elsewhere in this file (`f"{label} is not a prunable node bucket."`)

nucypher/acumen/perception.py
KPrasch: Yes, you are right - Ill include this in my next commit ( unless you beat me to it :wink:  )

nucypher/acumen/perception.py
```diff
@@ -168,8 +218,91 @@ def abridged_state_details(state):
...
+        for _label in self.BUCKETS:
+            if node in self.__marked[_label]:
+                self.__marked[_label].remove(node)
+
+    def prune_bucket(self, label):
+        """Apply pruning strategies to a single node bucket once"""
+        try:
+            strategies = self.__pruning_strategies[label]
+        except KeyError:
+            raise self.UnknownLabel
```
fjarri, line 302: This will benefit from a message too

nucypher/characters/lawful.py
KPrasch: I agree. Despite Alice running the discovery loop for such a short amount of time it will still be useful, although this will be much more effective it she continues to run node discovery as a damon.

nucypher/acumen/perception.py
```diff
@@ -168,8 +218,91 @@ def abridged_state_details(state):
...
-    def mark_as(self, label: Exception, node: "Teacher"):
-        self._marked[label].append(node)
-
-        if self._nodes.get(node):
-            del self._nodes[node]
+    def __reset_node_label_tracking(self, node: "Teacher"):
+        for strategy in self.__pruning_strategies:
+            strategy(node=node, reset=True)
+
+    def register_node_pruning_strategy(self,
```
fjarri, line 225: Is it called from anywhere?

KPrasch: Not sure yet. This essentially stems from the potential need to allow arbitrary pruning strategies to be set dynamically while still having a path to reset it's internal state.

nucypher/acumen/perception.py
```diff
@@ -62,29 +108,33 @@ def __setitem__(self, checksum_address, node_or_sprout):
...
-        return bool(self._nodes)
+        return bool(self.__nodes)
 
     def __contains__(self, item):
-        return item in self._nodes.keys() or item in self._nodes.values()
+        return item in self.__nodes.keys() or item in self.__nodes.values()
 
     def __iter__(self):
-        yield from self._nodes.values()
+        yield from self.__nodes.values()
```
fjarri, line 120: So, despite the bucketing, we still return all the nodes? Would it be better to exclude the bucketed ones here and not wait for pruning to get around to it?

KPrasch: Tough to answer right now.  I think we can remove this method in favor of `FleetSensor.get_nodes`...  This largely depends on what public facing perspective on peers is.  Some buckets are positive health indicators, like the `VERIFIED` bucket - and perhaps some operations (like granting) can use a view of the network constrained on this basis.

nucypher/acumen/perception.py
fjarri: I don't think there's an immediate need for that. The user can't change the node's behavior in runtime (would be cool to process SIGHUP for that though).

nucypher/acumen/pruning.py
```diff
@@ -0,0 +1,63 @@
...
+
+def reject_node(node: "Teacher", reset: bool = False) -> bool:
+    """Indicate a node can be deleted now."""
+    return False
+
+
+def construct_node_stalecheck(max_seconds: int, start_time: Optional[maya.MayaDT] = None) -> Callable:
+    start_time = start_time or maya.now()
+
+    def check(node: "Teacher", reset: bool = False) -> bool:
```
fjarri, line 40: Not sure how I feel about that. As much as I like FP, would a `Strategy` ABC be more understandable in code?

KPrasch: Initially @vepkenez and I were compelled to keep these strategies completely stateless, but I'm not sure how that's *truly possible* if we're going to track an a per-node basis.  There might be some way to rework the rules of invalidation with statelessness in mind... :thinking:

nucypher/acumen/perception.py
fjarri: > I think we can remove this method in favor of `FleetSensor.get_nodes`...

I would actually be in favor of that. I think `FleetSensor` is too complicated of a class (especially with the addition of buckets) to be used as a simple collection.

nucypher/acumen/pruning.py
fjarri: Hm, well, for example you could do the following:
- Have a single dict (`node -> label`) (or `checksum_address -> label`). A label here is an object that incorporates a strategy.
- In `prune_bucket()`, you iterate over nodes and `__call__` the label of each. It either returns another label (or the same one), or `None`, the latter meaning that the node is to be removed.

This way you can have a `MaxAttempts` label that will return a `MaxAttempts(label.count - 1)` on each call, or `None` if `count` is 0. Similarly for `Stalecheck`.

fjarri: Of course, you'll lose the ability to quickly iterate over nodes with a chosen label. You'll have to maintain supporting structures if that's desirable.

nucypher/acumen/perception.py
KPrasch: :heavy_check_mark:

nucypher/acumen/perception.py
KPrasch: :heavy_check_mark:

KPrasch:
Rebased @ fd541af82afb8cd2e604492bc263d6612cb0166f

nucypher/acumen/pruning.py
KPrasch: Reworked as a callable ABC, although much more simply than described above.  There may need to be more development iterations but I've enough ground work to continue on.

nucypher/acumen/comprehension.py
```diff
@@ -0,0 +1,124 @@
...
+
+class StaleCheck(PruningStrategy):
+
+    def __init__(self, max_seconds: int):
+        self.max_seconds = max_seconds
+        super().__init__()
+
+    def __call__(self, node: "Teacher") -> bool:
+        delta = maya.now() - node.last_seen
+        return delta.seconds() > self.max_seconds
```
derekpierre, line 69: Did you mean the following i.e. return `False` when the timeout is reached to erase the node.

```suggestion
        return delta.seconds() < self.max_seconds
```

nucypher/acumen/comprehension.py
```diff
@@ -0,0 +1,124 @@
...
+
+    def __init__(self, max_attempts: int):
+        self.attempts = Counter()
+        self.max_attempts = max_attempts
+        super().__init__()
+
+    def __call__(self, node: "Teacher") -> bool:
+        self.attempts[node.checksum_address] += 1
+        if self.attempts[node.checksum_address] > self.max_attempts:
+            del self.attempts[node.checksum_address]  # stop tracking.
```
derekpierre, line 82: Just call reset then?

```suggestion
            self.reset(node)  # stop tracking.
```

nucypher/acumen/comprehension.py
```diff
@@ -0,0 +1,124 @@
...
+# Buckets that do not need pruning
+# UNVERIFIED
+# VERIFIED
+# SEEDNODE
+
+PRUNING_STRATEGIES = {
+    UNAVAILABLE: [
+        StaleCheck(max_seconds=60 * 60 * 72),
+        MaxAttempts(max_attempts=20)
+    ],
```
derekpierre, line 109: 👍 🔥

nucypher/acumen/perception.py
```diff
@@ -168,8 +187,82 @@ def abridged_state_details(state):
...
+                keep = strategy(node=node)
+                if not keep:
+                    del self.__nodes[node.checksum_address]  # prune node
+                    self.unlabel(node=node, label=label)     # prune corresponding label
+                    break  # this node is already doomed anyways
+                    # TODO: forget node from disk too
+                    # TODO: Trash can label?
+            else:
+                # Reinstate this node to good standing by un/relabeling
+                if label in PRUNING_STRATEGIES:
```
derekpierre, line 260: For clarity, based on the operations done in this if-block, should we use BUCKETS here instead of PRUNING_STRATEGIES?

```suggestion
                if label in BUCKETS:
```

nucypher/network/nodes.py
```diff
@@ -388,51 +502,21 @@ def remember_node(self,
                 # This node is already known.  We can safely return.
                 return False
 
-        self.known_nodes[node.checksum_address] = node  # FIXME - dont always remember nodes, bucket them.
-
-        if self.save_metadata:
-            self.node_storage.store_node_metadata(node=node)
+        retained = self._remember_essential(node=node)
+        unlabeled = self.known_nodes.get_label(node=node) is None
+        if retained and unlabeled:
```
derekpierre, line 507: 👍

nucypher/acumen/perception.py
KPrasch: Not every bucket has a strategy associated with it.

derekpierre: 👍 Right! ignore that suggestion.

Side note, why are we re-instating the node in this block? We are resetting the tracking if the node is determined to be kept for now and not yet pruned. Do we lose tracking data by resetting the node label tracking eg. max attempts?

KPrasch:
```
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/twisted/python/context.py", line 122, in callWithContext
	    return self.currentContext().callWithContext(ctx, func, *args, **kw)
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/twisted/python/context.py", line 85, in callWithContext
	    return func(*args,**kw)
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/twisted/internet/defer.py", line 460, in callback
	    self._startRunCallbacks(result)
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/twisted/internet/defer.py", line 568, in _startRunCallbacks
	    self._runCallbacks()
	  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/twisted/internet/defer.py", line 654, in _runCallbacks
	    current.result = callback(current.result, *args, **kw)
	  File "/home/kieran/Git/nucypher/nucypher/network/nodes.py", line 687, in _discover_or_abort
	    result = self.learn_from_teacher_node(eager=False, canceller=self._discovery_canceller)
	  File "/home/kieran/Git/nucypher/nucypher/network/nodes.py", line 913, in learn_from_teacher_node
	    self.known_nodes.label(node=current_teacher, label=UNAVAILABLE)
	  File "/home/kieran/Git/nucypher/nucypher/acumen/perception.py", line 229, in label
	    raise self.UnknownNode(f'Cannot label an unknown node ({node}).')
	nucypher.acumen.perception.UnknownNode: Cannot label an unknown node ((Ursula)⇀Silver Zero YellowGreen Three↽ (0x1A787A78acCa5fDaca642Ad267af2bd70973C197)).
	
```

derekpierre:
@KPrasch Presumably the node was running for some time before this? I wonder if this was the result of the potential concurrency issue we chatted about. Pruning nodes in one task while potentially trying to use them in another task.

KPrasch:
> @KPrasch Presumably the node was running for some time before this? 

Actually this happened about 3-4 minutes after launching on ibex testnet with know nodes on disk.

> I wonder if this was the result of the potential concurrency issue we chatted about. Pruning nodes in one task while potentially trying to use them in another task.

Possibly.  Lets think about this some more.  Also - what's preventing a "pruned" node from re-entering the loop with a fresh status?

KPrasch:
Rebased @ 24a57e1c810aa6408ecfc24942956925146aa024

KPrasch:
I'm considering closing the PR for staleness and trying again with a fresh approach atop latest the version.  Any other opinions on this?

KPrasch:
Great teamwork and good ideas are contained in this PR, but it's too stale to recover and we'll be better off cherry-picking or starting fresh (picking up inspiration from this diff).

PR #2515, opened 2021-01-12
When rendering /status, don't fail, but log on StampNotSigned

vepkenez:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Allows for a gentler handling of UnsignedStamps only when in the context of reading known nodes while rendering ursula's /status endpoint

Resolves https://sentry.io/organizations/nucypher/issues/2134995939/?project=1480080&query=is%3Aunresolved
Fixes the /status?json=1 which is currently broken network wide.

**Notes for reviewers:**
This will not affect https://github.com/nucypher/nucypher/issues/1698 or any of the related issues as it only will change anything when called from /status

I have already tested this on several NuCo nodes over the period of 2021-01-01 to 2021-01-10 or so

### DIFF
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -1331,10 +1331,15 @@ class Teacher:
             address_first6=self.checksum_address[2:8]
         )
 
-    def known_nodes_details(self) -> dict:
+    def known_nodes_details(self, raise_invalid=True) -> dict:
         abridged_nodes = {}
         for checksum_address, node in self.known_nodes._nodes.items():
-            abridged_nodes[checksum_address] = self.node_details(node=node)
+            try:
+                abridged_nodes[checksum_address] = self.node_details(node=node)
+            except self.StampNotSigned:
+                if raise_invalid:
+                    raise
+                self.log.error(f"encountered unsigned stamp for node with checksum: {checksum_address}")
         return abridged_nodes
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


nucypher/network/nodes.py
```diff
@@ -1331,10 +1331,15 @@ def nickname_icon_details(self):
...
+    def known_nodes_details(self, raise_invalid=True) -> dict:
         abridged_nodes = {}
         for checksum_address, node in self.known_nodes._nodes.items():
-            abridged_nodes[checksum_address] = self.node_details(node=node)
+            try:
+                abridged_nodes[checksum_address] = self.node_details(node=node)
+            except self.StampNotSigned:
+                if raise_invalid:
+                    raise
+                self.log.error(f"encountered unsigned stamp for node with checksum: {checksum_address}")
```
derekpierre, line 1342: I'm leaning towards `warn` instead of `error` since we are ignoring and not raising. Thoughts?

derekpierre (commented):
Newsfragment?

nucypher/network/nodes.py
vepkenez: The main reason I went with error is so we get good context on sentry.
https://sentry.io/organizations/nucypher/issues/2134995939/?project=1480080&query=is%3Aunresolved

KPrasch (approved):
Approved - pending follow-up in #2513

derekpierre (approved):
🎸 - Approved pending addition of a newsfragment.

KPrasch:
@vepkenez - I think 2 approvals is sufficient for this given the issue number is included as a comment :+1:

PR #2521, opened 2021-01-14
Fix CI build

fjarri:
**Type of PR:**
Bugfix

**Required reviews:** 
1

**What this does:**
Fix a bug introduced in #2486. [Example of a failed CI](https://app.circleci.com/pipelines/github/nucypher/nucypher/6922/workflows/86ddb3fc-8793-4899-89e2-7890ee71708c/jobs/118136)

Basically, the code in `_handle_failed_transaction()`
```
        try:
            code = int(response['code'])
            message = response['message']
        except (KeyError, ValueError):
```
May result in a `TypeError` if `response` is not subscriptable. As a result, `TypeError` is raised instead of `exception`, which results in multiple tests failing. 

The PR introduces two changes: 
- Let `response = exception.args[0]` fail naturally. If we're replacing the exception, we may as well know what exactly failed there.
- Catch all `Exception`s in the block above instead of enumerating possible failures. Would be even better to actually check the type of `response`, perhaps?

### DIFF
--- a/newsfragments/2521.bugfix.rst
+++ b/newsfragments/2521.bugfix.rst
@@ -1,0 +1,1 @@
+Restore the re-raising behavior in ``BlockchainInterface._handle_failed_transaction()``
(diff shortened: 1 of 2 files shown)


KPrasch (approved):
Thanks for the fix - I had not noticed this yet.  Since you're allowing the exception to occur naturally, I think type checking is overkill, especially since this is an internal API.

Issue #2234, opened 2020-09-08
A family recipe for 'TreasureMap con KFrags'


jMyles:
## Like herding KFrags

The custody of `KFrag` material is one of the clunkiest aspects of `Ursula` state, and is the source of some brittleness for NuCypher.  The diligence of wallet management (or, thought of in slightly wider-scoped terms, keystore management) is insufficient to ensure robustness of Policy cohorts.  For example, if `n - m +1` Ursulas lose their KFrags, it will be of no comfort to Bob that their keystore state is intact.

Instead, we can craft a mechanism wherein Bob can provide, during his journey, all of the cryptographic material required for a given Ursula (for whom this material has been made by Alice specifically to match a trinket-secret pair of Ursula's) to craft and supply his sought-after `CFrag`.  In this scenario, this material is included by Alice in the TreasureMap for Bob to use this way.  

*This is a solution that we've come to call **TreasureMap con KFrags**.*

----
## Our choices at present
Cryptographically, there are at least two ways to achieve this:

**Scenario A**: A full KFrag, tantamount to the material currently custodied in Ursula's datastore, is added to each Arrangement in the TreasureMap.  The material is encrypted for Ursula.

**Scenario B**: A more primordial object, an "evaluation index", is stored in the Arrangement.  In order for PRE service to be applied, a multi-party computation is required which involves a secret held by Bob.  The resulting ethereal state is useful only for the requested operation; subsequent operations require subsequent MPCs.

----
### The state of due diligence
[...]

Issue #2225, opened 2020-09-04
Use deterministic key derivation to produce Keyring keys


cygnusv:
This can be used to regenerate Ursula's keys from a single master secret.

Potentially consider including additional contextual information like domains into the mix when derivating.

KPrasch:
I suggest we remove this issue from the scope of Blue-oyster, and remove it from the before activation milestone.

cygnusv:
This issue becomes very important in the context of TreasureMaps con Kfrags (#2234): if KFrags are encrypted in the TMap under Ursulas' public keys, it's critical that such key is persistent; otherwise, Ursula may not be able to open a KFrag encrypted for her.

KPrasch:
Here is a BIP-39 compliant library that is already part of our dependencies via trezorlib that we may be able to employ to solve this issue: <https://github.com/trezor/python-mnemonic>

Issue #2071, opened 2020-06-04
Make Alice granting logic more resilient to potential TX failures/delays


cygnusv:
Due to many factors, a TX may be initially reported as failed, but still be mined eventually. Such factors include gas prices, timeouts, etc.

Currently, our code deals with TX failures, but I'm not sure if it deals with TX that were initially failed, leading to a failure in the parent call, but end up succeeding in the background. 
The following is a hypothetical example: Imagine that Alice uses `grant()`, creates a policy, sends the TX to `PolicyManager`, and after waiting for some time, we timeout and the call to `grant()` fails. It's not impossible, however, that the TX does succeed eventually. Is our `grant()` logic resilient to that? Would it be possible for Alice to recover the situation and resume the operation now that the TX exists? Or is she condemned to live with a zombie policy?

The case of Alice may be the most flagrant, but there are other places to where this can happen. 

Related to #2070

KPrasch:
@cygnusv - This issue may be too broad for closure - the original motive here was mitigation of worker auto-commitments which has been mitigated by new strategies and max price settings, so I'm removing oily rags.

cygnusv:
I'm trying to re-scope this issue to be more Alice-specific and I'm seeing this has ramifications. 

Essentially, our client assumes that the whole granting logic is going to work successfully from start to end, but this is way too optimistic, and also very brittle. There are several places where there may be, at least, some delay, for example in the on-chain TX to set the policy (due to raising gas prices), but in the worst-case there may be no confirmation at all in a reasonable time. In such cases, our logic is bound to fail (which can be acceptable to a certain extent), but what's more worrying is that it's currently not recoverable: e.g., if the policy creation TX clears several hours later, it's obvious that the policy should be valid in the network, which implies (1) having the nodes already storing the KFrag, and (2) have some TMap distribution mechanism ready (being that a deterministic TMap, or a locally-stored TMap, or a TMap already flooded to the network).

For background on this, see the high-level description from @Bogdan here: https://discord.com/channels/411401661714792449/786726745892978719/796922401971241011
```
The thing is, granting is a three-step process: 
1) propose to Ursulas and select n of the ones that agreed
2) post a transaction with the addresses of these n Ursulas
3) send kfrags to Ursulas, along with the transaction hash
So you won't be able to start 3) until your transaction is posted.
Ursulas want to see the transaction to know they are being paid
```

Some ideas/comments:
[...]

Issue #1701, opened 2020-02-16
Ursula needs to, herself, check if her Staker has valid stakes given the expiration of a proposed arrangement.


jMyles:

jMyles:
@cygnusv asks: Why bother storing the `Arrangement` at all?  If we instead simply sign an "OK, I'm good" message, then the matter of having to store a row goes away, and thus the gravity of the situation lowers for Ursula.  Thereinafter, if `Alice` still pushes forward with the Arrangement, despite Ursula not having sufficient stake, it's Alice's fault alone.

@tuxxy rightly points out, and refuses to relent from his claim to glorious recognition of the fact that ~~he~~ (@tuxxy points out that I have misrepresented his pronouns) they discussed this in vigorous "weeks ago".

KPrasch:
Closed by #2808 and the introduction of SubscriptionManager.

PR #2530, opened 2021-01-18
[WIP] My first batch of TreasureMaps con KFrags

cygnusv:
This PR is mainly intended as an approximation of how to introduce "TreasureMaps con KFrags"

Issues to address here:
* #2234: "TreasureMaps con KFrags" 
* #2071: Improve Alice granting resilience
* #343: Versioning TMaps
* #2225: Deterministic keyring derivation 

Other issues that can potentially fit here:
* #234: Ursula responds to consider_arrangement
* #1700 
* #1701


nucypher/policy/collections.py
```diff
@@ -66,7 +67,8 @@ class IsDisorienting(Bob.NotEnoughNodes):
         leaves Bob disoriented.
         """
 
-    node_id_splitter = BytestringSplitter((to_checksum_address, ETH_ADDRESS_BYTE_LENGTH), ID_LENGTH)
+    ursula_con_kfrag_splitter = BytestringSplitter((to_checksum_address, ETH_ADDRESS_BYTE_LENGTH),
```
fjarri, line 70: I don't think it's a good idea to insert Spanish into variable names.

nucypher/policy/collections.py
```diff
@@ -76,7 +78,11 @@ def __init__(self,
                  destinations=None,
                  message_kit: UmbralMessageKit = None,
                  public_signature: Signature = None,
-                 hrac: Optional[bytes] = None) -> None:
+                 hrac: Optional[bytes] = None,
+                 version: bytes = None) -> None:
```
fjarri, line 82: In what cases would we want to specify the version externally? A version determines how an object is serialized to bytes, which is hardcoded into the class definition.

nucypher/policy/collections.py
```diff
@@ -54,6 +54,7 @@
 
 class TreasureMap:
     ID_LENGTH = 32
+    version = bytes.fromhex("42")  # FIXME: This is a placeholder versioning, while we devise a better data serialization format
```
fjarri, line 57: Does that mean we're going the way of having versioned classes instead of a versioned protocol? I guess it provides more flexibility, but there might be problems with processing nested classes, each with its own version.

nucypher/policy/collections.py
cygnusv: Yes, I agree, this is temporary, and a pun based on the the original issue concept (#2234).

nucypher/policy/collections.py
cygnusv: As hinted in the comment, this is just a temporary scheme to facilitate _some_ versioning, reusing the existing serialization approach; it's not intended to be definitive.

nucypher/network/server.py
```diff
@@ -218,56 +218,6 @@ def consider_arrangement():
         # TODO: Make this a legit response #234.
         return Response(b"This will eventually be an actual acceptance of the arrangement.", headers=headers)
 
-    @rest_app.route("/kFrag/<id_as_hex>", methods=['POST'])
-    def set_policy(id_as_hex):
```
vepkenez, line 222: ❤️

KPrasch:
Replaced by #2687

PR #2551, opened 2021-02-07
Completes CLI Interactivity for Grant, Encrypt and Retrieve

KPrasch:
**Type of PR:**
Feature 

**Required reviews:** 
2

**What this does:**
Completes CLI interactivity and parameter collection for CLI usage of Alice, Bob, and Enrico.

The following usages are now supported, all flags are optional:

```
nucypher alice grant
nucypher enrico encrypt
nucypher bob retrieve
```

**Why it's needed:**
Improves usability and legibility of grant and retrieve while using the CLI.  This is particularly impactful for application developers  and demo presenters.

### DIFF
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -585,10 +585,10 @@ class BlockchainInterface:
         #
 
         try:  # TODO: Handle block confirmation exceptions
+            waiting_for = 'receipt'
             if confirmations:
-                emitter.message(f'Waiting for {confirmations} confirmations', color='yellow')
-            else:
-                emitter.message(f'Waiting {self.TIMEOUT} seconds for receipt', color='yellow')
+                waiting_for = f'{confirmations} confirmations'
+            emitter.message(f'Waiting {self.TIMEOUT} seconds for {waiting_for}', color='yellow')
             receipt = self.client.wait_for_receipt(txhash, timeout=self.TIMEOUT, confirmations=confirmations)
         except TimeExhausted:
             # TODO: #1504 - Handle transaction timeout
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -569,10 +569,11 @@ class BlockchainInterface:
         #
         # Broadcast
         #
-        emitter.message(f'Broadcasting {transaction_name} Transaction ({cost} ETH @ {price_gwei} gwei)...',
+        emitter.message(f'Broadcasting {transaction_name} Transaction ({cost} ETH @ {price_gwei} gwei)',
                         color='yellow')
         try:
             txhash = self.client.send_raw_transaction(signed_raw_transaction)  # <--- BROADCAST
+            emitter.message(f'TXHASH {txhash.hex()}', color='yellow')
         except (TestTransactionFailed, ValueError):
             raise  # TODO: Unify with Transaction failed handling -- Entry point for _handle_failed_transaction
         else:
@@ -584,6 +585,10 @@ class BlockchainInterface:
         #
 
         try:  # TODO: Handle block confirmation exceptions
+            if confirmations:
+                emitter.message(f'Waiting for {confirmations} confirmations', color='yellow')
+            else:
+                emitter.message(f'Waiting {self.TIMEOUT} seconds for receipt', color='yellow')
             receipt = self.client.wait_for_receipt(txhash, timeout=self.TIMEOUT, confirmations=confirmations)
         except TimeExhausted:
             # TODO: #1504 - Handle transaction timeout
--- a/nucypher/characters/control/specifications/bob.py
+++ b/nucypher/characters/control/specifications/bob.py
@@ -42,23 +42,30 @@ class JoinPolicy(BaseSchema):  #TODO:  this doesn't have a cli implementation
 
 class Retrieve(BaseSchema):
     label = fields.Label(
-        required=True, load_only=True,
-        click=options.option_label(required=True))
+        required=True,
+        load_only=True,
+        click=options.option_label(required=False)
+    )
     policy_encrypting_key = fields.Key(
         required=True,
         load_only=True,
-        click=options.option_policy_encrypting_key(required=True))
+        click=options.option_policy_encrypting_key(required=False)
+    )
     alice_verifying_key = fields.Key(
-        required=True, load_only=True,
+        required=False,
+        load_only=True,
         click=click.option(
             '--alice-verifying-key',
             '-avk',
             help="Alice's verifying key as a hexadecimal string",
             type=click.STRING,
-            required=False))
+            required=False)
+    )
     message_kit = fields.UmbralMessageKit(
-        required=True, load_only=True,
-        click=options.option_message_kit(required=True))
+        required=True,
+        load_only=True,
+        click=options.option_message_kit(required=False)
+    )
 
     cleartexts = fields.List(fields.Cleartext(), dump_only=True)
 
--- a/nucypher/cli/actions/collect.py
+++ b/nucypher/cli/actions/collect.py
@@ -72,7 +72,7 @@ def collect_bob_public_keys(
     if not bob_encrypting_key:
         bob_encrypting_key = click.prompt("Enter Bob's encrypting key")
     if not bob_verifying_key:
-        bob_verifying_key = click.prompt("Enter Bob verifying key")
+        bob_verifying_key = click.prompt("Enter Bob's verifying key")
 
     public_keys = PublicKeys(encrypting_key=bob_encrypting_key, verifying_key=bob_verifying_key)
     return public_keys
--- a/nucypher/cli/actions/collect.py
+++ b/nucypher/cli/actions/collect.py
@@ -1,0 +1,198 @@
+"""
+ This file is part of nucypher.
+
+ nucypher is free software: you can redistribute it and/or modify
+ it under the terms of the GNU Affero General Public License as published by
+ the Free Software Foundation, either version 3 of the License, or
+ (at your option) any later version.
+
+ nucypher is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Affero General Public License for more details.
+
+ You should have received a copy of the GNU Affero General Public License
+ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+"""
+
+from collections import namedtuple
+
+import click
+import maya
+from constant_sorrow.constants import FEDERATED
+from datetime import timedelta
+from typing import Tuple
+from web3.main import Web3
+
+from nucypher.characters.control.emitters import StdoutEmitter
+from nucypher.characters.lawful import Bob, Alice
+from nucypher.cli.painting.help import enforce_probationary_period, paint_probationary_period_disclaimer
+from nucypher.cli.painting.policies import paint_single_card
+from nucypher.cli.types import GWEI
+from nucypher.policy.identity import Card
+
+PublicKeys = namedtuple('PublicKeys', 'encrypting_key verifying_key')
+PolicyParameters = namedtuple('PolicyParameters', 'label m n expiration value rate')
+
+
+def collect_keys_from_card(emitter: StdoutEmitter, card_identifier: str, force: bool):
+    emitter.message(f"Searching contacts for {card_identifier}\n", color='yellow')
+    card = Card.load(identifier=card_identifier)
+
+    if card.character is not Bob:
+        emitter.error('Grantee card is not a Bob.')
+        raise click.Abort
+    paint_single_card(emitter=emitter, card=card)
+
+    if not force:
+        click.confirm('Is this the correct grantee (Bob)?', abort=True)
+
+    bob_encrypting_key = card.encrypting_key.hex()
+    bob_verifying_key = card.verifying_key.hex()
+    public_keys = PublicKeys(encrypting_key=bob_encrypting_key, verifying_key=bob_verifying_key)
+    return public_keys
+
+
+def collect_bob_public_keys(
+        emitter: StdoutEmitter,
+        force: bool,
+        card_identifier: str,
+        bob_encrypting_key: str,
+        bob_verifying_key: str
+        ) -> PublicKeys:
+    """helper function for collecting Bob's public keys interactively in the Alice CLI."""
+
+    # From Bob card
+    if card_identifier and any((bob_encrypting_key, bob_verifying_key)):
+        message = '--bob cannot be used with --bob-encrypting-key or --bob-verifying key'
+        raise click.BadOptionUsage(option_name='--bob', message=message)
+
+    if card_identifier:
+        public_keys = collect_keys_from_card(
+            emitter=emitter,
+            card_identifier=card_identifier,
+            force=force)
+        return public_keys
+
+    # From hex public keys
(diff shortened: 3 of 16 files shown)


nucypher/blockchain/eth/interfaces.py
```diff
@@ -584,6 +585,10 @@ def sign_and_broadcast_transaction(self,
         #
 
         try:  # TODO: Handle block confirmation exceptions
+            if confirmations:
+                emitter.message(f'Waiting for {confirmations} confirmations', color='yellow')
+            else:
+                emitter.message(f'Waiting {self.TIMEOUT} seconds for receipt', color='yellow')
```
derekpierre, line 591: I don't think confirmations and timeout are mutually exclusive. It's a "keep trying until one condition, sufficient confirmations or timeout, happens first", no? If so. we may need to reword.

Perhaps something like:
```suggestion
            waiting_for = 'receipt'
            if confirmations:
                waiting_for = f'{confirmations} confirmations'
            emitter.message(f'Waiting {self.TIMEOUT} seconds for {waiting_for}', color='yellow')
```

nucypher/characters/control/specifications/bob.py
```diff
@@ -42,23 +42,30 @@ class JoinPolicy(BaseSchema):  #TODO:  this doesn't have a cli implementation
 
 class Retrieve(BaseSchema):
     label = fields.Label(
-        required=True, load_only=True,
-        click=options.option_label(required=True))
+        required=True,
+        load_only=True,
+        click=options.option_label(required=False)
```
derekpierre, line 47: Presumably, there are no side effects of this change, and the only effect is what the PR intended i.e. that Bob will be prompted for the label if not explicitly specified.

nucypher/cli/commands/alice.py
```diff
@@ -483,6 +484,13 @@ def grant(general_config,
         bob_encrypting_key = card.encrypting_key.hex()
         bob_verifying_key = card.verifying_key.hex()
 
+    if not bob_encrypting_key or bob_verifying_key:
+        emitter.message("*Caution: Only enter public keys*")
+    if not bob_encrypting_key:
+        bob_encrypting_key = click.prompt("Enter Bob's encrypting key")
+    if not bob_verifying_key:
+        bob_verifying_key = click.prompt("Enter Bob verifying key")
+
```
derekpierre, line 493: Do we need to be careful here of `force` is true, or is that case handled earlier?

nucypher/cli/commands/alice.py
```diff
@@ -500,14 +508,12 @@ def grant(general_config,
...
     if not force and not expiration:
-        if ALICE.duration_periods:
-            # TODO: use a default in days or periods?
-            expiration = maya.now() + timedelta(days=ALICE.duration_periods)  # default
-            if not click.confirm(f'Use default policy duration (expires {expiration})?'):
-                expiration = click.prompt('Enter policy expiration datetime', type=click.DateTime())
-        else:
-            # No policy duration default default available; Go interactive
-            expiration = click.prompt('Enter policy expiration datetime', type=click.DateTime())
+        expiration_prompt = 'Enter policy expiration (Y-M-D H:M:S)'
```
derekpierre, line 511: I know we allowed `click.DateTime()` before which allowed for formats `["%Y-%m-%d", "%Y-%m-%dT%H:%M:%S", "%Y-%m-%d %H:%M:%S"]`, but can we actually do anything more granular than day i.e. hours, minutes, seconds? If we can't, then we should explicitly specify the format `"%Y-%m-%d"` to click.DateTime, and only prompt for `Y-m-d` to be specified.

nucypher/cli/commands/enrico.py
```diff
@@ -63,25 +63,38 @@ def run(general_config, policy_encrypting_key, dry_run, http_port):
...
     ENRICO = _create_enrico(emitter, policy_encrypting_key)
-    if not (bool(message) ^ bool(file)):
-        emitter.error(f'Pass either --message or --file. Got {message}, {file}.')
+    if message and file:
+        emitter.error(f'Pass either --message or --file, not both.')
         raise click.Abort
 
+    if not message:
+        message = click.prompt('Enter plaintext to encrypt', type=click.STRING)
+
```
derekpierre, line 88: confused here - what if `file` was passed; doesn't this nullify the original XOR check?

nucypher/cli/utils.py
```diff
@@ -101,6 +101,7 @@ def make_cli_character(character_config,
     if character_config.federated_only:
         emitter.message(FEDERATED_WARNING, color='yellow')
 
+    emitter.message(f"Loaded {CHARACTER.__class__.__name__} {CHARACTER.checksum_address} ({CHARACTER.domain})", color='green')
```
derekpierre, line 104: 👍

nucypher/cli/commands/alice.py
```diff
@@ -525,14 +531,15 @@ def grant(general_config,
     # Policy Value
     policy_value_provided = bool(value) or bool(rate)
     if not ALICE.federated_only and not policy_value_provided:
-        rate = ALICE.default_rate  # TODO #1709 - Fine tuning and selection of default rates
+        # TODO #1709 - Fine tuning and selection of default rates
+        rate = ALICE.default_rate or Web3.toWei(50, 'gwei')  # TODO: Unhardcode, workaround for lynx/ibex testnet.
```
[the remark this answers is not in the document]
KPrasch: Current this value is 0 in lynx, and I wanted to show some number here for a demo.  What is better to show in this case?

nucypher/characters/control/specifications/bob.py
KPrasch: Yes, that's correct, the Click options need be be optional to allow the fallback to interactive collection.

nucypher/cli/commands/alice.py
vepkenez: Yes.  If find these precise format requirements to be very frictional.

KPrasch: > can we actually do anything more granular than day i.e. hours, minutes, seconds?

Yes, epoch seconds is used at the contract-level.

nucypher/cli/commands/enrico.py
KPrasch: Unsure - this is more or less an "application level" decision.  What's a sensible default behavior here?

derekpierre: I guess my thought was that the original code ensured that only one of them is set by using the XOR - `if not (bool(message) ^ bool(file)):`. In this change, we ensure that both aren't set, `if message and file` but we don't know which one, `file` or `message` is actually set, or if they are set at all.

Subsequently, you check on L86 whether `message` is not set, which could mean that either `file` is or is not set`. I think to do what you are trying to do here you might have to do something like:

```suggestion
    if not file and not message:
        message = click.prompt('Enter plaintext to encrypt', type=click.STRING)

```

nucypher/cli/commands/alice.py
derekpierre: Interesting. I guess in that case there is payment for periods and any part thereof i.e. payment for fees is period granular (effectively UTC day), but enaction of a policy is second granularity (@vzotova do I have this right?).

nucypher/cli/commands/enrico.py
KPrasch: > I guess my thought was that the original code ensured that only one of them is set by using the XOR

It doesn't play nicely with the interactivity -  When might this check be effective if we're prompting for manual input?

>  I think to do what you are trying to do here you might have to do something like

I'm liking this simple approach here, with the default behavior focused around encrypting messages (good for first-time users, hobbyist, demoers).  `--file` can be added if needed.

```suggestion
    if not message or file:
        message = click.prompt('Enter plaintext to encrypt', type=click.STRING)

```

nucypher/cli/commands/alice.py
KPrasch: Policy rates are calculated on a per-period basis, but retain their original timeatamps:

https://github.com/nucypher/nucypher/blob/main/nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol#L239

nucypher/cli/commands/alice.py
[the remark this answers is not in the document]
vzotova: I'm for setting default value on `lynx`

nucypher/cli/commands/enrico.py
KPrasch: :heavy_check_mark:

nucypher/cli/commands/alice.py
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark: - Handled 0 value for rate in 55f2ede773390c6197ca4f830029a719a78f6cce.  This is an exceptional case, so maybe it;s better to leave out this commit.

nucypher/cli/commands/alice.py
KPrasch: :heavy_check_mark:

nucypher/cli/actions/collect.py
```diff
@@ -0,0 +1,198 @@
...
+        if not force:
+            default_gwei = Web3.fromWei(rate, 'gwei')  # wei -> gwei
+            prompt = "Confirm rate of {node_rate} gwei * {n} nodes ({period_rate} gwei per period)?"
+
+            if not click.confirm(prompt.format(node_rate=default_gwei, period_rate=default_gwei * n, n=n), default=True):
+                interactive_rate = click.prompt('Enter rate per period in gwei', type=GWEI)
+                # TODO: Interactive rate sampling & validation (#1709)
+                interactive_prompt = prompt.format(node_rate=interactive_rate, period_rate=interactive_rate * n, n=n)
+                click.confirm(interactive_prompt, default=True, abort=True)
+                rate = Web3.toWei(interactive_rate, 'gwei')  # gwei -> wei
```
cygnusv, line 146: Does this solve the problem I found?
```
Confirm rate of 50 gwei * 3 (150 gwei per period)? [Y/n]: n
Enter rate per period in gwei: 60
Confirm rate of 50000000000 gwei * 3 (150000000000 gwei per period)? [Y/n]:
```

cygnusv (approved):
Tested live on Ethereum Denver...Works like a charm! Thanks for this @KPrasch !
The collect and confirm modules are a great approach to keep the code tidy. The CLI code looks much better now.

nucypher/cli/actions/collect.py
```diff
@@ -0,0 +1,198 @@
...
+        force: bool,
+        card_identifier: str,
+        bob_encrypting_key: str,
+        bob_verifying_key: str
+        ) -> PublicKeys:
+    """helper function for collecting Bob's public keys interactively in the Alice CLI."""
+
+    # From Bob card
+    if card_identifier and any((bob_encrypting_key, bob_verifying_key)):
+        message = '--bob cannot be used with --bob-encrypting-key or --bob-verifying key'
```
derekpierre, line 67: Feels like a little leakage here...`--bob` as it relates to `card_identifier`

We could leave it or move verifications to the originating source before we call `collect_bob_public_keys()`. Something to think about...

nucypher/cli/actions/collect.py
```diff
@@ -0,0 +1,198 @@
...
+
+    if card_identifier:
+        public_keys = collect_keys_from_card(
+            emitter=emitter,
+            card_identifier=card_identifier,
+            force=force)
+        return public_keys
+
+    # From hex public keys
+    if not bob_encrypting_key or bob_verifying_key:
```
derekpierre, line 78:
```suggestion
    if not bob_encrypting_key or not bob_verifying_key:
```

vepkenez:
> Would you mind if I move all this validation logic into the Serializer since that's the whole point of that?

Or we should get rid of the serializers since this becomes more difficult if validation logic is pepperer all over the place... but I still think it's useful to validate at the "model" level than on the CLI interface level since much of this logic still applies.

nucypher/cli/actions/collect.py
KPrasch: :+1:  Yes it does.

nucypher/cli/commands/alice.py
```diff
@@ -525,14 +531,15 @@ def grant(general_config,
...
         if not force:
             default_gwei = Web3.fromWei(rate, 'gwei')
-            prompt = "Confirm rate of {node_rate} gwei ({total_rate} gwei per period)?"
-            if not click.confirm(prompt.format(node_rate=default_gwei, total_rate=default_gwei*n), default=True):
+            prompt = "Confirm rate of {node_rate} gwei * {n} ({total_rate} gwei per period)?"
```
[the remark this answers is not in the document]
KPrasch: :heavy_check_mark:

nucypher/cli/actions/collect.py
KPrasch: :heavy_check_mark:

nucypher/cli/actions/collect.py
KPrasch: :heavy_check_mark:

KPrasch:
> I still think it's useful to validate at the "model" level than on the CLI interface level since much of this logic still applies to other interaction contexts.

@vepkenez Yes, exactly. In fact I think we have developed just the right amount of granularity in validation.  Serializers in `character.control.specifications` are performing an essential role by validating the grant quest before it is passed to the character, on the "model layer" as you mentioned.  The CLI validators relocated in 005e819 (at the "application layer") check flag combinations that include fields that are not part of `Alice.grant` data model, like `--card.`   There may be an abstraction leak by deriving the click fields in the serializes, but it's doing a fine job - no reason to rework it IMO.

nucypher/cli/actions/collect.py
```diff
@@ -0,0 +1,180 @@
...
+        public_keys = collect_keys_from_card(
+            emitter=emitter,
+            card_identifier=card_identifier,
+            force=force)
+        return public_keys
+
+    if not bob_encrypting_key:
+        bob_encrypting_key = click.prompt("Enter Bob's encrypting key")
+    if not bob_verifying_key:
+        bob_verifying_key = click.prompt("Enter Bob verifying key")
```
derekpierre, line 75:
```suggestion
        bob_verifying_key = click.prompt("Enter Bob's verifying key")
```

nucypher/cli/commands/enrico.py
```diff
@@ -63,25 +63,38 @@ def run(general_config, policy_encrypting_key, dry_run, http_port):
...
     """Encrypt a message under a given policy public key."""
 
-    # Setup
     emitter = setup_emitter(general_config=general_config, banner=policy_encrypting_key)
+
+    if ipfs:
+        try:
+            import ipfshttpclient
+        except ImportError:
+            raise ImportError("IPFS HTTP client not installed.  Run pip install ipfshttpclient then try again.")
```
derekpierre, line 72:
```suggestion
            raise ImportError("IPFS HTTP client not installed.  Run 'pip install ipfshttpclient' then try again.")
```

nucypher/cli/commands/enrico.py
```diff
@@ -63,25 +63,38 @@ def run(general_config, policy_encrypting_key, dry_run, http_port):
...
+
+        # Connect to IPFS before proceeding
+        ipfs_client = ipfshttpclient.connect(ipfs)
+        emitter.message(f"Connected to IPFS Gateway {ipfs}")
+
+    if not policy_encrypting_key:
+        policy_encrypting_key = click.prompt("Enter policy public key", type=click.STRING)
```
derekpierre, line 79:
```suggestion
        policy_encrypting_key = click.prompt("Enter policy encrypting key", type=click.STRING)
```

derekpierre (approved):
🎸 - looks good! A few minor suggestions.

vepkenez (approved):

PR #2555, opened 2021-02-07
Cache Alice's transacting power for later activation

KPrasch:
**Type of PR:**
Bugfix

**Required reviews:** 
1

**Why it's needed:**
Provides a handle for `Alice` signers to be used while sharing a blockchain connection with other character in-scope.


derekpierre (approved):
🎸

PR #2557, opened 2021-02-07
Stop the threadpool when all the worker results are processed

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 2

**What this does:**
Currently if the main thread stops before `TreasureMapPublisher` finishes distributing treasure maps (e.g. in a CLI granting situation), the process hangs indefinitely. It happens because if `WorkerPool.join()` is never called, the thread pool is still active and waiting for workers to be enqueued. 

In this PR, we stop the thread pool explicitly when all the worker results are processed, so it is guaranteed to be empty and to not receive any more `callInThread()` calls. 

It also makes the logic more strict:
- disallows races in `_threadpool.stop()`
- makes sure `join()` waits for all threads to finish for all callers, not just for the first one
- in case of an unexpected error in the producer thread, raises the error for all `join()` and `block()` callers, not just the first one

### DIFF
--- a/nucypher/utilities/concurrency.py
+++ b/nucypher/utilities/concurrency.py
@@ -32,9 +32,9 @@ class Success:
 
 
 class Failure:
-    def __init__(self, value, exception):
+    def __init__(self, value, exc_info):
         self.value = value
-        self.exception = exception
+        self.exc_info = exc_info
 
 
 class Cancelled(Exception):
@@ -246,7 +246,7 @@ class WorkerPool:
         except Cancelled as e:
             self._result_queue.put(e)
         except BaseException as e:
-            self._result_queue.put(Failure(value, str(e)))
+            self._result_queue.put(Failure(value, sys.exc_info()))
 
     def _process_results(self):
         """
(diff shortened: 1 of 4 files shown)


nucypher/utilities/concurrency.py
```diff
@@ -139,29 +133,35 @@ def cancel(self):
...
+        # and we also attempt to stop the pool from the `_process_results()` thread).
+        with self._threadpool_stop_lock:
+            if not self._threadpool_stopped:
+                self._threadpool.stop()
+                self._threadpool_stopped = True
+
+    def _check_for_unexpected_error(self):
+        if self._unexpected_error.is_set():
+            e = self._unexpected_error.get()
+            raise RuntimeError(f"Unexpected error in the producer thread: {e}")
```
KPrasch, line 148: Is the error's stack trace recoverable here? What do you think about raising the existing exception, instead of rendering it as a string?

fjarri: Raising the original exception here now.

Also saving the traceback in failures, but not sure how to log it - currently most connection failures are reported as exceptions, so there will be a lot of trash in the logs if we were to just print it all.

KPrasch: Great! Even if the stack traces are not logged, it's still useful In a debugging scenario to have them readily available.

KPrasch (approved):

vepkenez (approved):

Issue #1051, opened 2019-06-05
Ensure CLI-launched Bobs can operate On/Off Chain


KPrasch:
At the moment Bob CLI attempts to connect to a web3 provider by default.  Instead, ensure that bob can operate with or without a blockchain connection.

PR #2560, opened 2021-02-09
Alternate finnegans wake demo for Lynx testnet + API Refinements

KPrasch:
**Type of PR:**
Documentation / Examples

**Required reviews:** 
2 including @fjarri 's approval :smile: 

**What this does:**

- Reduces the number of arguments needed to initialize characters in the Grant and Retrieve flow.

- Provides an alternative example script of finnegans wake demo for the `lynx` testnet on goerli.
- Removes `client_password` where possible from characters init
- Requires a signer be used for direct Actors (`ContractAdministrator`, `Trustee`, `Bidder`, etc.)
- Require a `signer` be used to init `TransactingPower`
- Introduces `connect_web3_provider` utility
- Ensure Bob can operate without a blockchain connection (do not verify node bonding by default).

Closes https://github.com/nucypher/nucypher/issues/1051

**Why it's needed:**
Improves application development on-boarding and understanding of nucypher usage in a decentralized network.

### DIFF
--- a/examples/finnegans_wake_demo/README.md
+++ b/examples/finnegans_wake_demo/README.md
@@ -1,4 +1,4 @@
-# Ursula exchange example over live network
+# Ursula exchange example over local network
 
 This illustrates Alice sharing data with Bob over the NuCypher network using proxy re-encryption,
 without revealing private keys to intermediary entities.
@@ -14,10 +14,6 @@ without revealing private keys to intermediary entities.
 
 ### Run a fleet of federated Ursulas
 `python ../run_demo_ursula_fleet.py`
-
-
-### Download the Book!
-`./download_finnegans_wake.sh` 
 
 
 ### Run
--- a/examples/finnegans_wake_demo/README.md
+++ b/examples/finnegans_wake_demo/README.md
@@ -1,4 +1,4 @@
-# Ursula exchange example over local network
+# Finnegan's Wake Demo
 
 This illustrates Alice sharing data with Bob over the NuCypher network using proxy re-encryption,
 without revealing private keys to intermediary entities.
@@ -11,10 +11,25 @@ without revealing private keys to intermediary entities.
 5. Bob receives and reconstructs the Enrico from Policy public key and Enrico public key
 6. Bob retrieves the original message from Enrico and MessageKit
 
+There are two version of the example, one federated example using a local network
+and another example using the nucypher application development tesnet: Lynx.
 
-### Run a fleet of federated Ursulas
+### Federated Demo
+
+First run the local federated network:
 `python ../run_demo_ursula_fleet.py`
 
+Then run the demo:
+`python finnegans-wake-demo-federated.py`
 
-### Run
-`python finnegans-wake-demo.py`
+### Testnet Demo
+
+First, configure the demo.  Be sure tat alice's address has some Goerli ETH.
+```bash
+export DEMO_PROVIDER_URI=<GOERLI RPC ENDPOINT>
+export DEMO_ALICE_ETH_ADDRESS=<ETH ADDRESS>
+export DEMO_SIGNER_URI=keystore://<PATH TO KEYSTORE>
+```
+
+Then run the demo:
+`python finnegans-wake-demo-testnet.py`
--- a/examples/finnegans_wake_demo/finnegans-wake-demo-federated.py
+++ b/examples/finnegans_wake_demo/finnegans-wake-demo-federated.py
@@ -17,17 +17,17 @@
 
 
 import datetime
-import os
 import maya
+import os
 import sys
 from pathlib import Path
-from umbral.keys import UmbralPublicKey
 
-from crypto.powers import SigningPower, DecryptingPower
 from nucypher.characters.lawful import Alice, Bob, Ursula
 from nucypher.characters.lawful import Enrico as Enrico
 from nucypher.config.constants import TEMPORARY_DOMAIN
+from nucypher.crypto.powers import SigningPower, DecryptingPower
 from nucypher.utilities.logging import GlobalLoggerSettings
+
 
 ######################
 # Boring setup stuff #
@@ -79,6 +79,12 @@ encrypting_key = LOCAL_BOB.public_keys(DecryptingPower)
 LOCAL_ALICE = Alice(federated_only=True,
                     domain=TEMPORARY_DOMAIN,
                     known_nodes=[ursula])
+
+
+# Start node discovery and wait until 8 nodes are known in case
+# the fleet isn't fully spun up yet, as sometimes happens on CI.
+LOCAL_ALICE.start_learning_loop(now=True)
+LOCAL_ALICE.block_until_number_of_known_nodes_is(8, timeout=30, learn_on_this_thread=True)
 
 # Alice can get the public key even before creating the policy.
 # From this moment on, any Data Source that knows the public key
--- a/examples/finnegans_wake_demo/finnegans-wake-demo-federated.py
+++ b/examples/finnegans_wake_demo/finnegans-wake-demo-federated.py
@@ -15,122 +15,87 @@
  along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
+
 import datetime
 import maya
-import os
+import sys
+from pathlib import Path
 from umbral.keys import UmbralPublicKey
-from web3.main import Web3
 
-from nucypher.blockchain.eth.signers.base import Signer
 from nucypher.characters.lawful import Alice, Bob, Ursula
 from nucypher.characters.lawful import Enrico as Enrico
-from nucypher.crypto.powers import SigningPower, DecryptingPower
+from nucypher.config.constants import TEMPORARY_DOMAIN
 from nucypher.utilities.logging import GlobalLoggerSettings
-
-# Twisted Logger
-GlobalLoggerSettings.set_log_level(log_level_name='debug')
-GlobalLoggerSettings.start_console_logging()
-
-# Execute the download script (download_finnegans_wake.sh) to retrieve the book
-BOOK_PATH = os.path.join('.', 'finnegans-wake.txt')
-
-# Change this value to to perform more or less total re-encryptions
-# in order to avoid processing the entire book's text. (it's long)
-NUMBER_OF_LINES_TO_REENCRYPT = 25
-
-# Seednode
-SEEDNODE_URI = "https://lynx.nucypher.network:9151"
-SEEDNODE = Ursula.from_seed_and_stake_info(SEEDNODE_URI)
 
 ######################
 # Boring setup stuff #
 ######################
 
-# Configuration
-DOMAIN = 'lynx'  # testnet
+BOOK_PATH = Path('finnegans-wake-excerpt.txt')
 
-# Replace with ethereum RPC endpoint
-ETH_PROVIDER = ''
-PROVIDER_URI = os.environ.get('NUCYPHER_PROVIDER_URI', ETH_PROVIDER)
+# Twisted Logger
+GlobalLoggerSettings.set_log_level(log_level_name='debug')
+GlobalLoggerSettings.start_console_logging()
 
-# Replace with signer URI
-WALLET_FILEPATH = ''
-ETH_WALLET = f'keystore://{WALLET_FILEPATH}'
-SIGNER_URI = os.environ.get('NUCYPHER_SIGNER_URI', ETH_WALLET)
+# if your ursulas are NOT running on your current host,
+# run like this: python finnegans-wake-demo.py 172.28.1.3:11500
+# otherwise the default will be fine.
 
-# Replace with alice's ethereum address
-ETH_ADDRESS = ''
-ALICE_ETH_ADDRESS = os.environ.get('ALICE_ETH_ADDRESS', ETH_ADDRESS)
+try:
+    SEEDNODE_URI = sys.argv[1]
+except IndexError:
+    SEEDNODE_URI = "localhost:11500"
 
+##############################################
+# Ursula, the Untrusted Re-Encryption Proxy  #
+##############################################
+ursula = Ursula.from_seed_and_stake_info(seed_uri=SEEDNODE_URI,
+                                         federated_only=True,
+                                         minimum_stake=0)
 
-#####################
-# Bob the BUIDLer  ##
-#####################
-
-# First there was Bob.
-BOB = Bob(
-    known_nodes=[SEEDNODE],
-    domain=DOMAIN,
-    provider_uri=PROVIDER_URI
-)
-
-# Bob gives his public keys to alice.
-verifying_key = BOB.public_keys(SigningPower)
-encrypting_key = BOB.public_keys(DecryptingPower)
+# Here are our Policy details.
+policy_end_datetime = maya.now() + datetime.timedelta(days=1)
+m, n = 2, 3
+label = b"secret/files/and/stuff"
 
 ######################################
 # Alice, the Authority of the Policy #
 ######################################
 
-# Alice ethereum wallet
-wallet = Signer.from_signer_uri(SIGNER_URI)
-password = input(f'Enter password to unlock {ALICE_ETH_ADDRESS}: ')
-wallet.unlock_account(account=ALICE_ETH_ADDRESS, password=password)
-
-ALICE = Alice(
-
-    # Connection details
-    domain=DOMAIN,
-    known_nodes=[SEEDNODE],
-    provider_uri=PROVIDER_URI,
-
-    # Wallet details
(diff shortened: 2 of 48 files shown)


examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
```diff
@@ -0,0 +1,191 @@
...
+
+# Execute the download script (download_finnegans_wake.sh) to retrieve the book
+BOOK_PATH = os.path.join('.', 'finnegans-wake.txt')
+
+# Change this value to to perform more or less total re-encryptions
+# in order to avoid processing the entire book's text. (it's long)
+NUMBER_OF_LINES_TO_REENCRYPT = 25
+
+# Seednode
+SEEDNODE_URI = "https://lynx.nucypher.network:9151"
```
cygnusv, line 42: Can we fetch this from `RestMiddleware.TEACHER_NODES`?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
```diff
@@ -0,0 +1,191 @@
...
+SEEDNODE = Ursula.from_seed_and_stake_info(SEEDNODE_URI)
```
cygnusv, line 43: This makes me think about a method like `Ursula.seednode_for_network(network_name)` or something like that, removing the need for the seednode uri above.

cygnusv (approved):
Thanks @KPrasch !! It's a simple yet extremely valuable addition to the demos!
Can we get at least some reference in the docs for this?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
derekpierre: 👍

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
```diff
@@ -0,0 +1,191 @@
...
+from nucypher.characters.lawful import Enrico as Enrico
+from nucypher.crypto.powers import SigningPower, DecryptingPower
+from nucypher.utilities.logging import GlobalLoggerSettings
+
+# Twisted Logger
+GlobalLoggerSettings.set_log_level(log_level_name='debug')
+GlobalLoggerSettings.start_console_logging()
+
+# Execute the download script (download_finnegans_wake.sh) to retrieve the book
+BOOK_PATH = os.path.join('.', 'finnegans-wake.txt')
```
derekpierre, line 35: pathlib? 😅

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
```diff
@@ -0,0 +1,191 @@
...
+    # Wallet details
+    checksum_address=ALICE_ETH_ADDRESS,
+    signer=wallet,
+    client_password=password
+)
+
+# Here are the Policy details.
+policy_end_datetime = maya.now() + datetime.timedelta(days=5)
+m, n = 2, 3
+label = b"secret/files/and/stuff"
```
derekpierre, line 106: Is there any concern of reuse of the label through multiple runs of the demo?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
```diff
@@ -0,0 +1,191 @@
...
+    provider_uri=PROVIDER_URI,
+
+    # Wallet details
+    checksum_address=ALICE_ETH_ADDRESS,
+    signer=wallet,
+    client_password=password
+)
+
+# Here are the Policy details.
+policy_end_datetime = maya.now() + datetime.timedelta(days=5)
```
derekpierre, line 104: Depending on how much use this demo gets, would making the duration shorter (perhaps a day?) help save lynx nodes from keeping too much cruft? Do we even care?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
fjarri, line 35: If we're only using the first 25 lines, why not just make it a constant in the file?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
```diff
@@ -0,0 +1,191 @@
...
+ETH_ADDRESS = ''
+ALICE_ETH_ADDRESS = os.environ.get('ALICE_ETH_ADDRESS', ETH_ADDRESS)
+
+
+#####################
+# Bob the BUIDLer  ##
+#####################
+
+# First there was Bob.
+BOB = Bob(
```
fjarri, line 71: Is there a reason to have `BOB` and `bob`, which are different objects (same goes for Alice)? It's rather confusing. Maybe `bob` and `remote_bob`? Or `bob` and `public_bob`?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
```diff
@@ -0,0 +1,191 @@
...
+    n=n,
+    rate=Web3.toWei(50, 'gwei'),
+    expiration=policy_end_datetime
+)
+
+# Alice puts her public key somewhere for Bob to find later...
+alices_pubkey_bytes_saved_for_posterity = bytes(ALICE.stamp)
+
+# ...and then disappears from the internet.
+ALICE.disenchant()
```
fjarri, line 131: A question the user will have at this point: why do I need to call `disenchant()` for some character objects, but not for others?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: You mean modify the book text after it's been downloaded?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: Yes, but I think there's a small tradeoff for readability there. IMO default seed URLs need to move somewhere else especially considering that they are entirely unused in `RestMiddleware`.

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
```diff
@@ -0,0 +1,191 @@
...
+    enrico_as_understood_by_bob = Enrico.from_public_keys(
+        verifying_key=data_source_public_key,
+        policy_encrypting_key=policy_pubkey
+    )
+
+    # Now Bob can retrieve the original message.
+    alice_pubkey_restored_from_ancient_scroll = UmbralPublicKey.from_bytes(alices_pubkey_bytes_saved_for_posterity)
+    delivered_cleartexts = BOB.retrieve(single_passage_ciphertext,
+                                        enrico=enrico_as_understood_by_bob,
+                                        alice_verifying_key=alice_pubkey_restored_from_ancient_scroll,
```
fjarri, line 184: There's a strange term transformation going on. First, we have "Alice's stamp", which we save as "Alice's pubkey" and later use as "Alice's verifying key". Could they all be named the same?

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
fjarri: No, just have a constant with these 25 lines and use it for the example. To skip the downloading step.

KPrasch: :heavy_check_mark:

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: :heavy_check_mark:

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: :heavy_check_mark: - used `Ursula.seednode_for_network` instead.

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: :heavy_check_mark:

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: :heavy_check_mark: - Reduced to one day.

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: No, It's a new Alice and Bob every run.

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: Valid, but I am not prepared to solve that in this PR.

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: :heavy_check_mark:

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: :heavy_check_mark:

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
fjarri: I was saying that maybe a comment is justified at this point. Although ideally yes, it should be hidden if possible.

derekpierre (approved):
🎸

examples/finnegans_wake_demo/finnegans-wake-demo-lynx.py
KPrasch: Sure - what can I say here that will be helpful / not confusing?  I'd also like to remove `client_password` from `Alice` above too, but at least on my configuration, it's still required.

fjarri: Maybe something like

> Local characters (`alice` and `bob`), as opposed to objects representing remote characters constructed from public data (`remote_alice` and `remote_bob`) run a learning loop in a background thread and need to be stopped explicitly. 

It might have been a little more intuitive if we had separate objects for "local" and "remote" characters.

Also, do we really need explicit `del`s?

KPrasch: While this is excellent documentation text which I hope is included into the official docs, I actually think adding it here is too distracting in an introductory /demo setting.

> Also, do we really need explicit dels?

IMO, Yes.  The `del` adds a very pronounced description of a unique aspect of PRE and separating actor's domains.

fjarri: For me it would be more distracting to have a `disenchant()` call on one `Alice` but not the other, without any explanation, with the method name not being exactly self-descriptive either. Up to you.

KPrasch: :heavy_check_mark: - Okay - sold.  Included this commentary in both the federated and testnet examples.

KPrasch:
Since I removed the book downloading step as an RFC, our CI has started failing since it does not have the book inside the docker container:  <https://app.circleci.com/pipelines/github/nucypher/nucypher/7191/workflows/883b1db4-bba8-4a39-a4a6-618d60b3545b/jobs/122168>.

KPrasch:
@vepkenez - Thanks for the docker fix!

@fjarri - Thanks for your thoughtful reviewing.  What do I need to do to get your approval on this PR? :smile:

fjarri:
Hm, are you planning to rename `ALICE` and `BOB`? My comment about `disenchant()` assumed they would be renamed, now it's a little out of sync.

KPrasch:
> Hm, are you planning to rename ALICE and BOB?

:heavy_check_mark: Done.

examples/finnegans_wake_demo/finnegans-wake-demo-federated.py
```diff
@@ -61,55 +51,69 @@
...
 m, n = 2, 3
 label = b"secret/files/and/stuff"
 
+
+#####################
+# Bob the BUIDLer  ##
+#####################
+
+# First there was Bob.
+LOCAL_BOB = Bob(federated_only=True,
```
fjarri, line 67: Can't say I like all caps for regular variables. What's the rationale here? To distinguish the character objects?

examples/finnegans_wake_demo/README.md
```diff
@@ -11,14 +11,25 @@ without revealing private keys to intermediary entities.
...
+Then run the demo:
+`python finnegans-wake-demo-federated.py`
 
-### Download the Book!
-`./download_finnegans_wake.sh` 
+### Testnet Demo
 
+First, configure the demo.  Be sure tat alice's address has some Goerli ETH.
+```bash
+export DEMO_PROVIDER_URI=<GOERLI RPC ENDPOINT>
```
fjarri, line 29: Are they not defined right in the examples so that the users don't forget to set them?

examples/finnegans_wake_demo/finnegans-wake-demo-federated.py
fjarri: Actually, I'm a little confused. At some point there were two variables, `BOB` and `bob`, weren't there? My comment about `bob` and `remote_bob` applied then, but doesn't make sense now. If there's only one Bob, there's probably no need in emphasizing its locality. Sorry about that.

examples/finnegans_wake_demo/README.md
KPrasch: There is an example of a seednode address in the worker docs but not an ethereum endpoint.  For demo purposes I think this will most often be a remote ethereum service like Infura.

examples/finnegans_wake_demo/finnegans-wake-demo-federated.py
KPrasch: >  At some point there were two variables, BOB and bob, weren't there?

Yep, `BOB` was renamed to `LOCAL_BOB`, and `bob` was renamed to `remote_bob`.  

> If there's only one Bob, there's probably no need in emphasizing its locality. Sorry about that.

It's no problem! :smile: Sometimes it takes a few iterations of revision and reading to get the best outcome.  There are lots of subtle communications and hints baked into this demo, it's turning out nicely.  I agree, let's try out just `alice` and `bob`, as normally names lowercase variables.  It's less dramatic/distracting.

KPrasch: :heavy_check_mark:

fjarri (approved):

cygnusv (commented):
Looking great 👍 
One question: What's the motivation to remove the client password from character init?

KPrasch:
> One question: What's the motivation to remove the client password from character init?

Passing and ethereum password into the characters init felt a bit mysterious, compared to explicitly handling/unlocking the users ethereum wallet.  It's unavoidable in the case of the transacting power since it currently depends on internal password caching to remain unlocked.  In the Finnegan's Wake demo, and in the worker documentation I've demonstrated how to handle the users ethereum wallet and keyring, passing them both as parameters, no practical need to explicitly unlock.  This gives a higher degree of user confidence and control IMO.

cygnusv (approved):

nucypher/blockchain/eth/actors.py
```diff
@@ -253,7 +252,7 @@ def __repr__(self):
     def recruit_sidekick(self, sidekick_address: str, sidekick_password: str):
         self.sidekick_power = TransactingPower(account=sidekick_address, password=sidekick_password, cache=True)
         if self.sidekick_power.is_device:
-            raise ValueError("Holy Wallet! Sidekicks can only be SW accounts")
+            raise ValueError("Holy Wallet! Sidekicks can only be SW accounts.")
```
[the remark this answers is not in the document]
KPrasch: Opened #2566

examples/finnegans_wake_demo/README.md
```diff
@@ -11,14 +11,25 @@ without revealing private keys to intermediary entities.
...
 `python ../run_demo_ursula_fleet.py`
 
+Then run the demo:
+`python finnegans-wake-demo-federated.py`
 
-### Download the Book!
-`./download_finnegans_wake.sh` 
+### Testnet Demo
 
+First, configure the demo.  Be sure tat alice's address has some Goerli ETH.
```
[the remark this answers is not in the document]
KPrasch: ✔️

PR #2562, opened 2021-02-10
raise 403 instead of 500 on invalid Alice Signature on kfrag

vepkenez:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
500s will not be raised in response to HTTP requests.
Also, 400s all have sensible messages.

**Why it's needed:**
Better http behavior.

### DIFF
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -258,7 +258,7 @@ def _make_rest_app(datastore: Datastore, this_node, domain: str, log: Logger) ->
         kfrag = KFrag.from_bytes(kfrag_bytes)
 
         if not kfrag.verify(signing_pubkey=alices_verifying_key):
-            raise InvalidSignature("{} is invalid".format(kfrag))
+            return Response(f"Signature on {kfrag} is invalid", status=403)
 
         with datastore.describe(PolicyArrangement, id_as_hex, writeable=True) as policy_arrangement:
             if not policy_arrangement.alice_verifying_key == alice.stamp.as_umbral_pubkey():
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -232,7 +232,7 @@ def _make_rest_app(datastore: Datastore, this_node, domain: str, log: Logger) ->
             cleartext = this_node.verify_from(alice, policy_message_kit, decrypt=True)
         except InvalidSignature:
             # TODO: Perhaps we log this?  Essentially 355.
-            return Response(status_code=400)
+            return Response("Invalid Signature", status_code=400)
 
         if not this_node.federated_only:
             # This splitter probably belongs somewhere canonical.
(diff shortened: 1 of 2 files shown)


KPrasch:
This PR needs a rebase over master for CI to pass.

nucypher/network/server.py
```diff
@@ -258,7 +258,7 @@ def set_policy(id_as_hex):
         kfrag = KFrag.from_bytes(kfrag_bytes)
 
         if not kfrag.verify(signing_pubkey=alices_verifying_key):
-            raise InvalidSignature("{} is invalid".format(kfrag))
+            return Response(f"Signature on {kfrag} is invalid", status=403)
```
derekpierre, line 261: Should we also do something similar on L265 where we raise `alice.SuspiciousActivity`?

vepkenez: Yes that’s a good idea.  Will add it now.

nucypher/network/server.py
```diff
@@ -246,23 +246,23 @@ def set_policy(id_as_hex):
...
 
         with datastore.describe(PolicyArrangement, id_as_hex, writeable=True) as policy_arrangement:
             if not policy_arrangement.alice_verifying_key == alice.stamp.as_umbral_pubkey():
-                raise alice.SuspiciousActivity
+                return Response(f"Policy key does not match Alice stamp.", status=403)
```
derekpierre, line 265: Tricky wording here. Perhaps something like:

```suggestion
                return Response("Policy arrangement's signing key does not match sender's", status=403)
```

nucypher/network/server.py
```diff
@@ -246,23 +246,23 @@ def set_policy(id_as_hex):
...
                 # Alice didn't pay.  Return response with that weird status code.
                 this_node.suspicious_activities_witnessed['freeriders'].append((alice, f"No transaction matching {tx}."))
-                return Response(status=402)
+                return Response(f"No paid transaction matching {tx} for this node", status=402)
 
             this_node_has_been_arranged = this_node.checksum_address in arranged_addresses
             if not this_node_has_been_arranged:
                 this_node.suspicious_activities_witnessed['freeriders'].append((alice, f"The transaction {tx} does not list me as a Worker - it lists {arranged_addresses}."))
-                return Response(status=402)
+                return Response(f"No transaction arrangement matching {tx} for this node", status=402)
```
derekpierre, line 254: wdyt?

```suggestion
                return Response(f"This node was not listed as servicing the policy in transaction {tx}", status=402)
```

derekpierre (approved):
🎸 - thanks for updating the other spots without messages.

A couple suggestions - let me know what you think.

derekpierre:
Oh, perhaps we can add a newsfragment as well since we're changing the status code.

vepkenez:
> Oh, perhaps we can add a newsfragment as well since we're changing the status code.

yeah yeah 🙄 😆

nucypher/network/server.py
```diff
@@ -246,23 +246,23 @@ def set_policy(id_as_hex):
...
                 # Alice didn't pay.  Return response with that weird status code.
                 this_node.suspicious_activities_witnessed['freeriders'].append((alice, f"No transaction matching {tx}."))
-                return Response(status=402)
+                return Response(f"No paid transaction matching {tx} for this node", status=402)
 
             this_node_has_been_arranged = this_node.checksum_address in arranged_addresses
             if not this_node_has_been_arranged:
                 this_node.suspicious_activities_witnessed['freeriders'].append((alice, f"The transaction {tx} does not list me as a Worker - it lists {arranged_addresses}."))
-                return Response(status=402)
+                return Response(f"This node was not listed as servicing the policy in transaction {tx}", status=402)
```
KPrasch, line 254: Gotta love the usage of 402 here.  I've waited a lifetime to use it.

> 402 Payment Required 
   Reserved for future use. The original intention was that this code might be used as part of some form of digital cash or micropayment scheme

KPrasch (approved):
:cowboy_hat_face:

Issue #2342, opened 2020-10-07
Require signer prefix (API, CLI, and Configuration)


KPrasch:
The implicit default signer URI of `Web3Signer` can mislead some users to creating an unintentional configuration when the uri is a filepath without a prefix.

https://gist.github.com/fjarri/bfd244f7bc814aaba385ebd473f2e2e9

PR #2572, opened 2021-02-19
Injected Transacting Power

### LINKED ISSUE #2163: Why is transacting_power composed instead of using the ordinary API?
In several places in the codebase, we have `ursula.transacting_power` instead of `ursula._crypto_power.power_ups(TransactingPower)`.  Is there a reason for this?  Either seems to work, so why the extra composition for this power?

### LINKED ISSUE #2564: Unintentional power / signer caching
In order to cache ethereum provider connections, The `BlockchainInterface` is a singleton.  Agent methods currently sign transactions using a `TransactingPower`, which is composed on the interface.  This is causes implicit fallback to the first signer cached on the interface without explicit management of the "active" signer.  In retrospect this is a clear design flaw, especially pronounced in concurrent applications.

As a temporary work-around call `character.transacting_power.activate()` before sending transactions.

****Possible Solutions****

- Register and track multiple signers on the interface
- Rework agent methods to accept a `transacting_power` kwarg
- Rework agents to return a bound `ContractFunction`s, then handle sign & send of transactions on the layer above.

### LINKED ISSUE #2566: Deprecate Preallocation Escrow SideKicks
> This reminds me that the Sidekick thing is a relic of the time where we planned to use PreallocationEscrows.

_Originally posted by @cygnusv in https://github.com/nucypher/nucypher/pull/2560#r576068876_

KPrasch:
**Type of PR:**
Bugfixes, Rework, and Deprecation

**Required reviews:** 
3

**What this does:**
- Decomposes `TransactingPower` from `BlockchainInterface`.
- Reworks Character, Actors, Agents, and Interfaces to use dependency injection for `TransactingPower`.
- Removes unlocked status tracking on `TransactingPower` - Let `Signers` handle this responsibility.
- Do not use implicit `Web3Signer` as a default for `Worker`s (Advances #2342). 
- Simplification of `StakeHolder` assimilation (Use as a simple wrapper of `Staker` instead of subclass).
- Deprecation of the `Wallet` Abstraction.
- Deprecation of Preallocation agent, deployer, agency, registry, and CLI tools.
- Require `address` in `NucypherTokenAgent.get_balance`.
- Better encapsulation and reuse of CLI-launched character password unlocking (@cygnusv).
- Use `ContractAgency` in all tests, do not return "three agents" from `agency` fixture.
- `ContractAdministrator`  is a subclass of `BaseActor`
- Use the presence of `transacting_power` or `signer` kwarg to set `is_me` and `transacting` on actors.

**Issues Closed**
- Closes #2564 
- Closes #2566
- Closes #2163 

**Why it's needed:**
- Supports multiple transaction signers to be used in the same context and improves thread safety of blockchain interfaces. 
- Deprecates unused code paths and entities


nucypher/crypto/powers.py
```diff
@@ -127,42 +126,30 @@ def __init__(self,
...
-        return self.unlock_account()
+        return self.unlock()
 
     def __exit__(self, exc_type, exc_val, exc_tb):
         return self.lock_account()
 
+    def __eq__(self, other):
+        if not isinstance(other, TransactingPower):
+            return False
+        result = bool(self.account == other.account)
```
derekpierre, line 142: I guess technically we would try to match signers as well (URI?), but Web3Signer being a passthrough would cause issues if we tried that...?

derekpierre (commented):
Looks good! 
Have to get more tea 🍵  - mostly test file changes remaining.

tests/acceptance/blockchain/interfaces/test_transacting_power.py
```diff
@@ -75,9 +63,6 @@ def test_transacting_power_sign_transaction(testerchain):
                              signer=Web3Signer(testerchain.client),
                              account=eth_address)
 
-    assert power.is_active is False
-    assert power.is_unlocked is False
```
derekpierre, line 79: How come we removed the assertions?

tests/contracts/test_contracts_upgradeability.py
```diff
@@ -102,6 +102,7 @@ def download_github_file(source_link: str, target_folder: str):
 
 def deploy_base_contract(blockchain_interface: BlockchainDeployerInterface,
                          deployer: BaseContractDeployer,
+                         transacting_power: TransactingPower,
                          skipt_test: bool):
```
derekpierre, line 106: Should this be `skip_test`...?

nucypher/crypto/powers.py
KPrasch: Hmm interesting idea.  Ultimately if the same account is controlled, it might be safe enough to assume equality. The only place this equality check is used is in a test in any case so I don't mind removing it either.

derekpierre (approved):
🎸

tests/acceptance/blockchain/interfaces/test_transacting_power.py
KPrasch: In this PR I've removed locked/unlocked status tracking on the `TransactingPower` (leaving it up to the signer directly), and the notion of activation (it used to represent being cached on the chain interface.)

tests/fixtures.py
```diff
@@ -624,10 +579,12 @@ def test_registry_source_manager(testerchain, test_registry):
...
            token_economics,
-           test_registry_source_manager) -> Tuple[NucypherTokenAgent, StakingEscrowAgent, PolicyManagerAgent]:
+           test_registry_source_manager,
+           deployer_transacting_power):
     agents = _make_agency(testerchain=testerchain,
                           test_registry=test_registry,
-                          token_economics=token_economics)
+                          token_economics=token_economics,
+                          deployer_transacting_power=deployer_transacting_power)
     yield agents
```
vzotova, line 588: I thought `_make_agency` returns nothing, how `yield` works in this case?

vzotova (commented):
Wow, a lot of good refactoring

nucypher/blockchain/eth/actors.py
```diff
@@ -234,9 +244,7 @@ def __init__(self,
             if not signer:
                 raise ValueError('signer is required to make a transacting ContractAdministrator.')
             self.deployer_power = TransactingPower(signer=signer, account=deployer_address)
-
             self.transacting_power = self.deployer_power
-            self.transacting_power.activate()
         else:
             self.deployer_power = None
             self.transacting_power = None
```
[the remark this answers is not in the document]
vzotova: Now looks much better 👍  Just curious: what changed that we don't need sidekick (two addresses) more? @KPrasch @cygnusv

vzotova (approved):

docs/source/references/environment_variables.rst
```diff
@@ -27,6 +27,13 @@ Alice
     Password for Ethereum account used by Alice.
 
 
+Bob
+----
+
+* `NUCYPHER_BOB_ETH_PASSWORD`
+    Password for Ethereum account used by Bob.
```
cygnusv, line 34: Does this PR change something wrt Bob's need of an ETH account? My understanding is that we don't require it at all.

nucypher/cli/actions/configure.py
```diff
@@ -40,7 +40,7 @@
 from nucypher.cli.types import WORKER_IP
 from nucypher.config.characters import StakeHolderConfiguration
 from nucypher.config.constants import NUCYPHER_ENVVAR_WORKER_IP_ADDRESS
-from nucypher.config.node import CharacterConfiguration
+from nucypher.config.base import CharacterConfiguration
```
cygnusv, line 43: Great renaming :)

nucypher/blockchain/eth/actors.py
```diff
@@ -109,24 +112,30 @@ class ActorError(Exception):
...
     def __init__(self,
+                 domain: Optional[str],
                  registry: BaseContractRegistry,
-                 domain: Optional[str] = None,
+                 transacting_power: Optional[TransactingPower] = None,
                  checksum_address: Optional[ChecksumAddress] = None):
 
-        # TODO: Consider this pattern - None for address?.  #1507
-        # Note: If the base class implements multiple inheritance and already has a checksum address...
+        if not bool(checksum_address) ^ bool(transacting_power):
```
cygnusv, line 120: Can you clarify operator precedence here? At first sight, I wouldn't know if this is `(not A) xor B` or `not (A xor B)`.

nucypher/blockchain/eth/token.py
```diff
@@ -782,16 +784,16 @@ def __fire_commitment(self):
         transacting_power = self.worker.transacting_power
         with transacting_power:
             txhash = self.worker.commit_to_next_period(fire_and_forget=True)  # < --- blockchain WRITE
-        self.log.info(f"Making a commitment to period {self.current_period} - TxHash: {txhash.hex()}")
-        return txhash
+            self.log.info(f"Making a commitment to period {self.current_period} - TxHash: {txhash.hex()}")
+            return txhash
```
cygnusv, line 788: Isn't it preferable to remain in the context only for the transacting operation? Just wondering what's the motivation for doing it differently.

docs/source/references/environment_variables.rst
KPrasch: There was a slight change to character configuration tests, which includes bobs that support providers.  It's not *required* but we still want to check the cases where it is optionally used.

nucypher/cli/commands/stake.py
```diff
@@ -587,22 +556,16 @@ def create(general_config: GroupGeneralConfig,
     # Last chance to bail
     click.confirm(CONFIRM_BROADCAST_CREATE_STAKE, abort=True)
 
-    # Authenticate
-    password = get_password(stakeholder=STAKEHOLDER,
-                            blockchain=blockchain,
-                            client_account=client_account,
-                            hw_wallet=transacting_staker_options.hw_wallet)
-    STAKEHOLDER.assimilate(password=password)
```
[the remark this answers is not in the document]
KPrasch: Yes, you're right, It's not longer uniform across all commands.  This was required after changing the function signature of `get_password`.  I'd prefer to improve this in a future UX PR.

cygnusv (approved):

PR #2574, opened 2021-02-20
Help nucypher-monitor get the info it needs

fjarri:
**Type of PR:**
- Other

**Required reviews:** 
- 2

**What this does:**
Some adjustments and bugfixes for node and fleet state metadata to help `nucypher-monitor`:
- `FleetSensor.record_fleet_state()` now returns `None` if the fleet state was not updated and the new `ArchivedFleetState` otherwise.
- `Nickname.to_json()` now has an `icon` field (a little redundant, but it is needed often).
- `status_info()` is now a method of `Ursula` instead of `Teacher`. It is used to get the status of the local node. It returns a `LocalUrsulaStatus` object (a namedtuple with a `to_json()` method).
- `FleetSensor.status_info(checksum_or_node)` is used for remote nodes (those from `FleetSensor` itself) and returns a `RemoteUrsulaStatus` object.
- Removed `Learner.last_seen` and keeping track of it in `FleetSensor` instead.

Paired with https://github.com/nucypher/nucypher-monitor/pull/94

### DIFF
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -19,7 +19,7 @@
 import random
 import weakref
 from collections.abc import KeysView
-from typing import Optional, Dict, Iterable, List, Tuple, NamedTuple, Union
+from typing import Optional, Dict, Iterable, List, Tuple, NamedTuple, Union, Any
 
 import binascii
 import itertools
@@ -423,7 +423,7 @@ class RemoteUrsulaStatus(NamedTuple):
     recorded_fleet_state: Optional[ArchivedFleetState]
     last_learned_from: Optional[maya.MayaDT]
 
-    def to_json(self):
+    def to_json(self) -> Dict[str, Any]:
         if self.recorded_fleet_state is None:
             recorded_fleet_state_json = None
         else:
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -400,10 +400,12 @@ class FleetSensor:
 
         recorded_fleet_state = self._remote_states.get(node.checksum_address, None)
         last_learned_from = self._remote_last_seen.get(node.checksum_address, None)
+        worker_address = node.worker_address if node.verified_node else None
 
-        return RemoteUrsulaStatus(nickname=node.nickname,
+        return RemoteUrsulaStatus(verified=node.verified_node,
+                                  nickname=node.nickname,
                                   staker_address=node.checksum_address,
-                                  worker_address=node.worker_address,
+                                  worker_address=worker_address,
                                   rest_url=node.rest_url(),
                                   timestamp=node.timestamp,
                                   last_learned_from=last_learned_from,
@@ -412,9 +414,10 @@ class FleetSensor:
 
 
 class RemoteUrsulaStatus(NamedTuple):
+    verified: bool
     nickname: Nickname
     staker_address: ChecksumAddress
-    worker_address: str
+    worker_address: Optional[ChecksumAddress]
     rest_url: str
     timestamp: maya.MayaDT
     recorded_fleet_state: Optional[ArchivedFleetState]
@@ -429,7 +432,8 @@ class RemoteUrsulaStatus(NamedTuple):
             last_learned_from_json = None
         else:
             last_learned_from_json = last_learned_from.iso8601()
-        return dict(nickname=self.nickname.to_json(),
+        return dict(verified=self.verified,
+                    nickname=self.nickname.to_json(),
                     staker_address=self.staker_address,
                     worker_address=self.worker_address,
                     rest_url=self.rest_url,
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -32,13 +32,12 @@ from nucypher.utilities.logging import Logger
 from .nicknames import Nickname
 
 
-class ArchivedFleetState:
+class ArchivedFleetState(NamedTuple):
 
-    def __init__(self, checksum: str, nickname: Nickname, timestamp: maya.MayaDT, population: int):
-        self.checksum = checksum
-        self.nickname = nickname
-        self.timestamp = timestamp
-        self.population = population
+    checksum: ChecksumAddress
+    nickname: Nickname
+    timestamp: maya.MayaDT
+    population: int
 
     def to_json(self):
         return dict(checksum=self.checksum,
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -19,7 +19,7 @@
 import random
 import weakref
 from collections.abc import KeysView
-from typing import Optional, Dict, Iterable, List, Tuple
+from typing import Optional, Dict, Iterable, List, Tuple, NamedTuple, Union
 
 import binascii
 import itertools
@@ -239,7 +239,8 @@ class FleetSensor:
 
         self._current_state = FleetState.new(this_node)
         self._archived_states = [self._current_state.archived()]
-        self.remote_states = {}
+        self._remote_states = {}
+        self._remote_last_seen = {}
 
         # temporary accumulator for new nodes to avoid updating the fleet state every time
         self._nodes_to_add = set()
@@ -329,7 +330,7 @@ class FleetSensor:
     def unpack_snapshot(data):
         return FleetState.unpack_snapshot(data)
 
-    def record_fleet_state(self, skip_this_node: bool = False):
+    def record_fleet_state(self, skip_this_node: bool = False) -> Optional[ArchivedFleetState]:
         new_state = self._current_state.with_updated_nodes(nodes_to_add=self._nodes_to_add,
                                                            nodes_to_remove=self._nodes_to_remove,
                                                            skip_this_node=skip_this_node)
@@ -344,7 +345,11 @@ class FleetSensor:
         # 2. (possible) keep a dictionary of known states
         #    and bump the timestamp of a previously encountered one
         if new_state.checksum != self._archived_states[-1].checksum:
-            self._archived_states.append(new_state.archived())
+            archived_state = new_state.archived()
+            self._archived_states.append(archived_state)
+            return archived_state
+        else:
+            return None
 
     def shuffled(self):
         return self._current_state.shuffled()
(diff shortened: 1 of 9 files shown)


nucypher/characters/lawful.py
```diff
@@ -1735,6 +1738,62 @@ def _reencrypt(self, kfrag: KFrag, work_order: 'WorkOrder', alice_verifying_key:
...
+    # Status metadata
+    #
+
+    def status_info(self, raise_invalid=True, omit_known_nodes=False, is_local=True):
+        # FIXME: is anyone using `raise_invalid=True`? Or is it always `False`?
+
+        try:
+            worker_address = self.worker_address
+        # FIXME: how did such a node end up in `known_nodes`?
+        except self.StampNotSigned:
```
KPrasch, line 1754: > how did such a node end up in `known_nodes`?

Fair point.  This was almost certainly a hot-fix once to keep the centralized monitor running.  Perhaps we can catch a more generic error here to prevent any unintentional crash behavior?

fjarri: I'm starting to think that crashing here is a desired behavior. In fact, we should have crashed much earlier. We got a node here that's invalid yet still present in `known_nodes`. It means it can be selected for granting, or to be a teacher. This is a serious problem, and we can't just sweep it under the rug like that. 

Another consideration is that we need to trust the objects we create. Otherwise, to be consistent, we would have to add `raise_invalid` to pretty much every method call. 

If we were to allow for an exception in `status_info`, this should be handled in `server.py`. I removed the `try` block from `status_info` for now, but I will wait on adding it to the status page endpoint - it creates some API complications, and I'd rather wait until we decide if it is necessary.

nucypher/acumen/perception.py
```diff
@@ -358,8 +361,14 @@ def record_remote_fleet_state(self,
                                   state_checksum: str,
                                   timestamp: maya.MayaDT,
                                   population: int):
+
+        if checksum_address not in self._current_state:
+            raise IndexError(f"A node {checksum_address} if not present in the current fleet state")
```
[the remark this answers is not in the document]
fjarri: Thanks, fixed.

nucypher/characters/lawful.py
KPrasch: This raises the question, if not here, where then do we expect this kind of exception to be raised in `server.py` - Is this essentially a side-effect of calling `NodeSprout.mature` on every peer?

fjarri: For this specific exception, I expect the node without a worker address to not appear in `known_nodes` at all and be weeded out somewhere in `remember_node()`. Whether we bucket it, or just ignore it, is details. This exception shouldn't even reach `server.py`.

fjarri: After some investigation it turns out that `FleetSensor` can indeed contain unverified nodes. This may or may not change in #2513. For now I'll just add a verified flag in the returned node status info, and make `worker_address` nullable. `nucypher-monitor` doesn't use `worker_address` anyway, so for it nothing will change (although we may add some visualization of verified/unverified nodes later). Ursula's status page will separate verified and unverified nodes.

nucypher/acumen/perception.py
```diff
@@ -93,24 +101,28 @@ def archived(self) -> ArchivedFleetState:
...
 
+        nodes_updated = []
         for node in nodes_to_add:
             if node.checksum_address in nodes_to_remove:
                 continue
-            if node.checksum_address not in self._nodes:
-                return True
-            if bytes(self._nodes[node.checksum_address]) != bytes(node):
-                return True
+            if (    node.checksum_address not in self._nodes
```
KPrasch, line 114: There are some extra whitespace characters here.

nucypher/acumen/perception.py
```diff
@@ -358,8 +378,65 @@ def record_remote_fleet_state(self,
...
+                                  staker_address=node.checksum_address,
+                                  worker_address=worker_address,
+                                  rest_url=node.rest_url(),
+                                  timestamp=node.timestamp,
+                                  last_learned_from=last_learned_from,
+                                  recorded_fleet_state=recorded_fleet_state,
+                                  )
+
+
+class RemoteUrsulaStatus(NamedTuple):
```
KPrasch, line 416: 💥

nucypher/acumen/perception.py
```diff
@@ -358,8 +378,65 @@ def record_remote_fleet_state(self,
...
+                                   timestamp=timestamp,
+                                   population=population)
+
+        self._remote_last_seen[checksum_address] = maya.now()
+        self._remote_states[checksum_address] = state
+
+    def status_info(self, checksum_address_or_node: Union[ChecksumAddress, 'Ursula']) -> 'RemoteUrsulaStatus':
+
+        if isinstance(checksum_address_or_node, str):
+            node = self[checksum_address_or_node]
```
KPrasch, line 397: Do you think it is necessary to handle a `KeyError` here in case of invalid or unexpected input?

nucypher/acumen/perception.py
```diff
@@ -358,8 +378,65 @@ def record_remote_fleet_state(self,
...
+    verified: bool
+    nickname: Nickname
+    staker_address: ChecksumAddress
+    worker_address: Optional[ChecksumAddress]
+    rest_url: str
+    timestamp: maya.MayaDT
+    recorded_fleet_state: Optional[ArchivedFleetState]
+    last_learned_from: Optional[maya.MayaDT]
+
+    def to_json(self):
```
KPrasch, line 426:
```suggestion
    def to_json(self) -> Dict[str, str]:
```

nucypher/characters/lawful.py
```diff
@@ -1735,6 +1738,62 @@ def _reencrypt(self, kfrag: KFrag, work_order: 'WorkOrder', alice_verifying_key:
...
         return cfrag_byte_stream
 
+    def status_info(self, omit_known_nodes=False) -> 'LocalUrsulaStatus':
+
+        domain = self.domain
+        version = nucypher.__version__
+
+        latest_states = self.known_nodes.latest_states(5)
+
+        fleet_state = latest_states[-1]
```
KPrasch, line 1748: What is the expected behavior here if `latest_states` is empty?

nucypher/characters/lawful.py
```diff
@@ -1735,6 +1738,62 @@ def _reencrypt(self, kfrag: KFrag, work_order: 'WorkOrder', alice_verifying_key:
         # ... and finally returns all the re-encrypted bytes
         return cfrag_byte_stream
 
+    def status_info(self, omit_known_nodes=False) -> 'LocalUrsulaStatus':
```
KPrasch, line 1741:
```suggestion
    def status_info(self, omit_known_nodes: bool = False) -> 'LocalUrsulaStatus':
```

nucypher/characters/lawful.py
```diff
@@ -1735,6 +1738,62 @@ def _reencrypt(self, kfrag: KFrag, work_order: 'WorkOrder', alice_verifying_key:
...
+    worker_address: str
+    rest_url: str
+    timestamp: maya.MayaDT
+    domain: str
+    version: str
+    fleet_state: ArchivedFleetState
+    previous_fleet_states: List[ArchivedFleetState]
+    known_nodes: Optional[List[RemoteUrsulaStatus]]
+
+    def to_json(self):
```
KPrasch, line 1781:
```suggestion
    def to_json(self) -> Dict[str, str]:
```

tests/fixtures.py
```diff
@@ -953,6 +953,9 @@ def fleet_of_highperf_mocked_ursulas(ursula_federated_test_config, request):
                 all_ursulas = {u.checksum_address: u for u in _ursulas}
 
                 for ursula in _ursulas:
+                    # FIXME: FleetSensor should not own fully-functional Ursulas.
```
KPrasch, line 956: Let's track this in an issue and record the Issue number here.

KPrasch (commented):
I absolutely love the `namedtuple` subclasses and namespacing in this PR!

nucypher/acumen/perception.py
fjarri: I was trying to align the conditions following the spirit of our codestyle... Perhaps this would be better?
```
if     (node.checksum_address not in self._nodes
        or bytes(self._nodes[node.checksum_address]) != bytes(node)):
    nodes_updated.append(node.checksum_address)
```
In my own code I'd go with
```
if (node.checksum_address not in self._nodes
        or bytes(self._nodes[node.checksum_address]) != bytes(node)):
    nodes_updated.append(node.checksum_address)
```

nucypher/acumen/perception.py
fjarri: It's more of `Dict[str, Any]`. If we were being especially formal we could define a `JSON = Union[str, int, float, NoneType, list[JSON], dict[str, JSON]]` or something.

nucypher/characters/lawful.py
fjarri: Fixed.

nucypher/characters/lawful.py
fjarri: It is never empty, one state is created on `FleetSensor` creation. Although, admittedly, it is not obvious from this context.

fjarri: To make it more obvious from the API there could be a `latest_state()` (returning a single state) and `previous_states()` (returning `states[...:-1]`) methods. That's how the info is used in building the status anyway.

nucypher/characters/lawful.py
fjarri: Fixed.

nucypher/acumen/perception.py
KPrasch: I tend to use more lines in this kind of situation, but ill chalk it up to style:

```python
unknown = node.checksum_address not in self._nodes
outdated = bytes(self._nodes[node.checksum_address]) != bytes(node)

if unknown or outdated:
    nodes_updated.append(node.checksum_address)

File: nucypher/characters/lawful.py (hunk 7 above)
Comment (KPrasch) [reply nucypher/characters/lawful.py:1745]: That sounds great.  An elegant and simple way to improve readability here.

File: nucypher/characters/lawful.py [hunk 11]
```diff
@@ -1735,6 +1735,62 @@ def _reencrypt(self, kfrag: KFrag, work_order: 'WorkOrder', alice_verifying_key:
...
+    rest_url: str
+    timestamp: maya.MayaDT
+    domain: str
+    version: str
+    fleet_state: ArchivedFleetState
+    previous_fleet_states: List[ArchivedFleetState]
+    known_nodes: Optional[List[RemoteUrsulaStatus]]
+
+    def to_json(self) -> Dict[str, Any]:
+        if self.known_nodes is None:
```
KPrasch, line 1779: Do you anticipate that `self.known_nodes` will be something other than `None` or an instance?

```suggestion
    if not self.known_nodes:
```

If it's a constant sorrow constant it's stronger to check identity against constant explicitly.

KPrasch (approved):

nucypher/characters/lawful.py
fjarri: It can potentially be an empty list (if known node info was requested, but `FleetSensor` doesn't know any nodes). So I think the check for `None` is justified here.

nucypher/acumen/perception.py
fjarri: This variant does look better, but `outdated` line will fail if `unknown` is True (and also there's a minor performance hit of eager `bytes()` calculation)

fjarri: Went for a half of this solution. At least the condition fits in one line now.

nucypher/characters/lawful.py
fjarri: Fixed.

derekpierre (approved):
🎸

PR #2583, opened 2021-02-26
Deprecates worker IP address as enviorment variable

### LINKED ISSUE #2512: Remove worker IP as enviorment variable

KPrasch:
**Type of PR:**
Deprecation

**Required reviews:** 
1

**What this does:**
Removes support of worker IP address as an environment variable.

**Issues fixed/closed:**
Closes #2512


derekpierre (commented):
We need to remove the entry from our "Environment Variables" docs - https://docs.nucypher.com/en/latest/references/environment_variables.html#ursula-worker.

The newsfragment should contain a short blurb about removing the environment variable.

KPrasch:
> We need to remove the entry from our "Environment Variables" docs - https://docs.nucypher.com/en/latest/references/environment_variables.html#ursula-worker.

👌🏻 Will do.

> The newsfragment should contain a short blurb about removing the environment variable.

Why "*should*" it?  I intentionally included a blank one here.  It will be rendered in the release notes similar to web3.py's docs: <https://web3py.readthedocs.io/en/stable/releases.html#misc>

derekpierre:
> Why "should" it? I intentionally included a blank one here. It will be rendered in the release notes similar to web3.py's docs:

I'm looking at it from the view of someone who has a set up already using the env variable, and then wondering why it is no longer having an effect - is it a bug? what changed?. Viewing the release notes which contain a blank removal-type newsfragment, with only a link to the PR without context makes it obscure to understand - as opposed to text stating that the variable was deprecated along with the link to the PR.

I'm with you though that internal type changes eg. refactoring, efficiency, testing changes etc. don't need newsfragment text.

newsfragments/2583.removal.rst
```diff
@@ -0,0 +1 @@
+Deprecates worker IP address as enviorment variable.
```
derekpierre, line 1:
```suggestion
Deprecate worker IP address as environment variable (``NUCYPHER_WORKER_IP_ADDRESS``).

```

derekpierre (commented):
Don't forget the docs 🙂 - https://docs.nucypher.com/en/latest/references/environment_variables.html#ursula-worker

derekpierre (approved):
🎸

PR #2587, opened 2021-02-27
Prevent WorkLock from failing when increasing the min sub-stake size

vzotova:
**Based on #2549**

**Type of PR:**
- [X] Feature

**Required reviews:** 
- [X] 3

**What this does:**
Allow second part of this: https://dao.nucypher.com/t/1-improve-staker-p-l-by-1-increasing-period-duration-2-editing-min-stake-size/110

**Why it's needed:**
`WorkLock` still has some stakers who did not call commit (~49) and some of them have min deposit. With current contract code this will lead to fail (`WorkLock.commit()`) if min stake size will be more than 15k even for 1 NuNit

**Notes for reviewers:**
Inceasing min sub-stake size will affect only new sub-stakes (create and divide)


cygnusv (approved):

vzotova:
Postponed for now

PR #2594, opened 2021-03-12
[EPIC] DAO proposal #1 Improve staker P/L by increasing period duration

### LINKED ISSUE #2226: Two near-term options for reducing worker gas costs
Following discussions with @vzotova. These choices are actually feasible in the near-term but neither is a barrel of laughs. 

#### 1) Separating `commitToNextPeriod` (formerly `confirmActivity`) and `mint`. 

Advantages: 
- Stakers can choose **when** to mint, thereby enabling proactive gas price tactics (i.e. waiting for low prices). This is constrained by the transaction becoming too large, which may occur as quickly as following a few dozen daily commits, up to hundreds of daily commits (a more narrow range estimate to be provided).  
- Stakers can choose a gas price corresponding to a longer execution time, given there is no longer the 24h constraint/risk. 
- Minting accounts for approximately 50% of the gas costs associated with staking, so there is scope to significantly bring down overheads. 

Disadvantages:
- Once rewards are minted, the least-bad approach to restaking is creating a new sub-stake (it is problematic to add these rewards to existing sub-stakes since that would retroactively change past reward rates, creating a discrepancy between the sum minted and what 'should have been' minted). The upshot is that that gas costs will increase at each successive `mint` since the sum of rewards is calculated by looping through (growing) number of sub-stakes. 
- Flexibility on when to execute `mint` does not guarantee any savings. Indeed it may end up being more expensive if gas prices steadily rise, given the hard limit on the `mint` transaction size, which grows with each `commitToNextPeriod`. 

Other: 
- Stakers can restake with the same cadence as that chosen for minting (as opposed to on a period-to-period basis), which in almost all cases will reduce the compounding of rewards. This decreases subsidy-based income, especially over the long-term, but **it is ultimately up to each staker, for various minting frequencies, to weigh extra gas costs against the extra income from more regular restaking.** Overall, a smaller proportion of rewards distributed to restakers may be beneficial to the network as a whole, since it weakens the advantage of larger / earlier-adopting stakers, which impedes token supply consolidation. 
- `commitToNextPeriod` costs remain the same, and depend on the A) place in line the staker's commitment is relative to others each period and B) the number of sub-stakes they've set to wind down, if any.

#### 2) Increase period duration globally (e.g. to 1 week) 

Advantages and non-disadvantages: 
- The primary model of rationality for Ursula uptime currently leaned upon by the protoocl – i.e. that relatively minuscule savings associated with being offline / ignoring requests are far outweighed by the risks of damage to the service's reliability/reputation, potentially causing depreciation of the native token's value and the bulk of stakers' assets/income early on – is basically **orthogonal to the enforced cadence** of `commitToNextPeriod`. 
- There isn't a large economic difference between a _confirmactivitybot_ executing once a day or once a week. In other words, the fact a period is 24 hours is not what is compelling Ursulas to provide a reliable service.
- A seven day period would reduce the gas costs of `commitToNextPeriod` AND `mint` operations by 7x, rather than just the latter.

Disadvantages: 
- Involves increasing the minimum temporal unit of a sharing policy. However, it is probable that the vast majority of use cases will necessitate far longer policy durations. The likely fee per sharing policy per week has to be extremely low to accommodate end-user budgets (see https://github.com/nucypher/whitepaper/pull/9), so reduced granularity should have minimal impact. Importantly, use cases that require shorter durations (e.g. vehicle-to-vehicle communication) will regardless be unsatisfied by the current minimum policy duration of 24 hours.
- Requires changes the length and breadth of the the nucypher/nucypher codebase, so may break stuff. 

Bonus option:
`commitToNextPeriod` occurs once per week, without changing the period duration. 
_details and {a,disa}dvantages to be added_

vzotova:
**Type of PR:**
Feature

**Required reviews:** 
?

**What this does:**

- [x] Period increasing and nodes migration in contracts
- [x] Adjustment of economics parameters

**Issues fixed/closed:**
Closes #2226

**Why it's needed:**
See #2226 and https://dao.nucypher.com/t/1-improve-staker-p-l-by-increasing-period-duration/110

**Notes for reviewers:**
This EPIC is to organize the proposal. Probably (?) no need to review when PR is ready, only merge


nucypher/blockchain/economics.py
```diff
@@ -55,11 +56,12 @@ class BaseEconomics:
     nunits_per_token = 10 ** __token_decimals  # Smallest unit designation
 
     # Period Definition
-    _default_hours_per_period = 24
+    _default_hours_per_period = 24 * 7
```
vepkenez, line 59: `_default_hours_per_period = 24 * os.getenv('NUCYPHER_HOURS_PER_PERIOD', 7)`

vepkenez: this is not a suggestion... just a thought.

vzotova: this variable is close to constant, consider this like different eras: genesis 1 period - 1 day, current era - 1 period - 7 days.

KPrasch:
Rebased over main @ 1a35130d651e21dcf8d273097dac3624c0b6e58d

KPrasch:
Rebased over main @ 7415324ddbfa3ab095f020883d2b820fcf01dcf9

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
```diff
@@ -154,6 +174,19 @@ contract PolicyManager is Upgradeable {
...
+    /**
+    * @notice Migrate from the old period length to the new one
+    * @param _node Node address
+    */
+    function migrate(address _node) external onlyEscrowContract {
+        NodeInfo storage nodeInfo = nodes[_node];
+        // with previous period length any previousFeePeriod will be greater than current period
+        // this is a sign of not migrated node
+        require(nodeInfo.previousFeePeriod >= getCurrentPeriod());
+        nodeInfo.previousFeePeriod = recalculatePeriod(nodeInfo.previousFeePeriod);
```
arjunhassard, line 186: Ok, so if a node has reached period 1400, but the new protocol is on period 200, then clearly they haven't yet migrated. When they do migrate,  the `recalculatePeriod` function converts to the correct period? 

This makes sense intuitively, but the genesis protocol began with period 18550, so I expected a more complicated calculation than simply multiplying the `previousFeePeriod` by `genesisSecondsPerPeriod / secondsPerPeriod` (= 1/7). Unless on the migration date both the new and old protocol will reset the period count to 0?

nucypher/blockchain/eth/token.py
```diff
@@ -752,9 +752,15 @@ def _do_work(self) -> None:
         # Measure working interval
         interval = onchain_period - self.worker.last_committed_period
         if interval < 0:
-            self.__reset_tracker_state()
-            return  # No need to commit to this period.  Save the gas.
-        if interval > 0:
+            # Handle period migrations
+            last_commitment = self.worker.last_committed_period
+            next_period = onchain_period + 1
+            if last_commitment > next_period:
```
arjunhassard, line 758: This is to warn stakers who pointlessly commit during the 2 day migration window?

arjunhassard (commented):
Fantastic work 👏🏼
Just a couple questions about period conversions and migration

nucypher/blockchain/eth/agents.py
```diff
@@ -581,8 +581,8 @@ def withdraw(self, transacting_power: TransactingPower, amount: NuNits, replace:
     @contract_api(CONTRACT_CALL)
     def get_flags(self, staker_address: ChecksumAddress) -> StakerFlags:
         flags: tuple = self.contract.functions.getFlags(staker_address).call()
-        wind_down_flag, restake_flag, measure_work_flag, snapshot_flag = flags
-        return StakerFlags(wind_down_flag, restake_flag, measure_work_flag, snapshot_flag)
+        wind_down_flag, restake_flag, measure_work_flag, snapshot_flag, migration_flag = flags
+        return StakerFlags(wind_down_flag, restake_flag, measure_work_flag, snapshot_flag, migration_flag)
```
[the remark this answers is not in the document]
vzotova: There is no way back after migration

nucypher/blockchain/eth/sol/source/contracts/PolicyManager.sol
vzotova: But this is really simple, just divide and that's it. In some moments we use idea that all genesis periods more than all new periods.

nucypher/blockchain/eth/token.py
vzotova: We can't get date of upgrading automatically, this block for finding migration case (similar case led to not committing)

derekpierre:
Before merging, we need to merge the following:
- PR #2609 
- PR #2610 

Query used:
https://github.com/nucypher/nucypher/pulls?q=is%3Apr+is%3Aopen+base%3Adao-proposal-1

KPrasch (approved):

PR #2607, opened 2021-03-22
Handle period migration in work tracker.

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
1

**What this does:**
Handles an edge case in `WorkTracker` during period extension migration.

**Why it's needed:**
Prepares nodes for graceful transition to an extended period duration.

**Notes for reviewers**
This PR is a cherry-pick from #2594

### DIFF
--- a/nucypher/blockchain/eth/token.py
+++ b/nucypher/blockchain/eth/token.py
@@ -752,9 +752,15 @@ class WorkTracker:
         # Measure working interval
         interval = onchain_period - self.worker.last_committed_period
         if interval < 0:
-            self.__reset_tracker_state()
-            return  # No need to commit to this period.  Save the gas.
-        if interval > 0:
+            # Handle period migrations
+            last_commitment = self.worker.last_committed_period
+            next_period = onchain_period + 1
+            if last_commitment > next_period:
+                self.log.warn(f"PERIOD MIGRATION DETECTED - proceeding with commitment.")
+            else:
+                self.__reset_tracker_state()
+                return  # No need to commit to this period.  Save the gas.
+        elif interval > 0:
             # TODO: #1516 Follow-up actions for missed commitments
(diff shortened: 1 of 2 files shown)


nucypher/blockchain/eth/token.py
```diff
@@ -752,9 +752,15 @@ def _do_work(self) -> None:
...
-            return  # No need to commit to this period.  Save the gas.
-        if interval > 0:
+            # Handle period migrations
+            last_commitment = self.worker.last_committed_period
+            next_period = onchain_period + 1
+            if last_commitment > next_period:
+                self.log.warn(f"PERIOD MIGRATION DETECTED - proceeding with commitment.")
+            else:
+                self.__reset_tracker_state()
+                return  # No need to commit to this period.  Save the gas.
```
derekpierre, line 762: Could you clarify what's the scenario here? Just don't have the picture in my head...

KPrasch: Sure, essentially this bit of logic will not work during the period migration:

```python
interval = onchain_period - self.worker.last_committed_period

if interval < 0:
  self.__reset_tracker_state()
  return  # No need to commit to this period.  Save the gas.
```

Period numbers will be recalculated as `epoch // 168` instead of `epoch // 24` which leads to the following scenerio:

```python 
>>> onchain_period
9353

>>> self.worker.last_committed_period
18704

>>> onchain_period - self.worker.last_committed_period
-9351
```

Given the existing logic, this inaccurately indicates that no commitment is needed, since `-9351` is indeed less than `0`.
To consider the full picture we can add a new check: if last_commitment > next_period` which will evaluate Truthy in the case that the current commitment is occurring during the migration, and proceeding with the transaction.

derekpierre: Right! 👍

derekpierre (approved):
🎸

PR #2611, opened 2021-03-23
Bring back a few bits of blockchain node status that were accidentally removed

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

**What this does:**
Returns back some local Ursula status info that was accidentally removed in #2352. Sorry @vepkenez !


KPrasch (approved):
Thanks for bringing these stats back!

derekpierre (approved):
🎸

Issue #2028, opened 2020-05-22
Bob.retrieve is too complex


KPrasch:
Radon CC Class E Score 33

KPrasch:
Related to #2244 ( @fjarri )

fjarri:
Less complex now after PR #2244 , but if #2482 comes through, I'm thinking of giving it the same treatment I gave to `Policy`.

fjarri:
Not anymore after #2730

PR #2612, opened 2021-03-24
Updates for the new Umbral API

fjarri:
**Type of PR:**
- Feature

**Required reviews:** 
- 3

**What this does:**
Makes `nucypher` use the new Umbral API.

`umbral_adapter.py` is out in the top commit, although the intermediate commit with the adapter and the minimal changes is still present, for reference.

Notable changes on on top of API adjustments:

- `scrypt`-based key generation moved from `umbral` (and from `config/keyring.py`) to `crypto/passwords.py`. 
- since `Capsule` objects are immutable, keys and cfrags are now accumulated in `UmbralMessageKit` objects. Some functions called by `Bob.retrieve()` were modified accordingly (see the changes in `lawful.py`).
- In particular, `Bob._reencrypt()` now has a bit of an awkward signature (takes an additional `message_kits_dict`), which I propose to fix in a separate PR addressing the changes from #259 (and the whole data flow in `retrieve()`)
- Using splitters defined in `crypto/splitters.py` to use Umbral objects in larger splitters, instead of passing the types directly
- `ecdsa_sign()` and `verify_ecdsa()` removed from `crypto/api.py` (using Umbral's `Signer` and `Signature` for that).
- assertions checking for specific fields in Umbral objects removed from `test_umbral_deserializer.py`

The highperf tests were broken before (see #2715), and now they are probably broken even more - I tried to modify the mocks accordingly, but it's likely I forgot something. I think they should be dealt with separately.


deploy/docker/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.8.7-slim
 
 # Update
 RUN apt update -y && apt upgrade -y
-RUN apt install patch gcc libffi-dev wget -y
+RUN apt install patch gcc libffi-dev wget git -y
```
derekpierre, line 5: I guess once we stop using your pyumbral branch and have a pyumbral release with those changes, we can remove the need for 'git'.

derekpierre (approved):
🎸 Looks good - the adapter isolation definitely made it easier to look through.

deploy/docker/Dockerfile
fjarri: Yep, it's in a separate commit and can be removed as soon as pyumbral 0.2 is out.

piotr-roslaniec (approved):

Pipfile
```diff
@@ -42,6 +42,7 @@ tabulate = "*"
 # Felix
 flask_sqlalchemy = "*"
 sqlalchemy = "*"
+umbral = {ref = "premade-public-key", git = "https://github.com/fjarri/pyumbral"}
```
cygnusv, line 45: 🚀

Pipfile
```diff
@@ -8,7 +8,6 @@ python_version = "3"
 
 [packages]
 # NuCypher
-umbral = "==0.1.3a2"
```
cygnusv, line 11: 😢

nucypher/characters/lawful.py
```diff
@@ -684,8 +682,8 @@ def get_treasure_map_from_known_ursulas(self, network_middleware, map_identifier
                 raise _MapClass.NowhereToBeFound(f"Asked {len(self.known_nodes)} nodes, but none had map {map_identifier} ")
 
     def work_orders_for_capsules(self,
-                                 *capsules,
-                                 alice_verifying_key: UmbralPublicKey,
+                                 *message_kits,
+                                 alice_verifying_key: PublicKey,
```
cygnusv, line 686: Why the API changes here from capsules to message_kits? I've read the description of the PR but I still don't get why the fact that a capsule is immutable should change anything at this level.

nucypher/crypto/passwords.py
```diff
@@ -0,0 +1,125 @@
+"""
```
cygnusv, line 1: This module is a great idea

nucypher/characters/base.py
```diff
@@ -505,9 +504,7 @@ class - whatever type of object that may be.
     def derive_federated_address(self):
         if self.federated_only:
             verifying_key = self.public_keys(SigningPower)
-            uncompressed_bytes = verifying_key.to_bytes(is_compressed=False)
-            without_prefix = uncompressed_bytes[1:]
-            verifying_key_as_eth_key = EthKeyAPI.PublicKey(without_prefix)
+            verifying_key_as_eth_key = EthKeyAPI.PublicKey.from_compressed_bytes(bytes(verifying_key))
```
derekpierre, line 507: 🚀

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -40,7 +41,7 @@ def _deserialize(self, value, attr, data, **kwargs):
 
     def _validate(self, value):
 
-        splitter = BytestringSplitter(Signature,
+        splitter = BytestringSplitter(signature_splitter,
```
derekpierre, line 44: @fjarri - I modified this for Porter, interested to get your thoughts there. -see https://github.com/derekpierre/nucypher/blob/porter-impl/nucypher/characters/control/specifications/fields/treasuremap.py#L37

nucypher/characters/lawful.py
```diff
@@ -751,34 +750,20 @@ def join_policy(self, label, alice_verifying_key, node_list=None, block=False):
...
-            # If all the capsules are now activated, we can stop here.
-            if not remaining_capsules:
-                break
-
-            if not work_order_is_useful:
-                # None of the Capsules for this particular WorkOrder need to be activated.  Move on to the next one.
-                continue
-
-            remaining_work_orders.append(work_order)
+            if not work_order.tasks.keys().isdisjoint(remaining_capsules):
```
derekpierre, line 759: 🔥

nucypher/characters/lawful.py
fjarri: Yes, you are right, this method only needs a capsule iterable.

fjarri: Fixed

Pipfile
```diff
@@ -8,11 +8,11 @@ python_version = "3"
...
 constant-sorrow = ">=0.1.0a9"
 bytestring-splitter = ">=1.3.0"
 hendrix = ">=3.4"
 lmdb = "*"
+umbral = ">=0.2"
```
KPrasch, line 15: 👏🏻

nucypher/__init__.py
```diff
@@ -24,10 +24,3 @@
 __all__ = [
     "__title__", "__summary__", "__version__", "__author__", "__license__", "__copyright__", "__email__", "__url__"
 ]
-
-
-# Set Default Curve #
-#####################
-
-from umbral.config import set_default_curve
-set_default_curve()
```
KPrasch, line 33: Good to see this removed.  This is not needed any more since `umbral >= 0.2` only uses a single curve?

nucypher/characters/lawful.py
KPrasch: Indeed, very concise and pythonic.

nucypher/characters/lawful.py
```diff
@@ -751,34 +749,20 @@ def join_policy(self, label, alice_verifying_key, node_list=None, block=False):
...
+                remaining_work_orders.append(work_order)
 
         return remaining_work_orders, remaining_capsules
 
     def _reencrypt(self,
                    work_order: 'WorkOrder',
+                   message_kits_map,
```
KPrasch, line 765: Can we include a type annotation here?

nucypher/characters/lawful.py
```diff
@@ -808,27 +792,37 @@ def _reencrypt(self,
         self._completed_work_orders.save_work_order(work_order, as_replete=retain_cfrags)
 
         the_airing_of_grievances = []
+        verified_cfrags = []
```
KPrasch, line 795: In the interest in keeping this ever-growing method tidy (it's expanded in #2687) what do you think about making this block into a method?  It also might improve testability.

nucypher/characters/lawful.py
```diff
@@ -918,30 +911,31 @@ def retrieve(self,
             # TODO Optimization: Block here (or maybe even later) until map is done being followed (instead of blocking above). #1114
             the_airing_of_grievances = []
 
-            remaining_work_orders, capsules_to_activate = self._filter_work_orders_and_capsules(
-                new_work_orders, capsules_to_activate, m)
+            remaining_work_orders, remaining_capsules = self._filter_work_orders_and_capsules(
+                new_work_orders, message_kits, m)
```
KPrasch, line 915: Phew - this method is a beast (#2028)!   We don't need to address it in this PR of course.

newsfragments/2612.misc.rst
```diff
@@ -0,0 +1 @@
+Add support for the new Umbral API
```
KPrasch, line 1: Perhaps include a version number here for posterity.

nucypher/__init__.py
fjarri: Yep, no need to worry about curves or parameter objects anymore.

nucypher/config/keyring.py
```diff
@@ -576,9 +557,7 @@ def generate(cls,
             signing_private_key, signing_public_key = _generate_signing_keys()
 
             if checksum_address is FEDERATED_ADDRESS:
-                uncompressed_bytes = signing_public_key.to_bytes(is_compressed=False)
-                without_prefix = uncompressed_bytes[1:]
-                verifying_key_as_eth_key = EthKeyAPI.PublicKey(without_prefix)
+                verifying_key_as_eth_key = EthKeyAPI.PublicKey.from_compressed_bytes(bytes(signing_public_key))
```
KPrasch, line 560: As you already know, this module has been completely reworked in #2701.  Leaving this comment as a note-to-self to revisit this area.

nucypher/crypto/keypairs.py
```diff
@@ -41,8 +39,8 @@ class Keypair(object):
     A parent Keypair class for all types of Keypairs.
     """
 
-    _private_key_source = UmbralPrivateKey.gen_key
-    _public_key_method = "get_pubkey"
+    _private_key_source = SecretKey.random
+    _public_key_method = "public_key"
```
KPrasch, line 43: Such a relief 😆.

nucypher/characters/lawful.py
fjarri: For this PR I wanted to limit the changes to Umbral compatibility as much as possible, but I did plan to go through the whole retrieval flow in a separate PR (which will include the changes for #259). The problem here is that we still need to check the cfrag somewhere within `_reencrypt()` so that we could report a fail and fall back on the next Ursula. I'm sure it can be improved, of course.

nucypher/crypto/kits.py
```diff
@@ -55,11 +54,55 @@ def __init__(self,
                  ciphertext=None,
                  signature=NOT_SIGNED) -> None:
 
+        if type(capsule) == bytes:
+            raise Exception
```
KPrasch, line 58: - I typically prefer `isinstance` since it's a little faster and considers inheritance, but perhaps neither of those considerations matter here.  If you're going to stick with `type` I suggest going right for `type(capsule) is bytes`.
- Any meaningful exception to raise here, or perhaps a message?  If not consider a TODO.

nucypher/crypto/kits.py
```diff
@@ -55,11 +54,55 @@ def __init__(self,
...
         self.capsule = capsule
         self.sender_verifying_key = sender_verifying_key
         self._signature = signature
 
+        self._cfrags = set()
+        self._delegating_key = None
+        self._receiving_key = None
+        self._verifying_key = None
+
+    def attach_cfrag(self, cfrag: VerifiedCapsuleFrag):
```
KPrasch, line 70: Type annotations for inputs and outputs will look nice here 😄

nucypher/crypto/passwords.py
KPrasch: Agreed!

nucypher/crypto/kits.py
```diff
@@ -55,11 +54,55 @@ def __init__(self,
...
+        return dict(delegating=self._delegating_key,
+                    receiving=self._receiving_key,
+                    verifying=self._verifying_key)
+
+    def __len__(self):
+        # TODO: may be better to just have a "has enough cfrags" method
+        return len(self._cfrags)
+
+    def __str__(self):
+        return f"{self.__class__.__name__}({self.capsule}, {len(self)} cfrags"
```
KPrasch, line 104: This is my favorite way to construct `__repr__` and `__str__`.

Missing closing parenthesis:

```suggestion
        return f"{self.__class__.__name__}({self.capsule}, {len(self)} cfrags)"
```

nucypher/crypto/kits.py
fjarri: I think it is a remainder from sanity checks I used for testing. I will remove it completely - we shouldn't be passing bytes around like that anyway.

nucypher/crypto/kits.py
```diff
@@ -55,11 +54,55 @@ def __init__(self,
...
+
+    def clear_cfrags(self):
+        self._cfrags = set()
+
+    def set_correctness_keys(self, delegating=None, receiving=None, verifying=None):
+
+        if self._delegating_key is None:
+            self._delegating_key = delegating
+        elif delegating is not None and delegating != self._delegating_key:
+            raise Exception("Replacing an existing delegating key")
```
KPrasch, line 82: Similar to above, might benefit from a fancier exception class, but not a big deal.  Might warrant a TODO.

nucypher/crypto/passwords.py
```diff
@@ -0,0 +1,125 @@
...
+        algorithm=__HKDF_HASH_ALGORITHM,
+        length=__WRAPPING_KEY_LENGTH,
+        salt=salt,
+        info=__WRAPPING_KEY_INFO,
+        backend=default_backend()
+    ).derive(key_material)
+    return wrapping_key
+
+
+class Scrypt:
```
KPrasch, line 54: Nice work here.  I'll end up using this in a future version of #2701.

KPrasch (approved):
Left a handful of comments and suggestions.  Looking good.

nucypher/crypto/kits.py
fjarri: Added a TODO for #2028. These are also sanity checks and can be avoided with a proper API design (so that one simply cannot re-assign the keys)

nucypher/crypto/kits.py
fjarri: Fixed

nucypher/crypto/kits.py
fjarri: Fixed

nucypher/crypto/kits.py
fjarri: Removed the check and added annotations

newsfragments/2612.misc.rst
fjarri: Fixed

nucypher/characters/lawful.py
fjarri: Fixed

cygnusv (approved):

examples/heartbeat_demo/alicia.py
```diff
@@ -102,7 +102,7 @@
 policy_pubkey = alicia.get_policy_encrypting_key_from_label(label)
 
 print("The policy public key for "
-      "label '{}' is {}".format(label.decode("utf-8"), policy_pubkey.to_bytes().hex()))
+      "label '{}' is {}".format(label.decode("utf-8"), bytes(policy_pubkey).hex()))
```
jMyles, line 105: You love to see it.

nucypher/characters/lawful.py
```diff
@@ -223,7 +221,7 @@ def generate_kfrags(self,
         bob_encrypting_key = bob.public_keys(DecryptingPower)
         delegating_power = self._crypto_power.power_ups(DelegatingPower)
         return delegating_power.generate_kfrags(bob_pubkey_enc=bob_encrypting_key,
-                                                signer=self.stamp,
+                                                signer=self.stamp.as_umbral_signer(),
```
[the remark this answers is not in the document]
fjarri: The signer has to go inside Umbral, so it has to be an Umbral object. We have an arbitrary Python callable inside Rust code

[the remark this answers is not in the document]
jMyles: To clarify my question:

Why pass the umbral signer instead of the stamp and then use its `__call__` like we were doing?  What's the advantage of the signer object that the stamp itself can't satisfy?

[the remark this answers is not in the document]
fjarri: It's called inside `generate_kfrags()`, which is a Rust function. There is no Python available at that level. The signer object is a wrapped Rust object, so it can go through.

vepkenez (approved):

PR #2616, opened 2021-03-26
Logic that deployed Proposal #1 to NuCypher DAO

### LINKED ISSUE #2426: DAO registry for mainnet
We have a DAO registry for IBEX but not for mainnet.

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [] 3

**What this does:**
This PR contains the code used to generate the DAO Proposal #1. In particular, it contains the logic for

Proposal #1 actually comprised 3 different interactions:
1. Upgrade `StakingEscrow` dispatcher to new implementation
2. Upgrade `PolicyManager` dispatcher to new implementation
3. Call `PolicyManager.setMinFeeRateRange()` to make fee rate range parameters consistent with the period extension.

In order to bundle everything into a single TX, we make use of Aragon's forwarding functionality, which uses a custom EVM-level scripting. 

From a high-level perspective, the above interactions are directed to the NuCypher contracts, however these are owned by the DAO Agent. In turn, the Agent's execution rights belong to the Voting app; in order to continue, we must use the `forward` method of the Voting app, which creates a vote that only works for those who have voting power on the `VotingAggregator`. So, the proposer need to call the `forward` method of the latter in order to interact with the Voting app. As you can see, this a sequence of enclosed calls using `forward` and EVM scripts, generated by the code in this PR. The actual sequence of interactions is encoded in the reverse order, so first `VotingAggregator.forward()`, which calls `Voting.forward()`, which calls `Agent.execute()`, which makes the desired 3 interactions.

**Issues fixed/closed:**
We need a registry of mainnet's DAO entities (Voting app, Agent, etc). Closes #2426 

**Why it's needed:**
It's needed for several reasons:
1. To serve as a "historic artifact" of the logic that deployed the proposal #1 to the DAO.
2. We will probably use something like this in the future. This was the first proposal that was complex (3 transactions in one), and this PR shows how to do it.

**Notes for reviewers:**
The code on this PR is closer to a script but for simplicity we include it as an extension of the CLI (`nucypher dao propose`)


vzotova:
Could you rebase? contracts tests should pass after that

derekpierre (approved):
🎸

piotr-roslaniec (approved):

PR #2623, opened 2021-03-30
Fix bug with merged sub-stake and migration (when last period is 1)

vzotova:
**Type of PR:**
- [X] Bugfix

**Required reviews:** 
- [X] 3

**What this does:**
Fix small bug when inactive sub-stake become active for current period only after migration. It leads to lock tokens (without reward) when it should not


KPrasch (approved):
Thank You @vzotova!

derekpierre (approved):
🎸

arjunhassard (approved):
🪕 good catch

KPrasch:
Note for reviewers: The failing CLI test on CI is unrelated.

cygnusv (approved):

PR #2625, opened 2021-04-05
Fixes intermittently failing CLI test on CI

KPrasch:
**Type of PR:**
Testing Bugfix

**Required reviews:** 
1

**What this does:**
Mature known nodes before comparing them in test to prevent intermittent failure.

### DIFF
--- a/tests/integration/network/test_node_storage.py
+++ b/tests/integration/network/test_node_storage.py
@@ -46,6 +46,7 @@ def test_one_node_stores_a_bunch_of_others(federated_ursulas, lonely_ursula_make
 
     yield deferToThread(start_lonely_learning_loop)
 
-    assert list(newcomer.known_nodes)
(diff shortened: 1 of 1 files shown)


tests/integration/network/test_node_storage.py
```diff
@@ -46,6 +46,6 @@ def start_lonely_learning_loop():
 
     yield deferToThread(start_lonely_learning_loop)
 
-    assert list(newcomer.known_nodes)
+    assert list(node.mature() for node in newcomer.known_nodes)
```
derekpierre, line 49: Is L51 the one that fails? This one asserts that there are known_nodes only?

KPrasch: > Is L51 the one that fails?

Yes

> This one asserts that there are known_nodes only?

Yep

derekpierre (approved):
🎸

PR #2626, opened 2021-04-07
Fix missing provider_uri

piotr-roslaniec:
**Type of PR:**
Bugfix

**Required reviews:** 
1

**What this does:**
Fixes `ValueError: Provider URI is required to init a decentralized character.` thrown on `Bob()` and `Alice()` constructors.

### DIFF
--- a/examples/finnegans_wake_demo/finnegans-wake-demo-testnet.py
+++ b/examples/finnegans_wake_demo/finnegans-wake-demo-testnet.py
@@ -69,7 +69,7 @@ TESTNET = 'lynx'
 
 # Then, there was bob. Bob learns about the
 # rest of the network from the seednode.
-bob = Bob(domain=TESTNET)
+bob = Bob(domain=TESTNET, provider_uri=PROVIDER_URI)
 
 # Bob puts his public keys somewhere alice can find them.
 verifying_key = bob.public_keys(SigningPower)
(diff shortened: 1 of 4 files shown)


cygnusv (approved):
Good eye! 
Thanks for noticing @piotr-roslaniec :D

jMyles (approved):
Kickass.  Two greens and two reds.  

Welcome aboard.

examples/finnegans_wake_demo/finnegans-wake-demo-testnet.py
```diff
@@ -69,7 +69,7 @@
 
 # Then, there was bob. Bob learns about the
 # rest of the network from the seednode.
-bob = Bob(domain=TESTNET)
+bob = Bob(domain=TESTNET, provider_uri=PROVIDER_URI)
```
KPrasch, line 72: As a design goal, Bob can operate entirely offchain without an ethereum provider connection.  If this is a required parameter, we'll need to take a deeper look an make it optional!

KPrasch (commented):
First PR, Welcome!  👏🏻

nucypher/characters/base.py
```diff
@@ -84,6 +84,7 @@ def __init__(self,
                  signer: Signer = None,
                  registry: BaseContractRegistry = None,
                  include_self_in_the_state: bool = False,
+                 is_offchain: bool = False,
```
KPrasch, line 87: Hmm - this kwarg becomes a bit superfluous IMO since it doesn't affect any other character's construction.  Alternately, I suggest relocating `provider_uri` into each `Character` subclass init.

piotr-roslaniec: But because Bob is the only character that works without `provider_uri` we still want to raise `ValueError('Provider URI is required to init a decentralized character.')` in the character's respective init, correct?

KPrasch: Yes, I that's why I'm thinking! 👍🏻

KPrasch (approved):
Thanks @piotr-roslaniec!

PR #2628, opened 2021-04-08
Change nickname delimiter in card filename

### LINKED ISSUE #2471: Use a period instead of colon for nickname deliniation in card filenames
Some services, like discord, will strip out characters from uploaded files.  When sharing a character card, it's essential that the filename is preserved, so a `.` character need to be used instead.

piotr-roslaniec:
**Type of PR:**
Other

**Required reviews:** 
1

**What this does:**
- Changes nickname delimiter in card filename from colon `:` to dot `.`.
- Adds a test for a QR code output.

**Issues fixed/closed:**
- Fixes #2471 

**Notes for reviewers:**
- It could break compatibility for existing card files.

### DIFF
--- a/newsfragments/2628.misc.rst
+++ b/newsfragments/2628.misc.rst
@@ -1,0 +1,1 @@
+Change filepath delimiter to dot (".") in Card Storage API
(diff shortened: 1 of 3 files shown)


derekpierre (approved):
🎸

KPrasch (approved):
🤠 - Tested live with @derekpierre

Issue #2571, opened 2021-02-16
Document our newsfragment policy/github check as part of the Contributing Guide


derekpierre:
Related to #2396 , #2415 .

Document existence of a Github Check to ensure that PRs have a newsfragment entry, and the newsfragment naming convention - https://github.com/nucypher/nucypher/blob/main/newsfragments/README.md.

PR #2629, opened 2021-04-08
Use constant for loopback address

### LINKED ISSUE #2538: Use constant for loopback address (127.0.0.1) across codebase
See https://github.com/nucypher/nucypher/pull/2536/files#r563198899.

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other - (Code Quality)

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
- Replaces occurrences of `127.0.0.1` with a `LOOPBACK_ADDRESS` constant.

**Issues fixed/closed:**
- Fixes #2538 

**Why it's needed:**

**Notes for reviewers:**
- [One occurrence](https://github.com/nucypher/nucypher/blob/main/scripts/local_fleet/run_local_ursula_fleet.py#L33) of `127.0.0.1` was ignored because it's a script without dependency on the codebase.

### DIFF
--- a/newsfragments/2629.misc.rst
+++ b/newsfragments/2629.misc.rst
@@ -1,0 +1,1 @@
+Use constant for loopback address across the codebase.
--- a/examples/run_demo_ursula_fleet.py
+++ b/examples/run_demo_ursula_fleet.py
@@ -24,11 +24,12 @@ from twisted.internet import reactor
 
 from nucypher.characters.lawful import Ursula
 from nucypher.config.constants import APP_DIR, TEMPORARY_DOMAIN
+from nucypher.utilities.networking import LOOPBACK_ADDRESS
 
 FLEET_POPULATION = 12
 DEMO_NODE_STARTING_PORT = 11500
 
-ursula_maker = partial(Ursula, rest_host='127.0.0.1',
+ursula_maker = partial(Ursula, rest_host=LOOPBACK_ADDRESS,
                        federated_only=True,
                        domain=TEMPORARY_DOMAIN)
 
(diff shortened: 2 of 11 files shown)


derekpierre:
Great issue to tackle - @piotr-roslaniec you're on fire 🔥 !

There have been some discussions (and disagreements) around which PRs require a newsfragment. I think changes that are internal / code quality  / testing / refactoring changes don't need one unless they directly affect users (app developers or node operators).

In this case, I don't think we need one, but others can speak up if they disagree. Definitely something we need to reach agreement on and officially document - perhaps within the README https://github.com/nucypher/nucypher/blob/main/newsfragments/README.md.

piotr-roslaniec:
Thanks, @derekpierre! 
I was under an impression that newsfragments are a hard requirement for any contribution (due to GH Action checks). But I think I agree with your judgment.
I actually intended to tackle a related issue: #2571. Maybe we could have a discussion in a wider circle there?

KPrasch (approved):
🤠

derekpierre (approved):
🎸  - Looks good

GH action check is there just to ensure that the addition/non-addition of a newsfragment has been thought about - it is tough to have it to understand PR context 😕

newsfragments/2629.misc.rst
```diff
@@ -0,0 +1 @@
+Use constant for loopback address across the codebase.
```
KPrasch, line 1: Have a look at the towncrier documentation (https://github.com/twisted/towncrier#philosophy). IMO this PR does not need a newsfragment.  If we had a formal changelog I think it might be better suited there.  In any case, I prefer to err on the side of verbosity.

Issue #2153, opened 2020-07-22
Implement a queries module in datastore


tuxxy:
With #2099 expected to be merged, there's a need for importable queries across the codebase. Specifically in places like Workorder metrics, arrangement pruning, etc.

These can be conveniently placed in a `queries` modules in `datastore`.

PR #2632, opened 2021-04-09
Refactor queries into a separate file

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other - Refactor

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Refactors queries previously expressed as `Datastore.query_by` into a separate file

**Issues fixed/closed:**
- Fixes: #2153

**Why it's needed:**

**Notes for reviewers:**
Some `Datastore.query_by` queries were left without change - I felt they were better expressed in their original form, being used in `test_*.py` files

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -90,8 +90,8 @@ from nucypher.crypto.powers import (
     TransactingPower
 )
 from nucypher.crypto.signing import InvalidSignature
-from nucypher.datastore.datastore import DatastoreTransactionError, RecordNotFound
-from nucypher.datastore.models import PolicyArrangement, TreasureMap as DatastoreTreasureMap
+from nucypher.datastore.datastore import DatastoreTransactionError
+from nucypher.datastore.queries import find_expired_policies, find_expired_treasure_maps
 from nucypher.network.exceptions import NodeSeemsToBeDown
 from nucypher.network.middleware import RestMiddleware
 from nucypher.network.nodes import NodeSprout, Teacher
@@ -1221,35 +1221,29 @@ class Ursula(Teacher, Character, Worker):
         """Deletes all expired arrangements, kfrags, and treasure maps in the datastore."""
         now = maya.MayaDT.from_datetime(datetime.fromtimestamp(self._datastore_pruning_task.clock.seconds()))
         try:
-            with self.datastore.query_by(PolicyArrangement,
-                                         filter_field='expiration',
-                                         filter_func=lambda expiration: expiration <= now,
-                                         writeable=True) as expired_policies:
+            with find_expired_policies(self.datastore, now) as expired_policies:
                 for policy in expired_policies:
                     policy.delete()
                 result = len(expired_policies)
-        except RecordNotFound:
-            self.log.debug("No expired policy arrangements found.")
         except DatastoreTransactionError:
             self.log.warn(f"Failed to prune policy arrangements; DB session rolled back.")
         else:
-            if result > 0:
+            if result == 0:
+                self.log.debug("No expired policy arrangements found.")
+            elif result > 0:
                 self.log.debug(f"Pruned {result} policy arrangements.")
 
         try:
-            with self.datastore.query_by(DatastoreTreasureMap,
-                                         filter_field='expiration',
-                                         filter_func=lambda expiration: expiration <= now,
-                                         writeable=True) as expired_treasure_maps:
+            with find_expired_treasure_maps(self.datastore, now) as expired_treasure_maps:
                 for treasure_map in expired_treasure_maps:
                     treasure_map.delete()
                 result = len(expired_treasure_maps)
-        except RecordNotFound:
-            self.log.debug("No expired treasure maps found.")
         except DatastoreTransactionError:
             self.log.warn(f"Failed to prune expired treasure maps; DB session rolled back.")
         else:
-            if result > 0:
+            if result == 0:
+                self.log.debug("No expired treasure maps found.")
+            elif result > 0:
                 self.log.debug(f"Pruned {result} treasure maps.")
 
     def __preflight(self) -> None:
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -1226,7 +1226,7 @@ class Ursula(Teacher, Character, Worker):
                     policy.delete()
                 result = len(expired_policies)
         except RecordNotFound:
-            self.log.debug("No expired treasure maps found.")
+            self.log.debug("No expired policy arrangements found.")
         except DatastoreTransactionError:
             self.log.warn(f"Failed to prune policy arrangements; DB session rolled back.")
         else:
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -90,7 +90,7 @@ from nucypher.crypto.powers import (
     TransactingPower
 )
 from nucypher.crypto.signing import InvalidSignature
-from nucypher.datastore.datastore import DatastoreTransactionError
+from nucypher.datastore.datastore import DatastoreTransactionError, RecordNotFound
 from nucypher.datastore.queries import find_expired_policies, find_expired_treasure_maps
 from nucypher.network.exceptions import NodeSeemsToBeDown
 from nucypher.network.middleware import RestMiddleware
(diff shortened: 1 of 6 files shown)


nucypher/datastore/queries.py
```diff
@@ -0,0 +1,56 @@
...
+ (at your option) any later version.
+
+ nucypher is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Affero General Public License for more details.
+
+ You should have received a copy of the GNU Affero General Public License
+ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+"""
```
KPrasch, line 16: This module is looking great.  It's nice to get these queries isolated and away from their usage.  Based on the inclusion of the license, I assume you've got the pre-commit and pre-push hooks working?

tests/acceptance/characters/test_freerider_attacks.py
```diff
@@ -46,12 +47,8 @@ def test_policy_simple_sinpa(blockchain_ursulas, blockchain_alice, blockchain_bo
...
         # Reset the Ursula for the next test.
         ursula.suspicious_activities_witnessed['freeriders'] = []
-        try:
-            with ursula.datastore.query_by(PolicyArrangement, writeable=True) as arrangements:
-                [arrangement.delete() for arrangement in arrangements]
-        except RecordNotFound:
-            # No records were found; this Ursula didn't have the arrangement.
-            continue
+        with find_policy_arrangements(ursula.datastore, writeable=True) as arrangements:
+            [arrangement.delete() for arrangement in arrangements]
```
KPrasch, line 51: I know this was not introduced in this PR, but I find this to be an add pattern, since it's producing a list of `None`s.  This is an abuse of list comprehensions IMO.

KPrasch (approved):

nucypher/datastore/queries.py
piotr-roslaniec: Yes, it was included automatically.

KPrasch:
I'm re-running CI now to be sure, but think this PR has is causing an integration test failure.

https://app.circleci.com/pipelines/github/nucypher/nucypher/7553/workflows/6a21496e-7852-4581-9698-17cb53eed4cd

nucypher/datastore/queries.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+def find_expired_policies(ds: Datastore, now) -> List[Type['DatastoreRecord']]:
+    try:
+        return ds.query_by(PolicyArrangement,
+                           filter_field='expiration',
+                           filter_func=lambda expiration: expiration <= now,
+                           writeable=True)
+    except RecordNotFound:
+        return []
```
cygnusv, line 32: Ah yes, I like this pattern a lot, it makes the calling code much clearer.

cygnusv (approved):

nucypher/datastore/queries.py
piotr-roslaniec: Unfortunately, this pattern doesn't work and this actually is the reason for failing tests - the evaluation of query is lazy so the `RecordNotFound` will be raised further down the road in a different scope.

I'm going to fix this.

piotr-roslaniec: I committed an alternative query pattern. It's not as ergonomic, but it contains a necessary workaround for writable records: a lot of times, we want to `find_*` records and then `record.delete()` them.

KPrasch:
@piotr-roslaniec - This PR needs a rebase over the latest changes on `main` for CI fixes.

piotr-roslaniec:
Rebase done.

nucypher/characters/lawful.py
```diff
@@ -1237,10 +1234,7 @@ def __prune_datastore(self) -> None:
                 self.log.debug(f"Pruned {result} policy arrangements.")
 
         try:
-            with self.datastore.query_by(DatastoreTreasureMap,
-                                         filter_field='expiration',
-                                         filter_func=lambda expiration: expiration <= now,
-                                         writeable=True) as expired_treasure_maps:
+            with find_expired_treasure_maps(self.datastore, now) as expired_treasure_maps:
```
tuxxy, line 1237: Wow, very nice.

tuxxy: Looks very clean, and I much prefer this pattern in this file.

nucypher/datastore/queries.py
```diff
@@ -0,0 +1,64 @@
...
+import functools
+from typing import List, Type
+
+from nucypher.datastore.base import DatastoreRecord
+from nucypher.datastore.datastore import Datastore, RecordNotFound
+from nucypher.datastore.models import PolicyArrangement, TreasureMap, Workorder
+
+
+def fetch(func):
+    """Used to fetch lazy results. Breaks writeable records."""
```
tuxxy, line 26: Can you go into more detail on "Breaks writeable records"? How so?

nucypher/datastore/queries.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def find_expired_treasure_maps(ds: Datastore, now) -> List[Type['DatastoreRecord']]:
+    return ds.query_by(TreasureMap,
+                       filter_field='expiration',
+                       filter_func=lambda expiration: expiration <= now,
+                       writeable=True)
+
+
+@fetch
```
tuxxy, line 53: tbh, I find the usage of `fetch` and `find_*` a bit confusing. It would be nice to get more details on this, as well maybe find some more intuitive terminology. It's not very clear to me the differentiation between "eager" and "lazy" here, specifically with respect to expected behavior. When would I use one or the other, especially in the case when I'm constructing another query?

Otherwise, everything here is looking 10/10.

nucypher/utilities/prometheus/collector.py
```diff
@@ -129,11 +128,9 @@ def _collect_internal(self) -> None:
...
         else:
             self.metrics["availability_score_gauge"].set(-1)
-        try:
-            with self.ursula.datastore.query_by(Workorder) as work_orders:
-                self.metrics["work_orders_gauge"].set(len(work_orders))
-        except RecordNotFound:
-            self.metrics["work_orders_gauge"].set(0)
+
+        work_orders = fetch_work_orders(self.ursula.datastore)
+        self.metrics["work_orders_gauge"].set(len(work_orders))
```
tuxxy, line 133: Wow, love the simplicity here.

tuxxy (changes requested):

nucypher/datastore/queries.py
piotr-roslaniec: @tuxxy Thanks for your comments! I added some explanation in a form of typings and docs. Please let me know if it makes it more understandable - I'll be happy to iterate on that.

tuxxy:
Just reviewed the changes from 0af3af7 -- that's definitely much more clear. So far so good.

KPrasch (approved):

nucypher/datastore/queries.py
```diff
@@ -0,0 +1,71 @@
...
+        try:
+            with func(*args, **kwargs) as results:
+                return results
+        except RecordNotFound:
+            return []
+
+    return wrapper
+
+
+def find_expired_policies(ds: Datastore, now) -> DatastoreQueryResult:
```
fjarri, line 46: Perhaps the `now` parameter is better called something like `cutoff`? `now` feels more like a value calculated on the spot in runtime. 

Also a type annotation can be added here - I believe `now` is supposed to be `maya.MayaDT`.

nucypher/datastore/queries.py
```diff
@@ -0,0 +1,71 @@
...
+
+def find_expired_treasure_maps(ds: Datastore, now) -> DatastoreQueryResult:
+    return ds.query_by(TreasureMap,
+                       filter_field='expiration',
+                       filter_func=lambda expiration: expiration <= now,
+                       writeable=True)
+
+
+@unwrap_records
+def get_work_orders(ds: Datastore) -> List[Workorder]:
```
fjarri, line 61: Should this return `DatastoreQueryResult` too?

fjarri: Ah, disregard that, I see the decorator.

nucypher/characters/lawful.py
derekpierre: +1 - that's great!

derekpierre (approved):
🎸 - very clean 🧹 

Agree with @fjarri about adding a type annotation for `now` parameters within queries.py. I'm ok with the name `now` though 🙂 .

piotr-roslaniec:
@KPrasch One of the tests didn't pass, I can also see it failed on `main` branch after merging this PR. I think this issue was introduced after rebase. Should we revert?

KPrasch:
Thanks for noticing! In this particular case the reason for the CI failure was known (the policy probationary period is ~2 weeks away and a grant CLI test has a policy duration beyond that), and a subsequent PR's CI passes resolving the issue (#2650).

PR #2634, opened 2021-04-13
Improve Rewards command

### LINKED ISSUE #2509: Improve Rewards command
The UX for obtaining rewards information from the `stake` command is very limited. There's even more info in `nucypher status`.

Some missing stuff:
- Available rewards to claim
- Claiming partial amounts
- Historic rewards per period (maybe even APY)
- Deposits and Withdrawals

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
- Moves `stake collect-reward` to `stake rewards withdraw`
- Adds `stake rewards <--periods>` command for viewing available and historical staking rewards. Example output:

```
$ nucypher stake rewards show
Available staking rewards: 1 NU.

$ nucypher stake rewards show 30
Staking rewards in the last 30 periods (210 days):
╒═════════════╤════════════════╤══════════╤══════════════╕
│ Date        │   Block Number │   Period │   Value (NU) │
╞═════════════╪════════════════╪══════════╪══════════════╡
│ Jun 25 2257 │      632349976 │    15000 │         1    │
├─────────────┼────────────────┼──────────┼──────────────┤
│ Jun 18 2257 │      632306676 │    14999 │         1.01 │
├─────────────┼────────────────┼──────────┼──────────────┤
│ Jun 11 2257 │      632263376 │    14998 │         1.04 │
╘═════════════╧════════════════╧══════════╧══════════════╛
Total staking rewards: 3.05 NU.

$ nucypher stake rewards show --periods 10 # no rewards there
No staking rewards found.
```

**Issues fixed/closed:**
- Fixes: #2509

**Why it's needed:**

**Notes for reviewers:**
Tested locally on account with no past staking rewards.


nucypher/cli/commands/stake.py
```diff
@@ -1348,3 +1347,83 @@ def mint(general_config: GroupGeneralConfig,
...
+
+
+@rewards.group()
+@group_staker_options
+@option_config_file
+@group_general_config
+@click.option('--tokens', help="Deposit NU tokens", type=click.STRING)
+@click.option('--amount', help="Amount to deposit", type=click.FLOAT)
+def deposit(general_config, staker_options, config_file, tokens, amount):
+    """Deposit rewards."""
```
KPrasch, line 1412: I cannot seem to find the original explainer comment - what will be the role of this command?

piotr-roslaniec: There's an ongoing [discussion](https://github.com/nucypher/nucypher/issues/2509#issuecomment-817954554) about that, but I suspect it's superfluous.

KPrasch:
Looks like this PR needs a rebase to resolve conflicts and possibly fix failing CI tests.

tests/integration/cli/test_stake_cli_functionality.py
```diff
@@ -1403,3 +1407,70 @@ def test_stake_list_all(click_runner, surrogate_stakers, surrogate_stakes, token
...
+
+    mock_staking_agent.calculate_staking_reward.assert_called_once_with(staker_address=surrogate_stakers[0])
+
+
+@pytest.mark.usefixtures("test_registry_source_manager", "patch_stakeholder_configuration")
+def test_show_rewards_for_period(click_runner, surrogate_stakers, mock_staking_agent, mocker):
+    days = 30
+    seconds_in_day = 24 * 60 * 60
+    periods_per_day = 2
+    seconds_per_period = periods_per_day * seconds_in_day
```
derekpierre, line 1436: Not sure I understand the `periods_per_day` value. If you use economics (see previous comment), you can mock it and use your own values if you like for the test, or use the StandardTokenEconomics class.

derekpierre (commented):
Good job jumping into this! 💪

nucypher/cli/painting/staking.py
```diff
@@ -267,3 +272,42 @@ def paint_min_rate(emitter, staker):
     ~ Previously set ....... {prettify_eth_amount(raw_minimum)}
     ~ Effective ............ {prettify_eth_amount(minimum)}"""
     emitter.echo(rate_payload)
+
+
+def paint_staking_rewards(STAKEHOLDER, blockchain, emitter, period, staking_address, staking_agent):
```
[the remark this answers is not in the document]
piotr-roslaniec: I thought it was some sort of convention in this file.

piotr-roslaniec:
@derekpierre Thanks for your comments! I've updated examples in the issue description.

Rebase is done.

nucypher/cli/painting/staking.py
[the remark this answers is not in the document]
derekpierre: I believe the caps are only really used in `cli.commands` package.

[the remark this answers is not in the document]
KPrasch: I think we can move away from this convention now too, It has served it's purpose 😄

nucypher/cli/painting/staking.py
```diff
@@ -267,3 +274,50 @@ def paint_min_rate(emitter, staker):
     ~ Previously set ....... {prettify_eth_amount(raw_minimum)}
     ~ Effective ............ {prettify_eth_amount(minimum)}"""
     emitter.echo(rate_payload)
+
+
+def paint_staking_rewards(stakeholder, blockchain, emitter, num_past_periods, staking_address, staking_agent):
+    if num_past_periods:
+        economics = stakeholder.staker.economics
+        seconds_per_period = economics.seconds_per_period
+        days_per_period = economics.hours_per_period // 24
```
KPrasch, line 283: Consider moving `days_per_period` into a constant or attribute on `economics`, it may be useful elsewhere in the codebase or tests.

nucypher/cli/commands/stake.py
```diff
@@ -1399,3 +1400,35 @@ def migrate(general_config: GroupGeneralConfig,
...
+@stake.group()
+def rewards():
+    """Manage staking rewards."""
+
+
+@rewards.command('show')
+@group_staker_options
+@option_config_file
+@group_general_config
+@click.option('--num_past_periods', help="Number of past periods for which to calculate rewards", type=click.INT)
```
KPrasch, line 1419: Use hyphens instead of underscores for CLI options please 😄 

```suggestion
@click.option('--num-past-periods', help="Number of past periods for which to calculate rewards", type=click.INT)
 ```
 
 Also - what do you think about shortening this option to just `--periods`?  I tend to prefer shorter CLI flags.
 
 ```suggestion
@click.option('--periods', help="Number of past periods for which to calculate rewards", type=click.INT)
 ```

**Comment by KPrasch** [comment]
Let's include a news fragment for this PR since it introduces new CLI functionality.

**Review by KPrasch — COMMENTED** [review]
Looking good so far - The documentation may need an update to reflect this new functionality!

**Comment by piotr-roslaniec** [comment]
Should we should move `stake collect-reward` to `stake rewards collect` command?

**Comment by KPrasch** [comment]
This is a reasonable place and time to make that change IMO.  Also, after some consideration I think it's better named "withdraw" since the unlocked amount may not be rewards alone, it might also be the originally staked NU.

File: nucypher/cli/literature.py [hunk 6]
```diff
@@ -311,6 +311,14 @@
 
 NO_FEE_TO_WITHDRAW = "No policy fee can be withdrawn."
 
+TOKEN_REWARD_CURRENT = 'Available staking rewards: {reward_amount:.18g} NU.'
+
+TOKEN_REWARD_PAST_HEADER = 'Staking rewards in the last {periods} periods ({days} days):'
+
+TOKEN_REWARD_PAST = 'Total staking rewards: {reward_amount:.18g} NU.'
```
cygnusv, line 318: Rather than using floats, for dealing with NU tokens it's better to resort to the `NU` class is `nucypher.blockchain.eth.token`, which allows you to easily convert between NU and NuNits. It implements `__round__`, so you can say something like `round(amount, 2)` to get up to 2 decimal places.

nucypher/cli/painting/staking.py
```diff
@@ -267,3 +274,50 @@ def paint_min_rate(emitter, staker):
...
+            ])
+            rewards_total += token_reward
+
+        periods_as_days = math.floor(days_per_period * num_past_periods)
+        if rows:
+            emitter.echo(message=TOKEN_REWARD_PAST_HEADER.format(periods=num_past_periods, days=periods_as_days))
+            emitter.echo(tabulate.tabulate(rows,
+                                           headers=REWARDS_TABLE_COLUMNS,
+                                           tablefmt="fancy_grid",
+                                           floatfmt=".18g"))
```
cygnusv, line 317: Same comment as above.

tests/integration/cli/test_stake_cli_functionality.py
```diff
@@ -1403,3 +1412,95 @@ def test_stake_list_all(click_runner, surrogate_stakers, surrogate_stakes, token
...
+    assert result.exit_code == 0
+    assert TOKEN_REWARD_CURRENT.format(reward_amount=reward_amount) in result.output
+
+    mock_staking_agent.calculate_staking_reward.assert_called_once_with(staker_address=surrogate_stakers[0])
+
+
+@pytest.mark.usefixtures("test_registry_source_manager", "patch_stakeholder_configuration")
+def test_show_rewards_for_period(click_runner, surrogate_stakers, mock_staking_agent, token_economics, mocker):
+    num_past_periods = 30
+    periods_per_day = token_economics.hours_per_period / 24
```
cygnusv, line 1439:
```suggestion
    periods_per_day = token_economics.hours_per_period // 24
```

cygnusv:
> This is a reasonable place and time to make that change IMO. Also, after some consideration I think it's better named "withdraw" since the unlocked amount may not be rewards alone, it might also be the originally staked NU.

💯  I don't know how we never thought about this!

KPrasch (changes requested):
Got the following exception while trying to run `nucypher stake rewards show --periods 28`.  Not that this did not happen without the `--periods` flag.

```
nucypher stake rewards show --periods 28
...
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/bin/nucypher", line 33, in <module>
    sys.exit(load_entry_point('nucypher', 'console_scripts', 'nucypher')())
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/click/core.py", line 829, in __call__
    return self.main(*args, **kwargs)
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/click/core.py", line 782, in main
    rv = self.invoke(ctx)
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/click/core.py", line 1259, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/click/core.py", line 1066, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/click/core.py", line 610, in invoke
    return callback(*args, **kwargs)
  File "/home/kieran/Git/nucypher/nucypher/cli/options.py", line 169, in wrapper
    return func(**kwargs)
  File "/home/kieran/Git/nucypher/nucypher/cli/options.py", line 169, in wrapper
    return func(**kwargs)
  File "/home/kieran/Git/nucypher/nucypher/cli/options.py", line 169, in wrapper
    return func(**kwargs)
  File "/home/kieran/Git/nucypher/nucypher/cli/commands/stake.py", line 1354, in show_rewards
    paint_staking_rewards(stakeholder, blockchain, emitter, periods, staking_address, staking_agent)
  File "/home/kieran/Git/nucypher/nucypher/cli/painting/staking.py", line 308, in paint_staking_rewards
    int(event_record['block_number']),
  File "/home/kieran/.local/share/virtualenvs/nucypher-9CH2v_st/lib/python3.8/site-packages/web3/datastructures.py", line 51, in __getitem__
    return self.__dict__[key]  # type: ignore
KeyError: 'block_number'
```

newsfragments/2634.feature.rst
```diff
@@ -0,0 +1,2 @@
+Adds a new CLI command to show past and present staking rewards, "stake rewards show".
+Moves "stake collect-reward" to "stake rewards withdraw" command.
```
derekpierre, line 2: I think this second line entry can be put in a `2634.removal.rst` newsfragment (really as a deprecation not a removal). Although @KPrasch may feel differently 😜

nucypher/cli/painting/staking.py
```diff
@@ -267,3 +274,49 @@ def paint_min_rate(emitter, staker):
...
+        rewards_total += token_reward
+
+    if not rows:
+        emitter.echo(TOKEN_REWARD_NOT_FOUND)
+        return
+
+    periods_as_days = math.floor(economics.days_per_period * past_periods)
+    emitter.echo(message=TOKEN_REWARD_PAST_HEADER.format(periods=past_periods, days=periods_as_days))
+    emitter.echo(tabulate.tabulate(rows, headers=REWARDS_TABLE_COLUMNS, tablefmt="fancy_grid"))
+    rewards_total = NU(rewards_total, 'NU').to_tokens()
```
derekpierre, line 321: you don't have to convert NU object to a decimal, it can be represented as a string, and takes care of the details for you eg.
```python
In [1]: from nucypher.blockchain.eth.token import NU

In [2]: n = NU(1_000_000, 'NU')

In [3]: print(f"Total rewards: {n}")
Total rewards: 1000000 NU
```

You can save yourself from calling `to_tokens()` and remove the `NU` text in `TOKEN_REWARD_PAST` string.

derekpierre: Same comment/idea for `TOKEN_REWARD_CURRENT` higher up in the file.

nucypher/types.py
```diff
@@ -42,6 +42,7 @@ class WorklockParameters(Tuple):
 
 class StakingEscrowParameters(Tuple):
     seconds_per_period: int
+    days_per_period: int
```
derekpierre, line 45: I'm not sure why this change is needed?

@vzotova / @cygnusv are there broad implications to adding something here that modifies indices of other other values in the Tuple? Some values from this tuple are retrieved via indices, if I'm not mistaken, so changing the indices could cause issues.

nucypher/cli/painting/staking.py
derekpierre: and incorporate `round` as @cygnusv indicated. 

```python
In [9]: print(f"Total rewards: {round(n, 2)}")
Total rewards: 1000.23 NU

In [10]: n = NU(1000.23132132313, 'NU')

In [11]: print(f"Total rewards: {n}")
Total rewards: 1000.23132132313 NU

In [12]: print(f"Total rewards: {round(n, 2)}")
Total rewards: 1000.23 NU

In [13]: print(f"Total rewards: {round(n, 5)}")
Total rewards: 1000.23132 NU
```

nucypher/types.py
piotr-roslaniec: Yes, this is a mistake and actually a required change.

piotr-roslaniec:
> Got the following exception while trying to run `nucypher stake rewards show --periods 28`. Not that this did not happen without the `--periods` flag. I was using Infura as an ethereum provider for this review - Perhaps this is an RPC endpoint difference with our test backend?

@KPrasch This is actually a mistake. The `KeyError` is caused by using `block_number` instead of `blockNumber`. The `event_record` dict is following Solidity naming convention.

Nevertheless I appreciate you testing since I don't have enough rewards to test it locally.

newsfragments/2634.feature.rst
piotr-roslaniec: I think it's fair because this is a breaking change.

docs/source/references/cli_reference.rst
```diff
@@ -166,6 +166,15 @@ Manage stakes and other staker-related operations.
...
 +-------------------------+---------------------------------------------+
 
+**Stake Rewards Command Actions**
+
++----------------------+-------------------------------------------------------------------------------+
+| Action               |  Description                                                                  |
++======================+===============================================================================+
+|  ``show``            | Show available and historical staking rewards                                 |
++----------------------+-------------------------------------------------------------------------------+
+|  ``withdraw``        | Withdraw staking compensation from the contract to your wallet                |
```
cygnusv, line 176: I love the documentation changes <3  The command is much clearer now. Just a minor comment here:
```suggestion
|  ``withdraw``        | Withdraw unlocked tokens from the contract to your wallet                |
```

docs/source/staking/stake_management.rst
```diff
@@ -422,15 +422,15 @@ Collect Staker Rewards
 ~~~~~~~~~~~~~~~~~~~~~~
 
 NuCypher nodes earn two types of rewards: staking rewards (in NU) and policy fees (i.e., service fees in ETH).
-To collect these rewards use ``nucypher stake collect-reward`` with flags ``--staking-reward`` and ``--policy-fee``
+To collect these rewards use ``nucypher stake rewards withdraw`` with flags ``--staking-reward`` and ``--policy-fee``
```
cygnusv, line 425: What do you think about simplifying the `--policy-fee` flag to merely `--fees` and `--staking-reward` to `--tokens`? If not, the original flags should be plural, I just realized that.

cygnusv (commented):
Looking great!

docs/source/staking/stake_management.rst
piotr-roslaniec: Sounds good. We should also rename negative switches, so `--no-tokens` and `--no-fees`.

nucypher/cli/painting/staking.py
```diff
@@ -267,3 +275,49 @@ def paint_min_rate(emitter, staker):
...
+            int(period),
+            round(token_reward, TOKEN_DECIMAL_PLACE),
+        ])
+        rewards_total += token_reward
+
+    if not rows:
+        emitter.echo(TOKEN_REWARD_NOT_FOUND)
+        return
+
+    periods_as_days = math.floor(economics.days_per_period * past_periods)
```
derekpierre, line 319: I don't think you need `math.floor` anymore since the `economics.days_per_period` calculation already does a floor operation ('//')
```python
self.days_per_period = hours_per_period // 24
```

```suggestion
    periods_as_days = economics.days_per_period * past_periods
```

nucypher/cli/painting/staking.py
```diff
@@ -267,3 +275,49 @@ def paint_min_rate(emitter, staker):
...
+                                                  latest_block=latest_block)
+
+    argument_filters = {'staker': staking_address}
+    event_type = staking_agent.contract.events['Minted']
+    entries = event_type.getLogs(fromBlock=from_block,
+                                 toBlock='latest',
+                                 argument_filters=argument_filters)
+
+    rows = []
+    rewards_total = 0
```
derekpierre, line 302: You can save yourself the hassle of having to later convert this value to NU (L322) by just tracking it as a NU object. Arithmetic operations can be performed on NU objects.

For example
```python
In [24]: n = NU(0, 'NU')

In [25]: n += NU(1000.123, 'NU')

In [26]: print(f"{n}")
1000.123 NU

In [27]: n += NU(2345.456, 'NU')

In [28]: print(f"{n}")
3345.579 NU
...
In [38]: nunit_value = NU(123456789123456789123, 'NuNit')

In [39]: str(nunit_value)
Out[39]: '123.456789123456789123 NU'

In [40]: n += nunit_value

In [41]: str(n)
Out[41]: '3469.035789123456789123 NU'
```

derekpierre (commented):
Really good stuff here @piotr-roslaniec .

Seems like something is up with the dates - this is from mainnet:

```
╒═════════════╤════════════════╤══════════╤══════════════╕
│ Date        │   Block Number │   Period │   Value (NU) │
╞═════════════╪════════════════╪══════════╪══════════════╡
│ Nov 09 2327 │       11858179 │    18672 │     <value>  │
├─────────────┼────────────────┼──────────┼──────────────┤
│ Nov 16 2327 │       11865285 │    18673 │     <value>  │
├─────────────┼────────────────┼──────────┼──────────────┤

...

│ Dec 19 2328 │       12234999 │    18730 │     <value>  │
├─────────────┼────────────────┼──────────┼──────────────┤
│ Apr 14 2021 │       12286723 │     2676 │     <value>  │
╘═════════════╧════════════════╧══════════╧══════════════╛
...
```

We may have to use the `economics.genesis_seconds_per_period` for periods prior to the period length change.

Perhaps a check like (?)
if _event_blocknumber > network launch block number (constant)_ and _event_period number > first period at network launch (constant)_  
      then use gensis_seconds_per_period
else use current seconds per period ...

piotr-roslaniec:
> if _event_blocknumber > network launch block number (constant)_ and _event_period number > first period at network launch (constant)_
> then use gensis_seconds_per_period
> else use current seconds per period ...

@derekpierre  For selecting the period duration shouldn't  we use the date for the enactment of the [7 days period duration](https://docs.nucypher.com/en/latest/architecture/period_duration.html#day-period)? Example:

if _event_blocknumber > block_number_at_date("2021-04-15T00:00:00.0Z")_
then use gensis_seconds_per_period
else use current seconds per period ...

block_number_at_date("2021-04-15T00:00:00.0Z") would equal 12239394 according to this [transaction](https://etherscan.io/tx/0xa32a1a12d06fc9dc880cfdcf02675606b363afbbd02884fb6eb1b7f99e610383) from NuCypher DAO agent.

derekpierre:
> shouldn't we use the date for the enactment of the 7 days period duration?
> ...
> block_number_at_date("2021-04-15T00:00:00.0Z") would equal 12239394

🤔 . It's possible, but technically inexact - but may still work. The period migration occurred after 3pm UTC on April 14th. At the moment of the upgrade, the period numbers were switched and any subsequent events moving forward would use the update period number. Technically, there was a shortened period between the upgrade of the contracts and April 15th midnight UTC - although for only `Minted` events there wouldn't have been any in that shortened period, because there were no prior commitments.


Come to think of it, can we simply use `blockNumber` to determine the timestamp?

```python
>>> staking_agent.blockchain.client.w3.eth.getBlock(11858179).timestamp
1613350355
>>> n = MayaDT(epoch=1613350355)
>>> n.rfc3339()
'2021-02-15T00:52:35.0Z'
>>> n.local_datetime().strftime("%b %d %Y")
'Feb 14 2021'
```

piotr-roslaniec:
@derekpierre Yes, you're right - block `timestamp` is enough. I think I initially misunderstood how periods were calculated after migration and this lead to overcomplicated logic.

nucypher/blockchain/eth/clients.py
```diff
@@ -386,6 +386,9 @@ def get_blocktime(self):
         now = highest_block['timestamp']
         return now
 
+    def get_block(self, block_number):
+        return self.w3.eth.getBlock(block_number)
```
derekpierre, line 390: The convention for `web3.eth.getBlock(...)` is that it can take a block number, or 'latest', or a hash, and takes a `block_identifier` as the parameter. We should do the same.

```suggestion
    def get_block(self, block_identifier):
        return self.w3.eth.getBlock(block_identifier)
```

nucypher/cli/painting/staking.py
```diff
@@ -267,3 +275,50 @@ def paint_min_rate(emitter, staker):
...
+        event_block_number = int(event_record['blockNumber'])
+        event_period = event_record['args']['period']
+        event_reward = NU(event_record['args']['value'], 'NuNit')
+        timestamp = blockchain.client.get_block(event_block_number).timestamp
+        event_date = maya.MayaDT(epoch=timestamp).local_datetime().strftime("%b %d %Y")
+        rows.append([
+            event_date,
+            event_block_number,
+            int(event_period),
+            round(event_reward.to_tokens(), TOKEN_DECIMAL_PLACE),
```
derekpierre, line 313: Don't need `.to_tokens()`
```suggestion
            round(event_reward, TOKEN_DECIMAL_PLACE),
```

docs/source/references/cli_reference.rst
```diff
@@ -166,6 +166,15 @@ Manage stakes and other staker-related operations.
...
 +-------------------------+---------------------------------------------+
 
+**Stake Rewards Command Actions**
+
++----------------------+-------------------------------------------------------------------------------+
+| Action               |  Description                                                                  |
++======================+===============================================================================+
+|  ``show``            | Show available and historical staking rewards                                 |
++----------------------+-------------------------------------------------------------------------------+
+|  ``withdraw``        | Withdraw unlocked tokens from the contract to your wallet                     |
```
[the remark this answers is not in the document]
derekpierre: We should update the text to account for fees as well, not just staking rewards.

derekpierre (commented):
We're in the 5-yard line, almost there!

derekpierre:
I ran the `rewards show` command and the dates look good! 🎸

nucypher/cli/commands/stake.py
```diff
@@ -1399,3 +1320,116 @@ def migrate(general_config: GroupGeneralConfig,
...
+
+    paint_staking_rewards(stakeholder, blockchain, emitter, periods, staking_address, staking_agent)
+
+
+@rewards.command('withdraw')
+@group_transacting_staker_options
+@option_config_file
+@click.option('--replace', help="Replace any existing pending transaction", is_flag=True)
+@click.option('--tokens/--no-tokens', is_flag=True, default=False)
+@click.option('--fees/--no-fees', is_flag=True, default=False)
```
derekpierre, line 1362: Need some help text here.

derekpierre:
There is one thing swirling around in my head. 

- If I want to see my currently unlocked staking rewards, I run `nucypher stake rewards show`, and that returns `Available staking rewards: ...` 👍 
- If I want to withdraw my staking rewards, I run `nucypher stake rewards withdraw --tokens` 👍 
- If I want to withdrawn my fees, I run `nucypher stake rewards withdraw --fees` 👍 
- If I want to see my withdrawable fees...? 😕 🤷 . This is available as part of `nucypher status stakers` but that is a separate command. Given the prior functionality did not provide this anyway (`nucypher stake collect-reward`) this could be addressed/thought about in a separate PR.

Just something to think about.

cygnusv (approved):
LGTM! I also tested locally to see my rewards and historic info. Haven't tried to use it to withdraw but looks good. Great job!

derekpierre (approved):
🎸 - haven't tested `withdraw` either - presumably it was basically a copy and paste from before.

nucypher/cli/commands/stake.py
```diff
@@ -1399,3 +1320,118 @@ def migrate(general_config: GroupGeneralConfig,
...
+            click.confirm(CONFIRM_COLLECTING_WITHOUT_MINTING, abort=True)
+
+        # Authenticate and Execute
+        password = get_password(stakeholder=STAKEHOLDER,
+                                blockchain=blockchain,
+                                client_account=client_account,
+                                hw_wallet=transacting_staker_options.hw_wallet)
+        STAKEHOLDER.assimilate(checksum_address=client_account, password=password)
+
+        staking_receipt = STAKEHOLDER.staker.collect_staking_reward(replace=replace)
```
KPrasch, line 1413: Minor suggestion, a confirmation prompt might be a nice addition here.

KPrasch (approved):
I've tested all commands manually, looking good.  

Thanks @piotr-roslaniec!

PR #2637, opened 2021-04-14
Use an estimation multiplier for mint.

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
1

**What this does:**
Use a 25% gas estimation multiplier for mint.

**Issues fixed/closed:**
Related to #2337

**Why it's needed:**
Avoids intermittent insufficient gas estimation for mint transactions.

### DIFF
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -627,7 +627,7 @@ class BlockchainInterface:
                          transacting_power: TransactingPower,
                          payload: dict = None,
                          transaction_gas_limit: Optional[int] = None,
-                         gas_estimation_multiplier: Optional[float] = 1.25,  # TODO: Workaround for #2635, #2337
+                         gas_estimation_multiplier: Optional[float] = 1.15,  # TODO: Workaround for #2635, #2337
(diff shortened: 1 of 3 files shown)


cygnusv (commented):
I'm thinking we just should use a default multiplier always.

derekpierre:
Based on the failing test, it looks like having a default value (not None) for `gas_estimation_multiplier` causes issues when a fixed transaction gas limit is provided.

```
ValueError: 'transaction_gas_limit' and 'gas_estimation_multiplier' can't be used together.
```

nucypher/blockchain/eth/interfaces.py
```diff
@@ -627,7 +628,7 @@ def send_transaction(self,
                          transacting_power: TransactingPower,
                          payload: dict = None,
                          transaction_gas_limit: Optional[int] = None,
-                         gas_estimation_multiplier: Optional[float] = None,
+                         gas_estimation_multiplier: Optional[float] = 1.15,  # TODO: Workaround for #2635, #2337
```
KPrasch, line 631: 👍🏻

derekpierre (approved):
🎸

PR #2651, opened 2021-04-16
Post migration contracts changes

### LINKED ISSUE #2579: Clean contracts after upgrading up to "migration" update
Some places in `StakingEscrow` and `PolicyManager` is hacky to make possible `verifyState` works (and not only)
Remove them only after upgrade happened

### LINKED ISSUE #2580: Period overflow in contract tests
Some contract test use 1 hour as period length which lead to overflow (though tests still work well). Consider setting minimum possible value for period in constructor as 1 day

### LINKED ISSUE #2638: Allow worker to call StakingEscrow.mint()
We need to soften the restrictions on this method, particularly to fix one of the issues with #1422. 
A safe requirement could be either the staker is the caller, or the worker at a terminal period.

vzotova:
**Type of PR:**
- [X] Bugfix
- [X] Feature

**Required reviews:** 
- [X] 3

**What this does:**
Proper fix for `previousFeePeriod` migration bug.
Removed `verifyState` migration-only updates from contracts.
Softened restrictions for `mint()` method.
Additional restriction for hours per period to prevent potential overflow during period calculations.

**Issues fixed/closed:**
Fixes #2579 
Fixes #2638
Fixes #2580


KPrasch:
Let's give this PR a rebase to check for a passing CI suite.

nucypher/blockchain/eth/sol/source/contracts/Issuer.sol
```diff
@@ -98,7 +98,8 @@ abstract contract Issuer is Upgradeable {
         require(localTotalSupply > 0 &&
             _issuanceDecayCoefficient != 0 &&
             _hoursPerPeriod != 0 &&
-            _genesisHoursPerPeriod != 0 &&
+            // prevent overflow during period calculations
+            _genesisHoursPerPeriod >= 10 &&
```
derekpierre, line 102: 😕 ...could you expand on why 10? testing?

vzotova: it comes from minimum seconds per period when overflow occurs (around 6-7 hours), so I took value of 10 hours to have some periods reserve for time traveling (in worst case)

tests/contracts/contracts/migration/PolicyManagerOld.sol
```diff
@@ -719,39 +719,7 @@ contract PolicyManagerOld is Upgradeable {
     /// @dev the `onlyWhileUpgrading` modifier works through a call to the parent `verifyState`
     function verifyState(address _testTarget) public override virtual {
```
fjarri, line 720: I am not sure about Solidity syntax, but if it just calls the superclass, does it need to be overridden at all?

tests/contracts/contracts/migration/StakingEscrowOld.sol
```diff
@@ -1630,63 +1630,7 @@ contract StakingEscrowOld is IssuerOld, IERC900History {
     /// @dev the `onlyWhileUpgrading` modifier works through a call to the parent `verifyState`
     function verifyState(address _testTarget) public override virtual {
```
fjarri, line 1631: Same question here — if it just calls the superclass, does it need to be overridden at all?

tests/contracts/contracts/migration/PolicyManagerOld.sol
vzotova: You're right, it was not necessary, I just wanted to show there's no verifyState check now (and this is hack), but comment will be enough

vzotova: ✔️

tests/contracts/contracts/migration/StakingEscrowOld.sol
vzotova: ✔️

derekpierre (approved):
🎸

vzotova:
This PR will be useful only in case intermediary upgrade before #2775. Close for now

Issue #2269, opened 2020-09-18
Rich Trezor Signer URIs


KPrasch:
Support additional trezor configuration data via Signer URI - Here are some examples:

```
simple = 'trezor'

prefix_only = 'trezor://'

uri_with_device_id = "trezor://1209:53c1:01"

uri_with_device_id_and_path = "trezor://1209:53c1:01/m/44'/60'/0'/0/0"

uri_with_path = "trezor:///m/44'/60'/0'/0/0"

uri_with_checksum_address = "trezor://0xdeadbeef"

```

KPrasch:
obsoleted by #2989

PR #2653, opened 2021-04-18
Initial implementation of native support for Ledger Nano hardware signer

svenski123:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Adds a new hardware signer `LedgerSigner` supporting Ledger Nano devices.

**Why it's needed:**
Using Ledger wallets with clef is a complete PITA - this make it much simpler.

**Notes for reviewers:**
- This uses [ledgereth](https://github.com/mikeshultz/ledger-eth-lib) which is an MIT licensed wrapper around [ledgerblue](https://github.com/LedgerHQ/blue-loader-python)
- All development/testing done on Linux/x86_64 using Ledger Nano X with latest firmware and ethereum app
- Tested with CLI tools in production/mainnet, querying commands but transactions only to the point of rejecting on the ledger
- Not clear to me if multiple ledgers attached to the computer will work (as the device handle is buried in the APIs)
- Signed transactions are older style (i.e. v=27/28) and not EIP-155 style (which the existing Trezor signer makes a big deal of requiring); likely can be implemented but ledgereth may require root canal surgery to do so
- Testnet accounts (i.e. with derivation paths beginning 44'/1'/...) required changing implementation global variables in ledgereth; this should ideally either be fixed upstream (it is an alpha project) or ledgereth should be forked / replaced
- Observations from testing with `nucypher stake restake`:
  - Two ledger signers get created as the account list is iterated twice which doesn't happen with `nucypher stake list`
  - If `--hw-wallet` is not specified, the CLI asks for a wallet unlock passwork which seems silly as the implementation of Signer.is_device() always returns true (i.e. the signer module knows if it's a hardware wallet or not, so another CLI option to tell it what it is is unnecessary, unless I'm missing some strange use case here).
 - Error/exception handling - not sure what the appropriate exceptions signers should through to allow for good UX without hiding details; as usual various stuff comes out of ledger blue, ledgereth does some exception mapping itself and so at the signer level it's a bit hodge podge - needs polish

### DIFF
--- a/nucypher/blockchain/eth/signers/hardware.py
+++ b/nucypher/blockchain/eth/signers/hardware.py
@@ -20,15 +20,17 @@ import rlp
 from eth_account._utils.transactions import assert_valid_fields, Transaction
 from eth_utils.address import to_canonical_address
 from eth_utils.applicators import apply_key_map, apply_formatters_to_dict
-from eth_utils.conversions import to_int
+from eth_utils.conversions import decode_hex, to_int
 from functools import wraps
 from hexbytes import HexBytes
+from ledgereth import Transaction as LedgerTransaction, get_account_by_path as ledger_get_account_by_path, sign_transaction as ledger_sign_transaction
 from toolz.dicttoolz import dissoc
 from trezorlib import ethereum
 from trezorlib.client import get_default_client, TrezorClient
 from trezorlib.tools import parse_path, Address, H_
 from trezorlib.transport import TransportException
 from typing import List, Tuple, Union
+from urllib.parse import urlparse, parse_qsl
 from web3 import Web3
 
 from nucypher.blockchain.eth.decorators import validate_checksum_address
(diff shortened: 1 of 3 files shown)


KPrasch:
@svenski123 - Thank you very much for your contribution, we'll get to reviewing it as soon as we can 😄

nucypher/blockchain/eth/signers/hardware.py
```diff
@@ -277,3 +279,194 @@ def sign_transaction(self,
...
+        paths = (path,) if path else self.__get_paths(testnet, count, legacy)
+
+        for item in paths:
+            acct = ledger_get_account_by_path(item)
+            self.__addresses[acct.address] = acct
+            if testnet and acct.path.startswith(self.__BIP44_ETH):
+                raise self.SignerError(f'cannot use production path/account in testnet: {acct.path}/{acct.address}')
+            if not testnet and acct.path.startswith(self.__BIP44_TEST):
+                raise self.SignerError(f'cannot use testnet path/account in production: {acct.path}/{acct.address}')
+            emitter.message(f"Derived {acct.address} ({acct.path})")
```
fjarri, line 334: Do we need to print it here? Can it go to the log?

KPrasch: > Do we need to print it here? Can it go to the log?

Personally, I like knowing when/what HW interactions are happening when using the CLI.  This is the same style of message we print for the trezor as well.

svenski123: FYI I modeled this after the trezor implementation hence the same key derivation messages.
Also some logic would likely be common to all walllet implementation (such as BIP32 key derivation logic, etc.) and could be factored out but I wanted this PR to be independent / isolated from other wallets to make code review easier, etc.

I'll be reworking this to implement EIP-155 transactions instead of the older style so I may need to ditch ledgereth (or include a modified version of it.)

KPrasch: Thanks for following up on this issue - I has some kind of suspicion that this might be the case but have not yet had the opportunity to review it in-depth.

KPrasch:
Hey @svenski123 Thanks for opening this PR, it was helpful for ledger users when it was opened!  Ultimately we've decided to close this PR and #2269 due to incoming changes that will deprecate the need for native wallet support and CLI stake creation in favor of web-based UI on the Threshold Network.

PR #2585, opened 2021-02-26
Extend policy probationary period for 2 months to 2021-04-30T23:59:59 UTC

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

**What this does:**
> High-level idea of the changes introduced in this PR. 
> List relevant API changes (if any), as well as related PRs and issues.

Extend probationary period for policy creation to 2021-04-30 23:59:59 UTC.

**Issues fixed/closed:**
> - Fixes #...

Fixes #2584 .
Related to #2353 , #2460 

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

Original probationary period ended on 2021-02-28 23:59:59 UTC.

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #2656, opened 2021-04-19
Extend policy probationary period until May 31st, 2021.

KPrasch:
**Type of PR:**
Other

**Required reviews:** 
1

**What this does:**
Extends policy probationary period until May 31st, 2021.

**Why it's needed:**
Prevent CLI CI failures and give additional development runway before permitting long-lived and high-value policies.


derekpierre (approved):
🎸 - Do we want to do a newsfagment (done for #2585 ) or since this is only temporary until we have a discussion, no need for a newsfragment...? It does affect app developers.

I guess after an in-depth discussion it is more likely to extend the probationary period than reduce 😅 , so a newsfragment this time would not be a bad thing.

KPrasch:
Yes, let's do a newsfragment, it's exactly the kind of thing it's good at!  Standby.

derekpierre:
One nitpick - May has 31 days 😅 , we should do the 31st. After that merge at will.

nucypher/config/constants.py
```diff
@@ -73,4 +73,4 @@
 NUCYPHER_EVENTS_THROTTLE_MAX_BLOCKS = 'NUCYPHER_EVENTS_THROTTLE_MAX_BLOCKS'
 
 # Probationary period (see #2353, #2584)
-END_OF_POLICIES_PROBATIONARY_PERIOD = MayaDT.from_iso8601('2021-04-30T23:59:59.0Z')
+END_OF_POLICIES_PROBATIONARY_PERIOD = MayaDT.from_iso8601('2021-05-30T23:59:59.0Z')
```
derekpierre, line 76:
```suggestion
END_OF_POLICIES_PROBATIONARY_PERIOD = MayaDT.from_iso8601('2021-05-31T23:59:59.0Z')
```

newsfragments/2656.misc.rst
```diff
@@ -0,0 +1 @@
+Extends policy probationary period until May 30th, 2021.  No policies may be created on the network beyond this date.
```
derekpierre, line 1:
```suggestion
Extends policy probationary period until May 31st, 2021.  No policies may be created on the network beyond this date.
```

PR #2657, opened 2021-04-21
Fallback chain of teacher nodes

### LINKED ISSUE #2481: Fallback chain of teacher nodes
Current code has only a single teacher node for each domain:
```python
    TEACHER_NODES = {
        'ibex': ('https://ibex.nucypher.network:9151',),
        'mainnet': ('https://mainnet.nucypher.network:9151',),
    }
```
If this node is down, learning crashes. We should have a more resilient approach with a chain of alternatives. Additionally, perhaps we should use `closest-seed.nucypher.network` as the default.

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
- Adds `https://closest-seed.nucypher.network` and `https://mainnet.nucypher.network:9151` nodes to `TEACHER_NODES`
- Moves `TEACHER_NODES` from `RestMiddleware` to `nodes.py`
- Adds some tests to confirm the correctness of using `TEACHER_NODES` as a fallback when local node storage is empty. 

**Issues fixed/closed:**
- Fixes #2481 

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -16,49 +16,53 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
 import contextlib
-import datetime
 import time
 from collections import defaultdict, deque
 from contextlib import suppress
 from queue import Queue
-from typing import Iterable, List
-from typing import Set, Tuple, Union
+from typing import Iterable, List, Set, Tuple, Union
 
 import maya
 import requests
+from bytestring_splitter import BytestringSplitter, BytestringSplittingError, PartiallyKwargifiedBytes, \
+    VariableLengthBytestring
+from constant_sorrow import constant_or_bytes
+from constant_sorrow.constants import (CERTIFICATE_NOT_SAVED, FLEET_STATES_MATCH, NOT_SIGNED, NO_KNOWN_NODES,
+                                       NO_STORAGE_AVAILIBLE, RELAX, UNKNOWN_VERSION)
 from cryptography.x509 import Certificate
 from eth_utils import to_checksum_address
 from requests.exceptions import SSLError
 from twisted.internet import reactor, task
 from twisted.internet.defer import Deferred
+from umbral.signing import Signature
 
-import nucypher
-from bytestring_splitter import BytestringSplitter, BytestringSplittingError, PartiallyKwargifiedBytes, \
-    VariableLengthBytestring
-from constant_sorrow import constant_or_bytes
-from constant_sorrow.constants import (CERTIFICATE_NOT_SAVED, FLEET_STATES_MATCH, NOT_SIGNED,
-                                       NO_KNOWN_NODES, NO_STORAGE_AVAILIBLE, UNKNOWN_FLEET_STATE, UNKNOWN_VERSION,
-                                       RELAX)
 from nucypher.acumen.nicknames import Nickname
 from nucypher.acumen.perception import FleetSensor
 from nucypher.blockchain.economics import EconomicsFactory
 from nucypher.blockchain.eth.agents import ContractAgency, StakingEscrowAgent
 from nucypher.blockchain.eth.constants import NULL_ADDRESS
+from nucypher.blockchain.eth.networks import NetworksInventory
 from nucypher.blockchain.eth.registry import BaseContractRegistry
 from nucypher.config.constants import SeednodeMetadata
 from nucypher.config.storages import ForgetfulNodeStorage
-from nucypher.crypto.api import recover_address_eip_191, verify_eip_191, InvalidNodeCertificate
+from nucypher.crypto.api import InvalidNodeCertificate, recover_address_eip_191, verify_eip_191
 from nucypher.crypto.kits import UmbralMessageKit
-from nucypher.crypto.powers import DecryptingPower, NoSigningPower, SigningPower, TransactingPower
+from nucypher.crypto.powers import DecryptingPower, NoSigningPower, SigningPower
 from nucypher.crypto.signing import signature_splitter
 from nucypher.network import LEARNING_LOOP_VERSION
 from nucypher.network.exceptions import NodeSeemsToBeDown
 from nucypher.network.middleware import RestMiddleware
 from nucypher.network.protocols import SuspiciousActivity
-from nucypher.network.server import TLSHostingPower
 from nucypher.utilities.logging import Logger
-from umbral.signing import Signature
 
+TEACHER_NODES = {
+    NetworksInventory.MAINNET: (
+        'https://seeds.nucypher.network:9151',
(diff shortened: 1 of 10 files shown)


nucypher/network/nodes.py
```diff
@@ -16,49 +16,53 @@
...
-from umbral.signing import Signature
 
+TEACHER_NODES = {
+    NetworksInventory.MAINNET: (
+        'https://seeds.nucypher.network:9151',
+        'https://closest-seed.nucypher.network:9151',
+    ),
+    NetworksInventory.LYNX: ('https://lynx.nucypher.network:9151',),
+    NetworksInventory.IBEX: ('https://ibex.nucypher.network:9151',),
+}
```
KPrasch, line 65: 👍🏻 Good relocation.

nucypher/network/nodes.py
```diff
@@ -16,49 +16,53 @@
...
-from typing import Set, Tuple, Union
+from typing import Iterable, List, Set, Tuple, Union
 
 import maya
 import requests
+from bytestring_splitter import BytestringSplitter, BytestringSplittingError, PartiallyKwargifiedBytes, \
+    VariableLengthBytestring
+from constant_sorrow import constant_or_bytes
+from constant_sorrow.constants import (CERTIFICATE_NOT_SAVED, FLEET_STATES_MATCH, NOT_SIGNED, NO_KNOWN_NODES,
+                                       NO_STORAGE_AVAILIBLE, RELAX, UNKNOWN_VERSION)
```
KPrasch, line 31: Small style nitpick (will be good to perform import optimization, sorting, and formatting as part of pre-commit hooks):

```suggestion
from bytestring_splitter import (
    BytestringSplitter,
    BytestringSplittingError,
    PartiallyKwargifiedBytes,
    VariableLengthBytestring
)
from constant_sorrow import constant_or_bytes
from constant_sorrow.constants import (
    CERTIFICATE_NOT_SAVED, 
    FLEET_STATES_MATCH,
    NOT_SIGNED,
    NO_KNOWN_NODES,
    NO_STORAGE_AVAILIBLE,
    RELAX,
    UNKNOWN_VERSION
)
```

nucypher/utilities/networking.py
```diff
@@ -114,14 +114,15 @@ def get_external_ip_from_default_teacher(network: str,
...
+    # Prevents circular imports
     from nucypher.characters.lawful import Ursula
+    from nucypher.network.nodes import TEACHER_NODES
 
     if federated_only and registry:
         raise ValueError('Federated mode must not be true if registry is provided.')
     base_error = 'Cannot determine IP using default teacher'
     try:
-        top_teacher_url = RestMiddleware.TEACHER_NODES[network][0]
+        top_teacher_url = TEACHER_NODES[network][0]
```
derekpierre, line 125: In this function should we also be trying the next entry if the first one fails i.e.  if `NodeSeemToBeDown` L144. i.e. use a for-loop for all teachers available in TEACHER_NODES[network] and as soon as one is able to be connected to, L141, we short circuit the for-loop. I know it doesn't do that now, but we didn't have more than one entry before.

Or am I totally missing the mark?

^ @KPrasch / @cygnusv ?

KPrasch: You are totally correct!  I missed the `[0]`

nucypher/network/nodes.py
cygnusv: NO_STORAGE_AVAILIBLE -> NO_STORAGE_AVAILABLE

nucypher/utilities/networking.py
```diff
@@ -136,12 +134,21 @@ def get_external_ip_from_default_teacher(network: str,
...
-    except NodeSeemsToBeDown:
-        # Teacher is unreachable.  Move on.
+    teacher = None
+    for teacher_uri in TEACHER_NODES[network]:
+        try:
+            teacher = Ursula.from_teacher_uri(teacher_uri=teacher_uri,
+                                              federated_only=federated_only,
+                                              min_stake=0)  # TODO: Handle customized min stake here.
+            # Found a reachable teacher, return from loop
+            break
```
derekpierre, line 144: Do we need to also incorporate L155 as a check in this loop? That being said if `from_teacher_uri` works (it makes a connection to the node) chances are `_request_from_node` should work anyway (it also makes a connection to the node) - so it may be superfluous. Interested in what others think (@KPrasch )?

piotr-roslaniec: I think it's a good idea because `_request_from_node` is handling parsing and related errors.

nucypher/network/nodes.py
```diff
@@ -16,49 +16,63 @@
...
+    PartiallyKwargifiedBytes,
+    VariableLengthBytestring
+)
+from constant_sorrow import constant_or_bytes
+from constant_sorrow.constants import (
+    CERTIFICATE_NOT_SAVED,
+    FLEET_STATES_MATCH,
+    NOT_SIGNED,
+    NO_KNOWN_NODES,
+    NO_STORAGE_AVAILIBLE,
```
derekpierre, line 38: Seems like this was misspelt from before - https://github.com/nucypher/nucypher/blob/main/nucypher/network/nodes.py#L40, https://github.com/nucypher/nucypher/blob/main/nucypher/network/nodes.py#L244

nucypher/network/nodes.py
```diff
@@ -16,49 +16,63 @@
...
 from nucypher.network.middleware import RestMiddleware
 from nucypher.network.protocols import SuspiciousActivity
-from nucypher.network.server import TLSHostingPower
 from nucypher.utilities.logging import Logger
-from umbral.signing import Signature
 
+TEACHER_NODES = {
+    NetworksInventory.MAINNET: (
+        'https://seeds.nucypher.network:9151',
+        'https://closest-seed.nucypher.network:9151',
```
cygnusv, line 71: @vepkenez What do you think of this? I remember suggesting something similar in the past and you had some comments but I forgot where was that discussion. Related to this, wouldn't be better to put closest-seed first?

vepkenez: `https://seeds.nucypher.network:9151` should be `https://seeds.nucypher.network`

may as well add
`https://mainnet.nucypher.network`
at the end.

And yes.  It wouldn't hurt to put `https://closest-seed.nucypher.network:9151` first.
It will generally be the same result as `seeds` but there seem to be some occasional timeouts or some other unknown issues with that one that i'm still not sure about.

derekpierre (approved):
🎸

fjarri (approved):

PR #2664, opened 2021-04-26
[EPIC] Porter MVP - "Infura for NuCypher"

derekpierre:
porter_review_Logo2017_860x250-1-1

*("Porter" is also a Canadian airline based out of Toronto - https://blog.flyporter.com/)*

**CHAMPION:** @derekpierre 

Goal
-----
Simplifies and abstracts the complexities surrounding the nucypher protocol. It aims to negate the need for running the python client by applications to interact with the network - think web and mobile applications. It would leverage `rust-umbral` and associated JS bindings to achieve cross-platform functionality via the browser, without needing to rewrite the entire python client.

`web app <--> "js-nucypher" <--> porter <--> network`

Related to https://github.com/nucypher/productdev/issues/7.

MVP Milestone: https://github.com/nucypher/nucypher/milestone/18

High-level areas of focus for MVP:
* Scope the functionality that needs to be provided eg. learning loop, sampling, kfrag distribution, payment,...?  ( #2666 , #2667 )
* Porter Basic Learner (#2668)
* REST App/CLI Plumbing (#2702)
* REST API Implementation (based on work done in #2666 and #2667  )
    * Alice functionality (#2703 )
    * Bob functionality (#2704)
* Containerization and Deployment/Running of Porter, eg. with docker (#2669 ) 
* Documentation (#2670 )

* Association with `js-nucypher` - connective tissue between `rust-umbral`/webapps and `porter` (Issue(s) TBD)
    * API and functionality (probably becomes clearer as porter work moves along)
     
PR Ramp Order
----------------

- #2713 
- #2717 
- #2720 
- #2724 
- #2726 
- #2721 
- #2732 
- #2733 
- #2746 
- #2729


docs/source/application_development/porter.rst
```diff
@@ -0,0 +1,633 @@
...
+                "uri": "https://3.36.66.164:9151"
+             },
+             {
+                "encrypting_key": "02e43a623c24db4f62565f82b6081044c1968277edfdca494a81c8fd0826e0adf6",
+                "checksum_address": "0xfBeb3368735B3F0A65d1F1E02bf1d188bb5F5BE6",
+                "uri": "https://128.199.124.254:9151"
+             }
+          ]
+       },
+       "version": "6.0.0"
```
fjarri, line 465: What is this the version of? `nucypher`?

docs/source/application_development/porter.rst
```diff
@@ -0,0 +1,633 @@
...
+
+Parameters
+++++++++++
++----------------------------------+---------------+----------------------------------------+
+| **Parameter**                    | **Type**      | **Description**                        |
++==================================+===============+========================================+
+| ``ursula``                       | String        | | Checksum address corresponding to    |
+|                                  |               | | the Ursula to execute the work order.|
++----------------------------------+---------------+----------------------------------------+
+| ``work_order_payload``           | String        | Work order payload encoded as base64.  |
```
fjarri, line 591: Where would one get it from? I don't think it's exposed to the user.

nucypher/blockchain/eth/agents.py
```diff
@@ -252,7 +257,10 @@ class StakingEscrowAgent(EthereumContractAgent):
         'finishUpgrade'
     )
 
-    DEFAULT_PAGINATION_SIZE: int = 30    # TODO: Use dynamic pagination size (see #1424)
+    DEFAULT_STAKER_PAGINATION_SIZE_LIGHT_NODE: int = int(os.environ.get(
```
fjarri, line 260: This pattern worries me. First, the pagination size feels like a more high-level setting; whoever calls `StakingEscrowAgent` should worry about setting it. Second, the low-level class's behavior is influenced by an envvar, which may lead to some hard to diagnose problems. The only other place where we do that is `ContractEventsThrottler`.

nucypher/blockchain/eth/agents.py
```diff
@@ -321,15 +330,36 @@ def get_all_active_stakers(self, periods: int, pagination_size: Optional[int] =
             start_index: int = 0
             n_tokens: int = 0
             stakers: Dict[int, int] = dict()
-            active_stakers: Tuple[NuNits, List[List[int]]]
+            attempts = 0
```
fjarri, line 333: Should this be set inside the `while` loop? Or do we only get three pagination decreases in total?

nucypher/blockchain/eth/agents.py
```diff
@@ -321,15 +330,36 @@ def get_all_active_stakers(self, periods: int, pagination_size: Optional[int] =
...
+                        # exception unrelated to pagination size and timeout
+                        raise e
+                    elif pagination_size == 1 or attempts >= 3:
+                        # we tried
+                        raise e
+                    else:
+                        # reduce pagination size and retry
+                        old_pagination_size = pagination_size
+                        pagination_size = old_pagination_size // 2
+                        if pagination_size == 0:
```
fjarri, line 351: Wouldn't that possibility be already eliminated by `if pagination_size == 1` above?

nucypher/characters/control/specifications/fields/datetime.py
```diff
@@ -17,8 +17,10 @@
 
 import maya
 from marshmallow import fields
+from pendulum.parsing import ParserError
```
fjarri, line 20: May be better to just use `maya.pendulum.parsing.ParserError`. When one package raises an exception from another package, it looks weird (`maya` could have wrapped it, but since they decided not to...).

nucypher/utilities/porter/porter.py
```diff
@@ -0,0 +1,274 @@
...
+    _LONG_LEARNING_DELAY = 30
+    _ROUNDS_WITHOUT_NODES_AFTER_WHICH_TO_SLOW_DOWN = 25
+
+    DEFAULT_EXECUTION_TIMEOUT = 10  # 10s
+
+    DEFAULT_PORT = 9155
+
+    _interface_class = PorterInterface
+
+    class UrsulaInfo:
```
fjarri, line 72: Could be a named tuple

tests/integration/porter/conftest.py
```diff
@@ -0,0 +1,73 @@
...
+
+from nucypher.crypto.umbral_adapter import PublicKey
+
+
+@pytest.fixture(scope='session')
+def random_federated_treasure_map_data():
+    random_bob_encrypting_key = PublicKey.from_bytes(
+        bytes.fromhex("026d1f4ce5b2474e0dae499d6737a8d987ed3c9ab1a55e00f57ad2d8e81fe9e9ac"))
+    random_treasure_map_id = "cc1c29dd2305483cc838cbcc5ecb5ef6edfd69ecadb0aa52b6b084b630989187"  # federated is 32 bytes
+    random_treasure_map = b64decode("VE0AAZ3jsXPynMD9dm+Fjwi49bxkOzUjwsNI0Y0p8bGB9F60OXGmJqibK0Ki4FSWti2Y"
```
fjarri, line 29: Would be helpful to have some note about how this treasure map was generated, for the event when its format changes. Or, if it's possible, maybe even generate it on the fly? It's a session-scoped fixture, after all.

docs/source/application_development/porter.rst
derekpierre: Yes, it is the version of nucypher. I'm just proactively setting the version to 6.0.0 in the docs so that I don't have to update it later.

docs/source/application_development/porter.rst
derekpierre: Bob has to construct one. Is this not possible currently?

nucypher/blockchain/eth/agents.py
derekpierre: The caller of `StakingEscrowAgent` can still proactively specify a `pagination_size` see `get_all_active_stakers()` which is the only place a pagination size is needed/used. If the user doesn't specify a pagination_size (i.e. `None`), only then is the default used.

We were already using a default value for light nodes in prior code because they act differently than regular nodes. However, the number was fixed and could not be modified meaning that if we were wrong on the value it couldn't be modified in the field without changing the code. To me, this was the best compromise between allowing the user to specify a pagination size, if not use a default, if the default failed then it could be modified proactively.

The goal here is not to have the env var ever set, but rather have it act as a fallback just in case, for some reason, our beliefs about what the pagination size should be are incorrect.

Regarding "hard to diagnose" problems, for extra help, we log when the default pagination size is used and its value. see L324.

nucypher/blockchain/eth/agents.py
derekpierre: Three pagination decreases in total. Because of the start_index, if a larger size worked once but not a subsequent time, we just continue from where we were with a decreased size e.g. if 250 pagination size worked the first time (i.e. we got the first 250 stakers), and then failed the second time for a timeout reason, we would try again with a pagination size of 125 starting from start_index 250.

nucypher/blockchain/eth/agents.py
derekpierre: ✔️  Yep. Was overly cautious here and this check can be removed.

nucypher/characters/control/specifications/fields/datetime.py
derekpierre: ✔️

nucypher/utilities/porter/porter.py
derekpierre: ✔️

tests/integration/porter/conftest.py
derekpierre: Added a note about the values being obtained from running the heartbeat demo locally. It is specific to porter tests, and generating it on the fly seems excessive.

docs/source/application_development/porter.rst
fjarri: It is possible, but it is sort of internal, and may change significantly after #2730. Currently one can obtain some work orders via `work_orders_for_capsules()` (which is public, but not documented), but there are no public methods that take them. What's the scenario here? Why would the user want work orders?

derekpierre: The underlying scenario is what does Bob need to provide to Porter to get data re-encrypted by an Ursula, without disclosing anything sensitive...I assumed that was simply a WorkOrder payload that gets passed to the Ursula `/reencrypt` endpoint  See https://github.com/nucypher/nucypher/issues/2667. 

Do we need to make something public for work orders, or reconsider the strategy?

nucypher/blockchain/eth/agents.py
fjarri: I don't mind the default, or that there's an envvar for the default, it just seems that this specific spot is too deep in the call stack to depend on external environment. I admit I can't think of a good solution at the moment.

> Regarding "hard to diagnose" problems, for extra help, we log when the default pagination size is used and its value.

It helps, but it's still a surprising side effect.

docs/source/application_development/porter.rst
fjarri: Why not just make the whole `retrieve()` a Porter method? Why would Bob need a fine-grained control over work orders?

fjarri: Ah ok, `retrieve()` returns cleartexts. Then actually the `retrieve_cfrags()` method from #2730 would work quite well.

derekpierre: Looks like `work_order_payload` would be the equivalent of the bytes representation of the `RemoteWorkOrder` object

fjarri: Does it need to? I don't see why we should expose work orders at all. `retrieve_cfrags()` seems to cover all the Bob's needs without exposing the actual cleartexts. The only thing that may be missing here is the ability to select Ursulas to contact, but I'm not sure if that's necessary.

derekpierre: Unsure what you mean. Do you mean Porter should take the same parameters as `retrieve_cfrags()` and then perform the same functionality?

> The only thing that may be missing here is the ability to select Ursulas to contact, but I'm not sure if that's necessary.
Could you clarify this as well?

fjarri: Sorry, I just wanted to understand the requirements here to make sure I address them in 2730. For now this indeed seems to be the only way to retrieve via Porter without disclosing cleartexts to it. 

> The only thing that may be missing here is the ability to select Ursulas to contact, but I'm not sure if that's necessary.

So, currently, Bob can pick Ursulas he wants to contact for reencryption (since he provides work orders to Porter). If 2730 API (in its current form) is merged, Porter will only expose `retrieve_cfrags`, where Ursulas will be chosen by Porter, and Bob won't be able to affect it. I wonder if it is important to give Bob the ability to choose the Ursulas too.

derekpierre: > I just wanted to understand the requirements here to make sure I address them in 2730

👍 . The philosophical requirement is that Porter shouldn't gain access to any sensitive material when performing a re-encryption on behalf of Bob.

> currently, Bob can pick Ursulas

A somewhat limited "pick" since the set of ursulas is from the treasure map, and each work order is associated with a single ursula - although I guess you mean to achieve the threshold - in that case yes. Based on the current code, I was effectively trying to avoid dealing with Capsules in Porter.

Seems like we can follow-up on what this endpoint may look like/change to in #2730 , if the retrieval process is going to be refactored/updated. If that gets cleared up, then modifying the endpoint is relatively easy by just updating the endpoint schema and refactoring the common retrieval code to be used both by Porter and in the pure Python context.

piotr-roslaniec (approved):
I reviewed some of the PRs I previously missed out on. LGTM :raccoon:

cygnusv (approved):
Checked individual PRs, and ran porter via CLI and docker. This is an awesome work! 💪

vepkenez (approved):

derekpierre:
> @derekpierre - What do you make of the CI failure? https://app.circleci.com/pipelines/github/nucypher/nucypher/8212/workflows/73bacae3-e320-4803-a63e-14ab170a8210/jobs/139336

@KPrasch, it was introduced by the latest rebase which excluded a needed code change - should be fixed now. Thanks for checking!

KPrasch (approved):
Stellar effort here @derekpierre 👏🏻

PR #2673, opened 2021-04-28
Check allowance value before transactions

### LINKED ISSUE #2636: Check allowance value before executing transactions that depend on its value - approve() and approveAndCall()
**Describe the Bug**
Increasing a stake size involves two transactions 1) `increaseallowance` and 2) `depositAndIncrease`.  The `increaseallowance` transaction was successful, but the `depositAndIncrease` transaction failed.  Therefore, after the failed attempt to increase my stake size, the allowance value stored in the Token Contract was > 0.

I tried to increase my stake again thinking that perhaps it was a transient error. Again the same thing happened (`increaseallowance` was successful, but `depositAndIncrease` failed). So now my allowance in the Token Contract was 2x. 

(failure related to #2635)

In this case, perhaps trying to increase a stake should check the value of allowance and act accordingly; prompt the user, or provide some message or get the user to adjust the amount or skip the `increaseallowance` entirely and just do `depositAndIncrease` or ...?

--

After my failed attempts to increase my stake, and with my allowance at 2x, I tried to create a new stake via the CLI, and that failed with `Reason: execution reverted` because the allowance value was not 0 - we should provide better error handling by checking the allowance value and providing a more contextual message.

This seems to apply to any functionality that uses `approve()` and `approveAndCall()` calls.

 - OS Version: MacOS 11.2.3
 - Nucypher Version: v4.8.2
 - Ethereum Node Version:  Infura (Geth/v1.10.1-omnibus-2230b5e6/linux-amd64/go1.16) [version checked days later not on day of actual issue)

piotr-roslaniec:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Updates token allowance when needed for `depositAndIncrease()` transaction
- Raises error if token allowance is not 0 prior to `approve()` and `approveAndCall()` tranasctions

**Issues fixed/closed:**
- Fixes #2636 

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -787,9 +787,14 @@ class Staker(NucypherTokenActor):
 
     def _deposit_and_increase(self, stake_index: int, amount: int) -> TxReceipt:
         """Public facing method for deposit and increasing stake."""
-        self.token_agent.increase_allowance(increase=amount,
-                                            transacting_power=self.transacting_power,
-                                            spender_address=self.staking_agent.contract.address)
+        current_allowance = self.token_agent.get_allowance(owner=self.checksum_address,
+                                                           spender=self.staking_agent.contract.address)
+        if amount > current_allowance:
+            to_increase = amount - current_allowance
+            self.token_agent.increase_allowance(increase=to_increase,
+                                                transacting_power=self.transacting_power,
+                                                spender_address=self.staking_agent.contract.address)
+            self.log.info(f"{self.checksum_address} increased token allowance for spender {self.staking_agent.contract.address} to {amount}")
         receipt = self.staking_agent.deposit_and_increase(transacting_power=self.transacting_power,
                                                           stake_index=stake_index,
                                                           amount=amount)
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -764,9 +764,7 @@ class Staker(NucypherTokenActor):
 
     def _deposit(self, amount: int, lock_periods: int) -> TxReceipt:
         """Public facing method for token locking."""
-        self.token_agent.approve_transfer(amount=0,
-                                          spender_address=self.staking_agent.contract_address,
-                                          transacting_power=self.transacting_power)
+        self._ensure_allowance_equals(0)
         receipt = self.token_agent.approve_and_call(amount=amount,
                                                     target_address=self.staking_agent.contract_address,
                                                     transacting_power=self.transacting_power,
@@ -790,22 +788,28 @@ class Staker(NucypherTokenActor):
 
     def _deposit_and_increase(self, stake_index: int, amount: int) -> TxReceipt:
         """Public facing method for deposit and increasing stake."""
-        self._ensure_allowance_at_least(amount)
+        self._ensure_allowance_equals(amount)
         receipt = self.staking_agent.deposit_and_increase(transacting_power=self.transacting_power,
                                                           stake_index=stake_index,
                                                           amount=amount)
         return receipt
 
-    def _ensure_allowance_at_least(self, amount):
-        current_allowance = self.token_agent.get_allowance(owner=self.checksum_address,
-                                                           spender=self.staking_agent.contract.address)
+    def _ensure_allowance_equals(self, amount: int):
+        owner = self.checksum_address
+        spender = self.staking_agent.contract.address
+        current_allowance = self.token_agent.get_allowance(owner=owner, spender=spender)
         if amount > current_allowance:
             to_increase = amount - current_allowance
             self.token_agent.increase_allowance(increase=to_increase,
                                                 transacting_power=self.transacting_power,
-                                                spender_address=self.staking_agent.contract.address)
-            self.log.info(
-                f"{self.checksum_address} increased token allowance for spender {self.staking_agent.contract.address} to {amount}")
+                                                spender_address=spender)
+            self.log.info(f"{owner} increased token allowance for spender {spender} to {amount}")
+        elif amount < current_allowance:
+            to_decrease = current_allowance - amount
+            self.token_agent.decrease_allowance(decrease=to_decrease,
+                                                transacting_power=self.transacting_power,
+                                                spender_address=spender)
+            self.log.info(f"{owner} decreased token allowance for spender {spender} to {amount}")
 
     def _lock_and_increase(self, stake_index: int, amount: int) -> TxReceipt:
         """Public facing method for increasing stake."""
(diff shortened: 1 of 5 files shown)


cygnusv (commented):
I think we could make this reusable, as this pattern occurs in other places. There's potential to encapsulate all approve calls with a method that uses increaseAllowance underneath, with the extreme case of even skipping this step if there's enough allowance

vzotova (approved):

KPrasch:
> I think we could make this reusable, as this pattern occurs in other places.

Yes, I agree - as demonstrated in a discord discussion this can also happen while trying to create a new stake with an account that has previously created stakes.

<https://discord.com/channels/411401661714792449/497481850382450710/837152228811866162>

piotr-roslaniec:
`approve` method requires prior allowance [to be 0](https://github.com/piotr-roslaniec/nucypher/blob/6d225e422ff17edf0e608a9b210a74d1ae66f98c/nucypher/blockchain/eth/sol/source/zeppelin/token/ERC20/ERC20.sol#L82-L82), and this also applies to the `approveAndCall` method. We could rewrite `approve` and `approveAndCall` usages to rely on `increaseAllowance` instead, but that [in some cases](https://github.com/piotr-roslaniec/nucypher/blob/6d225e422ff17edf0e608a9b210a74d1ae66f98c/nucypher/blockchain/eth/actors.py#L767-L772) would require making additional transactions. WDYT?

vzotova:
> `approve` method requires prior allowance [to be 0](https://github.com/piotr-roslaniec/nucypher/blob/6d225e422ff17edf0e608a9b210a74d1ae66f98c/nucypher/blockchain/eth/sol/source/zeppelin/token/ERC20/ERC20.sol#L82-L82), and this also applies to the `approveAndCall` method. We could rewrite `approve` and `approveAndCall` usages to rely on `increaseAllowance` instead, but that [in some cases](https://github.com/piotr-roslaniec/nucypher/blob/6d225e422ff17edf0e608a9b210a74d1ae66f98c/nucypher/blockchain/eth/actors.py#L767-L772) would require making additional transactions. WDYT?

I'd prefer have one tx for `approve_and_call` because this is cheaper and simpler. For cases when more than one tx - `increaseAllowance` will be better

nucypher/blockchain/eth/actors.py
```diff
@@ -764,6 +764,9 @@ def _prolong_stake(self, stake_index: int, lock_periods: int) -> TxReceipt:
 
     def _deposit(self, amount: int, lock_periods: int) -> TxReceipt:
         """Public facing method for token locking."""
+        self.token_agent.approve_transfer(amount=0,
+                                          spender_address=self.staking_agent.contract_address,
+                                          transacting_power=self.transacting_power)
```
piotr-roslaniec, line 769: Because `approve_and_call` requires allowance to be 0, we either have to keep this horrific workaround

vzotova: in `approve_and_call` on contract side we can rewrite only `call` part because `approve` part happens inside token contract (which is not upgradeable). So in normal case allowance is already zero and first tx should be skipped

nucypher/blockchain/eth/agents.py
```diff
@@ -185,6 +185,8 @@ def approve_transfer(self,
                          transacting_power: TransactingPower
                          ) -> TxReceipt:
         """Approve the spender address to transfer an amount of tokens on behalf of the sender address"""
+        self._validate_zero_allowance(spender_address, transacting_power)
```
vzotova, line 188: if amount is zero and allowance is not then `_validate_zero_allowance` will fail (but it should not)

nucypher/blockchain/eth/actors.py
piotr-roslaniec: What happens if we have multiple `approveAndCall` method calls? Assuming allowance is zero, the first call will work and thus set allowance to some amount larger than 0. Now, because the allowance is non-zero, the second call will fail. I think this is what happened in the issue described by @KPrasch 
>Yes, I agree - as demonstrated in a discord discussion this can also happen while trying to create a new stake with an account that has previously created stakes. 
I updated the code to handle this case.

nucypher/blockchain/eth/agents.py
```diff
@@ -178,13 +178,27 @@ def increase_allowance(self,
...
         return receipt
 
+    @contract_api(TRANSACTION)
+    def decrease_allowance(self,
+                           transacting_power: TransactingPower,
+                           spender_address: ChecksumAddress,
+                           decrease: NuNits
+                           ) -> TxReceipt:
+        """Decrease the allowance of a spender address funded by a sender address"""
+        contract_function: ContractFunction = self.contract.functions.increaseAllowance(spender_address, decrease)
```
derekpierre, line 188:
```suggestion
        contract_function: ContractFunction = self.contract.functions.decreaseAllowance(spender_address, decrease)
```

Curious how the test didn't catch this...?

derekpierre (changes requested):

nucypher/blockchain/eth/actors.py
vzotova: after executing `approveAndCall` allowance will be zero again, so it's not zero only when something fail

nucypher/blockchain/eth/agents.py
piotr-roslaniec: It wasn't tested, I addressed it in the new commit.

derekpierre (approved):
🎸

nucypher/blockchain/eth/actors.py
```diff
@@ -787,14 +788,29 @@ def _divide_stake(self, stake_index: int, additional_periods: int, target_value:
...
-                                            transacting_power=self.transacting_power,
-                                            spender_address=self.staking_agent.contract.address)
+        self._ensure_allowance_equals(amount)
         receipt = self.staking_agent.deposit_and_increase(transacting_power=self.transacting_power,
                                                           stake_index=stake_index,
                                                           amount=amount)
         return receipt
 
+    def _ensure_allowance_equals(self, amount: int):
+        owner = self.checksum_address
```
cygnusv, line 798: I lost track about the invariant around `self.checksum_address`. Is it always the case that this address is the same than the transacting one? Just in case, I would do something like:
```suggestion
        owner = self.transacting_power.account
```

cygnusv (commented):
Just one comment. Apart from that, this is a great idea :)  Thanks @piotr-roslaniec !

KPrasch (approved):

PR #2681, opened 2021-05-01
Cleaned up each DB on shutdown, and removed reactor.run to give user console control

### LINKED ISSUE #2232: Proper cleanup of demo ursula db files
Related to #1741 - see https://github.com/nucypher/nucypher/pull/1741/files#r480311013

```python  
  finally:
        with suppress(FileNotFoundError):
            os.remove("sage.db")
        for u in ursulas[1:]:
            with suppress(FileNotFoundError):
                os.remove(f"{u.rest_interface.port}.db")
```

The removal of `sage.db` and `{u.rest_interface.port}.db` need to be relative to the `APP_DIR, user_cache_dir` since they were created relative to it. 

i.e. `{Path(APP_DIR.user_cache_dir) / 'sage.db'}` and `{Path(APP_DIR.user_cache_dir) / u.rest_interface.port}.db`.

ghostmonk:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Removed reactor.run in the demo fleet script because it was stealing console input from the user
- cleaned up the temp db files

**Issues fixed/closed:**
- Fixes #2232

**Why it's needed:**
This is just a simple db clean up for the fleet demo in the example scripts
Removed `reactor.run` because it steals console input from the user

### DIFF
--- a/examples/run_demo_ursula_fleet.py
+++ b/examples/run_demo_ursula_fleet.py
@@ -20,6 +20,9 @@ import shutil
 from functools import partial
 from pathlib import Path
 
+from twisted.internet import reactor
+from contextlib import suppress
+
 from nucypher.characters.lawful import Ursula
 from nucypher.config.constants import APP_DIR, TEMPORARY_DOMAIN
 from nucypher.utilities.networking import LOOPBACK_ADDRESS
@@ -63,17 +66,11 @@ def spin_up_federated_ursulas(quantity: int = FLEET_POPULATION):
         print(f"{u}: {deployer._listening_message()}")
 
     try:
-        while True:
-            input("'Ctl + C' to quit\r\n")
+        reactor.run()
     finally:
-        last_u = None
         for u in ursulas:
-            shutil.rmtree(u.datastore.db_path)
-            last_u = u
-
-        # This kills the whole process.
-        # TODO: Is there a graceful way to shut each Ursula down?
-        last_u.get_deployer().stop()
+            with suppress(FileNotFoundError):
+                shutil.rmtree(u.datastore.db_path)
 
 
 if __name__ == "__main__":
--- a/examples/run_demo_ursula_fleet.py
+++ b/examples/run_demo_ursula_fleet.py
@@ -47,12 +47,11 @@ def spin_up_federated_ursulas(quantity: int = FLEET_POPULATION):
 
     ursulas.append(sage)
     for index, port in enumerate(ports[1:]):
-        db = f"{USER_CACHE / port}.db"
         u = ursula_maker(
             rest_port=port,
             seed_nodes=[sage.seed_node_metadata()],
             start_learning_now=True,
-            db_filepath=db,
+            db_filepath=f"{USER_CACHE / port}.db",
         )
         ursulas.append(u)
 
--- a/examples/run_demo_ursula_fleet.py
+++ b/examples/run_demo_ursula_fleet.py
@@ -16,11 +16,9 @@
 """
 
 import os
-from contextlib import suppress
+import shutil
 from functools import partial
 from pathlib import Path
-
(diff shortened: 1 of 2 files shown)


examples/run_demo_ursula_fleet.py
```diff
@@ -52,23 +51,29 @@ def spin_up_federated_ursulas(quantity: int = FLEET_POPULATION):
...
+
     for u in ursulas:
         deployer = u.get_deployer()
         deployer.addServices()
         deployer.catalogServers(deployer.hendrix)
         deployer.start()
         print(f"{u}: {deployer._listening_message()}")
+
     try:
-        reactor.run()  # GO!
```
KPrasch, line 65: I'm slightly confused here - without the reactor running, how do the Ursulas startup services?  Are you certain they were not still running from a previous test?

ghostmonk: Hmmm... ok that's a good question. Let me double check. 
If I use reactor.run... how can I shut the ursulas down from the command line?

ghostmonk: ok.. you are right... so, `reactor.run` is problematic, because the user loses control of the terminal. The only way to shut the program down is to kill the terminal, or to kill the python program. (In a separate terminal)

Even if this were an acceptable way to stop this script, killing the program skips the finally block.

Do you know how I can call `reactor.run` in a way where the user still has control of the terminal, and how I would shut the Ursulas down gracefully?

KPrasch: > ok.. you are right... so, reactor.run is problematic, because the user loses control of the terminal.

In your view, why is this problematic?

> Even if this were an acceptable way to stop this script, killing the program skips the finally block.

`KeyboardInterrupt` skips the finally block?


> Do you know how I can call reactor.run in a way where the user still has control of the terminal, and how I would shut the Ursulas down gracefully?

Yes, it's possible to use a twisted `LineReceiever` for this purpose, but I don't think it's worth the complexity just for a demo.

ghostmonk: I mean... I can't stop the demo from running in the shell once it starts.
So as far as I can tell.. there is no way to keyboard interrupt.
I would have to kill the terminal... and the finally block doesn't fire when I do

KPrasch: > So as far as I can tell.. there is no way to keyboard interrupt.

What happens when you CTRL+C ?

ghostmonk: nothing... I cannot stop this process.
Can you?

KPrasch: > Can you?

Yes, but it's not consistent.  CTRL+C early on in the runtime seems to stop normally.  However if I wait about ten or twenty seconds for services to fully startup, that does not work anymore, so I end up halting the process with a core dump (CTRL+\ on ubuntu).

This is not expected behavior, if this process does not respond to a `KeyboardInterrupt`, it's most likely an indication that there is an unhandled exception or missing twisted error handler

newsfragments/2681.bugfix.rst
```diff
@@ -0,0 +1 @@
+examples/run_demo_ursula_fleet.py - Clean up each DB on shutdown, and removed reactor.run to give user console control
```
cygnusv, line 1:
```suggestion
examples/run_demo_ursula_fleet.py - Clean up each DB on shutdown.
```

cygnusv (approved):

KPrasch (approved):
Thanks for your contribution @ghostmonk!

PR #2686, opened 2021-05-06
Update WorkTracker consistency check

### LINKED ISSUE #2429: Too lenient consistency check in WorkTracker
Two related problems in `WorkTracker.__tracking_consistency_check()`:

* Currently, if `txs_in_mempool < len(self.__pending)` it returns `None` (will return `True` after #2425 is merged) quietly. But this case still indicates some inconsistency, and even if we choose to ignore it, it at least needs to be logged. We need to investigate what events can cause this condition to be true.
* In `__track_pending_commitments()`, if an inconsistency was detected, `UNTRACKED_PENDING_TRANSACTION` is added to the list of pending transactions. This means, if before that the inconsistency was that `txs_in_mempool` was greater than `len(self.__pending)` by 1, now `WorkTracker` suddenly becomes consistent. This may cause problems somewhere up the hierarchy, depending on how it is used. It is better to store the inconsistency marker somewhere separately.

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
- Handles an edge case, where worker's commitment transaction is dropped from mempool
- Checks whether a pending transaction is actually a commitment transaction

**Issues fixed/closed:**
- Fixes #2429

**Why it's needed:**

**Notes for reviewers:**
- CLI test is broken on `main`

### DIFF
--- a/newsfragments/2686.misc.rst
+++ b/newsfragments/2686.misc.rst
@@ -1,1 +1,1 @@
-Ursula will now detect absence of active stakes during startup and in worker loop.
+Ursula will now detect commitment transactions dropped from mempool.
--- a/newsfragments/2686.misc.rst
+++ b/newsfragments/2686.misc.rst
@@ -1,0 +1,1 @@
+Ursula will now detect absence of active stakes during startup and in worker loop.
--- a/nucypher/blockchain/eth/token.py
+++ b/nucypher/blockchain/eth/token.py
@@ -558,7 +558,7 @@ class WorkTracker:
         self._tracking_task.clock = self.CLOCK
 
         self.__pending = dict()  # TODO: Prime with pending worker transactions
-        self.__num_dropped_txs = 0
+        self.__has_untracked_tx = False
         self.__requirement = None
         self.__current_period = None
         self.__start_time = NOT_STAKING
@@ -634,7 +634,6 @@ class WorkTracker:
             self._consecutive_fails += 1
             self.start(commit_now=commit_now)
 
-
     def __work_requirement_is_satisfied(self) -> bool:
         # TODO: Check for stake expiration and exit
         if self.__requirement is None:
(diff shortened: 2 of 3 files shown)


nucypher/blockchain/eth/token.py
```diff
@@ -649,29 +648,42 @@ def pending(self) -> Dict[int, HexBytes]:
...
+            self.log.info(f'{self.__num_untracked_commitments} pending commitment transaction{s}'
+                          f' has been dropped from mempool.')
+            return False
+
+        # txs_in_mempool < len(self.__pending)
+        # We are missing some Txs, but none of them are commitment TXs
+        return True
+
+    def __get_tx_pending(self):
+        pending_block = self.client.w3.eth.getBlock(block_identifier='pending', full_transactions=True)
```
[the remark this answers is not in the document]
piotr-roslaniec: Once per `WorkTracker` interval, so at most once in 15 mins.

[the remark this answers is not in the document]
cygnusv: That's approximately 100 times per day. We need to make sure that doesn't incur in any rate limitation for, e.g. Infura or alchemy providers. Also, how long does it take to get the full "block" of pending transactions? It seems that can be heavy...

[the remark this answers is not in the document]
piotr-roslaniec: Previously, we were making a different kind of RPC call here (`eth_getTransactionCount`), so introducing `__get_tx_pending` (or `eth_getBlockByNumber`) doesn't increase the Infura request rate. 
According to my benchmark `eth_getBlockByNumber` RPC call takes 250-300ms.

nucypher/blockchain/eth/token.py
```diff
@@ -649,29 +648,42 @@ def pending(self) -> Dict[int, HexBytes]:
...
 
         if txs_in_mempool > len(self.__pending):  # We're missing some pending TXs
             return False
-        else:  # TODO #2429: What to do when txs_in_mempool < len(self.__pending)? What does this imply?
-            return True
+
+        self.__num_untracked_commitments = len(set(tx_pending) - set(self.__pending))
+        # Mempool has limit on the number on TXs, and may drop them resulting in missing pending TXs
+        if self.__num_untracked_commitments != 0:
+            s = "s" if self.__num_untracked_commitments > 1 else ""
```
derekpierre, line 664: I'm a little confused. We define `__get_tx_pending()` as:
```python
    def __get_tx_pending(self):
        pending_block = self.client.w3.eth.getBlock('pending', True)
        tx_pending = [
            tx.hash for tx in pending_block['transactions']
            if tx.hash in self.__pending.values()
        ]
        return tx_pending
```

So basically `tx_pending` is always a subset of `self.__pending`. Therefore, what am I missing that allows `self.__num_untracked_commitments`, which is the difference between the two, to ever be > 1?

nucypher/blockchain/eth/token.py
```diff
@@ -687,12 +699,8 @@ def __track_pending_commitments(self) -> bool:
             s = "s" if unmined_transactions > 1 else ""
             self.log.info(f'{unmined_transactions} pending commitment transaction{s} detected.')
 
-        inconsistent_tracker = not self.__commitments_tracker_is_consistent()
-        if inconsistent_tracker:
-            # If we detect there's a mismatch between the number of internally tracked and
-            # pending block transactions, create a special pending TX that accounts for this.
-            # TODO: Detect if this untracked pending transaction is a commitment transaction at all.
-            self.__pending[0] = UNTRACKED_PENDING_TRANSACTION
+        is_inconsistent = not self.__commitments_tracker_is_consistent()
```
derekpierre, line 702: 👍

nucypher/blockchain/eth/token.py
piotr-roslaniec: That's a mistake, it should actually be the other way around:
`self.__num_untracked_commitments = len(set(self.__pending)) - set(tx_pending)`.

The logic is that if the number of internally tracked pending transactions `self.__pending` is larger than the number of pending transactions in the block `tx_pending`, it would mean that one of the commitment transactions was dropped from the mempool.

Thanks for catching this!

newsfragments/2686.misc.rst
```diff
@@ -0,0 +1 @@
+Ursula will now detect absence of active stakes during startup and in worker loop.
```
derekpierre, line 1: Is this the correct newsfragment for this PR?

KPrasch (commented):
Has anyone tested this on a testnet or mainnet node?

derekpierre:
> Has anyone tested this on a testnet or mainnet node?

Admittedly, I have not.

newsfragments/2686.misc.rst
KPrasch: Bump - I see the edge case handling, but not the check for active stakes.  Am I missing it right in front of me?

piotr-roslaniec: No, this newsfragment is entirely incorrect and was meant for a different PR. That was a blunder.

piotr-roslaniec:
I tested it locally by replacing some code fragments to stub out the blockchain state. I will provide a `diff` so that we can evaluate it and test PR locally.

PR #2687, opened 2021-05-08
[EPIC] Chili: TreasureMaps con KFrags

jMyles:
This replaces #2530 and incorporates subsequent work from @KPrasch (with touches from @jMyles).

Description modified from #2530:

Issues addressed:
* #2234: "TreasureMaps con KFrags" 
* #234: Ursula responds to consider_arrangement
* #1701

Issues to address in a follow-up or related PR:
* #2736 
* #2737 
* #343: Versioning TMaps (Partially implemented / needs follow-up)
* #2071: Improve Alice granting resilience
* ~~#2225: Deterministic keyring derivation~~ Merged
* #1700 

----
*Note:* This PR underwent a conflict-fraught rebase from 64c8a789c to f098aac3d.  Some of @fjarri's work, particularly 26076b1d9, may have been impacted.


nucypher/characters/lawful.py
```diff
@@ -604,7 +624,7 @@ def _try_orient(self, treasure_map, alice_verifying_key):
             raise  # TODO: Maybe do something here?  NRN
 
     def get_treasure_map(self, alice_verifying_key, label):
-        map_identifier = self.construct_map_id(verifying_key=alice_verifying_key, label=label)
+        map_identifier = self.construct_map_id(relayer_verifying_key=alice_verifying_key, label=label)
```
jMyles, line 627: Nice turn-of-phrase.

nucypher/characters/lawful.py
```diff
@@ -1166,7 +1242,10 @@ def __init__(self,
             certificate_filepath = self._crypto_power.power_ups(TLSHostingPower).keypair.certificate_filepath
             certificate = self._crypto_power.power_ups(TLSHostingPower).keypair.certificate
 
-            # only you can prevent forest fires
+            # Only *YOU* can prevent forest fires
```
jMyles, line 1245: phew... glad this line got fixed.

nucypher/characters/lawful.py
```diff
@@ -1727,6 +1806,37 @@ def interface_info_with_metadata(self):
     # Re-Encryption
     #
 
+    def verify_kfrag_authorization(self,
```
jMyles, line 1809: Most significant new symbol in this PR, IMO.

nucypher/characters/lawful.py
```diff
@@ -1727,6 +1806,37 @@ def interface_info_with_metadata(self):
...
+
+        try:
+            self.verify_from(alice, reconstructed_writ, signature=writ_signature)
+        except InvalidSignature:
+            raise Policy.Unauthorized  # This isn't from Alice (relayer).
+
+        if writ_hrac != work_order.hrac:  # Funky Workorder
+            raise Policy.Unauthorized  # Bob, what the *hell* are you doing?
+
+        kfrag_checksum = keccak_digest(bytes(kfrag))[:WRIT_CHECKSUM_SIZE]
```
jMyles, line 1828: Any particular reason we're using a separate constant here instead of `KECCAK_DIGEST_LENGTH`?

nucypher/characters/lawful.py
KPrasch: 🔥 🌲

KPrasch (approved):
I am one of the author's of this PR and I approve! 😅 

Seriously, it's good to get a merge on this long-awaiting and impactful PR so we can begin to work with the new design assumptions.

nucypher/blockchain/eth/actors.py
```diff
@@ -1202,6 +1206,26 @@ def missing_commitments(self) -> int:
         missing = self.staking_agent.get_missing_commitments(checksum_address=staker_address)
         return missing
 
+    def verify_policy_payment(self, hrac: bytes) -> None:
+        arrangements = self.policy_agent.fetch_policy_arrangements(policy_id=hrac)
```
fjarri, line 1210: Let's choose whether to call it HRAC or PolicyID

nucypher/blockchain/eth/agents.py
```diff
@@ -792,6 +795,8 @@ class PolicyManagerAgent(EthereumContractAgent):
         'finishUpgrade'
     )
 
+    ArrangementInfo = namedtuple('ArrangementInfo', ['node', 'downtime_index', 'last_refunded_period'])
```
fjarri, line 798: Probably better to use subclass of `NamedTuple`

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -39,13 +35,17 @@ def _deserialize(self, value, attr, data, **kwargs):
...
         try:
-            signature, hrac, tmap_message_kit = splitter(value)
-            return True
-        except InvalidNativeDataTypes as e:
-            raise InvalidInputData(f"Could not parse {self.name}: {e}")
+            # Unsigned TreasureMap (Federated)
+            from nucypher.policy.maps import TreasureMap as UnsignedTreasureMap
+            splitter = UnsignedTreasureMap.get_splitter(value)
+            splitter(value)
+        except InvalidNativeDataTypes:
```
fjarri, line 43: Why not just always sign the treasure map? Do we still use 65-byte signatures for the blockchain case? If that's a normal signature, everything will be perfectly compatible.

nucypher/characters/lawful.py
```diff
@@ -602,7 +630,7 @@ def _try_orient(self, treasure_map, alice_verifying_key):
             raise  # TODO: Maybe do something here?  NRN
 
     def get_treasure_map(self, alice_verifying_key, label):
```
fjarri, line 632: Should be `relayer_verifying_key`?

nucypher/characters/lawful.py
```diff
@@ -804,6 +829,8 @@ def _reencrypt(self,
                                                        receiving_pk=keys['receiving'])
             except VerificationError:
                 # TODO: WARNING - This block is untested.
+                from nucypher.policy.disputes import IndisputableEvidence
```
fjarri, line 832: Rebase artefact

nucypher/characters/lawful.py
```diff
@@ -886,11 +874,13 @@ def retrieve(self,
 
         for message_kit in message_kits:
             message_kit.set_correctness_keys(receiving=self.public_keys(DecryptingPower))
-            message_kit.set_correctness_keys(verifying=alice_verifying_key)
+            message_kit.set_correctness_keys(verifying=relayer_verifying_key)
```
fjarri, line 877: Should be Alice's key here

nucypher/characters/lawful.py
```diff
@@ -962,6 +964,90 @@ def retrieve(self,
...
+                 *message_kits: UmbralMessageKit,
+                 label: bytes,
+
+                 # Alice(s)
+                 alice_verifying_key: Union[PublicKey, bytes],
+                 relayer_verifying_key: Optional[Union[PublicKey, bytes]] = None,
+
+                 # Source Authentication
+                 enrico: "Enrico" = None,
+                 policy_encrypting_key: PublicKey = None,
```
fjarri, line 1009: Not really source authentication

nucypher/network/middleware.py
```diff
@@ -191,14 +190,7 @@ def propose_arrangement(self, node, arrangement):
...
                                     data=bytes(arrangement),
-                                    timeout=2)
-        return response
-
-    def enact_policy(self, ursula, kfrag_id, payload):
-        response = self.client.post(node_or_sprout=ursula,
-                                    path=f'kFrag/{kfrag_id.hex()}',
-                                    data=payload,
-                                    timeout=2)
+                                    timeout=120)  # TODO: What is an appropriate timeout here?
```
fjarri, line 193: Left from debugging?

nucypher/network/server.py
```diff
@@ -359,17 +297,18 @@ def receive_treasure_map():
...
+
         if not this_node.federated_only:
-            from nucypher.policy.collections import SignedTreasureMap as _MapClass
+            from nucypher.policy.maps import SignedTreasureMap as _MapClass
         else:
-            from nucypher.policy.collections import TreasureMap as _MapClass
+            from nucypher.policy.maps import TreasureMap as _MapClass
 
         # Step 1: First, we verify the signature of the received treasure map.
-        # This step also deserializes the treasure map iff it's signed correctly.
```
fjarri, line 372: Was it a typo or "if and only if"?

nucypher/policy/policies.py
```diff
@@ -43,40 +42,30 @@ class Arrangement:
     """
     A contract between Alice and a single Ursula.
     """
-    ID_LENGTH = 32
 
-    splitter = BytestringSplitter(key_splitter,  # alive_verifying_key
-                                  (bytes, ID_LENGTH),  # arrangement_id
-                                  (bytes, VariableLengthBytestring))  # expiration
+    splitter = BytestringSplitter((PublicKey, PublicKey.serialized_size()),  # alice_verifying_key
```
fjarri, line 46: Isn't it what `key_splitter` is?

nucypher/policy/policies.py
```diff
@@ -115,8 +104,24 @@ def block_until_success_is_reasonably_likely(self):
...
+
+        successes = self._worker_pool.get_successes()
+        responses = {ursula.checksum_address: status for ursula, status in successes.items()}
+        if not all(response.status_code == 201 for response in responses.values()):
+            report = "\n".join(f"{address}: {status}" for address, status in responses.items())
+            self.log.debug(f"Policy enactment failed. Request statuses:\n{report}")
+
+            # OK, let's check: if any Ursulas claimed we didn't pay,
+            # we need to re-evaluate our situation here.
+            claims_of_freeloading = sum(response.status_code == 402 for response in responses.values())
```
fjarri, line 118: Small thing, but, given how it is used, could be just `any` instead of `sum`

nucypher/policy/policies.py
```diff
@@ -171,23 +176,42 @@ class Policy(ABC):
     An edict by Alice, arranged with n Ursulas, to perform re-encryption for a specific Bob.
     """
 
-    POLICY_ID_LENGTH = 16
+    ID_LENGTH = POLICY_ID_LENGTH
```
fjarri, line 179: Why do we have an additional constant here? Could it be just `ID_LENGTH = 16`, and whoever needs it would refer to it as `Policy.ID_LENGTH`?

nucypher/types.py
```diff
@@ -92,3 +92,18 @@ class StakerInfo(NamedTuple):
...
+
+class Policy(NamedTuple):
+    disabled: bool
+    sponsor: ChecksumAddress
+    owner: ChecksumAddress
+    fee_rate: Wei
+    start_timestamp: int
+    end_timestamp: int
+
+    # reserved_slot_1
```
fjarri, line 105: In what sense is it reserved?

tests/unit/test_collections.py
```diff
@@ -0,0 +1,88 @@
+"""
```
fjarri, line 1: Since there's no `collections.py` anymore, perhaps better named `test_treasure_map` or something?

nucypher/blockchain/eth/agents.py
```diff
@@ -814,29 +819,21 @@ def create_policy(self,
         return receipt
 
     @contract_api(CONTRACT_CALL)
-    def fetch_policy(self, policy_id: str, with_owner=False) -> Union[tuple, list]:
+    def fetch_policy(self, policy_id: bytes) -> Policy:
```
derekpierre, line 822: was there a reason we needed to change the type from `str` to `bytes`  for the `policy_id`?

nucypher/blockchain/eth/agents.py
```diff
@@ -68,6 +70,7 @@
     StakerInfo,
     PeriodDelta,
     StakingEscrowParameters,
+    Policy
```
derekpierre, line 73: I like that `ArrangementInfo` is defined in PolicyManagerAgent, and find it a little strange that we have some other datatypes that are specific to contracts declared somewhere other than in the respective agent eg. can `Policy` be declared in PolicyAgent instead of in `nucypher/types.py`; can `StakingEscrowParameters` be declared in StakingEscrowAgent instead of in `nucypher/types.py`...?

May not apply to all classes defined in `nucypher/types.py`, but perhaps some/most?

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -39,13 +35,17 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+            try:
```
derekpierre, line 44: This nested try-except will be improved in the Porter EPIC, but knowing the context (federated or not) when this validation is performed.

nucypher/blockchain/eth/agents.py
derekpierre: Also, in keeping with the naming of other similar classes - wdyt about `PolicyInfo` instead of `Policy`?

nucypher/characters/lawful.py
```diff
@@ -683,61 +710,59 @@ def get_treasure_map_from_known_ursulas(self, network_middleware, map_identifier
...
-        if treasure_map:
-            treasure_map_to_use = treasure_map
-        else:
-            try:
-                treasure_map_to_use = self.treasure_maps[map_id]
-            except KeyError:
-                raise KeyError(f"Bob doesn't have the TreasureMap {map_id}; can't generate work orders.")
+        if not alice_verifying_key:
+            # Assume the policy publisher is the same as the KFrag generator by default.
+            alice_verifying_key = relayer_verifying_key
```
derekpierre, line 724: I might be missing something, but this seems backward to me i.e. shouldn't alice_verifying_key should always be provided, and the relayer_verifying_key should be optional...? and then we get:

```python
if not relayer_verifying_key:
    relayer_verifying_key = alice_verifying_key
```

The KFrag generator would be alice_verifying_key, which is always the case, and then the publisher i.e. relayer could either be alice or a separate entity...

nucypher/characters/lawful.py
```diff
@@ -962,6 +964,90 @@ def retrieve(self,
...
+
+            self._try_orient(treasure_map, relayer_verifying_key)
+            # self.treasure_maps[treasure_map.public_id()] = treasure_map # TODO: Can we?
+        else:
+            map_id = self.construct_map_id(relayer_verifying_key, label)
+            try:
+                treasure_map = self.treasure_maps[map_id]
+            except KeyError:
+                # If the treasure map is not known, join the policy as part of retrieval.
+                self.join_policy(label=label, alice_verifying_key=relayer_verifying_key)
```
derekpierre, line 991: Should join_policy take a `relayer_verifying_key` as the parameter then instead of an `alice_verifying_key`?

nucypher/characters/lawful.py
```diff
@@ -1716,7 +1806,41 @@ def interface_info_with_metadata(self):
...
+            raise Policy.Unauthorized  # Bob, Seriously?
+
+        try:
+            verified_kfrag = kfrag.verify(verifying_pk=alice.stamp.as_umbral_pubkey())
+        except VerificationError:
+            raise Policy.Unauthorized  # WTF, Alice did not generate these KFrags.
+
+        if writ_hrac in self.revoked_policies:
+            # Note: This is only an off-chain and in-memory check.
+            raise Policy.Unauthorized  # Denied
```
derekpierre, line 1839: Do we want to differentiate the reason for these raised `Policy.Unauthorized` exceptions (L1830, L1835, L1839) by contextualizing with a string message? Or is it unsafe to leak the reason?

nucypher/network/server.py
```diff
@@ -203,148 +157,132 @@ def node_metadata_exchange():
...
+                message = f"{bob_identity_message} Policy {hrac.hex()} is not a published policy."
+                return Response(message, status=404)  # 404 - Not Found
+
+            # Verify Active Policy (onchain)
+            try:
+                this_node.verify_active_policy(hrac=hrac)
+            except Policy.Inactive:
+                message = f"{bob_identity_message} Policy {hrac.hex()} is not active."
+                return Response(message, status=403)  # 403 - Forbidden
+            except this_node.Policy.Expired:
```
derekpierre, line 256: Love these exception types - so clean ❤️

nucypher/network/server.py
```diff
@@ -397,56 +336,92 @@ def receive_treasure_map():
...
 
             # Check that this treasure map is from Alice per the Policy.
-            if not received_treasure_map.verify_blockchain_signature(checksum_address=alice_checksum_address):
+            if not received_treasure_map.verify_blockchain_signature(checksum_address=policy.owner):
                 log.info(f"Bad TreasureMap ID; not storing {map_identifier}")
                 return Response("This TreasureMap doesn't match a paid Policy.", status=402)
 
             # Check that this treasure map is valid for the Policy datetime and that it's not disabled.
-            if policy_data[0] or datetime.utcnow() >= datetime.utcfromtimestamp(policy_data[5]):
+            if policy.disabled or datetime.utcnow() >= datetime.utcfromtimestamp(policy.end_timestamp):
```
derekpierre, line 351: 🔥 to use named members.

tests/metrics/estimate_gas.py
```diff
@@ -288,8 +288,8 @@ def transact(function, transaction):
     #
     # Create policy
     #
-    policy_id_1 = os.urandom(int(Policy.POLICY_ID_LENGTH))
-    policy_id_2 = os.urandom(int(Policy.POLICY_ID_LENGTH))
+    policy_id_1 = os.urandom(int(POLICY_ID_LENGTH))
```
derekpierre, line 291: Seems strange that `POLICY_ID_LENGTH` is no longer defined in the `Policy` class - any reason in particular for that?

tests/unit/test_collections.py
KPrasch: ✔️

nucypher/characters/lawful.py
KPrasch: ✔️ - In isolation I think this is more meaningful as "publisher" instead of "relayer".  what do you think?

nucypher/characters/lawful.py
KPrasch: ✔️ Adding more keys is definitely a trade-off between API flexibility and complexity here.

nucypher/blockchain/eth/actors.py
fjarri: What worries me is that we have `HRAC_LENGTH` (in `crypto.constants`) and `POLICY_ID_LENGTH` (in `blockchain.eth.constants`). Currently they're equal. What happens if we decide to change one of them? 

There would be a point to having both of them if we merely used HRAC as a policy ID, but also for other purposes (and even than, I'd prefer to have something like a `hrac_as_policy_id()` function), but now it *is* a policy ID, isn't it? We can just rename `TreasureMap.derive_hrac()` to `TreasureMap.derive_policy_id()` and not lose any meaning, can't we?

nucypher/characters/control/specifications/fields/treasuremap.py
fjarri: > It might be enough to use the stamp instead of the transacting power to sign

We can use the same "trampoline" we had for `IndisputableEvidence`: sign the stamp with transacting power, and stamp the map. But then we'll have to have some mechanism of discovery for these "identity evidences"... Which may be hard, Alice is not Ursula and isn't always online.

nucypher/characters/lawful.py
fjarri: I agree, "publisher" provides more intuition to what the character does. Do we still use names for characters? Pablo the Publisher?

nucypher/network/middleware.py
fjarri: It feels that when the maximum wait is 2 minutes, some form of a progress bar is justified (or at the very least, a message for the user). I'd think the whole thing hung if I had to wait a minute.

nucypher/network/server.py
fjarri: I think I was over-cautious, and it's just a typo. There's no need for mathematical precision in this context (in fact, "iff" would be wrong, since that would imply that if the map is correctly signed, it is *always* deserialized, which may not be the case).

nucypher/types.py
fjarri: Ah, I see. Could you add a short comment here, something along the lines of "reserved but unused fields in the corresponding Solidity structure"?

nucypher/blockchain/eth/agents.py
fjarri: I am with Derek here regarding `PolicyInfo` vs `Policy`. We already have a `Policy` (defined in `policies.py`), which is a different type (with more complexity too, not just a collection of simple values). Makes sense to follow the example of `Arrangement` (a Python class) vs `ArrangementInfo` (a mapping of a Solidity structure).

KPrasch: Renamed `Policy` -> `PpolicyInfo`

tests/metrics/estimate_gas.py
fjarri: I think, the general guideline is that if the constant is used locally in the class, then it should be a class constant. In our case, `POLICY_ID_LENGTH` is a part of the interop protocol with Solidity, so it makes sense to have it in `eth.constants`. (That invalidates my other comment - after some thought I agree that it was a correct decision to move it there). 

There are some weird parts related to that in `Policy`. We have in `Policy.__init__()`
```
self._id = construct_policy_id(self.label, bytes(self.bob.stamp))
```
but it's not really a policy ID, and it does not seem to be used anywhere except for `repr()` (which will be confusing on debug). `construct_policy_id()` does not use `POLICY_ID_LENGTH` either. 

`POLICY_ID_LENGTH` is used in `_make_enactment_payload()`, to cut the HRAC which is already the correct length (because we use HRAC as a policy ID elsewhere). This relates to my previous comment about HRACs - wouldn't it be much more logical to just have it as a policy ID?

nucypher/policy/policies.py
fjarri: Disregard that, see my comment below.

tests/metrics/estimate_gas.py
KPrasch: > This relates to my previous comment about HRACs - wouldn't it be much more logical to just have it as a policy ID?

To clarify - you mean, remove the notion of HRAC entirely in favor of Policy ID?  

I am a bit concerned that I may have been too presumptuous to change policy ID to HRAC in this PR.  Ultimately it;s changing how policies are indexed on-chain.

fjarri: > you mean, remove the notion of HRAC entirely in favor of Policy ID?

Yep. And only create it in one place, too. Currently we have in `TreasureMap.derive_hrac()`
```
self._hrac = keccak_digest(bytes(alice_stamp) + bytes(bob_verifying_key) + label)[:HRAC_LENGTH]
```
And in `Policy.__init__` + `BlockchainPolicy._make_enactment_payload()`:
```
self.hrac = keccak_digest(bytes(self.alice.stamp) + bytes(self.bob.stamp) + self.label)[:HRAC_LENGTH]
...
bytes(self.hrac)[:self.ID_LENGTH]
```
But these methods are supposed to produce the same result, which is a policy ID.

KPrasch: I agree about consolidating the derivation functions themselves, I'll do that now.  I still have two outstanding concerns:

1. I'm not completely convinced that we want to solidify the union of HRAC and Policy ID and I'd like to have a focused discussion on the implications of deprecating policy ID completely.
2. From a strictly naming point of view, I much prefer the notion of HRAC over policy ID.  It's more semantically dense and accurate since knowledge of the HRAC bytes represent some degree of access/authentication.

nucypher/characters/lawful.py
KPrasch: Renamed `relayer_verifying_key` -> `publisher_verifying_key` for now.

nucypher/types.py
KPrasch: ✔️

tests/metrics/estimate_gas.py
fjarri: > From a strictly naming point of view, I much prefer the notion of HRAC over policy ID.

That makes sense. As long as we avoid having two entities which are really the same - I don't think we used HRAC for anything other than a policy IDs, right?

nucypher/characters/lawful.py
KPrasch: ✔️ Renamed relayer -> publisher.

nucypher/policy/policies.py
KPrasch: ✔️

nucypher/policy/policies.py
KPrasch: ✔️

tests/acceptance/blockchain/actors/test_investigator.py
```diff
@@ -0,0 +1,108 @@
...
+                                  signer=Signer(ursula_privkey))
+
+    signed_stamp = testerchain.client.sign_message(account=account,
+                                                   message=bytes(ursula_stamp))
+
+    ursula = mocker.Mock(stamp=ursula_stamp, decentralized_identity_evidence=signed_stamp)
+    return ursula
+
+
+@pytest.mark.skip("David, send help!")
```
KPrasch, line 44: @cygnusv - Heads up for you here.

nucypher/characters/lawful.py
KPrasch: ✔️

nucypher/characters/lawful.py
KPrasch: ✔️

nucypher/characters/lawful.py
KPrasch: Help me clarify here: We expect that the KFrag generator constructed this MessageKit, and not the publisher?

nucypher/blockchain/eth/agents.py
derekpierre: Presumably, the contract API always took bytes i.e. from before and would not be due to new code ...

I only ask because I think we've tried to limit the surface area passing bytes around.

nucypher/characters/lawful.py
derekpierre: Feels like it - the one raised from the `VerificationError` (L1835) may be interesting...

nucypher/blockchain/eth/agents.py
derekpierre: > Do you have a preference for consistency? If it's the namepacing that matters most to you perhaps we make a typing module per contract?

I think some sort of more direct association between the types defined and the contracts would make them more intuitive. Having them live in `nucypher/types.py` just seems vague; it could be a typing module per contract or even keeping it general like `nucypher.blockchain.eth.types`...?

tests/metrics/estimate_gas.py
derekpierre: 👍 . It slipped my mind that policy_id in the Solidity contract is `bytes16` and that's the association - was too focused on the fact that the hrac gets truncated in Policy python class.

nucypher/characters/lawful.py
fjarri: The verifying key here is the verifying key that was used to sign kfrags, so it should be the Alice's key. It's actually not related to the creator of the message kit at all (and we don't even serialize it with the kit); currently `UmbralMessageKit` only carries kfrag verification keys for convenience.

nucypher/blockchain/eth/agents.py
derekpierre: ^ BTW Doesn't have to be this PR - maybe the namespacing can be looked into as part of a separate PR.

nucypher/blockchain/eth/actors.py
KPrasch: Opened https://github.com/nucypher/nucypher/issues/2736

nucypher/characters/lawful.py
KPrasch: 👍🏻 Renamed to "publisher" for accuracy and readability.  Meant to indicate "Policy and Treasure Map Publisher".

nucypher/characters/control/specifications/fields/treasuremap.py
KPrasch: Opened https://github.com/nucypher/nucypher/issues/2737

nucypher/network/middleware.py
KPrasch: Opened https://github.com/nucypher/nucypher/issues/2738

nucypher/policy/maps.py
```diff
@@ -0,0 +1,384 @@
...
+                                   alice_stamp=alice.stamp)
+
+        # Sign the map if needed before sending it out into the world.
+        treasure_map.prepare_for_publication(bob_encrypting_key=bob.public_keys(DecryptingPower),
+                                             alice_stamp=alice.stamp)
+
+        return treasure_map
+
+
+# FIXME: a dirty hack to make the tests pass. Fix it ASAP.
```
KPrasch, line 322: See #343 #281

derekpierre (approved):
🎸 - approving in the interest of time/traction; my 2 outstanding comments can be addressed (or looked into and rejected) as part of follow-up issues/PRs.

https://github.com/nucypher/nucypher/pull/2687#discussion_r664955563 (error messages)
https://github.com/nucypher/nucypher/pull/2687#discussion_r665380700 (namespacing of types)

nucypher/characters/lawful.py
derekpierre: Alternatively, we could do sub-classes of Policy.Unauthorized if differentiation is worth it.

nucypher/blockchain/eth/agents.py
derekpierre: Filed #2739 .

nucypher/characters/lawful.py
derekpierre: Filed https://github.com/nucypher/nucypher/issues/2740.

nucypher/blockchain/eth/agents.py
KPrasch: My impression is that the hint for `str` was just incorrect.  The `PolicyManager` itself contract takes `bytes16`.  Using a string encoding and format actually seems more complex to me.

derekpierre: 👍

fjarri (approved):

PR #2688, opened 2021-05-09
Check for active stakes when starting worker and firing commitments

### LINKED ISSUE #1515: Automatically shutdown worker at end of terminal stake.
When all stakes for a stakeholder end, we should notify the stakeholder with better error handling and output.

Currently:
`missing commitments` will continue to increment each period.
an exception (detailed below) is generated on the repeated attempts to commit by the worktracker.

Ideally, the lack of an active stake for the bonded worker should prevent `Ursula run` from starting.

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Stops worker during its initialization if it has no stakes to take care of 
- Doesn't fire commitment for a given period if there are no active stakes

**Issues fixed/closed:**
- Fixes #1515

**Why it's needed:**

**Notes for reviewers:**
- CLI test is broken on `main`


nucypher/blockchain/eth/token.py
```diff
@@ -883,3 +889,10 @@ def __read_stakes(self) -> None:
 
         # Record most recent cache update
         self.__updated = maya.now()
+
+    @property
+    def has_active_substakes(self) -> bool:
+        current_period = self.staking_agent.get_current_period()
+        for stake in self.data:
+            if not stake.status(current_period=current_period).is_child(Stake.Status.INACTIVE):
```
derekpierre, line 897: You don't need to pass `current_period` here, if `status()` doesn't receive a current_period value it derives it - not a big deal but would make this method look even cleaner.

(From `status()` method)
```python
current_period = current_period or self.staking_agent.get_current_period()  # TODO #1514 this is online only.
```

piotr-roslaniec: I should have left the comment there: I did it intentionally to avoid multiple calls to contract. I thought it was neat, but maybe it's a bit pendantic.

derekpierre: `get_current_period()` would only be called once either way no? Either in this function, `has_active_substakes` or in `status()` - my comment was purely a simplicity change, not a functional change.

piotr-roslaniec: Yes, it would be called once per stake in a loop. So I figured I could just call it once outside the loop.

derekpierre: 🤦 - You are totally right - ignore me.

newsfragments/2688.misc.rst
```diff
@@ -0,0 +1 @@
+Ursula will now check for active stakes on startup.
```
derekpierre, line 1: wdyt about this as a feature newsfragment instead? Feels like it should be more prominent since nodes will not start. Just food for thought...

derekpierre (approved):
🎸

KPrasch (approved):
Good idea.  This will address a good portion of folks with odd end-of-stake behavior.

PR #2692, opened 2021-05-12
[WIP] Use pathlib

### LINKED ISSUE #1954: Migrate os.path joins to Pathlib
Throughout the codebase, we can gain better OS level compadibillity, and better readability by migrating our code to use [Pathlib](https://docs.python.org/3/library/pathlib.html) wherever possible.

#### Current API

```python
import os.path

data_folder = os.path.join("source_data", "text_files")

file_to_open = os.path.join(data_folder, "raw_data.txt")
```

#### Proposed API
```python
from pathlib import Path

data_folder = Path("source_data/text_files/")

file_to_open = data_folder / "raw_data.txt"
```

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
- Replaces `str` usage with `pathlib.Path`
- Replaces calls to functions in `os.path` module with appropriate calls to `pathlib.Path` methods
- Introduces changes to `serialize` and `deserialize` character methods

**Issues fixed/closed:**
- Fixes #1954 

**Why it's needed:**

**Notes for reviewers:**
- Line count is high, but 3/4 of that is essentially "search and replace"


derekpierre:
@piotr-roslaniec there is a CI failure with the updated version of trezor, you may need to relock dependencies i.e. run `scripts/dependencies/relock_dependencies.sh`.

nucypher/config/base.py
```diff
@@ -293,9 +303,19 @@ def deserialize(cls, payload: str, deserializer=json.loads, payload_label: Optio
             label = f"'{payload_label}' " if payload_label else ""
             raise cls.OldVersion(f"Configuration {label}is the wrong version "
                                  f"Expected version {cls.VERSION}; Got version {version}")
+
+        # TODO: Move to implementation of `deserialize` in their respective classes
```
derekpierre, line 307: definitely!

nucypher/config/base.py
```diff
@@ -188,17 +188,17 @@ def generate_filename(cls, modifier: str = None) -> str:
...
-    def default_filepath(cls, config_root: str = None) -> str:
+    def default_filepath(cls, config_root: Path = None) -> Path:
         """
         Generates the default configuration filepath for the class.
 
         :return: The generated filepath string
         """
         filename = cls.generate_filename()
-        default_path = os.path.join(config_root or cls.DEFAULT_CONFIG_ROOT, filename)
+        default_path = Path(config_root or cls.DEFAULT_CONFIG_ROOT) / filename
```
derekpierre, line 198: DEFAULT_CONFIG_ROOT is a Path object already can this be simplified to the following?
```suggestion
        default_path = (config_root or cls.DEFAULT_CONFIG_ROOT) / filename
```

nucypher/config/base.py
```diff
@@ -606,20 +621,23 @@ def produce(self, **overrides) -> CHARACTER_CLASS:
...
                                              federated_only=payload['federated_only'])
-        domain = payload['domain']
         max_gas_price = payload.get('max_gas_price')  # gwei
         if max_gas_price:
             max_gas_price = Decimal(max_gas_price)
 
         # Assemble
-        payload.update(dict(node_storage=node_storage, domain=domain, max_gas_price=max_gas_price))
+        payload.update(dict(node_storage=node_storage, max_gas_price=max_gas_price))
+        for key in ('keyring_root', 'filepath', 'config_root', 'registry_filepath'):
```
derekpierre, line 637: I worry a little about code maintenance here i.e. ensuring that this set tuple of keys is kept up to date. Unsure what can be done though...

nucypher/config/storages.py
```diff
@@ -476,29 +475,26 @@ def remove(self, checksum_address: str, metadata: bool = True, certificate: bool
...
+        def __destroy_dir_contents(path: Path) -> None:
+            if not path.is_dir():
                 return
-            else:
-                for file in paths_to_remove:
-                    file_path = os.path.join(path, file)
-                    if os.path.isfile(file_path):
-                        os.unlink(file_path)
+            for file in path.iterdir():
+                file_path = path / file
```
derekpierre, line 495: I think (?) the objects returned by `iterdir()` are already Path objects to the relevant file...

nucypher/config/storages.py
```diff
@@ -476,29 +475,24 @@ def remove(self, checksum_address: str, metadata: bool = True, certificate: bool
...
+        def __destroy_dir_contents(path: Path) -> None:
+            if not path.is_dir():
                 return
-            else:
-                for file in paths_to_remove:
-                    file_path = os.path.join(path, file)
-                    if os.path.isfile(file_path):
-                        os.unlink(file_path)
+            for dir_item in path.iterdir():
+                dir_item.unlink()
```
derekpierre, line 495: To remain in line with the old code do we need to check for `is_file()` before unlinking? Unless the logic was already flawed and this fixes it...

piotr-roslaniec: `known_nodes/certificates` and `known_nodes/metadata` are flat directories with files only so I thought having anything out of ordinary there should raise an error.

derekpierre: Understood. I only pointed it out because of the original logic that was there. 

These are folders on the system that people can manipulate directly however they want eg. play around with directories/files (hopefully they don't 😅 ), and we probably don't want the node to necessarily crash/not function properly because the user created a folder there that doesn't directly affect the functioning of the node.

I would vote to keep the original logic that checks `isfile()` and perhaps log a warning if not a file, instead of raising an error.

Perhaps others have different opinions.

piotr-roslaniec: I see your point, I think the original logic may even be extended to take more ownership of those directories:

```
        def __destroy_dir_contents(path: Path) -> None:
            if not path.is_dir():
                return
            for dir_item in path.iterdir():
                if dir_item.is_file():
                    dir_item.unlink()
                elif dir_item.is_dir():
                    __destroy_dir_contents(dir_item)
            path.rmdir()```

Comment (derekpierre) [reply nucypher/config/storages.py:495]: In the context of what this PR is aiming to achieve, we shouldn't get too out of scope - perhaps something to look at in a follow-up PR - eg. original code did not delete the parent directory. I think simply checking for a file and unlinking could be done here (same as original code), and then you can look at a follow-up PR.

**Comment by KPrasch** [comment]
Hefty amount of conflicts now, but I still think it will be good to get this PR merged!

**Comment by piotr-roslaniec** [comment]
Rebased on `main`

File: nucypher/characters/chaotic.py [hunk 6]
```diff
@@ -112,7 +113,7 @@ def __init__(self,
         # Database
         self.db_filepath = db_filepath
         self.db = NO_DATABASE_AVAILABLE
-        self.db_engine = create_engine(f'sqlite:///{self.db_filepath}', convert_unicode=True)
+        self.db_engine = create_engine(f'sqlite:///{self.db_filepath.absolute()}', convert_unicode=True)
```
KPrasch, line 116: I  cannot seem to find the documentation for this method - Does it return a path with a leading slash?

piotr-roslaniec: It [creates](https://docs.sqlalchemy.org/en/14/core/engines.html#sqlalchemy.create_engine) a new `sqlachemy.engine.Engine` instance.

KPrasch: Oh, I meant the `absolute` method.

piotr-roslaniec: Yes. That should give us four leading slashes in total, which is a correct [connection URI](https://flask-sqlalchemy.palletsprojects.com/en/2.x/config/#connection-uri-format).

docs-requirements.txt
```diff
@@ -6,7 +6,7 @@ babel==2.9.1; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2,
 certifi==2021.5.30
 chardet==4.0.0; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'
 click-default-group==1.2.2
-click==8.0.1; python_version >= '3.6'
+click==7.1.2; python_version >= '3.6'
```
KPrasch, line 9: Curious - what made you rollback the click version here?

tests/integration/config/test_storages.py
```diff
@@ -105,25 +105,25 @@ class TestTemporaryFileBasedNodeStorage(BaseTestNodeStorageBackends):
...
 
         # Let's break the metadata (but not the version)
-        metadata_path = os.path.join(self.storage_backend.metadata_dir, some_node)
+        metadata_path = self.storage_backend.metadata_dir / some_node
         with open(metadata_path, 'wb') as file:
             file.write(Learner.LEARNER_VERSION.to_bytes(4, 'big') + b'invalid')
 
         with pytest.raises(TemporaryFileBasedNodeStorage.InvalidNodeMetadata):
-            self.storage_backend.get(stamp=some_node[:-5],
+            self.storage_backend.get(stamp=str(some_node.name)[:-5],
```
derekpierre, line 116: is the `str(...)` needed? Is `.name` not a string already? Same for L126.

derekpierre (commented):
`keyring.py` and `test_keyring.py` seemed to have made their way back as empty files even though they were removed by the keystore work @KPrasch did - probably something with the rebase...

docs-requirements.txt
piotr-roslaniec: To avoid conflict with `requirements.txt`. Come to think of it, these dependencies should never conflict because we treat them as separate environments. WDYT?

derekpierre: If the docs requirements have a different version, conflicts can occur for local development and for circle ci builds.

If I run `pip install -e . -r docs-requirements.txt` locally I get the following:
```
ERROR: Cannot install click==8.0.1 and nucypher==5.3.0 because these package versions have conflicting dependencies.

The conflict is caused by:
    The user requested click==8.0.1
    nucypher 5.3.0 depends on click==7.1.2

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict
```

We've probably gotten away with this mismatch because of circle ci caching. It's best for the requirements to match. If updating one, then update the other, which I thought we did in the `relock_dependencies.sh` script...

The update of the docs-requirements to use the newer version of click seemed to occur in https://github.com/nucypher/nucypher/commit/b91dd386b849c748c077b4bfe34fdbeccb6b38cd as part of #2612 .

derekpierre: Probably need to do a relock of dependencies, and increment the circle ci cache.

nucypher/config/base.py
```diff
@@ -233,53 +233,62 @@ def _ensure_config_root_exists(self) -> None:
...
             raise FileExistsError(f"{filepath} exists and no filename modifier supplied.")
         with open(filepath, 'w') as file:
             file.write(self.serialize())
         return filepath
 
     def serialize(self, serializer=json.dumps) -> str:
         """Returns the JSON serialized output of `static_payload`"""
+        def _stringify_paths(d: dict):
+            for key, value in d.items():
+                if isinstance(value, PosixPath):
```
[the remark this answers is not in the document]
piotr-roslaniec: It should be `Path` instead, to maintain cross-platform compatibility (`nucypher` on Windows)

piotr-roslaniec:
I can see that rebase contains some changes with `str` usage - I will address them in this PR.

KPrasch (approved):
LGTM!

nucypher/config/base.py
```diff
@@ -293,9 +296,14 @@ def deserialize(cls, payload: str, deserializer=json.loads, payload_label: Optio
             label = f"'{payload_label}' " if payload_label else ""
             raise cls.OldVersion(f"Configuration {label}is the wrong version "
                                  f"Expected version {cls.VERSION}; Got version {version}")
+
+        if 'keyring_root' in deserialized_payload:
```
derekpierre, line 300: `keyring_root` became `keystore_path` as part of #2701. See https://github.com/nucypher/nucypher/commit/37a944df442f9df8ae8689408aa8a34472398287#diff-72cdbadc52be031579383783a54fab0b18127c72b9415b101270b55360be9f52R342

This was a concern I had regarding checking for specific keys, and what happens if they were changed. If we could somehow easily link them, that would be 👌 

Interesting that a test didn't fail, even something like that would be helpful.

nucypher/blockchain/eth/signers/software.py
```diff
@@ -105,8 +104,7 @@ def sign_transaction(self, transaction_dict: dict) -> HexBytes:
 
 
 class ClefSigner(Signer):
-
-    DEFAULT_IPC_PATH = '~/Library/Signer/clef.ipc' if sys.platform == 'darwin' else '~/.clef/clef.ipc'  #TODO: #1808
+    DEFAULT_IPC_PATH = '~/Library/Signer/clef.ipc' if sys.platform == 'darwin' else '~/.clef/clef.ipc'  # TODO: #1808
```
derekpierre, line 107: Any reason we didn't make this a Path as well? and L121?

tests/acceptance/cli/test_alice.py
```diff
@@ -114,28 +113,28 @@ def test_initialize_alice_with_custom_configuration_root(custom_filepath, click_
...
 
     # Auth
     assert COLLECT_NUCYPHER_PASSWORD in result.output, 'WARNING: User was not prompted for password'
     assert 'Repeat for confirmation:' in result.output, 'User was not prompted to confirm password'
 
 
 def test_alice_control_starts_with_preexisting_configuration(click_runner, custom_filepath):
-    custom_config_filepath = os.path.join(custom_filepath, AliceConfiguration.generate_filename())
+    custom_config_filepath = custom_filepath / AliceConfiguration.generate_filename()
     run_args = ('alice', 'run', '--dry-run', '--lonely', '--config-file', custom_config_filepath)
```
derekpierre, line 130: Am I missing something or do we need to update the value in args or else it would be a Path object...? Here and in other spots.

```suggestion
    run_args = ('alice', 'run', '--dry-run', '--lonely', '--config-file', str(custom_config_filepath.absolute()))
```

tests/unit/crypto/test_keystore.py
```diff
@@ -80,7 +80,7 @@ def test_invalid_keystore_file_type(tmp_path, tmp_path_factory):
 
     # Not an existing file
     invalid_path = Path('does-not-exist')
-    with pytest.raises(Keystore.NotFound, match=f"Keystore '{str(invalid_path)}' does not exist."):
+    with pytest.raises(Keystore.NotFound, match=f"Keystore '{(invalid_path.absolute())}' does not exist."):
```
derekpierre, line 83:
```suggestion
    with pytest.raises(Keystore.NotFound, match=f"Keystore '{invalid_path.absolute()}' does not exist."):
```

tests/acceptance/cli/test_alice.py
piotr-roslaniec: `Path` objects passed to CLI runner seem to be working. But I think it would be better to maintain the proposed convention of  passing `str` containing absolute paths instead.

nucypher/blockchain/eth/signers/software.py
piotr-roslaniec: Ah, I simply missed that.

nucypher/config/util.py
```diff
@@ -0,0 +1,33 @@
...
+from typing import Optional, get_type_hints
+
+
+def cast_paths_from(cls, payload):
+    constructor_args = get_type_hints(cls.__init__)
+    for ancestor in cls.__mro__:
+        constructor_args.update(get_type_hints(ancestor.__init__))
+    paths_only = [
+        arg for (arg, type_) in constructor_args.items()
+        if type_ == Path or type_ == Optional[Path]
```
derekpierre, line 28: Interesting strategy - cool! 😎

derekpierre (approved):
🎸

vepkenez (approved):
The Lord's work.

nucypher/blockchain/eth/multisig.py
```diff
@@ -93,7 +94,7 @@ def decode_transaction_data(self, contract: Contract = None, registry=None) -> T
         else:
             raise ValueError("This proposed TX doesn't have data")
 
-    def write(self, filepath: str = None) -> str:
+    def write(self, filepath: Path = None) -> Path:
```
fjarri, line 97: `Optional`?

nucypher/blockchain/eth/multisig.py
```diff
@@ -102,7 +103,7 @@ def write(self, filepath: str = None) -> str:
         return filepath
 
     @classmethod
-    def from_file(cls, filepath: str = None) -> 'Proposal':
+    def from_file(cls, filepath: Path = None) -> 'Proposal':
```
fjarri, line 106: `Optional`?

nucypher/cli/actions/configure.py
```diff
@@ -104,20 +106,20 @@ def destroy_configuration(emitter: StdoutEmitter,
 
 def handle_missing_configuration_file(character_config_class: Type[CharacterConfiguration],
                                       init_command_hint: str = None,
-                                      config_file: str = None) -> None:
+                                      config_file: Path = None) -> None:
```
fjarri, line 109: `Optional`

nucypher/cli/actions/select.py
```diff
@@ -208,9 +208,9 @@ def select_network(emitter: StdoutEmitter) -> str:
 
 def select_config_file(emitter: StdoutEmitter,
                        config_class: Type[CharacterConfiguration],
-                       config_root: str = None,
+                       config_root: Path = None,
```
fjarri, line 211: `Optional`

nucypher/cli/utils.py
```diff
@@ -130,21 +129,20 @@ def make_cli_character(character_config,
 
 def establish_deployer_registry(emitter,
                                 network: str = None,
-                                registry_infile: str = None,
-                                registry_outfile: str = None,
+                                registry_infile: Path = None,
```
fjarri, line 132: `Optional`

nucypher/cli/utils.py
```diff
@@ -167,15 +165,15 @@ def establish_deployer_registry(emitter,
     return registry
 
 
-def get_registry(network: str, registry_filepath: str = None) -> BaseContractRegistry:
+def get_registry(network: str, registry_filepath: Path = None) -> BaseContractRegistry:
```
fjarri, line 168: `Optional`

nucypher/config/base.py
```diff
@@ -133,8 +133,8 @@ class OldVersion(InvalidConfiguration):
         pass
 
     def __init__(self,
-                 config_root: str = None,
-                 filepath: str = None,
+                 config_root: Path = None,
```
fjarri, line 136: `Optional`

nucypher/config/base.py
```diff
@@ -188,17 +188,17 @@ def generate_filename(cls, modifier: str = None) -> str:
...
 
         :return: The generated filepath string
         """
         filename = cls.generate_filename()
-        default_path = os.path.join(config_root or cls.DEFAULT_CONFIG_ROOT, filename)
+        default_path = (config_root or cls.DEFAULT_CONFIG_ROOT) / filename
         return default_path
 
-    def generate_filepath(self, filepath: str = None, modifier: str = None, override: bool = False) -> str:
+    def generate_filepath(self, filepath: Path = None, modifier: str = None, override: bool = False) -> Path:
```
fjarri, line 201: `Optional`

nucypher/config/base.py
```diff
@@ -233,53 +233,62 @@ def _ensure_config_root_exists(self) -> None:
...
+    def peek(cls, filepath: Path, field: str) -> Union[str, None]:
         payload = cls._read_configuration_file(filepath=filepath)
         try:
             result = payload[field]
         except KeyError:
             raise cls.ConfigurationError(f"Cannot peek; No such configuration field '{field}', options are {list(payload.keys())}")
         return result
 
-    def to_configuration_file(self, filepath: str = None, modifier: str = None, override: bool = False) -> str:
+    def to_configuration_file(self, filepath: Path = None, modifier: str = None, override: bool = False) -> Path:
```
fjarri, line 251: `Optional`

nucypher/config/base.py
```diff
@@ -233,53 +233,62 @@ def _ensure_config_root_exists(self) -> None:
...
         filepath = self.generate_filepath(filepath=filepath, modifier=modifier, override=override)
         self._ensure_config_root_exists()
         filepath = self._write_configuration_file(filepath=filepath, override=override)
         return filepath
 
     @classmethod
-    def from_configuration_file(cls, filepath: str = None, **overrides) -> 'BaseConfiguration':
+    def from_configuration_file(cls, filepath: Path = None, **overrides) -> 'BaseConfiguration':
```
fjarri, line 258: `Optional`

nucypher/config/base.py
```diff
@@ -233,53 +233,62 @@ def _ensure_config_root_exists(self) -> None:
...
 
     def serialize(self, serializer=json.dumps) -> str:
         """Returns the JSON serialized output of `static_payload`"""
+        def _stringify_paths(d: dict):
+            for key, value in d.items():
+                if isinstance(value, Path):
+                    d[key] = str(value)
+                if isinstance(value, dict):
+                    d[key] = _stringify_paths(value)
+            return d
```
fjarri, line 288: May be excessive to return it seeing as how it's mutated inplace

nucypher/config/base.py
```diff
@@ -233,53 +233,62 @@ def _ensure_config_root_exists(self) -> None:
...
-        if os.path.exists(str(filepath)) and not override:
+        if filepath.exists() and not override:
             raise FileExistsError(f"{filepath} exists and no filename modifier supplied.")
         with open(filepath, 'w') as file:
             file.write(self.serialize())
         return filepath
 
     def serialize(self, serializer=json.dumps) -> str:
         """Returns the JSON serialized output of `static_payload`"""
+        def _stringify_paths(d: dict):
```
fjarri, line 282: In general it feels like it would be a good application for `marshmallow` (with `Path` being a field, so it will be serialized/deserialized appropriately), but perhaps it's not for this PR.

nucypher/config/base.py
```diff
@@ -615,20 +627,21 @@ def produce(self, **overrides) -> CHARACTER_CLASS:
         return character
 
     @classmethod
-    def assemble(cls, filepath: str = None, **overrides) -> dict:
+    def assemble(cls, filepath: Path = None, **overrides) -> dict:
```
fjarri, line 630: `Optional`

nucypher/config/base.py
```diff
@@ -722,7 +735,7 @@ def dynamic_payload(self) -> dict:
 
         return payload
 
-    def generate_filepath(self, filepath: str = None, modifier: str = None, override: bool = False) -> str:
+    def generate_filepath(self, filepath: Path = None, modifier: str = None, override: bool = False) -> Path:
```
fjarri, line 738: `Optional`

nucypher/config/characters.py
```diff
@@ -55,7 +55,8 @@ def __init__(self,
                  rest_host: str = None,
                  worker_address: str = None,
                  dev_mode: bool = False,
-                 db_filepath: str = None,
+                 db_filepath: Path = None,
```
fjarri, line 58: `Optional`

nucypher/config/characters.py
```diff
@@ -138,10 +139,22 @@ def callInThread(self, f, *args, **kwargs):
...
         super().destroy()
 
+    @classmethod
+    def deserialize(cls, payload: str, deserializer=json.loads, payload_label: Optional[str] = None) -> dict:
+        deserialized_payload = super().deserialize(payload, deserializer, payload_label)
+        deserialized_payload['db_filepath'] = Path(deserialized_payload['db_filepath'])
+        return deserialized_payload
+
+    @classmethod
+    def assemble(cls, filepath: Path = None, **overrides) -> dict:
```
fjarri, line 153: `Optional`

derekpierre:
@fjarri I sense a theme... 🤔 ... @piotr-roslaniec perhaps add `Optional[Path]` wherever we default a parameter of type Path to `None`.

fjarri:
Yeah, I'm just marking the occurrences. I mean, if you're modifying these lines anyway, why not fix the types :)

nucypher/config/characters.py
```diff
@@ -241,14 +254,13 @@ class FelixConfiguration(CharacterConfiguration):
...
     DEFAULT_DB_NAME = '{}.db'.format(NAME)
-    DEFAULT_DB_FILEPATH = os.path.join(DEFAULT_CONFIG_ROOT, DEFAULT_DB_NAME)
     DEFAULT_REST_PORT = 6151
     DEFAULT_LEARNER_PORT = 9151
     DEFAULT_REST_HOST = LOOPBACK_ADDRESS
     __DEFAULT_TLS_CURVE = ec.SECP384R1
 
     def __init__(self,
-                 db_filepath: str = None,
+                 db_filepath: Path = None,
```
fjarri, line 263: `Optional`

nucypher/config/characters.py
```diff
@@ -313,7 +325,7 @@ def _setup_node_storage(self, node_storage=None) -> None:
         pass
 
     @classmethod
-    def assemble(cls, filepath: str = None, **overrides) -> dict:
+    def assemble(cls, filepath: Path = None, **overrides) -> dict:
```
fjarri, line 328: `Optional`

nucypher/config/characters.py
```diff
@@ -322,19 +334,19 @@ def assemble(cls, filepath: str = None, **overrides) -> dict:
...
-    def generate_runtime_filepaths(cls, config_root: str) -> dict:
+    def generate_runtime_filepaths(cls, config_root: Path) -> dict:
         """Dynamically generate paths based on configuration root directory"""
         filepaths = dict(config_root=config_root,
-                         config_file_location=os.path.join(config_root, cls.generate_filename()))
+                         config_file_location=config_root / cls.generate_filename())
         return filepaths
 
-    def initialize(self, password: str = None) -> str:
+    def initialize(self, password: str = None) -> Path:
```
fjarri, line 343: `Optional`

derekpierre:
> Yeah, I'm just marking the occurrences. I mean, if you're modifying these lines anyway, why not fix the types :)

Oh I'm with you, but find and replace for "`: Path = None`" is relatively easy.

nucypher/config/util.py
```diff
@@ -0,0 +1,33 @@
...
+
+ You should have received a copy of the GNU Affero General Public License
+ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+"""
+
+from pathlib import Path
+from typing import Optional, get_type_hints
+
+
+def cast_paths_from(cls, payload):
```
fjarri, line 22: Some docstring would be useful. 

Relying on typehints being present and correct is kind of brittle. Personally, I'd go for
- corresponding classes accepting either a `Path` or a `str` and casting it; or
- using `marshmallow` to deserialized from JSON

nucypher/blockchain/eth/multisig.py
piotr-roslaniec: I think `filepath` shouldn't ever be `None` - otherwise it will throw at `open(filepath)`.

nucypher/blockchain/eth/multisig.py
piotr-roslaniec: Also here - `filepath=None` would throw at line 108

nucypher/config/characters.py
piotr-roslaniec: I think it always returns `Path`: if `self.dev==True` it assings new `self.config_root`, otherwise it throws inside `self._ensure_config_root_exists()`

nucypher/config/util.py
piotr-roslaniec: I would prefer the latter, as IMHO we would benefit from having the serialization format described in `marshmallow`. I can address this in a different PR.

nucypher/config/characters.py
derekpierre: Bogdan meant for the password parameter i.e. `password: Optional[str] = None`.

nucypher/blockchain/eth/multisig.py
```diff
@@ -18,7 +18,7 @@
 import json
 from collections import namedtuple
 from pathlib import Path
-from typing import Any, Dict, Tuple
+from typing import Any, Dict, Optional, Tuple
```
derekpierre, line 21: Is Optional actually used anywhere?

PR #2699, opened 2021-05-18
Remove IndisputableEvidence and related machinery

fjarri:
**Type of PR:**
- Other

**Required reviews:** 
- 3

**What this does:**
Removes `IndisputableEvidence`, utilities that were only used in it, and tests that check it.

**Why it's needed:**
Currently `IndisputableEvidence` can only lead to the slashing of an Ursula if she signs an incorrect cfrag and sends it to Bob. A minimally competent malicious Ursula won't expose herself that way, so it is not particularly useful.

Additionally, it is discarded after creation in the current codebase - it is not even sent out if the cfrag is invalid.

As discussed on a meeting, we remove `IndisputableEvidence` (to help with the adoption of the new Umbral), but keep the `Adjudicator` contract for now because of additional actions required to change the deployment process. But eventually it needs to be updated to do something useful. See #305.

**List of changes:**

- removed `IndisputableEvidence`
- removed `recover_pubkey_from_signature`
- removed `nucypher.types.Evidence`
- `Bob._reencrypt()` returns a list of offender Ursulas on failure instead of list of evidences
- removed slashing sections from `estimate_gas.py`
- `_mock_ursula_reencrypts()` returns a `PRETask` instead of `IndisputableEvidence`, and does not accept a `corrupt_cfrag` kwarg
- removed `test_slashing()` from `test_intercontract_integration.py` (note that I had to add the commitments from this tests to the next one, `test_withdraw()`). There's already a slashing test for `StakingEscrow`.
- removed `get_coordinates_as_bytes()` and simplified `canonical_address_from_umbral_key()`
- moved `get_signature_recovery_value()` to the only test where it is used.
- removed tests:
  - `tests/acceptance/blockchain/actors/test_investigator.py`
  - `tests/acceptance/blockchain/agents/test_adjudicator_agent.py`
  - `tests/contracts/lib/test_reencryption_validator.py`
  - `tests/contracts/main/adjudicator` (whole folder)
  - `tests/unit/crypto/test_coordinates_serialization.py`
  - `tests/unit/crypto/test_umbral_signatures.py`

Note that I kept `AdjudicatorAgent` in place for now, because it seems to be used in deployment.


newsfragments/2699.misc.rst
```diff
@@ -0,0 +1 @@
+Remove IndisputableEvidence
```
KPrasch, line 1: I think there is a removal newgragment category

KPrasch (approved):
Looks like the changes we discussed in the call.  

Let's revisit our approach to slashing soon 🙏🏻

newsfragments/2699.misc.rst
fjarri: That's right, it's not mentioned in README, but it is allowed by the validator. Fixed.

cygnusv (approved):
Cleanly done. Thanks @fjarri !

derekpierre (approved):
🎸

PR #2700, opened 2021-05-18
[WIP] Update Ursulas certificate

### LINKED ISSUE #2674: How does an Ursula update their TLS certificate if eg. the IP address changes
Conversation from discrod - https://discord.com/channels/411401661714792449/497481850382450710/836957333354840064

A node's IP address changed, therefore, making the existing TLS cert invalid. This node is now regarded as unverified by the rest of the network because the IP in the cert does not match the node's IP address.

Related to #2536, which had outstanding questions about cert renewal:
```
May introduce new side-effects: When do certificates need to be recreated or renewed?
IP address changes
Certificate Expiration
How/when to force re-validation and download a fresh copy of a certificate from a peer
```

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
- Fixes #2674 

**Why it's needed:**

**Notes for reviewers:**
- This PR will be heavily rewritten

### DIFF
--- a/nucypher/network/middleware.py
+++ b/nucypher/network/middleware.py
@@ -14,7 +14,7 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
-
+import urllib
 
 import requests
 import socket
@@ -34,6 +34,8 @@ EXEMPT_FROM_VERIFICATION.bool_value(False)
 
 
 class NucypherMiddlewareClient:
+    log = Logger()
+
     library = requests
     timeout = 1.2
 
@@ -54,6 +56,7 @@ class NucypherMiddlewareClient:
         if node_or_sprout:
             if node_or_sprout is not EXEMPT_FROM_VERIFICATION:
                 node = node_or_sprout.mature()  # Morph into a node.
+                # TODO: Why do we ignore verification results here? Are we only interested in raised exceptions?
                 node.verify_node(network_middleware_client=self, registry=self.registry)
         return self.parse_node_or_host_and_port(node_or_sprout, host, port)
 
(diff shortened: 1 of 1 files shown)


nucypher/network/middleware.py
```diff
@@ -115,6 +118,16 @@ def method_wrapper(path,
...
 
+            if node_or_sprout:
+                parsed = urllib.parse.urlparse(host)
+                try:
+                    certificate = ssl.get_server_certificate(addr=(parsed.hostname, parsed.port))
+                    certificate_filepath = node_or_sprout.node_storage.store_node_certificate(certificate)
+                    self.log.warn(f"Replaced SSL certificate for host: {host}")
+                except:
+                    self.log.warn(f"Failed to replace SSL certificate for host: {host}")
+                    pass
```
KPrasch, line 129: What exceptions are you anticipating here?  Do you think it will be okay to suppress *all* of them?

piotr-roslaniec: Just to clarify, this PR is not in its intended state, and the exceptions should be handled on a case-to-case basis. I'm probably going to take a different approach in solving this issue.

KPrasch:
This PR is on my mind today - I think adopting the principal of "always replace" the TLS certificate is the most simple and elegant way to ensure that the cert always matches the latest remote node's runtime, regardless of state changes, restarts, rederivation of keys, etc.

derekpierre:
> This PR is on my mind today - I think adopting the principal of "always replace" the TLS certificate is the most simple and elegant way to ensure that the cert always matches the latest remote node's runtime, regardless of state changes, restarts, rederivation of keys, etc.

@KPrasch , @piotr-roslaniec any thoughts on https://github.com/nucypher/nucypher/issues/2674#issuecomment-836718139? It isn't an "always replace" solution but a "check and replace" solution.

piotr-roslaniec:
An attempt to sum up our discussion:
1) On server-side - Check and replace your own certificate during startup
2)  On client-side - Check and replace server certificate on every connection

I think these two combined cover all relevant cases. 

On a different note, should "1)" be enabled by default? Or should the node just fail on startup? In case someone uses a regular certificate (not self-signed).

derekpierre:
>  Or should the node just fail on startup?
If we do that, would we basically have to provide some tool for users to generate their own cert? Feels like it, since the situation would need to be resolved somehow.

> In case someone uses a regular certificate (not self-signed).
Possible but seems highly unlikely to me since I think (?) they would have to muck with the files `nucypher` produced and possibly the code...

PR #2709, opened 2021-05-20
Provide required info for nucypher-monitor without unnecessarily maturing sprouts

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

**What this does:**
Removes a `mature()` call that was introduced in https://github.com/nucypher/nucypher/pull/2574/commits/50da675692fba33ab76ece5c4a5da86fd7549ce6 for the purposes of `nucypher-monitor`. The only thing it needs that a sprout doesn't have is `rest_url`, and we can provide that.

Additional changes were necessary to allow some tests to pass, since they now got a mix of `Ursula` and `NodeSprout` objects:
- `NodeSprout.__eq__` added (identical to `Character.__eq__`, comparing stamps)
- `NodeSprout.__hash__` was changed to conform to `__eq__` (and to `Character.__hash__`), that is taking the hash of a stamp.

### DIFF
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -269,7 +269,7 @@ class FleetSensor:
             # Replace the existing object with a newer object, even if they're equal
             # (this object can be mutated externally).
             # This behavior is supposed to be consistent with that of the node storage
-            # (where a newer obhect with the same `checksum_address` replaces an older one).
+            # (where a newer object with the same `checksum_address` replaces an older one).
             if node in self._nodes_to_add:
                 self._nodes_to_add.remove(node)
             self._nodes_to_add.add(node)
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -150,7 +150,6 @@ class FleetState:
             nodes_to_add_dict = {node.checksum_address: node for node in nodes_to_add}
             for checksum_address in diff.nodes_updated:
                 new_node = nodes_to_add_dict[checksum_address]
-                new_node.mature()
                 nodes[checksum_address] = new_node
             for checksum_address in diff.nodes_removed:
                 del nodes[checksum_address]
--- a/nucypher/acumen/perception.py
+++ b/nucypher/acumen/perception.py
@@ -266,6 +266,12 @@ class FleetSensor:
     def record_node(self, node: 'Ursula'):
 
         if node.domain == self._domain:
+            # Replace the existing object with a newer object, even if they're equal
+            # (this object can be mutated externally).
+            # This behavior is supposed to be consistent with that of the node storage
+            # (where a newer obhect with the same `checksum_address` replaces an older one).
+            if node in self._nodes_to_add:
+                self._nodes_to_add.remove(node)
             self._nodes_to_add.add(node)
 
             if self._auto_update_state:
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -133,6 +133,9 @@ class NodeSprout(PartiallyKwargifiedBytes):
             self._nickname = Nickname.from_seed(self.checksum_address)
         return self._nickname
 
+    def rest_url(self):
+        return self.rest_interface.uri
+
     def mature(self):
         if self._is_finishing:
             return self._finishing_mutex.get()
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -93,7 +93,11 @@ class NodeSprout(PartiallyKwargifiedBytes):
         self._finishing_mutex = Queue()
 
     def __eq__(self, other):
-        return isinstance(other, NodeSprout) and self._checksum_address == other._checksum_address
+        try:
+            other_stamp = other.stamp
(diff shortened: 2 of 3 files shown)


KPrasch (dismissed):

KPrasch:
Rerunning CI to see if the failure is intermittent.

fjarri:
Some results from the failure investigation.

`test_one_node_stores_a_bunch_of_others`: started at commit 37929b36a88b373356296759d7b23b1aa86f53e5

In `Learner.load_seednodes()`, a sprout corresponding to the seed node from the test was created twice: once as `maybe_sage_node`, and the second time as `seed_node` (in the iteration over `self._seed_nodes`). Since `NodeSprout` does not have `__eq__`, these objects were treated as different ones by `FleetSensor` (`record_node()` puts new nodes in a set `_nodes_to_add`). Then, since sets are unordered, either of the two objects could end up in the actual fleet state during `record_fleet_state()`.

On the other hand, `node_storage.store_node_metadata()` call had a different logic - since it did not accumulate changes and instead just saved the given one in the dictionary, the newer object always overwrote the older one. 

Therefore, during the assertion in the test, where `known_nodes` is compared with the `node_storage`, two things could happen:
- The seednode sprout never had `mature()` called on it. In that case the lists would differ, since two different sprout objects with the same checksum addresses were not considered equal.
- The seednode sprout (the newer one) had `mature()` called on it. In that case, one would see an `Ursula` object in one list, and a `NodeSprout` in the other, and the test would fail just the same.

Solution to this may vary depending on what semantics exactly we want. Currently, I committed the following solution:
- add `NodeSprout.__eq__()` (two sprouts are equal if their checksum addresses are equal)
- force the replacement behavior of the older object with the newer one in `FleetSensor.record_node()`, to match that of the node storage.

This seems to fix the issue. Although perhaps it is not the best solution, since it does not fix the similarly looking failure in `test_learner_uses_both_nodes_from_storage_and_fallback_nodes`.

To be continued.

fjarri:
More investigation.

The introduction of `NodeSprout.__eq__` in this PR broke logic in `remember_node()`. At the start there, we're comparing `node == self`, and if `True`, we don't remember the node. This worked because if `node` was a `NodeSprout`, it didn't have `__eq__`, and the `Character.__eq__` (of `self`) was called, which only needed `other.stamp`. The initial version of `NodeSprout.__eq__` I added only compared sprouts to sprouts, so it allowed node to remember itself in `FleetSensor`, causing test failures.

Additionally, `set`s of mixed `Ursula` and `NodeSprout` objects could not be compared in tests because even when a sprout and a full Ursula were equal, their `hash()` was different - sprouts used `public_address`, and Ursulas used `stamp`. Since the new `NodeSprout.__eq__` uses `stamp`, it is logical to change `NodeSprout.__hash__` to use it as well, because the assumption is that if two objects are equal, their hashes should be equal too.

fjarri:
The `cli` failure seems to be present in `main` as well

fjarri:
> The cli failure seems to be present in main as well

Ah, nevermind, it's the probationary period ending again.

nucypher/network/nodes.py
```diff
@@ -133,6 +137,9 @@ def nickname(self):
             self._nickname = Nickname.from_seed(self.checksum_address)
         return self._nickname
 
+    def rest_url(self):
+        return self.rest_interface.uri
```
derekpierre, line 141: This was already being stored, but just not previously made accessible?

fjarri: Yes, before this method was only present for an Ursula, so it required a `mature()` call at some point. So in #2574 the call that was now being removed was added. But as it seems, the required data is already present in a sprout, so we can just expose it.

That said, I still think there we need to review this approach and make the maturation event less random, to provide some guarantees for the users of `FleetSensor`. This will make testing easier, too.

derekpierre: > That said, I still think there we need to review this approach and make the maturation event less random, to provide some guarantees for the users of FleetSensor. This will make testing easier, too.

👍 - perhaps we can explore as a follow-up issue.

derekpierre (approved):
🎸

nucypher/network/nodes.py
fjarri: Filed #2715 - it's more specific, but changes to how `mature()` is called will probably be a part of the solution.

nucypher/acumen/perception.py
```diff
@@ -267,6 +266,12 @@ def __init__(self, domain: str, this_node: Optional['Ursula'] = None):
     def record_node(self, node: 'Ursula'):
 
         if node.domain == self._domain:
+            # Replace the existing object with a newer object, even if they're equal
+            # (this object can be mutated externally).
+            # This behavior is supposed to be consistent with that of the node storage
+            # (where a newer obhect with the same `checksum_address` replaces an older one).
```
KPrasch, line 272:
```suggestion
            # (where a newer object with the same `checksum_address` replaces an older one).
```

KPrasch (approved):

Issue #2704, opened 2021-05-20
Implement Porter functionality for Bob


derekpierre:
REST endpoints, associated Porter Learner functions, and think about failure HTTP codes.

Draft API provided by #2667 . Can re-iterate here if required, but I'll just summarize - see #2667 for full context.
For Bob::retrieve:
* `/get_treasure_map` - https://github.com/nucypher/nucypher/issues/2667#issuecomment-842416581
* `/exec_work_order` - https://github.com/nucypher/nucypher/issues/2667#issuecomment-842416581

derekpierre:
Closed by #2729 .

Issue #2703, opened 2021-05-20
Implement Porter functionality for Alice


derekpierre:
REST endpoints, associated Porter Learner functions, and think about failure HTTP codes.

Draft API provided by #2666 . Can re-iterate here if required, but I'll just summarize - see #2666 for full context.
For Alice::grant:
* `/get_ursulas` - https://github.com/nucypher/nucypher/issues/2666#issuecomment-845169535
* `/publish_treasure_map` - https://github.com/nucypher/nucypher/issues/2666#issuecomment-843383687

For Alice::revoke
* `/revoke` - https://github.com/nucypher/nucypher/issues/2666#issuecomment-844425997

derekpierre:
`/publish_treasure_map` will return:

```json
{
   "result": {
      "published": True
   },
   "version": "5.2.0"
}
```

derekpierre:
`/get_ursulas` and `/publish_treasure_map` are both already implemented. 

`/revoke` functionality is currently not implemented and TBD because of the work done for TMapConKFrags (see https://github.com/nucypher/nucypher/blob/main/nucypher/network/server.py#L278). Consequently revoke work will be completed separately (outside of the MVP Milestone) via #2749

PR #2717, opened 2021-06-02
POC Porter implementation for `/get_treasure_map`, `/publish_treasure_map`, and `/get_ursulas` (unaffected by TMapConKFrags work)

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**Issues fixed/closed:**
Related to #2703, #2704.

**Why it's needed:**
Simple POC implementation for Porter functions `/get_treasure_map` (Bob), `/publish_treasure_map` (Alice), and `/get_ursulas` (Alice) that are unaffected by the work to be completed for TMapConKFrags. 

Outstanding questions for reviewers:
1. What do we do about timeouts (default timeouts) for the Porter Learner functionality? Used existing default values in original code.
2. Is this the best way of doing connectivity checks with sampled Ursulas - <https://github.com/derekpierre/nucypher/blob/porter-impl/nucypher/utilities/porter/porter.py#L142>
3. How come `NotStaking` wasn't caught before when trying to get a treasure map from nodes? I added it to the refactored common code here - <https://github.com/derekpierre/nucypher/blob/porter-impl/nucypher/characters/utils.py#L56>
4. TreasureMapID validation only checks length; 32 bytes for federated, 16 for decentralized - is this correct/sufficient? <https://github.com/derekpierre/nucypher/blob/porter-impl/nucypher/utilities/porter/control/specifications/fields/treasuremapid.py#L28>
5. TreasureMap specification validation now uses TreasureMap.splitter() and SignedTreasureMap.splitter() - it can't tell whether the treasure map bytes provided are federated or not, so the validation uses both splitters to check and if both fail, combines the two exceptions into one. Is there a way to define one splitter that encompasses both types of expected treasure map bytes? See https://github.com/derekpierre/nucypher/blob/porter-impl/nucypher/characters/control/specifications/fields/treasuremap.py#L37. Porter is really meant for decentralized mode but having it handle federated mode is great for testing.


derekpierre:
Only tested in federated setting so far.

1. Run `$ python examples/run_demo_ursula_fleet.py`
2. Run Porter `$ nucypher porter run --federated-only --teacher localhost:11500`
3. Run the following in ipython shell
```python
In [1]: import requests

In [2]: payload = {"bob_encrypting_key": "026d1f4ce5b2474e0dae499d6737a8d987ed3c9ab1a55e00f57ad2d8e81fe9e9ac", "treasure_map_id": "f6ec73c93084ce91d5542a4ba6070071f5565112fe19b26ae9c960f9d658903a"}

In [3]: r = requests.get("http://localhost:9155/get_treasure_map", json=payload)

# no treasure map available since non published as yet

In [6]: r
Out[6]: <Response [500]>

# publish treasure map

In [7]: payload_publish = {"bob_encrypting_key": "026d1f4ce5b2474e0dae499d6737a8d987ed3c9ab1a55e00f57ad2d8e81fe9e9ac", "treasure_map": "Qld7S8sbKFCv2B8KxfJo4oxiTOjZ4VPyqTK5K1xK6DND6TbLg2hvlGaMV69aiiC5QfadB82w/5q1Sw+SNFHN2esWgAbs38QuUVUGCzDoWzQAAAGIAuhw12ZiPMNV8LaeWV8uUN+au2HGOjWilqtKsaP9fmnLAzFiTUAu9/VCxOLOQE88BPoWk1H7OxRLDEhnBVYyflpifKbOYItwLLTtWYVFRY90LtNSAzS8d3vNH4c3SHSZwYsCKY+5LvJ68GD0CqhydSxCcGckh0unttHrYGSOQsURUI4AAAEBsSMlukjA1WyYA+FouqkuRtk8bVHcYLqRUkK2n6dShEUGMuY1SzcAbBINvJYmQp+hhzK5m47AzCl463emXepYZQC/evytktG7yXxd3k8Ak+Qr7T4+G2VgJl4YrafTpIT6wowd+8u/SMSrrf/M41OhtLeBC4uDKjO3rYBQfVLTpEAgiX/9jxB80RtNMeCwgcieviAR5tlw2IlxVTEhxXbFeopcOZmfEuhVWqgBUfIakqsNCXkkubV0XS2l5G1vtTM8oNML0rP8PyKd4+0M5N6P/EQqFkHH93LCDD0IQBq9usm3MoJp0eT8N3m5gprI05drDh2xe/W6qnQfw3YXnjdvf2A="}

In [8]: r_pub = requests.post("http://localhost:9155/publish_treasure_map", json=payload_publish)

In [9]: r_pub
Out[9]: <Response [200]>

In [10]: r_pub.json()
Out[10]: {'result': {}, 'version': '5.2.0'}

# now that it is published try getting it again

In [11]: r = requests.get("http://localhost:9155/get_treasure_map", json=payload)

In [12]: r
Out[12]: <Response [200]>

In [13]: r.json()
Out[13]:
{'result': {'treasure_map': 'Qld7S8sbKFCv2B8KxfJo4oxiTOjZ4VPyqTK5K1xK6DND6TbLg2hvlGaMV69aiiC5QfadB82w/5q1Sw+SNFHN2esWgAbs38QuUVUGCzDoWzQAAAGIAuhw12ZiPMNV8LaeWV8uUN+au2HGOjWilqtKsaP9fmnLAzFiTUAu9/VCxOLOQE88BPoWk1H7OxRLDEhnBVYyflpifKbOYItwLLTtWYVFRY90LtNSAzS8d3vNH4c3SHSZwYsCKY+5LvJ68GD0CqhydSxCcGckh0unttHrYGSOQsURUI4AAAEBsSMlukjA1WyYA+FouqkuRtk8bVHcYLqRUkK2n6dShEUGMuY1SzcAbBINvJYmQp+hhzK5m47AzCl463emXepYZQC/evytktG7yXxd3k8Ak+Qr7T4+G2VgJl4YrafTpIT6wowd+8u/SMSrrf/M41OhtLeBC4uDKjO3rYBQfVLTpEAgiX/9jxB80RtNMeCwgcieviAR5tlw2IlxVTEhxXbFeopcOZmfEuhVWqgBUfIakqsNCXkkubV0XS2l5G1vtTM8oNML0rP8PyKd4+0M5N6P/EQqFkHH93LCDD0IQBq9usm3MoJp0eT8N3m5gprI05drDh2xe/W6qnQfw3YXnjdvf2A='},
 'version': '5.2.0'}
```

derekpierre:
For `/get_ursulas`:

NOTE: the list of Ursulas returned will vary every time the following is run so other values need to be modified eg. include/exclude lists.

Federated Example
-------------------
1. Run `$ python examples/run_demo_ursula_fleet.py`
2. Run Porter `$ nucypher porter run --federated-only --teacher localhost:11500`
3. Run the following in python shell:

```python
In [1]: import requests

In [2]: payload = {"quantity": 5, "duration_periods": 4}

In [3]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [4]: r
Out[4]: <Response [200]>

In [5]: r.content
Out[5]: b'{"result": {"ursulas": [{"encrypting_key": "03440a4c53bd8a215334694ff9e262bff2bdd015afb9b00e3bb60341b220166198", "checksum_address": "0xee94Bdb476510F447D2697A821E159CbcC7Db3d9", "ip_address": "https://127.0.0.1:11507"}, {"encrypting_key": "0341b821ad2e7fc814ccecc770d66ac67fc116f8ca01f0a077010498e408b485db", "checksum_address": "0x39Dbe2270780aaB9d20a8E58f62d441851AB96cD", "ip_address": "https://127.0.0.1:11511"}, {"encrypting_key": "026d9a993e37370d135867bd8c8108e7533def19e70ff26b7127cf2ad8bf3c5674", "checksum_address": "0x519A5EF6c5f66527d6888caf1221f52d64eD7C34", "ip_address": "https://127.0.0.1:11504"}, {"encrypting_key": "026ddd7ed9f8b7e649ad2f0a994894332688421cd64bb7eabdc506ec1ead48aa71", "checksum_address": "0x226f652aE5BD2A07097085729b7109EF62D16388", "ip_address": "https://127.0.0.1:11509"}, {"encrypting_key": "03add66a40607f44525aeed31e9c8979601704e3aba23d1e6378d89dffc5262eca", "checksum_address": "0x2b49d8a024CEc7e5E4D1fB953624ADEa9308ca9F", "ip_address": "https://127.0.0.1:11500"}]}, "version": "5.2.0"}'

In [6]: include_ursulas = ["0x7b5D8E3ADf81F53c815B731295E4677D28137935", "0x519A5EF6c5f66527d6888caf1221f52d64eD7C34"]

In [7]: payload["include_ursulas"] = include_ursulas

In [8]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [9]: r
Out[9]: <Response [200]>

In [10]: r.content
Out[10]: b'{"result": {"ursulas": [{"encrypting_key": "02f1999d954d88745ce9f4f9b8e7348e0a8ccda05711d14e5dc206bd272bdf0493", "checksum_address": "0x64FaCbb9AD35f1E5f6aAd2Dc5f626f3A38E2247f", "ip_address": "https://127.0.0.1:11501"}, {"encrypting_key": "021bad03835961be222e7d6b9ea161a2ba902545e8a796528308dad19f073dbf45", "checksum_address": "0x41f5f375521259E8377775DEEE045E5F6eA3fC35", "ip_address": "https://127.0.0.1:11510"}, {"encrypting_key": "026d9a993e37370d135867bd8c8108e7533def19e70ff26b7127cf2ad8bf3c5674", "checksum_address": "0x519A5EF6c5f66527d6888caf1221f52d64eD7C34", "ip_address": "https://127.0.0.1:11504"}, {"encrypting_key": "030fa280e84ff8c2e10b63bc32d1f4bb487beb98a8e8949a756f254c97f2052982", "checksum_address": "0x7b5D8E3ADf81F53c815B731295E4677D28137935", "ip_address": "https://127.0.0.1:11506"}, {"encrypting_key": "03ee7b66057b75f97f8b7ea51d0c2bfb5ab3838ef66a3c1c4dfe4014a0dd9d59a0", "checksum_address": "0x1f31F8F2988f1313c4B36CE211E708c946c66fee", "ip_address": "https://127.0.0.1:11505"}]}, "version": "5.2.0"}'

In [11]: exclude_ursulas = ["0x1f31F8F2988f1313c4B36CE211E708c946c66fee", "0x64FaCbb9AD35f1E5f6aAd2Dc5f626f3A38E2247f", "0x8fEE90e395a47cfB3f4e8ce96306337B701A6ca8", "0x39Dbe2270780aaB9d20a8E58f62d441851AB96cD"]

In [12]: payload["exclude_ursulas"] = exclude_ursulas

In [13]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [14]: r
Out[14]: <Response [200]>

In [15]: r.content
Out[15]: b'{"result": {"ursulas": [{"encrypting_key": "026d9a993e37370d135867bd8c8108e7533def19e70ff26b7127cf2ad8bf3c5674", "checksum_address": "0x519A5EF6c5f66527d6888caf1221f52d64eD7C34", "ip_address": "https://127.0.0.1:11504"}, {"encrypting_key": "0350d0ec52ffa63691d4a71748b0e5a6fe7c8aa751bedb8873c52ad43684b51fa0", "checksum_address": "0x538D64a582C8a3A6A2b8480c5e4c231b85B747f9", "ip_address": "https://127.0.0.1:11508"}, {"encrypting_key": "030fa280e84ff8c2e10b63bc32d1f4bb487beb98a8e8949a756f254c97f2052982", "checksum_address": "0x7b5D8E3ADf81F53c815B731295E4677D28137935", "ip_address": "https://127.0.0.1:11506"}, {"encrypting_key": "020faf809b76ef4ee45ced7929f0d7697aed958e39327d31ceb7275dbe9307e751", "checksum_address": "0x14c1c069216e2b1Fd7aB41f7e21a6d9dc9D631Bc", "ip_address": "https://127.0.0.1:11503"}, {"encrypting_key": "026ddd7ed9f8b7e649ad2f0a994894332688421cd64bb7eabdc506ec1ead48aa71", "checksum_address": "0x226f652aE5BD2A07097085729b7109EF62D16388", "ip_address": "https://127.0.0.1:11509"}]}, "version": "5.2.0"}'

In [16]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [17]: r
Out[17]: <Response [200]>

In [18]: r.content
Out[18]: b'{"result": {"ursulas": [{"encrypting_key": "0350d0ec52ffa63691d4a71748b0e5a6fe7c8aa751bedb8873c52ad43684b51fa0", "checksum_address": "0x538D64a582C8a3A6A2b8480c5e4c231b85B747f9", "ip_address": "https://127.0.0.1:11508"}, {"encrypting_key": "026ddd7ed9f8b7e649ad2f0a994894332688421cd64bb7eabdc506ec1ead48aa71", "checksum_address": "0x226f652aE5BD2A07097085729b7109EF62D16388", "ip_address": "https://127.0.0.1:11509"}, {"encrypting_key": "030fa280e84ff8c2e10b63bc32d1f4bb487beb98a8e8949a756f254c97f2052982", "checksum_address": "0x7b5D8E3ADf81F53c815B731295E4677D28137935", "ip_address": "https://127.0.0.1:11506"}, {"encrypting_key": "026d9a993e37370d135867bd8c8108e7533def19e70ff26b7127cf2ad8bf3c5674", "checksum_address": "0x519A5EF6c5f66527d6888caf1221f52d64eD7C34", "ip_address": "https://127.0.0.1:11504"}, {"encrypting_key": "020faf809b76ef4ee45ced7929f0d7697aed958e39327d31ceb7275dbe9307e751", "checksum_address": "0x14c1c069216e2b1Fd7aB41f7e21a6d9dc9D631Bc", "ip_address": "https://127.0.0.1:11503"}]}, "version": "5.2.0"}'
```

Decentralized Example
-----------------------
1. Run Porter `$ nucypher porter run --provider https://mainnet.infura.io/v3/<APP_ID> --network mainnet`

```python
In [1]: import requests

In [2]: payload = {"quantity": 5, "duration_periods": 4}

In [3]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [4]: r
Out[4]: <Response [200]>

In [5]: r.content
Out[5]: b'{"result": {"ursulas": [{"encrypting_key": "025a335eca37edce8191d43c156e7bc6b451b21e5258759966bbfe0e6ce44543cb", "checksum_address": "0x5cF1703A1c99A4b42Eb056535840e93118177232", "ip_address": "https://3.236.144.36:9151"}, {"encrypting_key": "02b0a0099ee180b531b4937bd7446972296447b2479ca6259cb6357ed98b90da3a", "checksum_address": "0x7fff551249D223f723557a96a0e1a469C79cC934", "ip_address": "https://54.218.83.166:9151"}, {"encrypting_key": "02761c765e2f101df39a5f680f3943d0d993ef9576de8a3e0e5fbc040d6f8c15a5", "checksum_address": "0x9C7C824239D3159327024459Ad69bB215859Bd25", "ip_address": "https://92.53.84.156:9151"}, {"encrypting_key": "0258b7c79fe73f3499de91dd5a5341387184035d0555b10e6ac762d211a39684c0", "checksum_address": "0x9919C9f5CbBAA42CB3bEA153E14E16F85fEA5b5D", "ip_address": "https://3.36.66.164:9151"}, {"encrypting_key": "02e43a623c24db4f62565f82b6081044c1968277edfdca494a81c8fd0826e0adf6", "checksum_address": "0xfBeb3368735B3F0A65d1F1E02bf1d188bb5F5BE6", "ip_address": "https://128.199.124.254:9151"}]}, "version": "5.2.0"}'

In [6]: include_ursulas = ["0x5cF1703A1c99A4b42Eb056535840e93118177232", "0x7fff551249D223f723557a96a0e1a469C79cC934"]

In [7]: payload["include_ursulas"] = include_ursulas

In [8]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [9]: r
Out[9]: <Response [200]>

In [10]: r.content
Out[10]: b'{"result": {"ursulas": [{"encrypting_key": "030b29c5bbcba278a962e76cbb0f404b74ef8c50892dd33664a24c8b079369fa41", "checksum_address": "0xCef1B4Bf8F05F623A2A688b56d9dA679D302EBa7", "ip_address": "https://54.39.181.135:9151"}, {"encrypting_key": "039397a359f537a34e72859c55563ea224433ecfb02bf030da46485197868e66fc", "checksum_address": "0x02D4Fe8f98129F578fb2e160404033B8e4627A36", "ip_address": "https://54.39.190.30:9151"}, {"encrypting_key": "025a335eca37edce8191d43c156e7bc6b451b21e5258759966bbfe0e6ce44543cb", "checksum_address": "0x5cF1703A1c99A4b42Eb056535840e93118177232", "ip_address": "https://3.236.144.36:9151"}, {"encrypting_key": "02b0a0099ee180b531b4937bd7446972296447b2479ca6259cb6357ed98b90da3a", "checksum_address": "0x7fff551249D223f723557a96a0e1a469C79cC934", "ip_address": "https://54.218.83.166:9151"}, {"encrypting_key": "0389b7552f5724bd831cb7dc3d2e8ba8389575b52d544d2b091a8a65dbeffa287e", "checksum_address": "0xB04FcDF9327f65AB0107Ea95b78BB200C07FA752", "ip_address": "https://18.210.24.63:9151"}]}, "version": "5.2.0"}'

In [11]: exclude_ursulas = ["0xB04FcDF9327f65AB0107Ea95b78BB200C07FA752", "0xCef1B4Bf8F05F623A2A688b56d9dA679D302EBa7", "0x02D4Fe8f98129F578fb2e160404033B8e4627A36", "0x9C7C824239D3159327024459Ad69bB215859Bd25", "0x9919C9f5CbBAA42CB3bEA153E14E16F85fEA5b5D", "0xfBeb3368735B3F0A65d1F1E02bf1d188bb5F5BE6"]

In [12]: payload['exclude_ursulas'] = exclude_ursulas

In [13]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [14]: r
Out[14]: <Response [200]>

In [15]: r.content
Out[15]: b'{"result": {"ursulas": [{"encrypting_key": "025a335eca37edce8191d43c156e7bc6b451b21e5258759966bbfe0e6ce44543cb", "checksum_address": "0x5cF1703A1c99A4b42Eb056535840e93118177232", "ip_address": "https://3.236.144.36:9151"}, {"encrypting_key": "02b0a0099ee180b531b4937bd7446972296447b2479ca6259cb6357ed98b90da3a", "checksum_address": "0x7fff551249D223f723557a96a0e1a469C79cC934", "ip_address": "https://54.218.83.166:9151"}, {"encrypting_key": "030b41e643823db37ca98f69362e8665dc2b7dbfa3aa412993665795afcecc2d9b", "checksum_address": "0xbdA6Ba3Ad624582495C55451A1840e563fc16389", "ip_address": "https://34.209.206.31:9151"}, {"encrypting_key": "029efc301b857e184da14fcb29b55d8a38bf28d906763c811cc641b4c916fd423b", "checksum_address": "0x93EC08943257DC1abeFb46875C7AAF92aeE0ecC6", "ip_address": "https://44.235.74.191:9151"}, {"encrypting_key": "02921281b5b0ca222a9bf5e5c6d6525381677b7ed429e2deab79eb245b6461a472", "checksum_address": "0x42D2875a71Cb25DDcF05Dbed55490cB905498Bac", "ip_address": "https://13.212.162.22:9151"}]}, "version": "5.2.0"}'

In [16]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [17]: r
Out[17]: <Response [200]>

In [18]: r.content
Out[18]: b'{"result": {"ursulas": [{"encrypting_key": "025a335eca37edce8191d43c156e7bc6b451b21e5258759966bbfe0e6ce44543cb", "checksum_address": "0x5cF1703A1c99A4b42Eb056535840e93118177232", "ip_address": "https://3.236.144.36:9151"}, {"encrypting_key": "02f80e41b8269c4a57a8b8dfb3787746fcea620883e5a0a09b2c8c89cae4488bc1", "checksum_address": "0xFb626333099A91Ab677BCd5e9C71bc4Dbe0238a8", "ip_address": "https://34.209.70.116:9151"}, {"encrypting_key": "02b0a0099ee180b531b4937bd7446972296447b2479ca6259cb6357ed98b90da3a", "checksum_address": "0x7fff551249D223f723557a96a0e1a469C79cC934", "ip_address": "https://54.218.83.166:9151"}, {"encrypting_key": "029efc301b857e184da14fcb29b55d8a38bf28d906763c811cc641b4c916fd423b", "checksum_address": "0x93EC08943257DC1abeFb46875C7AAF92aeE0ecC6", "ip_address": "https://44.235.74.191:9151"}, {"encrypting_key": "027bb2220dfd1f9d02c0ccf973678b9d811806daf0e6e4439c253854746c393c77", "checksum_address": "0xb011b6794F271B1d8D2328eDf332535343CB4D44", "ip_address": "https://138.197.186.215:9151"}]}, "version": "5.2.0"}'
```

nucypher/utilities/porter/control/specifications/porter_schema.py
```diff
@@ -102,6 +104,15 @@ def check_valid_quantity_and_include_ursulas(self, data, **kwargs):
         if ursulas_to_include and len(ursulas_to_include) > data['quantity']:
```
piotr-roslaniec, line 104: Just my 2 cents - Maybe we could clarify a bit what `quantity` means here:
a) "Give me a `quantity` of Ursulas, and also include Ursulas from  the `include_ursulas`  list", for a total of `quantity` + `len(include_ursulas)` Ursulas
b) "Give me a maximum of `quantity` of Ursulas, with a preference for Ursulas in `include_ursulas`",  for a total of `quantity` Ursulas
If it's "b)" maybe we could go for `max_quantity` instead of `quantity`

nucypher/utilities/porter/porter.py
```diff
@@ -49,15 +65,37 @@ class Porter(Learner):
...
+                 provider_uri: str = None,
                  *args, **kwargs):
-        self.federated_only = False  # TODO Start with non-federated for now
-        self.registry = registry or InMemoryContractRegistry.from_latest_publication(network=domain)
+        self.federated_only = federated_only
+
+        if not self.federated_only:
+            if not provider_uri:
+                if not provider_uri:
+                    raise ValueError('Provider URI is required for decentralized Porter.')
```
piotr-roslaniec, line 89:
```suggestion
            if not provider_uri:
                raise ValueError('Provider URI is required for decentralized Porter.')
```

nucypher/utilities/porter/control/specifications/porter_schema.py
derekpierre: In this case it is more like b), except that if we use `max_quantity` it can imply that the returned number of ursulas has the option to be less than the max. It is more of a *"Give me `quantity` ursulas, with a preference for Ursulas in `include_ursulas"`*. 

Note that It isn't guaranteed that all of the Ursulas in `include_ursulas` will be included since they may have connectivity issues at the time of the operation, in which case they will not be included (tests don't have this problem 😄 )

The current terms used in the python API are `n` and `handpicked_ursulas` for a grant, but felt too ambiguous for the REST API.

nucypher/utilities/porter/porter.py
derekpierre: 🤦 - thanks for catching this.

nucypher/characters/utils.py
```diff
@@ -0,0 +1,113 @@
...
+            raise _MapClass.NowhereToBeFound(f"Asked {len(learner.known_nodes)} nodes, "
+                                             f"but none had map {map_identifier}")
+
+
+def matching_nodes_among(nodes: FleetSensor,
+                         bob_encrypting_key: UmbralPublicKey,
+                         no_less_than=7):  # Somewhat arbitrary floor here.
+    # Look for nodes whose checksum address has the second character of Bob's encrypting key in the first
+    # few characters.
+    # Think of it as a cheap knockoff hamming distance.
```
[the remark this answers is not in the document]
derekpierre: These are nodes picked for the treasure map distribution which is different from nodes picked for policies. The "hamming distance" algorithm was derived by @jMyles , and makes finding treasure maps on the network faster by being more deterministic; Alice has to distribute the treasure map on the network, and then Bob has to find it on the network - if they both use the same algorithm for deciding which nodes to use/check with then the process is more efficient. 

This was previous functionality that I simply moved to be in a common location so that both Characters and Porter could use the same code.

In the TMapConKFrags world, distributing the treasure map would not be required, and Bob would just store it himeself.

nucypher/characters/utils.py
```diff
@@ -0,0 +1,113 @@
+"""
```
derekpierre, line 1: Would this file be better situated in `nucypher.network`?

nucypher/utilities/porter/porter.py
derekpierre: ✔️

nucypher/characters/utils.py
derekpierre: ✔️ - moved file

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -40,13 +35,19 @@ def _deserialize(self, value, attr, data, **kwargs):
...
-                                  )  # TODO: USe the one from TMap
-        try:
-            signature, hrac, tmap_message_kit = splitter(value)
-            return True
-        except InvalidNativeDataTypes as e:
-            raise InvalidInputData(f"Could not parse {self.name}: {e}")
+        from nucypher.policy.collections import SignedTreasureMap
+        from nucypher.policy.collections import TreasureMap as TreasureMapBase
+
+        # TODO unsure which is which is there a way for one splitter to work in both cases?
```
fjarri, line 41: This should be easier with the `get_splitter()` in #2719.

Although looking at a slightly bigger picture, how come we don't know at this point whether it's a signed treasure map or a not signed one? Shouldn't there rather be a separate `class SignedTreasureMap` definition here?

nucypher/network/utils.py
```diff
@@ -0,0 +1,113 @@
+"""
```
fjarri, line 1: I'm not sure if `utils` is the best name for this module. So far it seems treasure-map-oriented, do you think something more specific would suit it better?

nucypher/policy/policies.py
```diff
@@ -96,15 +86,33 @@ class TreasureMapPublisher:
...
+                 network_middleware,
                  percent_to_complete_before_release=5,
                  threadpool_size=120,
                  timeout=20):
 
         self._total = len(nodes)
         self._block_until_this_many_are_complete = math.ceil(len(nodes) * percent_to_complete_before_release / 100)
-        self._worker_pool = WorkerPool(worker=worker,
+
+        def put_treasure_map_on_node(node):
```
fjarri, line 99: Thank you for that, I should've done it when I made this class. There was no reason to keep the worker function outside.

nucypher/characters/control/specifications/fields/treasuremap.py
derekpierre: `_validate()` is called as part of broader `deserialize()` function for Fields, and is passed the result of `_deserialize()` (function above it). This means that the `value` parameter passed to `_validate()` is in bytes, and therefore it's not clear to me that we can differentiate whether the bytes are expected to be (note not what we receive, but rather what we expect to receive) i.e. a signed tmap or a federated tmap

In practice, it will probably be just a signed treasure map since Porter is intended for decentralized workflow, but I've found accommodating the federated workflow to be good for testing.

nucypher/network/utils.py
derekpierre: 👍 - Yeah probably, I was on the fence about whether other things may be added, but probably best to keep it scoped to treasure map stuff.

nucypher/policy/policies.py
derekpierre: 👍

nucypher/characters/control/specifications/fields/treasuremap.py
fjarri: What I mean is that, for example, we don't try to decide from a stream of bytes whether it's a public key or a signature. We already know what it should be, and if it's not that, we raise and exception. Why doesn't it work the same for treasure maps? 

I am not familiar with this part of the code. Is it because this validator is used on something the user is passed via CLI? Something like `nucypher bob --treasure-map ...`. So it is not a good idea to make a separate `--signer-treasure-map` option, and even though we know whether we in a federated or a blockchain context, there is currently no mechanism to pass that information through `deserialize()`?

nucypher/policy/reservoir.py
```diff
@@ -0,0 +1,115 @@
...
+along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+"""
+from typing import Iterable, Optional, List
+
+from eth_typing import ChecksumAddress
+
+from nucypher.blockchain.eth.agents import StakersReservoir, StakingEscrowAgent
+
+
+def make_federated_staker_reservoir(learner: 'Learner',
```
fjarri, line 24: May be enough to just pass the `FleetSensor` object here

nucypher/characters/control/specifications/fields/treasuremap.py
derekpierre: It would be something like this:

```bash
$ curl <PORTER_URI>/publish_treasure_map -X GET -H "Content-Type: application/json" -d ' {"treasure_map": "Qld7S8sbKFCv2B8KxfJo4oxiTOjZ4VPyqTK5K1xK6DND6TbLg2hvlGaMV69aiiC5QfadB82w/5q1Sw+SNFHN2esWgAbs38QuUVUGCzDoWzQAAAGIAuhw12ZiPMNV8LaeWV8uUN+au2HGOjWilqtKsaP9fmnLAzFiTUAu9/VCxOLOQE88BPoWk1H7OxRLDEhnBVYyflpifKbOYItwLLTtWYVFRY90LtNSAzS8d3vNH4c3SHSZwYsCKY+5LvJ68GD0CqhydSxCcGckh0unttHrYGSOQsURUI4AAAEBsSMlukjA1WyYA+FouqkuRtk8bVHcYLqRUkK2n6dShEUGMuY1SzcAbBINvJYmQp+hhzK5m47AzCl463emXepYZQC/evytktG7yXxd3k8Ak+Qr7T4+G2VgJl4YrafTpIT6wowd+8u/SMSrrf/M41OhtLeBC4uDKjO3rYBQfVLTpEAgiX/9jxB80RtNMeCwgcieviAR5tlw2IlxVTEhxXbFeopcOZmfEuhVWqgBUfIakqsNCXkkubV0XS2l5G1vtTM8oNML0rP8PyKd4+0M5N6P/EQqFkHH93LCDD0IQBq9usm3MoJp0eT8N3m5gprI05drDh2xe/W6qnQfw3YXnjdvf2A=", "bob_encrypting_key": "026d1f4ce5b2474e0dae499d6737a8d987ed3c9ab1a55e00f57ad2d8e81fe9e9ac"}'
```

The received base64 encoded data would be deserialized to bytes.

Of course, validation isn't completely necessary as part of deserialization i.e. one option is that we can just take the bytes from deserialization, not do any validation, and pass them along to the `Porter.publish_treasure_map function` that ultimately passes them to the `TreasureMapPublisher` which takes bytes for treasure map. So we don't "need"  to validate, and therefore don't care about federated vs decentralized. If the bytes are invalid, presumably the publisher would fail.

nucypher/policy/reservoir.py
```diff
@@ -0,0 +1,115 @@
...
+                                    exclude_addresses: List[str] = None,
```
fjarri, line 25: Would `ChecksumAddress` fit better here instead of `str`? And, perhaps, `Optional[Iterable]` instead of `List`.

fjarri: Or you could make `()` the default value, thus avoiding the need for `Optional`, and creating `exclusion_set` just as `exclusion_set = set(include_addresses) | set(exclude_addresses)`

nucypher/utilities/porter/control/specifications/fields/treasuremapid.py
```diff
@@ -19,11 +19,14 @@
...
+from nucypher.crypto.constants import HRAC_LENGTH
 from nucypher.policy.collections import TreasureMap
 
 
 class TreasureMapID(BaseField, fields.String):
 
     def _validate(self, value):
-        if len(bytes.fromhex(value)) != TreasureMap.ID_LENGTH:
+        treasure_map_id = bytes.fromhex(value)
+        # FIXME federated has map id length 32 bytes but decentralized has length 16 bytes ... huh?
```
fjarri, line 30: Wait, aren't they different things? `TreasureMap` creates an ID based on HRAC (`self._id = keccak_digest(bytes(self._verifying_key) + bytes(self._hrac)).hex()`)

nucypher/utilities/porter/porter.py
```diff
@@ -69,6 +107,80 @@ def __init__(self,
             self.make_cli_controller()
         self.log.info(self.BANNER)
 
+    def get_treasure_map(self, map_identifier: str, bob_encrypting_key: UmbralPublicKey, timeout=3):
```
fjarri, line 110: Isn't `map_identifier` a `bytes` object? Also, perhaps I am being too picky, but it worries me that we're passing around plain byte objects (map ID here, or `treasure_map_bytes` below). Feels like it can be a source of errors. Ideally we should serialize at the latest possible moment.

nucypher/utilities/porter/control/specifications/fields/treasuremapid.py
fjarri: Disregard that, apparently in federated mode it's just `hrac.hex()`. Not a great thing to have, should really be made less error-prone in some way. A `MapID` class, perhaps?

nucypher/utilities/porter/porter.py
fjarri: My bad, it is in fact a string. Still it would be better off at least as a type alias (like `ChecksumAddress`).

nucypher/network/utils.py
derekpierre: ✔️

nucypher/policy/reservoir.py
derekpierre: ✔️

nucypher/policy/reservoir.py
derekpierre: Not a fan of the default empty set, but made the param Optional[Iterable] and used ChecksumAddress - ✔️

nucypher/utilities/porter/control/specifications/fields/treasuremapid.py
derekpierre: `MapID` could be good but that affects what `TreasureMap.public_id()` should return - trying to keep my surface area of changes limited.

nucypher/utilities/porter/porter.py
derekpierre: Same comment for `MapID` in your other comment (https://github.com/nucypher/nucypher/pull/2717#discussion_r651036501). If we are going to do it, we should address it in all areas.

nucypher/policy/reservoir.py
fjarri: Yes, I know that mutable defaults is a controversial thing, but I was talking about a `()`, an empty tuple. Should be fine, it's immutable.

derekpierre: I rather ensure that the `None` case is handled by accepting `None` as the default, as opposed to using a tuple as the default, and `None` could be passed by callers and not handled appropriately. Using a tuple as default forces extra code in callers of the function eg. `Porter.get_ursulas()` to prevent `None` from being passed to the reservoir function.

I did some cleanup using the recommendation about set intersection though 👍 .

piotr-roslaniec (approved):

nucypher/network/treasuremap.py
```diff
@@ -0,0 +1,116 @@
...
+import maya
+from umbral.keys import UmbralPublicKey
+
+from nucypher.acumen.perception import FleetSensor
+from nucypher.crypto.signing import InvalidSignature
+from nucypher.network.exceptions import NodeSeemsToBeDown
+from nucypher.network.nodes import Learner
+
+
+def get_treasure_map_from_known_ursulas(learner: Learner,
```
cygnusv, line 28: Great use of the Learner abstraction 👍

nucypher/policy/policies.py
```diff
@@ -15,40 +15,29 @@
...
 from nucypher.crypto.api import keccak_digest, secure_random
 from nucypher.crypto.constants import HRAC_LENGTH, PUBLIC_KEY_LENGTH
 from nucypher.crypto.kits import RevocationKit
 from nucypher.crypto.powers import DecryptingPower, SigningPower, TransactingPower
 from nucypher.crypto.utils import construct_policy_id
-from nucypher.network.exceptions import NodeSeemsToBeDown
 from nucypher.network.middleware import RestMiddleware
 from nucypher.utilities.concurrency import WorkerPool, AllAtOnceFactory
 from nucypher.utilities.logging import Logger
+from .reservoir import make_federated_staker_reservoir, MergedReservoir, PrefetchStrategy, \
```
cygnusv, line 39: What's the preferred way, `.reservoir` or `nucypher.policy.reservoir`?

nucypher/utilities/porter/control/interfaces.py
```diff
@@ -31,23 +34,26 @@ def __init__(self, porter: 'Porter' = None, *args, **kwargs):
...
                              bob_encrypting_key: bytes) -> dict:
-        # Steps (analogous to nucypher.character.control.interfaces):
-        # 1. creation of relevant objects / setup
-        # 2. call self.implementer.some_function() i.e. Porter learner has an associated function to call
-        # 3. create response
-        pass
+        bob_enc_key = UmbralPublicKey.from_bytes(bob_encrypting_key)
+        self.implementer.publish_treasure_map(treasure_map_bytes=treasure_map,
+                                              bob_encrypting_key=bob_enc_key)
+        return {}
```
cygnusv, line 56: I'm wondering what can we return here, like the list of Ursulas holding the TMap.

nucypher/utilities/porter/control/specifications/fields/treasuremapid.py
cygnusv:
```suggestion
        # FIXME federated has map id length 32 bytes but decentralized has length 16 bytes ... huh? - #2725
```
Opened issue #2725

nucypher/policy/policies.py
derekpierre: ✔️ - that may have been something weird with my PyCharm.

nucypher/utilities/porter/control/interfaces.py
derekpierre: I guess my question would be what's the advantage there? The distribution of the treasure map is based on the hamming distance algorithm based on bob's encrypting key. Bob will inherently use that same algorithm through Porter via `/get_treasure_map`.

nucypher/utilities/porter/control/specifications/fields/treasuremapid.py
```diff
@@ -19,11 +19,14 @@
...
+from nucypher.crypto.constants import HRAC_LENGTH
 from nucypher.policy.collections import TreasureMap
 
 
 class TreasureMapID(BaseField, fields.String):
 
     def _validate(self, value):
-        if len(bytes.fromhex(value)) != TreasureMap.ID_LENGTH:
+        treasure_map_id = bytes.fromhex(value)
+        # FIXME federated has map id length 32 bytes but decentralized has length 16 bytes ... huh? - #2725
```
KPrasch, line 30: Needs to be revised with special consideration towards revocation.  16 bytes is the length of the field in `PolicyManager`, so let's stick to that unless we plan to modify the contract.  HRAC / Policy ID length must be the first 16 bytes, truncated from the keccack digest regardless of operating mode.

nucypher/utilities/porter/control/specifications/porter_schema.py
```diff
@@ -102,6 +104,15 @@ def check_valid_quantity_and_include_ursulas(self, data, **kwargs):
...
 
+    @validates_schema
+    def check_include_and_exclude_are_mutually_exclusive(self, data, **kwargs):
+        ursulas_to_include = data.get('include_ursulas') or []
+        ursulas_to_exclude = data.get('exclude_ursulas') or []
+        common_ursulas = set(ursulas_to_include).intersection(ursulas_to_exclude)
+        if len(common_ursulas) > 0:
+            raise InvalidArgumentCombo(f"Ursulas to include and exclude are not mutually exclusive; "
+                                       f"common entries {common_ursulas}")
+
```
KPrasch, line 115: Attention to details 👍🏻

nucypher/utilities/porter/porter.py
```diff
@@ -14,12 +14,30 @@
  You should have received a copy of the GNU Affero General Public License
  along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
-from constant_sorrow.constants import NO_CONTROL_PROTOCOL
```
KPrasch, line 17: Porter living inside a `utilities` module is short-selling it's greatness, and a case in favor of moving to a separate repo IMO (while I still do greatly appreciate the benefits of a unified CI).

nucypher/utilities/porter/porter.py
```diff
@@ -69,6 +108,80 @@ def __init__(self,
             self.make_cli_controller()
         self.log.info(self.BANNER)
 
+    def get_treasure_map(self, map_identifier: str, bob_encrypting_key: UmbralPublicKey, timeout=3):
+        return treasuremap.get_treasure_map_from_known_ursulas(learner=self,
+                                                               map_identifier=map_identifier,
+                                                               bob_encrypting_key=bob_encrypting_key,
+                                                               timeout=timeout)
+
+    def publish_treasure_map(self, treasure_map_bytes: bytes, bob_encrypting_key: UmbralPublicKey):
```
KPrasch, line 117:
```suggestion
    def publish_treasure_map(self, treasure_map_bytes: bytes, bob_encrypting_key: UmbralPublicKey) -> None:
```

KPrasch (commented):
This is the meant and potatoes of Porter, nice work.  There's also some excellent code reorg in here that introduces a higher degree of testability.  Still looking it over, but left a few comments and questions while I get my head around this somewhat large changeset.

nucypher/utilities/porter/porter.py
cygnusv: @derekpierre and I talked about this a while ago and we agreed that a separate repo is indeed a potential outcome. We also discussed the difficulties that we've seen with nucypher-monitor living in a separate repo in terms of synchronization. I'd personally lean towards keeping it inside the same repo, potentially relocated to a more prominent namespace.

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -40,13 +35,19 @@ def _deserialize(self, value, attr, data, **kwargs):
...
+        exceptions = ""
+        for splitter_type, splitter in splitters:
+            try:
+                _ = splitter(value)
+                return True
+            except InvalidNativeDataTypes as e:
+                # store exception
+                exceptions += f"[{splitter_type}]: {e} ;"
+
+        raise InvalidInputData(f"Could not parse {self.name}: {exceptions}")
```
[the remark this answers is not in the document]
derekpierre: Potential fix via https://github.com/nucypher/nucypher/pull/2717/commits/f18db740a496fd013a2613363c4836e9224e1183. @vepkenez you should take a look as well. A bit of indirection but works decently.

nucypher/characters/control/specifications/fields/treasuremap.py
derekpierre: Potential fix via https://github.com/nucypher/nucypher/pull/2717/commits/f18db740a496fd013a2613363c4836e9224e1183.

KPrasch (approved):
Looking good!

PR #2719, opened 2021-06-02
TreasureMap con KFrags fixes, mostly with compatibility and splitting previous versions 

jMyles:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

Versions TreasureMaps, replaces some of @KPrasch's work within the class with the work @vepkenez had previously done to achieve very similar functionality.  Also "brands" TreasureMaps - and distinctly, SignedTreasureMaps, for easy ascertainment across the network.

Updates depedencies.


nucypher/policy/maps.py
```diff
@@ -36,18 +35,19 @@
...
+class TreasureMap:
+    VERSION_NUMBER = 1  # Increment when serialization format changes.
 
-    _DELIMITER = b':'
-    _PREFIX = b'TM' + _DELIMITER
-    _VERSION = int(VERSION_NUMBER).to_bytes(1, 'big')
-    _HEADER = _PREFIX + _VERSION
-    _HEADER_SIZE = len(_HEADER)
+    _BRAND = b'TM'
+    _VERSION = int(VERSION_NUMBER).to_bytes(2, 'big')
```
fjarri, line 50: Maybe it's better to have `VersioningMixin.HEADER_LENGTH` instead of `2`. And perhaps some control for the `_BRAND` length as well.

nucypher/policy/maps.py
```diff
@@ -121,37 +122,47 @@ def __len__(self):
...
+            # This representation is compatible with a known stencil.
+            splitter = cls._splitters[version]
+        else:
+            # It's possible that this is a preversioned representation.
+            splitter = cls._splitters['unversioned']  # TODO: In this case, it's still a map from a previous version - how will we handle sin KFrags?
+        return splitter
+
+    _splitters = {}
+
+    def __new__(cls, *args, **kwargs):
```
fjarri, line 149: Can `_splitters` be defined as a class attribute? Looks like everything here is known at compile time.

nucypher/policy/maps.py
```diff
@@ -324,21 +335,34 @@ def author(cls,
...
-                                    public_signature=Signature,
-                                    hrac=(bytes, HRAC_LENGTH),
-                                    message_kit=(UmbralMessageKit, VariableLengthBytestring),
-                                    blockchain_signature=65)
+    def __new__(cls, *args, **kwargs):
+        # TODO: This is a little too brittle-and-repeaty for my tastes; maybe we modify
+        # BSS to allow summing *before* the prefix(es)?
+
+        cls._splitters['unversioned'] = BytestringKwargifier(cls,
+                                                 blockchain_signature=65,
```
fjarri, line 349: `65` warrants a constant, e.g. in `nucypher.crypto.constants`. Maybe dependent on `Signature` size.

nucypher/policy/maps.py
```diff
@@ -351,4 +375,6 @@ def __bytes__(self):
             raise self.InvalidSignature(
                 "Can't cast a SignedTreasureMap to bytes until it has a blockchain signature "
                 "(otherwise, is it really a 'SignedTreasureMap'?")
```
fjarri, line 377: Interesting point. Is it really a ` SignedTreasureMap` object if it doesn't have a blockchain signature?

tests/acceptance/characters/control/test_rpc_control_blockchain.py
```diff
@@ -74,7 +74,7 @@ def test_bob_rpc_character_control_retrieve_with_tmap(
     wrong_tmap.prepare_for_publication(
         bob_encrypting_key=blockchain_bob.public_keys(DecryptingPower),
         alice_stamp=blockchain_alice.stamp)
-    wrong_tmap._blockchain_signature = os.urandom(64)    # ...because it only matters when Ursula looks at it.
+    wrong_tmap._blockchain_signature = b"this is not a signature, but we don't need one for this test....."  # ...because it only matters when Ursula looks at it. (65 bytes)
```
fjarri, line 77: See comment above about a constant for `65`

tests/integration/learning/test_domains.py
```diff
@@ -163,4 +163,4 @@ def test_learner_uses_both_nodes_from_storage_and_fallback_nodes(lonely_ursula_m
     learner.learn_from_teacher_node()
     all_nodes = {teacher}
     all_nodes.update(other_nodes)
-    assert set(learner.known_nodes) == all_nodes
+    assert set(n.mature() for n in learner.known_nodes) == all_nodes
```
fjarri, line 166: Won't be necessary with #2709

nucypher/acumen/perception.py
```diff
@@ -149,9 +149,7 @@ def with_updated_nodes(self,
             nodes = dict(self._nodes)
             nodes_to_add_dict = {node.checksum_address: node for node in nodes_to_add}
             for checksum_address in diff.nodes_updated:
-                new_node = nodes_to_add_dict[checksum_address]
-                new_node.mature()
-                nodes[checksum_address] = new_node
+                nodes[checksum_address] = nodes_to_add_dict[checksum_address]
```
fjarri, line 152: Fixed in #2709

nucypher/policy/maps.py
jMyles: Agreed.

nucypher/policy/maps.py
jMyles: Oh yeah, that might now be the case.

tests/acceptance/characters/control/test_rpc_control_blockchain.py
jMyles: Not sure I understand this one - are you suggesting going back to the random variant?

I like having pre-determined text as the placeholder here for debugging purposes.

fjarri: Yes, sorry, it was ambiguous. I think it is worth it to add an assertion to make sure the length of the text is actually the expected length of the signature.

KPrasch:
The CI failure for gas estimation looks like it can be resolved with a simple rename, might as well include it in this PR.

KPrasch (approved):
Great advancement  - Definitely good enough to merge into the epic.  I even suggest reducing the amount of required approvals to two.

nucypher/policy/maps.py
jMyles: Actually, they do need the `cls` object.  Unless.... there's some better way... which there probably is....?

fjarri: Hm, I don't understand. What happens if you just create the `_splitters` class attribute with those two values?

jMyles: Ughh, what?

`NameError` for `cls` I suppose.

nucypher/crypto/constants.py
```diff
@@ -21,6 +21,7 @@
 # Policy component sizes
 HRAC_LENGTH = 16
 SIGNATURE_SIZE = 64
+EIP712_MESSAGE_SIGNATURE_SIZE = 65
```
fjarri, line 24: I just realized we already have `LENGTH_ECDSA_SIGNATURE_WITH_RECOVERY = 65` in `blockchain/eth/constants.py`. Is it the same format, or just the same length?

nucypher/policy/maps.py
```diff
@@ -121,37 +122,47 @@ def __len__(self):
...
-            except ValueError:
-                raise ValueError('Invalid treasure map header.')
-            raise cls.OldVersion(f'Treasure map is an old version ({version} but the latest version is {cls.VERSION_NUMBER}')
+        representation_metadata = TreasureMapSplitter.get_metadata(bytes_representation)
+
+        # header = bytes_representation[:cls._HEADER_SIZE]
+        brand_matches = representation_metadata['brand'] == cls._BRAND
+        version = representation_metadata['version']
+
+        if version <= cls.VERSION_NUMBER and brand_matches:
```
fjarri, line 139: Maybe `if version in cls._splitters` would be a more fitting check

nucypher/policy/maps.py
fjarri: That's right, I didn't realize `cls` itself is used here.

tests/integration/learning/test_domains.py
jMyles: Groovy.  I just yanked this commit entirely.

nucypher/policy/maps.py
```diff
@@ -121,37 +122,47 @@ def __len__(self):
...
-                                    hrac=(bytes, HRAC_LENGTH),
-                                    message_kit=(UmbralMessageKit, VariableLengthBytestring))
+        return self._BRAND + self._VERSION + self._payload
 
     @classmethod
-    def _check_version(cls, bytes_representation: bytes) -> None:
+    def get_splitter(cls, bytes_representation: bytes) -> None:
         """
         Takes a bytes representation of a treasure map and raises OldVersion
         error is the version is incompatible or ValueError if the header is malformed.
```
fjarri, line 131: Is it still the case?

fjarri: Also the return annotation seems to be outdated.

nucypher/crypto/constants.py
jMyles: Arguably the same format, though `LENGTH_ECDSA_SIGNATURE_WITH_RECOVERY` is more ambiguous.

jMyles: Presumably it's plausible that we might later accept a different ECDSA signature with recovery, or even more than one of them.  Obviously there are many different bytes representations to go with different schemes.  Whereas, EIP712 is unambiguously a 65-byte standard.

fjarri (approved):
Approved with reservations to avoid blocking further work:
- I am still not sure that the `__new__` approach is the best here, I'll revisit the topic when we have another versioned class and have to copy-paste or generalize.
- In fact, I am not even certain that object versioning as opposed to protocol interactions versioning is the way to go, but this PR does not close the way to the latter yet.
- I think it is possible to make `SignedTreasureMap` valid on creation, and not after manually setting all the required fields with additional method calls - I'll try it out later.

PR #2720, opened 2021-06-09
Allow Porter to launch with TLS

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Based over #2717 (which should be reviewed first).

`--tls-key-filepath` and `--certificate-filepath` options can now be provided to `nucypher porter run ... ` for Porter to be launched with TLS i.e. https.

For testing you can create a tls key and certificate for the localhost using the openssl command:
```bash
openssl req -x509 -out fullchain.pem -keyout key.pem \
  -newkey rsa:2048 -nodes -sha256 \
  -subj '/CN=localhost' -extensions EXT -config <( \
   printf "[dn]\nCN=localhost\n[req]\ndistinguished_name = dn\n[EXT]\nsubjectAltName=DNS:localhost\nkeyUsage=digitalSignature\nextendedKeyUsage=serverAuth")
```

### DIFF
--- a/nucypher/control/controllers.py
+++ b/nucypher/control/controllers.py
@@ -269,18 +269,17 @@ class WebController(InterfaceControlServer):
     def start(self,
               port: int,
               tls_key_filepath: str = None,
-              certificate_filepath: str = None,
+              tls_certificate_filepath: str = None,
               dry_run: bool = False):
-
-        self.log.info("Starting HTTP/HTTPS Control...")
         if dry_run:
             return
 
-        if tls_key_filepath and certificate_filepath:
+        if tls_key_filepath and tls_certificate_filepath:
+            self.log.info("Starting HTTPS Control...")
             # HTTPS endpoint
             hx_deployer = HendrixDeployTLS(action="start",
                                            key=tls_key_filepath,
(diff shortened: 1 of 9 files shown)


nucypher/control/controllers.py
```diff
@@ -265,15 +266,36 @@ def make_control_transport(self):
...
 
-    def start(self, http_port: int, dry_run: bool = False):
+    def start(self,
+              port: int,
+              tls_key_filepath: str = None,
+              certificate_filepath: str = None,
+              dry_run: bool = False):
 
-        self.log.info("Starting HTTP Control...")
+        self.log.info("Starting HTTP/HTTPS Control...")
```
KPrasch, line 275: Might as well move the log lower since you know whether or not the process is being launched with HTTPS or not.

KPrasch (approved):
🤠

nucypher/control/controllers.py

piotr-roslaniec (approved):

Issue #2706, opened 2021-05-20
Porter Authentication


derekpierre:
How does Porter authenticate REST API requesters?

While there is nothing to leak (in the current scope of functionality) - there is the possibility of DDOS-type attacks.

Mentioned in https://github.com/nucypher/productdev/issues/7#issuecomment-730612379.
* Infura by default requires project id API key - can also be extended to use basic auth - https://infura.io/docs/gettingStarted/authentication
* Filecoin uses basic auth: username (project id) and password (project secret) - https://infura.io/docs/filecoin#section/Securing-Your-Credentials'

derekpierre:
It is simplified to only Basic Authentication over HTTPS via #2726 for now unless users decide they want something more robust. However, this is a sufficient starting point. More intricate functionality can be inherently obtained via Nginx proxy (reverse proxy server) sitting in front of Porter.

PR #2724, opened 2021-06-16
Porter Docker

### LINKED ISSUE #2669: Containerization/Deployment of Porter
For the initial iteration of Porter, it would be beneficial for it to be run centrally by application providers, or potentially as a service that 3rd parties can provide.

At the very least, some sort of containerization (eg. docker) would be good. This has security implications if Porter does anything requiring trust other than as a conduit to the network protocol - but security may be out-of-scope for the version 1.

Current docker is `Ursula`-focused, what can be done for `Porter` (ports etc).

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Docker compose for Porter service. Based over #2720 .

**Issues fixed/closed:**

Fixes #2669

### DIFF
--- a/deploy/docker/porter/docker-compose.yml
+++ b/deploy/docker/porter/docker-compose.yml
@@ -1,0 +1,37 @@
+version: '3'
+
+services:
+  porter-http:
+    restart: on-failure
+    image: nucypher:latest
+    build:
+      context: ../../..
+      dockerfile: deploy/docker/Dockerfile
+    ports:
+      # Default Porter port
+      - "80:9155"
+    volumes:
+      - .:/code
+      - ~/.local/share/nucypher:/nucypher
+    command: ["nucypher", "porter", "run",
+              "--provider", "${WEB3_PROVIDER_URI}",
+              "--network", "${NUCYPHER_NETWORK}"]
+
+  porter-https:
+    restart: on-failure
+    image: nucypher:latest
+    build:
+      context: ../../..
+      dockerfile: deploy/docker/Dockerfile
+    ports:
+      # Default Porter port
+      - "443:9155"
+    volumes:
+      - .:/code
+      - ~/.local/share/nucypher:/nucypher
+      - "${TLS_DIR}:/etc/porter-tls/"
+    command: [ "nucypher", "porter", "run",
+               "--provider", "${WEB3_PROVIDER_URI}",
+               "--network", "${NUCYPHER_NETWORK}",
+               "--tls-key-filepath", "/etc/porter-tls/key.pem",
+               "--tls-certificate-filepath", "/etc/porter-tls/cert.pem"]


deploy/docker/porter/docker-compose.yml
```diff
@@ -0,0 +1,37 @@
...
+      - "443:9155"
+    volumes:
+      - .:/code
+      - ~/.local/share/nucypher:/nucypher
+      - "${TLS_DIR}:/etc/porter-tls/"
+    command: [ "nucypher", "porter", "run",
+               "--provider", "${WEB3_PROVIDER_URI}",
+               "--network", "${NUCYPHER_NETWORK}",
+               "--tls-key-filepath", "/etc/porter-tls/key.pem",
+               "--tls-certificate-filepath", "/etc/porter-tls/cert.pem"]
```
KPrasch, line 37: Essentially, you expect that users will modify these paths to match their own system?

local TLS termination is a nice feature here.

derekpierre: There is an associated volume mapping on L32.

```
"${TLS_DIR}:/etc/porter-tls/"
```

so my thought was that users would create a folder with `key.pem` and `cert.pem` files and then set the `TLS_DIR` environment variable to be that folder.

vepkenez:
I actually really think that all of this stuff with TLS is not where we want to be spending our time.  Nginx really does that well, and it is trivial and normal and best practice to use that highly optimized software for TLS termination.

Outside of places where we want to tie together identity with certificates (which has led to some minor ruin), it doesn't really make sense to have code in NuCypher that deals with https.

Obviously I had forever to comment on this and it's a little late right now and I deeply respect and appreciate the time spent dealing with it and I'm sorry I didn't chime in with all this at a more appropriate hour but all the lines of code dealing with TLS CLI options are lines of code that are already in NGINX and could be gotten for free with docker-compose.

derekpierre:
@vepkenez, as part of running Porter we are already running a web server (Hendrix) so at the moment putting nginx on top of that seems superfluous.

That being said, as a view to the future, I have been playing around a little with using nginx as a reverse proxy, to potentially allow for more complex authentication mechanisms for users of Porter (see #2706 ). It is unclear how Porter will be run, and how the users of the Porter service will be authenticated. Simple basic auth can be baked into Porter directly relatively easily, but anything more involved like OAuth2 or Token Authentication may necessitate something like nginx.

To that end this was the beginning of the docker-compose to get nginx working as a reverse proxy - without any authentication:

```yml
version: '3'

services:

  nginx-proxy:
    restart: always
    image: nginxproxy/nginx-proxy:alpine
    ports:
      - "80:80"
      - "443:443"
    volumes:
      - /var/run/docker.sock:/tmp/docker.sock:ro
      - "${TLS_DIR}:/etc/nginx/certs/"

  porter:
    restart: on-failure
    image: nucypher:latest
    build:
      context: ../../..
      dockerfile: deploy/docker/Dockerfile
    expose:
      # Default Porter port
      - "9155"
    volumes:
      - .:/code
      - ~/.local/share/nucypher:/nucypher
    command: [ "nucypher", "porter", "run",
               "--provider", "${WEB3_PROVIDER_URI}",
               "--network", "${NUCYPHER_NETWORK}" ]
    environment:
      - VIRTUAL_HOST=porter.local
      - VIRTUAL_PORT=9155
    depends_on:
      - nginx-proxy
```

deploy/docker/porter/docker-compose.yml
KPrasch: Ah - I see it now.  Thanks for clarifying.

KPrasch (approved):
🤠

deploy/docker/porter/docker-compose.yml
derekpierre: 👍

vepkenez (approved):

PR #2726, opened 2021-06-21
Basic Authentication for Porter

### LINKED ISSUE #2706: Porter Authentication

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Based over #2724 .

Allows Porter to facilitate Basic Authentication over HTTPS based on a provided `htpasswd` file via `--basic-auth-filepath`. Updated Docker/docker-compose files accordingly.

**Issues fixed/closed:**
Fixes #2706  - does it fully "fix" it? perhaps for now unless users decide they want something more robust. However, this is a good starting point.

**Why it's needed:**
User authentication in some form will be needed if Porter is to be used as a public service.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -436,7 +436,7 @@ commands:
       - run: sudo chown -R circleci:circleci /usr/local/bin
       - run: sudo chown -R circleci:circleci /usr/local/lib/python3.7/site-packages
       - save_cache:
-          key: pip-v6-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
+          key: pip-v7-{{ .Branch }}-{{ checksum "Pipfile.lock" }}
           paths:
             - "~/.local/bin"
             - "~/.local/lib/python3.7/site-packages"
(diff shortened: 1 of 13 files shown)


vepkenez:
I am thinking that a very common use case for Porter will require CORS "allow-origin" style access control.  Essentially the ability to whitelist domains that requests are allowed to come from.

Similar to how Infura's "Allowlist Origins" work.
Screen Shot 2021-06-21 at 13 39 46

So people will be able to make websites and run a porter server to service them.

I guess this would be doable through an nginx proxy.
Maybe Hendrix/Twisted has some facility for this as well.

Have we talked about all this?

derekpierre:
> I guess this would be doable through an nginx proxy.
Maybe Hendrix/Twisted has some facility for this as well.
Have we talked about all this?

@vepkenez the question becomes what do we provide out of the box vs what do we leave for people who intend to run Porter to customize for their own needs. As you mentioned, some of the more intricate functionality can be obtained via nginx proxy sitting in front of Porter, so we should think of where to draw the line.

Let's chat more tomorrow.

vepkenez (approved):

KPrasch (approved):

PR #2727, opened 2021-06-24
Fix highperf tests for the new Umbral

fjarri:
Minimal changes, for starters.

**Type of PR:**
- Bugfix

**Required reviews:** 
- 2

**What this does:**
Fixes high-performance mocks to work with Umbral >=0.2 (including the Rust backend).

Fun fact: with `rust-umbral` mocking `SecretKey`/`PrivateKey` gives only a marginal performance improvement. On my machine, I have the following results for `test_alice_can_learn_about_a_whole_bunch_of_ursulas`:

rust-umbral, test / PyUmbral, test - rust_umbral, setup / PuUmbral, setup
Mocked: 4.77 / 4.63 - 27.86 / 27.78
Not mocked: 5.51 / 12.20 - 28.76 / 41.93


piotr-roslaniec (approved):

newsfragments/2727.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fix large-scale tests for the new Umbral API
```
derekpierre, line 1: Don't think that we need a newsfragment for fixing of tests.

tests/integration/learning/test_discovery_phases.py
```diff
@@ -101,33 +99,40 @@ def test_alice_verifies_ursula_just_in_time(fleet_of_highperf_mocked_ursulas,
...
-                                    publish_treasure_map=False)
+    def mock_encrypt(public_key, plaintext):
+        if not isinstance(public_key, PublicKey):
+            public_key = public_key.i_want_to_be_a_real_boy()
+        return encrypt(public_key, plaintext)
+
+    mocks = (
+        NotARestApp.replace_route("receive_treasure_map", mock_receive_treasure_map),
+        NotARestApp.replace_route("set_policy", mock_set_policy),
+        patch('nucypher.crypto.umbral_adapter.PublicKey.__eq__', lambda *args, **kwargs: True),
```
derekpierre, line 116: We no longer need to scope these patches? That is now handled by the code in the `fleet_of_highperf_mocked_ursulas` fixture?

fjarri: They are still applied in the same way as before, I just wanted to flatten 10 nested `with`s - that's what `ExitStack` is for.

derekpierre: 👍

derekpierre (approved):

newsfragments/2727.bugfix.rst
fjarri: Removed.

Issue #2714, opened 2021-05-28
Review some of the existing character control fields for serialize/deserlization consistency and validation


derekpierre:
Follow-up issue related to work done for #2713 .

Related to #2702 .

Some of the field definitions for the character control mechanism are inconsistent when comparing serialization/deserialization.

A demonstrative example would be the `UmbralSignature` field.

```python
class UmbralSignature(BaseField, fields.Field):

    def _serialize(self, value: Signature, attr, obj, **kwargs):
        return b64encode(bytes(value)).decode()

    def _deserialize(self, value, attr, data, **kwargs):
        if isinstance(value, bytes):
            return value
        try:
            return Signature.from_bytes(b64decode(value))
        except InvalidNativeDataTypes as e:
            raise InvalidInputData(f"Could not parse {self.name}: {e}")

    def _validate(self, value):
        try:
            Signature.from_bytes(value)
        except InvalidNativeDataTypes as e:
            raise InvalidInputData(f"Could not parse {self.name}: {e}")
 ```
 
Within the `marshmallow.fields` library that we use for our fields, the result of the `_deserialize` function gets passed to the `_validate` function. In this case either bytes or an actual `Signature` object can be returned by `_deserialize`, but `_validate`  seems to only care about bytes and would fail validation if a `Signature` object was passed to it.
 
Also, why doesn't `_deserialize` not just return a `Signature` object, why does it need to be either `bytes` or `Signature`? I suspect it was working within a previously written python API to work, so it was required to work with legacy code.
 
 For example it could be made to instead be:
 ```python
 class UmbralSignature(BaseField, fields.Field):

    def _serialize(self, value: Signature, attr, obj, **kwargs):
        return b64encode(bytes(value)).decode()

    def _deserialize(self, value, attr, data, **kwargs):
        if not isinstance(value, bytes):
            value = b64decode(value)

        try:
            return Signature.from_bytes(value)
        except InvalidNativeDataTypes as e:
            raise InvalidInputData(f"Could not parse {self.name}: {e}")
```

Another inconsistency is the `Cleartext` field:
```python
class Cleartext(BaseField, fields.String):

    def _serialize(self, value, attr, data, **kwargs):
        return value.decode()

    def _deserialize(self, value, attr, data, **kwargs):
        return b64encode(bytes(value, encoding='utf-8')).decode()
```
The `_serialize` and `_deserialize` functions are not opposites i.e. the output of serialize can't be passed to deserialize which is a bit strange.

[...]

PR #2730, opened 2021-06-25
Retrieval protocol

### LINKED ISSUE #259: Define format for re-encryption metadata

### LINKED ISSUE #2791: Make `Bob.follow_treasure_map()` less obscure
It seems that the purpose of this method is to wait until enough nodes become available to retrieve with a given treasure map. But in its current state it has a few problems:
- The name does not really convey what it does. Something like `prepare_for_retrieval()` would be more straightforward.
- The docstring is misleading. It says "Determines which Ursulas are known and which are unknown." We don't really use its return value, its main purpose is to wait until Ursulas are known. The user should be given a clear strategy of how this method is supposed to be used.
- It takes a decrypted treasure map, which is not trivial to obtain (with public API)
- Might be more intuitive if by default it waited for `threshold` Ursulas to be online, not all of them - that's what we really need

fjarri:
**Type of PR:**
- Bugfix
- Code quality

**Required reviews:** 
- 3

**What this does:**
- removes the unused `bob` parameter of `TreasureMap.construct_by_publisher()` (introduced in #2773)
- removes `Bob.matching_nodes_among()`
- removes `EncryptedTreasureMap` datastore model
- removes the `UmbralMessageKit` alias of `PolicyMessageKit`
- moves `TLSHostingPower` to `powers.py` from `server.py` (a proper place for it, also helps avoid some import cycles)
- renames `PolicyMessageKit` to `MessageKit`. This is the externally exposed message-for-reencryption, so I think it makes sense to avoid unnecessary qualifiers - previously `MessageKit` wasn't even instantiated anywhere. 
- moves `kits.py` to `policies/` (seems like the proper place for it, the objects there are all policy-related)
- makes `encrypt_and_sign()` a classmethod constructor of `MessageKit` (renamed to `author()`)
- `author()` from above only returns `MessageKit` and not a tuple of the kit and the signature - the latter is already a part of the kit
- fixes #259 by removing excess parts of the data Bob sends to Ursula
- WorkOrder/PRETask/MessageKit streamlining (the existing behavior of externally distributed message kits is preserved; more changes may be made in other PRs depending on the discussion in #2743)
- Bob supports an external caching of cfrags (through `PolicyMessageKit` objects)
- New protocol objects: `ReencryptionRequest` and `ReencryptionResponse` 
- `encrypted_treasure_map` is now required in `retrieve()`; `label` and `publisher_verifying_key` are not needed
- `enrico` parameter is removed from `retrieve`; `policy_encrypting_key` is mandatory
- reworked the availability check before retrieval (former `Bob.follow_treasure_map`). Fixes #2791 
- (TODO) support serialization of `PolicyMessageKit` objects? For Bob to store them somewhere.
- (TODO) naming stuff. `retrieve()`, `retrieve_cfrags()`, `PolicyMessageKit`, `RetrievalResult` are the main offenders.

The whole retrieval procedure has been significantly simplified. Bob creates a `RetrievalPlan` plan based on Ursulas from the treasure map, the capsules in the kits, and possibly the attached cfrags in the kits. From it work orders for Ursulas are spawned and the results saved back into the plan.

You can see the new usage in `test_bob_handles_frags.py`. A lot of tests there were removed (because they tested the internal workings of `retrieve()`) and replaced with new ones.

**For reviewers:**
- is `retrieve()` is a good name? Maybe `retrieve_and_decrypt()`?
- should `RetrievalClient` be a class? It is pretty much stateless and could probably be rewritten as a function.


nucypher/characters/control/interfaces.py
```diff
@@ -220,6 +221,19 @@ def retrieve(self,
 
         self.character.join_policy(label=label, publisher_verifying_key=alice_verifying_key)
 
+        if self.character.federated_only:
+            from nucypher.policy.maps import TreasureMap as _MapClass
+        else:
+            from nucypher.policy.maps import SignedTreasureMap as _MapClass
+
+        # TODO: This LBYL is ugly and fraught with danger.  NRN
+        if isinstance(treasure_map, bytes):
```
derekpierre, line 232: I'm a tad bit confused here, the `/retrieve` endpoint schema `Bob.Retrieve` (https://github.com/fjarri/nucypher/blob/retrieval-protocol/nucypher/characters/control/specifications/bob.py#L43) does not include a treasure map in the request data - how is a treasure map even passed here via a request? Presumably, it is always `None` (L207).

Pinging @vepkenez / @jMyles, since the treasure_map parameter in this function is prior to @fjarri 's current change.

nucypher/blockchain/eth/actors.py
```diff
@@ -1202,6 +1206,26 @@ def missing_commitments(self) -> int:
         missing = self.staking_agent.get_missing_commitments(checksum_address=staker_address)
         return missing
 
+    def verify_policy_payment(self, hrac: bytes) -> None:
+        arrangements = self.policy_agent.fetch_policy_arrangements(policy_id=hrac)
+        members = set()
```
jMyles, line 1211: It took me a while to understand what's happening here - we populate a set only to later use it to determine error message to supply?

Can we just use a `frozenset` here instead?

Better yet, why not just use a sentinel for the candidate name (ie, `arrangement = None`) and then check that?

nucypher/blockchain/eth/actors.py
```diff
@@ -1202,6 +1206,26 @@ def missing_commitments(self) -> int:
...
+            if not members:
+                raise Policy.Unknown(f'{hrac.hex()} is not a published policy.')
+            raise Policy.Unpaid(f"{hrac.hex()} is unpaid.")
+
+    def verify_active_policy(self, hrac: bytes) -> None:
+        policy = self.policy_agent.fetch_policy(policy_id=hrac)
+        if policy.disabled:
+            raise Policy.Inactive(f'{hrac.hex()} is a disabled policy.')
+        expired = datetime.utcnow() >= datetime.utcfromtimestamp(policy.end_timestamp)
+        if expired:
```
jMyles, line 1226: Does this mean that if a policy is disabled, and then subsequently the expiration date passes, we'll get a different error?  Is that what we want?

nucypher/characters/control/interfaces.py
jMyles: Hmm, yeah, I'm also not sure.  It looks like it was added when @michwill crafted the 'bring your own map' logic, but then we never went the extra step of making any of our controllers pass it.

nucypher/policy/orders.py
```diff
@@ -0,0 +1,282 @@
...
+        # keys in partial and complete results are guaranteed to be distinct
+        results = {capsule: dict(ursulas_and_cfrags) for capsule, ursulas_and_cfrags in self._partial_results.items()}
+        results.update({capsule: dict(ursulas_and_cfrags) for capsule, ursulas_and_cfrags in self._complete_results.items()})
+        return results
+
+
+class WorkOrder:
+
+    def __init__(self, ursula_address, capsules):
+        self.ursula_address = ursula_address
```
jMyles, line 159: Why do we only include `ursula_address` now instead of the entire Ursula?  This makes it much more possible to send a WorkOrder to a wrong Ursula, no?

nucypher/blockchain/eth/actors.py
fjarri: Not actually added in this PR, it's from `chili-con-carne` on which this PR is based

nucypher/policy/orders.py
fjarri: I don't see how, an address identifies an Ursula just as well as an Ursula object.

nucypher/characters/control/specifications/fields/messagekit.py
```diff
@@ -40,6 +40,12 @@ def _deserialize(self, value, attr, data, **kwargs):
...
     def _validate(self, value):
         try:
-            PolicyMessageKitClass.from_bytes(value)
+            MessageKitClass.from_bytes(value)
+        # FIXME: remove when https://github.com/nucypher/bytestringSplitter/pull/40 is fixed.
+        except AttributeError as e:
+            if "'BytestringSplitter' object has no attribute '__name__'" in str(e):
+                raise InvalidInputData(f"Could not parse {self.name}: {e}")
+            else:
+                raise
```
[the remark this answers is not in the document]
derekpierre: Also, this is one of the classes I want to take a 2nd look at a part of https://github.com/nucypher/nucypher/issues/2714.

nucypher/utilities/porter/control/specifications/fields/workorder.py
```diff
@@ -16,13 +16,13 @@
...
 from nucypher.control.specifications.fields import Base64BytesRepresentation
-from nucypher.policy.orders import WorkOrder as WorkOrderClass
+from nucypher.policy.orders import ReencryptionRequest as ReencryptionRequestClass
 
 
-class WorkOrder(Base64BytesRepresentation):
-    def _serialize(self, value: WorkOrderClass, attr, obj, **kwargs):
-        return super()._serialize(value.payload(), attr, obj, **kwargs)
+class ReencryptionRequest(Base64BytesRepresentation):
+    def _serialize(self, value: ReencryptionRequestClass, attr, obj, **kwargs):
```
derekpierre, line 23: Don't need to define a _serialize function if `bytes(value)` can be used for bytes representation since this is what the super class uses. WorkOrder needed to override the _serialize function because the `.payload()` function was needed for the corresponding bytes representation.

nucypher/characters/control/interfaces.py
derekpierre: Fixed in #2768.

derekpierre (commented):
Good stuff! Only comment is to add text to the newsfragment so that it isn't empty.

derekpierre:
PS is this ready to come off of "Draft"?

fjarri:
It came off WIP, I'm just keeping it as draft for the sake of merge protection.

nucypher/policy/kits.py
```diff
@@ -0,0 +1,247 @@
...
+        self.queried_addresses = queried_addresses
+
+    def __bytes__(self):
+        return (bytes(self.capsule) +
+                b''.join(to_canonical_address(address) for address in self.queried_addresses))
+
+    @classmethod
+    def from_bytes(cls, data: bytes):
+        capsule, remainder = capsule_splitter(data, return_remainder=True)
+        addresses_as_bytes = checksum_address_splitter.repeat(remainder)
```
derekpierre, line 159: If it is possible for `queried_addresses` to be an empty list, this line will fail recreating such a `RetrievalKit` from its bytes. Either we prevent empty lists of addresses or handle them.

```python
In [7]: retrieval_kit = RetrievalKit(capsule, [])

In [8]: retrieval_kit_bytes = bytes(retrieval_kit)

In [9]: retrieval_kit_bytes
Out[9]: b'\x02%\x01\xb8f\xd2I\x05\x1a\xb8\xa2\xeb_l\rv\xca\x94\x14\xec\xbaoE\xd7\x0c\xbf\x80\x91_\xb9V\xb5\x93\x03\xd9\x1c\xd7\xa18\t\x11\x0b\xbb\xb3\x06\x0b\x13\x94\xc9\x0b\xe5\xbc\xfc\x1d\xb1X\x8d9#Vi\xe9\xa0\xee\xc2\x85\xfcl\x1f\x9d\x8a\xdd^\xfe\x04`\x9c,*\xb8\xc0\xb1\x8b8\xda[\xe7\x08\x99a\xa3\xf4>\x11\xfc\xbb!\xa1'

In [10]: r = RetrievalKit.from_bytes(retrieval_kit_bytes)
BytestringSplittingError                  Traceback (most recent call last)
<ipython-input-10-b61f49bde05f> in <module>
----> 1 r = RetrievalKit.from_bytes(retrieval_kit_bytes)

~/Documents/Github/repos/forks/derek/nucypher/nucypher/policy/kits.py in from_bytes(cls, data)
    157     def from_bytes(cls, data: bytes):
    158         capsule, remainder = capsule_splitter(data, return_remainder=True)
--> 159         addresses_as_bytes = checksum_address_splitter.repeat(remainder)
    160         return cls(capsule, [to_checksum_address(address) for address in addresses_as_bytes])
...
BytestringSplittingError: Not enough bytes to constitute message types bytes:20 - need 20, got 0
```

nucypher/policy/orders.py
```diff
@@ -38,310 +38,260 @@
...
-        self.hrac = hrac
-        self.ursula = ursula
-        self.encrypted_kfrag = encrypted_kfrag
-        self.publisher_verifying_key = publisher_verifying_key
-        self.alice_verifying_key = alice_verifying_key
-        self.tasks = tasks
-        self.receipt_signature = receipt_signature  # not a blockchain receipt
-        self.completed = False
+        # Records the addresses of Ursulas that were already queried, indexed by capsule.
+        self._queried_addresses = {retrieval_kit.capsule: retrieval_kit.queried_addresses
```
derekpierre, line 87: Side effect of changing `RetrievalKit.queried_addresses` from a `Set` to `List` causes this line to have subsequent effects on:
1. L102 where we try to do `ursulas_to_contact_last |= queried_addresses` but would have to do `ursulas_to_contact_last |= set(queried_addresses)` instead
2. L136 where we try to do `self._queried_addresses[capsule].add(work_order.ursula_address)` but would have to do `self._queried_addresses[capsule].append(work_order.ursula_address)` instead.

Some of my Porter tests in #2768 are now failing because of it.
Either we use a set here:`self._queried_addresses = {retrieval_kit.capsule: set(retrieval_kit.queried_addresses)`, or we make the above changes.

fjarri: Good catch, the last is especially bed since I'm mutating the externally passed objects. Fixed.

nucypher/policy/kits.py
fjarri: Good point, fixed.

nucypher/policy/kits.py
```diff
@@ -0,0 +1,250 @@
...
+    All the components needed to transmit and verify an encrypted message.
+    """
+
+    @classmethod
+    def author(cls,
+               recipient_key: PublicKey,
+               plaintext: bytes,
+               signer: 'SignatureStamp',
+               sign_plaintext: bool = True
+               ) -> Tuple['MessageKit', Signature]:
```
derekpierre, line 49:
```suggestion
               ) -> 'MessageKit':
```

nucypher/policy/kits.py
```diff
@@ -0,0 +1,250 @@
...
+                capsule, ciphertext = umbral.encrypt(recipient_key, sig_header + plaintext)
+                signature = signer(ciphertext)
+                signature_in_kit = signature
+            message_kit = cls(ciphertext=ciphertext,
+                              capsule=capsule,
+                              sender_verifying_key=signer.as_umbral_pubkey(),
+                              signature=signature_in_kit)
+        else:
+            # Don't sign.
+            signature = sig_header = NOT_SIGNED
```
derekpierre, line 70:
```suggestion
            sig_header = NOT_SIGNED
```

nucypher/policy/kits.py
```diff
@@ -0,0 +1,250 @@
...
+            capsule, ciphertext = umbral.encrypt(recipient_pubkey_enc, sig_header + plaintext)
```
derekpierre, line 71: Doesn't seem like this line gets tested - ideally a test should have failed.

```suggestion
            capsule, ciphertext = umbral.encrypt(recipient_key, sig_header + plaintext)
```

nucypher/characters/lawful.py
```diff
@@ -1874,13 +1852,12 @@ def __init__(self,
         if is_me:
             self.log.info(self.banner.format(policy_encrypting_key))
 
-    def encrypt_message(self, plaintext: bytes) -> Tuple[UmbralMessageKit, Signature]:
+    def encrypt_message(self, plaintext: bytes) -> Tuple[MessageKit, Signature]:
```
derekpierre, line 1855:
```suggestion
    def encrypt_message(self, plaintext: bytes) -> MessageKit:
```

nucypher/policy/maps.py
```diff
@@ -54,7 +53,7 @@ class NowhereToBeFound(RestMiddleware.NotFound):
 
     ursula_and_kfrag_payload_splitter = BytestringSplitter(
         (to_checksum_address, ETH_ADDRESS_BYTE_LENGTH),
-        (UmbralMessageKit, VariableLengthBytestring),
+        (MessageKit, VariableLengthBytestring),
     )
 
     @classmethod
```
derekpierre, line 59: `bob` parameter on L63 is no longer used and can be removed - was introduced as part of #2773.

KPrasch (approved):
Thanks for the detailed walkthrough!  Excellent (re)work.

nucypher/policy/orders.py
```diff
@@ -38,310 +38,261 @@
...
+        return cls(cfrags, signature)
+
+    def __init__(self, cfrags: List[CapsuleFrag], signature: Signature):
+        self.cfrags = cfrags
+        self.signature = signature
+
+    @classmethod
+    def from_bytes(cls, data: bytes):
+        signature, cfrags_bytes = signature_splitter(data, return_remainder=True)
+        cfrags = cfrag_splitter.repeat(cfrags_bytes)
```
derekpierre, line 290: Do we need to check whether `cfrags_bytes`  is not None first? Similar to what we did for `RetrievalKit`?

tests/acceptance/characters/test_freerider_attacks.py
```diff
@@ -127,12 +127,12 @@ def test_pay_a_flunky_instead_of_the_arranged_ursula(blockchain_alice,
     # Enrico becomes
     enrico = Enrico(policy_encrypting_key=bupkiss_policy.public_key)
     plaintext = b"A crafty campaign"
-    message_kit, _signature = enrico.encrypt_message(plaintext)
+    message_kit = enrico.encrypt_message(plaintext)
 
-    with pytest.raises(RestMiddleware.PaymentRequired):
-        blockchain_bob.retrieve(message_kit,
+    with pytest.raises(Ursula.NotEnoughUrsulas):
```
derekpierre, line 132: Based on the name of this test, how did the exception change?

nucypher/characters/lawful.py
```diff
@@ -996,6 +835,153 @@ def retrieve():
...
+            raise RuntimeError(message) from e
+        except middleware.UnexpectedResponse:
+            raise # TODO: Handle this
+
+        try:
+            reencryption_response = ReencryptionResponse.from_bytes(response.content)
+        except Exception as e:
+            message = f"Ursula ({ursula}) returned an invalid response: {e}."
+            self.log.warn(message)
+            raise RuntimeError(message)
```
derekpierre, line 886:
```suggestion
            raise RuntimeError(message) from e
```

nucypher/characters/lawful.py
```diff
@@ -996,6 +835,153 @@ def retrieve():
...
+        middleware = self._learner.network_middleware
+
+        try:
+            response = middleware.reencrypt(ursula, bytes(reencryption_request))
+        except NodeSeemsToBeDown as e:
+            # TODO: What to do here?  Ursula isn't supposed to be down.  NRN
+            message = (f"Ursula ({ursula}) seems to be down "
+                       f"while trying to complete ReencryptionRequest: {reencryption_request}")
+            self.log.info(message)
+            raise RuntimeError(message) from e
```
derekpierre, line 868: Any reason we don't want to just reraise the original exception instead of using RuntimeException?

docs/source/application_development/getting_started.rst
```diff
@@ -302,7 +302,7 @@ Encrypt
    from nucypher.characters.lawful import Enrico
 
    enrico = Enrico(policy_encrypting_key=policy_encrypting_key)
-   ciphertext, signature = enrico.encrypt_message(plaintext=b'Peace at dawn.')
+   ciphertext = enrico.encrypt_message(plaintext=b'Peace at dawn.')
```
jMyles, line 305: The idea here is that the user can get the signature from the `MessageKit` if they want to, right?  Why are we calling it `ciphertext` in the docs?

fjarri: Yes, signature is either in the plaintext (if we signed the plaintext) or kept inside the message kit (if we signed the ciphertext). Now that I think about it, why can't we keep the signature outside in both cases?

> Why are we calling it ciphertext in the docs?

Honestly, I don't know. Probably some historical reasons.

nucypher/policy/orders.py
fjarri: A response with zero cfrags makes no sense because we won't ever make a request with zero capsules. But I guess it can be checked for a more informative error.

nucypher/characters/lawful.py
derekpierre: ^ Should we include the underlying exception here - for ease of debugging?

derekpierre (dismissed):
🎸

nucypher/characters/lawful.py
derekpierre: nvm - it is caught and ignored anyway

nucypher/policy/orders.py
fjarri: Fixed.

tests/acceptance/characters/test_freerider_attacks.py
fjarri: Filed #2789 - possibly for another PR? Note though that if we decide to save errors in `RetrievalResult`, this will change the protocol.

docs/source/application_development/getting_started.rst
fjarri: Anyway, changed to `message_kit`.

nucypher/characters/lawful.py
```diff
@@ -615,7 +615,7 @@ def decryptor(ciphertext):
 
         return encrypted_treasure_map.decrypt(decryptor)
 
-    def retrieve_cfrags(
+    def try_retrieve(
```
derekpierre, line 618: This function name is pretty vague. 

Perhaps something more explicit like `retrieve_without_decrypt`? or `retrieve_cfrags_for_kits` or `get_cfrags_for_kits`...?

nucypher/characters/control/specifications/bob.py
```diff
@@ -45,15 +41,15 @@ class Retrieve(BaseSchema):
...
+    message_kit = character_fields.MessageKit(
         required=True,
         load_only=True,
         click=options.option_message_kit(required=False)
     )
 
-    treasure_map = EncryptedTreasureMap(required=False,
-                                        load_only=True,
-                                        click=options.option_treasure_map)
+    encrypted_treasure_map = EncryptedTreasureMap(required=False,
```
derekpierre, line 50: This is required now, no? If so, then we need to update the `requred` parameter
 
```suggestion
    encrypted_treasure_map = EncryptedTreasureMap(required=True,
```

**AND**

 make the `option_treasure_map` be required.

examples/finnegans_wake_demo/finnegans-wake-demo-federated.py
```diff
@@ -140,7 +140,7 @@
     # single passage from James Joyce's Finnegan's Wake.
     # The matter of whether encryption makes the passage more or less readable
     # is left to the reader to determine.
-    single_passage_ciphertext, _signature = enrico.encrypt_message(plaintext)
+    single_passage_ciphertext = enrico.encrypt_message(plaintext)
```
derekpierre, line 143: did we want to rename this to `single_passage_message_kit`. Analogous to the change done in the getting started docs.

docs/source/application_development/http_character_control.rst
```diff
@@ -167,6 +167,6 @@ This endpoint controls the ``Enrico.encrypt_message`` method.
 - HTTP Method: ``POST``
 - Required arguments:
     - ``message`` -- encoded as base64
-- Returns: ``message_kit`` and ``signature`` encoded as base64
+- Returns: ``message_kit`` encoded as base64
```
derekpierre, line 170: Required arguments for `/retrieve` (L149)  need to remove `label` and include `encrypted_treasure_map`

examples/finnegans_wake_demo/finnegans-wake-demo-testnet.py
```diff
@@ -144,7 +144,7 @@
     # The matter of whether encryption makes the passage more or less readable
     # is left to the reader to determine.  Many data sources (Enricos) can
     # encrypt fot the policy's public key.
-    ciphertext, _signature = enrico.encrypt_message(plaintext)
+    ciphertext = enrico.encrypt_message(plaintext)
```
derekpierre, line 147: update this as well to be `message_kit` (like in getting started docs)? Same for L156.

nucypher/characters/control/specifications/bob.py
```diff
@@ -45,15 +41,15 @@ class Retrieve(BaseSchema):
             type=click.STRING,
             required=False)
     )
-    message_kit = character_fields.UmbralMessageKit(
+    message_kit = character_fields.MessageKit(
         required=True,
         load_only=True,
         click=options.option_message_kit(required=False)
     )
```
derekpierre, line 48: Are we going to keep the REST API to only one message kit for the character control, or a list of message kits? For `Bob.retrieve()`python API it takes a list of message kits now.

If desired we can do it in a separate PR if need be.

nucypher/characters/control/specifications/bob.py
```diff
@@ -45,15 +41,15 @@ class Retrieve(BaseSchema):
             type=click.STRING,
             required=False)
     )
-    message_kit = character_fields.UmbralMessageKit(
+    message_kit = character_fields.MessageKit(
         required=True,
         load_only=True,
         click=options.option_message_kit(required=False)
```
derekpierre, line 47: This should be required

```suggestion
        click=options.option_message_kit(required=True)
```

nucypher/characters/control/specifications/bob.py
fjarri: Technically, it did take a list before as well, it's just that it was spliced. That is, it was `retrieve(kit1, kit2, ...)`, now it's `retrieve([kit1, kit2, ...], ...)`. For Porter it probably makes sense to support batch retrieve.

derekpierre: Yep - but currently the Bob character control specification here for the `/retrieve` endpoint only accepts a single kit and not multiple - do we want it to only allow for 1 kit to be passed?

Porter uses a batch retrieve i.e. a list of retrieval kits can currently be provided - https://github.com/derekpierre/nucypher/blob/porter-retrieval/nucypher/utilities/porter/control/specifications/porter_schema.py#L135.

fjarri: Would it be better done as a part of #2768?

derekpierre: Sure - I can add it in

nucypher/characters/lawful.py
fjarri: Renamed to `retrieve()`/`retrieve_and_decrypt()`.

nucypher/network/retrieval.py
```diff
@@ -0,0 +1,410 @@
...
+        # How many nodes over the threshold we want to know (just in case)
+        redundancy = 0
+
+        # If we know enough to decrypt, we can proceed.
+        known_ursulas = ursulas_in_map & all_known_ursulas
+
+        if len(known_ursulas) - redundancy >= treasure_map.threshold:
+            return
+
+        allow_missing = max(len(unknown_ursulas) - redundancy, 0)
```
derekpierre, line 282: This calculation seems off or am I missing something? If you get here the `len(known_ursulas)` is not sufficient to make the threshold (L279).

Now, suppose shares are 8, and threshold is 7. Also suppose that `len(known_ursulas)` is 5 i.e. not met the threshold. Therefore, `len(unknown_ursulas) = shares - len(known_ursulas) = 8 - 5 = 3`

Therefore `allow_missing` gets set to `max(3 - 0, 0) = 3`, whereas technically it should be 1 because out of the 3 unknown we need 2 to meet the threshold, and it's ok if we don't learn about 1 i.e. `len(unknown_ursulas) - (threshold - len(known_ursulas))` = 3 - (7 - 5) = 1

nucypher/network/retrieval.py
```diff
@@ -0,0 +1,410 @@
...
+                       f"while trying to complete ReencryptionRequest: {reencryption_request}")
+            self.log.info(message)
+            raise RuntimeError(message) from e
+        except middleware.NotFound as e:
+            # This Ursula claims not to have a matching KFrag.  Maybe this has been revoked?
+            # TODO: What's the thing to do here?
+            # Do we want to track these Ursulas in some way in case they're lying?  #567
+            message = (f"Ursula ({ursula}) claims not to have the KFrag "
+                       f"to complete ReencryptionRequest: {reencryption_request}. "
+                       f"Has access been revoked?")
```
derekpierre, line 316: Since we are now sending the kfrags as part of the treasure map in the reencryption request this wording may need to be updated...?

newsfragments/2730.misc.rst
```diff
@@ -0,0 +1 @@
+Simplifies the retrieval protocol (see #259 for the discussion). `PolicyMessageKit` is renamed to `MessageKit`. `Bob.retrieve()` is renamed to `retrieve_and_decrypt()`, and its signature is simplified: it only requires the treasure map, Alice's verifying key, and the policy encrypting key. A lower-level `Bob.retrieve()` is added that does not decrypt, but only attempts to retrieve the capsule frags.
```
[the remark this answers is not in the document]
derekpierre:
```suggestion
Simplifies the retrieval protocol (see `#259 <https://github.com/nucypher/nucypher/issues/259>`_ for the discussion). `PolicyMessageKit` is renamed to `MessageKit`. `Bob.retrieve()` is renamed to `retrieve_and_decrypt()`, and its signature is simplified: it only requires the treasure map, Alice's verifying key, and the policy encrypting key. A lower-level `Bob.retrieve()` is added that does not decrypt, but only attempts to retrieve the capsule frags.
```

nucypher/network/retrieval.py
```diff
@@ -0,0 +1,410 @@
...
+        ursulas_in_map = treasure_map.destinations.keys()
+        all_known_ursulas = self._learner.known_nodes.addresses()
+
+        # Push all unknown Ursulas from the map in the queue for learning
+        unknown_ursulas = ursulas_in_map - all_known_ursulas
+        if unknown_ursulas:
+            self._learner.learn_about_specific_nodes(unknown_ursulas)
+
+        # How many nodes over the threshold we want to know (just in case)
+        redundancy = 0
```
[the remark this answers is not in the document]
fjarri: It was kind of an "extended TODO", illustrating that we can take redundancy into account. Removed for now, it just creates complications.

nucypher/network/retrieval.py
fjarri: You are right, my bad. Adjusted the formula.

nucypher/characters/control/specifications/bob.py
```diff
@@ -24,36 +24,32 @@
...
-        click=options.option_message_kit(required=False)
+        click=options.option_message_kit(required=True)
     )
 
-    treasure_map = EncryptedTreasureMap(required=False,
-                                        load_only=True,
-                                        click=options.option_treasure_map)
+    encrypted_treasure_map = EncryptedTreasureMap(required=True,
+                                                  load_only=True,
+                                                  click=options.option_treasure_map)
```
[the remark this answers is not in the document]
fjarri: It's only used in one place, so I just made it required by default.

nucypher/network/retrieval.py
fjarri: Yes, now we get `NotFound` if policy HRAC is not found. Adjusted the message.

nucypher/network/retrieval.py
```diff
@@ -0,0 +1,409 @@
...
+        # If we know enough to decrypt, we can proceed.
+        known_ursulas = ursulas_in_map & all_known_ursulas
+        if len(known_ursulas) >= treasure_map.threshold:
+            return
+
+        # | <--- shares                                            ---> |
+        # | <--- threshold               ---> |
+        # | <--- known_ursulas ---> |         | <--- allow_missing ---> |
+        # |                         | <--- unknown_ursulas         ---> |
+        allow_missing = len(treasure_map.destinations) - treasure_map.threshold
```
derekpierre, line 282: I think we might still be off here. 

The `allow_missing` value in this case represents "from the list of unknown ursulas, how many ursulas do we want to allow to not be learned about". That would be dependent on how many ursulas we know about already relative to the threshold.

If we didn't know about any ursulas, this calc would be correct, but if we already know about some ursulas, the value needs to be adjusted for that; basically, at this point in the code, we are short of the threshold by `(threshold - len(known_ursulas))` ursulas.

```suggestion
        allow_missing = len(unknown_ursulas) - (treasure_map.threshold - len(known_ursulas))
```

Alternatively, I think (?) you can skip all of this logic for known vs unknown ursulas and just pass the entire treasure map destinations since there is code in `block_until_specific_nodes_are_known()` to account for already known nodes anyway, and then use allow_missing=(len(destinations) - threshold)  i.e. something like

```python
self._learner.block_until_specific_nodes_are_known(
    addresses=treasure_map.destinations,
    timeout=timeout,
    allow_missing=(len(treasure_map.destinations) - treasure_map.threshold))
```

fjarri: > `len(unknown_ursulas) - (treasure_map.threshold - len(known_ursulas))`

Well, that amounts to the same thing, since `len(unknown_ursulas) + len(known_ursulas) == len(treasure_map.destinations)`

> I think (?) you can skip all of this logic for known vs unknown ursulas and just pass the entire treasure map destinations

Yes, that'll probably be simpler. The current calculation sort of evolved from nonzero redundancy which is now gone.

derekpierre: > Well, that amounts to the same thing, since len(unknown_ursulas) + len(known_ursulas) == len(treasure_map.destinations)

You are right - that was less clear to me on first look 😊

derekpierre (approved):
🎸 - great work @fjarri !

nucypher/characters/lawful.py
```diff
@@ -554,404 +541,97 @@ def get_card(self) -> 'Card':
...
+        # Normalize input
+        message_kits: List[PolicyMessageKit] = [
+            message_kit.as_policy_kit(policy_encrypting_key, treasure_map.threshold)
+                if isinstance(message_kit, MessageKit) else message_kit
+            for message_kit in message_kits
+            ]
+
+        # Clear up all unrelated information from message kits before retrieval.
+        # This call may go to Porter, and we don't want to send it the ciphertexts
+        # which it cannot decrypt anyway.
```
derekpierre, line 596: Just did a rebase for #2768, and noticed this comment - presumably, this comment can be removed.

nucypher/datastore/queries.py
```diff
@@ -53,16 +53,9 @@ def find_expired_policies(ds: Datastore, cutoff: maya.MayaDT) -> DatastoreQueryR
...
-    return ds.query_by(EncryptedTreasureMap,
-                       filter_field='expiration',
-                       filter_func=lambda expiration: expiration <= cutoff,
-                       writeable=True)
-
-
 @unwrap_records
-def get_work_orders(ds: Datastore) -> List[Workorder]:
-    return ds.query_by(Workorder)
+def get_reencryption_requests(ds: Datastore) -> List[ReencryptionRequest]:
```
derekpierre, line 57: The prometheus call to this function needs to be updated - https://github.com/fjarri/nucypher/blob/retrieval-protocol/nucypher/utilities/prometheus/collector.py#L131. Only the call itself, **not** the metric key string called "work_orders_gauge".

fjarri: Why not the metric too? Backward compatibility?

derekpierre: Yep, backward compatibility for the dashboard.

fjarri: Is "work orders" set in stone there, or will we be able to change it eventually? It creates kind of a confusing situation.

derekpierre: Linked to the wrong line (associated but not the exact line), it should be https://github.com/fjarri/nucypher/blob/retrieval-protocol/nucypher/utilities/prometheus/collector.py#L101 i.e. `ursula_work_orders` field.

It's not set in stone. It's more that the Grafana dashboard configuration file would need to be updated as well - https://github.com/p2p-org/nucypher-monitoring/blob/master/grafana/provisioning/dashboards/nucypher.json#L1820. We'll need to coordinate with p2p to update that file - separate from that, we could think about whether we want to distribute that file separate from p2p.

Unsure if @KPrasch has already given this some thought - but I don't think it is necessary to resolve at this moment.

fjarri: Filed #2797 and added a TODO in the code

nucypher/crypto/splitters.py
```diff
@@ -25,3 +25,4 @@
 capsule_splitter = BytestringSplitter((Capsule, Capsule.serialized_size()))
 cfrag_splitter = BytestringSplitter((CapsuleFrag, CapsuleFrag.serialized_size()))
 kfrag_splitter = BytestringSplitter((KeyFrag, KeyFrag.serialized_size()))
+checksum_address_splitter = BytestringSplitter((bytes, 20)) # TODO: is there a pre-defined constant?
```
piotr-roslaniec, line 28: I don't think there is one, not even in `eth_utils` or `eth_typings`

nucypher/policy/orders.py
jMyles: The full `Ursula` has its `EncryptingPower` (making it possible only to encrypt for that Ursula) and ships with its server address.

fjarri: But we don't need that power in this context. Everything that had to be encrypted for the Ursula (kfrag) is already encrypted.

nucypher/policy/kits.py
```diff
@@ -0,0 +1,249 @@
...
+
+    @classmethod
+    def from_bytes(cls, data: bytes):
+        splitter = BytestringSplitter(
+            capsule_splitter,
+            (bytes, 1))
+
+        capsule, signature_flag, remainder = splitter(data, return_remainder=True)
+
+        if signature_flag == b'\x00':
```
piotr-roslaniec, line 114: Nitpick: Could `b'\x00'` and `b'\x01'` be translated into constants, for example: `NO_SIG`, `HAS_SIG`?

piotr-roslaniec (approved):
LGTM :tada:  Great work!

KPrasch (approved):
Thanks for the walkthrough in-person.  This PR represents the introduction of many new internal objects and altered system metaphors, it will take me a while to get used to this version of the codebase.  Overall I think we're introducing some complexity to gain immutability and reduce the number of invalid instance states (as far as I understand this is precisely your goal). 👍🏻

nucypher/policy/kits.py
fjarri: Fair point, but this code might prove to be unnecessary depending on the decision on #2743; otherwise it is better fixed in BSS (as a generic support for Optional objects). Added TODOs.

PR #2721, opened 2021-06-10
Initial Porter Documentation

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [X] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Initial documentation of Porter. Based over #2726  .

**Issues fixed/closed:**
Related to #2670 .

**Notes for reviewers:**
Build artifact available here - https://136953-100446516-gh.circle-artifacts.com/0/docs/build/html/application_development/porter.html.

This is an initial writeup of already implemented functionality up to "basic authentication" (#2726 ). `/revoke` and `/exec_work_order` APIs haven't been implemented as yet.

I would like to also add some application-specific documentation about specifics of using Porter aside from just stating the API. I intend to add this in separate PRs.

One mistake I made is doing documentation as a separate task, instead of providing a simple base document and then adding to the document as part of implementation PRs so that it builds up over time. To rectify that I'll merge this PR now as an initial base, and then add to it over time in subsequent PRs moving forward.

PR #1424, opened 2019-10-19
[WIP] CLI demo

cygnusv:
* Added support for `--rate`

PR #2732, opened 2021-06-30
Use default/dynamic active staker pagination size and standardize Porter execution timeouts

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Based over #2721 (Initial Porter Documentation), so only the latest commit is additional work.

Default/Dynamic Pagination Size
---------------------------------
When attempting to get active stakers from StakingEscrow as part of sampling logic, this eth_call can be quite heavy due to network size, and can sometimes timeout
```
In [51]: r = requests.get("http://127.0.0.1:9155/get_ursulas", json=payload)

In [52]: r
Out[52]: <Response [500]>

In [53]: r.text
Out[53]: "{'code': -32000, 'message': 'execution aborted (timeout = 5s)'}"
```
This causes Porter to sometimes fail when attempting to sample ursulas. The timeout occurs intermittently, probably because of some caching done by the underlying eth node. In my testing, I've been using Infura which uses Geth. Geth has a hard-coded timeout of 5s for contract calls - https://github.com/ethereum/go-ethereum/blob/master/internal/ethapi/api.go#L950.

I assume that since Porter and Alice use the same code, and since Alice doesn't specify a pagination size either, Alice will encounter this same problem. So instead of just fixing it for Porter, I made a generalized default solution.

By default, the pagination size used is 0 (unlimited) if no pagination size is specified which is currently the case for both Porter and Alice when creating the staker reservoir. Therefore the call attempts to obtain all active stakers, which is a heavy call and therefore sometimes times out. Instead, I propose that we use a default size if no pagination value (i.e. `pagination_size` is `None`) is provided to the API - 0 will still hold if explicitly provided as the pagination size. 

The value of this size is tricky and can vary by node types (geth etc.), and whether the node is a light node or not. Consequently, the pagination size for regular and light nodes have default values that can be overridden via environment variables (1000 for regular node, 30 for light node). The environment variable provides flexibility for us not to have to determine the "correct" value for a variety of scenarios.

Additionally, to provide some dynamism, if there is a failure in getting the active stakers based on the magnitude of the pagination size used, the size is reduced by half and then retried - process repeated for a maximum of 3 total attempts. 

For example, if the pagination size value used is 1000
- try pagination size of 1000 (attempt 1.)
- if 1000 fails, retry with pagination size 500 (attempt 2.)
- if 500 fails, retry with pagination size 250 (attempt 3.)

if attempt 3 fails, then raise failure exception. Note that this sequence will be attempted for any non-zero pagination size provided, including defaults.

Porter Standardized Execution Timeout
---------------------------------------
Standardized execution timeout used by Porter. All executions that take a timeout will use this default timeout (10s). It felt weird some executions using different values than others.

**Issues fixed/closed:**
Related to https://github.com/nucypher/nucypher/pull/1424.

**Notes for reviewers:**
- Thoughts about the environment variable approach?
- Thoughts about how dynamic to make the retries and pagination size? - I went for simplicity
- Thoughts on Porter default execution timeout, as opposed to using default timeout values where available.


nucypher/blockchain/eth/agents.py
```diff
@@ -319,15 +327,36 @@ def get_all_active_stakers(self, periods: int, pagination_size: Optional[int] =
...
-                n_tokens = n_tokens + temp_locked_tokens
-                stakers.update(temp_stakers_map)
-                start_index += pagination_size
+                try:
+                    attempts += 1
+                    active_stakers = self.contract.functions.getActiveStakers(periods,
+                                                                              start_index,
+                                                                              pagination_size).call()
+                except Exception as e:
+                    if 'timeout = 5s' not in str(e):
```
KPrasch, line 338: This seems a bit fragile, since it depends on the exception containing this string exactly.  Not an RFC, just commentary.

nucypher/blockchain/eth/agents.py
```diff
@@ -250,7 +254,10 @@ class StakingEscrowAgent(EthereumContractAgent):
         'finishUpgrade'
     )
 
-    DEFAULT_PAGINATION_SIZE: int = 30    # TODO: Use dynamic pagination size (see #1424)
+    DEFAULT_STAKER_PAGINATION_SIZE_LIGHT_NODE: int = int(os.environ.get(
```
KPrasch, line 257: Why the cast to `int` here?

nucypher/blockchain/eth/agents.py
derekpierre: I hear ya - feels like we can do one of the following:
- simply check for the word "timeout" instead of including the length (it is currently a geth setting that can't be set)
OR
- don't perform any specific check on the text, and assume that any exception is caused by the timeout (i.e. remove the "if" check')
OR
...?

nucypher/blockchain/eth/agents.py
derekpierre: `os.enrion.get(...)` always returns a string.

KPrasch:
> Thoughts about the environment variable approach?

Seems like a good approach for a minimal configuration.  Let's not overthink it for now until we start getting some actual usage.

> Thoughts about how dynamic to make the retries and pagination size? - I went for simplicity

Simplicity is the best principal at this stage of Porter's development 👍🏻 

> Thoughts on Porter default execution timeout, as opposed to using default timeout values where available.

Same s above.  Use a high timeout by default, ans let usage dictate optimization.

KPrasch (approved):

vzotova (approved):

PR #2734, opened 2021-07-03
Fix TreasureMap splitter dict population bug

fjarri:
- Fix the issue with `__new__` not being called quickly enough and not populating the `_splitters` objects.
- Fix the issue with POLICY_ID_LENGTH in `estimate_gas`


KPrasch:
This PR can be understood as a temporary workaround until a more comprehensive versioning solution is implemented.

KPrasch (approved):
🤠

PR #2742, opened 2021-07-09
Bring your own key material

### LINKED ISSUE #2655: Bring Your Own Keys
Similarly to what causes the problem on #2654, we used a monolithic approach for designing NuCypher keyring, without taking into consideration that Alice, Bob, Ursula or any other new character in town, can have different requirements and expectations when it comes to their key custody.

In the case of Alice, and other user-side characters, it's reasonable to allow them to bring their own keys. What that means is still to be decided: is it "importing" keys from a different context? or is it allowing some sort of external signer/encryptor to interact with the nucypher logic?

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
- Introduces `Keystore.import_secure` method for user-supplied entropy blobs for usage as keystore secrets.
- Introduces `nucypher ursula|alice|bob init --key-material xxx` to accept a user-supplied hex entropy blob to import.

**Issues fixed/closed:**
- Fixes #2655 

**Why it's needed:**
- Allow additional flexibility and portability of keystores.

**Notes for reviewers**
- Is supporting the import of the secret key material enough portability for browser wallet interoperability or do we also need to support importing keys for individual powers?

### DIFF
--- a/nucypher/cli/commands/ursula.py
+++ b/nucypher/cli/commands/ursula.py
@@ -161,7 +161,7 @@ class UrsulaConfigOptions:
                 # TODO: Exit codes (not only for this, but for other exceptions)
                 return click.get_current_context().exit(1)
 
-    def generate_config(self, emitter, config_root, force, secret_key):
+    def generate_config(self, emitter, config_root, force, key_material):
 
         if self.dev:
             raise RuntimeError('Persistent configurations cannot be created in development mode.')
@@ -180,7 +180,7 @@ class UrsulaConfigOptions:
             self.rest_host = collect_worker_ip_address(emitter, network=self.domain, force=force)
 
         return UrsulaConfiguration.generate(password=get_nucypher_password(emitter=emitter, confirm=True),
-                                            secret_key=bytes.fromhex(secret_key),
+                                            key_material=bytes.fromhex(key_material) if key_material else None,
                                             config_root=config_root,
                                             rest_host=self.rest_host,
                                             rest_port=self.rest_port,
@@ -296,8 +296,8 @@ def ursula():
 @option_force
 @option_config_root
 @group_general_config
-@click.option('--secret-key', help="An custom pre-secured secret hex blob to use for key derivations", type=click.STRING)
-def init(general_config, config_options, force, config_root, secret_key):
+@click.option('--key-material', help="An custom pre-secured secret hex blob to use for key derivations", type=click.STRING)
+def init(general_config, config_options, force, config_root, key_material):
     """Create a new Ursula node configuration."""
     emitter = setup_emitter(general_config, config_options.worker_address)
     _pre_launch_warnings(emitter, dev=None, force=force)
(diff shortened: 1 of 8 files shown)


nucypher/cli/commands/ursula.py
```diff
@@ -295,7 +296,8 @@ def ursula():
 @option_force
 @option_config_root
 @group_general_config
-def init(general_config, config_options, force, config_root):
+@click.option('--secret-key', help="An custom pre-secured secret hex blob to use for key derivations", type=click.STRING)
```
fjarri, line 299: I have reservations about calling it `secret_key`, seeing how we already have `SecretKey` objects, and that's not it. "Key seed" or "key material" perhaps?

KPrasch: ✅

tests/unit/crypto/test_keystore.py
```diff
@@ -241,6 +241,33 @@ def test_restore_keystore_from_mnemonic(tmpdir, mocker):
     assert keystore._Keystore__secret == secret
 
 
+def test_import_custom_keystore(tmpdir):
+
+    # Too short - 32 bytes is required
+    custom_secret = b'tooshort'
+    with pytest.raises(ValueError, match="Entropy must be at least 32 bytes."):
```
fjarri, line 248: Shouldn't we also raise an error if it's more than 32 bytes? Currently any size from 32 to 64 bytes will work, and for anything over 64 an error from the depths of Umbral will be raised.

nucypher/cli/options.py
```diff
@@ -48,6 +48,7 @@
 option_force = click.option('--force', help="Don't ask for confirmation", is_flag=True)
 option_gas_price = click.option('--gas-price', help="Set a static gas price (in GWEI)", type=GWEI)
 option_gas_strategy = click.option('--gas-strategy', help="Operate with a specified gas price strategy", type=click.STRING)  # TODO: GAS_STRATEGY_CHOICES
+option_key_material = click.option('--key-material', help="An custom pre-secured secret hex blob to use for key derivations", type=click.STRING)
```
derekpierre, line 51:
```suggestion
option_key_material = click.option('--key-material', help="A custom pre-secured secret hex to use for key derivations", type=click.STRING)
```

OR

seed?
```suggestion
option_key_material = click.option('--key-material', help="A custom pre-secured secret hex seed to use for key derivations", type=click.STRING)
```

derekpierre (approved):
🎸

tests/unit/crypto/test_keystore.py
KPrasch: ✔️

nucypher/cli/options.py
KPrasch: ✔️ - Made similar edits.

piotr-roslaniec (approved):

nucypher/cli/options.py
```diff
@@ -48,6 +48,7 @@
 option_force = click.option('--force', help="Don't ask for confirmation", is_flag=True)
 option_gas_price = click.option('--gas-price', help="Set a static gas price (in GWEI)", type=GWEI)
 option_gas_strategy = click.option('--gas-strategy', help="Operate with a specified gas price strategy", type=click.STRING)  # TODO: GAS_STRATEGY_CHOICES
+option_key_material = click.option('--key-material', help="A pre-secured hex-encoded secret to use for private key derivations", type=click.STRING)
```
cygnusv, line 51: Instead of directly inputting the key material here, an alternative is to pass a file path that contains the key material. The advantage here is not disclosing the key material from the shell history. Geth does something similar for passwords (see https://geth.ethereum.org/docs/interface/managing-your-accounts)

KPrasch: Great suggestion.  I'll make the changes.

vepkenez (approved):

Issue #2737, opened 2021-07-07
Unify Treasure Maps


KPrasch:
Presently we're maintaining two kinds of treasure maps: `TreasureMap` and `SignedTreasureMap`.  This stems from a difference in serialization formats due to different signing mechanisms.  In blockchain mode Alice signs the treasure map with her ethereum wallet so that it can be verified that it was authored by the same actor that published the policy on-chain.   An EIP-191 (65-byte) signature is applied which is different than the signatures that can be created using stamps in federated mode.  This results in the following patterns appearing in several places in the codebase:

```python
if federated:
    from nucypher.policy.maps import TreasureMap as MapClass
else:
    from nucypher.policy.maps import SignedTreasureMap as MapClass
``` 

Or this monster from interface validators:

```python
def _validate(self, value):
    try:
        # Unsigned TreasureMap (Federated)
        from nucypher.policy.maps import TreasureMap as UnsignedTreasureMap
        splitter = UnsignedTreasureMap.get_splitter(value)
        splitter(value)
    except InvalidNativeDataTypes:
        try:
            # Signed TreasureMap (Blockchain)
            from nucypher.policy.maps import SignedTreasureMap
            splitter = SignedTreasureMap.get_splitter(value)
            splitter(value)
        except InvalidNativeDataTypes as e:
            raise InvalidInputData(f"Could not parse {self.name}: {e}")
    return True
```

- Is treasure map signing by ethereum wallet a valid protocol security concern?  If not we can simply sign with stamps or not at all.
- Can we implement an EIP-191 compatible signature for federated mode?

The goal of this issue is to unify treasure maps into a single class or entity that can be used in both operating modes, if possible.

derekpierre:
Would this now be close via #2773 ?

fjarri:
Fixed in #2773

PR #2745, opened 2021-07-12
AuthorizedKeyFrag and assorted fixes

fjarri:
**Type of PR:**
- Other

**Required reviews:** 
- 2

**What this does:**
- moves deserialization out of `Bob.retrieve()` and into the respective interface
- renames remaining "alices" that should be "publishers" according to the new terminology
- wraps authorized kfrag logic into an `AuthorizedKeyFrag` class

See #2752 regarding the renaming. 

One thing not changed in this PR (should it be?) is that the `grant` interface returns only one verifying key (still named `alice_verifying_key`). Technically it should return both Alice's and publisher's key, but using web interface implies that Alice *is* the publisher, so perhaps it's ok?


nucypher/policy/maps.py
```diff
@@ -35,12 +35,63 @@
...
+        )
+
+    @staticmethod
+    def _kfrag_checksum(kfrag: Union[KeyFrag, VerifiedKeyFrag]) -> bytes:
+        # TODO: assuming here that KeyFrag and VerifiedKeyFrag have the same byte representation.
+        # It's a little hacky... Should we have `verified_kfrag.unverify()` or something?
+        return keccak_digest(bytes(kfrag))[:WRIT_CHECKSUM_SIZE]
+
+    @classmethod
+    def create(cls, hrac: bytes, kfrag: VerifiedKeyFrag, alice_stamp: SignatureStamp) -> 'AuthorizedKeyFrag':
```
KPrasch, line 60: Perhaps `from_verified_kfrag` adds a bit more semantic density here.

fjarri: Yes, that's a good variant. I didn't particularly like the name myself, it was just a placeholder. Another possibility is `construct_by_alice`.

fjarri: Or rather `construct_by_publisher`, I guess

nucypher/policy/maps.py
```diff
@@ -35,12 +35,63 @@
...
+        # The writ and the KFrag together represent a complete kfrag kit: the entirety of
+        # the material needed for Ursula to assuredly service this policy.
+        return cls(hrac, kfrag_checksum, writ_signature, kfrag)
+
+    def __init__(self, hrac: bytes, kfrag_checksum: bytes, writ_signature: Signature, kfrag: VerifiedKeyFrag):
+        self.hrac = hrac
+        self.kfrag_checksum = kfrag_checksum
+        self.writ = hrac + kfrag_checksum
+        self.writ_signature = writ_signature
+        self.kfrag = kfrag
```
derekpierre, line 77: Should we name this `verified_kfrag`?

It gets a bit confusing between Kfrag, VerifiedKFrag, AuthorizedKeyfrag...

nucypher/characters/control/specifications/bob.py
```diff
@@ -27,11 +27,11 @@ class JoinPolicy(BaseSchema):  #TODO:  this doesn't have a cli implementation
     label = fields.Label(
         load_only=True, required=True,
         click=options.option_label(required=True))
-    alice_verifying_key = fields.Key(
+    publisher_verifying_key = fields.Key(
```
derekpierre, line 30: Does the "Retreive" schema also need to be changed. Bob.retrieve(...) will call join policy as part of its operations. Currently the Retrieve schema only takes `alice_verifying_key`, but does it also need to take `publisher_verifying_key` since they can be two different keys?  See https://github.com/fjarri/nucypher/blob/chili-aftertaste/nucypher/characters/control/specifications/bob.py#L43 and https://github.com/fjarri/nucypher/blob/chili-aftertaste/nucypher/characters/control/interfaces.py#L222.

nucypher/characters/control/interfaces.py
```diff
@@ -220,6 +221,19 @@ def retrieve(self,
...
+            from nucypher.policy.maps import SignedTreasureMap as _MapClass
+
+        # TODO: This LBYL is ugly and fraught with danger.  NRN
+        if isinstance(treasure_map, bytes):
+            treasure_map = _MapClass.from_bytes(treasure_map)
+
+        if isinstance(treasure_map, str):
+            tmap_bytes = treasure_map.encode()
+            treasure_map = _MapClass.from_bytes(b64decode(tmap_bytes))
+
```
derekpierre, line 236: Ideally, this would be handled as part of the "TreasureMap" field deserialization. In Porter, I do something similar for validation, but I only use the splitter since I only care about the bytes representation that I then pass on to ursula(s) - https://github.com/nucypher/nucypher/pull/2664/files#diff-54f7a125485aa11f633996b09e5f6b0468b8b91cd06a04cb097a808b70e14c1eR59.

However, we can think about changing the field to always make the `_deserialize()` function return a TreasureMap/SignedTreasureMap object (instead of bytes) and then use the object directly in this case, or convert the object back to bytes for usage for Porter.

nucypher/policy/maps.py
fjarri: Actually at this point it can either be `KeyFrag` or `VerifiedKeyFrag`, depending on whether it `AuthorizedKeyFrag` was created by `from_bytes()` or `create()`. I guess we just need an un-verification in the latter. Not sure if a separate method is worth it, `KeyFrag.from_bytes(bytes(verified_kfrag))` will do the trick.

nucypher/characters/control/specifications/bob.py
fjarri: For now we can leave only `alice_verifying_key`, `publisher` can be added later without breaking compatibility. But yes, eventually we will want both.

nucypher/characters/control/interfaces.py
fjarri: The problem here is that in the current form it needs to know if the character is federated, and this is information is not available in `Field._deserialize()`. We really need to try and get rid of this dependency.

nucypher/policy/maps.py
KPrasch: I like `construct_by_publisher` too 👍🏻

nucypher/policy/maps.py
derekpierre: 🤯

nucypher/characters/control/interfaces.py
derekpierre: This was the motivation behind the following in the Porter EPIC, https://github.com/nucypher/nucypher/pull/2664/commits/1eda693733bcefbc970786e63afd366901d5a33c. You know the nature of the character (federated vs non-federated) when the interface is initiated so you can set a schema context.

Perhaps as a follow-up PR that already includes Porter changes.

KPrasch: The best solution in the long-term is to unify federated and unfederated treasure maps #2737 but this will require some additional footwork.

KPrasch (approved):
Approved with comments and one renaming RFC.

nucypher/characters/control/specifications/bob.py
derekpierre: Filed https://github.com/nucypher/nucypher/issues/2750.

nucypher/characters/control/interfaces.py
derekpierre: Filed #2751 as a pretty easy interim solution until #2737 can actually be done.

nucypher/policy/maps.py
derekpierre: I'm a little confused. Is the type hint of `VerifiedKeyFrag` wrong then...?

Is the 'authorized' dictated by the `writ`? and then it's up to whoever is using this AuthorizedKeyFrag to check whether the key frag included is verified or not? Missing a bit of the larger picture.

fjarri: Yes, as it is called now it should be `Union[KeyFrag, VerifiedKeyFrag]`. But if it actually is a VKF, it will cause problems later to whoever tries to verify it (if for some reason you created an AKF and try to verify without it passing through a serialization-deserialization), so it's better to just un-verify in `create()`.

derekpierre: Ok, so `AuthorizedKeyFrag` should only ever take a `KeyFrag` in `__init__` and never a `VerifiedKeyFrag` then. And to achieve that we should add `KeyFrag.from_bytes(bytes(verified_kfrag))` in the `create()` function.

fjarri: Added an un-verifiation step.

nucypher/policy/maps.py
fjarri: Renamed, along with remaining "alice" usage in `maps.py`

nucypher/policy/maps.py
derekpierre: 👍

nucypher/characters/control/interfaces.py
```diff
@@ -189,6 +191,19 @@ def retrieve(self,
 
         self.implementer.join_policy(label=label, publisher_verifying_key=alice_verifying_key)
 
+        if self.implementer.federated_only:
+            from nucypher.policy.maps import TreasureMap as _MapClass
+        else:
+            from nucypher.policy.maps import SignedTreasureMap as _MapClass
+
+        # TODO: This LBYL is ugly and fraught with danger.  NRN
```
derekpierre, line 199: Let's give the issue number of the follow-up issue. I'm pretty sure none of this code gets used because of  https://github.com/nucypher/nucypher/pull/2730#discussion_r667964421.

```suggestion
        # TODO: This LBYL is ugly and fraught with danger.  NRN - #2751
```

nucypher/characters/control/interfaces.py
```diff
@@ -96,7 +97,8 @@ def grant(self,
 
         response_data = {'treasure_map': new_policy.treasure_map,
                          'policy_encrypting_key': new_policy.public_key,
-                         'alice_verifying_key': new_policy.alice_verifying_key}
+                         # TODO: technically we should return both Alice's and Publisher's keys here.
```
derekpierre, line 100: If not being addressed in the PR, does this need a follow-up issue?

nucypher/characters/lawful.py
```diff
@@ -1748,34 +1735,27 @@ def interface_info_with_metadata(self):
...
-        writ_hrac, writ_kfrag_checksum, writ_signature = WorkOrder.signed_writ_splitter(signed_writ)
-        reconstructed_writ = writ_hrac + writ_kfrag_checksum
+        # TODO: should it be a method of AuthorizedKeyFrag?
 
         try:
-            self.verify_from(alice, reconstructed_writ, signature=writ_signature)
+            self.verify_from(publisher, authorized_kfrag.writ, signature=authorized_kfrag.writ_signature)
         except InvalidSignature:
-            raise Policy.Unauthorized  # This isn't from Alice (relayer).
+            raise Policy.Unauthorized  # This isn't from Alice (publisher).
```
derekpierre, line 1748: Wanted this included somewhere in PR #2687 but it wasn't before getting merged - https://github.com/nucypher/nucypher/pull/2687#discussion_r664046936. Piggy-backing off of your PR 😅  
```suggestion
            # TODO differentiate cases for Policy.Unauthorized - #2740
            raise Policy.Unauthorized  # This isn't from Alice (publisher).
```

tests/integration/characters/test_bob_handles_frags.py
```diff
@@ -206,13 +206,13 @@ def test_bob_can_issue_a_work_order_to_a_specific_ursula(enacted_federated_polic
...
     encrypted_kfrag = enacted_federated_policy.treasure_map.destinations[ursula.checksum_address]
-    alice = Alice.from_public_keys(verifying_key=enacted_federated_policy.alice_verifying_key)
+    alice = Alice.from_public_keys(verifying_key=enacted_federated_policy.publisher_verifying_key)
     plaintext_kfrag_payload = ursula.verify_from(stranger=alice,
                                                  message_kit=encrypted_kfrag,
                                                  decrypt=True)
     _signed_writ, the_kfrag = work_order.kfrag_payload_splitter(plaintext_kfrag_payload)
 
-    verified_kfrag = the_kfrag.verify(enacted_federated_policy.alice_verifying_key.as_umbral_pubkey())
+    verified_kfrag = the_kfrag.verify(enacted_federated_policy.publisher_verifying_key)
```
derekpierre, line 215: How come the verification of a kfrag requires the `publisher_verifying_key`, as opposed to an `alice_verifying_key`? I understand that for some cases they can be one in the same, but if they are not, we should be able to obtain the `alice_verifying_key` instead.

derekpierre (commented):
Any reason for the empty newsfragment file?

tests/integration/characters/test_bob_handles_frags.py
KPrasch: Agreed - this seems like the one situation when it's actually meant to be the kfrag generator's key, right?

fjarri:
> Any reason for the empty newsfragment file?

Initially it only contained internal changes, but I guess now the Alice/publisher stuff has leaked into the public API. I'll add one.

nucypher/characters/control/interfaces.py
fjarri: I am actually not sure now if the interface warrants it. Can a policy with Provider and Publisher being different characters be even created at the moment, in any way?

tests/integration/characters/test_bob_handles_frags.py
fjarri: Yes, you are right - it was essentially a no-context fix, but looking at a slightly bigger picture, `enacted_federated_policy` should not have been used here at all (same goes for the `alice` above), and the key from `federated_alice` should have been used instead.

nucypher/characters/control/interfaces.py
derekpierre: that's a question for @KPrasch I guess...

fjarri: After some thought, I think it's justified to return only the Alice's key - currently for this interface there's no way to have a separate publisher. When this capability is added, we can extend the return value.

tests/integration/characters/test_bob_handles_frags.py
fjarri: Actually, in this whole bit, `AuthorizedKeyFrag` should have been used - fixed that.

nucypher/characters/lawful.py
```diff
@@ -1748,34 +1735,28 @@ def interface_info_with_metadata(self):
...
+        if authorized_kfrag.hrac != hrac:  # Funky Workorder
             raise Policy.Unauthorized  # Bob, what the *hell* are you doing?
 
-        kfrag_checksum = keccak_digest(bytes(kfrag))[:WRIT_CHECKSUM_SIZE]
-        if kfrag_checksum != writ_kfrag_checksum:
-            raise Policy.Unauthorized  # Bob, Seriously?
-
         try:
-            verified_kfrag = kfrag.verify(verifying_pk=alice.stamp.as_umbral_pubkey())
+            verified_kfrag = authorized_kfrag.kfrag.verify(verifying_pk=publisher.stamp.as_umbral_pubkey())
```
derekpierre, line 1755: This reads a bit weird - similar to my comment in the `test_bob_handles_frags.py`. The "publisher" is used to verify kfrags...as opposed to alice? I know you mentioned there isn't a differentiation between the two at the moment, but naming the parameter "publisher" makes this weird.

fjarri: Yes, I fixed it - added a `provider` parameter to `verify_kfrag_authorization()`. It would be weird to call the parameter `alice` - do we have an accepted term for this? Is `provider` ok?

KPrasch: Sorry if I'm missing something obvious here, but why is it weird to call this parameter "Alice"?  Isn't this always the KFrag generator?

fjarri: Well... we're confusing two naming schemes there. Publisher is named by its role in what's happening, and Alice is a character name. If it was Pablo and not Publisher that would have been at least consistent... Although to me, "Alice" is an instance and not a type, so to speak. When it was only one Alice at least it wasn't that bad, but now `alice` is `Alice`, and `publisher` is also `Alice`.

nucypher/characters/lawful.py
```diff
@@ -1748,34 +1735,29 @@ def interface_info_with_metadata(self):
     #
 
     def verify_kfrag_authorization(self,
-                                   alice: Alice,
-                                   kfrag: KeyFrag,
-                                   signed_writ: bytes,
-                                   work_order: 'WorkOrder'
+                                   hrac: bytes,
+                                   provider: Alice,
```
derekpierre, line 1739: Repeating my comment from discord here.

"""
I'm iffy on `provider` since it is less specific; there is the generator of Kfrags and the publisher of policies

perhaps something like `kfrag_creator`/`kfrag_generator`... 🤷 
"""

derekpierre: "author" and "publisher" dynamic could be good

KPrasch: Yeah "author and publisher" is pretty elegant.

fjarri: > "author" sounds exactly like Alice

To me, Alice is a specific instance (character, person, whatever), Author is a type (or a role). Like, we can have `alice = Author()` in some example code, meaning that we define a character Alice who is an Author. `Alice` as a type always felt to me like it should be a singleton or something.

nucypher/characters/lawful.py
```diff
@@ -1750,34 +1737,29 @@ def interface_info_with_metadata(self):
...
+        if authorized_kfrag.hrac != hrac:  # Funky Workorder
             raise Policy.Unauthorized  # Bob, what the *hell* are you doing?
 
-        kfrag_checksum = keccak_digest(bytes(kfrag))[:WRIT_CHECKSUM_SIZE]
-        if kfrag_checksum != writ_kfrag_checksum:
-            raise Policy.Unauthorized  # Bob, Seriously?
-
         try:
-            verified_kfrag = kfrag.verify(verifying_pk=alice.stamp.as_umbral_pubkey())
+            verified_kfrag = authorized_kfrag.kfrag.verify(verifying_pk=author.stamp.as_umbral_pubkey())
```
derekpierre, line 1758: Nice! That reads a lot better. 🔥

Issue #2735, opened 2021-07-04
Consider allowing Porter API parameters to be passed via URL query strings


derekpierre:
Related to https://github.com/derekpierre/nucypher/pull/5.

Prior discussion on Discord with @vepkenez. 

```

damonC — Yesterday at 9:40 PM
@derek made a https://github.com/derekpierre/nucypher/pull/5  all this stuff is standard and web developers deal with long URLs all the time

---

derek — Today at 1:01 PM
Thanks for the PR :thumbsup: !

Question: how do I correctly provide a list as a parameter? i.e. for include/exclude ursulas parameters. Perhaps I am providing it incorrectly...?

Trying the URL parameter approach with the fix, I get a 400, and a message that include_ursulas parameter is not a valid list
>  r = requests.get('http://127.0.0.1:9155/get_ursulas?quantity=2&duration_periods=4&include_ursulas=["0xC23F4F463c779Dd27c08939dC495b8b35C952c89"]')
>  r.text
Out[211]: "{'include_ursulas': ['Not a valid list.']}"


and the request body has include_ursulas as a string
>>>>> request body: {'quantity': '2', 'duration_periods': '4', 'include_ursulas': '["0xC23F4F463c779Dd27c08939dC495b8b35C952c89"]'}


Providing multiple values yields a 400 as well:
>  r = requests.get('http://127.0.0.1:9155/get_ursulas?quantity=2&duration_periods=4&include_ursulas=0xC23F4F463c779Dd27c08939dC495b8b35C952c89&include_ursulas=0xbdA6Ba3Ad624582495C55451A')

and yields the following request body with a single string entry value for include_ursulas
>>>>> request body: {'quantity': '2', 'duration_periods': '4', 'include_ursulas': '0xC23F4F463c779Dd27c08939dC495b8b35C952c89'}

---

damonC — Today at 1:04 PM
Hmm it may be slightly complex… https://stackoverflow.com/questions/57914396/list-of-query-params-with-flask-request-args
I didn’t try include_ursulas[]=…
That is a common way to do it… but might need that getlist stuff… which is no fun
```

vepkenez:
Maybe the solution to avoiding `getlist()` is to change the format for list type data to something like 
`0x00000000000,0x000000000000,0x000000000000`
It's gross but simple.

derekpierre:
👍 Yep, that's the format I used when I got lists to work.

[...]

PR #2746, opened 2021-07-13
Allow query parameters to be passed in the URL

vepkenez:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1

**Issues fixed/closed:**
Closes https://github.com/nucypher/nucypher/issues/2735.

**Notes for reviewers:**
@derekpierre what do you think? I think it solves all our problems.

### DIFF
--- a/docs/source/application_development/porter.rst
+++ b/docs/source/application_development/porter.rst
@@ -341,6 +341,32 @@ getting involved. Please include contextual information about the sequence of st
 GitHub issue. For any questions, message us in our `Discord <https://discord.gg/7rmXa3S>`_.
 
 
+URL Query Parameters
+^^^^^^^^^^^^^^^^^^^^
+All parameters can be passed as either JSON data within the request or as query parameter strings in the URL.
+
+For ``List`` data types to be passed via a URL query parameter, the value should be provided as a comma-delimited
+String. For example, if a parameter is of type ``List[String]`` either a JSON list of Strings can be provided e.g.
+
+.. code:: bash
+
+    curl -X GET <PORTER URI>/<ENDPOINT> \
+        -H "Content-Type: application/json" \
+        -d '{"parameter_with_list_of_values": ["value1", "value2", "value3"]}'
+
+OR it can be provided via a URL query parameter
+
+.. code:: bash
+
+    curl -X GET <PORTER URI>/<ENDPOINT>?parameter_with_list_of_values=value1,value2,value3
+
+More examples shown below.
+
+.. important::
+
+    If URL query parameters are used and the URL becomes too long, the request may fail.
+
+
 GET /get_ursulas
 ^^^^^^^^^^^^^^^^
 Sample available Ursulas for a policy as part of Alice's ``grant`` workflow. Returns a list of Ursulas
@@ -364,6 +390,7 @@ Parameters
 |                                  |               | | include in the result.                      |
 +----------------------------------+---------------+-----------------------------------------------+
 
+
 Returns
 +++++++
 List of Ursulas with associated information:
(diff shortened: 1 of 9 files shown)


nucypher/control/specifications/fields/base.py
```diff
@@ -42,6 +42,21 @@ class List(BaseField, fields.List):
...
+    and returns list
+    """
+
+    def __init__(self, *args, **kwargs):
+        self.delimiter = kwargs.pop('delimiter', ',')
+        super().__init__(*args, **kwargs)
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if type(value) is not list:
+            value =value.split(self.delimiter)
```
vepkenez, line 57: oops... :|

derekpierre: have some unit tests I want to add, so I can address that.

nucypher/control/controllers.py
```diff
@@ -312,6 +312,11 @@ def handle_request(self, method_name, control_request, *args, **kwargs) -> Respo
             request_body = control_request.data or dict()
             if request_body:
                 request_body = json.loads(request_body)
+
+            # handle query string parameters for GET methods
+            if control_request.method == 'GET' and hasattr(control_request, 'args'):
```
derekpierre, line 317: @vepkenez just pushed a commit, wdyt about adding this additional check?

derekpierre:
🎸 - this looks great @vepkenez . Added some additional tests. Once the build passes we can merge.

nucypher/control/specifications/fields/base.py
derekpierre: ✔️

nucypher/control/controllers.py
vepkenez: POST/PATCH/WHATEVER requests can also have query params.
I say lets honor them and not check the request method.  The question is... which should take precedence?

Should they supersede the request_body?
Should we raise a 400 if the same key appears in both?

It's kind of an edge case and not worth spending a lot of time on in my opinion.

derekpierre: Ok. I'll remove the check.

Come to think of it, I'll also add something for the docs before merging.

derekpierre (approved):
🎸 - I have a couple of commits in here but still giving explicit approval.

derekpierre:
Failed tests are unrelated.

docs/source/application_development/porter.rst
```diff
@@ -455,6 +492,11 @@ Example Request
         -d '{"treasure_map": "Qld7S8sbKFCv2B8KxfJo4oxiTOjZ4VPyqTK5K1xK6DND6TbLg2hvlGaMV69aiiC5QfadB82w/5q1Sw+SNFHN2esWgAbs38QuUVUGCzDoWzQAAAGIAuhw12ZiPMNV8LaeWV8uUN+au2HGOjWilqtKsaP9fmnLAzFiTUAu9/VCxOLOQE88BPoWk1H7OxRLDEhnBVYyflpifKbOYItwLLTtWYVFRY90LtNSAzS8d3vNH4c3SHSZwYsCKY+5LvJ68GD0CqhydSxCcGckh0unttHrYGSOQsURUI4AAAEBsSMlukjA1WyYA+FouqkuRtk8bVHcYLqRUkK2n6dShEUGMuY1SzcAbBINvJYmQp+hhzK5m47AzCl463emXepYZQC/evytktG7yXxd3k8Ak+Qr7T4+G2VgJl4YrafTpIT6wowd+8u/SMSrrf/M41OhtLeBC4uDKjO3rYBQfVLTpEAgiX/9jxB80RtNMeCwgcieviAR5tlw2IlxVTEhxXbFeopcOZmfEuhVWqgBUfIakqsNCXkkubV0XS2l5G1vtTM8oNML0rP8PyKd4+0M5N6P/EQqFkHH93LCDD0IQBq9usm3MoJp0eT8N3m5gprI05drDh2xe/W6qnQfw3YXnjdvf2A=", \
              "bob_encrypting_key": "026d1f4ce5b2474e0dae499d6737a8d987ed3c9ab1a55e00f57ad2d8e81fe9e9ac"}'
 
+.. note::
+
+    If URL parameters are used for the ``/publish_treasure_map`` endpoint instead of JSON data, the request will fail because the URL becomes too long.
```
piotr-roslaniec, line 497: Do we know what the limit is? I know HTTP spec doesn't specify it, and most browsers are ok with 2048 characters. What is the error message in this case? `400 Bad Request` or `500 Internal Server Error`?

derekpierre: I hesitated to put a limit value since I've seen 2048, and 2083 for IE. The result in my testing was `400 Bad Request` but unclear if this is general or not.

derekpierre: Added a short vague note in the `URL Query Parameters` section.

piotr-roslaniec: Thanks!

piotr-roslaniec (approved):
:tada: Thanks for adding this feature, it will help me with `nucypher-ts`.

docs/source/application_development/porter.rst
vepkenez: “Publish” makes more sense as a POST anyway.

piotr-roslaniec: Agreed. I just wanted to know how Porter API users would handle this kind of error (it's not an endpoint-specific error).

PR #2747, opened 2021-07-13
Invalid node certificate fix

xela7:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**

Running on ibex/rinkeby i got the following error:

```
  File "/Users/user/.pyenv/versions/3.8.7/envs/project/lib/python3.8/site-packages/nucypher/network/nodes.py", line 931, in learn_from_teacher_node
    except sprout.SuspiciousActivity:
AttributeError: 'Ursula' object has no attribute 'SuspiciousActivity'


```

I believe this error is occurring because another error that should be caught is not caught.  I've added code to catch the other error and give a warning.  I've also commented out this error to allow for other errors to be raised and not get stopped by the attribute error.

**Why it's needed:**
When running a node on ibex and hitting it with a bob.retrieve request this error occurs.  so couldn't use our own node on a testnet successfully without this change.

**Notes for reviewers:**
I'm wondering why sprout.SuspiciousActivity failed in the first place. Seems like there is something else missing somewhere.  Sorry - I don't know the codebase well enough to suggest change.


nucypher/network/nodes.py
```diff
@@ -927,11 +927,17 @@ def learn_from_teacher_node(self, eager=False, canceller=None):
             # except sprout.Invalidsprout:
             #     self.log.warn(sprout.invalid_metadata_message.format(sprout))
 
-            except sprout.SuspiciousActivity:
```
derekpierre, line 930: Try simple adding the import `from nucypher.network.protocols import SuspiciousActivity`

and then changing this line to be `except SuspiciousActivity`.

xela7: i believe i have fixed this in the (latest) commit 274d3c410edbf5a9b37531c991d0cbcdc5f8be0c

KPrasch (commented):
@xela7 - Can you describe how this issue can be reproduced?

xela7:
@KPrasch  - I found this error when we switched from goerli/lynx to rinkeyby/ibex/.   I did not have the issue when running tests on the goerli/lynx network.

to reproduce:

using nucypher python API

1) create a policy for an alice /bob pair.
2) use enrico to encrypt then store encrypted stuff on ipfs
3) then use bob to retrieve

During retrieval,  bob.retrieve complains about at least one of the ursulas that it finds during the learning loop.  ( I think!?)


It could have been an error that only appeared with the right combination of Ursulas running on ibex and the error was transitory.  

If you need I can re-run testing with regular nucypher, instead of our fork to see if error is still present.

KPrasch:
This PR needs a rebase. Aside from that, I'm unsure this concern is still relevant now that checksum addresses are no longer part of the TLS certificate there is much less room for this kind failure to occur.

KPrasch:
bump

xela7:
ok, I've updated this pul lrequest.  I don't have an opinion as to its relevance given all the new changes.  That said, the changes are pretty minor.  hopefully they would be beneficial on balance.

KPrasch (approved):

PR #2748, opened 2021-07-14
Make `NodeSprout.stamp` be of type `SignatureStamp` and not `bytes`

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 2

**What this does:**
Makes the `stamp` attribute of `NodeSprout` a typed `SignatureStamp` instead of plain bytes. This eliminates the bugs like https://app.circleci.com/pipelines/github/nucypher/nucypher/8166/workflows/09c311e0-3772-47e1-984e-25efdb1e1d52/jobs/138296 where node metadata is created with a repr-of-bytestring name, because `__generate_metadata_filepath()` wasn't applying `hex()` if it just got the stamp as bytes. 

We could just add a check in `__generate_metadata_filepath()` instead, but it is better to make that attribute typed to avoid future similar problems.

### DIFF
--- a/newsfragments/2748.bugfix.rst
+++ b/newsfragments/2748.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed a problem with node metadata being stored to a file with an incorrect name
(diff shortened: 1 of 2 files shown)


KPrasch (approved):
🤠

Thank you for investigating and fixing this @fjarri

derekpierre (approved):
🎸

PR #2764, opened 2021-07-28
Don't include stakers with expired stakes in the `missing_stakers` results for `StakingEscrowAgent.partition_stakers_by_activity()`

### LINKED ISSUE #2760: StakingEscrowAgent overstates the missing stakers (i.e. Inactive Stakers) returned via `partition_stakers_by_activity()`
**Describe the Bug**

`StakingEscrowAgent.partition_stakers_by_activity()` returns a tuple of lists `active_stakers, pending_stakers, missing_stakers`. In separating the stakers for each bucket, stakers whose stakes have expired are not excluded and instead included in the `missing_stakers` list thereby overstating the number of Inactive stakers on the network.

This gets reflected in the status monitor.

Chatted with @vzotova, and this function should exclude expired stakers by checking whether the tokens locked for the current period (`get_locked_tokens()`) and the tokens locked for the next period (`get_locked_tokens(periods=1)`) are both 0. If the values are both 0 then the stake is expired.

derekpierre:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Ignores stakers with expired stakes.

**Issues fixed/closed:**
Fixes #2760 .

**Why it's needed:**
Including stakers with expired stakes causes an overstating of the number of inactive stakers on the network i.e. stakers who haven't committed to the current period.

### DIFF
--- a/newsfragments/2764.bugfix.rst
+++ b/newsfragments/2764.bugfix.rst
@@ -1,0 +1,1 @@
+``StakingEscrow.partition_stakers_by_activity()`` no longer includes stakers with expired stakes in the ``missing_stakers`` value returned, thereby no longer overstating the number of inactive stakers.
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -301,6 +301,10 @@ class StakingEscrowAgent(EthereumContractAgent):
             elif last_committed_period == current_period:
                 pending_stakers.append(staker)
             else:
+                locked_stake = self.non_withdrawable_stake(staker_address=staker)
+                if locked_stake == 0:
+                    # don't include stakers with expired stakes
+                    continue
                 missing_stakers.append(staker)
 
         return active_stakers, pending_stakers, missing_stakers


derekpierre:
@vzotova does the StakingEscrow contract need to be updated to really fix this i.e. the contract should do the pruning of its stakers list from time to time? Otherwise, with this agent-specific fix, we end up with some inconsistencies eg.

```python
>>> active_stakers, pending_stakers, missing_stakers = staking_agent.partition_stakers_by_activity()
>>> staking_agent.get_staker_population()
2129
>>> len(active_stakers) + len(pending_stakers) + len(missing_stakers)
638
```

2129 != 638.

Interested to get your thoughts here.

vzotova:
I'd prefer not to spend gas from anyone to make state clearer - just have this stakers and calculate real numbers offline. But it definitely possible

derekpierre:
@vzotova some times for you 😅 :

Using infura:
```
# partition function before checking for expired stakes on mainnet
473.30053091049194 s
493.7947690486908 s

# partition function when checking for expired stakes on mainnet
875.4592299461365 s
820.4715950489044 s
768.975784778595 s
```

derekpierre:
Ignore the numbers above, I was using the wrong web3 provider:

Using infura:
```
# before fix
262.1745069026947 s
266.5320801734924 s

# after fix
480.71446204185486 s
455.59283089637756 s
```

vzotova (approved):

piotr-roslaniec (approved):

PR #2766, opened 2021-07-29
force pull latest tagged image on external geth deployment

vepkenez:
**Type of PR:**
- [x] Bugfix

**Required reviews:** 
- [X] 1

**What this does:**
Fixes ansible geth deployment to force pull stable image even if a version of it is downloaded already.

**Why it's needed:**
If the node already had an image tagged `stable` it was not pulling the latest.

### DIFF
--- a/deploy/ansible/worker/include/run_external_geth.yml
+++ b/deploy/ansible/worker/include/run_external_geth.yml
@@ -10,6 +10,7 @@
         name: geth
         state: started
         restart: yes
+        pull: true
         image: ethereum/client-go:stable
         restart_policy: "unless-stopped"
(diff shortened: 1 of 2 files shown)


derekpierre:
Kind of interesting use of an env variable for forcing pulls when needed as opposed to every time - see https://stackoverflow.com/questions/55429231/my-playbook-is-not-downloading-the-updated-images-with-same-tag-name...don't know if that is sufficient for our needs. Thoughts?

vepkenez:
@derekpierre the intention is that we want the latest stable branch.  There is nothing complicated or conditional about it.  

I could see adding a way to insert a specific image as a cli arg to the ansible command as a future feature, but this is a fix for a bug which is that the intended result which is to pull the latest stable version, is not occurring.

derekpierre (approved):
🎸

KPrasch (approved):
In it goes.

PR #2767, opened 2021-08-04
Versioning of bytes serializable protocol entities

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
The goal of this PR is to gently introduce a unified version and branding scheme for the following entities (alphabetical):

| Entity         | Prefix       | V | Header |
|----------------|--------------|----|-----------|
| `Arrangement` | `b'AR'` | 1.0 | ` b'AR\x00\x01\x00\x00'` |
| `AuthorizedFrag` | `b'KF'` | 1.0 | ` b'KF\x00\x01\x00\x00'` |
| `EncryptedTreasureMap` | `b'EM'` | 1.0 | ` b'EM\x00\x01\x00\x00'` |
| `MessageKit` | `b'MK'` | 1.0 | ` b'MK\x00\x01\x00\x00'` |
| `RetrievalKit` | `b'RK'` | 1.0 | ` b'RK\x00\x01\x00\x00'` |
| `ReencryptionRequest` | `b'RQ'` | 1.0 | ` b'RQ\x00\x01\x00\x00'` |
| `ReencryptionResponse` | `b'RR'` | 1.0 | ` b'RR\x00\x01\x00\x00'` |
| `RevocationOrder` | `b'RV'` | 1.0 | ` b'RV\x00\x01\x00\x00'` |
| `TreasureMap` | `b'TM'` | 1.0 | ` b'TM\x00\x01\x00\x00'` |

Related changes: 
- Introduces `Versioned` ABC (inspired by @fjarri 's original gist).
- Subclasses `Versioned` and implement abstract methods on bytes-serializable entities listed above.
- Prepends six-byte header to bytestrings (two-byte brand, two-byte major version, two-byte minor version).
- Enforces Invalid or incompatible version and branding at deserialization time.
- Renames `Revocation` -> `RevocationOrder`.
- Adjusts size constant `AuthorizedKeyFrag.ENCRYPTED_SIZE` to accommodate headers.

**Issues fixed/closed:**
#343 #281 

**Why it's needed:**
- Required for management of future release backwards compatibility
- Supports immediate backwards-incompatibility with "Treasure Map Con KFrags"

**Notes for reviewers:**
- Model of this strategy by @fjarri <https://gist.github.com/fjarri/bad73af26927e6fb3b1aa5f61f8863e3>
- Provide backwards compatibility for unversioned entities?
- `Ursula` subclasses `Versioned` too? This will cause an additional major backwards incompatibility in node discovery.

### DIFF
--- a/nucypher/policy/policies.py
+++ b/nucypher/policy/policies.py
@@ -44,8 +44,7 @@ from nucypher.utilities.versioning import Versioned
 class Arrangement(Versioned):
     """A contract between Alice and a single Ursula."""
 
-    def __init__(self, publisher_verifying_key: PublicKey, expiration: maya.MayaDT, *args, **kwargs):
-        super().__init__(*args, **kwargs)
+    def __init__(self, publisher_verifying_key: PublicKey, expiration: maya.MayaDT):
         self.expiration = expiration
         self.publisher_verifying_key = publisher_verifying_key
 
--- a/nucypher/policy/policies.py
+++ b/nucypher/policy/policies.py
@@ -62,8 +62,8 @@ class Arrangement(Versioned):
         return b'AR'
 
     @classmethod
-    def _version(cls) -> int:
-        return 1
+    def _version(cls) -> Tuple[int, int]:
+        return 1, 0
 
     def _payload(self) -> bytes:
         """Returns the unversioned bytes serialized representation of this instance."""
--- a/nucypher/policy/policies.py
+++ b/nucypher/policy/policies.py
@@ -38,24 +38,19 @@ from nucypher.policy.reservoir import (
 from nucypher.policy.revocation import RevocationKit
 from nucypher.utilities.concurrency import WorkerPool
 from nucypher.utilities.logging import Logger
+from nucypher.utilities.versioning import Versioned
 
 
-class Arrangement:
-    """
-    A contract between Alice and a single Ursula.
-    """
+class Arrangement(Versioned):
+    """A contract between Alice and a single Ursula."""
 
-    splitter = BytestringSplitter(
-        key_splitter,                      # publisher_verifying_key
-        (bytes, VariableLengthBytestring)  # expiration
-    )
-
-    def __init__(self, publisher_verifying_key: PublicKey, expiration: maya.MayaDT):
+    def __init__(self, publisher_verifying_key: PublicKey, expiration: maya.MayaDT, *args, **kwargs):
+        super().__init__(*args, **kwargs)
         self.expiration = expiration
         self.publisher_verifying_key = publisher_verifying_key
 
-    def __bytes__(self):
-        return bytes(self.publisher_verifying_key) + bytes(VariableLengthBytestring(self.expiration.iso8601().encode()))
+    def __repr__(self):
+        return f"Arrangement(publisher={self.publisher_verifying_key})"
 
     @classmethod
     def from_publisher(cls, publisher: 'Alice', expiration: maya.MayaDT) -> 'Arrangement':
@@ -63,13 +58,30 @@ class Arrangement:
         return cls(publisher_verifying_key=publisher_verifying_key, expiration=expiration)
 
     @classmethod
-    def from_bytes(cls, arrangement_as_bytes: bytes) -> 'Arrangement':
-        publisher_verifying_key, expiration_bytes = cls.splitter(arrangement_as_bytes)
+    def _brand(cls) -> bytes:
+        return b'AR'
+
+    @classmethod
+    def _version(cls) -> int:
+        return 1
+
+    def _payload(self) -> bytes:
+        """Returns the unversioned bytes serialized representation of this instance."""
+        return bytes(self.publisher_verifying_key) + bytes(VariableLengthBytestring(self.expiration.iso8601().encode()))
+
+    @classmethod
+    def _old_version_handlers(cls) -> Dict:
+        return {}
+
+    @classmethod
+    def _from_bytes_current(cls, data: bytes):
+        splitter = BytestringSplitter(
+            key_splitter,  # publisher_verifying_key
+            (bytes, VariableLengthBytestring)  # expiration
+        )
+        publisher_verifying_key, expiration_bytes = splitter(data)
         expiration = maya.MayaDT.from_iso8601(iso8601_string=expiration_bytes.decode())
         return cls(publisher_verifying_key=publisher_verifying_key, expiration=expiration)
-
-    def __repr__(self):
-        return f"Arrangement(publisher={self.publisher_verifying_key})"
 
 
 class Policy(ABC):
--- a/nucypher/policy/revocation.py
+++ b/nucypher/policy/revocation.py
@@ -95,7 +95,7 @@ class Revocation(Versioned):
 
         splitter = BytestringSplitter(
             checksum_address_splitter,  # ursula canonical address
-            (bytes, AuthorizedKeyFrag.SERIALIZED_SIZE),  # encrypted kfrag payload (includes writ)
+            (bytes, Versioned._HEADER_SIZE+AuthorizedKeyFrag.SERIALIZED_SIZE),  # MessageKit version header + versioned ekfrag
             signature_splitter
         )
         ursula_canonical_address, ekfrag, signature = splitter(data)
--- a/nucypher/policy/revocation.py
+++ b/nucypher/policy/revocation.py
@@ -30,7 +30,7 @@ from nucypher.policy.maps import AuthorizedKeyFrag
 from nucypher.utilities.versioning import Versioned
 
 
-class Revocation(Versioned):
+class RevocationOrder(Versioned):
     """
     Represents a string used by characters to perform a revocation on a specific
     Ursula. It's a bytestring made of the following format:
@@ -110,9 +110,9 @@ class RevocationKit:
     def __init__(self, treasure_map, signer: SignatureStamp):
         self.revocations = dict()
         for node_id, encrypted_kfrag in treasure_map:
-            self.revocations[node_id] = Revocation(ursula_checksum_address=node_id,
-                                                   encrypted_kfrag=encrypted_kfrag,
-                                                   signer=signer)
+            self.revocations[node_id] = RevocationOrder(ursula_checksum_address=node_id,
+                                                        encrypted_kfrag=encrypted_kfrag,
+                                                        signer=signer)
 
     def __iter__(self):
         return iter(self.revocations.values())
--- a/nucypher/policy/revocation.py
+++ b/nucypher/policy/revocation.py
@@ -16,7 +16,7 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
 
-from typing import Optional, Dict
+from typing import Optional, Dict, Tuple
 
 from bytestring_splitter import BytestringSplitter
 from eth_typing.evm import ChecksumAddress
@@ -77,8 +77,8 @@ class RevocationOrder(Versioned):
         return b'RV'
 
     @classmethod
-    def _version(cls) -> int:
-        return 1
+    def _version(cls) -> Tuple[int, int]:
+        return 1, 0
 
     @classmethod
     def _old_version_handlers(cls) -> Dict:
--- a/nucypher/policy/revocation.py
+++ b/nucypher/policy/revocation.py
@@ -16,7 +16,7 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
 
-from typing import Optional
+from typing import Optional, Dict
 
 from bytestring_splitter import BytestringSplitter
 from eth_typing.evm import ChecksumAddress
@@ -27,9 +27,10 @@ from nucypher.crypto.splitters import signature_splitter, checksum_address_split
 from nucypher.crypto.umbral_adapter import Signature, PublicKey
 from nucypher.policy.kits import MessageKit
 from nucypher.policy.maps import AuthorizedKeyFrag
+from nucypher.utilities.versioning import Versioned
 
 
-class Revocation:
+class Revocation(Versioned):
     """
     Represents a string used by characters to perform a revocation on a specific
     Ursula. It's a bytestring made of the following format:
@@ -37,16 +38,9 @@ class Revocation:
     This is sent as a payload in a DELETE method to the /KFrag/ endpoint.
     """
 
-    PREFIX = b'REVOKE-'
-    revocation_splitter = BytestringSplitter(
-        (bytes, len(PREFIX)),
-        checksum_address_splitter, # ursula canonical address
-        (bytes, AuthorizedKeyFrag.ENCRYPTED_SIZE),  # encrypted kfrag payload (includes writ)
-        signature_splitter
-    )
-
     def __init__(self,
-                 ursula_checksum_address: ChecksumAddress,  # TODO: Use staker address instead (what if the staker rebonds)?
+                 ursula_checksum_address: ChecksumAddress,
+                 # TODO: Use staker address instead (what if the staker rebonds)?
                  encrypted_kfrag: MessageKit,
                  signer: Optional[SignatureStamp] = None,
                  signature: Optional[Signature] = None):
@@ -57,12 +51,9 @@ class Revocation:
         if not (bool(signer) ^ bool(signature)):
             raise ValueError("Either pass a signer or a signature; not both.")
         elif signer:
-            self.signature = signer(self.payload)
+            self.signature = signer(self._body())
         elif signature:
             self.signature = signature
-
-    def __bytes__(self):
-        return self.payload + bytes(self.signature)
 
     def __repr__(self):
         return bytes(self)
@@ -73,27 +64,45 @@ class Revocation:
     def __eq__(self, other):
         return bytes(self) == bytes(other)
 
-    @property
-    def payload(self):
-        return self.PREFIX                                          \
-               + to_canonical_address(self.ursula_checksum_address) \
-               + bytes(self.encrypted_kfrag)                        \
-
-    @classmethod
-    def from_bytes(cls, revocation_bytes):
-        prefix, ursula_canonical_address, ekfrag, signature = cls.revocation_splitter(revocation_bytes)
-        ursula_checksum_address = to_checksum_address(ursula_canonical_address)
-        return cls(ursula_checksum_address=ursula_checksum_address,
-                   encrypted_kfrag=ekfrag,
-                   signature=signature)
-
     def verify_signature(self, alice_verifying_key: PublicKey) -> bool:
         """
         Verifies the revocation was from the provided pubkey.
         """
-        if not self.signature.verify(self.payload, alice_verifying_key):
+        if not self.signature.verify(self._body(), alice_verifying_key):
             raise InvalidSignature(f"Revocation has an invalid signature: {self.signature}")
         return True
+
+    @classmethod
+    def _brand(cls) -> bytes:
+        return b'RV'
+
+    @classmethod
+    def _version(cls) -> int:
+        return 1
+
+    @classmethod
+    def _old_version_handlers(cls) -> Dict:
+        return {}
+
+    def _body(self) -> bytes:
+        return to_canonical_address(self.ursula_checksum_address) + bytes(self.encrypted_kfrag)
+
+    def _payload(self) -> bytes:
+        return self._body() + bytes(self.signature)
+
+    @classmethod
+    def _from_bytes_current(cls, data):
+
+        splitter = BytestringSplitter(
+            checksum_address_splitter,  # ursula canonical address
+            (bytes, AuthorizedKeyFrag.SERIALIZED_SIZE),  # encrypted kfrag payload (includes writ)
+            signature_splitter
+        )
+        ursula_canonical_address, ekfrag, signature = splitter(data)
+        ursula_checksum_address = to_checksum_address(ursula_canonical_address)
+        return cls(ursula_checksum_address=ursula_checksum_address,
+                   encrypted_kfrag=ekfrag,
+                   signature=signature)
 
 
 class RevocationKit:
--- a/nucypher/utilities/versioning.py
+++ b/nucypher/utilities/versioning.py
@@ -15,6 +15,7 @@
  along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
+
 from abc import abstractmethod, ABC
 from typing import Dict
 
--- a/nucypher/utilities/versioning.py
+++ b/nucypher/utilities/versioning.py
@@ -17,21 +17,27 @@
 
 
 from abc import abstractmethod, ABC
-from typing import Dict, Tuple, Callable
+from typing import Dict, Tuple, Callable, NamedTuple
 
 
 class Versioned(ABC):
     """Base class for serializable entities"""
 
-    _BRAND_LENGTH = 2  # bytes
-    _VERSION_LENGTH = 2
-    _HEADER_SIZE = _BRAND_LENGTH + _VERSION_LENGTH
+    _PARTS = 2  # bytes
+    _PART_SIZE = 2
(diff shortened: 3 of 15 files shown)


derekpierre (commented):
> Supports immediate backwards-incompatibility with "Treasure Map Con KFrags"

Given the backwards incompatibility with TMapConKfrags, in addition to the feature newsfragment, should we also add a `removal` newsfragment to indicate deprecation of the non- TMapConKFrags (TMapSinKFrags)?

> Provide backwards compatibility for unversioned entities? 
> Ursula subclasses Versioned too? This will cause an additional major backwards incompatibility in node discovery.

I don't think I understand enough about the future ramifications of not doing this. Could you clarify those for my benefit? Presumably, if we do some sort of Ursula level versioning, we may end up with a fragmented network that may also have implications with respect to sampling of Ursulas i.e. Alice's choice of Ursulas?

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,83 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def _payload(self) -> bytes:
+        """Returns the unversioned bytes serialized representation of this instance."""
+        raise NotImplementedError
+
+    @classmethod
+    @abstractmethod
+    def _old_version_handlers(cls) -> Dict:
```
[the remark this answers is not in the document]
KPrasch: Thanks for the feedback!

`Versioned._old_version_handlers` is an internal method.  We need a way to compose the old version handlers on each subclass.  Moving the handler lookup logic inside this method means repeating it in each subclass's implementation or at least introducing an additional lookup method.

> instead of a Dict with all of the handlers for all of the previous versions?

There will still need to be a mapping somewhere, right?  Where do you suggest relocating it to?

> A Dict for the entire version history does not have to be generated or returned for every call to this function

I agree this feels somewhat imperfect. however the generation of this dictionary must have the method handlers in scope, and using an abstract method is useful here.

> This function can return None if no handler exists for the specified version, and therefore a KeyError does not have to be handled on L49.

I don't know If this as an advantage 😄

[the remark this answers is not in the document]
derekpierre: > There will still need to be a mapping somewhere, right?

Not necessarily since within the function it could use a mapping (although I don't think it necessarily has to be), but it could also just be if-statements based on the version number?

> I agree this feels somewhat imperfect. however the generation of this dictionary must have the method handlers in scope, and using an abstract method is useful here.

It may help me to understand what an implementation of an old version handler would look like - I assumed it would probably be a reference to some Callable - can it be something else? Is scope an issue if presumably, these callables would be class methods or static methods (?) because `_old_version_hander(cls, version)` would be a class method?

Apologies if I am missing something obvious.

> I don't know If this as an advantage 😄

Haha fair - that one is more personal preference. I rather not have to catch `KeyError`. 😅

KPrasch:
> I don't think I understand enough about the future ramifications of not doing this. Could you clarify those for my benefit? Presumably, if we do some sort of Ursula level versioning, we may end up with a fragmented network that may also have implications with respect to sampling of Ursulas i.e. Alice's choice of Ursulas?

Let's start by understanding that new (de)serialization formats for the aforementioned entities will no longer work with previous versions of the code base.  This creates several surfaces of backwards-incompatibility, for example:

1. Existing treasure maps generated by previous versions will no longer be usable by Alice or Bob.
2. Existing `MessageKit`s will no longer be decryptable by Alice, Bob, or Ursula running an older version.
3. `WorkOrder`s (now `ReencryptionRequest`)` generated by older versions will no longer be usable by Ursulas.
4. Offline revocations will no longer be deserializable by Alice or Ursula.

> Provide backwards compatibility for unversioned entities?

Currently I expect the failure mode for using outdated bytes to be caused by a [missing brand](<https://gist.github.com/fjarri/bad73af26927e6fb3b1aa5f61f8863e3#file-versioning-py-L14>) or [invalid version bytes](https://gist.github.com/fjarri/bad73af26927e6fb3b1aa5f61f8863e3#file-versioning-py-L10) .  We can potentially improve this behavior by detect the missing header or catching an error and re-raise a specific error message.   One issue with this approach is that there is no way to determine if a given bytes blob is outdated or simply invalid since the start bytes will be random. We can potentially even attempt to maintain backwards-compatibility for old policy entities.  I think this is a bad idea right now but still worth noting for the future.

> Ursula subclasses Versioned too? This will cause an additional major backwards incompatibility in node discovery.

If we also consider Ursula, if we implement the same versioning scheme we'll definitely need to provide backwards compatibility to prevent segmenting the network, sampling, etc.  This will require some additional work in this PR if we decide that it belongs in this changeset.

nucypher/utilities/versioning.py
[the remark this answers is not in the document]
KPrasch: > Not necessarily since within the function it could use a mapping (although I don't think it necessarily has to be), but it could also just be if-statements based on the version number?

This is essentially just avoiding the creation of an actual dictionary.  I call premature optimization! 😝 

> It may help me to understand what an implementation of an old version handler would look like 

Have a look here:
<https://gist.github.com/fjarri/bad73af26927e6fb3b1aa5f61f8863e3#file-versioning-py-L61>

> Haha fair - that one is more personal preference. I rather not have to catch KeyError. sweat_smile

As the old python adages go: "Better to ask for forgiveness than permission." and "Don't look before you leap!"

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,94 @@
...
+        """Raised when an unexpected or invalid bytes header is encountered during deserialization"""
+
+    @classmethod
+    def from_bytes(cls, data):
+
+        # Parse brand
+        brand = data[:cls._BRAND_LENGTH]
+        if brand != cls._brand():
+            error = f"Incorrect brand. Expected {cls._brand()}, Got {brand}."
+            if not brand.isalpha():
```
fjarri, line 38: Is it just a convenience thing? Separating cases when an incorrect (but branded) object is passed vs some random bytestring?

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,94 @@
...
+            raise cls.InvalidHeader(error)
+
+        # Parse body
+        remainder = data[version_index:]
+
+        # Select deserializer and process
+        if version_number == cls._version():
+            return cls._from_bytes_current(remainder)
+        handlers = cls._old_version_handlers()
+        try:
```
fjarri, line 58: Is there a need for this `try`? We have just checked that the version number has a handler.

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,94 @@
...
+        except KeyError:
+            raise ValueError(f"Incorrect or unknown version number ({version_number}).")
+
+    def __bytes__(self):
+        return self._header() + self._payload()
+
+    @classmethod
+    def _header(cls) -> bytes:
+        version_bytes = cls._version().to_bytes(cls._VERSION_LENGTH, 'big')
+        return cls._brand() + version_bytes
```
fjarri, line 69: I have a minor concern that we do not enforce that `len(cls._brand) == cls._BRAND_LENGTH`, but perhaps it will be easy to notice.

nucypher/policy/revocation.py
```diff
@@ -27,26 +27,20 @@
...
-        (bytes, len(PREFIX)),
-        checksum_address_splitter, # ursula canonical address
-        (bytes, AuthorizedKeyFrag.ENCRYPTED_SIZE),  # encrypted kfrag payload (includes writ)
-        signature_splitter
-    )
-
     def __init__(self,
-                 ursula_checksum_address: ChecksumAddress,  # TODO: Use staker address instead (what if the staker rebonds)?
+                 ursula_checksum_address: ChecksumAddress,
+                 # TODO: Use staker address instead (what if the staker rebonds)?
```
fjarri, line 43: Not really the topic of this PR, but I think it is the staker address - it comes from the treasure map

fjarri:
Some general notes:
- the list in the description is missing `RetrievalKit` and `AuthorizedKeyFrag`, which are also versioned
- we may want to make `Arrangement` versioned as well
- Ursula info is currently versioned independently - do we bring it to this scheme too?
- Another possible protocol object is the stuff returned from `server.py::all_known_nodes()` (currently not even an object)
- I am worried about possible transition to `nucypher-core` where we may want to use `serde`-specific versioning packages which will be hard to connect to this scheme

nucypher/utilities/versioning.py
KPrasch: Yes, that's right.  I specifically am trying to control he path of execution for any "unbranded" bytes generated by previous versions of the codebase.  I also added an alphanumeric enforcement in `.header()`.  I don't think there are any reliable ways to determine if a blob of bytes is random or an old workorder for example, but at least we can say that it's completely incompatible either way.

nucypher/utilities/versioning.py
KPrasch: ✔️

nucypher/utilities/versioning.py
KPrasch: ✔️

nucypher/utilities/versioning.py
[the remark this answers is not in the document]
KPrasch: Ended up removing the `try` in favor of membership checking a subsequent refactor 😆 .

KPrasch:
> the list in the description is missing RetrievalKit and AuthorizedKeyFrag, which are also versioned
 ✔️ 
 
 > we may want to make Arrangement versioned as well
 ✔️ 

> Ursula info (see Ursula.payload_splitter()) is currently versioned independently - do we bring it to this scheme too? Another possible protocol object is the stuff returned from server.py::all_known_nodes() (currently not even an object)

For whatever reason I'm somewhat less concerned with node discovery metadata, but it's a great question.  It certainly will not hurt.  Shall I go ahead and also version `Ursula` and make a class for `UrsulaInfo` in this PR?  I'm thinking "yes".

> I am worried about possible transition to nucypher-core where we may want to use serde-specific versioning packages which will be hard to connect to this scheme.  

It may be good to have a scheme that is compatible across different language implementations.  Will `serde` add additional constraints in this regard, or can it be easily made compatible in python?  I'm eager to get *something* delivered for versioning in the next release (hopefully as soon as this PR is merged) since many entities are already completely backwards incompatible.  
 
 This system is elegant and is working *right now*.  This is a very simple scheme, why will it be hard to implement in nucypher core?

fjarri:
> It certainly will not hurt. Shall I go ahead and also version Ursula and make a class for UrsulaInfo in this PR? I'm thinking "yes".

I'm looking into it myself right now, to separate it into `nucypher-core`... and it's a massive can of worms. But perhaps if you just want to replace the existing versioning scheme, it will be easier to do.

> It may be good to have a scheme that is compatible across different language implementations. Will serde add additional constraints in this regard, or can it be easily made compatible in python?

Well, that's why we wanted to have a core in Rust and bindings to other languages :) But really, it will depend on what we choose as the "reference" binary implementation, and on the specific versioning solution. My guess is it can be made compatible, but perhaps not with the specific scheme in this PR.

> I'm eager to get something delivered for versioning in the next release (hopefully as soon as this PR is merged) since many entities are already completely backwards incompatible.

Yes, I agree with that. But I'm worried that if we do want to proceed with nucypher-core, we will have to make another major release for it.

vepkenez:
I think using this as a reference implementation and attempting to replicate it in nucypher-core is the way to go.  Hopefully any tweaks will not be substantial.  

Keeping this in Python and _not_ doing it in nucypher-core means that we need to copy/rewrite it in nucypher-ts and then in every other future language/platform forever.

fjarri:
> I think using this as a reference implementation and attempting to replicate it in nucypher-core is the way to go. Hopefully any tweaks will not be substantial.

We will have to write our own serializer for that and not use any machinery that `serde` provides.

vepkenez:
> We will have to write our own serializer for that and not use any machinery that `serde` provides.

Do you mean we won't be able to use the `serde` versioning libraries?  Or can't use `serde` _at all_?

fjarri:
One can't enforce any specific position of serialized data with `serde` alone, and we want the brand and the version to appear first in the bytestring. We can try and find some format that *happens* to do what we want (not sure if there is one), or write our own format implementation, and then we will have to implement version handling on top of that manually.

vepkenez:
So, the alternative within the `serde` ecosystem would be to use one of a few schemes that exist that would mostly accomplish the same goal of allowing for the identification of the version/brand of some bytestring and allowing for python/javascript level handling based on that info, without necessarily adhering to the structural constraint of ordering the bytes in this exact way?

fjarri:
Actually, the crate I wanted to use for versioning (https://github.com/doctorn/obake) appears to be non-serde-specific, and the other solution I saw (https://docs.rs/serde-version/0.5.1/serde_version/) is 2 years old and unmaintained.

So, given that we will have to fix a single binary format for serialization, we can use a hybrid approach: keep the brand/version requirement as the first four bytes, and the rest will be whatever our Python code, or later `serde` serializes to. These bytes can be separated and used for dispatch in Rust code as well. The only issue I see is that when we switch to Rust, we will have to write a custom BSS-mimicking deserializer (shouldn't be *too* hard...), and there is a small possibility there will be some pesky type that we won't be able to deserialize in Rust (low probability; I have some worries about `MayaDT`)

fjarri:
Also, while this PR is not merged yet, and seeing how there's already talk about supporting "ranges" of versions, I would like to revisit the suggestion to use "major.minor" versions instead of a single number.

KPrasch:
> Also, while this PR is not merged yet, and seeing how there's already talk about supporting "ranges" of versions, I would like to revisit the suggestion to use "major.minor" versions instead of a single number.

@fjarri, I've incorporated this suggestion in ba6a49e13cee32c840849ac5b3523c91f7d3d94b. I've made the assumption that when the major version changes, backwards-compatibility with older bytes strings is broken but maintained for minor versions.  In this sense the minor version number is just added metadata.  Do you have a different sense of how this might operate?

fjarri:
> Do you have a different sense of how this might operate?  

Basically what you implemented, but slightly more relaxed: I think, if `bytestring.major == class.major` and `bytestring.minor > class.minor`, we can pass the bytestring to `_from_bytes_current()` and expect it to handle it. In most cases, yes, it will mean that later versions only add some fields but don't change existing ones, but I do not exclude the possibility that sometimes it will be possible to add some "future-proofing" (most likely, default values for fields). 

I am not sure that ideologically it makes sense to keep deserializers for old major versions in the class as opposed to failing right away if `bytestring.major != class.major`; is it really still a different major version if we continue to support it?

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,160 @@
...
+
+class Versioned(ABC):
+    """Base class for serializable entities"""
+
+    _PARTS = 2  # bytes
+    _PART_SIZE = 2
+    _BRAND_SIZE = 2
+    _VERSION_SIZE = _PART_SIZE * _PARTS
+    _HEADER_SIZE = _BRAND_SIZE + _VERSION_SIZE
+    _MAXIMUM_VERSION = 99  # length
```
fjarri, line 31: Doesn't seem to be used

KPrasch: Whoops, I did not mean to commit this change. Thanks for noticing.

tests/unit/test_versioning.py
```diff
@@ -0,0 +1,181 @@
...
+import pytest
+
+from nucypher.utilities.versioning import Versioned
+
+
+def _check_valid_version_tuple(version: Any, cls: Type):
+    if not isinstance(version, tuple):
+        pytest.fail(f"Old version handlers keys for {cls.__name__} must be a tuple")
+    if not len(version) == Versioned._PARTS:
+        pytest.fail(f"Old version handlers keys for {cls.__name__} must be a {str(Versioned._BRAND_SIZE)}-tuple")
```
derekpierre, line 30: Is this value supposed to be `_PARTS`?

```suggestion
        pytest.fail(f"Old version handlers keys for {cls.__name__} must be a {str(Versioned._PARTS)}-tuple")
```

tests/unit/test_versioning.py
```diff
@@ -0,0 +1,181 @@
...
+    @classmethod
+    def _version(cls) -> Tuple[int, int]:
+        return 2, 1
+
+    def _payload(self) -> bytes:
+        return bytes(self.x)
+
+    @classmethod
+    def _old_version_handlers(cls):
+        return {
```
derekpierre, line 53: See my question about whether we maintain handlers for older major versions.

tests/unit/test_versioning.py
```diff
@@ -0,0 +1,181 @@
...
+
+    @classmethod
+    def _from_bytes_current(cls, data):
+        return cls(str(int(data, 16)))  # but now we use a string representation for some reason
+
+
+def test_unique_branding():
+    brands = tuple(v._brand() for v in Versioned.__subclasses__())
+    if len(brands) != len(set(brands)):
+        pytest.fail(f"Duplicated_brand {Versioned._brand()}.")
```
derekpierre, line 74: Doesn't this test just raise a `NotImplementedError` if the test fails - is this intended? Perhaps you can instead show the duplicate - not the most efficient but 1) it is a unit test, and 2) only happens in the failure case)

```suggestion
    brands = tuple(v._brand() for v in Versioned.__subclasses__())
    brands_set = set(brands)
    if len(brands) != len(brands_set):
        duplicate_brands = list(brands)
        for brand in brands_set:
            duplicate_brands.remove(brand)
        pytest.fail(f"Duplicated brand(s) {duplicate_brands}.")
```

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,160 @@
...
+        return payload
+
+    @classmethod
+    def _deserializers(cls) -> Dict[Tuple[int, int], Callable]:
+        """Return a dict of all known deserialization handlers for this class keyed by version"""
+        return {cls._version(): cls._from_bytes_current, **cls._old_version_handlers()}
+
+
+# Collects the brands of every serializable entity
+SERIALIZABLE_ENTITIES = {v.__class__.__name__: v._brand() for v in Versioned.__subclasses__()}
```
derekpierre, line 160: Is this used somewhere?

KPrasch:
> ...if bytestring.major == class.major and bytestring.minor > class.minor, we can pass the bytestring to _from_bytes_current() and expect it to handle it... 

> I am not sure that ideologically it makes sense to keep deserializers for old major versions in the class as opposed to failing right away if bytestring.major != class.major; is it really still a different major version if we continue to support it?

@fjarri, Considering the above two points. if we expect a minor version increase to be compatible with the current major version (`_from_bytes_current`) *and* we do not maintain previous major version handlers...  what's the point of having `_old_version_handlers` at all?

nucypher/utilities/versioning.py
KPrasch: ✔️ Commented it out and left an explainer.

tests/unit/test_versioning.py
KPrasch: ✔️

tests/unit/test_versioning.py
KPrasch: ✔️

tests/unit/test_versioning.py
KPrasch: ✔️  Good catch, thank you.

nucypher/utilities/versioning.py
KPrasch: ✔️

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @classmethod
+    def from_bytes(cls, data: bytes):
+        """"Public deserialization API"""
+        if len(data) < cls._HEADER_SIZE:
+            raise ValueError(f'Invalid bytes for {cls.__name__}.')
+        brand, version, payload = cls._parse(data)
+        major, minor = version
+        latest_major_version = cls._version()[0]
+        handlers = cls._deserializers()
```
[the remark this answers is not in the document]
KPrasch: ✔️

KPrasch:
> Basically what you implemented, but slightly more relaxed: I think, if bytestring.major == class.major and bytestring.minor > class.minor, we can pass the bytestring to _from_bytes_current() and expect it to handle it.

✔️ 

> I am not sure that ideologically it makes sense to keep deserializers for old major versions in the class as opposed to failing right away if bytestring.major != class.major

✔️

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,160 @@
...
+        compatible = major == latest_major_version
+        if not compatible:
+            message = f'Incompatible versioned bytes for {cls.__name__}. ' \
+                      f'Compatible version is {cls.version_string()}; Got {major}.{minor}.'
+            raise cls.IncompatibleVersion(message)
+        return handlers[version](payload)
+
+    @classmethod
+    def _parse(cls, data: bytes) -> Tuple[bytes, Tuple[int, int], bytes]:
```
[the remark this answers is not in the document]
KPrasch: ✔️

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,160 @@
...
+        if len(data) < cls._HEADER_SIZE:
+            raise ValueError(f'Invalid bytes for {cls.__name__}.')
+        brand, version, payload = cls._parse(data)
+        major, minor = version
+        latest_major_version = cls._version()[0]
+        handlers = cls._deserializers()
+        compatible = major == latest_major_version
+        if not compatible:
+            message = f'Incompatible versioned bytes for {cls.__name__}. ' \
+                      f'Compatible version is {cls.version_string()}; Got {major}.{minor}.'
```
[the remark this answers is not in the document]
KPrasch: ✔️

fjarri:
> if we expect a minor version increase to be compatible with the current major version (_from_bytes_current) and we do not maintain previous major version handlers... what's the point of having _old_version_handlers at all?

Ok, fair enough, you're right. I think the guideline is that X.Y can be deserialized by X.Z as long as Y >= Z, and when you can't guarantee that, you need to bump the major version. But we need some limit on how long the old major versions are supported. Maybe to the point that they can be dropped at any time without further notice.

nucypher/utilities/versioning.py
```diff
@@ -0,0 +1,166 @@
...
+    @classmethod
+    def from_bytes(cls, data: bytes):
+        """"Public deserialization API"""
+        brand, version, payload = cls._parse_header(data)
+        version = cls._get_compatible_handler(data=data, version=version)
+        handlers = cls._deserializers()
+        return handlers[version](payload)
+
+    @classmethod
+    def _get_compatible_handler(cls, data: bytes, version: Tuple[int, int]) -> Tuple[int, int]:
```
fjarri: A small nitpick - a better name would be `_get_compatible_handler_version()`

KPrasch: I *just* renamed it to `resolve_version` as you submitted this comment.

nucypher/policy/policies.py
```diff
@@ -38,39 +38,51 @@
...
-
-    splitter = BytestringSplitter(
-        key_splitter,                      # publisher_verifying_key
-        (bytes, VariableLengthBytestring)  # expiration
-    )
+class Arrangement(Versioned):
+    """A contract between Alice and a single Ursula."""
 
-    def __init__(self, publisher_verifying_key: PublicKey, expiration: maya.MayaDT):
+    def __init__(self, publisher_verifying_key: PublicKey, expiration: maya.MayaDT, *args, **kwargs):
```
fjarri, line 47: why `*args, **kwargs`?

derekpierre (approved):
🎸 

@KPrasch Any thoughts on also adding a `removal` newsfragment to indicate deprecation of non-versioned object serialization?

KPrasch:
> indicate deprecation of non-versioned object serialization?

I'm a bit hesitant to do so since there are still some unversioned (or differently versioned) entities.

I suppose it can say something like "Removes support for almost all unversioned bytestrings."

fjarri (approved):

Issue #2751, opened 2021-07-14
Update TreasureMap differentiation of federated vs non-federated for Character Control


derekpierre:
Related to #2745. See https://github.com/nucypher/nucypher/pull/2745#discussion_r668830635.

The TreasureMap field should differentiate between non-federated and federated maps.

This issue should be worked on post Porter (#2664 ) merge since it has the foundational changes that are needed - Porter had a similar problem that was resolved by setting a schema context. 

This should be sufficient until federated / non-federated treasure maps are unified (#2737) (which entails much more significant work).

Issue #2743, opened 2021-07-10
Message kit contents


fjarri:
There are some issues with `MessageKit`s that should be resolved before 6.0 since they change its serialization format, and it's a protocol object. The issues below are not orthogonal, and answering one of them may resolve or invalidate others.

1. Should we allow unsigned message kits (`encrypt_and_sign()` called without a `signer`)? Currently this capability is only used once in test, and message kits created this way cannot be deserialized (since the splitter assumes that there is a verifying key saved). So it is unlikely that anyone is currently using this functionality.

2. Should we give a choice of signing plaintext/ciphertext to the user? In what situations would they choose one over the other? Should we do it at all, or let the user handle signing of their own data? Also note that currently if we choose to sign the ciphertext, the resulting signature is *not* serialized, so the deserialized object is in an invalid state.

3. Assuming that we do not allow unsigned messages, does it still make sense to attach the signature presence flag to the plaintext and encrypting it? 

4. The signature presence flag (if it's still required) needs to be made portable - see #2556

5. What is the proper way of handling the sender's verifying key? It is useful if one gets it from a separate channel, but if it is attached to the message, and we just check the message signature against the message key, we won't notice if the whole message kit got replaced midway. 

    Currently in the code this happens if `retrieve` is called via `BobInterface` - `Enrico` is created there from the message kit's verifying key, so the whole sender check is useless. The only way to actually check it is to call `retrieve` programmatically and pass it an `Enrico` created using a data from a separate source. But: creating a whole character just to compare two public keys seems like an overkill, and it won't work if we want to decrypt messages from several Enricos.

6. Should the `encrypt_for` interface return the message kit *and* the signature? Is the signature ever used by itself? It's already embedded into the message kit.

fjarri:
Ok, let's get things going on this. #2802 is being blocked.

First, we don't want to encrypt, then sign: ciphertext can be replaced along with the signature, and there is a possibility of signature forgery by the receiver (granted, it's been demonstrated for RSA and El Gamal, but who knows). So let's not sign the ciphertext.

Second, we don't want to include the sender's verifying key in plaintext with the kit. Since we don't sign ciphertexts, it can be easily replaced, so it doesn't really tell us anything.

We currently have three scenarios where we need to decide on what to sign/what to include with the message kit. 

## Enrico encrypts for Alice [to reencrypt]

[...]

PR #2768, opened 2021-08-11
Update Porter endpoint for retrieval

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Changes the Porter endpoint intended for part Bob retrieval process to use the updated retrieval protocol in #2730.

**Issues fixed/closed:**
Based over #2730 .
Related to #2714 , #2751 .

**Questions for Reviewers**:

1. Any issues with defining `__eq__`:
    * https://github.com/nucypher/nucypher/pull/2768#discussion_r698674580
    * https://github.com/nucypher/nucypher/pull/2768#discussion_r698675021
2. Serialization format for retrieval kits:
    * https://github.com/nucypher/nucypher/pull/2768#discussion_r701952266
    *^ If keeping the status quo, then I'll need to add documentation about the expected byte structure of the retrieval kit*


derekpierre:
@fjarri, fyi, I used a dirty hack regarding decrypted treasure maps for now to be able to properly test Porter retrieve - see https://github.com/nucypher/nucypher/pull/2768/commits/511091f6059c986218f0afdd9e5fa4efce0b1120.

nucypher/utilities/porter/control/specifications/fields/retrieve.py
```diff
@@ -0,0 +1,53 @@
...
+from nucypher.control.specifications.exceptions import InvalidInputData
+from nucypher.control.specifications.fields import Base64BytesRepresentation
+from nucypher.crypto.umbral_adapter import CapsuleFrag as CapsuleFragClass
+from nucypher.policy.kits import RetrievalKit as RetrievalKitClass
+from nucypher.utilities.porter.control.specifications.fields import UrsulaChecksumAddress
+
+
+# TODO should this be moved to character control - would this be used by a Bob character control endpoint?
+class RetrievalKit(Base64BytesRepresentation):
+    def _deserialize(self, value, attr, data, **kwargs):
```
derekpierre, line 29: @fjarri instead of serializing/deserializing based on the bytes representation of a `RetrievalKit` object, should we expect to receive the broken down representation of a RetrievalKit from its constituent parts i.e. serialized to simpler representations and deserialize can create a RetrievalKit object from those values.

Instead of
```
'retrivalkit': <base64_encoded_bytes_of_retrievalkit_object>
```

should it be:
```
'retrieval_kit': {
    'capsule': <base64_encoded_bytes_of_capsule_object>,
    'queried_addresses': [
          <ursula_address_1>,
          <ursula_address_2>,
          ...
     ]
}
```

The latter would remove the need for a RetrievalKit object and bytes representation the web/js layer.

Thoughts?

fjarri: I guess it makes sense for `RetrievalKit` since nobody would probably want to serialize it into bytes. I just wish we had schemas linked to the objects and not halfway across the codebase, like I did in [this gist](https://gist.github.com/fjarri/35dfee392699eb67290c70a00ac86dbd)

fjarri: Also JSON would make it easier to serialize errors

derekpierre: Thought a bit more about the update to the structure mentioned above (which I updated to be more specific of what the structure actually looks like). While it works well for JSON, I think the nested structure doesn't work for URL query parameters which Porter currently supports (https://github.com/nucypher/nucypher/pull/2746). It would need a flatter structure to work for URL query parameters.

Possible options could include:
1. Keep the original structure of the list of base64 encoded retrieval kits bytes and document the expected byte structure - currently the serialized byte structure for a retrieval_kit is `<bytes of capsule><byte of ursula1 address><byte of ursula2 address>...`
2. Modify the API to not take any queried_addresses, and instead just the capsule for re-encryption i.e. I think (?) that means that any retries for the same capsule will be for all of the cfrags. Example (if we do this, would probably rename `retrieval_kits` to just `capsules`):
```
'retrival_kits': [
     <base64_encoded_bytes_of_capsule_object>,
     <base64_encoded_bytes_of_capsule2_object>
     ...
]
```
...
(other alternatives?)

Thoughts?

derekpierre: > I just wish we had schemas linked to the objects and not located halfway across the codebase, like I did in this gist

I could see that being pretty beneficial. We could address this in a separate PR once the fields are finalized.

newsfragments/2772.feature.rst
```diff
@@ -1 +1 @@
-Update WorkerPool error messages returned by Porter API.
\ No newline at end of file
+Update WorkerPool error messages returned by Porter API.
```
fjarri, line 1: Huh? What changed here?

nucypher/policy/maps.py
```diff
@@ -130,6 +127,9 @@ def from_bytes(cls, data: bytes):
         destinations = {u: k for u, k in ursula_and_kfrags}
         return cls(threshold, hrac, destinations)
 
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
fjarri: I don't see any problems with it. Where is it used, in tests?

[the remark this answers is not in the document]
fjarri: One thing though: if there's a different order in `destinations`, it will lead to the equality returning `False`, while the destinations might be the same.

nucypher/utilities/porter/control/specifications/fields/retrieve.py
fjarri: It kind of depends on whether the users plan to do multi-step retrievals. It is tempting to just send capsules, I admit.

tests/integration/porter/control/test_porter_web_control_federated.py
```diff
@@ -80,52 +84,96 @@ def test_get_ursulas(federated_porter_web_controller, federated_ursulas):
...
+                                      delegating_pk=enacted_federated_policy.public_key,
+                                      receiving_pk=federated_bob.public_keys(DecryptingPower))
+        verified_cfrags[ursula] = verified_cfrag
+    retrieval_result_object = RetrievalResult(cfrags=verified_cfrags)
+    policy_message_kit = policy_message_kit.with_result(retrieval_result_object)
+
+    assert policy_message_kit.is_decryptable_by_receiver()
+
+    cleartext_with_sig_header = federated_bob._crypto_power.power_ups(DecryptingPower).keypair.decrypt(policy_message_kit)
+    sig_header, remainder = default_constant_splitter(cleartext_with_sig_header, return_remainder=True)
```
fjarri, line 147: This should really be fixed as a part of #2743. The signature header (if any) should be separated on decryption.

fjarri (approved):
Approved with some minor comments.

newsfragments/2772.feature.rst
derekpierre: A newline was missing - failed one of the pre-commit checks - https://github.com/nucypher/nucypher/pull/2768/commits/ba0c2bd056930e99d139715af0917a2127b5edfa

tests/integration/porter/control/test_porter_web_control_federated.py
derekpierre: 👍, I was basically mimicking the decryption code from here - https://github.com/nucypher/nucypher/blob/main/nucypher/characters/base.py#L426. Didn't want to use Bob's Python API to decrypt the result, to better mimic how Porter would actually be used, and ensure that it worked - https://github.com/nucypher/nucypher/pull/2768/commits/7f20fbdc7bc7a9a0aff6c3a2714dacc672bd3008.

nucypher/policy/maps.py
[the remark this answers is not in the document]
derekpierre: It is used for tests.

> if there's a different order in destinations, it will lead to the equality returning False

Good point - perhaps I'll modify it to compare `threshold`, `destinations`, and `hrac` individually instead.

newsfragments/2772.feature.rst
KPrasch: Someone is not using git hooks? 🤔

nucypher/characters/control/interfaces.py
```diff
@@ -163,27 +161,14 @@ class BobInterface(CharacterPublicInterface):
...
         """
         Character control endpoint for re-encrypting and decrypting policy data.
         """
-        from nucypher.characters.lawful import Enrico
-
-        policy_encrypting_key = PublicKey.from_bytes(policy_encrypting_key)
-        alice_verifying_key = PublicKey.from_bytes(alice_verifying_key)
-        message_kit = MessageKit.from_bytes(message_kit)  # TODO #846: May raise UnknownOpenSSLError and InvalidTag.
-
-        if isinstance(encrypted_treasure_map, bytes):
```
KPrasch, line 179: I suppose the thinking here is that deserialization must be handled by the caller?

nucypher/policy/maps.py
```diff
@@ -311,3 +312,6 @@ def from_bytes(cls, data: bytes):
         result = cls(hrac, public_signature, message_kit, blockchain_signature=blockchain_signature)
         result._public_verify()
         return result
+
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
KPrasch: No concerns from me.

tests/utils/policy.py
```diff
@@ -36,28 +40,36 @@ def generate_random_label() -> bytes:
...
-                bob_encrypting_key=encode_bytes(bob.public_keys(DecryptingPower)),
-                bob_verifying_key=encode_bytes(bob.stamp.as_umbral_pubkey()),
-                policy_encrypting_key=encode_bytes(enacted_policy.public_key),
-                )
+    if not original_message:
+        original_message = ''.join(random.choice(string.ascii_lowercase) for i in range(20)).encode()  # random message
+
+    message_kit = enrico.encrypt_message(original_message)
+
+    encode_bytes = (lambda field, obj: field()._serialize(value=obj, attr=None, obj=None)) if encode_for_rest else (lambda field, obj: obj)
```
KPrasch, line 57: 😂

nucypher/control/interfaces.py
```diff
@@ -39,12 +40,19 @@ def __init__(self, implementer=None, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     @classmethod
-    def connect_cli(cls, action):
+    def connect_cli(cls, action, exclude: Optional[Set[str]] = None):
```
KPrasch, line 43: Where is this kwarg being called?  The idea here is to provide an alternative method for CLI option exclusion?

nucypher/characters/control/interfaces.py
derekpierre: The EncryptedTreasureMap field used in the RetrieveAndDecrypt Schema, automatically handles the deserialization - https://github.com/derekpierre/nucypher/blob/porter-retrieval/nucypher/characters/control/specifications/fields/treasuremap.py#L27.

nucypher/control/interfaces.py
derekpierre: The general idea is that the schema can still be used for populating CLI options, but allow for some CLI options to be modified. 

More specifically, it is used here - https://github.com/derekpierre/nucypher/blob/porter-retrieval/nucypher/cli/commands/bob.py#L347 with the `nucypher bob retrieve-and-decrypt` CLI command. In this case, the `retrieve-and-decrypt` command from the CLI can be used without requiring an `alice_verifying_key` to be specified via `--alice-verifying-key` option which is required by the schema. An Alice card can be specified from the CLI via `--alice` and the verifying key obtained from the Card API. 

In this case `--alice-verifying-key` needs to be optional and not required. Either Alice's verifying key is provided directly via `--alice-verifying-key` or indirectly via `--alice`.

nucypher/policy/maps.py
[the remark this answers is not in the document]
derekpierre: ✔️

nucypher/utilities/porter/control/specifications/fields/retrieve.py
derekpierre: @KPrasch any thoughts here on the structure of `retrieval_kits` with respect to multi-step (if retrying, allows for only retrying for reencryptions that failed - perhaps ursula was temporarily unavailable) vs single-step (if retrying, redo all reencryptions irrespective of previous successes) retrievals?

derekpierre: Sticking with the ability to retry reencryptions without repeating already successful reencryptions. i.e. option #1.

> Keep the original structure of the list of base64 encoded retrieval kits bytes and document the expected byte structure - currently the serialized byte structure for a RetrievalKit object is `<bytes of capsule><byte of ursula1 address><byte of ursula2 address>` ...

KPrasch (approved):

PR #2769, opened 2021-08-12
[Hotfix] Remove problematic gas feed for now, and make gas feeds more robust against errors.

derekpierre:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Ignores etherchain gas feed for now since it caused issues due to EIP-1559:
```
 File "/usr/local/lib/python3.8/site-packages/nucypher/utilities/datafeeds.py", line 115, in _parse_gas_prices
  self.gas_prices = {self.get_canonical_speed(k): int(Web3.toWei(v, \'gwei\')) for k, v in self._raw_data.items()}
 File "/usr/local/lib/python3.8/site-packages/nucypher/utilities/datafeeds.py", line 115, in <dictcomp>
  self.gas_prices = {self.get_canonical_speed(k): int(Web3.toWei(v, \'gwei\')) for k, v in self._raw_data.items()}
 File "/usr/local/lib/python3.8/site-packages/nucypher/utilities/datafeeds.py", line 97, in get_canonical_speed
  raise LookupError(message)
  builtins.LookupError: \'currentBaseFee\' is not a valid speed name.'
```

Issue #2770, opened 2021-08-12
Figure out long term solution to gas price feeds in a post EIP-1559 world


derekpierre:
Related to #2769.

@cygnusv mentioned potentially getting rid of external price feeds permanently.

cygnusv:
Moving TX finality stuff outside nucypher, specifically to nucypher/atxm

derekpierre:
> Moving TX finality stuff outside nucypher, specifically to nucypher/atxm

Is there an issue to remove the exiting price feeds code from `nucypher` if they are no longer used?

PR #2771, opened 2021-08-12
Include changes and release notes from v5.3.1 hotfix in `main`

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

Related to #2769 .

No newsfragment necessary, already incorporated as part of original hotfix, and updated releases page.

**For reviewers**
Is this the best way to reconcile with `tags/v5.3.1`?

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.3.0
+current_version = 5.3.1
 commit = True
 tag = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(-(?P<stage>[^.]*)\.(?P<devnum>\d+))?
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -47,7 +47,7 @@ author = 'NuCypher'
 # The short X.Y version
 version = ''
 # The full version, including alpha/beta/rc tags
-release = '5.3.0'
+release = '5.3.1'
 
 
 # -- General configuration ---------------------------------------------------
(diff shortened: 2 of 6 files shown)


vepkenez:
The question of whether or not to merge this into main is if we can figure out #2770 before our next release right?

derekpierre:
> The question of whether or not to merge this into main is if we can figure out #2770 before our next release right?

My thought here was that we need to merge this PR now since there is currently a release that we put out (v5.3.1) that is not documented in our docs (releases page - https://docs.nucypher.com/en/latest/release_notes/releases.html), and the code isn't in `main`. 

#2770 is really a follow-up type issue. This is the workaround until #2770 and we can prioritize it as necessary moving forward; it can adjust/remove the changes made as necessary.

vepkenez:
> > The question of whether or not to merge this into main is if we can figure out #2770 before our next release right?
> 
> My thought here was that we need to merge this PR now since there is currently a release that we put out (v5.3.1) that is not documented in our docs (releases page - https://docs.nucypher.com/en/latest/release_notes/releases.html), and the code isn't in `main`.
> 
> #2770 is really a follow-up type issue. This is the workaround until #2770 and we can prioritize it as necessary moving forward; it can adjust/remove the changes made as necessary.

Ok good point!  Lets do it.

vepkenez (approved):

derekpierre:
Do we need to pick - https://github.com/nucypher/nucypher/commit/2aa615c613f43cd4ab846084c995b7b6138f530b?

derekpierre:
> Do we need to pick - 2aa615c?

Picked and force-pushed.

derekpierre:
@KPrasch is this PR sufficient?

KPrasch (approved):

PR #2772, opened 2021-08-13
Porter - return meaningful error if there are not enough Ursulas

piotr-roslaniec:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
- Currently, if there are not enough Ursulas in `reservoir`, Porter will return `0 total failures recorded` error from `concurrency.py`
- This PR checks for the number of Ursulas before contacting them; if the number of Ursulas is too low, it returns a meaningful error message to the user
- Found this to be useful while testing on Rinkeby

### DIFF
--- a/nucypher/utilities/porter/porter.py
+++ b/nucypher/utilities/porter/porter.py
@@ -134,6 +134,10 @@ the Pipe for nucypher network operations
                     exclude_ursulas: Optional[Sequence[ChecksumAddress]] = None,
                     include_ursulas: Optional[Sequence[ChecksumAddress]] = None) -> List[UrsulaInfo]:
         reservoir = self._make_staker_reservoir(quantity, duration_periods, exclude_ursulas, include_ursulas)
+
+        if len(reservoir) < quantity:
+            raise ValueError(f"Requested quantity={quantity} Ursulas, but only {len(reservoir)} are available")
+
         value_factory = PrefetchStrategy(reservoir, quantity)
 
         def get_ursula_info(ursula_address) -> Porter.UrsulaInfo:
--- a/nucypher/policy/reservoir.py
+++ b/nucypher/policy/reservoir.py
@@ -14,7 +14,7 @@ GNU Affero General Public License for more details.
 You should have received a copy of the GNU Affero General Public License
 along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
-from typing import Iterable, Optional, List
+from typing import Iterable, List, Optional
 
 from eth_typing import ChecksumAddress
 
@@ -86,6 +86,8 @@ class MergedReservoir:
         else:
             return None
 
+    def __len__(self):
+        return len(self.values) + len(self.reservoir)
 
 class PrefetchStrategy:
     """


derekpierre:
Interesting - `block_until_number_of_known_nodes_is()` typically handles that case:

Ran the following on Lynx.

```python
In [2]: payload = {'quantity': 20, 'duration_periods': 5}

In [4]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [5]: r
Out[5]: <Response [500]>

In [6]: r.text
Out[6]: "After 10 seconds and 30 rounds, didn't find 20 nodes"
```

How did you test it? It may be related to Ibex being a bit of a shit show for nodes.

derekpierre:
I see why. On Ibex:
```
In [16]: payload = {'quantity': 20, 'duration_periods': 1000}

In [17]: r = requests.get("http://localhost:9155/get_ursulas", json=payload)

In [18]: r
Out[18]: <Response [500]>

In [19]: r.text
Out[19]: '0 total failures recorded'
```

The `WorkerPool` used by Porter will return either `TimedOut` or `OutOfValues` when `block_until_target_successes()` is called - https://github.com/nucypher/nucypher/blob/main/nucypher/utilities/concurrency.py#L233.

We probably have a better shot of more generally catching those exceptions instead, and raising something more informative. Analogous situation when publishing the treasure map - https://github.com/nucypher/nucypher/blob/main/nucypher/utilities/porter/porter.py#L129.

piotr-roslaniec:
@derekpierre I see, let me rewrite this. Btw. there are like 2 good Ursulas on Ibex

nucypher/utilities/porter/porter.py
```diff
@@ -159,14 +168,25 @@ def get_ursula_info(ursula_address) -> Porter.UrsulaInfo:
                                                   learn_on_this_thread=True,
                                                   eager=True)
 
+        timeout = 1  # self.DEFAULT_EXECUTION_TIMEOUT
```
derekpierre, line 171: This was supposed to be uncommented?

```suggestion
        timeout = self.DEFAULT_EXECUTION_TIMEOUT
```

derekpierre (changes requested):

nucypher/utilities/porter/porter.py
piotr-roslaniec: Yes, that's left-over from testing, sorry about that.

derekpierre:
I had a few thoughts after thinking about this some more:
1. The messaging from the WorkerPool could be improved
2. The WorkerPool is used as part of other calls within these functions that aren't caught by this change such as `_make_staker_reservoir()` and `block_until_number_of_known_nodes_is()` so the exceptions of the pool can still creep into the response from Porter.
3. I'm concerned about returning `RuntimeError` from the Porter learner class, since technically it can also be called via the Python API and returning `RuntimeError` is faulty in that case i.e. not really a runtime error scenario if get_ursulas can't return the expected number of ursulas or it times out. Potentially we could return a `PorterException` (and/or subclasses of PorterException) that wraps any underlying execution exception that needs to be raised by Porter or something like that.  However, when I originally wrote Porter I didn't want to create a `PorterException` because I wanted similar exceptions that were thrown by individual characters Alice/Bob to be also thrown by Porter on the python side and RPC side.

Played around with modifying the WorkerPool messaging which I think was too vague to begin with and this should help the original problem you had. See the PR I created to merge into this branch - https://github.com/piotr-roslaniec/nucypher/pull/1.  Interested to get your thoughts.

derekpierre:
^ @KPrasch , @vepkenez , @fjarri feel free to weigh in as well.

derekpierre:
@piotr-roslaniec I fixed the failing concurrency tests due to my changes. Could you add a newsfragment?

Since both of us have changes in this PR, lets get others to review.

piotr-roslaniec:
I liked the changes to the WorkerPool and I think it's a step in the right direction for error handling in Porter. 

I was wondering how much Porter API consumers should know about the reasons behind API call failures. For example (correct me if I'm wrong) the response will now contain WorkerPool error messages that are fairly vague to the caller ("something timed out", "something failed"). Is it the "right amount" of information? We could provide more ("Urulas timed-out"), or less (just an HTTP error code, 404).

I prefer the latter (less complexity for the caller), with the assumptions that Porter API calls are "robust" or "idempotent" (API caller can just repeat the call without unintended consequences).

Maybe I am overthinking this. Any thoughts?

derekpierre:
> I was wondering how much Porter API consumers should know about the reasons behind API call failures. For example (correct me if I'm wrong) the response will now contain WorkerPool error messages that are fairly vague to the caller ("something timed out", "something failed"). Is it the "right amount" of information? We could provide more ("Urulas timed-out"), or less (just an HTTP error code, 404).
>
> I prefer the latter (less complexity for the caller), with the assumptions that Porter API calls are "robust" or "idempotent" (API caller can just repeat the call without unintended consequences).

I hear ya. It's a tough balance - you want to return something to the client in the error case, that provides some context but doesn't overwhelm them. The "timed out" case is simple and in most cases users will understand and trying again suffices, so saying "Timed out" type message is probably sufficient. The "out of values" case is trickier because retrying doesn't necessarily mean it will work - depends on the reason - but we don't want the entire stack trace which the WorkerPool was doing previously.

The current Porter API functions are idempotent, so retrying does not have unintended consequences, it's just whether you can identify the reason, and do something about it. It will also help us when people come to us and say Porter returned some status code, and we have no idea what to tell them because there is only a status code. Remember that clients that use a Porter service may not have access to Porter's logs because the service is being run by a separate entity.

nucypher/utilities/concurrency.py
```diff
@@ -337,6 +311,27 @@ def _produce_values(self):
...
+            failure_result = f"Execution timed out after {self._timeout}s"
+        elif result == PRODUCER_STOPPED:
+            failure_result = f"Execution stopped before completion - not enough available values"
+
+        if failures:
+            # Using one random failure
+            # Most probably they're all the same anyway.
+            value = list(failures)[0]
+            _exception_cls, exception, tb = failures[value]
+            return f"{failure_result} ({len(failures)} failures recorded); for example, {exception}"
```
fjarri, line 331: Why did you get rid of printing the traceback? I think it was pretty useful.

derekpierre: The traceback was excessive for an exception string message. If necessary, the failures can be obtained from the WorkerPool directly if needed for logging - I guess one more direct alternative to this is that we can have the `OutOfValues` and `TimedOut` exceptions store the copy of the failures dict, so the failures can be directly accessed by anything that catches those exceptions and cares about the underlying failures. This way there will still be access to the failures...I think I'll add this in.

derekpierre: @fjarri See https://github.com/nucypher/nucypher/pull/2772/commits/5bde88630e4c98eb36d9b318ac9051edb93c1be9

fjarri: The thing that worries me about this approach is that these traceback are a very useful debugging tool. Generally it's hard to understand just from the exception what actually happened in a worker. With these changes, one would have to re-run the tests with `get_tracebacks()` put in to get better debugging information.

If the traceback is excessive for a user output, why not mask the `OutOfValues`/`TimedOut` with a high-level exception? 

In fact, ideally, we shouldn't show *any* exceptions to CLI users - any error we can think of should be handled and a user-friendly message shown, with tracebacks put in the log.

derekpierre: > The thing that worries me about this approach is that these traceback are a very useful debugging tool

It is useful...in a debugging context - but debugging vs regular running of code are different paradigms. For example, the traceback can be written log file if needed, and then whoever can look at the log file if more context is needed. Alternatively, the calling code can parse the traceback themselves and modify what they want to do with the failures or how they want to represent them. Therefore, client calls can decide what they want to do with the underlying exception traceback instead of being forced to have the stack trace (of a random failure) in the exception message.

Just like regular exceptions.

> If the traceback is excessive for a user output, why not mask the OutOfValues/TimedOut with a high-level exception?

Wrapping the exception is possible but only if something useful can be extracted from the message - what if the entire stack trace is not needed. That would entail something like `e.get_message_without_stack_trace()`

> In fact, ideally, we shouldn't show any exceptions to CLI users - any error we can think of should be handled and a user-friendly message shown, with tracebacks put in the log.

Not sure what you mean by this. 

Currently, the uses of the WorkerPool are for:
    * making arrangements - where the exceptions are caught and ignored - failures are obtained from the WorkerPool itself, not the exception
    * publishing of treasure maps - where the exception is raised from Alice.grant() i.e. called by client code and therefore can be handled (logged or otherwise) as part of writing that client-side code
   
It isn't that we need to show exceptions to users, it is that the exception messages are used for context of what failed. The underlying exception's message is still included in the WorkerPool's exception message, just not the entire stack trace.

derekpierre: Further to our conversation - https://github.com/nucypher/nucypher/pull/2772/commits/23140923f504552af289989aa0577dd4a348368c

derekpierre:
@KPrasch , @fjarri , @vepkenez: ready for review. We can be even more specific for Porter, but all REST endpoints will have to possibly deal with WorkerPool exceptions so just generalizing it for now. Thoughts?

PR #2773, opened 2021-08-13
Immutable treasure maps

### LINKED ISSUE #2725: Federated has map id length 32 bytes but decentralized has length 16 bytes ... huh?

### LINKED ISSUE #2736: Descrepancy between policy ID and HRAC
Currently PolicyManager contains functions with a `policyID` input which is mirrored in higher level python (agents, actors, etc.)  PR #2687 sets the policy ID as the HRAC leaving a naming discrepancy and duplicated length constants.  I'd like to revisit the motives behind policy Id construction, and unify the result.

- Is there any advantage to using a different hash digest for HRAC versus Policy ID? 
- Perhaps there's an anonymity consideration when involving alice's stamp?
- Do we *really* need to rename on-chain variables for this purpose?

(old) policyID = `hash(bob_stamp + label)`
HRAC = `hash(alice_stamp, bob_stamp, label)`

I suggest simply moving entirely in favor of HRAC for simplicity and unifying the usage across the codebase:  HRAC can be used for all kinds of policy identification including on-chain indexing.

### LINKED ISSUE #2751: Update TreasureMap differentiation of federated vs non-federated for Character Control

fjarri:
Based on top of #2745

**Type of PR:**
- Feature
- Other

**Required reviews:** 
- 3

**What this does:**
- splits treasure maps into unencrypted and encrypted ones 
- renames `TreasureMap` schema and datastore models to `EncryptedTreasureMap` (because that's what they are)
- Mend the schism! Merges blockchain and federated treasure maps
- removes map ID in favor of just using HRAC at all times
- removes policy ID in favor of HRAC (**NOTE:** not sure about this, but doesn't seem to break anything. Please check!)
- `HRAC` is now an object instead of just a bytestring (see `policy/hrac.py`)
- fixes #2751
- fixes #2725
- fixes #2736

Note: versioning was temporarily removed, since the blockchain signature processing does not fit the "`get_splitter()`" approach; to be fixed in #2767.

**Why it's needed:**
- needed for retrieval support in Porter

**For  reviewers:**
- Any other name for `EncryptedTreasureMap`? `PublishedTreasureMap`?
- What's the format for blockchain signature in the federated case? Possible variants:
  - Optional: `b'\x00'` if no blockchain signature, `b'\x01' + signature` otherwise (current version)
  - Placeholder: `b'\x00' * 65` if no blockchain signature
- What's with the comment about #180 at `Policy.hrac` attribute? It seems out of place.


derekpierre (commented):
Haven't looked through the whole PR just yet but one thought comes to mind.

hrac feels like an internal type of representation (within nucypher code) i.e. we derive an hrac and use it for identification purposes. Exposing it as an "hrac" for example in REST requests seems overly complex to me for exposure to users. We could still refer to it as an "id" externally (i.e. policy.id and treasuremap.id), but internally (within nucypher code) we know it is an hrac.

wdyt?

nucypher/policy/hrac.py
```diff
@@ -0,0 +1,68 @@
...
+
+    This way, Bob can generate it and use it to find the TreasureMap.
+    """
+
+    # Note: this corresponds to the hardcoded size in the contracts
+    # (which use `byte16` for this variable).
+    SIZE = 16
+
+    @classmethod
+    def derive(cls, publisher_verifying_key: PublicKey, bob_verifying_key: PublicKey, label: bytes) -> 'HRAC':
```
KPrasch, line 43: 👏🏻

nucypher/characters/control/interfaces.py
```diff
@@ -192,18 +192,14 @@ def retrieve(self,
 
         self.implementer.join_policy(label=label, publisher_verifying_key=alice_verifying_key)
 
-        if self.implementer.federated_only:
-            from nucypher.policy.maps import TreasureMap as _MapClass
-        else:
-            from nucypher.policy.maps import SignedTreasureMap as _MapClass
+        from nucypher.policy.maps import EncryptedTreasureMap
```
KPrasch, line 195: Such a relief to see this conditional importing cleaned up.  Can this line be moved to the top of the module without a circular dependency?

nucypher/characters/control/interfaces.py
```diff
@@ -107,17 +107,17 @@ def grant(self,
     def revoke(self, label: bytes, bob_verifying_key: bytes) -> dict:
 
         # TODO: Move deeper into characters
-        policy_id = construct_policy_id(label, bob_verifying_key)
-        policy = self.implementer.active_policies[policy_id]
+        policy_hrac = HRAC.derive(self.implementer.stamp.as_umbral_pubkey(), bob_verifying_key, label)
+        policy = self.implementer.active_policies[policy_hrac]
```
KPrasch, line 111: An issue for another PR, but I don't think this active policy cache is used anywhere.

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -56,23 +39,7 @@ def _deserialize(self, value, attr, data, **kwargs):
...
-        if self.federated_only is not None:
-            # defined by field itself
-            is_federated_context = self.federated_only
-        else:
-            # defined by schema
-            if self.parent is not None and self.parent.context.get('federated') is not None:
-                is_federated_context = self.context.get('federated')
-
         try:
-            splitter = SignedTreasureMap.get_splitter(value) if not is_federated_context else UnsignedTreasureMap.get_splitter(value)
```
KPrasch, line 73: 👍🏻

nucypher/characters/lawful.py
```diff
@@ -942,19 +932,19 @@ def _get_cleartexts(self,
     def _handle_treasure_map(self,
                              publisher_verifying_key: PublicKey,
                              label: bytes,
-                             treasure_map: Optional['TreasureMap'] = None,
+                             enc_treasure_map: Optional['EncryptedTreasureMap'] = None,
```
KPrasch, line 935:
```suggestion
                             encrypted_treasure_map: Optional['EncryptedTreasureMap'] = None,
 ```

File: nucypher/characters/lawful.py [hunk 6]
```diff
@@ -942,19 +932,19 @@ def _get_cleartexts(self,
...
                              ) -> 'TreasureMap':
-        if treasure_map is not None:
-            self._try_orient(treasure_map, publisher_verifying_key)
-            # self.treasure_maps[treasure_map.public_id()] = treasure_map # TODO: Can we?
+        if enc_treasure_map is not None:
```
KPrasch, line 937: Unless this can possibly be assigned to a CS constant, this can be a simple falsy check.

nucypher/characters/lawful.py
```diff
@@ -965,7 +955,7 @@ def retrieve(self,
                  *message_kits: UmbralMessageKit,
                  label: bytes,
                  policy_encrypting_key: Optional[PublicKey] = None,
-                 treasure_map: Optional['TreasureMap'] = None,
+                 treasure_map: Optional['EncryptedTreasureMap'] = None,
```
KPrasch, line 958: `encrypted_treasure_map`

nucypher/network/server.py
```diff
@@ -286,16 +287,22 @@ def revoke():
         # TODO: Implement offchain revocation.
         return Response(status=200)
 
-    @rest_app.route('/treasure_map/<identifier>')
-    def provide_treasure_map(identifier):
+    @rest_app.route('/treasure_map/<hrac>')
+    def provide_treasure_map(hrac):
```
KPrasch, line 291: Soon to be deprecated 😈

nucypher/network/server.py
```diff
@@ -286,16 +287,22 @@ def revoke():
...
         headers = {'Content-Type': 'application/octet-stream'}
+
+        try:
+            hrac_obj = HRAC.from_bytes(bytes.fromhex(hrac))
+        except ValueError:
+            return Response(f"Invalid HRAC format (got: {hrac}).", status=401)
```
KPrasch, line 297:
```suggestion
            return Response(f"Invalid HRAC format (got: {hrac}).", status=400)  # bad request
```

nucypher/policy/policies.py
```diff
@@ -537,43 +527,40 @@ def _publish_to_blockchain(self, ursulas) -> dict:
...
-        self.id = id # TODO: is it even used anywhere?
         self.hrac = hrac
         self.label = label
         self.public_key = public_key
         self.treasure_map = treasure_map
         self.treasure_map_publisher = treasure_map_publisher
         self.revocation_kit = revocation_kit
-        self.n = len(self.treasure_map.destinations)
+        self.m = m
+        self.n = len(self.revocation_kit)
```
KPrasch, line 563: Slightly awkward.  Needs to be revisited with the next iteration on Revocation.

nucypher/policy/maps.py
```diff
@@ -16,29 +16,131 @@
...
+    def __init__(self,
+                 m: int,
+                 hrac: HRAC,
+                 destinations: Dict[ChecksumAddress, bytes],
+                 ):
+        self.m = m
+        self.destinations = destinations
+        self.hrac = hrac
+
+    def prepare_for_publication(self,
```
KPrasch, line 108:
```suggestion
    def encrypt(self,
```

nucypher/policy/maps.py
```diff
@@ -103,329 +205,112 @@ def from_bytes(cls, data: bytes):
...
-        """Raised when a treasure map's version is too old or contents are incompatible."""
-
-    ursula_and_kfrag_payload_splitter = BytestringSplitter(
-        (to_checksum_address, ETH_ADDRESS_BYTE_LENGTH),
+    _splitter = BytestringSplitter(
+        signature_splitter,
+        hrac_splitter,
         (UmbralMessageKit, VariableLengthBytestring),
-    )
+        (bytes, 1))
```
KPrasch, line 214: a comment will be useful here.

nucypher/policy/maps.py
```diff
@@ -16,29 +16,131 @@
...
+
+
+class TreasureMap:
+
+    class NowhereToBeFound(RestMiddleware.NotFound):
+        """
+        Called when no known nodes have it.
+        """
+
+    class IsDisorienting(Exception):
```
KPrasch, line 50: Perhaps split this into two exceptions, `DecryptionFailed` and `Corrupted`.

nucypher/policy/maps.py
```diff
@@ -103,329 +205,112 @@ def from_bytes(cls, data: bytes):
...
-
-    def __len__(self):
-        return len(self.destinations)
-
-    def __bytes__(self):
-        if self._payload is None:
-            self._set_payload()
-        return self._BRAND + self._VERSION + self._payload
+    @staticmethod
+    def _make_blockchain_signature(blockchain_signer: Callable[[bytes], bytes],
```
KPrasch, line 220:
```suggestion
    def _sign(blockchain_signer: Callable[[bytes], bytes],
```

Perhaps this can be combined with the federated signing method too, appending all zeros to match the format.

nucypher/policy/maps.py
```diff
@@ -103,329 +205,112 @@ def from_bytes(cls, data: bytes):
...
-        else:
-            raise self.InvalidSignature("This TreasureMap is not properly publicly signed by Alice.")
+        self.hrac = hrac
+        self._public_signature = public_signature
+        self._verifying_key = encrypted_tmap.sender_verifying_key
+        self._encrypted_tmap = encrypted_tmap
+        self._blockchain_signature = blockchain_signature
 
-    def orient(self, compass: Callable):
+    def orient(self, compass: Callable[[bytes], Signature]) -> TreasureMap:
```
KPrasch, line 267:
```suggestion
    def decrypt(self, func: Callable[[bytes]]) -> TreasureMap:
```

nucypher/policy/maps.py
```diff
@@ -103,329 +205,112 @@ def from_bytes(cls, data: bytes):
...
-
-    def verify_blockchain_signature(self, checksum_address):
-        self._set_payload()
-        return verify_eip_191(message=self._payload,
+    def __bytes__(self):
+        return (bytes(self._public_signature) +
+                bytes(self.hrac) +
+                bytes(VariableLengthBytestring(self._encrypted_tmap.to_bytes())) +
+                # Other variants:
+                # - store all zeros if no blockchain signature
```
KPrasch, line 284: I think the zero-padding approach is the best way forward.  Then we can follow up with a federated/offchain version of signing.

nucypher/policy/maps.py
```diff
@@ -103,329 +205,112 @@ def from_bytes(cls, data: bytes):
...
+                # - store the flag too in addition to all zeros, for added sanity check
+                # - use a faux blockchain signer in federated case
+                (b'\x00' if self._blockchain_signature is None else (b'\x01' + bytes(self._blockchain_signature)))
+                )
+
+    def verify_blockchain_signature(self, checksum_address: ChecksumAddress) -> bool:
```
KPrasch, line 290: Adding to the above comment, if an offchain EIP-191 signature can be implemented, this method can simply be `verify`.

nucypher/characters/control/interfaces.py
fjarri: It is used once in `AliceInterface.revoke()`

derekpierre:
@fjarri , @KPrasch any thoughts here - https://github.com/nucypher/nucypher/pull/2773#pullrequestreview-730747852

nucypher/characters/control/specifications/fields/treasuremap.py
```diff
@@ -56,23 +39,7 @@ def _deserialize(self, value, attr, data, **kwargs):
             raise InvalidInputData(f"Could not parse {self.name}: {e}")
 
     def _validate(self, value):
```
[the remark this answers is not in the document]
KPrasch: Where is hrac exposed publicly that gives you concern?

derekpierre:
> Where is hrac exposed publicly that gives you concern?

I haven't looked through the entire PR yet - but of particular concern was the Porter REST API using 'hrac' as a REST parameter - https://github.com/fjarri/nucypher/blob/tmaps/tests/integration/porter/control/test_porter_web_control_federated.py#L106 (schema: https://github.com/fjarri/nucypher/blob/tmaps/nucypher/utilities/porter/control/interfaces.py#L71)

fjarri:
@derekpierre , regarding ID/HRAC: my position is that it would be better to have the same name throughout the code. I don't feel particularly strongly whether it's ID or HRAC in the end; HRAC is a little easier to look up in the code. 

The endpoint you cited is probably going away anyway - @KPrasch is thinking about getting rid of storing the treasure map on Ursula side. In this case, we won't even need to send the HRAC to Ursula.

nucypher/characters/lawful.py
```diff
@@ -624,17 +624,17 @@ def follow_treasure_map(self,
 
         return unknown_ursulas, known_ursulas, treasure_map.m
 
-    def _try_orient(self, treasure_map, publisher_verifying_key):
+    def _try_orient(self, encrypted_treasure_map, publisher_verifying_key):
         # TODO: should be rather a Publisher character
         alice = Alice.from_public_keys(verifying_key=publisher_verifying_key)
-        compass = self.make_compass_for_alice(alice)
+        decryptor = self.make_decryptor_for_alice(alice)
```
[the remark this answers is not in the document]
fjarri: Removed that method completely - now `_try_orient()` and `make_decryptor_for_alice()` are joined in `_decrypt_treasure_map()`

KPrasch (approved):

nucypher/policy/maps.py
fjarri: Filed #2778 . Depending on how it is handled, we can change the method names.

fjarri:
Regarding the schema changes - would it be better handled in #2768?

derekpierre:
> Regarding the schema changes - would it be better handled in #2768?

Sure. I guess that would save me from an even more rough rebase.

derekpierre:
> @derekpierre , regarding ID/HRAC: my position is that it would be better to have the same name throughout the code. I don't feel particularly strongly whether it's ID or HRAC in the end; HRAC is a little easier to look up in the code.
>
> The endpoint you cited is probably going away anyway - @KPrasch is thinking about getting rid of storing the treasure map on Ursula side. In this case, we won't even need to send the HRAC to Ursula.

@fjarri fair enough.

Updated PR looks good - do we need a newsfragment for this PR? Currently, it is empty.

fjarri:
CI fails:
- probationary period ending
- https://github.com/nucypher/bytestringSplitter/pull/42

derekpierre (approved):
🎸

PR #2774, opened 2021-08-15
Rename `m` to `threshold` and `n` to `shares`

fjarri:
Based on top of #2773

**Type of PR:**
- Documentation

**Required reviews:** 
- 2

**What this does:**
Gets rid of vague one-letter names for the threshold and the total number of kfrags.

Trying to gauge the amount of work that needs to be done.


KPrasch (commented):
I like the direction we're going in here, but I think we can improve on `num_kfrags`.  Here's some ideas we've had collectively to consider:

- overhead
- redundancy
- total
- shares

Personally, I think `overhead` is an excellent choice to continue the API metaphor here (it's intuitive and descriptive)  but am also not opposed to keeping `n`.

fjarri:
See also https://github.com/nucypher/rust-umbral/issues/48

newsfragments/2774.doc.rst
```diff
@@ -0,0 +1 @@
+Changed the names of `m` and `n` parameters to `threshold` and `num_kfrags`.
```
derekpierre, line 1: I think either we add an additional `removal` newsfragment or this might need to be changed to a `removal` newsfragment, since we are deprecating/modifying public-facing APIs that is potentially currently used:
1.  Character control: "m" and "n" parameters can no longer be used
2. CLI: "--m" and "--n" are no longer valid -> "-m" and "-n"
3. Python API: no longer takes "m" and "n" as parameter.

All in favour of `threshold` and `num_kfrags'

derekpierre (changes requested):
Looks good - but requesting addition/update to newsfragment for a `removal` newsfragment.

I quite like "shares" (and I'm *ok* with "num_kfrags") as opposed to "overhead". Until now, we've had "n" representing the total number of kfrags, changing that to be "threshold + overhead" seems a bit heavy to me, and would be less natural for existing users. Aside from that, saving users from doing the math (albeit simple math 😄 ) I think is worth it - particularly when you think about how we tend to explain the threshold concept "m-of-n", "threshold out of a number of shares" etc.

fjarri:
Alright, I think the consensus is to go with "shares". I updated the newsfragments and added a removal one. Waiting for CI to pass.

KPrasch (approved):
🤠

vepkenez (approved):
❤️

tests/metrics/grant_availability.py
```diff
@@ -75,8 +75,8 @@
 TEMP_ALICE_DIR: Path = Path('/', 'tmp', 'grant-metrics')
 
 # Policy Parameters
-M: int = 1
-N: int = 1
+THRESHOLD: int = 1
+NUM_KFRAGS: int = 1
```
derekpierre, line 79: Could probably change this to "SHARES", but...only if you want to...

derekpierre (approved):
🎸  - minor comment - @fjarri up to you if you want to update the constant in the test.

fjarri:
It actually was already renamed, but it was `NUM_KFRAGS`. Changed that to `SHARES`.

PR #2779, opened 2021-08-19
Extend the policy probation period.

KPrasch:


derekpierre:
@KPrasch we'll need to include a 'misc' newsfragment as we've done before when changing the probationary period - for example https://docs.nucypher.com/en/latest/release_notes/releases.html#v5-3-0-2021-06-17

derekpierre:
Should we do at least 2 months (maybe 3) - it would bring us more in line with keanu (@cygnusv ?)

cygnusv:
> Should we do at least 2 months (maybe 3) - it would bring us more in line with keanu (@cygnusv ?)

Yep, 2 months at least.

newsfragments/2779.misc.rst
```diff
@@ -0,0 +1 @@
+Extends policy probationary period until October 30th, 2021. No policies may be created on the network beyond this date.
```
derekpierre, line 1: Since October has 31 days.

```suggestion
Extends policy probationary period until October 31st, 2021. No policies may be created on the network beyond this date. 
```

nucypher/config/constants.py
```diff
@@ -77,4 +77,4 @@
 
 
 # Probationary period (see #2353, #2584)
-END_OF_POLICIES_PROBATIONARY_PERIOD = MayaDT.from_iso8601('2021-08-31T23:59:59.0Z')
+END_OF_POLICIES_PROBATIONARY_PERIOD = MayaDT.from_iso8601('2021-10-30T23:59:59.0Z')
```
derekpierre, line 80:
```suggestion
END_OF_POLICIES_PROBATIONARY_PERIOD = MayaDT.from_iso8601('2021-10-31T23:59:59.0Z')
```

nucypher/config/constants.py
```diff
@@ -77,4 +77,4 @@
 
 
 # Probationary period (see #2353, #2584)
```
derekpierre, line 79: We could get rid of this comment altogether, but if keeping - 

```suggestion
# Probationary period (see #2779)
```

KPrasch: ✔️

nucypher/config/constants.py
KPrasch: ✔️

newsfragments/2779.misc.rst
KPrasch: ✔️

derekpierre (approved):
🎸

PR #2785, opened 2021-08-26
Prometheus Memory

derekpierre:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

Investigation of potential prometheus memory issues.

*For reviewers*
It would be great if someone with a worklock allocation could run their node briefly with prometheus to ensure the worklock collectors work appropriately - worklock collectors are not tested on testnets.

### DIFF
--- a/newsfragments/2785.bugfix.rst
+++ b/newsfragments/2785.bugfix.rst
@@ -1,0 +1,1 @@
+Minor memory improvement when collecting staker/worker metrics for prometheus.
--- a/nucypher/blockchain/eth/events.py
+++ b/nucypher/blockchain/eth/events.py
@@ -98,8 +98,8 @@ class ContractEventsThrottler:
         self.to_block = to_block if to_block is not None else agent.blockchain.client.block_number
         # validity check of block range
         if self.to_block < self.from_block:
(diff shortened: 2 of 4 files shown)


KPrasch:
> collectors were holding on to references of Contract Agent instances

Why do you think this causes growing memory consumption?

derekpierre:
> Why do you think this causes growing memory consumption?

It's not that I thought this causes the memory consumption, it was just something that I noticed and thought was weird. My memory testing pre and post change did not yield any noticeable change - and that was expected. Figured it was good to improve the design anyway.

KPrasch:
Ah, gotcha.  So this PR doesn't really have anything to do with prometheus memory usage, rather it's just an adjustment to collectors?

derekpierre:
> Ah, gotcha. So this PR doesn't really have anything to do with prometheus memory usage, rather it's just an adjustment to collectors?

👍  - yep, the memory investigation didn't lead anywhere other than me discovering this non-associated issue.

KPrasch (approved):

PR #2798, opened 2021-09-14
Updates for Umbral 0.3

fjarri:
**Type of PR:**
- Other

**Required reviews:** 
- 2

**What this does:**
Updates the usage of Umbral, bumps its version to 0.3 and relocks dependencies.

### DIFF
--- a/tests/unit/test_datafeeds.py
+++ b/tests/unit/test_datafeeds.py
@@ -16,7 +16,7 @@ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
 
-from collections import Callable
+from collections.abc import Callable
 from statistics import median
 from unittest.mock import patch
 
--- a/Pipfile
+++ b/Pipfile
@@ -12,7 +12,7 @@ constant-sorrow = ">=0.1.0a9"
 bytestring-splitter = ">=2.4.0"
 hendrix = ">=3.4"
 lmdb = "*"
-umbral = ">=0.2"
+umbral = ">=0.3"
 # Cryptography
 pyopenssl = "*"
 cryptography = ">=3.2"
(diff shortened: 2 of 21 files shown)


KPrasch (approved):
🤠

deploy/docker/Dockerfile
```diff
@@ -2,7 +2,7 @@ FROM python:3.8.7-slim
 
 # Update
 RUN apt update -y && apt upgrade -y
-RUN apt install patch gcc libffi-dev wget -y
+RUN apt install patch gcc libffi-dev wget git -y
```
derekpierre, line 5: Could you do the same for `deploy/docker/porter/Dockerfile` as well?

tests/unit/test_datafeeds.py
```diff
@@ -16,7 +16,7 @@
 """
 
 
-from collections import Callable
+from collections.abc import Callable
```
derekpierre, line 19: This is for python 3.9 support?

derekpierre (commented):
We can probably add something about using Umbral 0.3 to the newsfragment.

deploy/docker/Dockerfile
fjarri: It's just a temporary measure, because I'm currently pulling `umbral` from a git commit. It will be removed when I set it to ==0.3

derekpierre: Oh I see - 👍

tests/unit/test_datafeeds.py
fjarri: Just to get rid of the warning that's been bugging me for months. It will be required in 3.10 though.

derekpierre (approved):
🎸

PR #2801, opened 2021-09-21
build and deploy monitor with ansible

vepkenez:
**Type of PR:**
- [x] Feature

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
Adds an automated build workflow for an existing ssh accessible cloud ubuntu host to deploy and fully configure a working nucypher-monitor instance.

**Why it's needed:**
Having standardized configs for our Ursulas, Eth nodes and now the monitor is nice.

### DIFF
--- a/deploy/ansible/monitor/include/run_monitor.yml
+++ b/deploy/ansible/monitor/include/run_monitor.yml
@@ -1,0 +1,52 @@
+- name: "Run Ursula"
+  hosts: "{{ play_hosts }}"
+  remote_user: "{{default_user}}"
+  gather_facts: no
+  tasks:
+    - name: Clone Monitor Repo
+      become: yes
+      become_user: nucypher
+      ansible.builtin.git:
+        repo: https://github.com/nucypher/nucypher-monitor.git
+        dest: /home/nucypher/nucypher-monitor
+        recursive: yes
+        force: yes
+
+    - name: setup certs dir
+      become: yes
+      ansible.builtin.file:
+        path: "{{NUCYPHER_LETSENCRYPT_DIR}}"
+        state: directory
+        owner: nucypher
+        group: nucypher
+        mode: '0700'
+
+    # - name: Clean any existing Influx DB directory
+    #   ansible.builtin.file:
+    #     path: /home/nucypher/influxdb
+    #     state: absent
+
+    - name: Copy SSL Certs
+      become: yes
+      ansible.builtin.copy:
+        src: "{{LOCAL_CERTS_DIR }}/{{item}}"
+        dest: "{{NUCYPHER_LETSENCRYPT_DIR}}/{{item}}"
+        owner: nucypher
+        group: nucypher
+        mode: '0700'
+      loop:
+        - cert.pem
+        - chain.pem
+        - fullchain.pem
+        - privkey.pem
+
+    - name: Run Monitor (docker-compose up)
+      become: yes
+      become_user: nucypher
+      ansible.builtin.command: docker-compose -f /home/nucypher/nucypher-monitor/deploy/docker-compose.yml up -d --build
+      args:
+        chdir: /home/nucypher/
+      environment:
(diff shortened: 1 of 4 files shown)


derekpierre:
@vepkenez thanks for starting this 🙏 .

We should put these scripts in the `nucypher-monitor` repos instead - we've endeavoured to keep monitor-specific code out of the `nucypher` codebase. The status monitor is more like an app that uses `nucypher` and not innate to `nucypher` itself.

vepkenez:
> We should put these scripts in the `nucypher-monitor` repos instead - we've endeavoured to keep monitor-specific code out of the `nucypher` codebase. The status monitor is more like an app that uses `nucypher` and not innate to `nucypher` itself.

The problem is that they rely heavily on the existing ansible node setup that we use for Ursula.  So maybe we should move _all_ of the ansible deployment stuff out of nucypher.  Does that seem like a good idea?  I would just need to add a thing to cloudworkers which would download it.  This would actually solve some other issues (the one where you need to git checkout v5.3.1 and not the latest main commit to use cloudworkers).

On the other hand, if you look at the current state of [this directory](https://github.com/nucypher/nucypher/tree/main/deploy/ansible), with respect to this discussion, it seems like we could hold our nose and add this in...

derekpierre:
I'm less familiar with ansible so take my comments with a pinch of salt.

> The problem is that they rely heavily on the existing ansible node setup that we use for Ursula. So maybe we should move all of the ansible deployment stuff out of nucypher. Does that seem like a good idea? I would just need to add a thing to cloudworkers which would download it. This would actually solve some other issues (the one where you need to git checkout v5.3.1 and not the latest main commit to use cloudworkers).

I'm intrigued about moving it to something external to `nucypher`, but have some questions:
1. How does moving it to something external affect the ease of use by users? Sounds like it would just be to get the playbooks from a separate repos which may not be too bad. Would users interact with that external repos directly, or only indirectly via the `nucypher cloudworkers` CLI? If directly, there would be this amalgamation of various playbooks in the repos that users would see, and could lead to some confusion - most users know nothing about the `nucypher-monitor` side of things.
2. Seeing the `moe`, `felix` folders in that folder already has my spider-sense tingling 🕷️ 
3. Instead of moving "everything", are there generic/abstract components/functionalities/playbooks that can be extracted from what we have, that can then be used/referenced by playbooks in both `nucypher` and `nucypher-monitor`? If possible, this could be pretty clean.

> On the other hand, if you look at the current state of this directory, with respect to this discussion, it seems like we could hold our nose and add this in...

Heh, I don't know if I can hold my nose tight enough to escape that stench 👃 🧄 🤮

vepkenez:
I think I could make it so cloudworkers worked as is by having it download the ansible stuff from a new nucypher-ansible repo.
I'll give it a shot.

Then we could easily just have a "go get the nucypher ansible stuff" script for the monitor repo as well.

KPrasch:
How many reviews do you want on this one? 😆 

Screenshot from 2021-09-21 14-03-51

vepkenez:
Definitely not 2!

KPrasch (approved):

derekpierre:
Did we want to merge this - I thought @vepkenez was working on a different strategy?

KPrasch:
Oh, is that so? I cannot not tell by looking at this PR that it was intended to be do not merge, it looked open and waiting for review with outstanding comments.  Let's use `[WIP]`, `[Do Not Merge]` or Draft mode to indicate this in the future!

What shall I do to remedy this situation, revert and close?

vepkenez:
Yeah I'm working on something but it's fine for now.  It won't hurt anything.

derekpierre:
@KPrasch, you are right that a resulting solution/step forward was not adequately reflected in the comments of PR  after the discussion - we should have updated it. I should have also added the `request changes` as part of the review.

In any case, perhaps we can have a reversion PR...

PR #2780, opened 2021-08-19
[WIP] Deprecate treasure map storage

KPrasch:
**Type of PR:**
Removal

**Required reviews:** 
2

**What this does:**
Removes treasure map storage functionality.

**Issues fixed/closed:**
WIP

**Why it's needed:**
WIP

**Notes for reviewers:**
WIP

PR #2804, opened 2021-09-30
Remove unused method `Amonia.use_ursula_as_an_involuntary_and_unbeknownst_cdn`

fjarri:
**Type of PR:**
- Other

**Required reviews:** 
- 1

**What this does:**
Removes an unused method (used in a test prior to #2780, but the test was removed since it dealt with storing treasure maps on Ursulas)


newsfragments/2804.misc.rst
```diff
@@ -0,0 +1 @@
+Remove an unused method of ``Amonia`` (deprecated since we do not store tmap on Ursulas anymore)
```
derekpierre, line 1: `removal` newsfragment instead?

fjarri: My impression was that "removal" is for public API deprecations, but sure, I'll change to it

derekpierre: I believe `Amonia` is technically a public API - although presumably would not be used by anyone other than us for testing.

fjarri: Fixed

derekpierre (approved):
🎸

KPrasch (approved):

PR #2805, opened 2021-09-30
Wider lanes

fjarri:
**Type of PR:**
- Feature

**Required reviews:** 
- 2

**What this does:**
- Extend the brand size in `Versioned` to 4 bytes to make it easier to resolve name clashes in the future.
- Allow all regexp `\w` characters in brands.

The exact values for new 4-byte brands are open for suggestions.

### DIFF
--- a/nucypher/policy/kits.py
+++ b/nucypher/policy/kits.py
@@ -122,7 +122,7 @@ class MessageKit(Versioned):
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'MK'
+        return b'MKit'
 
     @classmethod
     def _version(cls) -> Tuple[int, int]:
@@ -181,7 +181,7 @@ class RetrievalKit(Versioned):
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'RK'
+        return b'RKit'
 
     @classmethod
     def _version(cls) -> Tuple[int, int]:
--- a/nucypher/policy/maps.py
+++ b/nucypher/policy/maps.py
@@ -105,7 +105,7 @@ class TreasureMap(Versioned):
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'TM'
+        return b'TMap'
 
     @classmethod
     def _version(cls) -> Tuple[int, int]:
@@ -205,7 +205,7 @@ class AuthorizedKeyFrag(Versioned):
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'KF'
+        return b'AKF_'
 
     @classmethod
     def _version(cls) -> Tuple[int, int]:
@@ -330,7 +330,7 @@ class EncryptedTreasureMap(Versioned):
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'EM'
+        return b'EMap'
 
     @classmethod
     def _version(cls) -> Tuple[int, int]:
--- a/nucypher/policy/revocation.py
+++ b/nucypher/policy/revocation.py
@@ -74,7 +74,7 @@ class RevocationOrder(Versioned):
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'RV'
+        return b'Revo'
 
     @classmethod
     def _version(cls) -> Tuple[int, int]:
--- a/nucypher/utilities/versioning.py
+++ b/nucypher/utilities/versioning.py
@@ -17,16 +17,17 @@
 
 
 from abc import abstractmethod, ABC
+import re
 from typing import Dict, Tuple, Callable
 
 
 class Versioned(ABC):
     """Base class for serializable entities"""
 
-    _PARTS = 2  # bytes
-    _PART_SIZE = 2
-    _BRAND_SIZE = 2
-    _VERSION_SIZE = _PART_SIZE * _PARTS
+    _VERSION_PARTS = 2
+    _VERSION_PART_SIZE = 2  # bytes
+    _BRAND_SIZE = 4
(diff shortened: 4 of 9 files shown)


nucypher/policy/kits.py
```diff
@@ -122,7 +122,7 @@ def _payload(self) -> bytes:
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'MK'
+        return b'MK  '
```
derekpierre, line 125: I'm ok with spaces either way, but just throwing it out there that underscores could be used instead -it would make the 4-bytes a little more obvious, and it is also already part of the `\w` match for regex.

nucypher/utilities/versioning.py
```diff
@@ -135,7 +136,7 @@ def _parse_brand(cls, data: bytes) -> bytes:
         brand = data[:cls._BRAND_SIZE]
         if brand != cls._brand():
             error = f"Incorrect brand. Expected {cls._brand()}, Got {brand}."
-            if not brand.isalpha():
+            if not re.match(rb'^[\w\d ]+', brand):
```
derekpierre, line 139: We should be careful here:
1. Is `\d` redundant when `\w` is used?
2. Because of the `+` this could yield partial matches for example the first x bytes (for x < 4) are alphanumeric_or_space, but the rest are not; but what we want is all 4 bytes are alphanumeric_or_space. 

For example, the following incorrect brands would pass the check.
```python
In [97]: brand = b'MK&%'

In [98]: re.match(rb'^[\w\d ]+', brand)
Out[98]: <re.Match object; span=(0, 2), match=b'MK'>

In [99]: brand = b'M?&%'

In [100]: re.match(rb'^[\w\d ]+', brand)
Out[100]: <re.Match object; span=(0, 1), match=b'M'
```

Therefore, should the regex instead be:

```suggestion
            if not re.fullmatch(rb'^[\w ]+', brand):
```

derekpierre: Side note: we should add a partially valid brand to `test_versioning_invalid_brand` analogous to the ones illustrated above.

derekpierre (changes requested):

nucypher/policy/kits.py
fjarri: Yes, fair enough, we can just use a single `\w`. Fixed.

nucypher/utilities/versioning.py
fjarri: Both good points, added `fullmatch` and a test.

tests/unit/test_versioning.py
```diff
@@ -116,39 +116,45 @@ def test_versioning_header_prepend():
...
 
 def test_versioning_invalid_brand():
-    invalid = b'\x00\x03\x00\x0112'
+    invalid = b'\x01\x02\x00\x03\x00\x0112'
+    with pytest.raises(Versioned.InvalidHeader, match="Incompatible bytes for A."):
+        A.from_bytes(invalid)
+
+    # A partially invalid brand, to check that the regexp validates
+    # the whole brand and not just the beginning of it.
+    invalid = b'ABC \x00\x02\x00\x0112'
```
derekpierre, line 137: 🔥

nucypher/policy/kits.py
```diff
@@ -122,7 +122,7 @@ def _payload(self) -> bytes:
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'MK'
+        return b'MK__'
```
KPrasch, line 125:
```suggestion
        return b'MKit'
```

derekpierre (approved):
🎸

nucypher/policy/kits.py
```diff
@@ -181,7 +181,7 @@ def _payload(self) -> bytes:
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'RK'
+        return b'RK__'
```
KPrasch, line 184:
```suggestion
        return b'RKit'
```

nucypher/policy/maps.py
```diff
@@ -105,7 +105,7 @@ def construct_by_publisher(cls,
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'TM'
+        return b'TM__'
```
KPrasch, line 108:
```suggestion
        return b'TMap'
```

nucypher/policy/revocation.py
```diff
@@ -74,7 +74,7 @@ def verify_signature(self, alice_verifying_key: PublicKey) -> bool:
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'RV'
+        return b'RO__'
```
KPrasch, line 77:
```suggestion
        return b'Revo'
```

nucypher/policy/maps.py
```diff
@@ -330,7 +330,7 @@ def _payload(self) -> bytes:
 
     @classmethod
     def _brand(cls) -> bytes:
-        return b'EM'
+        return b'ETM_'
```
KPrasch, line 333:
```suggestion
        return b'EMap'
```

KPrasch (commented):
Just some branding suggestions.

nucypher/utilities/versioning.py
```diff
@@ -25,7 +26,7 @@ class Versioned(ABC):
 
     _PARTS = 2  # bytes
```
KPrasch, line 27: Just noticed this comment doesn't make sense on this line.  Perhaps just move it down inline with `_BRAND_SIZE`.

PR #2806, opened 2021-09-30
Remove additions from PR #2801 - the intended feature will be completed using a different strategy

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
Removes addition of nucypher-monitor ansible scripts to `nucypher` made via #2801 (sorry @vepkenez 😑 ). The PR was mistakenly merged, and a different strategy will be explored - https://github.com/nucypher/nucypher/pull/2801#issuecomment-929624909.

There is no newsfragment for this PR on purpose, since it undoes the work done in #2801.


deploy/ansible/worker/include/setup_docker.yml
```diff
@@ -39,25 +39,6 @@
         name: nucypher
         group: docker
 
-    - name: Download docker-compose
```
KPrasch, line 42: This code is in the worker directory, is it also unused?

derekpierre: I think (?) this code was added in order to have a common code/location to obtain `docker-compose`. Once obtained it can then be used with the monitor - I'm not sure that docker-compose was utilized anywhere else - @vepkenez ?

Basically this removal matches the additions made here - https://github.com/nucypher/nucypher/pull/2801/commits/cb8f40e6bc4ea19fe26f9f21d19fda042df23544#diff-43c64472a931bf0fb6bc4f21d485f2e3584c45ccb554dfda94fd17153f9c69c9R42 - unless it is still useful.

KPrasch (approved):

vepkenez:
I don't really see the point of this.  We would be taking something that works, that I'm actually using, and making it not work.   Yes I realize this code was merged by mistake, but it doesn't actually hurt or affect anything and having an actual programatic deployment for the monitor is a good thing.

Lets wait until we have a replacement for this that also cleans up all the other junk in the deploy directory, which we should do when that is the most important thing to do.  

I can't really approve this PR because I think it's a bad idea but if you want to remove all this... fine, I'll just keep my own branch with this useful feature that very much improved our monitor deployment that this PR removes.

derekpierre:
@vepkenez , based on our discussion in #2801 we talked about the code feeling out of place in `nucypher` especially since `nucypher-monitor` is an entirely different repos. You mentioned moving common ansible stuff outside of nucypher and possibly downloading it via cloudworkers CLI (https://github.com/nucypher/nucypher/pull/2801#issuecomment-924311477). If so, what's the point of keeping the code? If you want to keep it for now, a branch sounds like a good idea. We don't really want to do a release with this code in it, do we?

Sure it is convenient for us (at NuCypher), but adds unrelated extra bloat to the `nucypher` repos that then gets distributed to others.

How dependent is the ansible script on ursula ansible stuff? Looks like it is `setup_docker` and `setup_user`; any others?These seem like great candidates associated with your comment about common ansible scripts previously in #2801 .

I'm all for improving the monitor deployment - how can we better do that without the `nucypher` ansible dependencies? It doesn't seem that we currently publish a nucypher-monitor docker image, perhaps we should start doing that (save the docker-compose step). Would that be helpful?

vepkenez:
> Sure it is convenient for us (at NuCypher), but adds unrelated extra bloat to the `nucypher` repos that then gets distributed to others.

It actually doesn't get distributed except when you checkout the repo.  The ansible.yml files are not included in what gets `pip installed`

I guess what I'm saying is, lets remove this stuff _with_ the PR that reorganizes it.  Otherwise when, for example, I need to update the ssl certs on the monitor, there is no longer a documented way to do that and we are back to custom bespoke servers that only one person knows about as we were a few weeks ago.  That was bad.  

Right now the ansible stuff that is removed by this PR works with data in the knowledge-base. I completely agree with everything you say, but I am against actually merging this until there is an alternative documented method of maintaining and updating the monitor.

vepkenez:
I have always felt that all of deploy/ should be a separate repo as well as our contract deploy code.

deploy/ansible/worker/include/setup_docker.yml
vepkenez: As noted below, I am planning on potentially using docker-compose for future cloudworkers.  It is useful to keep.

derekpierre:
> I guess what I'm saying is, lets remove this stuff with the PR that reorganizes it. Otherwise when, for example, I need to update the ssl certs on the monitor, there is no longer a documented way to do that and we are back to custom bespoke servers that only one person knows about as we were a few weeks ago. That was bad.

I'm with you, *if* the PR wasn't merged - unfortunately it was, so now the code (that aims to be changed) is already in main. In any case, you weren't going to merge the PR, and were going to leave the code in that branch anyway until we got around to updating everything. Perhaps it can better be served as an `[WIP] EPIC` branch PR with your original commit cherry-picked? That way the code is clearly available and set aside from `main`?

KPrasch:
While this PR raises some larger code organization concerns we can still move forward.  Let's move the nucypher-monitor automation over to nucypher/nucypher-monitor.

vepkenez:
I vote that we close this PR and I will remove all the offending code and more deprecated deployment configs that nobody ever uses in a future PR that is in progress.

KPrasch:
> I vote that we close this PR and I will remove all the offending code and more deprecated deployment configs that nobody ever uses in a future PR that is in progress.

Sounds like a plan!  We're overdue for some cleanup in that area.  Closing on this basis.

PR #2809, opened 2021-10-14
Nucypher-Core part 2 (more ABI changes)

### LINKED ISSUE #2743: Message kit contents

fjarri:
Breaking changes:
- Instead of just signing the interface in `NodeMetadata`, sign the whole metadata. As a part of that, `interface_signature` and `timestamp` attributes were removed. `timestamp` is now a metadata attribute, since it's recorded when the metadata is created.
- `hrac`, `publisher_verifying_key`, and `_public_signature` are no longer bundled with `EncryptedTreasureMap`.
- `publisher_verifying_key` is added to `ReencryptionRequest`.
- `MessageKit` does not sign anything now, it's the user's responsibility.
- Added `EncryptedKeyFrag`, a separate versioned entity hosting an encrypted `AuthorizedKeyFrag`.
- Added `AuthorizedTreasureMap`, a separate versioned entity hosting an "unverifed" treasure map (similarly to how the pair `AuthorizedKeyFrag`-`VerifiedKeyFrag` works.
- Using the whole keyfrag to create a signature in `AuthorizedKeyFrag`, not just its checksum.
- The brand of `AuthorizedKeyFrag` changed from `AKF_` to `AKFr`
- Normalized the usage of `VariableLengthBytestring`. Now every class knows how to deserialize itself from the stream, instead of relying on the higher level class to provide it with a bytestring of correct length. 
- Denoting optional fields with a flag instead of a `VariableLengthBytestring`.

In particular, fixes #2743:
- `MessageKit` is just plain capsule + ciphertext
- `AuthorizedKeyFrag` (which is a plaintext for `EncryptedKeyFrag`) has its main payload (kfrag) + HRAC signed by the publisher
- `AuthorizedTreasureMap` (which is a plaintext for `EncryptedTreasureMap`) has its main payload (treasure map) + recipient key signed by the publisher
- Removed `alice_verifying_key` from `ReencryptionRequest` (as a part of #2812)

Other changes:
- Removed some unused imports
- Removed passing Alice's signing power to Enrico in `Enrico.from_alice()`, since Enrico doesn't sign anything anymore.
- Fixed an argument name in `Ursula._decrypt_kfrag()` - the verifying key passed to it is supposed to be Publisher's
- `UnbondedWorker` is not mysteriously caught and ignored in `verify_node()`, but instead propagated to be caught in `learn_from_teacher_node()` like the rest of exceptions.
- Fixed treasure map caching in `Bob`
- Moved `ReencryptionResponse` verification from `RetrievalClient._request_reencryption()` to `ReencryptionResponse.verify()`
- Normalized parameter names and order in protocol object constructors.
- Using `http.HTTPStatus` enum instead of hardcoded status codes.

**For reviewers:**
- Need to decide on #2812, as a result `publisher_verifying_key` may be removed from `TreasureMap`, or a cross-check added to see if it matches the HRAC
- See the TODO in `AuthorizedKeyFrag.verify()` - should we have a forced verification method for `KeyFrag` in Umbral?
- At the moment HRAC is not saved in `AuthorizedKeyFrag`, and the recipient key is not saved in `AuthorizedTreasureMap`. The externally passed value is used for verification. Another option is to store them in the respective objects (still passing an external value, of course) - this creates some redundancy, but may simplify error checking. A parallel can be drawn also with how capsules are signed in `ReencryptionResponse`, but not included in it. Thoughts?
- Make a trivial versioned plaintext class for `MessageKit`, to enable possible updates?
- Rename `MessageKit` to `EncryptedMessage` for the uniformity with kfrags and treasure maps?


fjarri:
The last commit attempts to normalize the `VariableLengthBytestring` usage in protocol objects (this was the remaining TODO item). The problem was that variable size objects and lists were treated differently in different classes. Some classes were relying on the higher-level class to supply a bytestring of correct length to `from_bytes()`, and just assumed that all that remained is the variable length part. Sometimes a class actually wrapped the variable-length part in `VariableLengthBytestring`. Sometimes it was both. 

So in that commit the guidelines were:
- Every class knows how to deserialize itself from the given bytestring, returning the unused remainder
- `Versioned` was changed to support that, adding `take()` and requiring deserialization handlers return the object *and* the remainder of the bytestring
- Whenever a list occurs in a class, it is wrapped in `VariableLengthBytestring`. This is done for both lists of types with constant sizes and lists of types with variable sizes, for uniformity (an alternative for constant sized types would be specifying the number of items). Since all objects know how to deserialize themselves from the stream, it is not necessary to wrap every element of the list in `VariableLengthBytestring` - we just `take()` until we're out of bytes.

The disadvantage of the `take()` approach is that you can't just skip a variable-sized field and deserialize it later (lazily) - you have to deserialize everything. I think that for protocol objects it is not a problem.

One small change to be added is a simpler support for optional fields - we just need a byte flag instead of wrapping them in a `VariableLengthBytestring`.

nucypher/network/server.py
```diff
@@ -169,14 +168,12 @@ def reencrypt():
         reenc_request = ReencryptionRequest.from_bytes(request.data)
         hrac = reenc_request.hrac
         bob = Bob.from_public_keys(verifying_key=reenc_request.bob_verifying_key)
-        log.info(f"Work Order from {bob} for policy {hrac}")
+        log.info(f"Reencryption request from {bob} for policy {hrac}")
 
         # Right off the bat, if this HRAC is already known to be revoked, reject the order.
         if hrac in this_node.revoked_policies:
-            return Response(response="Invalid KFrag sender.", status=401)  # 401 - Unauthorized
+            return Response(response="Policy with {hrac} has been revoked.", status=HTTPStatus.UNAUTHORIZED)
```
derekpierre, line 175:
```suggestion
            return Response(response=f"Policy with {hrac} has been revoked.", status=HTTPStatus.UNAUTHORIZED)
```

KPrasch (approved):
Already left all the comments and questions I have here.  LGTM.  I don't know if it really counts as a "part 3" but let's move onto the next PR for any follow-ups.  As discussed the tasks mentioned in the "for reviewers" section need to be resolve before the next major release.

fjarri:
The TODOs were filed as issues:

> See the TODO in AuthorizedKeyFrag.verify() - should we have a forced verification method for KeyFrag in Umbral?

Filed https://github.com/nucypher/rust-umbral/issues/73

> At the moment HRAC is not saved in AuthorizedKeyFrag, and the recipient key is not saved in AuthorizedTreasureMap. The externally passed value is used for verification. Another option is to store them in the respective objects (still passing an external value, of course) - this creates some redundancy, but may simplify error checking. A parallel can be drawn also with how capsules are signed in ReencryptionResponse, but not included in it. Thoughts?

Filed https://github.com/nucypher/nucypher/issues/2816

> Make a trivial versioned plaintext class for MessageKit, to enable possible updates?

Filed https://github.com/nucypher/nucypher/issues/2817

> Rename MessageKit to EncryptedMessage for the uniformity with kfrags and treasure maps?

Filed https://github.com/nucypher/nucypher/issues/2818

PR #2811, opened 2021-10-19
Include changes and release notes from v5.3.2 hotfix in `main`

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

Related to #2810.
Don't include regeneration of certs commit since deterministic keystore work in `main` always regenerates the cert on startup.
No newsfragment necessary, already incorporated as part of original hotfix.

### DIFF
--- a/docs/source/release_notes/releases.rst
+++ b/docs/source/release_notes/releases.rst
@@ -4,6 +4,21 @@ Releases
 
 .. towncrier release notes start
 
+v5.3.2 (2021-10-15)
+-------------------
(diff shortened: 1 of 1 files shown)


KPrasch:
Thanks for following-up on this.  We can make patch releases a bit easier to perform by making a few tweaks in the `Makefile`.

PR #2815, opened 2021-11-01
Remove the remnants of the old metadata versioning scheme

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

**What this does:**
Fixes the error handling to catch any problems with seed nodes without crashing the learning loop. 

**For reviewers:**
Would it be useful to print the traceback in the log messages? (see TODOs)

### DIFF
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -32,7 +32,6 @@ from constant_sorrow.constants import (
     NOT_SIGNED,
     NO_STORAGE_AVAILABLE,
     RELAX,
-    UNKNOWN_VERSION
 )
 from cryptography.x509 import Certificate, load_pem_x509_certificate
 from cryptography.hazmat.backends import default_backend
@@ -329,7 +328,6 @@ class Learner:
         self._learning_round = 0  # type: int
         self._rounds_without_new_nodes = 0  # type: int
         self._seed_nodes = seed_nodes or []
-        self.unresponsive_seed_nodes = set()
 
         if self.start_learning_now and not self.lonely:
             self.start_learning_loop(now=self.learn_on_same_thread)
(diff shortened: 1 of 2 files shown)


nucypher/network/nodes.py
```diff
@@ -359,36 +357,35 @@ def load_seednodes(self, read_storage: bool = True, record_fleet_state=False):
...
+
+            self.log.debug(f"Seeding from: {node_tag}")
+
+            try:
+                seed_node = self.node_class.from_seednode_metadata(seednode_metadata=seednode_metadata,
+                                                                   network_middleware=self.network_middleware,
+                                                                   )
+            except Exception as e:
+                # TODO: log traceback here?
+                # TODO: distinguish between versioning errors and other errors?
```
KPrasch, line 382: This is essentially the purpose of the `NodeSeemsToBeDown` tuple, otherwise this may be overtly generic.  I do think it's a good idea to have this catch-all for `Exception`.

KPrasch (commented):
This is a good entry point for seed nodes...   It's possible the same or similar problem may still occur in general node peering  A quick investigation is warranted to ensure we have the proper test for this.

nucypher/network/nodes.py
fjarri: We could check for that first specifically, but the action would be the same - just log it and move on. So I am not sure it is necessary at this stage. Maybe if we bucketed the node behavior, there would be a difference between an unresponsive node, a node with incompatible version, and a node returning some trash.

KPrasch (approved):
Considering this change will unblock near-term developments while we migrate our infrastructure, I'm approving 👍🏻 

Thanks for the quick turn-around on this.

PR #2819, opened 2021-11-03
Generate metadata if the timestamp is requested

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

**What this does:**
Make `Ursula.timestamp` available regardless of whether `metadata()` has been called before or not.


derekpierre (approved):
🎸

KPrasch (approved):

PR #2820, opened 2021-11-06
Fix a memory leak in Ursula

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 2

**What this does:**
- Scraps `suspicious_activities_witnessed` from Ursula. We're not doing anything with it (until #567 is implemented), and it's clogging the memory over time. It is still logged, and I modified the corresponding tests to check the logs instead.
- Limits the number of retained fleet states (to 5, which is what the status page needs). Older states are discarded as the new ones are pushed.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -1,5 +1,5 @@
 [bumpversion]
-current_version = 5.3.2
+current_version = 5.3.3
 commit = True
 tag = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(-(?P<stage>[^.]*)\.(?P<devnum>\d+))?
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -47,7 +47,7 @@ author = 'NuCypher'
 # The short X.Y version
 version = ''
 # The full version, including alpha/beta/rc tags
-release = '5.3.2'
+release = '5.3.3'
 
 
 # -- General configuration ---------------------------------------------------
--- a/docs/source/release_notes/releases.rst
+++ b/docs/source/release_notes/releases.rst
@@ -4,6 +4,15 @@ Releases
 
 .. towncrier release notes start
 
+v5.3.3 (2021-11-24)
+-------------------
+
+Bugfixes
+~~~~~~~~
+
+- Fixed a memory leak in Ursula; removed some teacher statistics accumulated over time. (`#2826 <https://github.com/nucypher/nucypher/issues/2826>`__)
+
+
 v5.3.2 (2021-10-15)
 -------------------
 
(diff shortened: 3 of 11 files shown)


derekpierre (approved):
🎸 

Do you view this change as something widespread enough that we should do a patch release, or should we just wait until v6.0 is published? I'm unsure of the current timing for v6.0.

fjarri:
Judging by Discord it affects some people, and apparently @vepkenez encountered it too. I personally was never bothered, probably because my node has more RAM than recommended, and I update it often enough. For me, the memory usage starts from 4% and the most I've seen is about 20% or so.

I lean towards making a release (although it's easy for me to say, since it's not me who will be doing that :), but I would like to be sure that it's the only source of the leak first.

vepkenez:
Wow.  This is so obviously a one way ticket to a memory leak now that you found it.  There isn't even really _any_ way to access that list on a running node.  Definitely lets turn this off!

vepkenez (approved):

vepkenez:
@derekpierre I think I will deploy this branch to our NuCypher nodes because they tend do die every couple of weeks if left on their own.  This could be the issue.  Especially in the case of nodes that are "seeds", where they are more likely to encounter issues that get added to that `suspicious_activities_witnessed` list.

derekpierre:
@vepkenez / @fjarri Any indications of the effectiveness of this code change on memory usage? You both were testing it on various nodes?

fjarri:
I've been running it on my mainnet node for a week, and it still has roughly the same memory footprint.

derekpierre:
@fjarri / @vepkenez - sounds like testing has been good...? If so, should we issue a release - presumably, a v5.3.3 based over v5.3.2 with these commits? (fyi @KPrasch)

vepkenez:
> @fjarri / @vepkenez - sounds like testing has been good...? If so, should we issue a release - presumably, a v5.3.3 based over v5.3.2 with these commits? (fyi @KPrasch)

Yes.  A pre 6.0 release of this would be good.

KPrasch:
I'm trying to make a release on this now, but it's non-trivial considering that this changeset is based on `main` and not `5.3.2` and there have been divergences.

derekpierre:
> I'm trying to make a release on this now, but it's non-trivial considering that this changeset is based on main and not 5.3.2 and there have been divergences.

@KPrasch - see #2826 .

KPrasch:
Included in v.5.3.3 patch release 9f7abd2283aca0a1b81ed4d3d02cd7971577570c

KPrasch:
Reopening to include changes in future maintained releases

derekpierre (approved):
🎸

PR #2822, opened 2021-11-18
Halting inflation

vzotova:
**Type of PR:**
- [X] Feature

**Required reviews:** 
- [X] 3

**What this does:**
`StakingEscrow` update only. Removes most of the code from `StakingEscrow`: halting inflation, removing all methods to control stake (including withdraw), exclude `PolicyManager` and `Adjudicator` from lifecycle (for now)

**Why it's needed:**
First step in merging networks

**Notes for reviewers:**
Only tests for contracts have been adjusted. Some code for `PolicyManger` and `Adjudicator` was not removed to use it together with `PREApplication` contract later

PR #2825, opened 2021-11-23
Integrate `StakingEscrow` with `TokenStaking`

vzotova:
**Type of PR:**
- [X] Feature

**Required reviews:** 
- [X] 3

**What this does:**
Update `StakingEscrow` to work with `TokenStaking`

**Notes for reviewers:**
Test upgradeability won't work because different versions of solidity also we assume that `StakingEscrow` will be upgraded to PRs #2831 and #2822 before upgrading to this PR.

### DIFF
--- a/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
@@ -6,7 +6,6 @@ pragma solidity ^0.8.0;
 import "aragon/interfaces/IERC900History.sol";
 import "contracts/NuCypherToken.sol";
 import "contracts/lib/Bits.sol";
-import "contracts/lib/Snapshot.sol";
 import "contracts/proxy/Upgradeable.sol";
 import "zeppelin/math/Math.sol";
 import "zeppelin/token/ERC20/SafeERC20.sol";
@@ -72,7 +71,6 @@ contract StakingEscrowStub is Upgradeable {
 contract StakingEscrow is Upgradeable, IERC900History {
 
     using Bits for uint256;
-    using Snapshot for uint128[];
     using SafeERC20 for NuCypherToken;
 
     /**
@@ -141,12 +139,11 @@ contract StakingEscrow is Upgradeable, IERC900History {
 
         uint256[] stub8; // former slot for pastDowntime
         uint256[] stub9; // former slot for subStakes
-        uint128[] history; // TODO keep or remove?
+        uint128[] stub10; // former slot for history
 
     }
 
     // indices for flags (0-4 were in use, skip it in future)
-//    uint8 internal constant SNAPSHOTS_DISABLED_INDEX = 3;
     uint8 internal constant MERGED_INDEX = 5;
 
     uint256 internal constant ACCEPTABLE_STAKING_ERROR = 10**15;
@@ -164,13 +161,13 @@ contract StakingEscrow is Upgradeable, IERC900History {
     mapping (address => address) private stub3; // former slot for stakerFromWorker
 
     mapping (uint16 => uint256) private stub4; // former slot for lockedPerPeriod
-    uint128[] public balanceHistory;  // outdated
+    uint128[] private stub5;  // former slot for balanceHistory
 
-    address private stub5; // former slot for PolicyManager
-    address private stub6; // former slot for Adjudicator
-    address private stub7; // former slot for WorkLock
+    address private stub6; // former slot for PolicyManager
+    address private stub7; // former slot for Adjudicator
+    address private stub8; // former slot for WorkLock
 
-    mapping (uint16 => uint256) private stub8; // last former slot for lockedPerPeriod
+    mapping (uint16 => uint256) private stub9; // last former slot for lockedPerPeriod
 
     /**
     * @notice Constructor sets address of token contract and parameters for staking
@@ -305,7 +302,6 @@ contract StakingEscrow is Upgradeable, IERC900History {
         );
         info.value -= _value;
 
-        addSnapshot(info, - int256(_value));
         token.safeTransfer(msg.sender, _value);
         emit Withdrawn(msg.sender, _value);
     }
@@ -431,16 +427,10 @@ contract StakingEscrow is Upgradeable, IERC900History {
     //------------------ ERC900 connectors ----------------------
 
     function totalStakedForAt(address _owner, uint256 _blockNumber) public view override returns (uint256) {
-        if (isUpgrade == UPGRADE_TRUE) {
-            return stakerInfo[_owner].history.getValueAt(_blockNumber);
-        }
         return 0;
     }
 
     function totalStakedAt(uint256 _blockNumber) public view override returns (uint256) {
-        if (isUpgrade == UPGRADE_TRUE) {
-            return balanceHistory.getValueAt(_blockNumber);
-        }
         return token.totalSupply();
     }
 
--- a/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
@@ -297,7 +297,7 @@ contract StakingEscrow is Upgradeable, IERC900History {
             "Not enough tokens unstaked in T staking contract"
         );
         require(
-            _value + getVestedTokens(msg.sender) <= info.value,
+            _value + getUnvestedTokens(msg.sender) <= info.value,
             "Not enough tokens released during vesting"
         );
         info.value -= _value;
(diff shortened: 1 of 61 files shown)


nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -258,6 +284,106 @@ contract StakingEscrow is Upgradeable, IERC900History {
...
+            "Not enough tokens released during vesting"
+        );
+        info.value -= _value;
+
+        token.safeTransfer(msg.sender, _value);
+        emit Withdrawn(msg.sender, _value);
+    }
+
+    /**
+    * @notice Returns amount of not released yet tokens for staker
```
derekpierre, line 310: This comment is a little confusing - makes it sound more like unvested as opposed to vested.

Is this analogous to what you intended?

```suggestion
    * @notice Returns tokens available for release to the staker based on vesting.
```

OR

```suggestion
    * @notice Returns tokens available to the staker based on vesting.
```

Unsure if there were already terminology discussions amongst yourselves?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -258,6 +284,106 @@ contract StakingEscrow is Upgradeable, IERC900History {
...
+        return info.value;
+    }
+
+    /**
+    * @notice Confirm migration to threshold network
+    * @param _staker Staker address
+    */
+    function confirmMerge(address _staker) external {
+        StakerInfo storage info = stakerInfo[_staker];
+        require(info.operator != address(0), "Staker didn't request merge");
```
derekpierre, line 374: Is "staker didn't request merge" the only outcome of the address is non-zero i.e. is it also possible that the merge already happened, or is this case prevented somewhere before this function is called?

nucypher/blockchain/eth/sol/source/threshold/IStaking.sol
```diff
@@ -0,0 +1,358 @@
...
+    ///         the owner or operator.
+    function topUpNu(address operator) external;
+
+    //
+    //
+    // Undelegating a stake (unstaking)
+    //
+    //
+
+    /// @notice Reduces the liquid T stake amount by the provided amount and
```
derekpierre, line 178: Just checking :
*"Reduces the liquid T stake amount"* - are we referring to staked T as *"liquid" T stake"*?

derekpierre (commented):
Still working my way through.

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: Terminology is not set and it looks like I did mistake during naming. Appreciate any help
So method doc is proper which means that this tokens are still locked. `getUnvestedTokens` ?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: `requestMerge` can be called only from T `TokenStaking` during `stakeNu`/`topUpNu`, only these two roots. `requestMerge` is only one entry point so I do not see other way to do it

nucypher/blockchain/eth/sol/source/threshold/IStaking.sol
vzotova: I think so, because there is no difference between KEEP/NU and T stakes except staker can withdraw T stake

tests/contracts/main/staking_escrow/test_staking_escrow.py
```diff
@@ -170,3 +174,371 @@ def test_slashing(testerchain, token, worklock, escrow):
...
+        tx = escrow.functions.confirmMerge(staker).transact()
+        testerchain.wait_for_receipt(tx)
+
+
+def test_withdraw(testerchain, token, worklock, threshold_staking, escrow):
+    creator, staker, operator = testerchain.client.accounts[0:3]
+    withdrawal_log = escrow.events.Withdrawn.createFilter(fromBlock='latest')
+
+    # Deposit some tokens
+    value = NU(ONE_HOUR, 'NU').to_nunits()  # Exclude rounding error
```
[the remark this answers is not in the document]
vzotova: I just need some number that will be divided by ONE_HOUR without remainder. Should I add one more variable to make it more clear?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
derekpierre: From what I gather I think it might just be a terminology thing. 

So based on the line:

```python
uint256 vestedTokens = (info.vestingReleaseTimestamp - block.timestamp) * info.vestingReleaseRate;
        return info.value < vestedTokens ? info.value : vestedTokens;
```

Does 'release' mean the tokens have vested and are now available to be withdrawn? If so, there are still two parts to the calculation: 1) how many tokens have vested (`vestedTokens`) 2) how many tokens are currently staked `info.value`. 

So the value returned is really either the tokens vested, or the amount of tokens remaining because perhaps some tokens have already been withdrawn. Is that correct?

If the above is correct, the value returned is the tokens that can be withdrawn by the staker based on the vesting schedule - or am I missing something.

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
derekpierre: Ah! I misread that line.

tests/contracts/main/staking_escrow/test_staking_escrow.py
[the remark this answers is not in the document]
derekpierre: 👍 hehe yeah I figured. Instead of a variable a comment could suffice - but up to you, I'm fine either way. it just jumped out at me, but not a problem.

nucypher/blockchain/eth/sol/source/threshold/IStaking.sol
derekpierre: Something being liquid/unlocked vs staked was different in our old terminology, but perhaps this has changed for Threshold.

I was wondering if the comment was more like "Reduces the T stake amount..."

vzotova: yeah, this interface is just a copy from Threshold repo
Main emphasis was to distinguish stake from legacy Keep/Nu and from liquid T

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: Renamed to `unvested`

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -258,6 +284,109 @@ contract StakingEscrow is Upgradeable, IERC900History {
...
+    */
+    function requestMerge(address _staker, address _operator)
+        external onlyTStakingContract returns (uint256)
+    {
+        StakerInfo storage info = stakerInfo[_staker];
+        require(
+            info.operator == address(0) ||
+            info.operator == _operator ||
+            (tStaking.stakedNu(info.operator) == 0 &&
+            !info.flags.bitSet(MERGED_INDEX)),
```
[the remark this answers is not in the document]
arjunhassard: The point of this is to make it more equitable between those who have vested tokens and those who have staked but non-vested tokens, by restricting the latter's ability to re-request a merge with a new operator address? That makes sense, particularly as doing so may have implications for their eligibility to authorize to tbtcv2 later. In general the more uniform the action space is for stakers (with different situations) the better, mostly to avoid irreversible moves where the benefits are massively outweighed by the costs, but the staker isn't aware of those downsides.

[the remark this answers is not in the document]
vzotova: Not really. The thing is after `confirmMerge` staker can't restake and change oeprator without withdrawing all NU. So for non-vested stakers it's easy (because they can withdraw everything in that case), but for vested-stakers it won't be possible before everything is unlocked. So the solution (for vested stakers) is - do not call `confirmMerge` before you're sure in operator address. And my question related to: it's not necessery to restrict like that (only before `confirmMerge`) but I have itchy feeling not to do that

cygnusv (approved):

derekpierre (approved):
🎸

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -298,67 +391,13 @@ contract StakingEscrow is Upgradeable, IERC900History {
...
-    * @notice Return the length of the array of sub stakes
-    */
-    function getSubStakesLength(address _staker) external view returns (uint256) {
-        return stakerInfo[_staker].subStakes.length;
-    }
-
-    /**
-    * @notice Return the information about sub stake
-    */
-    function getSubStakeInfo(address _staker, uint256 _index)
```
KPrasch, line 311: 💥

KPrasch (approved):

PR #2826, opened 2021-11-23
Fix Ursula Memory Leak Issue (v5.3.3 Candidate)

derekpierre:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Redid changes from #2820 over the tags/v5.3.2 branch in order to allow for a 5.3.3 release.

Originally authored by @fjarri .

**Issues fixed/closed:**
Related to #2820 

**Why it's needed:**
There are many conflicts trying to cherry-pick changes from #2820 onto 5.3.2.

**Notes for reviewers:**

### DIFF
--- a/newsfragments/2810.bugfix.rst
+++ b/newsfragments/2810.bugfix.rst
@@ -1,1 +1,0 @@
-Regenerate Ursula TLS certificates if the become invalid, e.g. become expired.
--- a/docs/source/release_notes/releases.rst
+++ b/docs/source/release_notes/releases.rst
@@ -4,6 +4,21 @@ Releases
 
 .. towncrier release notes start
 
+v5.3.2 (2021-10-15)
+-------------------
+
+Bugfixes
+~~~~~~~~
+
+- Regenerate Ursula TLS certificates if the become invalid, e.g. become expired. (`#2810 <https://github.com/nucypher/nucypher/issues/2810>`__)
+
+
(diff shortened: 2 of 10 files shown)


newsfragments/2810.bugfix.rst
```diff
@@ -1 +0,0 @@
-Regenerate Ursula TLS certificates if the become invalid, e.g. become expired.
```
fjarri, line 1: What's up with the removed newsfragments? Mis-commit?

fjarri (approved):

piotr-roslaniec (approved):

newsfragments/2810.bugfix.rst
derekpierre: There was some issue with the release the last time - see https://discord.com/channels/411401661714792449/412681039178366996/899655352767889469, where the release note that was supposed to be generated by the newsfragment wasn't included for some reason for the v5.3.2 release, so the newsfragment was never removed. 

That reminds me, I should add the release note for 5.3.2 to the releases docs page as part of this PR, or else it won't be included at all in the releases page i.e. there would only be release notes for 5.3.1 and 5.3.3. I've updated the release page now.

PR #2827, opened 2021-11-24
Include expiration timestamp in Kfrag authorization

### LINKED ISSUE #2813: Include expiration timestamp in kfrag authorization
In an effort to align federated and blockchain operation and increase error-resistance of reencryption requests, Alice can sign the policy expiration timestamp as part of the kfrag's authorization or "writ".  This can be checked by ursula just before performing reencyption for freshness and if needed prevent service completely offchain.  It's unclear what impact if any this may have on on-chain timestamp verification.

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
- Alice signs a 4-byte expiration timestamp
- After decrypting the writ, Ursula uses the signed timestamp as part of the kfrag validity check 

**Issues fixed/closed:**
Closes #2813 

**Why it's needed:**
Increases robustness of kfrag validity check and time-based policies without incurring any additional on-chain reads.  Since this is a signature-based approach this check is also fully compatible with federated node operation.


derekpierre (commented):
If the expiration check is purely off-chain i.e. we don't compare the expiry date of the kfrag to that of the on-chain policy, is it possible for the Publisher to do some nefarious/odd things (whether on purpose or not) involving having the kfrag expiration and the policy expiration be different...?

piotr-roslaniec (approved):

derekpierre (changes requested):
Formalize my prior comments.

nucypher/characters/lawful.py
```diff
@@ -282,10 +282,10 @@ def generate_policy_parameters(self,
         if self.federated_only:
             if not expiration:
                 raise TypeError("For a federated policy, you must specify expiration (payment_periods don't count).")
-            if expiration <= maya.now():
+            if expiration <= maya.now().epoch:
```
fjarri, line 285: `expiration` is of type `MayaDT` in the method signature, should it be `int` instead?

nucypher/characters/unlawful.py
```diff
@@ -167,7 +167,7 @@ def publish_wrong_payee_address_to_blockchain(policy, ursulas):
                 policy_id=bytes(policy.hrac),  # bytes16 _policyID
                 transacting_power=policy.publisher.transacting_power,
                 value=policy.value,
-                end_timestamp=policy.expiration.epoch,  # uint16 _numberOfPeriods
+                end_timestamp=policy.expiration,  # uint16 _numberOfPeriods
```
fjarri, line 170: Is the comment outdated?

nucypher/core.py
```diff
@@ -296,12 +316,14 @@ def __init__(self,
...
                  publisher_verifying_key: PublicKey,
-                 destinations: Dict[ChecksumAddress, EncryptedKeyFrag]):
+                 destinations: Dict[ChecksumAddress, EncryptedKeyFrag],
+                 expiration: int):
         self.threshold = threshold
         self.destinations = destinations
         self.hrac = hrac
         self.policy_encrypting_key = policy_encrypting_key
         self.publisher_verifying_key = publisher_verifying_key
+        self.expiration = expiration
```
fjarri, line 326: Do we really need to keep it in the TreasureMap too? I don't think it is used anywhere.

KPrasch:
Returning to WIP/Draft due to significant base divergence

fjarri:
Probably should be an issue in `nucypher-core`

piotr-roslaniec:
Has this change been implemented elsewhere or maybe superseded? I can't find it in  `nucypher-core`.

PR #2828, opened 2021-11-24
Keystore.generate can return optional mnemonic words

xela7:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
> I've added a field report_mnemonic that is by default False, so as to not affect any other tests.  When set to True, when interactive is set to False, generate method returns both keystore and mnemonic.   

> When interactive is True and report_mnemonic is also True, a ValueError is raised.

**Why it's needed:**
> Users who are using a function that calls Keystore.generate are not able to ever see their mnemonic and cannot store.

> For example, when a website calls a wrapper around nucypher code to generate a keystore, the user of that website can never see their mnemonic words the way code is currently structured.  With code change, wrapper code can call generate and return mnemonic __words to end user.

**Notes for reviewers:**
> Is name report_mnemonic OK?  I understand that code returns __words and not the mnemonic itself.

> Please pardon my branch name. Should have been something like: keystore-generate-report-mnemonic-words

### DIFF
--- a/nucypher/config/base.py
+++ b/nucypher/config/base.py
@@ -799,9 +799,15 @@ class CharacterConfiguration(BaseConfiguration):
                                                      password=password,
                                                      keystore_dir=self.keystore_dir)
         else:
-            self.__keystore, _ = Keystore.generate(password=password,
-                                                   keystore_dir=self.keystore_dir,
-                                                   interactive=interactive)
+            if interactive:
+                self.__keystore = Keystore.generate(password=password,
+                                                    keystore_dir=self.keystore_dir,
+                                                    interactive=interactive)
+            else:
+                self.__keystore, _ = Keystore.generate(password=password,
+                                                       keystore_dir=self.keystore_dir,
+                                                       interactive=interactive)
+
         return self.keystore
 
     @classmethod
--- a/nucypher/config/base.py
+++ b/nucypher/config/base.py
@@ -799,9 +799,9 @@ class CharacterConfiguration(BaseConfiguration):
                                                      password=password,
                                                      keystore_dir=self.keystore_dir)
         else:
-            self.__keystore = Keystore.generate(password=password,
-                                                keystore_dir=self.keystore_dir,
-                                                interactive=interactive)
+            self.__keystore, _ = Keystore.generate(password=password,
+                                                   keystore_dir=self.keystore_dir,
+                                                   interactive=interactive)
         return self.keystore
 
     @classmethod
--- a/nucypher/crypto/keystore.py
+++ b/nucypher/crypto/keystore.py
@@ -346,10 +346,10 @@ class Keystore:
             cls, password: str,
             keystore_dir: Optional[Path] = None,
             interactive: bool = True,
-            report_mnemonic: bool = False
+            return_mnemonic: bool = False
             ) -> Union['Keystore', Tuple['Keystore', str]]:
         """Generate a new nucypher keystore for use with characters"""
-        if report_mnemonic and interactive:
+        if return_mnemonic and interactive:
             raise ValueError("The two values: report_mnemonic and interactive,  may not both be `True`")
 
         mnemonic = Mnemonic(_MNEMONIC_LANGUAGE)
@@ -360,7 +360,7 @@ class Keystore:
         path = Keystore.__save(secret=__secret, password=password, keystore_dir=keystore_dir)
         keystore = cls(keystore_path=path)
 
-        if report_mnemonic:
+        if return_mnemonic:
             return keystore, __words
 
         return keystore
--- a/nucypher/crypto/keystore.py
+++ b/nucypher/crypto/keystore.py
@@ -346,12 +346,8 @@ class Keystore:
             cls, password: str,
             keystore_dir: Optional[Path] = None,
             interactive: bool = True,
-            return_mnemonic: bool = False
             ) -> Union['Keystore', Tuple['Keystore', str]]:
         """Generate a new nucypher keystore for use with characters"""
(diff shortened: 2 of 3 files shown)


nucypher/crypto/keystore.py
```diff
@@ -342,15 +342,27 @@ def restore(cls, words: str, password: str, keystore_dir: Optional[Path] = None)
...
+
         mnemonic = Mnemonic(_MNEMONIC_LANGUAGE)
         __words = mnemonic.generate(strength=_ENTROPY_BITS)
         if interactive:
             cls._confirm_generate(__words)
         __secret = bytes(mnemonic.to_entropy(__words))
         path = Keystore.__save(secret=__secret, password=password, keystore_dir=keystore_dir)
         keystore = cls(keystore_path=path)
+
+        if report_mnemonic:
```
KPrasch, line 363: Aside from being literal, I think the name `return_mnemonic` will help avoid confusion with the existing reporting (console echo) behavior.

nucypher/crypto/keystore.py
```diff
@@ -342,15 +342,27 @@ def restore(cls, words: str, password: str, keystore_dir: Optional[Path] = None)
         return keystore
 
     @classmethod
-    def generate(cls, password: str, keystore_dir: Optional[Path] = None, interactive: bool = True) -> 'Keystore':
+    def generate(
+            cls, password: str,
+            keystore_dir: Optional[Path] = None,
+            interactive: bool = True,
+            report_mnemonic: bool = False
+            ) -> 'Keystore':
```
KPrasch, line 350: Given the new optional return type this type annotation is no longer accurate.

```suggestion
            ) -> Union['Keystore', Tuple['Keystore', str]]:
```

tests/unit/crypto/test_keystore.py
```diff
@@ -116,6 +116,20 @@ def test_keystore_invalid_password(tmpdir):
...
+    _keystore, words = Keystore.generate(
+        INSECURE_DEVELOPMENT_PASSWORD,
+        keystore_dir=tmpdir,
+        interactive=False,
+        report_mnemonic=True)
+    assert len(words.split(" ")) == 24
+
+
+def test_keystore_generate_report_mnemonic_blocked_by_interactive(tmpdir):
+    with pytest.raises(ValueError):
```
KPrasch, line 129: If you decide to keep this exception at all, consider making the test more strict:

```suggestion
    with pytest.raises(ValueError, match="The two values: report_mnemonic and interactive,  may not both be `True`"):
```

nucypher/crypto/keystore.py
```diff
@@ -342,15 +342,27 @@ def restore(cls, words: str, password: str, keystore_dir: Optional[Path] = None)
...
         """Generate a new nucypher keystore for use with characters"""
+        if report_mnemonic and interactive:
```
KPrasch, line 352: It seems like this same logic can be expressed as a simple if/else against a single boolean flag. Why use two separate inputs?

xela7: I wanted to honor the existing interactive flag -- but never return the mnemonic.words when interactive was turned on since that seemed unnecessary and possibly(?) a bit loose with security.   If you think there is a good use case for doing both interactive and returning mnemonic.words, please let me know and i'll get rid of this functionality.

KPrasch: I don't see a use case in doing both, however the same behavior you have here can be expressed without `report_mnemonic`. Instead If `interactive=False` just return `keystore, __words`.

nucypher/crypto/keystore.py
xela7: I've changed `report_mnemonic` to  `return_mnemonic`.

nucypher/crypto/keystore.py
xela7: ok i've cleaned this up.  you are quite right.  i was originally doing it i think in part because i feared many tests would need to be changed.  I have found one just now that seems to error because of my change.  Working on that now:

pytest -s tests/integration/config/test_configuration_persistence.py

xela7: ok.  i think that the failing test is now passing with my latest change to base.py on line 802.

`            self.__keystore, _ = Keystore.generate(password=password,
`

xela7: uh oh - i think that this will or should break other tests... working on more changes...

xela7: I fixed the remaining broken test.

nucypher/config/base.py
```diff
@@ -799,9 +799,15 @@ def write_keystore(self, password: str, key_material: Optional[bytes] = None, in
...
         else:
-            self.__keystore = Keystore.generate(password=password,
-                                                keystore_dir=self.keystore_dir,
-                                                interactive=interactive)
+            if interactive:
+                self.__keystore = Keystore.generate(password=password,
+                                                    keystore_dir=self.keystore_dir,
+                                                    interactive=interactive)
+            else:
+                self.__keystore, _ = Keystore.generate(password=password,
```
KPrasch, line 807: `_` is unused here, and `self.keystore` is returned in both cases.  Is this the intended behavior?

KPrasch: @derekpierre - The difference here is whether or not to unpack the result, assigning `self.__keystore` to a tuple or a keystore instance..

derekpierre: Ah so we had it flipped, they care about the words when interactive is enabled; so line 803.

👍

PR #2829, opened 2021-11-24
Fix some remaining occurrences of `n` instead of `shares`

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

**What this does:**
Fixes a bug reported on Discord, where an old name for the `shares` parameter is used.


derekpierre (approved):
🎸

KPrasch (approved):

PR #2831, opened 2021-12-07
Removes snapshots logic from `StakingEscrow`

vzotova:
**Type of PR:**
- Feature

**Required reviews:** 
- 1

**Why it's needed:**
To remove staker power in DAO

### DIFF
--- a/newsfragments/2831.feature.rst
+++ b/newsfragments/2831.feature.rst
@@ -1,1 +1,1 @@
-Removes snapshots logic from `StakingEscrow`
+Removes snapshots logic from ``StakingEscrow``
--- a/newsfragments/2831.feature.rst
+++ b/newsfragments/2831.feature.rst
@@ -1,0 +1,1 @@
+Removes snapshots logic from `StakingEscrow`
--- a/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
+++ b/nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
@@ -7,7 +7,6 @@ import "aragon/interfaces/IERC900History.sol";
 import "contracts/NuCypherToken.sol";
 import "contracts/lib/Bits.sol";
 import "contracts/lib/Snapshot.sol";
-import "contracts/lib/AdditionalMath.sol";
 import "contracts/proxy/Upgradeable.sol";
 import "zeppelin/math/Math.sol";
 import "zeppelin/token/ERC20/SafeERC20.sol";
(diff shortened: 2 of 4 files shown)


newsfragments/2831.feature.rst
```diff
@@ -0,0 +1 @@
+Removes snapshots logic from `StakingEscrow`
```
cygnusv, line 1:
```suggestion
Removes snapshots logic from ``StakingEscrow``
```
Needs double-quotes for RST format

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -399,11 +349,17 @@ contract StakingEscrow is Upgradeable, IERC900History {
     //------------------ ERC900 connectors ----------------------
 
     function totalStakedForAt(address _owner, uint256 _blockNumber) public view override returns (uint256) {
-        return stakerInfo[_owner].history.getValueAt(_blockNumber);
+        if (isUpgrade == UPGRADE_TRUE) {
+            return stakerInfo[_owner].history.getValueAt(_blockNumber);
+        }
+        return 0;
```
cygnusv, line 355: Great trick! Does this mean that in the next upgrade, we can just return 0 always?

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
```diff
@@ -399,11 +349,17 @@ contract StakingEscrow is Upgradeable, IERC900History {
...
     }
 
-    function totalStakedAt(uint256 _blockNumber) public view virtual override returns (uint256) {
-        return balanceHistory.getValueAt(_blockNumber);
+    function totalStakedAt(uint256 _blockNumber) public view override returns (uint256) {
+        if (isUpgrade == UPGRADE_TRUE) {
+            return balanceHistory.getValueAt(_blockNumber);
+        }
+        return 0;
```
cygnusv, line 362: For greater guarantees, what about continuing to return the `balanceHistory` here? That way voting power for individual addresses will be 0% for everyone. Otherwise, relative voting power (i.e., individual voting power / total voting power)will be 0/0, which I have no idea how Aragon would handle.
Alternatively, instead of 0, we could return here something like `type(uint256).max` or something like that.

newsfragments/2831.feature.rst
vzotova: ✔️

nucypher/blockchain/eth/sol/source/contracts/StakingEscrow.sol
vzotova: yep, see #2825

Issue #1762, opened 2020-03-05
blake2b-py | Use the hashlib included in pyhton 3.6 instead of own package | ARM


WayneWayner:
blake2b-py is already included in python 3.6 hashlib. 
So is there any need to install it as a separate package?

I came up with this because the nucypher setup fails at this point on an ARM CPU.

Maturin is used to install Blake2b-py and is the point where it fails.
https://github.com/PyO3/maturin
The current maturin sources are said to have ARM support since the last pull but I still can't get any further because of new errors.

The best from my point of view would be to leave out the package and use the hashlib.

tuxxy:
Hi, in my opinion (other NuCypherino opinions may vary), launching the network with ARM support is a very bad idea.

At the beginning of launch, it's incredibly important that we have reliable nodes to bootstrap the network. Since NuCypher hasn't been reliably tested on ARM machines (such as Raspberry Pis), I think it's a bad idea to support this from launch.

If ARM support is something we'd like to add, then it should be first integrated into our testing workflow and ideally have a few ARM-based testnet nodes running for a while without issue.

WayneWayner:
I agree that it would be disadvantageous for mainnet to release untested ARMs. But can you prevent this from happening if someone goes for it? 
I thought it would be nice to do the first tests with an ARM on the testnet now.

KPrasch:
Closed by #2837

PR #2837, opened 2021-12-26
Bye bye, blake2b-py!

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [] 3

**What this does:**
Some users (including myself) are experiencing troubles to install nucypher due to an error with the `blake2b-py` dependency for Apple M1. Funny thing, we don't use blake2b anymore, and it's only just a subdependency of `py-evm` used only for blake precompiles in the EVM. `py-evm` already marked this dependency an optional in https://github.com/ethereum/py-evm/pull/2012, so the fix here is just to pin `py-evm >= 0.5.0a0` to remove the dependency of `blake2b-py`. This PR also relocks all dependencies files. 

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


dev-requirements.txt
```diff
@@ -1,49 +1,50 @@
...
-gitdb==4.0.7; python_version >= '3.4'
-gitpython==3.1.18; python_version >= '3.6'
-greenlet==1.1.1; python_version >= '3' and platform_machine in 'x86_64 X86_64 aarch64 AARCH64 ppc64le PPC64LE amd64 AMD64 win32 WIN32'
-hypothesis==6.21.3
-identify==2.2.14; python_full_version >= '3.6.1'
-idna==3.2; python_version >= '3'
+filelock==3.4.0; python_version >= '3.6'
+gitdb==4.0.9; python_version >= '3.6'
+gitpython==3.1.24; python_version >= '3.7'
+greenlet==1.1.2; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'
```
piotr-roslaniec, line 15: I may be misunderstanding how environment markers work, so let me ask a clarifying question here: 
We set `[requires] python_version = "3"` in `Pipfile`, `[mypy] python_version=3.6` in `mypy.ini`, and  `PYPI_CLASSIFIERS` in `setup.py` to be essentially `3.9 => python_version => 3.6`. That leaves us with documented compatibility with Python 3.6-3.9. 
When relocking dependencies, why do we ask dependencies to enable compatibility with (for example) `python_version >= '2.7'` etc. instead of `python_version >= '3.6'?` Is it to maintain backward-compatibility with previous Python 3 versions, perhaps caused by setting `[requires] python_version = "3"` in `Pipfile`?

piotr-roslaniec (approved):
LGTM :tada: 
Just a single question about relocking dependencies.

dev-requirements.txt
derekpierre: Good question. This is automatically produced by `pipenv` when running the `pipenv lock ... --requirements >  [requirements_file]` command to relock dependencies. There may be ways to refine it, or perhaps this has improved in newer versions of pipenv but haven't looked into it recently.

Maybe https://github.com/pypa/pipenv/pull/4751 which was included in version 2021.11.5 (https://pipenv.pypa.io/en/latest/changelog/#id24) may help (?)

derekpierre (approved):
🎸

afmsavage:
🔥

KPrasch:
related to https://github.com/nucypher/nucypher/issues/1762

requirements.txt
```diff
@@ -1,24 +1,26 @@
...
+aiosignal==1.2.0; python_version >= '3.6'
 appdirs==1.4.4
+async-timeout==4.0.2; python_version >= '3.6'
 attrs==21.4.0; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'
-autobahn==22.2.2; python_version >= '3.7'
+autobahn==22.3.2; python_version >= '3.7'
 automat==20.2.0
 base58==2.1.1; python_version >= '3.5'
 bitarray==1.2.2
-blake2b-py==0.1.3
```
KPrasch, line 8: Here is the removal of blake2b from dependencies.

Pipfile
```diff
@@ -32,12 +32,12 @@ flask = "*"
 # 2.1.0 removed `JSONWebSignatureSerializer`, which `flask_httpasswd` depends on
 # (and it does not cap the version of `itsdangerous` itself)
 itsdangerous = "<2.1.0"
+jinja2 = "<3.1.0"
```
KPrasch, line 35: This is a workaround for https://github.com/sphinx-doc/sphinx/issues/10291

KPrasch (commented):
🤠

PR #2824, opened 2021-11-22
[EPIC] Updates for threshold network

vzotova:
**Type of PR:**
- Feature

**Required reviews:** 
- 3

**What this does:**
- [X] Halting NU inflation
- [X] Remove snapshots
- [x] Deployment of updated StakingEscrow for halting NU inflation and update contract registries - #2836 
- [x] Integrate `StakingEscrow` with `TokenStaking`
- [x] Deploy updated StakingEscrow contract (with tStaking functionality) on IBEX for testing - #2847 
- [x] `PREApplication`
- [x] Adjust codebase

PR #2840, opened 2022-01-11
[Do Not Merge] Payment Methods

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
3

**What this does:**

##### Payments 
- Introduces `ReencryptionPrerequsite`, `PaymentMethod`, and `ContractPayment` baseclasses. (+100 lines)
- Supports swapable `PaymentMethod`s by Ursula and Alice facilitate layer-2 policy payment. (+175 lines)
    - `PolicyManagerPayment`
    - `SubscriptionManagerPayment`
    - `FreeReencryptions` (Federated)

##### Contracts
- Temporarily hosts `SubscriptionManager.sol` and ABI (+300 lines)
- Introduces `SubscriptionManagerAgent` & `SubscriptionManagerDeployer`

##### Actor & Character
- Removal of `BlockchainPolicyAuthor`
- Federated nodes use `FreeReencryptions` by default

##### Configuration & CLI
- Introduces `CharacterConfiguration.configure_payment_method`
- New configuration fields `payment_method`, `payment_network`, and `payment_provider` for `AliceConfiguration` and `UrsulaConfiguration`
- New CLI flags `--payment-method`, `--payment-network`, and `--payment-provider` for `alice` and `ursula`
- Example and docs updates for those flags (+200 lines)

**Issues fixed/closed:**
TBD

**Why it's needed:**

The rent is too damn high.  

- Allows node operators to specify which payment method to check before providing service
- Support for the acceptance/submission of payments on polygon/matic by default

**Notes for reviewers:**
- Design only assumes operation with asset-bridged EVM-compatible L2s.
- Contains targeted compromises and code inefficiencies (see `TODO`s)
- Open to suggestions on a more intuitive naming scheme for "payment-*"
- Preserves compatibility with L1 payments on `PolicyManager`
- Future support of multiple payment methods
- ~~Consider complete removal of `BlockchainPolicyAuthor`~~ Removed
- This PR does not include a L2 reward collection flow for Ursula (needs follow-up)
- We need network names for polygon deployments
- `ReencryptionPrerequsite` can be used to enforce other "policies" like expiration
- This PR is subject to rebase over #2824 
- Opens #2844 (@derekpierre )


derekpierre:
@KPrasch Some design questions for you - perhaps these have already been discussed or I'm jumping the gun since this is WIP or just a base for follow-ups 😅 .

I saw that in this PR, Ursulas need to run with a specific `--payment_network` and `--payment_provider` specified. Are we envisioning having Ursula only support one payment network/provider? Would Ursulas specify that they are only accepting payments from some specific network - would that  therefore affect/limit their ability to be sampled for a policy depending on where the payment for the policy would be made? 

I assume our eventual goal here is network agnosticism for all Ursulas. If so, it starts to feel like all Ursulas need to support the various payment network/providers that are supported by the PRE Application eg. Polygon, other L2s in the future, etc. Ursulas wouldn't know which provider to use to check payment until presumably a re-encryption request comes in. So specifying it at Ursula start time would be premature. This then leads me to think that really, payment(s) network/provider information may be something much more involved allowing multiple network/provider values that provides various configuration information for all of the supported network/providers in one place, and then re-encryption requests specify which one to use to verify the legitmacy of the policy (Polygon, mainnet, other L2 etc.). eg. a section of the configuration may be:

```json
{
      "payments_config": {
               "payment_network_1": "payment_provider_1",
               "payment_network_2": "payment_provider_2", 
               "payment_network_3": "payment_provider_3",
      }
}
```

OR 

```
{
      "payments_config": {
               "payment_network_1":  {
                       "payment_provider_1": "",
                       "other_config_entry_1: "",
                       "other_config_entry_2: "",
                       ...
               },
               "payment_network_2": {
                        ...
               }
      }
}
```

If the above holds, this leads me to the next question which is - how does the indication of what network to use to verify the policy get provided to an Ursula, so that Ursula can then subsequently check that the policy is paid for. Would this have to be part of the re-encryption request? separate parameter or included in the treasure map? Should this value also be included in the derivation of the policy HRAC - prevent collisions for different network, and potential malfeasance?

Interested to get your thoughts (cc @cygnusv ). Let me know if I've totally misunderstood something/everything here.

KPrasch:
subsumed by #2857

PR #2848, opened 2022-01-18
Retire Faucet

KPrasch:
**Type of PR:**
Removal

**Required reviews:** 
1

**What this does:**
- Removes `Felix`
- Removes `nucypher felix` CLI
- Removes `chaotic.py`
- Removes all faucet functionality

**Why it's needed:**
Reduces the scope of the nucypher codebase.

**Notes for reviewers:**
We can push a git tag marking the version to use for future support.  Ultimately this faucet mini-application is best served in it's own repository.


tests/integration/config/test_character_configuration.py
```diff
@@ -48,13 +44,13 @@
...
-blockchain_only_characters = (Felix, StakeHolder)
+blockchain_only_configurations = (StakeHolderConfiguration, )
+blockchain_only_characters = (StakeHolder, )
 
 # Assemble
 characters_and_configurations = list(zip(characters, configurations))
-all_characters = tuple(characters + blockchain_only_characters)
-all_configurations = tuple(configurations + blockchain_only_configurations)
+all_characters = tuple(characters, )
+all_configurations = tuple(configurations, )
```
derekpierre, line 53: Are we no longer including `blockchain_only_characters`(i.e. only `StakeHolder` character) and the associated `blockchain_only_configurations`(i.e. only`StakeHolderConfiguration`)?

I know we are going to get rid of it eventually - but just checking if we wanted to keep it for now.

KPrasch: Since it's EOL I preemptively removed it here, albeit eagerly.  I think the remaining tuple was empty so it was awkward to leave it dangling.

derekpierre: 👍

PR #2835, opened 2021-12-10
rust-python execution environment for CI

KPrasch:
**Type of PR:**
Ops

**Required reviews:** 
1

**What this does:**
- Includes Dockerfile for building rust-python container for use on CI
- Encapsulate path management and permissions in circle config
- Use `python3 -m <module>` everywhere possible
- Reintroduce `[docs]` pip extra in `setup.py`

**Why it's needed:**
Accommodate unreleased development installation nucypher with rust dependencies (including bindings) for execution on CI or developer workflows.

PR #2849, opened 2022-01-22
Use nucypher/rust-python

KPrasch:
**Type of PR:**
Ops

**Required reviews:** 
1

**What this does:**
- Use nucypher-hosted rust/python docker images for CI and deployments.
- Fixes missed renaming of examples script

**Why it's needed:**
- Follow-up for #2835 #2809
- Examples CI compatibility with locally-built nucypher-core

**Notes for Reviewers**
- The two demo failures on CI are expected and will be fixed in a follow-up PR.  This one is only for pointing configs at the nucypher docker repo instead of my personal one.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -386,28 +386,28 @@ python_36_base: &python_36_base
   parallelism: 1
   working_directory: ~/nucypher
   docker:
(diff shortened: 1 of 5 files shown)


derekpierre (approved):
🎸

PR #2850, opened 2022-01-24
Address derivation inside nucypher-core

fjarri:
- use `NodeMetadataPayload.derive_operator_address()` instead of deriving with Python libraries
- `TransactingPower.sign_message()` returns the signatures with the normalized recovery byte (that is, with the chain ID cleared), since they need to be compatible with RustCrypto
- `coincurve` dependency removed (it was only used in one place for address derivation, and that is already covered by `eth_account`)
- `umbral` dependency removed (it is re-exported from `nucypher-core`)
- Dependencies relocked (`py-evm` pinned at '<0.5`, the update is to be addressed by #2837)


piotr-roslaniec (approved):
LGTM :tada:

nucypher/network/nodes.py
```diff
@@ -1096,11 +1078,11 @@ def validate_worker(self, registry: BaseContractRegistry = None) -> None:
...
         else:
 
-            # Off-chain signature verification
-            if not self._stamp_has_valid_signature_by_worker():
-                message = f"Invalid signature {self.__decentralized_identity_evidence.hex()} " \
-                          f"from worker {self.worker_address} for stamp {bytes(self.stamp).hex()} "
-                raise self.InvalidWorkerSignature(message)
+            # Try to derive the worker address if it hasn't been derived yet.
+            try:
+                worker_address = self.worker_address
```
KPrasch, line 1083: I know it's a bit of extra lift however, this line will read much more explicitly and better fit the emerging conventions as a function.

nucypher/network/nodes.py
```diff
@@ -1221,8 +1203,5 @@ def decentralized_identity_evidence(self):
     @property
     def worker_address(self):
         if not self.__worker_address and not self.federated_only:
-            if self.decentralized_identity_evidence is NOT_SIGNED:
-                raise self.StampNotSigned  # TODO: Find a better exception  NRN
-            self.__worker_address = recover_address_eip_191(message=bytes(self.stamp),
-                                                            signature=self.decentralized_identity_evidence)
+            self.__worker_address = to_checksum_address(self.metadata().payload.derive_worker_address())
```
KPrasch, line 1206: This can be called for both local and remote nodes?

fjarri: Yes, they both have the `metadata()` method; for remote ones is what was received from the teacher, for the local one is what it built itself.

KPrasch: So is it possible that the derived address conflicts with an address provided by the worker (perhaps my accidental misconfiguration)?  If so how is this handled?

fjarri: Hm, what do you mean? The worker address is not currently a part of metadata.

KPrasch: Still thinking through it a bit... I'm wondering if there's a concern of mismatched addresses if the derived address does not match the value of the worker address obtained from user input/config, through `Ursula.__init__`, etc.  Ideally such behavior is not even possible.

fjarri: They way I understand the logic, we initialize an Ethereum account somehow (e.g. from a keystore), and use it to create the evidence. So in order for the derived address not to match, the evidence would have to be derived incorrectly.

Note also that the logic in these lines didn't change - the address was derived from the evidence before as well, it's just that now the evidence is taken from the metadata which takes it from the attribute, instead of directly taking it from the attribute.

KPrasch (approved):

fjarri:
Tried to move things around with the derivation of operator address/decentralized evidence, tests pass, but what's really needed is a proper separation of local and remote Ursulas.

nucypher/characters/lawful.py
```diff
@@ -702,7 +703,7 @@ def __init__(self,
                  checksum_address: ChecksumAddress = None,
                  operator_address: ChecksumAddress = None,  # TODO: deprecate, and rename to "checksum_address"
                  client_password: str = None,
-                 decentralized_identity_evidence=NOT_SIGNED,
+                 decentralized_identity_evidence_from_metadata=NOT_SIGNED,
```
KPrasch, line 706: Why so wordy?

nucypher/characters/lawful.py
```diff
@@ -1003,16 +1026,20 @@ def get_deployer(self):
...
         # so we can cache the result of this method.
         # TODO: should this be a method of Teacher?
         timestamp = maya.now()
-        if self.decentralized_identity_evidence is NOT_SIGNED:
+        if self.federated_only:
             decentralized_identity_evidence = None
         else:
             decentralized_identity_evidence = self.decentralized_identity_evidence
-        payload = NodeMetadataPayload(canonical_address=self.canonical_address,
+        payload = NodeMetadataPayload(staker_address=self.canonical_address,
```
KPrasch, line 1042: To follow the convention this can be `staking_provider_address` everywhere in this PR.

tests/acceptance/characters/test_ursula_prepares_to_act_as_worker.py
```diff
@@ -44,13 +44,11 @@ def test_blockchain_ursula_substantiates_stamp(blockchain_ursulas):
     first_ursula = list(blockchain_ursulas)[0]
     signature_as_bytes = first_ursula.decentralized_identity_evidence
     signature_as_bytes = to_standard_signature_bytes(signature_as_bytes)
+    # `worker_address` was derived in nucypher_core, check it independently
```
KPrasch, line 47: `worker_address` -> `operator_address`

nucypher/characters/lawful.py
```diff
@@ -853,6 +853,29 @@ def __substantiate_stamp(self):
...
         self.log.debug(message)
 
+    @property
+    def decentralized_identity_evidence(self):
+        return self.__decentralized_identity_evidence
+
+    @property
+    def operator_address(self):
+        if not self.federated_only:
+            # TODO: The reason for the fork here is the difference in available information
```
KPrasch, line 863: Definitely an awkward branching and getter logic here.  Deserves an issue to track.

nucypher/policy/revocation.py
```diff
@@ -29,7 +29,7 @@ def __init__(self, treasure_map, signer: SignatureStamp):
         self.revocations = dict()
         for ursula_address, encrypted_kfrag in treasure_map.destinations.items():
             self.revocations[ursula_address] = RevocationOrder(signer=signer.as_umbral_signer(),
-                                                               ursula_address=ursula_address,
+                                                               staker_address=ursula_address,
```
KPrasch, line 32: Was this an intentional change?

nucypher/characters/lawful.py
fjarri: Couldn't think of anything better, to be honest, and decided that explicit is at least clear. Suggestions?

nucypher/characters/lawful.py
fjarri: Ok, should I update the API & bump version on `nucypher-core` then?

nucypher/policy/revocation.py
fjarri: Yes, it's the same `staker_address` as the NodeMetadata has, so I thought it should be named the same too.

KPrasch: Hmm - I don't understand.  This change implies the changing of roles from the operater's address to the staking providers address.  At minimum this is a confusing line since `staker` and `ursula` are typically handled as separate actors/addresses.  Are you sure we don't have our wires crossed here?

`staker_address=ursula_address` - huh?

KPrasch: Ultimately the revocation API needs a little reinvention so this is not a very important change at this time.

fjarri: The `ursula_address` here is taken from `treasure_map.destinations`, so it is a staker address, not an operator address. That's why I renamed it - makes it more clear.

fjarri: I guess it's worth renaming the loop variable too.

KPrasch: Something is certainly incorrect here if we end up with `staking_provider_address=operator_address`, no?

fjarri: There's no operator addresses anywhere here, it's all staking addresses. I guess it the variable just got unfortunately named originally.

KPrasch: Ah - Got it! I agree you may as well fixup the loop var.  👍🏻

nucypher/characters/lawful.py
KPrasch: Heh, I suppose this is the meaning of integration of incoming changes with the rust-core!

Since the convention is so widespread I think it's worthwhile.

nucypher/characters/lawful.py
KPrasch: Kind of a tough one.  How about simply  `identity_evidence` or `wallet_signature`?

fjarri: Ah, I thought you're referring to the `_from_metadata` prefix. If you want to rename the whole thing, it'll have to be renamed in `nucypher-core` as well. `identity_evidence` is fine, or perhaps `operator_signature`?

KPrasch: Either of those seem fine to me 👍🏻

fjarri: I did not want to make too many changes in this PR, only focusing on bringing it up to speed with `nucypher-core`, but I guess I can rename that everywhere. The `nucypher-core` renaming will have to wait the next release though.

fjarri: Renamed to `operator_signature`, filed https://github.com/nucypher/nucypher-core/issues/6

tests/acceptance/characters/test_ursula_prepares_to_act_as_worker.py
fjarri: Fixed

nucypher/characters/lawful.py
fjarri: Filed https://github.com/nucypher/nucypher/issues/2875

nucypher/policy/revocation.py
fjarri: Fixed the loop variable, `nucypher-core` part to be fixed with https://github.com/nucypher/nucypher-core/issues/6

nucypher/characters/lawful.py
fjarri: Filed <https://github.com/nucypher/nucypher-core/issues/6>

KPrasch (approved):

PR #2834, opened 2021-12-10
[WIP] PRE application contract

vzotova:
**Type of PR:**
- [X] Feature

**Required reviews:** 
- [X] 3

**What this does:**
> High-level idea of the changes introduced in this PR. 
> List relevant API changes (if any), as well as related PRs and issues.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #2856, opened 2022-02-01
SimplePREApplication: confirmation can be checked using staking provider or operator

vzotova:
**Type of PR:**
- Feature

**Required reviews:** 
- 2

**What this does:**
Changes app contract ABI: `isOperatorConfirmed(address _operator)` -> `isOperatorConfirmed(address _stakingProviderOrOperator)`

**Why it's needed:**
Makes this check more convenient for both UI and PRE node

**Notes for reviewers:**
There is no guarantee 1to1 relation between `operator`<->`stakingProvider` in this version of contract. Implementation tries to prevent false negative in case when some `operator` is another `stakingProvider`
Still possible to have false positive for this case.


vzotova:
I'm closing for now because of this ambiguous case with false positive result. Probably will aplly those changes in #2834

PR #2877, opened 2022-02-25
Fix obtaining metadata from a node

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

Not sure how it worked before, but there hasn't been `bytes(node)` for a while. `metadata()` needs to be called explicitly.


KPrasch (approved):
🤠

PR #2885, opened 2022-03-21
Fix bug in git url of docs dependancy `jsonschema2rst`

theref:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Fixes bug in the git url of dependancy `jsonschema2rst`.
The cause of the issue is described [here](https://teknotopnews.com/otomotif-https-github.blog/2021-09-01-improving-git-protocol-security-github/) and [here](https://teknotopnews.com/a/70663683).
This bug was expected from Jan. 11th, 2022.

**Why it's needed:**
Docs cannot be built by circleci because it cannot install the dependancies, as seen here https://app.circleci.com/pipelines/github/nucypher/nucypher/9307/workflows/e680a4ca-71cb-4cbf-9106-29e50e9e2b10/jobs/157929

### DIFF
--- a/docs-requirements.txt
+++ b/docs-requirements.txt
@@ -11,7 +11,7 @@ coverage==6.3.2; python_version >= '3.7'
 docutils==0.17.1; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'
 dulwich==0.19.16
 execnet==1.9.0; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3, 3.4'
-git+git://github.com/inspirehep/jsonschema2rst.git@d211d9709046431a6b6a4594c97c22d8713d854b#egg=jsonschema2rst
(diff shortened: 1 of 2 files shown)


vzotova (approved):

scripts/dependencies/docs/Pipfile.lock
```diff
@@ -0,0 +1,507 @@
+{
```
derekpierre, line 1: We don't maintain a separate `Pipfile.lock` for the docs; so this file is not needed. It is purposely removed via this relocking of dependencies script which handles the repos dependencies - https://github.com/nucypher/nucypher/blob/main/scripts/dependencies/relock_dependencies.sh#L39.

The generated `docs-requirements.txt` file is sufficient for docs dependencies, and is used by ReadTheDocs when generating the documentation (cc @KPrasch).

derekpierre (approved):
🎸

KPrasch (approved):
🤠

KPrasch:
related to #2902

PR #2873, opened 2022-02-23
Grant & Retrieve via Polygon Payments; Pre-release cleanup.

KPrasch:
**Type of PR:**
Bugfix & Feature

**Required reviews:** 
2

**What this does:**

- Polygon CLI and configuration updates
- Mitigations for network update metadata pollution
- Handles SSL certificates as part of RestMiddleware with just-in-time cert fetching
- Forcibly clear all stored node metadata and certificates on startup
- Only store peers in memory during runtime
- Working demos for polygon testnet
- Relax invalid node logging and errors
- Bumps probationary period three months
- Minor code quality improvements (type hints and organization)
- `NUCYPHER_WORKER_ETH_PASSWORD` -> `NUCYPHER_OPERATOR_ETH_PASSWORD`

PR #2886, opened 2022-03-24
Fix additional issues from Airship

KPrasch:
**Type of PR:**
- [X] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Builds on the `airship` PR ( #2873 ), and fixes / addresses some issues observed during testing.

* Fixes large number of web3 requests made due to the continued running of WorkTracker - the repurposed task used to make sure the operator confirmed transaction is performed on startup. On the first run of Ursula, the WorkTracker is started and tries to confirm the operator, but once it does, the node continues running but the task never shuts down i.e. once the operator is confirmed the WorkTracker isn't proactively stopped. If Ursula is subsequently restarted then all is fine because the operator was already confirmed and the WorkTracker is not needed and is never started.

Screen_Shot_2022-03-24_at_9 06 26_AM

* Whenever an Ursula's operator address is no longer bonded to a staking provider, it should not keep running i.e. it should be proactively shutdown.

**Why it's needed:**
* Free infura accounts have daily limits, which will be exceeded if the WorkTracker isn't stopped when no longer needed
* An unbonded Ursula running in the network is useless.

### DIFF
--- a/nucypher/network/trackers.py
+++ b/nucypher/network/trackers.py
@@ -16,16 +16,58 @@
 """
 
 import random
+from typing import Union
 
 import maya
 from twisted.internet import reactor
 from twisted.internet.task import LoopingCall
-from typing import Union
+from twisted.python.failure import Failure
 
+from nucypher.blockchain.eth.agents import ContractAgency, PREApplicationAgent
+from nucypher.blockchain.eth.constants import NULL_ADDRESS
+from nucypher.control.emitters import StdoutEmitter
 from nucypher.network.exceptions import NodeSeemsToBeDown
 from nucypher.network.middleware import RestMiddleware
 from nucypher.network.nodes import NodeSprout
 from nucypher.utilities.logging import Logger
+from nucypher.utilities.task import SimpleTask
+
+
+class OperatorBondedTracker(SimpleTask):
+    INTERVAL = 60 * 60  # 1 hour
+
+    class OperatorNoLongerBonded(RuntimeError):
+        """Raised when a running node is no longer associated with a staking provider."""
+
+    def __init__(self, ursula):
+        self._ursula = ursula
+        super().__init__()
+
+    def run(self) -> None:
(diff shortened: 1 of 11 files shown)


nucypher/network/trackers.py
```diff
@@ -16,16 +16,54 @@
...
 from nucypher.network.nodes import NodeSprout
 from nucypher.utilities.logging import Logger
+from nucypher.utilities.task import RestartableTask
+
+
+class OperatorBondedTracker(RestartableTask):
+    INTERVAL = 60 * 60  # 1 hour
+
+    class OperatorNoLongerBonded(RuntimeError):
+        pass
```
KPrasch, line 38:
```suggestion
"""Raised when a running node is no longer associated with a staking provider"""
```

derekpierre: ✔️

derekpierre:
My IBEX testnet node is currently running this code.

nucypher/utilities/task.py
```diff
@@ -0,0 +1,64 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def handle_errors(self, *args, **kwargs):
+        """Error callback for error handling during execution."""
+        raise NotImplementedError
+
+    @staticmethod
+    def clean_traceback(failure: Failure) -> str:
+        # FIXME: Amazing.
```
cygnusv, line 62: What's this?

cygnusv (approved):

nucypher/utilities/task.py
derekpierre: It's code from before, just copied here. Has to do with the ability to log tracebacks and issues with curly braces

KPrasch (commented):
🤠

PR #2908, opened 2022-04-06
Optimizations for SSL, Porter etc.

### LINKED ISSUE #2907: Optimize just-in-time certificate fetching as part of REST middleware
**Describe the Bug**
General slowness for p2p communication. PR #2873 modified SSL certificate handling by always first fetching the node's self-signed certificate and then using that to hit other endpoints on the node, instead of possibly using a stale certificate that was previously obtained.

This change resulted in 2 roundtrips for every node endpoint that is hit. Slows down discovery loop, and Porter's ability to test whether a node is online or not when sampling.

We should consider optimizations to reduce the slowdown being observed in v6.0.0.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Middleware attempts to use cached copy of certificate first for the request, and if the certificate fails verification during the request, only then is a fresh copy of the SSL cert fetched and then the request retried.

**Issues fixed/closed:**
Fixes #2907 .

**Why it's needed:**
Slowdown of inter-node communication, and Porter functionality.

**Notes for reviewers:**


derekpierre:
@vepkenez / @KPrasch  this commit, https://github.com/nucypher/nucypher/pull/2908/commits/3827b69ad7c7c3b4c7f183cdb7b1607178f8eec0, should resolve the issue of Porter acting wonky when trying to sample for 1 node.

vepkenez:
> @vepkenez / @KPrasch this commit, [3827b69](https://github.com/nucypher/nucypher/commit/3827b69ad7c7c3b4c7f183cdb7b1607178f8eec0), should resolve the issue of Porter acting wonky when trying to sample for 1 node.

@derekpierre why does it do that?
Because of the removal of the `stagger timeout` arg?

derekpierre:
> Because of the removal of the stagger timeout arg?

@vepkenez the stagger timeout remains. It is the removal of the thread pool size limitation. When the quantity to sample was 1, the thread pool would max out at 1 thread - and therefore it becomes a sequential type execution and trying any node that would fail, slows down the whole process, which then leads to execution timeouts.

nucypher/control/controllers.py
```diff
@@ -302,8 +304,23 @@ def start(self,
...
+    def json_response_from_worker_pool_exception(exception):
+        json_response = {}
+        json_response['failure_message'] = f"{exception}; " \
+                                           f"{len(exception.failures)} concurrent failures recorded"
+        if exception.failures:
+            failures = []
+            for value, exc_info in exception.failures.items():
+                failure = {'value': value}
+                _, exception, tb = exc_info
+                failure['error'] = str(exception)
```
KPrasch, line 317: I think this can be done in fewer lines, and without the dict mutation?

nucypher/control/emitters.py
```diff
@@ -243,6 +243,14 @@ def __init__(self,
 
         self.log = Logger('web-emitter')
 
+    def _log_exception(self, e, error_message, log_level, response_code):
+        exception = f"{type(e).__name__}: {str(e)}" if str(e) else type(e).__name__
+        message = f"{self} [{str(response_code)} - {error_message}] | ERROR: {exception}"
+        logger = getattr(self.log, log_level)
+        # See #724 / 2156
+        message_cleaned_for_logger = message.replace("{", "<^<").replace("}", ">^>")
```
KPrasch, line 251: I know your just keeping up with status-quo here, but I think using a utility function in logging.py will make this more readable and less noisy.

nucypher/control/emitters.py
```diff
@@ -255,25 +263,35 @@ def exception(self,
...
         return self.sink(response_message, status=response_code)
 
-    def respond(self, response) -> Response:
-        assembled_response = self.assemble_response(response=response)
+    def exception_with_response(self,
+                                json_error_response,
+                                e,
+                                error_message: str,
+                                log_level: str = 'info',
+                                response_code: int = 500):
```
KPrasch, line 278: Just to recap -- under what conditions do you expect a 500 to be raised?  I think it's important we honor the 500 series errors a true "server exceptions".

nucypher/control/controllers.py

nucypher/control/emitters.py
derekpierre: ✔️ - Turns out there was one already.

nucypher/control/emitters.py
derekpierre: ✔️  - Generally, for the WebController any non-specific exceptions results in the response code of 500. In the future I think we should allow subclasses of WebController to specific an exception, http_code mapping so that it is more customizable - but that's later on.

For the only use of this method  the response code is specified (`404`) so 500 doesn't get used here. In any case, I removed the default value of `response_code` and it must now be explicitly specified when calling this function.

nucypher/control/controllers.py
```diff
@@ -302,8 +304,22 @@ def start(self,
     def __call__(self, *args, **kwargs):
         return self.handle_request(*args, **kwargs)
 
-    def handle_request(self, method_name, control_request, *args, **kwargs) -> Response:
+    @staticmethod
+    def json_response_from_worker_pool_exception(exception):
+        json_response = {}
+        json_response['failure_message'] = f"{exception}"
```
KPrasch, line 310: Why mutate here? IMO, initializing the dictionary with the first value is simple and direct.

nucypher/control/emitters.py
```diff
@@ -243,6 +243,13 @@ def __init__(self,
 
         self.log = Logger('web-emitter')
 
+    def _log_exception(self, e, error_message, log_level, response_code):
+        exception = f"{type(e).__name__}: {str(e)}" if str(e) else type(e).__name__
+        message = f"{self} [{str(response_code)} - {error_message}] | ERROR: {exception}"
+        logger = getattr(self.log, log_level)
+        message_cleaned_for_logger = Logger.escape_format_string(message)
```
KPrasch, line 250: 👍🏻 much better

nucypher/network/middleware.py
```diff
@@ -183,6 +182,41 @@ def method_wrapper(path,
...
+    def _execute_method(self,
+                        node_or_sprout,
+                        host,
+                        port,
+                        method,
+                        endpoint,
+                        *args, **kwargs):
+        # Use existing cached SSL certificate or fetch fresh copy and retry
+        cached_certificate_filepath = Path(
+            self.storage.generate_certificate_filepath(host=host, port=port))
```
KPrasch, line 194: A variable and two lines will be more readable here.

nucypher/network/middleware.py
```diff
@@ -183,6 +182,41 @@ def method_wrapper(path,
...
+        if cached_certificate_filepath.exists():
```
KPrasch, line 195: LBYL -- Can this be done without checking for the cached cert first?  It will be nice to unnest as many blocks as possible here.

nucypher/utilities/porter/porter.py
```diff
@@ -137,18 +139,22 @@ def get_ursula_info(ursula_address) -> Porter.UrsulaInfo:
...
-                                 stagger_timeout=1,
-                                 threadpool_size=quantity)
+                                 timeout=self.execution_timeout,
+                                 stagger_timeout=1)
         worker_pool.start()
-        successes = worker_pool.block_until_target_successes()
+        try:
+            successes = worker_pool.block_until_target_successes()
+        finally:
+            worker_pool.cancel()
```
KPrasch, line 155: Nice use of a try/finally here.

nucypher/network/middleware.py
```diff
@@ -183,6 +182,41 @@ def method_wrapper(path,
...
+                # successful use of cached certificate
+                return response
+            except SSLError as e:
+                # ignore this exception - probably means that our cached cert may not be up-to-date.
+                SSL_LOGGER.debug(f"Cached cert for {host}:{port} is invalid {e}")
+
+        # Fetch fresh copy of SSL certificate
+        try:
+            certificate, filepath = self.get_certificate(host=host, port=port)
+        except NodeSeemsToBeDown as e:
```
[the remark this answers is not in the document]
derekpierre: We need to chat more broadly about this exception. The introduction of RestMiddleware.Unreachable and potential propagation of it [here](https://github.com/nucypher/nucypher/blob/development/nucypher/network/middleware.py#L155) has had some unintended consequences.

Renaming it, won't be a problem - but let's do that in a separate PR.

nucypher/control/controllers.py
derekpierre: ✔️

nucypher/network/middleware.py
derekpierre: The concern I have here is that passing a non-existent filepath will cause an `IOError` to be raised (https://github.com/psf/requests/blob/main/requests/adapters.py#L228). However, `IOError` is very general, and it won't be clear if the `IOError` is specifically because of the non-existent cert, or because of some other reason related to the actual execution of the method (if the cert was actually valid). I'm also hesitant to filter on text in the exception message...

nucypher/network/middleware.py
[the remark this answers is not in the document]
derekpierre: ^ Filed #2920

KPrasch (approved):
I've been using this in staging over the past week, and it's noticeably faster.  The granular per-node error reporting it a huge improvement as well.

PR #2917, opened 2022-04-13
Improve handling of network selection in ursula init

### LINKED ISSUE #2899: Only show valid payment provider networks during `ursula init`
This issues can be closed when:
- interactive ursula initialization via CLI only presents the user with valid options
- `nucypher ursula init` does not default `payment-provider` in `ursula.json` to the `--eth-provider-uri` value

The only thing the REPL asked, re: payment provider, was which network

Screenshot from 2022-04-02 10-58-30

theref:
**Type of PR:**
- [x] Bugfix

**Required reviews:** 
- [X] 2

**What this does:**
- [x] interactive ursula initialization via CLI only presents the user with valid options
- [x]  nucypher ursula init does not default payment-provider in ursula.json to the --eth-provider-uri value

**Issues fixed/closed:**
Fixes #2899

### DIFF
--- a/nucypher/blockchain/eth/networks.py
+++ b/nucypher/blockchain/eth/networks.py
@@ -22,6 +22,7 @@ class NetworksInventory:  # TODO: See #1564
     IBEX = 'ibex'
     LYNX = 'lynx'
     ORYX = 'oryx'
+    ETH = 'ethereum'
 
     # TODO: Use naming scheme to preserve multiple compatibility with multiple deployments to a single network?
     POLYGON = 'polygon'
@@ -30,18 +31,22 @@ class NetworksInventory:  # TODO: See #1564
     UNKNOWN = 'unknown'  # TODO: Is there a better way to signal an unknown network?
     DEFAULT = MAINNET
 
-    __to_ethereum_chain_id = {
+    __to_chain_id_eth = {
         MAINNET: 1,      # Ethereum Mainnet
         ORYX: 3,         # Ropsten
         IBEX: 4,         # Rinkeby
         LYNX: 5,         # Goerli
-
+    }
+    __to_chain_id_polygon = {
         # TODO: Use naming scheme?
         POLYGON: 137,    # Polygon Mainnet
         MUMBAI: 80001,   # Polygon Testnet (Mumbai)
     }
 
-    NETWORKS = tuple(__to_ethereum_chain_id.keys())
+    ETH_NETWORKS = tuple(__to_chain_id_eth.keys())
+    POLY_NETWORKS = tuple(__to_chain_id_polygon.keys())
+
+    NETWORKS = ETH_NETWORKS + POLY_NETWORKS
 
     class UnrecognizedNetwork(RuntimeError):
         pass
--- a/nucypher/cli/actions/select.py
+++ b/nucypher/cli/actions/select.py
@@ -136,13 +136,23 @@ def select_client_account(emitter,
     return chosen_account
 
 
-def select_network(emitter: StdoutEmitter, message: Optional[str] = None) -> str:
+def select_network(emitter: StdoutEmitter, network_type: str, message: Optional[str] = None) -> str:
     """Interactively select a network from nucypher networks inventory list"""
-    emitter.message(message=message or str(), color='yellow')
-    rows = [[n] for n in NetworksInventory.NETWORKS]
-    emitter.echo(tabulate(rows, showindex='always'))
-    choice = click.prompt(SELECT_NETWORK, default=0, type=click.IntRange(0, len(NetworksInventory.NETWORKS)-1))
-    network = NetworksInventory.NETWORKS[choice]
+    emitter.message(message=message or str(), color="yellow")
+    if network_type == NetworksInventory.ETH:
+        network_list = NetworksInventory.ETH_NETWORKS
+    elif network_type == NetworksInventory.POLYGON:
+        network_list = NetworksInventory.POLY_NETWORKS
+    else:
+        raise(ValueError("Network type must be either 'eth' or 'polygon'"))
+    rows = [[n] for n in network_list]
+    emitter.echo(tabulate(rows, showindex="always"))
+    choice = click.prompt(
+        SELECT_NETWORK,
+        default=0,
+        type=click.IntRange(0, len(rows) - 1),
+    )
+    network = network_list[choice]
     return network
 
 
--- a/nucypher/config/base.py
+++ b/nucypher/config/base.py
@@ -487,6 +487,7 @@ class CharacterConfiguration(BaseConfiguration):
             blockchain_args = {'filepath': registry_filepath,
                                'poa': poa,
                                'eth_provider_uri': eth_provider_uri,
+                               'payment_provider': payment_provider,
                                'gas_strategy': gas_strategy,
                                'max_gas_price': max_gas_price}
             if any(blockchain_args.values()):
--- a/newsfragments/2917.bugfix.rst
+++ b/newsfragments/2917.bugfix.rst
@@ -1,0 +1,1 @@
+Only print relevant network options when running `nucypher ursula init`
--- a/nucypher/cli/commands/bond.py
+++ b/nucypher/cli/commands/bond.py
@@ -23,6 +23,7 @@ from eth_typing import ChecksumAddress
 
 from nucypher.blockchain.eth.agents import ContractAgency, PREApplicationAgent
 from nucypher.blockchain.eth.constants import NULL_ADDRESS
+from nucypher.blockchain.eth.networks import NetworksInventory
 from nucypher.blockchain.eth.signers import Signer
 from nucypher.cli.actions.auth import get_client_password
 from nucypher.cli.actions.select import select_network
@@ -103,7 +104,7 @@ def bond(registry_filepath, eth_provider_uri, signer_uri, operator_address, stak
         emitter.message('--signer is required', color='red')
         raise click.Abort()
     if not network:
-        network = select_network(emitter=emitter)
+        network = select_network(emitter=emitter, network_type=NetworksInventory.ETH)
 
     signer = Signer.from_signer_uri(signer_uri)
     transacting_power = TransactingPower(account=staking_provider, signer=signer)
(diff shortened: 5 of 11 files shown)


nucypher/cli/actions/select.py
```diff
@@ -93,64 +108,84 @@ def select_client_account(emitter,
...
-        emitter.echo(NO_ETH_ACCOUNTS, color='red', bold=True)
+        emitter.echo(NO_ETH_ACCOUNTS, color="red", bold=True)
         raise click.Abort()
     elif len(enumerated_accounts) == 1:
         # There are no choices if there is only one available address.
         return enumerated_accounts[0]
 
     # Display account info
-    headers = ['Account']
+    headers = ["Account"]
```
derekpierre, line 118: Is it a linter producing these extra quotation changes...might want to disable some of these - makes the PR bigger than necessary...

nucypher/cli/actions/select.py
```diff
@@ -93,64 +108,84 @@ def select_client_account(emitter,
...
+def select_network(emitter: StdoutEmitter, message: Optional[str] = None, eth: Optional[bool] = True, polygon: Optional[bool] = True) -> str:
     """Interactively select a network from nucypher networks inventory list"""
-    emitter.message(message=message or str(), color='yellow')
-    rows = [[n] for n in NetworksInventory.NETWORKS]
-    emitter.echo(tabulate(rows, showindex='always'))
-    choice = click.prompt(SELECT_NETWORK, default=0, type=click.IntRange(0, len(NetworksInventory.NETWORKS)-1))
-    network = NetworksInventory.NETWORKS[choice]
+    emitter.message(message=message or str(), color="yellow")
+    if eth and polygon:
+        network_list = NetworksInventory.NETWORKS
```
derekpierre, line 165: Is both eth and polygon being true possible? What's the use case there? 

If showing both lists, does it make sense to keep them as two separate lists, and show them as two separate lists...combining them into one list looks like what in the terminal? Might be confusing to have a mixed list to then select from.

nucypher/blockchain/eth/networks.py
```diff
@@ -24,24 +24,28 @@ class NetworksInventory:  # TODO: See #1564
     ORYX = 'oryx'
 
     # TODO: Use naming scheme to preserve multiple compatibility with multiple deployments to a single network?
-    POLYGON = 'polygon'
-    MUMBAI = 'mumbai'
+    POLYGON = 'polygon-mainnet'
+    MUMBAI = 'polygon-mumbai'
```
derekpierre, line 28: I like the names, but we need to be careful here - changing these values has side effects. The payment network value is stored in Ursula configuration files, which would already exist for someone running v6.0.0 (i.e. that configuration file would have `polygon` as the value). 

If we want to change these values, I think you'll need to incremenent the default configuration version and create a configuration migration script for converting the old version of the configuration file to this version of the configuration file that uses these values - for example, https://github.com/nucypher/nucypher/commit/82fb99022f8d43b6dbb8c0b7e888bb5049d6d194. (cc @KPrasch ).

derekpierre (changes requested):

nucypher/blockchain/eth/networks.py
derekpierre: You can also accept both values I guess as a workaround

nucypher/cli/actions/select.py
theref: you're right, having searched through the code, there's no instance where you'd want the full list

nucypher/blockchain/eth/networks.py
theref: ok cool, i think i'll remove this change and treat that as a separate issue/PR #2918

derekpierre: 👌  - nice!

nucypher/blockchain/eth/networks.py
```diff
@@ -30,18 +30,22 @@ class NetworksInventory:  # TODO: See #1564
...
         # TODO: Use naming scheme?
         POLYGON: 137,    # Polygon Mainnet
         MUMBAI: 80001,   # Polygon Testnet (Mumbai)
     }
 
-    NETWORKS = tuple(__to_ethereum_chain_id.keys())
+    ETH_NETWORKS = tuple(__to_ethereum_chain_id_eth.keys())
+    POLY_NETWORKS = tuple(__to_ethereum_chain_id_poly.keys())
+
+    NETWORKS = ETH_NETWORKS + POLY_NETWORKS
```
derekpierre, line 48: Is this used anymore anywhere? If it still is, I would look into whether its usage can simply just be changed to ETH_NETWOKRS...

nucypher/cli/actions/select.py
```diff
@@ -136,13 +136,23 @@ def select_client_account(emitter,
     return chosen_account
 
 
-def select_network(emitter: StdoutEmitter, message: Optional[str] = None) -> str:
+def select_network(emitter: StdoutEmitter, network_type: str, message: Optional[str] = None) -> str:
```
derekpierre, line 139: 👌 Nice reuse of the function for both networks.

Should we make the network type values ( "eth" / "polygon" ) constants? Callers of this function can use the contents as well.

derekpierre (commented):
Good call on the separate issue for configuration changes.

nucypher/blockchain/eth/networks.py
theref: it's used for verification in a few places, for example in `nucypher/blockchain/eth/registry.py`

```python
class CanonicalRegistrySource(ABC):

    logger = Logger('RegistrySource')

    name = NotImplementedError
    is_primary = NotImplementedError

    def __init__(self, network: str, registry_name: str, *args, **kwargs):
        if network not in NetworksInventory.NETWORKS:
            raise ValueError(f"{self.__class__.__name__} not available for network '{network}'. "
                             f"Valid options are: {list(NetworksInventory.NETWORKS)}")
        self.network = network
        self.registry_name = registry_name
```

nucypher/cli/actions/select.py
theref: I'm happy to make that change, but not sure what the benefits are? it's just going to be:
```
ETH = "eth"
```
is that more readable?

nucypher/blockchain/eth/networks.py
```diff
@@ -30,18 +30,22 @@ class NetworksInventory:  # TODO: See #1564
     UNKNOWN = 'unknown'  # TODO: Is there a better way to signal an unknown network?
     DEFAULT = MAINNET
 
-    __to_ethereum_chain_id = {
+    __to_ethereum_chain_id_eth = {
```
KPrasch, line 33: I don't think there are any inter-chain ID collisions, why use two dictionaries?

derekpierre: The eventual split tuple of `ETH_NETWORKS` and `POLY_NETWORKS` allows for easier validation, and selection listing in the CLI - https://github.com/theref/nucypher/blob/ursula-init/nucypher/cli/actions/select.py#L139.

Did you have an alternative thought here?

nucypher/cli/actions/select.py
derekpierre: My original thought here was not just for the function itself, but that any callers of the function would then do:
```python
network = select_network(emitter=emitter, network_type=ETH)
```
OR
```python
network = select_network(emitter=emitter, network_type=POLYGON)
```

instead of having "eth" or "polygon" be hardcoded everywhere this function is called (~ 5 spots right now). Also a thought towards easily supporting other payment networks in the future. Presumably, the constants can live in `NetworksInventory`...?

wdyt?

theref: 👍 if more networks end up being added this will be a cleaner solution

fjarri: > should really be "eth" and "payment"

In nuycpher-async I am calling them "identity" and "payment"

nucypher/config/base.py
```diff
@@ -398,6 +398,7 @@ def __init__(self,
                  poa: bool = None,
                  light: bool = False,
                  eth_provider_uri: str = None,
+                 payment_provider_uri: str = None,
```
derekpierre, line 401: Could you clarify why this is needed? See Line 409 which uses `payment_provider`.

theref: yep, you're totally right - there's a difference in naming convention here, we use `eth_provider_uri` and just `payment_provider`.  I think it would be cleaner if both were `*_uri`

nucypher/blockchain/eth/networks.py
KPrasch: Well in any case `__to_ethereum_chain_id_eth` is a bit redundant 🤔

nucypher/blockchain/eth/networks.py
```diff
@@ -30,18 +31,22 @@ class NetworksInventory:  # TODO: See #1564
...
         MAINNET: 1,      # Ethereum Mainnet
         ORYX: 3,         # Ropsten
         IBEX: 4,         # Rinkeby
         LYNX: 5,         # Goerli
-
+    }
+    __to_ethereum_chain_id_poly = {
```
KPrasch, line 40: Naming stuff: this name is confusing -- is it ethereum or polygon?

nucypher/config/base.py
derekpierre: Yep, it would be more consistent to use the "_uri" suffix, but therefore we should change the existing parameter used for that value on L409, and not add a new one - or else we'd have two parameters for the same thing (unless I'm misunderstanding).

I'm not sure about what the side effects are for changing the existing parameter name (cc @KPrasch ).

KPrasch: I think renaming this field to `payment_provider_uri` is okay, but I think a manual test against an existing configuration is warranted.  It's worth considering the TODO that reads: `Resolve code prefixing below, possibly with the use of nested configuration fields`. Also this appears to be a duplicate parameter to the one on line 409.

KPrasch (changes requested):
There's still a duplicated init parameter in base.py.

KPrasch:
This PR still has outstanding RFCs and needs a rebase to avoid merge conflicts.

KPrasch (approved):

derekpierre (approved):
🎸 

Minor question about adding a test. Looks good otherwise.

PR #2921, opened 2022-04-18
Add a pre-commit hook and github action for Darker

theref:
**Type of PR:**
- [x] Other

**Required reviews:** 
- [X] 3

**What this does:**
Add a pre-commit hook and github action for [Darker](https://github.com/akaihola/darker) to ensure all future changes conform to black and isort.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -52,6 +52,14 @@ repos:
         files: \.py$  # only target python files, for example - skip contract_registry.json
         args: [--license-filepath, '.nucypher-license', --detect-license-in-X-top-lines=16, --comment-style, '"""||"""']
 
+  - repo: https://github.com/akaihola/darker
+    rev: 1.4.2
+    hooks:
+    -   id: darker
+        args: [--isort]
+        additional_dependencies:
+        -   isort~=5.9
+
 
 # TODO: May be incrementally introduced
 #  - repo: https://github.com/PyCQA/bandit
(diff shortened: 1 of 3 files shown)


.pre-commit-config.yaml
```diff
@@ -52,6 +52,14 @@ repos:
         files: \.py$  # only target python files, for example - skip contract_registry.json
         args: [--license-filepath, '.nucypher-license', --detect-license-in-X-top-lines=16, --comment-style, '"""||"""']
 
+  - repo: https://github.com/akaihola/darker
+    rev: 1.4.2
+    hooks:
+    -   id: darker
+        args: [--isort]
```
derekpierre, line 59: Any thoughts on configuring line length - the default is `88`?

theref: I've never had a problem with 88 - i know monitors have got a lot bigger, but i regularly find myself working on laptops/opening multiple files side by side/looking at diffs on github.

Whatever we choose, someone is going to be unhappy! 😁

derekpierre: Haha fair enough 😄 .

derekpierre (approved):
🎸

cygnusv (approved):
LGTM. 
Re: line length, I usually have two visual rulers (80 and 120). 99% of lines I try to keep under 80, but sometimes it doesn't feel right. 120 is a hard stop for me in any case. Let's try with 88, we can always increase it a bit if we feel uncomfortable

KPrasch (approved):
Here we go 😅

PR #2924, opened 2022-04-21
Update github registry source path to use `development` instead of `main`

derekpierre:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
Retrieve contract registries from the `development` branch on GitHub instead of `main`. 

Extra changes are due to our new `darker` pre-commit hook.

**Why it's needed:**
Without this any subsequent contract_registries introduced as part of development, won't be accessible for GitHub download from the code until it gets included in a release and is on `main`.


derekpierre:
My local issues with the darker pre-commit hook and my workaround may have contributed to a larger diff than necessary.

derekpierre:
> My local issues with the darker pre-commit hook and my workaround may have contributed to a larger diff than necessary.

Should be resolved now.

KPrasch (approved):

KPrasch:
I can understand the counter-argument that only published releases should be specified here, but I think the agility of using the development branch here is worth it.

theref (approved):

PR #2943, opened 2022-05-17
work around for failure to estimate gas occuring on Ropsten

vepkenez:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Fixes the KeyError occurring on Ropsten/Oryx when node attempts to confirm worker address

### DIFF
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -562,9 +562,14 @@ class BlockchainInterface:
         #
 
         # TODO: Show the USD Price:  https://api.coinmarketcap.com/v1/ticker/ethereum/
-        max_tx_price = transaction_dict['maxFeePerGas']
-        max_priority_price = transaction_dict['maxPriorityFeePerGas']
-        max_price = max_tx_price + max_priority_price
+        
+        if transaction_dict.get('maxFeePerGas') and transaction_dict.get('maxPriorityFeePerGas'):
+            # TODO:  this probably only needs to be here to support Ropsten?
+            max_tx_price = transaction_dict['maxFeePerGas']
+            max_priority_price = transaction_dict['maxPriorityFeePerGas']
+            max_price = max_tx_price + max_priority_price
+        else:
+            max_price = transaction_dict['gasPrice']
         max_price_gwei = Web3.fromWei(max_price, 'gwei')
         max_cost_wei = max_price * transaction_dict['gas']
         max_cost = Web3.fromWei(max_cost_wei, 'ether')
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -564,10 +564,12 @@ class BlockchainInterface:
         try:
             # post-london fork transactions (Type 2)
             max_unit_price = transaction_dict['maxFeePerGas']
+            tx_type = 'EIP-1559'
         except KeyError:
             # pre-london fork "legacy" transactions (Type 0)
             max_unit_price = transaction_dict['gasPrice']
-            
+            tx_type = 'Legacy'
+
         max_price_gwei = Web3.fromWei(max_unit_price, 'gwei')
         max_cost_wei = max_unit_price * transaction_dict['gas']
         max_cost = Web3.fromWei(max_cost_wei, 'ether')
@@ -581,7 +583,7 @@ class BlockchainInterface:
         #
         # Broadcast
         #
-        emitter.message(f'Broadcasting {transaction_name} Transaction ({max_cost} ETH @ {max_price_gwei} gwei)',
+        emitter.message(f'Broadcasting {transaction_name} {tx_type} Transaction ({max_cost} ETH @ {max_price_gwei} gwei)',
                         color='yellow')
         try:
             txhash = self.client.send_raw_transaction(signed_raw_transaction)  # <--- BROADCAST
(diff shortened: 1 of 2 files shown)


derekpierre:
@vepkenez, just checking, has this workaround been tested on oryx and works?

@KPrasch is this the last PR for a v6.1.0 release?

nucypher/blockchain/eth/interfaces.py
```diff
@@ -562,9 +562,14 @@ def sign_and_broadcast_transaction(self,
...
         # TODO: Show the USD Price:  https://api.coinmarketcap.com/v1/ticker/ethereum/
-        max_tx_price = transaction_dict['maxFeePerGas']
-        max_priority_price = transaction_dict['maxPriorityFeePerGas']
-        max_price = max_tx_price + max_priority_price
+        
+        if transaction_dict.get('maxFeePerGas') and transaction_dict.get('maxPriorityFeePerGas'):
+            # TODO:  this probably only needs to be here to support Ropsten?
+            max_tx_price = transaction_dict['maxFeePerGas']
+            max_priority_price = transaction_dict['maxPriorityFeePerGas']
+            max_price = max_tx_price + max_priority_price
```
cygnusv, line 570: I don't think this is right, the TX price will never be higher than the `maxFeePerGas` (see https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.send_transaction).

The thing is that I don't see how web3py populates `maxFeePerGas` or `maxPriorityFeePerGas` to give you a complete opinion here, since both fields are optional, and to make it more complex, depend on the ETH provider. In principle, the simplest approach to make sure your TX will be mined is to just hard-code a max priority fee per gas (e.g., 1 or 1.5 gwei), and call it a day, but I haven't tried this with web3py (since they use to do a lot of weird processing of TX data on their side when building and sending).

Now, this code tries to estimate the actual cost of the TX. In that case, instead of querying `maxFeePerGas`, we should get the current `baseFee` and the estimated cost will be `baseFee + maxPriorityFeePerGas`

cygnusv (changes requested):

vepkenez:
@derekpierre yes this code is live on 3 Oryx nodes and allows the nodes to confirm worker address.
Without this code, but seemingly only when using Ropsten, the nodes do not confirm and instead fail repeatedly with a keyerror

```
    max_tx_price = transaction_dict[\'maxFeePerGas\']
builtins.KeyError: \'maxFeePerGas\'
```

the transaction_dict looks like
```
{'value': 0, 'gasPrice': 44797000000, 'chainId': 3, 'nonce': 0, 'from': '0x65F317Db8f48B59B97C9Abc7175D2bEf360caFc3', 'gas': 47018, 'to': '0xb6f98dA65174CE8F50eA0ea4350D96B2d3eFde9a', 'data': '0x9f795125'}
```

vepkenez:
Also, I attempted to install web3py== 6.0.0b2 but there were too many dependency conflicts with other things so I abandoned that.

derekpierre:
AFAIK this is the last issue to address before a v6.1.0 release - should we give it some priority? I know @cygnusv had some reservations about the current strategy, and @vepkenez responded.

derekpierre:
I've just encountered this issue on rinkeby (ibex), during the `unbond` operation.

```
  File "/Users/derek/Documents/Github/repos/forks/derek/nucypher/nucypher/cli/commands/bond.py", line 197, in unbond
    receipt = agent.bond_operator(operator=NULL_ADDRESS, transacting_power=transacting_power, staking_provider=staking_provider)
  File "/Users/derek/Documents/Github/repos/forks/derek/nucypher/nucypher/blockchain/eth/decorators.py", line 105, in wrapped
    return func(*args, **kwargs)
  File "/Users/derek/Documents/Github/repos/forks/derek/nucypher/nucypher/blockchain/eth/agents.py", line 557, in bond_operator
    transacting_power=transacting_power)
  File "/Users/derek/Documents/Github/repos/forks/derek/nucypher/nucypher/blockchain/eth/decorators.py", line 105, in wrapped
    return func(*args, **kwargs)
  File "/Users/derek/Documents/Github/repos/forks/derek/nucypher/nucypher/blockchain/eth/interfaces.py", line 668, in send_transaction
    fire_and_forget=fire_and_forget)
  File "/Users/derek/Documents/Github/repos/forks/derek/nucypher/nucypher/blockchain/eth/interfaces.py", line 565, in sign_and_broadcast_transaction
    max_tx_price = transaction_dict['maxFeePerGas']
KeyError: 'maxFeePerGas'
```

nucypher/blockchain/eth/interfaces.py
KPrasch: > The thing is that I don't see how web3py populates maxFeePerGas or maxPriorityFeePerGas 

Reading through the source code, I don't see an RPC call (unless I'm missing it?)  IT appears to just be a local gas estimation.  The lowest level call from the nucypher codebase is to `buildTransaction` which includes gas estimation.  Perhaps our usage is somehow signals a legacy transaction.

> In principle, the simplest approach to make sure your TX will be mined is to just hard-code a max priority fee per gas (e.g., 1 or 1.5 gwei), and call it a day, but I haven't tried this with web3py (since they use to do a lot of weird processing of TX data on their side when building and sending).

This particular block of code isn't used to populate a transaction to be broadcasted, it's just a console message for the user's convenience.  If it's too much trouble to debug this block, we can simply remove it with no consequences.

> Now, this code tries to estimate the actual cost of the TX. In that case, instead of querying maxFeePerGas, we should get the current baseFee and the estimated cost will be baseFee + maxPriorityFeePerGas

Since `maxFeePerGas = baseFee + maxPriorityFeePerGas` and I expect the ransaction fields to be filled by web3 for this type of transaction, I think we can juts set `max_price` here to `transaction_dict['maxFeePerGas']` and call it a day.

KPrasch (approved):

PR #2945, opened 2022-05-25
Add recommendations/information about auto restart capabilities when running a PRE node

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [X] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Add restart capabilities to docker/systemd documentation. Also document the restart strategy utilized by `nucypher-ops`.

This should be included in v6.1.0 so that it is included in the Read-the-docs `stable` documentation (cc @KPrasch)

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
It's a positive for nodes to be initially set up with auto restart functionality.

**Notes for reviewers:**

### DIFF
--- a/docs/source/pre_application/running_a_node.rst
+++ b/docs/source/pre_application/running_a_node.rst
@@ -222,7 +222,7 @@ This step starts the PRE node.
 
     $ docker run -d --rm \
     --name ursula      \
-    --restart unless-stopped          \
+    --restart on-failure              \
     -v ~/.local/share/nucypher:/root/.local/share/nucypher \
     -v ~/.ethereum/:/root/.ethereum   \
     -p 9151:9151                      \
@@ -234,9 +234,9 @@ This step starts the PRE node.
 
 .. note::
 
-    Setting the ``--restart`` parameter to ``unless-stopped`` ensures that the Docker
-    container will be automatically restarted if it exited, except if the container was
-    stopped using the ``docker stop`` command. See `Docker Restart Policies <https://docs.docker.com/engine/reference/run/#restart-policies---restart>`_
+    Setting the ``--restart`` parameter to ``on-failure`` ensures that the Docker
+    container will be automatically restarted if the application exited unexpectedly (e.g. an application crash).
+    See `Docker Restart Policies <https://docs.docker.com/engine/reference/run/#restart-policies---restart>`_
     for more information.
 
 View Node Logs


docs/source/pre_application/running_a_node.rst
```diff
@@ -230,6 +231,14 @@ This step starts the PRE node.
     nucypher/nucypher:latest          \
     nucypher ursula run
 
+
+.. note::
+
+    Setting the ``--restart`` parameter to ``unless-stopped`` ensures that the Docker
+    container will be automatically restarted if it exited, except if the container was
```
KPrasch, line 238: To make this a bit more useful for a less technical audience, perhaps we can say "crashed" instead of "exited" (or something similar)?

KPrasch (approved):

docs/source/pre_application/running_a_node.rst
derekpierre: Technically it does restart even if there was a valid exit (exit code of 0). Thinking more about this, perhaps the restart policy should be `on-failure` instead? 🤔  (cc @vepkenez )

I'm thinking about the case where a node is proactively shut down because it is no longer bonded to a staking provider.

vepkenez: > Technically it does restart even if there was a valid exit (exit code of 0). Thinking more about this, perhaps the restart policy should be `on-failure` instead? 🤔 (cc @vepkenez )
> 
> I'm thinking about the case where a node is proactively shut down because it is no longer bonded to a staking provider.

Like for example if we have a thing where we check the blockchain and kill the process if we are no longer bonded?
Yeah... maybe.  Or we could just handle that in the rest middleware and return 400s. Not sure.  It seems strange to just kill the process in that case.

derekpierre: We already have thing 😅  - https://github.com/nucypher/nucypher/pull/2886/commits/a8aa869c2d41a73f2fc11b8020e5655f0c1ad6d3.

derekpierre: ^ @KPrasch / @vepkenez

derekpierre: Went with `on-failure` instead of `unless-stopped`. @vepkenez something for you to think about with respect to `nucypher-ops`.

derekpierre:
Merging this PR. Any issues with using `on-failure` instead of `unless-stopped` (https://github.com/nucypher/nucypher/pull/2945#discussion_r890166268) can be addressed in a follow-up PR.

PR #2946, opened 2022-05-26
Show fleet state checksums as hex instead of an escaped bytestring

fjarri:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

**What this does:**
Turns
Screen Shot 2022-05-25 at 18 23 47
into
Screen Shot 2022-05-25 at 18 26 19

This has been bothering me for a while, and since it was originally my fault...


derekpierre (approved):
🎸

PR #2948, opened 2022-05-31
don't block on lack of funding

vepkenez:
**Type of PR:**
- [x] Bugfix

**Required reviews:** 
- [x] 1

**What this does:**
> Fix issue preventing confirmed nodes with 0 ETH on their operator address from starting

### DIFF
--- a/deploy/docker/porter/Dockerfile
+++ b/deploy/docker/porter/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.8.12-slim
+FROM nucypher/rust-python:3.8.12
 
 # Update
 RUN apt-get update -y && apt upgrade -y
--- a/newsfragments/2948.bugfix.rst
+++ b/newsfragments/2948.bugfix.rst
@@ -1,0 +1,1 @@
+Allow nodes which have already confirmed but have 0 ETH in their operator wallet to start.
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -350,7 +350,7 @@ class Operator(BaseActor):
         poll_rate = poll_rate or self.READY_POLL_RATE
         timeout = timeout or self.READY_TIMEOUT
         start, funded, bonded = maya.now(), False, False
-        while not (funded and bonded):
+        while not bonded:
 
             if timeout and ((maya.now() - start).total_seconds() > timeout):
                 message = f"x Operator was not qualified after {timeout} seconds"
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -357,21 +357,15 @@ class Operator(BaseActor):
                 emitter.message(message, color='red')
                 raise self.ActorError(message)
 
-            if not funded:
-                # check for funds
-                ether_balance = client.get_balance(self.operator_address)
-                if ether_balance:
-                    # funds found
-                    funded, balance = True, Web3.fromWei(ether_balance, 'ether')
-                    emitter.message(f"✓ Operator {self.operator_address} is funded with {balance} ETH", color='green')
-                else:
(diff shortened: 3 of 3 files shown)


derekpierre:
Posting our discussion from Discord here.

(Derek)
> Agreed. It would be cleaner to associate the funding with the confirm operator tx more directly if possible - just wondering why we didn't do that before 🤷

(Damon)
> I think it's because the earlier model needed to ensure funding for future transactions
the state of funding mattered even if there was no immediate transaction needed

(Derek)
> It's also the case that the worker that does the tx doesn't prohibit startup of the node whereas the bonded/funded checker does. The worker for the confirm operator tx just runs in the background. So if the node is not funded, do we want the node up and running (learning, being learned about) if it wasn't funded and the confirm operator tx continuing to fail i.e. is it ok for a node to be running for a prolonged period of time without the confirm_operator tx being executed...(?)

derekpierre:
> Why is it even run in the background? Is there a point in starting up the node if it's not confirmed?

Yep, that's my point indeed. Why allow the node to startup if it's not confirmed? 

For clarification the two steps we are talking about here are:
- Check bonding/funding worker (blocking task) that prevents further startup until conditions are satisfied
- Operator confirmed tx worker (background non-blocking task)

That operator confirmed tx worker is a vestige of the old continuous (always running) commitToNextPeriod tx worker that was repurposed. For Threshold, it was modified to run until the operator confirmed tx goes through and then shut itself down. 

However, it was never changed to halt startup until the tx is completed. The only prevention of startup that was/is done is the "bonding/funded" worker task in the step before.

In the current code, when funding was checked in the step before, it didn't really matter that the operator confirmed tx task ran in the background because once the node was assured to be funded before continuing node startup, the operator confirmed tx worker would be started and the tx would go through quickly. If we don't check funding in the blocking step before, and move it to the operator confirmed tx worker, the tx might not go through for quite some time if there is no funding, while the node is allowed to startup. 

So if we are moving the funding check to the operator confirmed worker, this step would need to have the ability to halt startup until the tx is executed since it could take a long time, and the node probably shouldn't start up.

In general, I think we tried to keep the change simple initially for the transition to Threshold, but this now needs some more thought - maybe even consolidation of these two separate tasks.

derekpierre:
@vepkenez , @KPrasch we still need to address this issue.

derekpierre:
> One reason to keep this structured similarly as before would be if we want to shut down the node in the event that it becomes unbonded,

We actually have that functionality already 😄 and in a separate task - https://github.com/nucypher/nucypher/commit/a8aa869c2d41a73f2fc11b8020e5655f0c1ad6d3.

vepkenez:
Ah, ok great I wasn't sure about that one.

KPrasch (commented):
LGTM - Are there any outstanding concerns blocking the merge of this PR?

derekpierre:
I remember now. Whether a node is bonded or not, is irrespective of funding. The only reason funding is checked is to ensure it can do the operator confirmed tx (if it hasn't done it already).

There is a simple (but less clean) fix here - https://discord.com/channels/411401661714792449/412681039178366996/981339908843130901. It takes an optimistic approach in the sense that if the operator address is funded with ETH, the operator confirmed tx is expected go through.

However, as @fjarri rightly pointed out in discord in that discussion, and here, there is a possible cleaner and more deterministic way to fix this; check that the node is funded just before executing the operator confirmed tx and then halt startup/shutdown. But this would be a more involved change since the current worker task that does the operator confirmed tx currently cannot shutdown the node/halt the node from starting up if the node is not funded. What we don't want is the node starting up without operator confirmed tx going through quickly which it won't if not funded.

(cc @KPrasch)

derekpierre (changes requested):
See previous comments.

KPrasch:
memorialized by https://github.com/nucypher/nucypher/issues/2983

PR #2951, opened 2022-06-25
tDec global keygen script

KPrasch:
Hackathon


scripts/tdec/keygen.py
```diff
@@ -0,0 +1,70 @@
...
+
+# Network Policy
+m, n = 1, 1
+label = 'dream-team'.encode()
+policy = alice.grant(bob=universal_bob,
+                     ursulas=ursulas,
+                     label=label,
+                     threshold=m,
+                     shares=n,
+                     duration=3600*365)
```
cygnusv, line 46: I think this policy only works for 365 hours (15 days), not a year

```suggestion
                     duration=86400*365)
```

cygnusv (changes requested):

derekpierre:
Rebase over the latest `development` for the `cli` build tests to pass - the issue is the probation period needing to be extended which it was in a recent PR.

scripts/tdec/network_config.json
```diff
@@ -0,0 +1,9 @@
...
+    "pek": "03a061abec02f3ffa2688a709a7499c666f871fccd39b021d3b36e266eeb0e4500",
+    "bob_verifying_key": "032581042cb430610cd60f447bd4a887d76deb1b339c188052d80a24614daf4cb6",
+    "bob_verifying_secret": "77956dd5f8c26681b1c87ed9af30eb3066d93d7277056e826f59b1a5f21ad0be",
+    "bob_encrypting_key": "03c7f656269b7b51301275667011ff203717bb5fc30b18998839a9d350eb445096",
+    "bob_encrypting_secret": "073e656ca21122485b3a2b4102f668c70436d91bdf9a140f5ea3fb45dc33fa4f"
```
arjunhassard, line 8: Just to be clear. Forthcoming versions will have gatekeeping conditions, right now the conditions are 'there are no conditions', so anyone can decrypt (and anyone can encrypt to the policy too!). And these private keys are basically there for the entire network – so everyone uses the same keys and they're not really private in any sense. Is this secure, even when there is strict conditionality?

scripts/tdec/keygen.py
```diff
@@ -0,0 +1,70 @@
...
+
+# Store artifacts
+network_tmap = policy.treasure_map
+network_pek = policy.public_key
+
+payload = {
+    'alice_public_key': bytes(alice.stamp).hex(),
+    'tmap': bytes(network_tmap).hex(),
+    'pek': bytes(network_pek).hex(),
+    # TODO: store Alice's secret?
```
arjunhassard, line 57: What power does Alice's secret give her in perpetuity? Does it mean she can never actually be revoked from the policy, even when she consents to be revoked – either as part of an interactive atomized NFT selling+signing function, or conditionally/non-interactively (if someone pays this amount, revoke me) – plus an honest majority of Ursulas that execute it?

KPrasch:
Obsoleted by https://github.com/nucypher/network-keys

PR #2961, opened 2022-08-22
Remove core shims; Refactor for nucypher-core conditions; Support web3.py v6+

KPrasch:
Title says it all. This PR supports the tDec epic (#2960)

Depends on https://github.com/nucypher/nucypher-core/pull/26
https://github.com/nucypher/nucypher-core/pull/27

PR #2962, opened 2022-08-30
Allow Porter to provide `context` for re-encryption conditions

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
Based over #2961 .
Related to https://github.com/nucypher/nucypher-ts/pull/82.
Related to https://github.com/nucypher/tdec/issues/39.

**Issues fixed/closed:**
Allows an optional `context` parameter to be provided as part of the retrieval flow.

**Why it's needed:**
`Context` is needed to provide the necessary information to determine whether the recipient satisfies the relevant conditions for getting data re-encrypted/decrypted.

**Notes for reviewers:**
CI tests will continue to fail until an updated version of `nucypher-core` is used.

### DIFF
--- a/nucypher/utilities/porter/control/specifications/porter_schema.py
+++ b/nucypher/utilities/porter/control/specifications/porter_schema.py
@@ -159,5 +159,18 @@ class BobRetrieveCFrags(BaseSchema):
             type=click.STRING,
             required=True))
 
+    # optional
+    context = base_fields.Base64JSON(
+        required=False,
+        load_only=True,
+        click=click.option(
+            "--context",
+            "-ctx",
+            help="Context data for retrieval conditions",
+            type=click.STRING,
+            required=False,
+        ),
+    )
+
     # output
     retrieval_results = marshmallow_fields.List(marshmallow_fields.Nested(fields.RetrievalResultSchema), dump_only=True)
--- a/nucypher/utilities/porter/control/specifications/porter_schema.py
+++ b/nucypher/utilities/porter/control/specifications/porter_schema.py
@@ -160,7 +160,7 @@ class BobRetrieveCFrags(BaseSchema):
             required=True))
 
     # optional
-    context = base_fields.Base64JSON(
+    context = base_fields.JSON(
         required=False,
         load_only=True,
         click=click.option(
--- a/nucypher/utilities/porter/control/specifications/porter_schema.py
+++ b/nucypher/utilities/porter/control/specifications/porter_schema.py
@@ -161,6 +161,7 @@ class BobRetrieveCFrags(BaseSchema):
 
     # optional
     context = base_fields.JSON(
+        expected_type=dict,
         required=False,
         load_only=True,
         click=click.option(
--- a/tests/acceptance/porter/control/test_porter_web_control_blockchain.py
+++ b/tests/acceptance/porter/control/test_porter_web_control_blockchain.py
@@ -18,15 +18,17 @@
 import json
 import os
 from base64 import b64encode
-from urllib.parse import urlencode
 
 from nucypher_core import RetrievalKit
 
 from nucypher.characters.lawful import Enrico
-from nucypher.control.specifications.fields import Base64JSON
+from nucypher.control.specifications.fields import JSON
 from nucypher.crypto.powers import DecryptingPower
 from nucypher.policy.kits import PolicyMessageKit, RetrievalResult
-from nucypher.utilities.porter.control.specifications.fields import RetrievalResultSchema, RetrievalKit as RetrievalKitField
+from nucypher.utilities.porter.control.specifications.fields import (
+    RetrievalResultSchema,
+    RetrievalKit as RetrievalKitField,
+)
 from tests.utils.middleware import MockRestMiddleware
 from tests.utils.policy import retrieval_request_setup, retrieval_params_decode_from_rest
 
@@ -173,27 +175,13 @@ def test_retrieve_cfrags(blockchain_porter,
     #
     # Use context
     #
-    context_field = Base64JSON()
+    context_field = JSON()
     multiple_retrieval_kits_params['context'] = context_field._serialize(random_context, attr=None, obj=None)
 
     response = blockchain_porter_web_controller.post('/retrieve_cfrags', data=json.dumps(
         multiple_retrieval_kits_params))
     assert response.status_code == 200
 
-    response_data = json.loads(response.data)
-    retrieval_results = response_data['result']['retrieval_results']
-    assert retrieval_results
-    assert len(retrieval_results) == 2
-
-    #
-    # Try same retrieval (with multiple retrieval kits) using query parameters
-    #
-    url_retrieve_params = dict(multiple_retrieval_kits_params)  # use multiple kit params from above
-    # adjust parameter for url query parameter list format
-    url_retrieve_params['retrieval_kits'] = ",".join(url_retrieve_params['retrieval_kits'])   # adjust for list
-    response = blockchain_porter_web_controller.post(f'/retrieve_cfrags'
-                                                     f'?{urlencode(url_retrieve_params)}')
-    assert response.status_code == 200
     response_data = json.loads(response.data)
     retrieval_results = response_data['result']['retrieval_results']
     assert retrieval_results
--- a/tests/acceptance/porter/control/test_porter_web_control_blockchain.py
+++ b/tests/acceptance/porter/control/test_porter_web_control_blockchain.py
@@ -23,6 +23,7 @@ from urllib.parse import urlencode
 from nucypher_core import RetrievalKit
 
 from nucypher.characters.lawful import Enrico
+from nucypher.control.specifications.fields import Base64JSON
 from nucypher.crypto.powers import DecryptingPower
 from nucypher.policy.kits import PolicyMessageKit, RetrievalResult
 from nucypher.utilities.porter.control.specifications.fields import RetrievalResultSchema, RetrievalKit as RetrievalKitField
@@ -91,7 +92,8 @@ def test_retrieve_cfrags(blockchain_porter,
                          blockchain_porter_web_controller,
                          random_blockchain_policy,
                          blockchain_bob,
-                         blockchain_alice):
+                         blockchain_alice,
+                         random_context):
     # Send bad data to assert error return
     response = blockchain_porter_web_controller.post('/retrieve_cfrags', data=json.dumps({'bad': 'input'}))
     assert response.status_code == 400
@@ -161,6 +163,21 @@ def test_retrieve_cfrags(blockchain_porter,
         retrieval_kit_field._serialize(value=retrieval_kit_2, attr=None, obj=None)
     ]
     response = blockchain_porter_web_controller.post('/retrieve_cfrags', data=json.dumps(multiple_retrieval_kits_params))
+    assert response.status_code == 200
+
+    response_data = json.loads(response.data)
+    retrieval_results = response_data['result']['retrieval_results']
+    assert retrieval_results
+    assert len(retrieval_results) == 2
+
+    #
+    # Use context
+    #
+    context_field = Base64JSON()
+    multiple_retrieval_kits_params['context'] = context_field._serialize(random_context, attr=None, obj=None)
+
+    response = blockchain_porter_web_controller.post('/retrieve_cfrags', data=json.dumps(
+        multiple_retrieval_kits_params))
     assert response.status_code == 200
 
     response_data = json.loads(response.data)
--- a/docs/source/application_development/web_development.rst
+++ b/docs/source/application_development/web_development.rst
@@ -674,9 +674,9 @@ Parameters
 +-------------------------------------------+---------------+----------------------------------------+
 | ``bob_verifying_key``                     | String        | Bob's verifying key encoded as hex.    |
 +-------------------------------------------+---------------+----------------------------------------+
-| ``context`` *(Optional)*                  | String        | | Associated data required during      |
+| ``context`` *(Optional)*                  | String        | | Associated JSON data required during |
 |                                           |               | | re-encryption e.g. data to satisfy   |
-|                                           |               | | conditions.                          |
+|                                           |               | | re-encryption conditions.            |
 +-------------------------------------------+---------------+----------------------------------------+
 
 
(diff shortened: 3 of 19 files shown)


tests/unit/conditions/test_core_shims.py
```diff
@@ -13,7 +13,7 @@ def test_conditional_message_kit_serialization(enacted_federated_policy, ERC1155
     )
     mk_bytes = bytes(mk)
     restored_mk = MessageKit.from_bytes(mk_bytes)
-    assert restored_mk.lingos is None
+    assert restored_mk.conditionss is None
```
piotr-roslaniec, line 16:
```suggestion
    assert restored_mk.conditions is None
```

tests/unit/conditions/test_core_shims.py
```diff
@@ -24,4 +24,4 @@ def test_conditional_message_kit_serialization(enacted_federated_policy, ERC1155
     )
     mk_bytes = bytes(mk)
     restored_mk = MessageKit.from_bytes(mk_bytes)
-    assert bytes(restored_mk.lingos) == bytes(ERC1155_balance_condition)
+    assert bytes(restored_mk.conditionss) == bytes(ERC1155_balance_condition)
```
piotr-roslaniec, line 27:
```suggestion
    assert bytes(restored_mk.conditions) == bytes(ERC1155_balance_condition)
```

derekpierre: Yep - see <https://github.com/nucypher/nucypher/pull/2961#discussion_r957407413> in #2961 . I don't want to touch that code and cause conflicts.

tests/unit/conditions/test_core_shims.py
derekpierre: Yep - see https://github.com/nucypher/nucypher/pull/2961#discussion_r957407413 in #2961 . Good eye 👍

nucypher/utilities/porter/control/specifications/porter_schema.py
```diff
@@ -159,5 +159,18 @@ class BobRetrieveCFrags(BaseSchema):
             type=click.STRING,
             required=True))
 
+    # optional
+    context = base_fields.Base64JSON(
```
piotr-roslaniec, line 163: Just what I needed :ok_hand:

derekpierre: @piotr-roslaniec: Question for you (need your web expertise).

One of the reasons I created Base64JSON was to allow for URL request parameters. Does this make sense to do? We can't pass nested dicts as URL parameters so if the context JSON had nested dicts in it, that would be problematic. The Base64 encoding takes care of handling this.

Interested to get your thoughts. Should we support URL parameters? Porter did before.

piotr-roslaniec: @derekpierre 
IIRC the original reason for using URL-encoded parameters was that the browsers don't like `GET` requests with JSON body. Base64JSON makes sense because it solves the issue of sending complex types (dictionaries) as parameters in URLs. And since context etc. is never going to be used on its own etc., the fact that the context type and structure are obfuscated (base64-encoded) will not cause any issues.

derekpierre: > IIRC the original reason for using URL-encoded parameters was that the browsers don't like GET requests with JSON body

Does it change your mind at all given that the retrieve endpoint is a `POST` and not a `GET`?

piotr-roslaniec: Well, actually, yes. In that case, `Base64JSON` is not _necessary_. 

So I think I should ask, is there any reason why we shouldn't send plain JSON then? Maybe we'd like to obfuscate the condition context because the structure/API will be unstable, so it's better to present it as Base64JSON? 

I'm just throwing ideas, but if there are no objections then I think plain JSON is fine. What do you think @derekpierre ?

derekpierre: > Maybe we'd like to obfuscate the condition context because the structure/API will be unstable
IMHO JSON is robust enough to handle instability in the API. I guess any obfuscation of condition context would probably take the form of a proof in the context...maybe (?)

On the simplicity/readability side, plain JSON is good. If there are no need for URL parameters for a POST endpoint, then I'm pro just plain JSON.

cc @KPrasch for his thoughts.

derekpierre: FYI - what the code would look like if we switched to plain JSON - https://github.com/nucypher/nucypher/pull/2962/commits/27330b4d5f2e3b59fa61327d46d39551dca1b5e8

piotr-roslaniec:
While testing locally, I've noticed that Porter throws `'nucypher_core.RetrievalKit' object has no attribute 'conditions'`. Reproducible example:

```
curl 'http://127.0.0.1/retrieve_cfrags' \
  -H 'Accept: application/json, text/plain, */*' \
  -H 'Accept-Language: en-US,en;q=0.9' \
  -H 'Cache-Control: no-cache' \
  -H 'Connection: keep-alive' \
  -H 'Content-Type: application/json' \
  -H 'Origin: http://127.0.0.1:8080' \
  -H 'Pragma: no-cache' \
  -H 'Referer: http://127.0.0.1:8080/' \
  -H 'Sec-Fetch-Dest: empty' \
  -H 'Sec-Fetch-Mode: cors' \
  -H 'Sec-Fetch-Site: same-site' \
  -H 'User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/102.0.5005.61 Safari/537.36' \
  -H 'sec-ch-ua: " Not A;Brand";v="99", "Chromium";v="102"' \
  -H 'sec-ch-ua-mobile: ?0' \
  -H 'sec-ch-ua-platform: "Linux"' \
  --data-raw '{"treasure_map":"VE1hcAABAACVAsQQ1PslxJp4amgrhJquipcK6ITEFFP3J4fTOymw8dVIX0sLEoH76OYTksRiA1l2MEJbqiRxVMIYApX1zwX8bvja+5DdQnynTJo/jFMxAmIhSO8d034KsVR7pE+b5Hx9cLy0+H6tTPmQwTSDXQ0YIEDG7+NYoENBg3UQuvOTZVXrszfq/+bH3w5y0tn4cd3FAXpfLOFArb4EXV+NhMQs+rSnfh6q2AP7uhVe7+QiiRvhScdoT2y4luyw37tZCKGsqXaijA/sxCL2tbd2gvDbUICMoauYK8Z/yAbNqCEJLEjMTlyWq5e1hE+1UvJzeCCxI+PpXG0gvvFGTn/zwrim782SZfrEyf56YkAXMmXYHLzMewJRRIJ8HBbq4hTfM8HRZLnUYMz3URQD/L+kYlXzylzdfrC6MQQ+r3OAdPRfR7YRhw/jb8l472uZTfaDy2Y18mASsPRSMn9GRegz/CfMAQ4bUBX5kE1drDmJH/etzC1GvOl0/bTz5+fBUxiYPo9/WhF6gh1l6S6vHgb/cN8hp+mlws1nZr8E/HDbTPF2gY+raLz5mzWgHIBqnRccsvW5Apa9n0kdX95D6VuPSCiV4q9ZbVls3onjNJ30IBAH5tLd26Op7hBiuBxY2NviUZ58RvlYKoJsH9ZT9KN434EkjTADpEKLIfkZeQfVHJyFqTWLNE4oaz3BI2wa7tzEFLBIinStBtwNogMcvXIFN8SNY+/XksRiA8U8zm17fYVNLT5mdb9ic1+vmAASvt8NTUQkkdog/dT1AnkRVrGrYLdOZChDt2eHVlpSFEIEaZooreaY3CGEXJn9pt2tsRhCBuRROug/IcB3vLnW20sMKbN0lEuKladlNDLFAXqJMUJdBLPdfPqGxR9GcsH+cBLWYelc5uocpZ0cp7gQUKMW4mMM2cqBRUVcb0K8yUVLSxUB9K028I8a+i/9QdKhIEGJJIB54iTYiOmoClK55toy2XrnPMD/ObP6sY6ubk5vmabLFZ5LgDG7DLKIzpgCEM5bEhC8AlBtsPajuM5iY6aiVXBWYdNAkfdzJYoWCQjEkNscRVRrbORfqgso1vPVah8kuTYW7XkYi3pcBiqRfAw+4I7i3w0c0AUvv9TFv0ubaamCZSOUrzrB1eg9gZUXCiOhqr7qjleMQ2YM06KBCrwndqayJmqgrhCY++Byv9k4MWoQ4CMdyOBC7SyYmsQtC7ErHYJtPIWWPIwDRSeZWxcxIsZQaLj3ZliaE8JVUeiobj+oAxgXJiK34nc5g8w8+H2Be4PK4x8euOcP9e9ysp2SIwtyLrjoEAGKimNu/gJWDKv/dLyCZACv6X6GkgHorWdGSAT6bQGduHP3pRQrl6fdvajtGzGtjvnEFLxArgBBubCuPQEqQEkrQtHg7ilPksRiA1JLRmPxmJWCm8N+TUCn75C1QTirtt6AOAm3Ca2p1hYCApXvQOU6rqeAAMqJNb4Cw9J4MwRVF7Dbe7AfhHp7mApADYwRuVYoRILgKrr/DWbyR/awhornOQPd1X08iSqxahLFAXqdNG6EY1eHYCqMvKc+uc1wdK2lGch4hUvCTdffXEYI2/Dj5+wU9OmgD/bRGPXpGBOXehDI/fPcSpYmisyx1nCPGOx1fW6SQ3Ip1HopU59yCT4+dc50dMSyBTKOStMw8t+JSgmXNUV4a0E6UQGpCvjXY8E3p9rYNpedrFT1n+fckBi4TJ1ogJmDX01cND+fcnryI2NiYUWEWiwEOLYM1w2CnPfO1rhiEKWvDUEGndXzDAyycu4nczKd+s30WVL/mjHSLaNWhTQLydHQwMMf75UQfENtGfxziofVuLQ/SD3/CvdbsdWMZJSSk2bRY6knOQzeZUDNFxEqntnyOnDBFe0OsueYyHfHes1Ylsa3qW7bmsZl+UPgFqwkMA2vK0xq3q7oAxjK6v7fHlODAyWvSXBBF/Fbtwq/EPr2+7E2csUukGetllL8YBo56T2gBCpT/ZNGwj/vrzvG6cpD0BH3+eNdREKuP5rEIvtHmzNNSPbqBrXc3tRBZpz/PqzEFL0n5BOxRg1R/htrS2Tl2vw1QbErksRiA1G751uvb+1qjAC/8EBTVBX2sFyPNNm902OnP0gI+dGtAp9ZIRLmluV0aBNzx9MkDXu2VPLFOVf+7HQ0bLND6rv5bNCjsgSINPmP8LEjXRhePWP5GWRf6KpXrKLHhp01vQ7FAXpvfJLErFCmLaTwsRfQ3lgTLwXjQAMKBsV7uNLyWEYwnDVbm8OGNoyw6i1uVldLc1fhkZsRYIgKzCfy/bFbNAmsH7ADNbK2anG8t7s/12Hlq8AhKH8Pt/qtpllTDsl9yXuRNnTQDgjAropw/FoQHDoWwGV1IG+UqVBP2FqOY2i/FIi9u7yqsNojEjt8b1mjmgUSUOhPS5AuwPivUwAHHkNJik97R0KPEHUBXfFqG8O6ZorkNjK0fi0Z0cxUcaq2tc5VSYg/VQ7dqLv64kgTU8vHvoj9ghQe9T+NjPUi+NqYlap/dzkyiZyiFFdc4/W5RXmzZqMwaiEQAc91YW09AJTuGX4HFDHFftKiUc8wMVvNWVlYqYpvn50hAdfkEuWQ62tpgp7lztSF8iB75kZWk1tRNbSX3veg3SaBKS6svc/x3+nc9w4S7TiMhzmLwB1TgpLcHCtkbI4CBtx+Q2sZlVuR3h6QMf9GbxmwmrXxpw3dUSh4CYzTqhi+NMLEIQOeJ/98fJLSlKnylfM5hvTEnz8cHS9WxyTw0A2A2ZqzRcQhA4S8E0JBW5h0VycxPRCuzRteuitWaIBNPGGXbl/ztuUu","retrieval_kits":["UktpdAABAACSxGIChApfylLkkDQbtcHrwxLdaTI3/qPWPJBv3CzCm76KcaAD2amoHlRIlRgYCUfqv15Ju2zmOy9CBY7KuBc+kwqUFKq0AOR+8q8J1lnUg49qkmC1/UOTIcZltxSSgq0g5ucC55A="],"alice_verifying_key":"0384bc1342415b98745727313d10aecd1b5eba2b5668804d3c61976e5ff3b6e52e","bob_encrypting_key":"03d1ed998900f47d956d2736019d71417716bcc0a83090c36c7e38adc88e99c8f9","bob_verifying_key":"03172e4d4b27e83641ecf539a1650d9f74b9e23455da597b96aa03e3caca3409da","context":"{\"signature\":\"0xc4029685c3ea31b226072b546876e533293a066d2539e5d9830f2aa69b75ee40283e61b3eabd83345c091d279a03c578f8bc7e33f84285899f4fad42c5121c5c1b\"}"}' \
  --compressed
```

KPrasch:
Hmmm possibly the result of a missing getter binding?

derekpierre:
@piotr-roslaniec Is that just a dependency issue with `nucypher-core`? Things may have changed since I've been OOO, but I needed to build `nucypher-core` (`fizzy` branch) locally to use the updated code since there isn't an associated released version of `nucypher-core` with the required changes. Once an updated version is released, the `nucypher-core` dependency version used by `nucypher` will need to be updated. 

In the meantime, you'll probably need to manually build/install the latest `nucypher-core` yourself.

nucypher/utilities/porter/control/specifications/porter_schema.py
derekpierre: Opted for plain JSON.

tests/acceptance/porter/control/test_porter_web_control_blockchain.py
```diff
@@ -169,14 +173,15 @@ def test_retrieve_cfrags(blockchain_porter,
...
-    url_retrieve_params = dict(multiple_retrieval_kits_params)  # use multiple kit params from above
-    # adjust parameter for url query parameter list format
-    url_retrieve_params['retrieval_kits'] = ",".join(url_retrieve_params['retrieval_kits'])   # adjust for list
-    response = blockchain_porter_web_controller.post(f'/retrieve_cfrags'
-                                                     f'?{urlencode(url_retrieve_params)}')
+    context_field = JSON()
+    multiple_retrieval_kits_params['context'] = context_field._serialize(random_context, attr=None, obj=None)
+
+    response = blockchain_porter_web_controller.post('/retrieve_cfrags', data=json.dumps(
+        multiple_retrieval_kits_params))
```
KPrasch, line 182: 👍🏻 👍🏻

KPrasch (approved):

PR #2965, opened 2022-09-13
Initial context variable work (No validation)

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Handle injection of context variable value into condition. Initially starting with `:userAddress` without any signature validation - simply obtain the address from the information provided in the `context`.

**Issues fixed/closed:**
Related to https://github.com/nucypher/tdec/issues/39.

**Why it's needed:**
Allows for simple evaluation of conditions that use the `:userAddress` context variable, and require the user address to be provided dynamically i.e. as part of the `context`.

**Notes for reviewers:**

### DIFF
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -27,11 +27,11 @@ from flask import Flask, Response, jsonify, request
 from mako import exceptions as mako_exceptions
 from mako.template import Template
 from nucypher_core import (
-    ReencryptionRequest,
-    RevocationOrder,
     MetadataRequest,
     MetadataResponse,
     MetadataResponsePayload,
+    ReencryptionRequest,
+    RevocationOrder,
 )
 
 from nucypher.config.constants import MAX_UPLOAD_CONTENT_LENGTH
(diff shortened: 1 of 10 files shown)


derekpierre:
@KPrasch / @piotr-roslaniec if this looks reasonable we can deploy to Ursulas for further testing.

piotr-roslaniec:
@derekpierre It does, I just have one question: Are context variables such as `:hrac` and `:addressToUse` just examples or should I somehow accommodate them in `nucypher-ts`?

derekpierre:
Currently, the only supported context variable with an associated function for handling is `:userAddress`. The current idea is that there would be specific context variables we would code special handling for (`:userAddress` for now) with possibly more to come later, but the user configuring the condition could also create customized ones themselves that would be set using key-value pairs in the `context` with no special handling. An example of that could be `:hrac`.

For now, I would only focus on `:userAddress` and wouldn't focus on the custom variables just yet - I did it just as an idea based on prior conversations with @KPrasch  - it's not confirmed functionality and possibly needs more thought, but it does add some user flexibility.

piotr-roslaniec (approved):
I see. Ok, then I'd say this PR is in line with my expectations and we're good to go

derekpierre:
@piotr-roslaniec / @vepkenez if there have been successful testing of this code from `nucypher-ts` i.e. using `:userAddress` context variables and having conditions pass / fail depending on the address used, can we merge this PR?

KPrasch (approved):
🤠

PR #2972, opened 2022-09-20
eval is evil 

cygnusv:
A small fix that reduces the risk of dangerous input that exploit our use of `eval()`, although this PR doesn't remove `eval()` completely.

To solve this once and for all I'd prefer using a different approach, which affects how we define compound conditions (see https://github.com/nucypher/tdec/issues/59)

**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -62,12 +62,12 @@ class ReturnValueTest:
         def make(self, data, **kwargs):
             return ReturnValueTest(**data)
 
-    def __init__(self, comparator: str, value: Union[int, str]):
-        comparator, value = self.sanitize(comparator, value)
+    def __init__(self, comparator: str, value):
+        comparator, value = self._sanitize(comparator, value)
         self.comparator = comparator
         self.value = value
 
-    def sanitize(self, comparator: str, value: str) -> Tuple[str, str]:
+    def _sanitize(self, comparator: str, value) -> Tuple[str, Any]:
         if comparator not in self.COMPARATORS:
             raise ValueError(f'{comparator} is not a permitted comparator.')
         return comparator, ast.literal_eval(str(value))
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -61,15 +61,13 @@ class ReturnValueTest:
         self.value = value
 
     def sanitize(self, comparator: str, value: str) -> Tuple[str, str]:
-        # Note that we're not sanitizing the value
         if comparator not in self.COMPARATORS:
             raise ValueError(f'{comparator} is not a permitted comparator.')
-        return comparator, value
+        return comparator, ast.literal_eval(str(value))
 
     def eval(self, data) -> bool:
-        # TODO: Sanitize input!!!
         left_operand = ast.literal_eval(str(data))
-        right_operand = ast.literal_eval(str(self.value))
+        right_operand = self.value
         result = eval(f'{left_operand}{self.comparator}{right_operand}')
         return result
 
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -1,3 +1,4 @@
+import ast
 import base64
 import json
 from typing import Union, Tuple, List, Dict, Any
@@ -60,13 +61,16 @@ class ReturnValueTest:
         self.value = value
 
     def sanitize(self, comparator: str, value: str) -> Tuple[str, str]:
+        # Note that we're not sanitizing the value
         if comparator not in self.COMPARATORS:
             raise ValueError(f'{comparator} is not a permitted comparator.')
         return comparator, value
 
     def eval(self, data) -> bool:
         # TODO: Sanitize input!!!
-        result = eval(f'{data}{self.comparator}{self.value}')
+        left_operand = ast.literal_eval(str(data))
+        right_operand = ast.literal_eval(str(self.value))
+        result = eval(f'{left_operand}{self.comparator}{right_operand}')
         return result
 
 
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -45,7 +45,14 @@ class Operator:
 
 
 class ReturnValueTest:
-    COMPARATORS = ('==', '>', '<', '<=', '>=')
+    _COMPARATOR_FUNCTIONS = {
+        '==': lambda x, y: x == y,
+        '>': lambda x, y: x > y,
+        '<': lambda x, y: x < y,
+        '<=': lambda x, y: x <= y,
+        '>=': lambda x, y: x >= y,
+    }
+    COMPARATORS = tuple(_COMPARATOR_FUNCTIONS.keys())
 
     class ReturnValueTestSchema(CamelCaseSchema):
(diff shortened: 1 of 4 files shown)


tests/unit/conditions/test_return_value.py
```diff
@@ -1,14 +1,19 @@
...
+    assert test.eval('1')
+    assert not test.eval('-1')
 
 
 def test_return_value_sanitization():
     with pytest.raises(ValueError):
         _test = ReturnValueTest('DROP', 'TABLE')
+
+def test_eval_is_evil():
+    test = ReturnValueTest(comparator='>', value="(lambda: (__import__('os').system(f'touch hola'), 42)[1])()")
```
KPrasch, line 18: 👾

tests/unit/conditions/test_return_value.py
```diff
@@ -1,14 +1,23 @@
...
 
 
 def test_return_value_sanitization():
     with pytest.raises(ValueError):
         _test = ReturnValueTest('DROP', 'TABLE')
+
+def test_eval_is_evil():
+    file = f"file_{os.urandom(10).hex()}"
+    assert not os.path.isfile(file)
+    test = ReturnValueTest(comparator='>', value=f"(lambda: (42, __import__('os').system('touch {file}'))[0])()")
```
derekpierre, line 20: 😱

nucypher/policy/conditions/lingo.py
```diff
@@ -60,13 +61,16 @@ def __init__(self, comparator: str, value: Union[int, str]):
         self.value = value
 
     def sanitize(self, comparator: str, value: str) -> Tuple[str, str]:
+        # Note that we're not sanitizing the value
         if comparator not in self.COMPARATORS:
             raise ValueError(f'{comparator} is not a permitted comparator.')
         return comparator, value
```
derekpierre, line 67: Any reason we don't do `ast.literal_eval(value)` here, during `sanitize` at creation time, instead of doing it during `eval(...)`?

cygnusv: Good point. It makes sense. I guess I was going it later for symmetry, since we need to validate also the `data` that's evaluated on the condition, but the sooner the better

arjunhassard (approved):
If eval() and its untrusted input risks are avoidable with prefix notation, this update would be obsolete right? Or is sanitization still necessary?

cygnusv:
> If eval() and its untrusted input risks are avoidable with prefix notation, this update would be obsolete right? Or is sanitization still necessary?

Correct, this PR is just a temporary fix due to the use of eval, but some level of sanitization will still be necessary

tests/unit/conditions/test_return_value.py
```diff
@@ -1,14 +1,23 @@
...
+    with pytest.raises(ValueError):
+        _ = test.eval('1000')
+    assert not os.path.isfile(file)
```
KPrasch, line 23: It's good that this attack is no longer exploitable but we can take it a step further by raising an error as early as possible (in `ReturnValueTest.__init__`) if the value is not acceptable, in another PR of course.

tests/unit/conditions/test_return_value.py
```diff
@@ -1,14 +1,23 @@
...
 from nucypher.policy.conditions.lingo import ReturnValueTest
 
 
 def test_return_result_test_simple():
-    test = ReturnValueTest(comparator='>', value=0)
-    assert test.eval(1)
-    assert not test.eval(-1)
+    test = ReturnValueTest(comparator='>', value='0')
+    assert test.eval('1')
+    assert not test.eval('-1')
```
KPrasch, line 10: Shall we continue to require string exclusively?  I see this kind of type casting spreading in the codebase a bit.

tests/unit/conditions/test_return_value.py
derekpierre: Yep, that's related to https://github.com/nucypher/nucypher/pull/2972#discussion_r977898875.

nucypher/policy/conditions/lingo.py
```diff
@@ -60,13 +61,16 @@ def __init__(self, comparator: str, value: Union[int, str]):
...
 
     def eval(self, data) -> bool:
         # TODO: Sanitize input!!!
-        result = eval(f'{data}{self.comparator}{self.value}')
+        left_operand = ast.literal_eval(str(data))
+        right_operand = ast.literal_eval(str(self.value))
```
KPrasch, line 72: For other reviewers this is how we're gaining safety in this PR

KPrasch (approved):

piotr-roslaniec (approved):
LGTM

tests/unit/conditions/test_return_value.py
cygnusv: I created a test that shows that literal values are accepted too.

cygnusv:
@KPrasch @derekpierre @piotr-roslaniec I updated the PR to remove completely one of the usages of `eval()`. Note that the second use, which evaluates compound boolean conditions, is still there, but it should be way less risky as inputs to that eval are only boolean values resulting from `ReturnValueTest`.

nucypher/policy/conditions/lingo.py
```diff
@@ -62,11 +70,12 @@ def __init__(self, comparator: str, value: Union[int, str]):
     def sanitize(self, comparator: str, value: str) -> Tuple[str, str]:
```
derekpierre, line 70: We should adjust the typehint here. If `value` is already a `str` then we shouldn't need to do `str(value)` on L73. 
Seems like our options are:
1. Not have a typehint for value
OR
2. Have the typehint for value match what is in the constructor i.e. `Union[int, str]` 
OR
3. We pass `str(value)` to `sanitize(...)` on L66. 

wdyt?

Side note: should we change `sanitize(...)` ->  `_sanitize(...)` since it is only used by the constructor? Is there any expected usage outside of that?

nucypher/policy/conditions/lingo.py
```diff
@@ -62,11 +70,12 @@ def __init__(self, comparator: str, value: Union[int, str]):
...
         if comparator not in self.COMPARATORS:
             raise ValueError(f'{comparator} is not a permitted comparator.')
-        return comparator, value
+        return comparator, ast.literal_eval(str(value))
```
derekpierre, line 73: 👍

jMyles (approved):
Some old school shit.

Attacks are back on the menu boys!

derekpierre (approved):
🎸

nucypher/policy/conditions/lingo.py
```diff
@@ -44,7 +45,14 @@ def to_json(self) -> str:
 
 
 class ReturnValueTest:
-    COMPARATORS = ('==', '>', '<', '<=', '>=')
+    _COMPARATOR_FUNCTIONS = {
+        '==': lambda x, y: x == y,
```
derekpierre, line 49: @cygnusv side note: found these python builtin methods you can use instead of the lambdas,https://docs.python.org/3/library/operator.html;  up to you if you want to use them or not.

PR #2973, opened 2022-09-21
Porter Error Handling for `/retrieve_cfrags` endpoint

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Initial schema for communicating errors that occur during `/retrieve_cfrags` back to the caller.

**Issues fixed/closed:**
Fixes https://github.com/nucypher/tdec/issues/57
Fixes https://github.com/nucypher/nucypher/issues/2789

**Why it's needed:**
Porter needs to provide better feedback on errors that occur during retrieval.

Initial version looks like:
```json
 "result": {
        "retrieval_results": [
            {
                "cfrags": {
                    // ...
                },
                "errors": {
                    "<ursula_address>": "<error_message>",
                    // category I - network is down, ursulas are wrong, everything is falling apart
                    "<ursula>": "Internal Server Error",
                     // sir, stop breaking the porter plz
                    "<ursula>": "Something wrong with your parameters buddy: ciphertext is undefined",
                    // condition verification on ursula - actionable feedback for the user
                    "<ursula>": "Decryption conditions not satisfied",
                    // auth - actionable feedback for the user; stale auth
                    "<ursula>": "Context variable data could not be verified"
                }
            }
        ]
    }, 
    "version": "6.1.0"
```

**Notes for reviewers:**

PR #2974, opened 2022-09-23
ERC721 `ownerOf` support

derekpierre:
**Type of PR:**
- [X] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
This change is really just a one-liner to ensure that we only check whether a parameter is a context variable or not, if it is provided as a string. Everything else is just testing.

*OF NOTE: ERC1155*
The code doesn't yet properly support ERC1155. I modified what the parameters for ERC1155 balanceOfBatch
```
balanceOfBatch(address[] accounts, uint256[] ids) → uint256[]
```
should look like in `test_conditions.json`. The parameter types in the condition need to match the parameter types in the contract call. It is only a definition for now, and if we agree, we can add support for ERC1155 later on, with proper processing of list parameters - which we don't currently support.

I also suspect how the result of the contract call is processed and compared needs to be adjusted since it returns a list of values, and not a single value.

**Issues fixed/closed:**
Based over #2973 .

**Why it's needed:**
@piotr-roslaniec was doing some testing with ERC721 `ownerOf` conditions and ran into the following issue.

When specifying the `tokenId` as an int for the `ownerOf` condition, the following error was returned by Ursula:
```json
"errors":{
    "0xbD27e413b1460d51fe1b6b4b64E5dAFC3541b12B":"UnexpectedResponse: b\"Unexpected exception while evaluating decryption condition (AttributeError): 'int' object has no attribute 'startswith'\"",
    "0xbC40aE0041b9b0AE3D012a40492B42d1E0EE294F":"UnexpectedResponse: b\"Unexpected exception while evaluating decryption condition (AttributeError): 'int' object has no attribute 'startswith'\"",
    "0xb0488A74Ad06DC0DA2031cBD720537C48d63EfD7":"UnexpectedResponse: b\"Unexpected exception while evaluating decryption condition (AttributeError): 'int' object has no attribute 'startswith'\"",
    "0x53f72787d33b29b0f1d5485f4b0B1281FbE8e613":"UnexpectedResponse: b\"Unexpected exception while evaluating decryption condition (AttributeError): 'int' object has no attribute 'startswith'\""
}
```

The issue here is that when checking whether a provided parameter is a context variable or not, while processing parameters we incorrectly assume that the parameter is already a string, and do the following:
```python
for p in parameters:
    if is_context_variable(p):
    ...
```

where the `is_context_variable(p)` function is defined as:
```python
def is_context_variable(parameter: str) -> bool:
    return parameter.startswith(_CONTEXT_PREFIX)
```
which fails because the provided `tokenId` parameter value is of type int.

However, the `tokenId` parameter in the ERC721 ownerOf condition can't just be made into a string, because the parameters are passed directly from the json loaded type to the contract, and as such must match the type in the contract call.

The definition of the ERC721 `ownerOf` contract function is
```
ownerOf(uint256 tokenId) → address owner external
```

To prove this when provided as a string, the follow error occurs:
```json
"errors":{
    "0xb0488A74Ad06DC0DA2031cBD720537C48d63EfD7":"b\"\\nCould not identify the intended function with name `ownerOf`, positional arguments with type(s) `str` and keyword arguments with type(s) `{}`.\\nFound 1 function(s) with the name `ownerOf`: ['ownerOf(uint256)']\\nFunction invocation failed due to no matching argument types.\"",
    "0x53f72787d33b29b0f1d5485f4b0B1281FbE8e613":"b\"\\nCould not identify the intended function with name `ownerOf`, positional arguments with type(s) `str` and keyword arguments with type(s) `{}`.\\nFound 1 function(s) with the name `ownerOf`: ['ownerOf(uint256)']\\nFunction invocation failed due to no matching argument types.\"",
    "0xbC40aE0041b9b0AE3D012a40492B42d1E0EE294F":"b\"\\nCould not identify the intended function with name `ownerOf`, positional arguments with type(s) `str` and keyword arguments with type(s) `{}`.\\nFound 1 function(s) with the name `ownerOf`: ['ownerOf(uint256)']\\nFunction invocation failed due to no matching argument types.\"",
    "0xbD27e413b1460d51fe1b6b4b64E5dAFC3541b12B":"b\"\\nCould not identify the intended function with name `ownerOf`, positional arguments with type(s) `str` and keyword arguments with type(s) `{}`.\\nFound 1 function(s) with the name `ownerOf`: ['ownerOf(uint256)']\\nFunction invocation failed due to no matching argument types.\""
}
```

**Notes for reviewers:**


nucypher/policy/conditions/evm.py
```diff
@@ -75,7 +75,9 @@ def _process_parameters(parameters, **request) -> List:
     """Handles request parameters"""
     processed_parameters = []
     for p in parameters:
-        if is_context_variable(p):
+        # TODO needs additional support for ERC1155 which has lists of values
+        # context variables can only be strings, but other types of parameters can be passed
+        if type(p) == str and is_context_variable(p):
```
piotr-roslaniec, line 80: :+1:

piotr-roslaniec (approved):
LGTM

cygnusv (approved):
Looks great. My only minor comment is that the type check before `is_context_variable()` seems to me like something that should be done inside that function.

derekpierre:
@cygnusv nice find! Fixed via https://github.com/nucypher/nucypher/pull/2974/commits/720bdc8b85690755cc00aa0810714778420a11ed

nucypher/policy/conditions/context.py
```diff
@@ -101,8 +101,8 @@ def _recover_user_address(**context) -> ChecksumAddress:
 }
 
 
-def is_context_variable(parameter: str) -> bool:
-    return parameter.startswith(_CONTEXT_PREFIX)
+def is_context_variable(parameter) -> bool:
+    return type(parameter) == str and parameter.startswith(_CONTEXT_PREFIX)
```
KPrasch, line 105:
```suggestion    
    return isinstance(parameter, str) and parameter.startswith(_CONTEXT_PREFIX)
```

nucypher/policy/conditions/evm.py
```diff
@@ -148,12 +157,17 @@ def _configure_provider(self, provider: BaseProvider):
...
         self._configure_provider(provider=provider)
-        parameters = _process_parameters(self.parameters, **contract_kwargs)  # resolve context variables
-        rpc_endpoint_, rpc_method = self.method.split('_', 1)
+        parameters = _process_parameters(
+            self.parameters, **contract_kwargs
+        )  # resolve context variables
+        return_value_test = _process_return_value_test(
+            self.return_value_test, **contract_kwargs
+        )  # resolve context variables
+        rpc_endpoint_, rpc_method = self.method.split("_", 1)
```
KPrasch, line 166: Perhaps move this to a helper / dehydration method to avoid duplication below.

nucypher/policy/conditions/evm.py
```diff
@@ -212,10 +226,17 @@ def _get_unbound_contract_function(self) -> ContractFunction:
...
+        )  # resolve context variables
+        return_value_test = _process_return_value_test(
+            self.return_value_test, **contract_kwargs
+        )  # resolve context variables
+        bound_contract_function = self.contract_function(
+            *parameters
+        )  # bind contract function
         contract_result = bound_contract_function.call()  # onchain read
-        eval_result = self.return_value_test.eval(contract_result)  # test
+        eval_result = return_value_test.eval(contract_result)  # test
```
KPrasch, line 239: 👍🏻

nucypher/policy/conditions/lingo.py
```diff
@@ -45,39 +67,69 @@ def to_json(self) -> str:
...
+        try:
+            return ast.literal_eval(str(value))
+        except Exception:
+            raise self.InvalidExpression(f'"{value}" is not a permitted value.')
 
     def eval(self, data) -> bool:
-        left_operand = ast.literal_eval(str(data))
-        right_operand = self.value
+        if is_context_variable(self.value):
+            # should never get in here
```
KPrasch, line 114:
```suggestion
            # programming error
```

nucypher/policy/conditions/lingo.py
```diff
@@ -45,39 +67,69 @@ def to_json(self) -> str:
...
+            raise RuntimeError(
+                "'value' should never be a context variable here - this object "
+                "should have been replaced by one whose context variable's value"
+                "has been populated"
```
KPrasch, line 118:
```suggestion
                "'value' must not be a context variable at evaluation time.
```

nucypher/policy/conditions/lingo.py
```diff
@@ -45,39 +67,69 @@ def to_json(self) -> str:
...
+            )
+        left_operand = self._sanitize_value(data)
+        right_operand = self._sanitize_value(self.value)
         result = self._COMPARATOR_FUNCTIONS[self.comparator](left_operand, right_operand)
```
KPrasch, line 122: Is a `KeyError` here is impossible?

nucypher/policy/conditions/lingo.py
```diff
@@ -45,39 +67,69 @@ def to_json(self) -> str:
...
         return result
 
+    def __eq__(self, other):
+        if not isinstance(other, ReturnValueTest):
+            return False
+        try:
+            return (self.comparator == other.comparator) and (self.value == other.value)
+        except AttributeError:
+            return False
```
KPrasch, line 131: This is for testing purposes, or how else might it be used?

tests/data/test_conditions.json
```diff
@@ -31,12 +31,12 @@
...
     "parameters": [
-      ":userAddress,:userAddress,:userAddress,:userAddress",
-      "1,2,10003,10004"
+      [":userAddress",":userAddress",":userAddress",":userAddress"],
+      [1,2,10003,10004]
     ],
     "returnValueTest": {
       "comparator": ">",
-      "value": "0"
+      "value": 0
```
KPrasch, line 39: 👍🏻

tests/unit/conditions/conftest.py
```diff
@@ -113,8 +128,16 @@ def timelock_condition():
...
-        conditions=[timelock_condition, OR, rpc_condition, AND, evm_condition]
+        conditions=[
+            erc721_evm_condition,
+            OR,
+            timelock_condition,
+            OR,
+            rpc_condition,
+            AND,
+            erc20_evm_condition,
+        ]
```
KPrasch, line 141: 🤩

nucypher/policy/conditions/lingo.py
derekpierre: For testing purposes only for now...open to an alternative idea...

derekpierre: I guess in testing I could add this function in the test...wdyt?

nucypher/policy/conditions/lingo.py
derekpierre: No, because we check the comparator on init.

nucypher/policy/conditions/lingo.py
```diff
@@ -66,40 +67,69 @@ def to_json(self) -> str:
 
 
 class ReturnValueTest:
+    class InvalidExpression(ValueError):
+        pass
```
cygnusv, line 71: 🔥

tests/unit/conditions/test_return_value.py
```diff
@@ -73,14 +128,80 @@ def test_return_value_test_string():
...
-def test_eval_is_evil():
-    file = f"file_{os.urandom(10).hex()}"
-    assert not os.path.isfile(file)
-    with pytest.raises(ValueError):
-        _test = ReturnValueTest(comparator='>', value=f"(lambda: (42, __import__('os').system('touch {file}'))[0])()")
+    # mixing types works because the value is evaluated as an int, not a string
+    test = ReturnValueTest(
+        comparator="==", value="0xaDD9D957170dF6F33982001E4c22eCCdd5539118"
+    )
+    assert test.eval(992513598061863249514433594012370110655092723992)
```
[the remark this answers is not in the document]
derekpierre: Sure. It's analogously tested in the `!=` case on L141, but I'll add it here as well to be explicit.

tests/unit/conditions/test_return_value.py
```diff
@@ -73,14 +128,80 @@ def test_return_value_test_string():
...
+
+def test_return_value_test_none():
+    test = ReturnValueTest(comparator="!=", value=None)
+    assert test.eval("0xaDD9D957170dF6F33982001E4c22eCCdd5539118")
+
+    test = ReturnValueTest(comparator="==", value=None)
+    assert not test.eval("0xaDD9D957170dF6F33982001E4c22eCCdd5539118")
+
+    test = ReturnValueTest(comparator="==", value=None)
+    assert test.eval("None")
```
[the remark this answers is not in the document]
derekpierre: Yes. Although note that "\"None\"" does not - see L154.

nucypher/policy/conditions/lingo.py
```diff
@@ -66,40 +67,69 @@ def to_json(self) -> str:
...
         ">=": operator.ge,
-        "!=": operator.ne,
     }
-    COMPARATORS = tuple(_COMPARATOR_FUNCTIONS.keys())
+    COMPARATORS = tuple(_COMPARATOR_FUNCTIONS)
 
     class ReturnValueTestSchema(CamelCaseSchema):
         comparator = fields.Str()
-        value = fields.Str()
+        value = fields.Raw(allow_none=True)  # any valid type including None
```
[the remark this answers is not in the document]
derekpierre: My thought here was that anything can be returned from a contract call including objects, and sometimes you don't care what the value/object is, just that something returned.

For example, `SubscriptionManager.getPolicy(...)` which returns a Policy object or not - if you just care that there is a policy you can just check that the result of the contract call is` != None` without caring about the object returned. Of course `SubscriptionManager.isPolicyActive(...)` allows you to potentially work around this, but `getPolicy(...)` is illustrative.

nucypher/policy/conditions/lingo.py
derekpierre: Without the `instanceof` check you would be able to pass any object with `comparator` and `value` fields eg. 

```python
class Rando(NamedTuple):
    comparator: str
    value
    other_member:...
```
that would pass the equality check.


> I'm having a hard time imagining a use for this equality test. What do you envision?

The equality test is only used for testing. See test `/tests/unit/conditions/test_return_value.py::test_return_value_serialization`.

I can modify the test to not require `__eq__` to be defined if that is preferable...?

derekpierre: Removed the `__eq__` - the comparison is done directly in the test.

nucypher/policy/conditions/lingo.py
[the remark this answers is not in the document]
derekpierre: Chatted more with @cygnusv , and solidity doesn't have any concept of `None`/null, so the benefit of `None` functionality is not that clear at the moment. I've removed it for now - and we can revisit later if that functionality will prove useful for a use case.

tests/acceptance/blockchain/conditions/test_conditions.py
```diff
@@ -253,44 +263,81 @@ def test_erc721_evm_condition_balanceof_evaluation(
...
+    zeroized_policy_struct = (
+        "0x0000000000000000000000000000000000000000",
+        0,
+        0,
+        0,
+        "0x0000000000000000000000000000000000000000",
+    )
+    context = {
+        ":hrac": bytes(enacted_blockchain_policy.hrac),
+        ":expectedPolicyStruct": zeroized_policy_struct,
```
cygnusv, line 319: Great use of context variables!

cygnusv (approved):
:chef_kiss:  (we need that emoji)

PR #2980, opened 2022-09-30
Handle RPC Condition call failure and other failure modes

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Just a simple musing for now about exception handling until we decide what we are doing regarding https://github.com/nucypher/tdec/issues/60. This is along the lines of the 2. option mentioned here - https://github.com/nucypher/tdec/issues/60#issuecomment-1262791297.

**Issues fixed/closed:**
- https://github.com/nucypher/tdec/issues/60

**Why it's needed:**

**Notes for reviewers:**

In https://github.com/nucypher/tdec/issues/60 there were other possible ideas about handling exceptions potentially checking them as condition results as well. This strategy just propagates the error message back to the caller. Other scenarios can be explored.

### DIFF
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -42,6 +42,7 @@ from nucypher.datastore.models import ReencryptionRequest as ReencryptionRequest
 from nucypher.network.exceptions import NodeSeemsToBeDown
 from nucypher.network.nodes import NodeSprout
 from nucypher.network.protocols import InterfaceInfo
+from nucypher.policy.conditions.base import ReencryptionCondition
 from nucypher.policy.conditions.context import (
     ContextVariableVerificationFailed,
     InvalidContextVariableData,
@@ -236,6 +237,11 @@ def _make_rest_app(datastore: Datastore, this_node, log: Logger) -> Flask:
                     # TODO: Can conditions return a useful value?
                     log.info(f'Evaluating decryption condition')
                     lingo.eval(**context)
+                except ReencryptionCondition.InvalidCondition as e:
+                    message = f"Incorrect value provided for condition: {e}"
+                    error = (message, HTTPStatus.BAD_REQUEST)
+                    log.info(message)
+                    return Response(message, status=error[1])
                 except RequiredContextVariable as e:
                     message = f"Missing required inputs: {e}"
                     # TODO: be more specific and name the missing inputs, etc
@@ -250,6 +256,11 @@ def _make_rest_app(datastore: Datastore, this_node, log: Logger) -> Flask:
                 except ContextVariableVerificationFailed as e:
                     message = f"Context variable data could not be verified: {e}"
                     error = (message, HTTPStatus.FORBIDDEN)
+                    log.info(message)
+                    return Response(message, status=error[1])
+                except ReencryptionCondition.ConditionEvaluationFailed as e:
+                    message = f"Decryption condition not evaluated: {e}"
+                    error = (message, HTTPStatus.BAD_REQUEST)
                     log.info(message)
                     return Response(message, status=error[1])
                 except lingo.Failed as e:
--- a/nucypher/policy/conditions/base.py
+++ b/nucypher/policy/conditions/base.py
@@ -64,6 +64,14 @@ class _Serializable:
 
 class ReencryptionCondition(_Serializable, ABC):
 
+    class InvalidCondition(ValueError):
+        """Invalid value for condition."""
+        pass
+
+    class ConditionEvaluationFailed(Exception):
+        """Could not evaluate condition for some reason."""
+        pass
+
     class Schema(Schema):
         name = NotImplemented
 
--- a/nucypher/policy/conditions/base.py
+++ b/nucypher/policy/conditions/base.py
@@ -66,11 +66,9 @@ class ReencryptionCondition(_Serializable, ABC):
 
     class InvalidCondition(ValueError):
         """Invalid value for condition."""
-        pass
 
     class ConditionEvaluationFailed(Exception):
         """Could not evaluate condition for some reason."""
-        pass
 
     class Schema(Schema):
         name = NotImplemented
--- a/nucypher/policy/conditions/evm.py
+++ b/nucypher/policy/conditions/evm.py
@@ -175,13 +175,17 @@ class RPCCondition(ReencryptionCondition):
             )
         return provider
 
-    def _execute_call(self, parameters: List[Any]) -> Any:
-        """Execute onchain read and return result."""
-        rpc_endpoint_, rpc_method = self.method.split("_", 1)
+    def _get_web3_py_function(self, rpc_method: str):
         web3_py_method = camel_case_to_snake(rpc_method)
         rpc_function = getattr(
             self.w3.eth, web3_py_method
         )  # bind contract function (only exposes the eth API)
+        return rpc_function
+
+    def _execute_call(self, parameters: List[Any]) -> Any:
+        """Execute onchain read and return result."""
+        rpc_endpoint_, rpc_method = self.method.split("_", 1)
+        rpc_function = self._get_web3_py_function(rpc_method)
         rpc_result = rpc_function(*parameters)  # RPC read
         return rpc_result
 
--- a/nucypher/policy/conditions/evm.py
+++ b/nucypher/policy/conditions/evm.py
@@ -26,7 +26,7 @@ from web3.contract import ContractFunction
 from web3.providers import BaseProvider
 
 from nucypher.blockchain.eth.clients import PUBLIC_CHAINS
-from nucypher.policy.conditions import STANDARD_ABIS
+from nucypher.policy.conditions import STANDARD_ABI_CONTRACT_TYPES, STANDARD_ABIS
 from nucypher.policy.conditions._utils import CamelCaseSchema
 from nucypher.policy.conditions.base import ReencryptionCondition
 from nucypher.policy.conditions.context import get_context_value, is_context_variable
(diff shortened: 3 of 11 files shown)


nucypher/network/server.py
```diff
@@ -252,6 +258,11 @@ def reencrypt():
                     error = (message, HTTPStatus.FORBIDDEN)
                     log.info(message)
                     return Response(message, status=error[1])
+                except ReencryptionCondition.ConditionEvaluationFailed as e:
+                    message = f"Decryption condition not evaluated: {e}"
```
KPrasch, line 262: In this block the condition was evaluated but the result was false.  It's more accurate to say something along the lines of "decryption conditions unfulfilled"

nucypher/policy/conditions/base.py
```diff
@@ -64,6 +64,14 @@ def from_bytes(cls, data: bytes) -> '_Serializable':
 
 class ReencryptionCondition(_Serializable, ABC):
 
+    class InvalidCondition(ValueError):
+        """Invalid value for condition."""
+        pass
```
KPrasch, line 69: `pass` is not needed here since you have a docstring (same below)

nucypher/policy/conditions/evm.py
```diff
@@ -140,32 +148,49 @@ def chain(self) -> str:
...
 
+    def _execute_call(self, parameters: List[Any]) -> Any:
+        """Execute onchain read and return result."""
+        rpc_endpoint_, rpc_method = self.method.split("_", 1)
+        web3_py_method = camel_case_to_snake(rpc_method)
+        rpc_function = getattr(
+            self.w3.eth, web3_py_method
+        )  # bind contract function (only exposes the eth API)
+        rpc_result = rpc_function(*parameters)  # RPC read
+        return rpc_result
```
KPrasch, line 179: 👍🏻 It's good to see this extracted.

nucypher/network/server.py
derekpierre: This is not the actual condition failing (i.e. there is no condition result), but rather the failure of the function that is needed to determine the value to use to check the condition.

KPrasch: Ah, okay I misunderstood.

derekpierre: In that case, perhaps the name of the exception can be made clearer? I was trying not to use `Function` in the name so that `ReencryptionCondition` can remain sufficiently abstract. Maybe something like `ConditionValueNotObtained` or `ConditionCheckNotEvaluated`...?

KPrasch: Is it a matter of validity?

nucypher/policy/conditions/evm.py
```diff
@@ -47,19 +47,23 @@ def _resolve_chain(chain: Union[str, int]) -> Tuple[str, int]:
...
+        )
 
 
 def _resolve_abi(standard_contract_type: str, method: str, function_abi: List) -> List:
     """Resolves the contract an/or function ABI from a standard contract name"""
     if not (function_abi or standard_contract_type):
         # TODO: Is this protection needed?
-        raise ValueError(f'Ambiguous ABI - Supply either an ABI or a standard contract name.')
+        raise ReencryptionCondition.InvalidCondition(
+            f"Ambiguous ABI - Supply either an ABI or a standard contract name."
```
KPrasch, line 60: Do you think it's too inflexible in list specifically supported standards in this error message?

nucypher/policy/conditions/evm.py
```diff
@@ -47,19 +47,23 @@ def _resolve_chain(chain: Union[str, int]) -> Tuple[str, int]:
...
+        )
     try:
         function_abi = STANDARD_ABIS[standard_contract_type]
     except KeyError:
         if not function_abi:
-            raise Exception(f'No function ABI found')
+            raise ReencryptionCondition.InvalidCondition(f"No function ABI found")
```
KPrasch, line 66:
```suggestion
            raise ReencryptionCondition.InvalidCondition(f"No function ABI supplied for {method}")
```

nucypher/policy/conditions/evm.py
```diff
@@ -100,6 +104,10 @@ class RPCCondition(ReencryptionCondition):
         'eth_getBalance',
     )
 
+    class RPCExecutionFailed(ReencryptionCondition.ConditionEvaluationFailed):
+        """Raised when an exception is raised from an RPC call."""
+        pass
```
KPrasch, line 109:
```suggestion
```

nucypher/policy/conditions/evm.py
```diff
@@ -140,32 +148,49 @@ def chain(self) -> str:
 
     def validate_method(self, method):
         if method not in self.ALLOWED_METHODS:
-            raise Exception(f'{method} is not a permitted RPC endpoint for conditions.')
+            raise ReencryptionCondition.InvalidCondition(
+                f"{method} is not a permitted RPC endpoint for conditions."
```
KPrasch, line 152:
```suggestion
                f"{method} is not a permitted RPC endpoint for condition evaluation."
```

KPrasch: If we don't know what other methods to include in the allowed list right now (besides `balanceOf` and eth_getBalance`), let's track it with an issue and corresponding TODO.

nucypher/policy/conditions/evm.py
```diff
@@ -140,32 +148,49 @@ def chain(self) -> str:
...
+            )
         if not method.startswith('eth_'):
-            raise Exception(f'Only eth RPC methods are accepted for conditions.')
+            raise ReencryptionCondition.InvalidCondition(
+                f"Only eth RPC methods are accepted for conditions."
```
KPrasch, line 156:
```suggestion 
                f"Only eth RPC methods are permitted for condition evaluation."
```

nucypher/policy/conditions/evm.py
```diff
@@ -140,32 +148,49 @@ def chain(self) -> str:
...
+            )
         return provider
 
+    def _execute_call(self, parameters: List[Any]) -> Any:
+        """Execute onchain read and return result."""
+        rpc_endpoint_, rpc_method = self.method.split("_", 1)
+        web3_py_method = camel_case_to_snake(rpc_method)
+        rpc_function = getattr(
+            self.w3.eth, web3_py_method
+        )  # bind contract function (only exposes the eth API)
```
KPrasch, line 177: Perhaps you can extract this bit to yet another helper method for further isolation and improved mocking capability.

nucypher/policy/conditions/evm.py
```diff
@@ -217,24 +242,23 @@ def _configure_provider(self, *args, **kwargs):
...
+            contract = self.w3.eth.contract(
+                address=self.contract_address, abi=self.function_abi
+            )
+            contract_function = getattr(
+                contract.functions, self.method
+            )  # TODO: Use function selector instead/also?
+            return contract_function
+        except Exception as e:
+            raise ReencryptionCondition.InvalidCondition(
+                f"Unable to obtain contract function, '{self.method}', for condition: {e}"
```
KPrasch, line 255: This line is suggests that including a "name" or "nickname" metadata field in the condition might be useful.

nucypher/policy/conditions/evm.py
derekpierre: you mean something like:

```suggestion
            f"Ambiguous ABI - Supply either an ABI or a standard contract name (ERC20, ERC721, ERC1155)."
```

KPrasch: Yes, exactly.

KPrasch: Although, it will be more maintainable if we have a single constant somewhere that is used both for the error message and the actual validity check.

derekpierre: Yeah, read my mind. I was thinking of having a set with the standard contract types. We don't have that currently.

There is some validation work we need to eventually do on the Schemas, and having the set of values will help with validation of that field.

cygnusv (commented):
Some preliminary comments

nucypher/policy/conditions/evm.py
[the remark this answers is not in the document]
KPrasch: Agreed - It feels a bit low-level and brittle.

nucypher/policy/conditions/evm.py
piotr-roslaniec: Should we also provide a list of currently allow-listed methods in the exception message?

tests/unit/conditions/test_condition_validity.py
```diff
@@ -0,0 +1,99 @@
...
+def test_invalid_contract_condition():
+    # no abi or contract type
+    with pytest.raises(ReencryptionCondition.InvalidCondition):
+        _ = ContractCondition(
+                contract_address="0xaDD9D957170dF6F33982001E4c22eCCdd5539118",
+                method="getPolicy",
+                chain="testerchain",
+                return_value_test=ReturnValueTest('!=', 0),
+                parameters=[
+                    ':hrac',
```
piotr-roslaniec, line 71: So the current list of context parameters is `[':hrac', ':userAddress', ':tokenId']`, correct?

derekpierre: No. The only "supported" context variable (by that I mean with actual special functionality behind it) is `:userAddress`. All others are free-form / user-defined and are just key-value pairs provided in the context.

nucypher/policy/conditions/evm.py
derekpierre: Might get a bit long over time depending on how https://github.com/nucypher/tdec/issues/64 shakes out. However, we can revisit then.

derekpierre:
Addressed various RFCs.

KPrasch (approved):

nucypher/policy/conditions/evm.py
derekpierre: 👍

piotr-roslaniec (approved):

PR #2985, opened 2022-10-18
Retire Alice, Bob, Contacts CLI

KPrasch:
**Type of PR:**
Code Removal

**Required reviews:** 
2

**What this does:**
Deprecates the `alice`, `bob`, and `contacts` CLI, supporting tests.  There may still be some dangling CLI option definitions which can be cleaned up.


tests/acceptance/cli/test_mixed_configurations.py
```diff
@@ -63,163 +58,6 @@ def test_destroy_with_no_configurations(click_runner, custom_filepath):
...
-                       '--registry-filepath', str(agency_local_registry.filepath.absolute()),
-                       '--config-root', str(custom_filepath.absolute()))
-
-    result = click_runner.invoke(nucypher_cli, alice_init_args, catch_exceptions=False, env=envvars)
-    assert result.exit_code == 0
-
-    # All configuration files still exist.
-    assert alice_file_location.is_file()
-
-    # Use the same local filepath to init a persistent Ursula
```
derekpierre, line 128: Should we keep the Ursula aspect of this test, or is similar testing already done separately?

newsfragments/2985.removal.rst
```diff
@@ -0,0 +1,2 @@
+Deprecates alice, bob, and card CLI commands.
```
derekpierre, line 1:
```suggestion
Deprecates alice, bob, and contacts CLI commands.
```

manumonti (approved):
LGTM!

tests/acceptance/cli/test_federated_cli_lifecycle.py
```diff
@@ -1,32 +0,0 @@
...
-from tests.acceptance.cli.lifecycle import run_entire_cli_lifecycle
-
-
-@pt.inlineCallbacks
-def test_federated_cli_lifecycle(click_runner,
-                                 random_policy_label,
-                                 federated_ursulas,
-                                 custom_filepath,
-                                 custom_filepath_2):
-    yield run_entire_cli_lifecycle(click_runner,
```
derekpierre, line 28: Should the `tests/acceptance/cli/lifecycle.py` file also be deleted? I believe the decentralized/federated cli_lifecycle files are the only ones using it.

tests/acceptance/cli/test_mixed_configurations.py
KPrasch: Good question -I had a good think about this.  I came to the conclusion that the original intent of this test is to prevent leakage between different types of configurations, and since we no longer expect someone to run (for example) and alice and an ursula on the same host, it's no longer relevant.

tests/acceptance/cli/test_federated_cli_lifecycle.py
KPrasch: You're right, it can be removed.  I will make an RFC commit.

tests/acceptance/cli/test_mixed_configurations.py
KPrasch: Indeed, there is good integration and unit test coverage for the config API.  It can't hurt to reintroduce a basic CLI config test for ursula to make sure we're testing that layer.

arjunhassard (approved):
The scope of PRE beyond the PoC version of CBD is not decided, but regardless a unified demo/walkthrough/sandbox style (i.e. browser-based) for all the access control products makes more sense

tests/acceptance/cli/test_mixed_configurations.py
KPrasch: Upon further investigation, there's ample ursula CLI config acceptance testing.

derekpierre (approved):
🎸

PR #2986, opened 2022-10-18
Payment provider adapted condition evaluation

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
- Supports condition evaluation by ursula on polygon chains
- Internal API support for generic multi-chain condition evaluation

**Notes**
Additional follow-up #3004


nucypher/policy/conditions/evm.py
```diff
@@ -32,24 +32,18 @@
...
+_CONDITION_CHAINS = {
+    1: "Ethereum",  # TODO: 60 ?
+    5: "Goerli",
+    # 6: "Kotti",
+    # 42: "Kovan",
+    # 77: "Sokol",
+    # 100: "xDai",
+    137: "Polygon/Mainnet",
+    80001: "Polygon/Mumbai"
+}
```
theref, line 46: Can we use the names as defined by https://chainid.network/chains_mini.json in order to maintain consistency?

KPrasch: Decided to remove the names altogether for this PR.

nucypher/policy/conditions/evm.py
```diff
@@ -25,31 +25,20 @@
...
-        if chain in pair:
-            chain_id, chain_name = pair
-            return chain_name, chain_id
-    else:
-        raise ReencryptionCondition.InvalidCondition(
-            f"{chain} is not a known blockchain."
-        )
+
+# Permitted blockchains for condition evaluation
+_CONDITION_CHAINS = (
```
theref, line 36: 🚀

tests/constants.py
```diff
@@ -143,3 +143,4 @@
 
 CLI_ENV = {NUCYPHER_ENVVAR_KEYSTORE_PASSWORD: INSECURE_DEVELOPMENT_PASSWORD,
            NUCYPHER_ENVVAR_OPERATOR_ETH_PASSWORD: INSECURE_DEVELOPMENT_PASSWORD}
+TESTERCHAIN_CHAIN_ID = 131277322940537
```
theref, line 146: Yes, my favourite ETH testnet

theref (commented):
Love the change to chain_id

theref (approved):

theref (changes requested):
I think chain id's will need to be added to `tests/data/test_conditions.json`

nucypher/network/retrieval.py
```diff
@@ -272,7 +272,6 @@ def _request_reencryption(self,
 
         ursula_verifying_key = ursula.stamp.as_umbral_pubkey()
 
-        # FIXME: Uncomment
```
derekpierre, line 275: 😅

nucypher/policy/conditions/evm.py
```diff
@@ -102,6 +91,13 @@ def _resolve_any_context_variables(
     return processed_parameters, processed_return_value_test
 
 
+def _validate_chain(chain: int):
+    if not isinstance(chain, int):
+        raise ValueError(f'"chain" must be a the integer of a chain ID (got "{chain}").')
```
derekpierre, line 96:
```suggestion
        raise ValueError(f'"chain" must be an integer representing the blockchain ID (got "{chain}").')
```

derekpierre: Feels like other than for our internal testing of old tests (since we started off with a non-integer chain id) the check below should be sufficient (L97), and this one is not needed...? Really just a remnant of the old way of doing things. wdyt?

nucypher/policy/conditions/_utils.py
```diff
@@ -66,3 +76,51 @@ def _deserialize_condition_lingo(data: Union[str, Dict[str, str]]) -> Union['Ope
     lingo_class = _resolve_condition_lingo(json_data=data)
     instance = lingo_class.from_dict(data)
     return instance
+
+
+def evaluate_conditions_for_ursula(lingo, context, log, ursula):
```
derekpierre, line 81: Is the `ursula` parameter needed here?

nucypher/network/server.py
```diff
@@ -156,30 +151,40 @@ def node_metadata_exchange():
...
+
+        # requester-supplied reencryption condition context
+        context = json.loads(str(reenc_request.context)) or dict()
+
+        # zip capsules with their respective conditions
         packets = zip(reenc_request.capsules, lingo)
 
+        # Populate default request context for decentralized nodes
+        if not this_node.federated_only:
+            context.update({'providers': this_node.condition_providers})
```
derekpierre, line 172: Any reason we shouldn't just pass `condition_providers` directly to `evaluate_conditions_for_ursula(lingo, context, log, ursula)` and subsequently `lingo.eval(...)` method instead of piggybacking ontop of `context`? Perhaps this was the original reason for the unused `ursula` parameter in `evaluate_conditions_for_ursula`?

I'm cautious of modifying a `context` object that was obtained directly via a user request. I like the idea of passing `condition_providers` and `context` to the `eval(...)` method (similar to `verify(...)`); feels like a good separation.

nucypher/policy/conditions/evm.py
```diff
@@ -117,42 +113,38 @@ class RPCExecutionFailed(ReencryptionCondition.ConditionEvaluationFailed):
 
     class Schema(CamelCaseSchema):
         name = fields.Str()
-        chain = fields.Str()
-        method = fields.Str()
+        chain = fields.Int(required=True)
```
derekpierre, line 116: 🔥 - these definitely need to be required.

nucypher/policy/conditions/evm.py
```diff
@@ -164,13 +156,24 @@ def validate_method(self, method):
...
+
+        # This next block validates that the actual web3 provider is *actually*
+        # connected to the condition's chain ID by reading its RPC endpoint.
         provider_chain = self.w3.eth.chain_id
-        if provider_chain != self.chain_id:
+        if provider_chain != self.chain:
             raise ReencryptionCondition.InvalidCondition(
-                f"This condition can only be evaluated on {self.chain_id} but the provider's "
+                f"This condition can only be evaluated on {self.chain} but the provider's "
                 f"connection is to chain {provider_chain}"
```
derekpierre, line 177:
```suggestion
                f"This condition can only be evaluated on chain ID {self.chain} but the provider's "
                f"connection is to chain ID {provider_chain}"
```

nucypher/policy/conditions/evm.py
KPrasch: I agree, It's mostly just a debugging step that I decided to leave since the highly-specific error message might be helpful if a string is passed.

nucypher/policy/conditions/_utils.py
KPrasch: It's unused however, I was considering including some ursula-specific metadata (checksum, connected chains, etc.) in log messages.  Do you think that adds any debugging value?

nucypher/network/server.py
KPrasch: I agree.  Removed this usage of context (Ursula no longer mutates user context).  In retrospect I was looking for the most elegant way support federated mode and settled on this imperfect solution.  Instead I added ternary operation to handle providers:

```python
providers = this_node.condition_provider if not this_node.federated_only else dict()
```

In the highly unlikely situation where an onchain condition is submitted to evaluation by an ursula, the empty dictionary here will fire this `KeyError` in `nucypher/policy/conditions/evm.py`:

```python
def _configure_provider(self, providers: Dict[int, BaseProvider]):
    """Binds the condition's contract function to a blockchian provider for evaluation"""
    try:
        provider = providers[self.chain]
    except KeyError:
        raise Exception(f'This node does not have a connection to chain {self.chain}')

```

nucypher/policy/conditions/_utils.py
derekpierre: Interesting thought. I don't know if they do right now... 

If you think not right now as well we can rename this function to just be `evaluate_conditions`. Dealer's choice.

nucypher/network/server.py
```diff
@@ -208,65 +213,17 @@ def reencrypt():
             return Response(message, status=HTTPStatus.BAD_REQUEST)
 
         # Enforce Reencryption Conditions
-        # TODO: back compatibility for PRE?
-
-        if not this_node.federated_only:
-            # TODO: Detect whether or not a provider is required by introspecting the condition instead.
-            context.update({'provider': this_node.application_agent.blockchain.provider})
-
+        providers = this_node.condition_providers if not this_node.federated_only else dict()
```
derekpierre, line 216: 👍

derekpierre (approved):
🎸 - RFCs can be addressed in a separate PR.

PR #2987, opened 2022-10-25
Assorted deprecations (RPC and Web servers, Enrico CLI, Literature, Specifications, Ursula Interactivity)

KPrasch:
**Type of PR:**
Removal 

**Required reviews:** 
2

**Why**
- Resolves technical debt through removal of unused and non-essential code entities clearing the way for near-term future product development. 
- Eases potential integration with `nucypher-async`

**What this does:**
- Removes all RPC server functionality
- Removes character WebControllers
- Removes unused literature
- Removes unused CLI option definitions
- Removes supporting CLI functionality to follow-up PR #2985
- Removes interactive Ursula mode
- Removes Enrico CLI commands
- Removes unused character banners

**Notes**
Porter is intended to remain fully-functional and is now the only implementer of `WebController` and `WebEmitter`.  I recommend relocating porter to a stand alone repository.

PR #2988, opened 2022-10-26
Removes lmdb and datastore modules

KPrasch:
**Type of PR:**
Removal

**Required reviews:** 
1

**Notes**
- Based on #2987 
- May require a configuration update script to remove the `db_filepath` field from node operator's configurations.
- Postponing a full dependency relock to isolate unrelated dependency test failures

**Why**
Resolves technical debt through removal of unused and non-essential code entities clearing the way for near-term future product development.


Pipfile
```diff
@@ -11,7 +11,6 @@ python_version = "3"
 constant-sorrow = ">=0.1.0a9"
 bytestring-splitter = ">=2.4.0"
 hendrix = ">=3.4"
-lmdb = "*"
```
derekpierre, line 14: Should we relock dependencies? Or are we waiting until the code massacre is over 😄 , and then do a fresh relock?

derekpierre (approved):
🎸 Glad that we are finally getting rid of this.

Pipfile
KPrasch: Yep - I'd like to perform a full relock in an isolated PR to avoid unrelated issues. 😅

derekpierre: 👌 Makes sense.

PR #2989, opened 2022-10-26
Removes Trezor and Clef Signer Support

KPrasch:
**Type of PR:**
Removal

**Required reviews:** 
1

**What this does:**
- Removes ClefSigner
- Removes TrezorSigner

**Notes**
- Based on #2988 
- postponing a dependency relock to isolate unrelated dependency test failures

**Why**
- Resolves technical debt by removing code pathways that have been subsumed by nucypher/nucypher-ts.  Although the Alice, Bob, and Enrico python APIs are still supported (using software wallets), the recommended and most common implementation is browser-based.  This implies that wallet signing operations will be handed by browser-based wallets (like metamask), so we can drop support for CLI & HW based grant-and-retrieve user flows.


derekpierre (approved):
🎸 - You are on fire 🔥

PR #2991, opened 2022-10-27
Remove clouddeploy.py

KPrasch:
**Type of PR:**
Removal

**Required reviews:** 
1

**What this does:**
- Removes `clouddeploy.py`

**Why it's needed:**
This functionality has grown into its own project nucypher/nucypher-ops


arjunhassard (approved):
End of an era. This kind of automation has just as much impact on the number of independent stakers as the reward mechanism – we should reuse a lot of this thinking for new apps on Threshold

manumonti (approved):
Let's 🪓

PR #3002, opened 2022-11-07
Add custom abi conditions and key/value return values

### LINKED ISSUE #2984: Support Return Value Tests for Structs
Suggestions from @theref 

Presently, conditions can only be evaluated against contract functions that return integers and strings.  Conditions can be more flexible by also evaluating against functions that return structs, and introspecting a user-supplied key.

### LINKED ISSUE #3003: Support custom ABI conditions

theref:
**Type of PR:**
- [x] Feature

**Required reviews:** 
- [X] 2

**What this does:**
- Allow the user to specify a custom abi for EVM conditions
- Allow the user to specify a `key` to be used when evaluating the return conditions

**Issues fixed/closed:**
- Fixes #2984
- Fixes #3003


nucypher/policy/conditions/lingo.py
```diff
@@ -83,12 +83,13 @@ class InvalidExpression(ValueError):
...
         comparator = fields.Str()
         value = fields.Raw(allow_none=False)  # any valid type (excludes None)
+        key = fields.Raw(allow_none=True)
 
         @post_load
         def make(self, data, **kwargs):
             return ReturnValueTest(**data)
 
-    def __init__(self, comparator: str, value):
+    def __init__(self, comparator: str, value, key: str = ""):
```
KPrasch, line 92:
```suggestion
    def __init__(self, comparator: str, value, key: Optional[Union[int, str]] = None):
```

nucypher/policy/conditions/lingo.py
```diff
@@ -116,6 +118,19 @@ def eval(self, data) -> bool:
...
+                f"'{self.value}' is an unprocessed context variable and is not valid "
+                f"for condition evaluation."
+            )
+        if self.key:
+            try:
+                data = data[self.key]
+            except KeyError:
+                raise self.InvalidExpression(
+                    f"Key '{self.key}' not found in return data."
+                )
```
KPrasch, line 133: 👍🏻

nucypher/policy/conditions/lingo.py
```diff
@@ -116,6 +118,19 @@ def eval(self, data) -> bool:
                 f"'{self.value}' is an unprocessed context variable and is not valid "
                 f"for condition evaluation."
             )
+        if is_context_variable(self.key):
+            # programming error if we get here
+            raise RuntimeError(
+                f"'{self.value}' is an unprocessed context variable and is not valid "
```
KPrasch, line 124: `self.value` -> `self.key` (I know this PR is still WIP, and still needs preprocessing of key's context variables)

nucypher/policy/conditions/lingo.py
```diff
@@ -83,12 +83,13 @@ class InvalidExpression(ValueError):
...
         comparator = fields.Str()
         value = fields.Raw(allow_none=False)  # any valid type (excludes None)
+        key = fields.Raw(allow_none=True)
 
         @post_load
         def make(self, data, **kwargs):
             return ReturnValueTest(**data)
 
-    def __init__(self, comparator: str, value):
+    def __init__(self, comparator: str, value, key: Union[int, str] = None):
```
KPrasch, line 92: `Optional` type hint here?

nucypher/policy/conditions/lingo.py
theref: eugh, stupid copy/paste

I think the preprocessing will be the same as for `value` - I couldn't come up with any other scenarios

KPrasch (commented):
Looking nice so far.  Custom ABI conditions are already supported, unless you have something more id mind?

nucypher/policy/conditions/lingo.py
theref: i got linting errors....hold on...

theref: error occurred between the chair and the keyboard

theref:
> Looking nice so far. Custom ABI conditions are already supported, unless you have something more in mind?

I added:
```json
"TStaking": {
    "contractAddress": "0x01B67b1194C75264d06F808A921228a95C765dd7",
    "chain": "ethereum",
    "method": "stakes",
    "functionAbi": {
      "inputs": [
        {
          "internalType": "address",
          "name": "stakingProvider",
          "type": "address"
        }
      ],
      "name": "stakes",
      "outputs": [
        {
          "internalType": "uint96",
          "name": "tStake",
          "type": "uint96"
        },
        {
          "internalType": "uint96",
          "name": "keepInTStake",
          "type": "uint96"
        },
        {
          "internalType": "uint96",
          "name": "nuInTStake",
          "type": "uint96"
        }
      ],
      "stateMutability": "view",
      "type": "function"
    },
    "parameters": [
      ":userAddress"
    ],
    "returnValueTest": {
      "key": "tStake",
      "comparator": ">",
      "value": 0
    }
  },
```
to `test_conditions.json` and got error: `marshmallow.exceptions.ValidationError: {'functionAbi': ['Not a valid string.']}`

Have I misunderstood how the `functionAbi` should be defined?

KPrasch (commented):
it's not exactly obvious, but the value of `function_abi` must be a list.  Here's an example: https://github.com/nucypher/nucypher/pull/2986/commits/812d0a1b62c835b7d67787a1a2acbb7b1e7d805f#diff-c97439f8b650945d4bd8cbd051de8ceb852934885c1aae284f2cbda6f416aab2R182

nucypher/policy/conditions/evm.py
```diff
@@ -95,9 +95,14 @@ def _resolve_any_context_variables(
         processed_parameters.append(p)
 
     v = return_value_test.value
+    k = return_value_test.key
```
derekpierre, line 98: What's the scenario where the key for the test would be a context variable?

Presumably the `key` is determined at condition creation based on the specific ABI, and not at condition evaluation? But, maybe I'm missing something.

KPrasch (commented):
I don't think any of the current JSON test data implements custom ABIs, that's an excellent addition you've got there.

nucypher/policy/conditions/lingo.py
```diff
@@ -83,12 +83,13 @@ class InvalidExpression(ValueError):
...
         comparator = fields.Str()
         value = fields.Raw(allow_none=False)  # any valid type (excludes None)
+        key = fields.Raw(allow_none=True)
 
         @post_load
         def make(self, data, **kwargs):
             return ReturnValueTest(**data)
 
-    def __init__(self, comparator: str, value):
+    def __init__(self, comparator: str, value, key: Optional[Union[int, str]] = None):
```
derekpierre, line 92: Key can be a `string` or `int`...does `key` being an `int` represent an index for the result we care about eg. index from values in a tuple?

nucypher/policy/conditions/lingo.py
```diff
@@ -116,6 +118,19 @@ def eval(self, data) -> bool:
...
             )
+        if is_context_variable(self.key):
+            # programming error if we get here
+            raise RuntimeError(
+                f"'{self.key}' is an unprocessed context variable and is not valid "
+                f"for condition evaluation."
+            )
+        if self.key:
+            try:
+                data = data[self.key]
```
derekpierre, line 129: Is `key` always allowed to be an `int` or `string` here i.e. if you are expecting a struct/tuple to be returned I'm not sure that `string` will work here. Perhaps I'm missing the ideal scenario.

I could be wrong, but try adding some acceptance tests. You can test using `getPolicy` tests on the SubscriptionManager contract which returns a Policy struct. See https://github.com/nucypher/nucypher/blob/development/tests/acceptance/blockchain/conditions/test_conditions.py#L316.

derekpierre (commented):
Question for you @theref . Take the `stakes` function on TokenStaking contract. which returns `(tStake, keepInTStake, nuInTStake)` tuple. Is the goal to be able to say `key=1` to check the `keepInTStake` value, or would it be key=`keepInTStake` to check the value or both should be valid?

nucypher/policy/conditions/evm.py
theref: if the function returned a struct mapping addresses to something, and we wanted to key by :userAddress

KPrasch: It's plausible that struct can be returned that is keyed by addresses, which might be the user's address.  Whether or not that address needs to be recovered/verified by a digital signature is unclear from an authentication purposes, but it's certainly convenient UX opposed to having the user paste their own address into the condition.

theref:
> Question for you @theref . Take the `stakes` function on TokenStaking contract. which returns `(tStake, keepInTStake, nuInTStake)` tuple. Is the goal to be able to say `key=1` to check the `keepInTStake` value, or would it be key=`keepInTStake` to check the value or both should be valid?

Good question, I hadn't considered keying by index, so my gut would be just `keepInTStake`, but I'm open to suggestions here

KPrasch:
Hmm, it's possible to use `key` to represent looksups in mappings as well as arrays, is that acceptable?  Alternatively we can introduce some strict type-checking to support `key` and also `index` in return value tests.

nucypher/policy/conditions/lingo.py
KPrasch: In this function it can be agnostic - mapping of ints or an array are both compatible (additional commentary below).

nucypher/policy/conditions/evm.py
derekpierre: > Whether or not that address needs to be recovered/verified by a digital signature is unclear from an authentication purposes

^ With `:userAddress` it would need to be authenticated - that's the requirement of the processing directive for that context variable; the case for *needing* authentication for this case is not clear to me

KPrasch: Consider that it might just be a matter of convenience since there is not an *unauthenticated* context variable for user address.

theref: i can't think of a case where the users address would need to be used without it being authenticated somehow.
Let's say i had a whitelisting contract with an attribute `whitelist` that mapped addresses to booleans.

I would want my return value test to look something like:
```
    "returnValueTest": {
      "key": :userAddress,
      "comparator": "==",
      "value": true
    }
```

I **need** the `:userAddress` to be authenticated here

KPrasch:
> This returns a dict which you're then placing inside the list. 

Exactly, the serialized form needs to be a list at the top-level for web3.py consumption.

> My issue is that marshmallow doesn't want to read in the functionAbi

Ah, I see! Sounds like a deserialization issue. You'll need to do something different in the schema (using fields.Str is too simple here because it does not represent the inner types of the ABI):

https://github.com/nucypher/nucypher/blob/development/nucypher/policy/conditions/evm.py#L212

nucypher/policy/conditions/evm.py
derekpierre: This is where my confusion lies 😅 , but perhaps there are other examples out there.

Perhaps it is possible that someone could write a contract that way, but seems unlikely that they would. To me, the case for whitelisting is a contract function that takes the `:userAddress` as a param and returns true/false i.e. `function isWhitelisted(address user) returns(bool)`; and it is as a param that the address would be authenticated, not as a key on the return value. Otherwise, the entire dictionary of who is whitelisted is returned, which:
1. could get long
2. reveals/leaks other addresses that are whitelisted

With the `isWhitelisted` function, the ReturnValueTest just checks for `true`/`false` and does not need the key. 

Basically what I'm wondering is, in any of the cases where some mapping keyed by address could be returned, would it be common to also have functions in the same contract that just take an address and return information specific to the address, since returning the entire internal mapping is not efficient.

Maybe @vzotova may have some thoughts here given her expertise in contracts, and common practices.

KPrasch: I'm not sure we need to uncover every possible example.  Another way to approach this design decision: Do we want every possible input and output value to support `:userAddress` for maximum flexibility, or is less control still sufficient to cover practical use cases?

derekpierre: > or is less control still sufficient to cover practical use cases?
That's along the lines I'm thinking.  The base case for a `key` to be specified is a great one. My thought here is to start simple (actual value), which could be sufficient to cover most practical use cases - and then iterate if needed. 

That being said, I accept if others feel differently.

nucypher/policy/conditions/evm.py
```diff
@@ -62,7 +50,8 @@ def _resolve_abi(standard_contract_type: str, method: str, function_abi: List) -
 
     if standard_contract_type:
         try:
-            function_abi = STANDARD_ABIS[standard_contract_type]
+            contract_abi = STANDARD_ABIS[standard_contract_type]
+            function_abi = [x for x in contract_abi if x.get("name") == method][0]
```
derekpierre, line 54: What happens if an invalid method name is provided?

nucypher/policy/conditions/lingo.py
```diff
@@ -116,6 +120,24 @@ def eval(self, data) -> bool:
...
+                f"'{self.key}' is an unprocessed context variable and is not valid "
+                f"for condition evaluation."
+            )
+        if self.key and type(data) is dict:
+            try:
+                data = data[self.key]
+            except KeyError:
+                raise KeyError(f"Key '{self.key}' not found in return data.")
+
+        if type(self.key) is int and type(data) is list:
```
derekpierre, line 135: What about tuples? I believe structs get returned as tuples

nucypher/policy/conditions/lingo.py
```diff
@@ -116,6 +120,24 @@ def eval(self, data) -> bool:
                 f"'{self.value}' is an unprocessed context variable and is not valid "
                 f"for condition evaluation."
             )
+        if is_context_variable(self.key):
+            # programming error if we get here
+            raise RuntimeError(
+                f"'{self.key}' is an unprocessed context variable and is not valid "
+                f"for condition evaluation."
+            )
+        if self.key and type(data) is dict:
```
derekpierre, line 129: To keep things clean, maybe pull the processing of the key into a helper method.

tests/unit/conditions/test_evm_condition_serializers.py
```diff
@@ -22,14 +22,24 @@
...
+
 def test_evm_condition_json_serializers(ERC1155_balance_condition_data):
     original_data = ERC1155_balance_condition_data
     condition = ContractCondition.from_json(original_data)
     serialized_data = condition.to_json()
 
-    # TODO functionAbi is present in serialized data
     deserialized_data = json.loads(serialized_data)
     deserialized_data.pop("functionAbi")
+    deserialized_data["returnValueTest"].pop("key")
```
derekpierre, line 41: Why is this needed?

The original reason for L40, is that the ContractCondition automatically derives the functionABI from the standard contract type. However `key` isn't required and isn't specified in the original condition's return value test, so why is it part of the serialized data?

tests/unit/conditions/test_return_value.py
```diff
@@ -23,6 +23,34 @@
 from nucypher.policy.conditions.lingo import ReturnValueTest
 
 
+def test_return_value_key():
```
derekpierre, line 26: If we are supporting context variables as keys, you could also add a test to `test_return_value_test_with_context_variable_cant_run_eval` to ensure that of a context variable for a key it fails if not converted before use.

derekpierre (dismissed):
We need to add some conditions acceptance tests to properly test the use of "key" with Return Value Tests, and the context variable replacement functionality.

You can take a look at the SubscriptionManager acceptance tests (at least with respect to handling structs). If you need some guidance for those let me know.

tests/unit/conditions/test_evm_condition_serializers.py
derekpierre: I think if `key` is optional, and doesn't have a value (or is `None`), it shouldn't be serialized.

We should also fix the serialization of `functionAbi` while we are at it. That was a stop gap for the demo. If the standard contract type and method are provided, there is no need to serialize the `functionAbi` property either.

@KPrasch wdyt?

nucypher/policy/conditions/evm.py
```diff
@@ -62,11 +54,16 @@ def _resolve_abi(standard_contract_type: str, method: str, function_abi: List) -
...
+            contract_abi = STANDARD_ABIS[standard_contract_type]
         except KeyError:
             raise ReencryptionCondition.InvalidCondition(
                 f"Invalid standard contract type {standard_contract_type}; Must be one of {STANDARD_ABI_CONTRACT_TYPES}"
             )
+        if method not in STANDARD_ABI_CONTRACT_METHODS[standard_contract_type]:
+            raise ReencryptionCondition.InvalidCondition(
+                f"Invalid method {method} for standard contract type {standard_contract_type}; Must be one of {STANDARD_ABI_CONTRACT_METHODS[standard_contract_type]}"
+            )
+        function_abi = [x for x in contract_abi if x.get("name") == method][0]
```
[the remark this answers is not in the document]
derekpierre: > I've just realized that it's possible to get multiple matches for a given function name.

I had the same thought. Technically can we pass the abis for all of the methods? Previously, this did allow the passing of the entire ABI for the contract (remember the subscription manager test)? Or can we use the params to determine which overridden method to use?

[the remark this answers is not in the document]
KPrasch: > Technically can we pass the abis for all of the methods?

Actually yes. Subsequently web3.py will resolve the correct ABI given the input paramters.  That's probably the best way to handle it.

```python
functions = Web3().eth.contract(abi=contract_abi).find_functions_by_name(method)
function_abis = [func.abi for func in functions]
```

nucypher/policy/conditions/evm.py
```diff
@@ -51,11 +50,22 @@ def _resolve_abi(standard_contract_type: str, method: str, function_abi: List) -
...
+        functions = Web3().eth.contract(abi=contract_abi).find_functions_by_name(method)
+        if not functions:
+            raise ReencryptionCondition.InvalidCondition(
+                f"Invalid method {method} for standard contract type {standard_contract_type}"
+            )
+        if len(functions) > 1:
+            raise ReencryptionCondition.InvalidCondition(
+                f"Multiple functions with name {method} for standard contract type {standard_contract_type}"
+            )
+        function_abi = [func.abi for func in functions][0]
```
KPrasch, line 67: I opened a PR (#1) into your fork <https://github.com/theref/nucypher/pull/1> that simplifies this function a bit more and adds some strict type annotations 🙏🏻

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +105,46 @@ def __init__(self, comparator: str, value: Any):
...
         self.value = value
+        self.key = key
 
     def _sanitize_value(self, value):
         try:
             return ast.literal_eval(str(value))
         except Exception:
             raise self.InvalidExpression(f'"{value}" is not a permitted value.')
 
+    def _process_data(self, data: Any) -> Any:
```
KPrasch, line 116: A descriptive function docstring will be useful here.

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +105,46 @@ def __init__(self, comparator: str, value: Any):
...
+        if self.key and isinstance(data, dict):
+            try:
+                processed_data = data[self.key]
+            except KeyError:
+                raise KeyError(f"Key '{self.key}' not found in return data.")
+        elif isinstance(self.key, int) and isinstance(data, (list, tuple)):
+            try:
+                processed_data = data[self.key]
+            except IndexError:
+                raise IndexError(f"Index '{self.key}' not found in return data.")
```
KPrasch, line 126: This section is looking excellent. ⚡

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +105,46 @@ def __init__(self, comparator: str, value: Any):
 
         self.comparator = comparator
         self.value = value
+        self.key = key
```
KPrasch, line 108: This is a good area to prevent an invalid `ReturnValue` state by enforcing a valid type of `key`.

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +105,46 @@ def __init__(self, comparator: str, value: Any):
...
                 f"'{self.value}' is an unprocessed context variable and is not valid "
                 f"for condition evaluation."
             )
-        left_operand = self._sanitize_value(data)
+        if is_context_variable(self.key):
+            # programming error if we get here
+            raise RuntimeError(
+                f"'{self.key}' is an unprocessed context variable and is not valid "
+                f"for condition evaluation."
+            )
```
KPrasch, line 144: Even though the error message will be less descriptive, it's possible to combine these two checks into check with a more generic message.  Considering this is a "Programming Error" that is not intended as part of normal flow-control I think it's worth the trade-off.

tests/acceptance/blockchain/conditions/conftest.py
```diff
@@ -181,7 +185,9 @@ def subscription_manager_is_active_policy_condition(test_registry, agency):
     )
     condition = ContractCondition(
         contract_address=subscription_manager.contract.address,
-        function_abi=[subscription_manager.contract.find_functions_by_name('isPolicyActive')[0].abi],
+        function_abi=subscription_manager.contract.find_functions_by_name(
+            "isPolicyActive"
+        )[0].abi,
```
KPrasch, line 190: Can be simplified by using `get_function_by_name` (see PR https://github.com/theref/nucypher/pull/1).

tests/acceptance/blockchain/conditions/conftest.py
```diff
@@ -199,7 +205,9 @@ def subscription_manager_get_policy_zeroized_policy_struct_condition(
     )
     condition = ContractCondition(
         contract_address=subscription_manager.contract.address,
-        function_abi=[subscription_manager.contract.find_functions_by_name('getPolicy')[0].abi],
+        function_abi=subscription_manager.contract.find_functions_by_name("getPolicy")[
+            0
```
KPrasch, line 209: Sometimes blackened code format is less readable... 🤔

KPrasch: Same as above, can be simplified by using `get_function_by_name` (see PR https://github.com/theref/nucypher/pull/1).

tests/data/test_conditions.json
```diff
@@ -1,4 +1,46 @@
...
+    },
+    "parameters": [
+      ":userAddress"
+    ],
+    "returnValueTest": {
+      "key": "tStake",
+      "comparator": ">",
+      "value": 0
+    }
+  },
```
KPrasch, line 43: Nice to see a custom ABI test vector here 👍🏻

tests/unit/conditions/test_condition_validity.py
```diff
@@ -89,12 +89,25 @@ def test_invalid_contract_condition():
...
-                function_abi=["rando ABI"],
-                return_value_test=ReturnValueTest('!=', 0),
-                parameters=[
-                    ':hrac',
-                ]
-            )
+            contract_address="0xaDD9D957170dF6F33982001E4c22eCCdd5539118",
+            method="getPolicy",
+            chain=TESTERCHAIN_CHAIN_ID,
+            function_abi=["rando ABI"],
```
KPrasch, line 95: (included in https://github.com/theref/nucypher/pull/1)

```suggestion
            function_abi={"rando": "ABI"},
```

tests/unit/conditions/test_evm_condition_serializers.py
KPrasch: Marshmallow schemas can be configured to exclude fields that have a `None` value during de/serialization.  This is currently implemented in `ContractCondition`:

```python
class ContractCondition(RPCCondition):
    class Schema(RPCCondition.Schema):
        SKIP_VALUES = (None,)
```

The same can be done for `key` in `ReturnValueTest.ReturnValueTestSchema` which will avoid the situation seen in this test with `.pop`:

```python

class ReturnValueTestSchema(CamelCaseSchema):
    SKIP_VALUES = (None,)
    ...
    key = fields.Raw(allow_none=True)
```

tests/unit/conditions/test_return_value.py
```diff
@@ -23,6 +23,47 @@
...
+    test = ReturnValueTest(comparator=">", value="0", key=0)
+    assert test.eval((1,))
+    assert not test.eval((-1,))
+
+
+def test_return_value_with_context_variable_key_cant_run_eval():
+    # known context variable
+    test = ReturnValueTest(comparator="==", value="0", key=":userAddress")
+    with pytest.raises(RuntimeError):
+        test.eval({"0xaDD9D957170dF6F33982001E4c22eCCdd5539118": 0})
```
KPrasch, line 64: 🔥

tests/unit/conditions/test_return_value.py
```diff
@@ -23,6 +23,47 @@
...
+        test.eval({5: 1})
+
+
+def test_return_value_index():
+    test = ReturnValueTest(comparator=">", value="0", key=0)
+    assert test.eval([1])
+    assert not test.eval([-1])
+
+    test = ReturnValueTest(comparator="==", value='"james"', key=3)
+    assert test.eval([0, 1, 2, '"james"'])
```
KPrasch, line 48: Why use triple quotation here?

KPrasch:
Rebased over `development` @ 806f08b1bec77298405b2c0b3b8baec43d9f5cb1

tests/unit/conditions/test_return_value.py
theref: it's not triple, it's a single outside of a double, similar to how the other tests are done

tests/unit/conditions/test_return_value.py
```diff
@@ -23,6 +23,47 @@
...
+    assert not test.eval({4: -1})
+
+    with pytest.raises(KeyError):
+        test.eval({5: 1})
+
+
+def test_return_value_index():
+    test = ReturnValueTest(comparator=">", value="0", key=0)
+    assert test.eval([1])
+    assert not test.eval([-1])
```
KPrasch, line 45: Interesting situation here - in python a `TypeError` is raised when using operators (like `>` and `<`) of different types but it's not obvious where/how we resolve such cases with `ReturnValueTest` 🤔

tests/unit/conditions/test_return_value.py
```diff
@@ -23,6 +23,47 @@
...
+    assert test.eval([0, 1, 2, '"james"'])
+
+    with pytest.raises(IndexError):
+        test.eval([0, 1, 2])
+
+
+def test_return_value_index_tuple():
+    test = ReturnValueTest(comparator=">", value="0", key=0)
+    assert test.eval((1,))
+    assert not test.eval((-1,))
```
KPrasch, line 57: This implies that `0` is equal to `"0"` - is this intentional, or just a side-effect of python's `eval`?  I'm aware that this behavior predates this PR, but it's illuminated here.

theref: I believe this is to do with `eval` and is related to why `'"james"'` is needed in the above comment

KPrasch: Is an exception raised when using single quotes (perhaps a `NameError` or `TypeError`)?

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +106,44 @@ def __init__(self, comparator: str, value: Any):
...
+        if self.key and isinstance(data, dict):
+            try:
+                processed_data = data[self.key]
+            except KeyError:
+                raise KeyError(f"Key '{self.key}' not found in return data.")
+        elif isinstance(self.key, int) and isinstance(data, (list, tuple)):
+            try:
+                processed_data = data[self.key]
+            except IndexError:
+                raise IndexError(f"Index '{self.key}' not found in return data.")
```
derekpierre, line 131: Should this instead be something like:
```python
if self.key:
    if isinstance(data, dict):
          ...
     elif instance(self.key, int) and isinstance(data, (list, tuple)):
          ...
     else:
           ... <some exception about incompatible data>
else:
    processed_data = data
```

Feels like we are missing the exception case for incompatible data i.e. the data is not a dict, list, or tuple.

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +106,44 @@ def __init__(self, comparator: str, value: Any):
...
+        return processed_data
+
     def eval(self, data) -> bool:
-        if is_context_variable(self.value):
+        if is_context_variable(self.value) or is_context_variable(self.key):
             # programming error if we get here
             raise RuntimeError(
-                f"'{self.value}' is an unprocessed context variable and is not valid "
+                f"Either '{self.value}' or '{self.key}' is an unprocessed context variable and is not valid "
                 f"for condition evaluation."
```
derekpierre, line 142: What about something more concrete like:
```suggestion
                f"Return value comparator contains an unprocessed context variable (key={self.key}, value={self.value}) and is not valid "
                f"for condition evaluation."
```

tests/acceptance/blockchain/conditions/test_conditions.py
```diff
@@ -333,11 +334,7 @@ def test_subscription_manager_get_policy_policy_struct_condition_evaluation(
 
     # zeroized policy struct
     zeroized_policy_struct = (
-        "0x0000000000000000000000000000000000000000",
-        0,
-        0,
-        0,
-        "0x0000000000000000000000000000000000000000",
+        NULL_ADDRESS, 0, 0, 0, NULL_ADDRESS,
```
derekpierre, line 337: 👍

tests/unit/conditions/test_return_value.py
```diff
@@ -23,6 +23,47 @@
...
+
+
+def test_return_value_with_context_variable_key_cant_run_eval():
```
derekpierre, line 60: Nice - 👍 . Can we also add a test that includes actual context variable replacement for key?

See context variable replacemet tests in `tests/acceptance/blockchain/conditions/test_conditions.py`

derekpierre: We should add tests for when a data type that is not list, tuple, or dict, to ensure the code handles it appropriately.

derekpierre: > if key is supplied, and the value is not in list, tuple, or dict, then the value is just returned as-is

Feels like this case should fail. Key is optional so if supplied, it should presumably be adhered to...? Usage of key to get value from other data types that don't allow it (basically invalid use of key given the data type returned) eg. the value returned for comparison testing is an `int`, and can't use the `key` on it - this should presumably also be an exception

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +106,44 @@ def __init__(self, comparator: str, value: Any):
...
+    def _process_data(self, data: Any) -> Any:
+        """
+        If a key is specified, return the value at that key in the data if data is a dict or list-like.
+        Otherwise, return the data.
+        """
+        if self.key and isinstance(data, dict):
+            try:
+                processed_data = data[self.key]
+            except KeyError:
+                raise KeyError(f"Key '{self.key}' not found in return data.")
```
derekpierre, line 126: Since this method is executed at condition evaluation time, I think this should raise an exception caught via `_utils.evaluate_condition(...)`, eg. something like `InvalidCondition`. Same for the `IndexError` on L131.

tests/unit/conditions/test_return_value.py
theref: Not sure i understand the problem?

tests/unit/conditions/test_return_value.py
theref: The line that requires this is `lingo.py:116` if you run that line on a regular string you get:
```
$python
>>> v = "james"
>>> import ast
>>> ast.literal_eval(str(v))
```

theref: There is even a pre-existing test for it:
```python
def test_return_value_test_string():
    # string values must be quoted
    with pytest.raises(ReturnValueTest.InvalidExpression):
        _test = ReturnValueTest(comparator="==", value="foo")
```

KPrasch: Got it. I'm glad we reviewed it since it's not obvious why its needed.

tests/unit/conditions/test_return_value.py
KPrasch: I'm not sure if there's a problem, but I'm curios what the expected behavior is when comparing mismatched types (`1 > "2" for example.`.  In Python this is a `TypeError`.  Is the same true via `ast.literal_eval`?

nucypher/policy/conditions/lingo.py
KPrasch: Bumping this one - perhaps we can type-check `key` in case it's an invalid type here to promote a fast failure:

```python
if not any(isinstance(self.k, t) for t in (str, int)): raise
```

nucypher/policy/conditions/lingo.py
KPrasch: I'd also like to see a higher-order exception raised here to follow the emerging convention.  

> The condition isn't necessarily invalid, they key just isn't present.

Interesting take - IMO if the return value of an on-chain funtion returns a different type than what is expected by the condition, the condition is invalid, malformed, or at least incorrect.  Perhaps you can introduce a new type if exception class here: `nucypher.policy.conditions.exceptions.ReturnValueTypeError` or similar.  This can potentially also encompass other type-related problems mentioned in my prior comments.

derekpierre: Don't get me wrong, I wasn't forcing an exception on you, my point was that we shouldn't raise a `KeyError`. If you raise a `KeyError` here, the `lingo.eval()` raises a `KeyError` which then probably gets turned into an HTTP 500 generalized internal error? (I think it does). If so, this is not really helpful.

The question becomes is a key not found in the data, 1) a condition failure or 2) an invalid condition. In your case you are considering it a failure, for which we can do better than just a generalized 500. It can be a separate exception (in exception.py) but I don't think it should be a generic `KeyError`, but rather something more specific and therefore informative. Same for `IndexError`. This exception would then be caught and handled appropriately (message and status code) in `evaluate_conditions(...)` in `_utils.py`.

nucypher/policy/conditions/lingo.py
derekpierre: 👍 Nice - Good to catch this early

nucypher/policy/conditions/lingo.py
```diff
@@ -98,6 +99,11 @@ def __init__(self, comparator: str, value: Any, key: Optional[Union[int, str]] =
                 f'"{comparator}" is not a permitted comparator.'
             )
 
+        if not isinstance(key, (int, str)) and key is not None:
+            raise self.InvalidExpression(
+                f'"{key}" is not a permitted key. Must be a string or integer.'
+            )
```
KPrasch, line 105: perfect

nucypher/policy/conditions/lingo.py
```diff
@@ -123,12 +129,16 @@ def _process_data(self, data: Any) -> Any:
...
+                    f"Key '{self.key}' not found in return data."
+                )
         elif isinstance(self.key, int) and isinstance(data, (list, tuple)):
             try:
                 processed_data = data[self.key]
             except IndexError:
-                raise IndexError(f"Index '{self.key}' not found in return data.")
+                raise ReturnValueEvaluationError(
+                    f"Index '{self.key}' not found in return data."
+                )
```
KPrasch, line 141: 👍🏻

nucypher/policy/conditions/_utils.py
```diff
@@ -129,6 +130,11 @@ def evaluate_conditions(
         if not result:
             # explicit condition failure
             error = ("Decryption conditions not satisfied", HTTPStatus.FORBIDDEN)
+    except ReturnValueEvaluationError as e:
+        error = (
+            f"Unable to evaluate return value: {e}",
+            HTTPStatus.INTERNAL_SERVER_ERROR,
```
KPrasch, line 136: 500s are server errors, is it really a problem with the server?  At least to my sensibilities 500s are critical issues with the server that are outside the control of the client (syntax errors, unhanded exceptions, etc.).  In this case, it's really a matter of input to the request so perhaps it's better to use a 400-series response code?

nucypher/policy/conditions/evm.py
```diff
@@ -45,28 +46,39 @@
...
+        method: str,
+        standard_contract_type: Optional[str] = None,
+        function_abi: Optional[ABIFunction] = None,
+) -> ABIFunction:
     """Resolves the contract an/or function ABI from a standard contract name"""
+
     if not (function_abi or standard_contract_type):
-        # TODO: Is this protection needed?
-        raise InvalidCondition(
+        raise ReencryptionCondition.InvalidCondition(
```
KPrasch, line 58: `InvalidCondition` relocated to `nucypher.policy.conditions.exceptions`

nucypher/policy/conditions/evm.py
```diff
@@ -45,28 +46,39 @@
...
             raise InvalidCondition(
                 f"Invalid standard contract type {standard_contract_type}; Must be one of {STANDARD_ABI_CONTRACT_TYPES}"
             )
 
+        try:
+            # Extract all function ABIs from the contract's ABI.
+            # Will raise a ValueError if there is not exactly one match.
+            function_abi = w3.eth.contract(abi=contract_abi).get_function_by_name(method).abi
+        except ValueError as e:
+            raise ReencryptionCondition.InvalidCondition(str(e))
```
KPrasch, line 76: `InvalidCondition` relocated to `nucypher.policy.conditions.exceptions`

tests/unit/conditions/test_return_value.py
```diff
@@ -23,6 +23,47 @@
 from nucypher.policy.conditions.lingo import ReturnValueTest
 
 
+def test_return_value_key():
+    test = ReturnValueTest(comparator=">", value="0", key="james")
+    assert test.eval({"james": 1})
+    assert not test.eval({"james": -1})
+
+    with pytest.raises(KeyError):
```
derekpierre, line 31: Does this test pass? It shouldn't, right...

tests/unit/conditions/test_return_value.py
```diff
@@ -23,6 +23,47 @@
...
+
+def test_return_value_index():
+    test = ReturnValueTest(comparator=">", value="0", key=0)
+    assert test.eval([1])
+    assert not test.eval([-1])
+
+    test = ReturnValueTest(comparator="==", value='"james"', key=3)
+    assert test.eval([0, 1, 2, '"james"'])
+
+    with pytest.raises(IndexError):
```
derekpierre, line 50: Does this test pass? It shouldn't, right...

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +112,48 @@ def __init__(self, comparator: str, value: Any):
...
+    def _process_data(self, data: Any) -> Any:
+        """
+        If a key is specified, return the value at that key in the data if data is a dict or list-like.
+        Otherwise, return the data.
+        """
+        if self.key and isinstance(data, dict):
+            try:
+                processed_data = data[self.key]
+            except KeyError:
+                raise ReturnValueEvaluationError(
```
derekpierre, line 132: 👌

nucypher/policy/conditions/lingo.py
```diff
@@ -104,21 +112,48 @@ def __init__(self, comparator: str, value: Any):
...
+                )
+        elif isinstance(self.key, int) and isinstance(data, (list, tuple)):
+            try:
+                processed_data = data[self.key]
+            except IndexError:
+                raise ReturnValueEvaluationError(
+                    f"Index '{self.key}' not found in return data."
+                )
+        else:
+            processed_data = data
```
derekpierre, line 143: I think we chatted about this previously - but I don't remember the conclusion.

If `key` is specified, and the data instance is not tuple, list, or dict then then a `ReturnValueEvaluationError` should be raised, and not just continue on with the `data` as is - this case is not currently handled.

If `key` is not specified then obviously we do `processed_data = data`, and should do this. See prior comment here - https://github.com/nucypher/nucypher/pull/3002#discussion_r1019268711

theref: this is really strange, I definitely made this change. I think something may have gotten overwritten when I rebased

derekpierre: Yeah I thought I remembered you did...damn rebases 😬

theref: I very confused, I have 100% already fixed this :D i think i may have overwritten a bunch of stuff when rebasing

eugh

theref: and now my comments aren't adding correctly...

KPrasch (approved):
👏🏻 Excellent work on this PR. I feel this area of the codebase has taken an important step forward in completeness.

nucypher/policy/conditions/context.py
```diff
@@ -108,7 +108,7 @@ def get_context_value(context_variable: str, **context) -> Any:
         # fallback for context variable without directive - assume key,value pair
         # handles the case for user customized context variables
         value = context.get(context_variable)
-        if not value:
+        if value is None:
```
KPrasch, line 111: 💥

PR #3007, opened 2022-11-09
Ursula server update - fix abstraction leaks

derekpierre:
Better handle error cases and condition failure cases.

**Type of PR:**
- [X] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -224,7 +224,7 @@ def _make_rest_app(this_node, log: Logger) -> Flask:
                 if error:
                     # TODO: This response short-circuits the entire request on falsy condition
                     #  even if other unrelated capsules (message kits) are present.
-                    return Response(message=error.message, status=error.status_code)
+                    return Response(error.message, status=error.status_code)
             capsules_to_process.append(capsule)
 
         # FIXME: DISABLED FOR PRE-adapted-TDEC
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -185,7 +185,6 @@ def _make_rest_app(this_node, log: Logger) -> Flask:
 
         # Bob
         bob_ip_address = request.remote_addr
-        bob_verifying_key = bob.stamp.as_umbral_pubkey()
         bob_identity_message = f"[{bob_ip_address}] Bob({bytes(bob.stamp).hex()})"
 
         # Verify & Decrypt KFrag Payload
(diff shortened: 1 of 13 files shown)


nucypher/network/server.py
```diff
@@ -217,17 +217,13 @@ def reencrypt():
         capsules_to_process = list()
         for capsule, lingo in packets:
             # raises an exception or continues
-            result, error = evaluate_conditions_for_ursula(
+            error = evaluate_conditions(
```
KPrasch, line 220: 👍🏻 Good name change here

nucypher/policy/conditions/_utils.py
```diff
@@ -81,57 +99,64 @@ def _deserialize_condition_lingo(data: Union[str, Dict[str, str]]) -> Union['Ope
...
 
     # avoid using a mutable defaults and support federated mode
     context = context or dict()
     providers = providers or dict()
-
+    error = None
     if lingo is not None:
         # TODO: Evaluate all conditions even if one fails and report the result
         try:
             log.info(f'Evaluating access conditions {lingo.id}')
```
KPrasch, line 116: Here's another place where lingo's ID is being used.

nucypher/policy/conditions/evm.py
```diff
@@ -163,9 +160,10 @@ def _configure_provider(self, providers: Dict[int, BaseProvider]):
         try:
             provider = providers[self.chain]
         except KeyError:
-            # TODO Use a custom exception class, and catch bubble it up to include info about the node
-            # QUESTION Are nodes required to provide connections to all providers?
-            raise Exception(f'This node does not have a connection to chain {self.chain}')
+            raise self.NoConnectionToChain(
```
KPrasch, line 163: This exception can be relocated to `nucypher.policy.conditions.exceptions`.

nucypher/policy/conditions/exceptions.py
```diff
@@ -0,0 +1,50 @@
...
+class InvalidCondition(ValueError):
+    """Invalid value for condition."""
+
+
+class ConditionEvaluationFailed(Exception):
+    """Could not evaluate condition."""
+
+
+class RPCExecutionFailed(ConditionEvaluationFailed):
+    """Raised when an exception is raised from an RPC call."""
```
KPrasch, line 50: Looks excellent ✨

nucypher/network/server.py
```diff
@@ -217,7 +217,13 @@ def reencrypt():
...
         for capsule, lingo in packets:
             # raises an exception or continues
-            evaluate_conditions_for_ursula(lingo=lingo, providers=providers, context=context)
+            error = evaluate_conditions(
+                lingo=lingo, providers=providers, context=context
+            )
+            if error:
+                # error cases
+                return Response(*error)
+
```
KPrasch, line 226: This loop can be bypassed using a logical branch if there are no conditions in the `ReencryptionRequest`.

nucypher/policy/conditions/evm.py
derekpierre: Good find. It was relocated, but I need to fix the `self.`

nucypher/network/server.py
derekpierre: I don't think this is that simple, because ReeencryptionRequest.conditions is a list with null values if there are no conditions i.e.`reenc_request.conditions` can be [null, null, null]` which gets converted to a list of `[None, None, None]` via `json.loads(...)`.

Do we go through the entire list check whether all values are `None` - in which case we are basically the same loop that we do now, just earlier. in the method? Can no conditions and conditions be mixed i.e. [None, <ConditionLingo_object>, None]? If not i.e. if the first element is `None` then all the others are, then we can just check the first entry for None - that would be easy if that premise holds...? Unsure about this.

In any case, seems like some nuance here. Left this for now, but interested to get your thoughts.

derekpierre: Resolved via https://github.com/nucypher/nucypher/pull/3007/commits/f893aae94ec9f2deafae65df052b48378a805e79

KPrasch (approved):

PR #3011, opened 2022-11-14
Fixes missing call to response.json on HTTP 400.

KPrasch:
**Type of PR:**
bugfix

**Required reviews:** 
1


derekpierre (approved):
🎸 - suprising this lasted this long, and is not covered by tests.

PR #3018, opened 2022-11-16
Code coverage refinement

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
Code coverage seems to be misreporting. Fix the discrepancy

- Use `coverage` directly instead of via `pytest-cov`
- Improve efficiency by only running coverage on python 3.10, and don't run coverage otherwise
- Only upload coverage reports after all tests have passed

One possible reason is that when using pytest-cov, code coverage is started after functions are defined and have already been loaded instead of before. See https://coverage.readthedocs.io/en/stable/faq.html#q-why-do-the-bodies-of-functions-show-as-executed-but-the-def-lines-do-not

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -32,12 +32,145 @@ jobs:
       - name: Install Solidity Compiler
         run: python ./scripts/installation/install_solc.py
 
+      - name: Check CLI Entrypoint
+        run: nucypher --help
+
       - name: Check Python Entrypoint
         run: python -c "import nucypher; print(nucypher.__version__)"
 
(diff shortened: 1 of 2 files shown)


derekpierre:
Much better - overall coverage is now 81%.

Exception definition lines, import lines etc. are now properly covered. See https://app.codecov.io/gh/nucypher/nucypher/tree/code-cov-refine?displayType=list

KPrasch (approved):
Thank You!

PR #3023, opened 2022-11-17
Relock dependencies

KPrasch:
**Type of PR:**
Other

**Required reviews:** 
1

**What this does:**
- removes unused direct dependencies (`qrcode`, `bandit`, `itsdangerous`, `jinja`, `tzlocal`)
- introduces direct dependency on `mnemonic`
- pins `aiohttp` and `coverage` versions for compatibility
- updates web3.py accommodate in codebase
- Handle type-casting of rust `Address` type to `bytes` for internal usage


nucypher/network/retrieval.py
```diff
@@ -140,7 +140,7 @@ def results(self) -> Tuple[List["RetrievalResult"], List[RetrievalError]]:
             results.append(
                 RetrievalResult(
                     {
-                        to_checksum_address(address): cfrag
+                        to_checksum_address(bytes(address)): cfrag
```
derekpierre, line 143: This is a result of using rust's `Address`? These rust barriers are so explicit...

nucypher/network/resources.py
```diff
@@ -3,8 +3,9 @@
 import os
 from hendrix.facilities.resources import MediaResource
 
+
 def get_static_resources():
     resources = []
     if os.getenv('NUCYPHER_STATIC_FILES_ROOT'):
-        resources.append(MediaResource(os.getenv('NUCYPHER_STATIC_FILES_ROOT'), namespace='statics'))
+        resources.append(MediaResource(os.getenv('NUCYPHER_STATIC_FILES_ROOT').encode(), namespace=b'statics'))
```
derekpierre, line 10: This was the twisted change that led to the hendrix change, that led to this line needing to be changed?

KPrasch: It's related.  Twisted's `putChild` function expects bytes for paths as of 22.10.0.

nucypher/network/retrieval.py
KPrasch: Yes.  `to_checksum_address` does not accept the `Address` type.

derekpierre (dismissed):
🎸 - dev newsfragment?

PR #3032, opened 2022-12-05
Restrict return value `index` (previously `key`) to only be an integer

### LINKED ISSUE #3027: Web3 returns a list not dict, so named return values are not possible (only index is)
To show this:

```python
from web3 import Web3
from pathlib import Path
from nucypher.policy.conditions.evm import ContractCondition

# change the chain id in test_conditions

VECTORS_FILE = Path("tests")/ "data" / "test_conditions.json"
with open(VECTORS_FILE, 'r') as file:
    VECTORS = json.loads(file.read())

condition_json = json.dumps(VECTORS["TStaking"])
condition = ContractCondition.from_json(condition_json)
providers = {1: Web3.HTTPProvider("https://eth-mainnet.gateway.pokt.network/v1/lb/6244b16400b9f0003ad2637f")}
condition._configure_provider(providers)
condition._execute_call(['0x3B42d26E19FF860bC4dEbB920DD8caA53F93c600'])

# returns [0, 0, 0]
```

theref:
Fixes #3027

### DIFF
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -93,15 +93,15 @@ class ReturnValueTest:
         def make(self, data, **kwargs):
             return ReturnValueTest(**data)
 
-    def __init__(self, comparator: str, value: Any, key: Optional[Union[int, str]] = None):
+    def __init__(self, comparator: str, value: Any, key: int = None):
         if comparator not in self.COMPARATORS:
             raise self.InvalidExpression(
                 f'"{comparator}" is not a permitted comparator.'
             )
 
-        if not isinstance(key, (int, str)) and key is not None:
+        if not isinstance(key, int) and key is not None:
             raise self.InvalidExpression(
-                f'"{key}" is not a permitted key. Must be a string or integer.'
+                f'"{key}" is not a permitted key. Must be a an integer.'
             )
 
         if not is_context_variable(value):
@@ -122,19 +122,11 @@ class ReturnValueTest:
 
     def _process_data(self, data: Any) -> Any:
         """
-        If a key is specified, return the value at that key in the data if data is a dict or list-like.
-        Otherwise, return the data.
+        Solidity will only return a list
         """
         processed_data = data
         if self.key is not None:
-            if isinstance(data, dict):
-                try:
-                    processed_data = data[self.key]
-                except KeyError:
-                    raise ReturnValueEvaluationError(
-                        f"Key '{self.key}' not found in return data."
-                    )
-            elif isinstance(self.key, int) and isinstance(data, (list, tuple)):
+            if isinstance(self.key, int) and isinstance(data, (list, tuple)):
                 try:
                     processed_data = data[self.key]
                 except IndexError:
@@ -149,10 +141,10 @@ class ReturnValueTest:
         return processed_data
 
     def eval(self, data) -> bool:
-        if is_context_variable(self.value) or is_context_variable(self.key):
+        if is_context_variable(self.value):
             # programming error if we get here
             raise RuntimeError(
-                f"Return value comparator contains an unprocessed context variable (key={self.key}, value={self.value}) and is not valid "
+                f"Return value comparator contains an unprocessed context variable (value={self.value}) and is not valid "
                 f"for condition evaluation."
             )
 
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -87,21 +87,21 @@ class ReturnValueTest:
         value = fields.Raw(
             allow_none=False, required=True
         )  # any valid type (excludes None)
-        key = fields.Raw(allow_none=True)
+        index = fields.Raw(allow_none=True)
 
         @post_load
         def make(self, data, **kwargs):
             return ReturnValueTest(**data)
 
-    def __init__(self, comparator: str, value: Any, key: int = None):
+    def __init__(self, comparator: str, value: Any, index: int = None):
         if comparator not in self.COMPARATORS:
             raise self.InvalidExpression(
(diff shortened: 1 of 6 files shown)


nucypher/policy/conditions/lingo.py
```diff
@@ -93,15 +93,15 @@ class ReturnValueTestSchema(CamelCaseSchema):
...
 
-    def __init__(self, comparator: str, value: Any, key: Optional[Union[int, str]] = None):
+    def __init__(self, comparator: str, value: Any, key: int = None):
         if comparator not in self.COMPARATORS:
             raise self.InvalidExpression(
                 f'"{comparator}" is not a permitted comparator.'
             )
 
-        if not isinstance(key, (int, str)) and key is not None:
+        if not isinstance(key, int) and key is not None:
```
derekpierre, line 102: Minor nitpick, I would reverse the order to read as "is key set, and then is it an int" - but up to you.

```suggestion
        if key is not None and not isinstance(key, int):
```

nucypher/policy/conditions/lingo.py
```diff
@@ -122,19 +122,11 @@ def _sanitize_value(self, value):
 
     def _process_data(self, data: Any) -> Any:
         """
-        If a key is specified, return the value at that key in the data if data is a dict or list-like.
-        Otherwise, return the data.
+        Solidity will only return a list
```
derekpierre, line 125: Careful, Solidity can return a tuple. The prior docstring was good, just needed tweaking.

tests/acceptance/blockchain/conditions/test_conditions.py
```diff
@@ -490,7 +472,7 @@ def test_subscription_manager_get_policy_policy_struct_condition_key_context_var
         return_value_test=ReturnValueTest(
             comparator="!=",
             value=":sponsor",  # use sponsor sponsor context var
-            key=":sponsorIndex",
```
derekpierre, line 493: L458 (the ":sponsorIndex" entry in the `context`) should also go away.

The test name itself, `test_subscription_manager_get_policy_policy_struct_condition_key_context_var_evaluation`, could probably use a rename since no longer testing `key_context_var`, perhaps `key_and_value_context_var`?

tests/unit/conditions/test_return_value.py
```diff
@@ -37,19 +37,9 @@ def test_return_value_test_schema():
 
 
 def test_return_value_key():
-    test = ReturnValueTest(comparator=">", value="0", key="james")
-    assert test.eval({"james": 1})
-    assert not test.eval({"james": -1})
-
-    with pytest.raises(ReturnValueEvaluationError):
-        test.eval({"bond": 1})
+    with pytest.raises(ReturnValueTest.InvalidExpression):
```
derekpierre, line 40: 👍

derekpierre (commented):
After looking through the code, it does look like the "key" variable in ReturnValueTest should be called "index", as you alluded to here - https://github.com/nucypher/nucypher/issues/3027#issuecomment-1324811414

derekpierre:
This also needs to be removed - https://github.com/theref/nucypher/blob/key/nucypher/policy/conditions/evm.py#L77

derekpierre:
> After looking through the code, it does look like the "key" variable in ReturnValueTest should instead be called "index", as you alluded to here - [#3027 (comment)](https://github.com/nucypher/nucypher/issues/3027#issuecomment-1324811414)

@theref what are your thoughts about this rename?

theref:
> > After looking through the code, it does look like the "key" variable in ReturnValueTest should instead be called "index", as you alluded to here - [#3027 (comment)](https://github.com/nucypher/nucypher/issues/3027#issuecomment-1324811414)
> 
> @theref what are your thoughts about this rename?

You're totally right - and it's best to do it now whilst the ripple effects are small

nucypher/policy/conditions/lingo.py
```diff
@@ -122,37 +122,30 @@ def _sanitize_value(self, value):
...
         """
-        If a key is specified, return the value at that key in the data if data is a dict or list-like.
+        If an index is specified, return the value at that index in the data if data is list-like.
         Otherwise, return the data.
         """
         processed_data = data
-        if self.key is not None:
-            if isinstance(data, dict):
+        if self.index is not None:
+            if isinstance(self.index, int) and isinstance(data, (list, tuple)):
```
derekpierre, line 130: Is this `isinstance(self.index, int)` already checked by this point and therefore not needed ? It's from the `ReturnValueTest` object which already checks that index is an `int` in the constructor.

theref: completely, but I actually like having it there because I feel it makes the code more readable (it's really clear what that line is trying to achieve)

removing it would make the code more efficient, but it's extremely minor

derekpierre:
@KPrasch if you don't have any objections, I'm good to merge this PR.

Nice work @theref 🎸

derekpierre:
> Hmm let's check with a diff tool if this will cause any conflict with #3030. This change itself LGTM.
I did a test rebase and there were no conflicts, so we should be good.

@KPrasch #3030 does have a large surface area for potential conflicts, let's push for reviews/resolutions of comments there. I've already approved it, with one question about a file that could be removed since it is no longer needed.

theref:
> > Hmm let's check with a diff tool if this will cause any conflict with #3030. This change itself LGTM.
> 
> I did a test rebase and there were no conflicts, so we should be good.
> 
> @KPrasch #3030 does have a large surface area for potential conflicts, let's push for reviews/resolutions of comments there. I've already approved it, with one question about a file that could be removed since it may be no longer needed.

taking a look through #3030 now

KPrasch (approved):

PR #3040, opened 2022-12-13
Remove lingering references to CircleCI which is no longer used.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [X] 1
- [ ] 2
- [ ] 3

**What this does:**
Removes some references to Circle CI.

PR #3042, opened 2022-12-15
Fix weird python chars in exception messages/logging/REST responses

### LINKED ISSUE #3041: Weird python characters on PolicyMessageKit errors
I get this error message from PolicyMessageKit. As you can see, the string includes the Python byte literal syntax (b'...'). I think these characters should be escaped in order to get cleaner errors.

```json
{
    "cFrags": {},
    "errors": {
        "0xd274f0060256c186479f2b9f51615003cbcd19E6": "Unauthorized: b'Decryption conditions not satisfied'",
        "0x05Be6D76d2282D24691E28E3Dc1c1A9709d70fa1": "Unauthorized: b'Decryption conditions not satisfied'",
        "0x18F3a9ae64339E4FcfeBe1ac89Bc51aC3c83C22E": "Unauthorized: b'Decryption conditions not satisfied'",
        "0xcbE2F626d84c556AbA674FABBbBDdbED6B39d87b": "Unauthorized: b'Decryption conditions not satisfied'",
        "0xA7165c0229544c84b417e53a1D3ab717EA4b4587": "Unauthorized: b'Decryption conditions not satisfied'"
    }
}
```

The code I use to get this:

```js
const retrievedMessages = await decrypter.retrieve(
        [encryptedMessages],
        conditionContext
);
const decryptedMessages = retrievedMessages.map(
        (mk: PolicyMessageKit) => {
                console.log(mk);
        }
);
```

I'm using nucypher-core@0.4.0 and nucypher-ts@0.9.0.

derekpierre:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
We were using raw content from HTTP responses which is in bytes. Instead, use the decoded text.

Fixes #3041 . 

Need to update ursula nodes and porter (relock deps) on testnet once merged.

### DIFF
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -809,7 +809,11 @@ class Learner:
             self.cycle_teacher_node()
 
         if response.status_code != 200:
-            self.log.info("Bad response from teacher {}: {} - {}".format(current_teacher, response, response.content))
+            self.log.info(
+                "Bad response from teacher {}: {} - {}".format(
+                    current_teacher, response, response.text
+                )
+            )
             return
 
         # TODO: we really should be checking this *before* we ask it for a node list,
@@ -825,7 +829,9 @@ class Learner:
         try:
             metadata = MetadataResponse.from_bytes(response.content)
         except Exception as e:
-            self.log.warn(f"Failed to deserialize MetadataResponse from Teacher {current_teacher} ({e}): {response.content}")
+            self.log.warn(
+                f"Failed to deserialize MetadataResponse from Teacher {current_teacher} ({e}): {response.text}"
+            )
             return
 
         try:
@@ -833,7 +839,8 @@ class Learner:
         except Exception as e:
             # TODO (#567): bucket the node as suspicious
             self.log.warn(
-                f"Failed to verify MetadataResponse from Teacher {current_teacher} ({e}): {response.content}")
+                f"Failed to verify MetadataResponse from Teacher {current_teacher} ({e}): {response.text}"
+            )
             return
 
         # End edge case handling.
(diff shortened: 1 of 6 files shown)


derekpierre:
Will need a rebase after #3030 is merged - will have conflicts.

nucypher/network/nodes.py
```diff
@@ -833,15 +837,18 @@ def learn_from_teacher_node(self, eager=False, canceller=None):
         try:
             metadata = MetadataResponse.from_bytes(response.content)
```
manumonti, line 838: Should we use `response.text` instead of `response.contest` as well?

Same question for lines 838, 841 and 850.

manumonti (approved):
This is great, thank you @derekpierre 

Just a little question in comments, but LGTM ✌️

PS - Good error messages for the tests 👍

nucypher/network/nodes.py
derekpierre: The response code if we get here is 200 i.e. successful execution and therefore we only want the bytes that are returned.

Basically, on success (status code 200), use `response.content` for the bytes returned (serialized representation of various successful results eg. node metadata, reencrypted data etc.) from the endpoint. Those bytes then get deserialized to their corresponding objects such as `MetadataResponse` in this case.  It's not guaranteed that these bytes can be converted to a string for logging purposes.

Only on failure, we use the text because the data for the failure is expected to be a regular text string so calling `response.text` is used.

derekpierre: It does feel like we should not be logging bytes though. Perhaps I'll convert those bytes to hex.

manumonti: Got it, thank you Derek!

manumonti (dismissed):
LGTM ✌️

derekpierre:
Rebased over #3040 .

KPrasch (approved):
it's always nice to see the casual addition of a test to a PR like this one.

PR #3069, opened 2023-03-03
Revive Oryx

KPrasch:


derekpierre:
I was thinking more about this and I'm wondering if the following steps make maintenance easier:
1. Create a `tapir` branch from whatever commit the current tapir docker image uses. The `tapir` branch can be updated from `development` whenever we decide that what's on `lynx` (`development` branch) is good enough to be on `tapir`.
2. Create an`oryx` branch from the `tapir` branch and then merge this change into that.

The reasoning is that:
* Makes maintenance of `oryx` and `tapir` easier
* As Masterfile discovers issues that require fixing (general code issues and not specific to them), it will be easier to integrate into `lynx` (`development` branch), `tapir`, and `oryx` as needed.

It's unclear if `oryx` is going to be for Masterfile only, or potentially others, and whether it is transient or here to stay. I guess at the moment no one else is asking for this kind of set up so just Masterfile for now, but would we allow others on it...I don't see a reason not to to allows others at the moment if they requested. However, once CBD is on mainnet presumably, we would not want projects remaining on `oryx`.

wdyt? Interested in any thoughts/alternatives.

derekpierre:
Chatted with @KPrasch and we can use a tagging strategy instead of branches.

derekpierre (approved):
🎸

theref (approved):

PR #3071, opened 2023-03-21
Use fixed binary value for test

manumonti:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

**What this does:**
The use of randomly generated values for pytest parameters may cause errors in test explorer in Visual Studio Code. Specifically, `test_return_value_serialization` test returns the following error when a random parameter is used:

```
Test result not found for:
./tests/unit/conditions/test_return_value.py::test_return_value_serialization
[\xbd\x01\t\xecV\x05\x04\xe2\x9b(\xb6\xcef\xea\xa4\xfa]
```

Since the rest of the values used in this test are fixed, there is no reason to use a random value for the bytes parameter.


manumonti:
> Is this a specific issue with Visual Studio Code? It's not able to run the test? The underlying issue seems so specific.

Running the tests using the Pytest command works, but I'm not able to run it using VSC.

image

We can keep this random parameter if it is worthwhile, but in my opinion, using a fixed value is a quick fix for this issue and it doesn't affect the good quality of the tests.

derekpierre:
> We can keep this random parameter if it is worthwhile, but in my opinion, using a fixed value is a quick fix for this issue and it doesn't affect the good quality of the tests.

For this test, I agree it is of no material difference to change it. 

It's more so that I'm wondering if it's just a VS configuration issue, or some VS-specific setting.  For example (caveat: I'm no VS Code expert 😅 ):
Screenshot 2023-03-21 at 9 02 28 AM

We do have other tests that use `os.urandom(...)` (assuming that's the problem), so not sure if those will also be a problem. I would be very hesitant to change other tests if this same issue arises again in other tests.

However, if the answer is not obvious, and this one small change gets you unblocked for now, that's fine.

manumonti:
> However, if the answer is not obvious, and this one small change gets you unblocked for now, that's fine.

Actually, this is not blocking me. The intention of this PR is only to facilitate the use of VSCode and its testing suite. But this is not a critical change and of course, we can discard this PR if it isn't good.

> It's more so that I'm wondering if it's just a VS configuration issue, or some VS-specific setting. 

The rest of the tests work well.

I think the problem comes from here:

1. When you "load" the tests (for the first time that you are going to run them, or when you make any change in some test), VSCode runs a **_discovery_** process and gets the names of the tests:

image

2. If there are pytests parameters like in this test [@pytest.mark.parametrize](https://github.com/nucypher/nucypher/blob/c03c0199eb5c92334828841b3ba39a56b17f9a07/tests/unit/conditions/test_return_value.py#L211), VSCode takes each parameter like a different test and use the value of the parameter as test name:

image

Note that the name of the 6th parameter is `xad\xbc '|xe7\×98L\×89 xed\×9alxeb\xee"|×b4\×85\x1a`. This is the random value that `test_value` took during the **_discovery_** process.

3. When the tests are executed, this parameter takes a new value. So VSCode is unable to find the test called `xad\xbc '|xe7\×98L\×89 xed\×9alxeb\xee"|×b4\×85\x1a`.

image

In any case, I repeat, this is not a blocking issue and we can discard this PR.

derekpierre:
Ah, I see. Interesting. Thanks for the investigation. So it's only an issue when `os.urandom()` is used for pytest parametrization, since there is a discrepancy between the value used for discovery (test name) and the value used during execution.

In that case, updating this test is probably a good thing. I suspect there are very few cases (if any others) of `os.urandom` being used for pytest parametrization.

theref (approved):

PR #3054, opened 2023-02-10
Ferveo-DKG Protocol Pathfinder

KPrasch:
This PR targets an epic upstream branch.

##### outcomes

- represents the initial product transition (T1) by integrating the ferveo/tpke cryptosystem into a network-compatible protocol
- replaces *pre-adapted-cbd* with *ferveo-based-cbd*

##### what this does

- implements intranode DKG ceremony protocol with on-chain transcripts
- nodes periodically perform contract event scanning for ritual events emitted by the `Coodinator` contract
- asynchronously executes each round of the ritual participation (transcript generation and aggregation)
- nodes exchange ferveo public keys over P2P
- Integration with `ferveo_py`'s server API (thanks @piotr-roslaniec !)
- Initial integration with `nucypher_core.ThresholdDecryptionRequest` and `ThresholdDecryptionResponse`
- Adapts Enrico for use with ferveo `PublicKey` and `encrypt_for_dkg`
- Introduces new public Bob APIs for fetching decryption shares from the network and decrypting ciphertext

##### Dependency Changes

- `ferveo-python==0.1.7` (<https://github.com/nucypher/ferveo/pull/75>, <https://github.com/nucypher/ferveo/pull/92>)
- umbral `==0.9.2`
- `nucypher-core==0.7.x` (floating in development; depends on modified version of `nucypher_core`)
- `web3==6.2.0`

##### Notes

- this PR has temporary inclusion of `Coordinator.sol` for development purposes
- example test output <https://gist.github.com/KPrasch/7b5a387e96fb99547a9c10a43b19e728>

##### Issues

- ferveo key persistence (via mnemonic derivation)
- transcript persistence
- DKG public key finalization
- management ~~`g1_inv`~~ `DKGPublicParameters`
- staking providers are used for cohort identification, this will require the contracts to verify operator bonding/authorization
- more...

##### New Entities

- `Ritualist`:  superclass of `Ursula` encapsulating all dkg/tdec/cbd functionality
- `RitualTracker`:  handles on-chain ritual events and participation
    - `EventScanner`: sub-component of ritual tracker for handling contract event scanning
    - `EventActuator`: sub-component of `EventScanner` to manage the execution of tasks/rounds triggered by events
- `RitualisticPower`: powers API for the RitualKeypair
    - `RitualKeypair`: subcomponent of `RitualPower` Compatibility layer for Ferveo cryptographic keypairs
- `CoordinatorDeployer`: Testing utility
- `CoordinatorAgent`: Internal API for coordinator contract I/O

PR #3084, opened 2023-04-17
Ferveo-DKG - post public key with aggregated transcripts (missed changes from 3054 rebase)

derekpierre:
Missed commits after rebase for #3054 .

* Verify caller when posting aggregated transcripts to Coordinator contract
* Nodes now post the dkg public key when posting their aggregated transcripts.


KPrasch (approved):
🤠

PR #3086, opened 2023-04-18
Pin web3 version to >= 6.0.0

manumonti:
**Type of PR:**
- Other

**Required reviews:** 
- 2

**What this does:**
Nucypher code is intended to use v6 of web3 dependency. So this should be a dependency requirement.

**Notes for reviewers:**
relock_dependencies script has been run.


derekpierre (approved):
🎸

KPrasch (approved):

PR #3092, opened 2023-04-20
Bugfixes: coordinator event scanning and tdec request/response

derekpierre:
Based over #3093

### DIFF
--- a/nucypher/blockchain/eth/trackers/dkg.py
+++ b/nucypher/blockchain/eth/trackers/dkg.py
@@ -82,7 +82,7 @@ class ActiveRitualTracker:
 
         # Map events to handlers
         self.actions = {
-            contract.events.StartTranscriptRound: self.ritualist.perform_round_1,
+            contract.events.StartRitual: self.ritualist.perform_round_1,
             contract.events.StartAggregationRound: self.ritualist.perform_round_2,
         }
         self.events = list(self.actions)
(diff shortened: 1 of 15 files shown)


nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -140,7 +140,7 @@ def _get_start_block_number(self) -> int:
         ):
             expected_start_block = w3.eth.get_block(expected_start_block.number - 1)
 
-        return expected_start_block.number
```
derekpierre, line 136: Do we need to be careful here. Could this return -1 if the expected_start_block.number is 0 - might be possible in tests?

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -127,13 +127,20 @@ def _get_start_block_number(self) -> int:
...
         )
-        while expected_start_block.number > 0 and expected_start_block.timestamp > target_timestamp:
+        while (
+            expected_start_block.number > 0
+            and expected_start_block.timestamp > target_timestamp
+        ):
             expected_start_block = w3.eth.get_block(expected_start_block.number - 1)
 
-        return expected_start_block.number
+        return int(expected_start_block.number - 1)
```
derekpierre, line 143: Do we need to be careful here. Could this return -1 if the expected_start_block.number is 0 - might be possible in tests?

```suggestion
        return int(expected_start_block.number - 1) if expected_start_block.number > 0 else 0
```

derekpierre: Or maybe adjust the earlier logic to already accomodate the -1...

derekpierre: Resolved via https://github.com/nucypher/nucypher/pull/3092/commits/958c8ff9968ccd359c2f6f427af6244127c353e8

KPrasch (approved):

PR #3093, opened 2023-04-20
Ferveo keypair derived from HKDF via mnemonic

KPrasch:


piotr-roslaniec (approved):

nucypher/crypto/keypairs.py
```diff
@@ -101,19 +102,14 @@ def decrypt_treasure_map(self, etmap: EncryptedTreasureMap, publisher_verifying_
...
-    _private_key_source = FerveoKey
+    _private_key_source = ferveo_py.Keypair.random
     _public_key_method = "public_key"
 
-    def __init__(self, *args, **kwargs) -> None:
-        super().__init__(*args, **kwargs)
+    @classmethod
+    def from_secure_randomness(cls, randomness: bytes) -> 'RitualisticKeypair':
+        """Create a keypair from a precomputed secure source of randomness"""
+        keypair = FerveoKeypair.from_secure_randomness(randomness)
```
derekpierre, line 111: we could:
- assert the length of the bytes provided is the expected size here
```python
assert len(randomness) == ferveo_py.Keypair.secure_randomness_size()
```

OR

- move the following code into this method
```python
size = ferveo_py.Keypair.secure_randomness_size()
keypair = FerveoKeypair.from_secure_randomness(randomness[:size])
```

OR 
- both

```python
size = ferveo_py.Keypair.secure_randomness_size()
assert len(randomness) >= size
keypair = FerveoKeypair.from_secure_randomness(randomness[:size])
```


wdyt?

KPrasch: ✔️

PR #3103, opened 2023-04-21
Dkg Script

derekpierre:
Based over #3092 .

Script for testing optional ritual creation, and encryption/decryption of data using the cohort of nodes.

### DIFF
--- a/scripts/hooks/nucypher_dkg.py
+++ b/scripts/hooks/nucypher_dkg.py
@@ -15,8 +15,18 @@
  along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
 """
 
-# Get an interactive Python session with all the NuCypher agents loaded by running:
+# Run nucypher dkg by running:
+#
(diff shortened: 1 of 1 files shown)


scripts/hooks/nucypher_dkg.py
```diff
@@ -0,0 +1,251 @@
...
+ (at your option) any later version.
+
+ nucypher is distributed in the hope that it will be useful,
+ but WITHOUT ANY WARRANTY; without even the implied warranty of
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU Affero General Public License for more details.
+
+ You should have received a copy of the GNU Affero General Public License
+ along with nucypher.  If not, see <https://www.gnu.org/licenses/>.
+"""
```
KPrasch, line 16: IS this caused by a pre-commit hook?

KPrasch (approved):
This script will look better without the license header but the scripting itself LGTM!

scripts/hooks/nucypher_dkg.py
derekpierre: License removed.

PR #3108, opened 2023-04-25
Acceptance testing with Ape; Relocate contracts to `nucypher-contracts`

KPrasch:
##### what this does

- mostly closes https://github.com/nucypher/nucypher/issues/3055
- follow-up https://github.com/nucypher/nucypher-contracts/pull/66
- ports blockchain tests to Ape (`ape test`)
- removes all solidity source code and contract tests (relocated to `nucypher/nucypher-contracts`)
- adapts Testerchain fixture to handle ape-managed `EthereumTester` as a provider
- introduces `InMemoryContractRegistry.from_ape_deployments`

#####  Removals

- `tests.acceptance.blockchain.eth.sol` package
    - `source/` (solidity source code)
    -  `compile/` (solidity compilation tools)
- `blockchain.eth.deployers` module and `Deployers`
- `blockchain.eth.actors.ContractAdministrator`
- `blockchain.eth.interfaces.BlockchainDeployerInterface`
- `cli.commands.deploy` and supporting `nucypher-deploy` CLI functionality

#### todo

- [x] configure ape for local & remote source compilation
- [x] integrate ape compilation & deployments into nucypher pytest fixtures
- [x] importing ethpm sandard package JSON to contract registries
- [x] make remaining tests pass
    - [x] unit
    - [x] integration
    - [x] acceptance
        - [x] blockchain (ape project)
        - [x] everything else
    
##### follow-ups needed

- [ ] add `eth-ape` to dependencies and relock
    - [ ] remove direct project dependency on `web3` or other dependencies of `eth-ape` in `Pipfile`
- [ ] integrate into CI
    - [x] set `ape-config.yaml` to use `github: nucypher/nucypher-contracts` OR
    - ~~[ ] checkout a local copy of nucypher-contracts on CI and test against it (`local: /path/to/nucypher-contracts`)~~
    - [ ] determine a good default nucypher-contracts branch, ref, or tag to test against `ref: ...`
- [x] ~~install ape as a pytest plugin for debugging support~~ (launch via `pytest -s -m ape_test`)
- [ ] update contributing documentation

##### Co-Authors

@derekpierre 
@KPrasch

PR #3111, opened 2023-05-01
Compatibility changes for python 3.11

jMyles:


derekpierre:
Is this based over #3108 ?

KPrasch:
This commit includes changes from the base branch 🤔

tests/unit/crypto/test_keccak_sanity.py
```diff
@@ -37,15 +37,15 @@ def test_secure_random_range(self):
     def test_keccak_digest(self):
         data = b'this is a test'
 
-        digest1 = sha3.keccak_256(data).digest()
```
derekpierre, line 40: Does it make sense to store some well-known output from the `sha3` implementation as constants and then use for comparison, to confirm/reassure we are still backwards-compatible?

jMyles: ...I thought we decided we didn't care, and that we were OK with obseleting legacy fingerprints?

derekpierre: This test is only changing the library we use to compute the keccak_256 digest, no?  If so, we still compute the keccak digest so I was wondering if we want to ensure that the new library produces the same digest as the old library. Or is that excessive and/or irrelevant now?

KPrasch: Yep, I think it's okay to obsolete old fingerprints.

PR #3126, opened 2023-05-11
Ensure that Bob and Porter can use common code for making Threshold Decryption Requests

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
This is akin to what was done for PRE retrieval.

Bob and Porter will use the same type of client object for performing threshold decryption i.e. if it works for Bob it should work for Porter. Still unencrypted for now.

Relied upon by https://github.com/nucypher/nucypher-porter/pull/26.


newsfragments/3126.dev.rst
piotr-roslaniec, line 1: Is this file intentionally left empty?

derekpierre: Yes. It's basically a refactor of existing functionality and this is a dev release note.

piotr-roslaniec (approved):
:tada:

PR #3127, opened 2023-05-11
Trying nucypher-contracts as a submodule.

jMyles:


KPrasch:
We've chosen another path for now: Keep nucypher/nucypher-contracts as an external repo and consume it via ape.

PR #3130, opened 2023-05-16
Polygon ecosystem DKG coordination (Alpha 4)

KPrasch:
compatibility changes for polygon ecosystem deployments of dkg coordination, although to bulk of the code supports event scanning.

This PR needs additional follow-up to make the tests pass, but it's okay to review now!

Co-Authored by: 
@derekpierre 
@cygnusv 
@KPrasch


derekpierre (approved):
🎸

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -253,7 +261,10 @@ def scan(self):
         # Note that our chain reorg safety blocks cannot go negative
         if self.start_block is None:
             self.start_block = self._get_start_block_number()
-        start_block = max(self.state.get_last_scanned_block() - chain_reorg_safety_blocks, self.start_block)
+        start_block = min(
+            self.state.get_last_scanned_block() - chain_reorg_safety_blocks,
```
derekpierre, line 265: Hmmm 🤔 ...what's the value of `get_last_scanned_block()` on startup?

`start_block` is only set once on startup, so using `min` here could cause it to stay at the same block forever, as the scan tries to progress over time. That's why `max` was used.

derekpierre (changes requested):
Just drawing attention to my comment about `start_block`

derekpierre (dismissed):
Just drawing attention to my comment about `start_block`

nucypher/blockchain/eth/trackers/dkg.py
cygnusv: Ok, that makes sense, but something's weird here. The behavior that prompted us to look here was that in the logs, we saw patterns were blocks seemed to be skipped like the following:
```
Scanning blocks: 1234 - 1234
... 
Scanning blocks: 1244 - 1244
...
Scanning blocks: 1255 - 1254   (yeah sometimes the start block was higher than the end block)
```
Also, the code in this function seem overly complex and trying to do many things at the same time.

derekpierre: We also need some improved testing in this area.

KPrasch:
rebased @ 5c4cc708c7f352d7a5c7751a4ad7ff6a9f1b8082

tests/integration/learning/test_domains.py
```diff
@@ -124,6 +131,7 @@ def test_learner_ignores_stored_nodes_from_other_domains(
     assert pest not in other_staker.known_nodes  # But not anymore.
 
 
+@pytest.mark.skip
```
piotr-roslaniec, line 134: What is the reason for skipping these tests? I assume it's temporary?

KPrasch: Ah yes, thanks for the reminder, I'll attempt to restore these tests before merging.

nucypher/blockchain/eth/clients.py
```diff
@@ -63,11 +63,13 @@ class Web3ClientUnexpectedVersionString(Web3ClientError):
 
 # TODO: This list is incomplete, but it suffices for the moment - See #1857
 POA_CHAINS = {
-    4,    # Rinkeby
-    5,    # Goerli
-    42,   # Kovan
-    77,   # Sokol
+    4,  # Rinkeby
```
cygnusv, line 66: Rinkeby is officially dead ☠️

cygnusv (approved):

Issue #3051, opened 2023-01-30
Use multicalls when evaluating two or more CBD conditions


cygnusv:
Multicall contracts allow you to aggregate several contract calls into a single call. This is very relevant when we have two or more conditions.

Advantages:
* Less latency by reducing calls to ETH provider (i.e. `N` rpc calls can be substituted by 1)
* Ensures all conditions are evaluated on the same block

Disadvantages:
* More code complexity in case there's no good library support for multicalls
* Dependency on a deployed multicall contract (e.g. https://github.com/mds1/multicall), but this is actually not much of an issue

manumonti:
Multicall PoC in Brownie [WIP]: https://github.com/manumonti/multicall-brownie-poc

Also using Hardhat [WIP] (probably discontinued): https://github.com/manumonti/multicall-hardhat-poc

manumonti:
Some updates:

The main way that we are following to get multi-call working is using MakerDAO contracts https://github.com/mds1/multicall, specifically the V3.

But we have to think about how to integrate this into our Python-based nucypher code.


> * https://github.com/eth-brownie/brownie/blob/master/brownie/network/multicall.py

This implementation of multi-call implies using Brownie in production, which doesn't look pretty appropriate.


> * https://github.com/safe-global/safe-eth-py/blob/master/gnosis/eth/multicall.py

This Gnosis implementation uses multicall V2. It would be preferable if we implement V3.


> * https://github.com/banteg/multicall.py

Banteg's `multicall.py` implementation works fine as we have seen in this PoC: https://github.com/manumonti/multicall-brownie-poc

So one path could be integrating this implementation in Nucypher.

-> Adding the module in requirements.txt is not directly possible since a dependencies conflict exists. nucypher requires `eb3==6.0.0b6` and multicall requires web3<6.0 and >=5.27.

-> Currently I'm addressing the option of integrating the multicall.py code in our web3 client layer `clients.py`.

arjunhassard:
Until an adopter specifically requests a (more efficient) way to evaluate multiple conditions concurrently, this can be bumped to the next version

manumonti:
After some talks with my mates, my current path is to create a nucypher-specific multicall.py package based on https://github.com/banteg/multicall.py and integrate it into the nucypher repo.

This implies some problems that I'm trying to solve:

[...]

PR #3139, opened 2023-06-07
[WIP] TimeCondition should use block time instead of system time

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


nucypher/policy/conditions/time.py
```diff
@@ -27,30 +25,32 @@ def make(self, data, **kwargs):
...
     @property
     def timestamp(self):
         return self.return_value_test.value
 
-    def verify(self, *args, **kwargs) -> Tuple[bool, float]:
-        eval_time = time.time()   # system  clock
-        return self.return_value_test.eval(data=eval_time), eval_time
+    def _execute_call(self, parameters: List[Any]) -> Any:
+        """Execute onchain read and return result."""
+        latest_block = self.w3.eth.get_block("latest")
```
manumonti, line 55: In order to implement multicall, this should be replaced by a call to getCurrentBlockTimestamp() in the mds1/multicall contract. But this can be addressed in #3051.

manumonti (approved):
LGTM! ✌️

The use of block timestamp instead of Ursula's system time is pretty convenient when https://github.com/nucypher/nucypher/issues/3051 is implemented. This will allow us to check several conditions and be sure that all of them was evaluated in the same block.

nucypher/policy/conditions/time.py
derekpierre: 👍 . Indeed, let's do it as part of the `multicall` work.

I'm interested to look more deeply at the use of `multicall`  in `nucypher/nucypher`, I guess my uninformed expectation was that a condition object perhaps can returns a `Call` object for its intended condition action, that are collected and then executed as a single multi-call call (pun intended 😄 ).

cygnusv (approved):

newsfragments/3139.feature.rst
```diff
@@ -0,0 +1 @@
+`TimeCondition` now uses the timestamp of the latest block for evaluating conditions, instead of system time. System time condition evaluation is no longer supported.
```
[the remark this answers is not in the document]
derekpierre: ✔️

arjunhassard (approved):
After our discussion it seems there is no advantage to using system time, even for adopters who are just testing or building proof-of-concepts. It's worth noting that using blocktime prevents one category of Ursula error, and makes a condition evaluation correctness proof theoretically easier to produce (with respect to following time-based conditionality) in the future, which would reduce trust in the cohort.

PR #3143, opened 2023-06-12
Simplified tooling for local hacking, particularly python frontend dev

jMyles:
mammoth-cake-closeup

mammoth-cake

Subclasses of `Enrico` and `Bob` that work entirely locally, transparently simulate a DKG cohort and its rituals, and allow the user to force success or failure of decryption.

Uses:

* Rapid iteration of frontend components:  With these tools, the network I/O for both blockchain reads and share gathering is eliminated.

* Ease of styling success and failure cases: since the user can force success or failure, it's much easier to work on the full flow of each case, for multiple ciphertexts and conditions, even in the same runtime.  

* Offline development: when you're on a train in the middle of nowhere or a busy hackathon wifi, sometimes the connectivity is the most frustrating part of trying to iterate.

* Can still be used in situations of testnet downtime or version incompatibility.

* No more using Infura credits for dev.

* Some artists may want to use an original piece for the purposes of styling their interface, but might not want to share it with the testnet.

Next steps:

* A version of Porter which uses these characters to allow users to force success or failure for development with applications which use Porter.


examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,28 @@
+from examples.tdec.demo_utilities.demo_sample_conditions import ten_oclock_florida_time
+from nucypher.characters.chaotic import BobGonnaBob, NiceGuyEddie
```
theref, line 2: what do these characters do that is different to standard `Bob` and ?`Enrico`?

KPrasch:
> What is the benefit of allowing Bob to circumvent Ursula?

The benefit is offline and offchain local hacking tooling for python projects.

theref:
> > What is the benefit of allowing Bob to circumvent Ursula?
> 
> The benefit is offline and offchain local hacking tooling for python projects.

ok, so build the project using offchain objects, then switch over to standard on chain object when you're done hacking?

KPrasch:
> > > What is the benefit of allowing Bob to circumvent Ursula?
> > 
> > 
> > The benefit is offline and offchain local hacking tooling for python projects.
> 
> ok, so build the project using offchain objects, then switch over to standard on chain object when you're done hacking?

Essentially, yes.  This gives another option for fast-paced hackers to get going even in the face of bad conference wifi :-) Yes the testnet is a good option, but think of this as robustness.

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,28 @@
...
+
+plaintext = b"PEACE AD DAWN"
+THIS_IS_NOT_A_TRINKET = 55
+
+enrico = NiceGuyEddie(encrypting_key=THIS_IS_NOT_A_TRINKET)
+bob = BobGonnaBob(domain="lynx")
```
KPrasch, line 8: You almost have a crony theme here, but bob needs a better name (`Bobbykins`?).  Or for the unimaginative, `LocalBob` and `LocalEnrico`

cygnusv: Robert

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,28 @@
+from examples.tdec.demo_utilities.demo_sample_conditions import ten_oclock_florida_time
```
cygnusv, line 1: Maybe you missed to add this module?

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,28 @@
...
+)
+decrypted_cleartext_from_ciphertext = bob.threshold_decrypt(ciphertext=ciphertext)
+decrypted_cleartext_from_tdr = (
+    bob.get_decryption_shares_using_existing_decryption_request(tdr)
+)
+
+assert (
+    decrypted_cleartext_from_ciphertext
+    == plaintext
+    == decrypted_cleartext_from_ciphertext
```
cygnusv, line 27: Is `decrypted_cleartext_from_ciphertext` twice for a particular reason?

examples/tdec/simple_tdec_local.py
jMyles: https://getyarn.io/yarn-clip/c8fe0950-378c-4fc4-a417-b08b61441ec6

jMyles:
Whew - yeah I realize it's confusing at the moment - I only yesterday overcame a `ciphertext verification failed` due to overindentation.  It'll get better.  :-)

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,263 @@
...
+                ferveo.Keypair.random() for _ in range(0, self.shares_num)
+            ]
+
+            validators = [
+                ferveo.Validator(checksum_addresses[i], keypair.public_key())
+                for i, keypair in enumerate(validator_keypairs)
+            ]
+
+            # Validators must be sorted by their public key
+            validators.sort(key=lambda v: v.address)
```
meatballs, line 99: Why use a lambda when there's a perfectly good stdlib (`operator` module) function to do this?
```suggestion
            validators.sort(key=attrgetter("address")
```

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,263 @@
...
+        self, tau, threshold, shares_num, checksum_addresses, session_seed=None
+    ):
+        self.tau = tau
+        self.threshold = threshold
+        self.shares = shares_num
+        self.fake_nodes = [
+            FakeNode(checksum_address) for checksum_address in checksum_addresses
+        ]
+        self.checksum_addresses = checksum_addresses
+        if session_seed == None:
```
meatballs, line 52: This should be an identity comparison, not equality
```suggestion
        if session_seed is None:
```

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,263 @@
...
+
+            self.fake_ritual = _Blasphemy(
+                tau=self.tau,
+                threshold=self.security_threshold,
+                shares_num=self.shares_num,
+                checksum_addresses=checksum_addresses,
+            )
+
+            validator_keypairs = [
+                ferveo.Keypair.random() for _ in range(0, self.shares_num)
```
meatballs, line 90: The start of the range is redundnant. 0 is the default.
```suggestion
                ferveo.Keypair.random() for _ in range(self.shares_num)
```

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,263 @@
...
+            assert self.server_aggregate.verify(
+                self.shares_num, self.aggregation_messages
+            )
+
+    _dkg_insight = DKGInsight()
+
+
+class NiceGuyEddie(Enrico, DKGOmniscient):
+    """
+    Like Enrico, but from Reswervoir Dogs.
```
meatballs, line 137: typo
```suggestion
    Like Enrico, but from Reservoir Dogs.
```

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,263 @@
...
+                    dkg,
+                    ciphertext,
+                    self._learner._dkg_insight.conditions_bytes,
+                    validator_keypair,
+                )
+
+                decryption_share_bytes = bytes(decryption_share)
+
+                ##### Uncomment for sanity check
+                # ferveo.DecryptionShareSimple.from_bytes(decryption_share_bytes)  # No IOError!  Let's go!
```
meatballs, line 235: Why not make this an assertion?
```suggestion
                assert(ferveo.DecryptionShareSimple.from_bytes(decryption_share_bytes))   # No IOError!  Let's go
```

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,263 @@
+import json
+from typing import Dict, Tuple
+
+from ferveo_py import DkgPublicParameters
+from nucypher_core import (
+    EncryptedThresholdDecryptionResponse,
+    SessionSecretFactory,
+    SessionStaticKey,
```
meatballs, line 8: My linter says this isn't used

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,263 @@
...
+        security_threshold = 3
+        shares_num = 4
+
+        def __init__(self):
+            """ """
+
+            def gen_eth_addr(i: int) -> str:
+                return f"0x{i:040x}"
+
+            checksum_addresses = [gen_eth_addr(i) for i in range(0, self.shares_num)]
```
meatballs, line 80: Unless I'm missing something, I think this could be done within the comprehension:
```suggestion
            checksum_addresses = [f"0x{i:040x}" for i in range(self.shares_num)]
```

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,263 @@
...
+    ):
+        self.tau = tau
+        self.threshold = threshold
+        self.shares = shares_num
+        self.fake_nodes = [
+            FakeNode(checksum_address) for checksum_address in checksum_addresses
+        ]
+        self.checksum_addresses = checksum_addresses
+        if session_seed == None:
+            session_seed = b"ABytestringOf32BytesIsNeededHere"
```
meatballs, line 53: The comparison should really be identity rather than equality and this could be done with a (more declarative) ternary operator:
```suggestion
            session_seed = b"ABytestringOf32BytesIsNeededHere" if session_seed is None else session_seed
```

examples/tdec/simple_tdec_local.py
jMyles: Yes, it was a placeholder for both paths of the API (decrypting with the parameters, and decrypting with a pre-formed threshold decryption request).  Near-term vision is to turn this into a file format.

nucypher/characters/chaotic.py
jMyles: It's bad enough that it's a comprehension at all.  But shoving the string substitution in there as well is the stuff of nightmares.

meatballs: declarative > imperative every time!

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,49 @@
...
+    ritual_id=ANYTHING_CAN_BE_PASSED_AS_RITUAL_DATA,
+)
+
+decrypted_cleartext_from_ciphertext_list = bob.threshold_decrypt(
+    ciphertext=ciphertext,
+    ritual_id=ANYTHING_CAN_BE_PASSED_AS_RITUAL_DATA,
+    conditions=before_the_beginning_of_time,
+)
+
+THESE_CAN_BE_FAKE_FOR_THE_PURPOSES_OF_THIS_DEMO = []
```
KPrasch, line 33: Perhaps a comment here?  It's unclear what this is meant to be used for.

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,290 @@
...
+    He doesn't know who's shot, who's not.
+    """
+
+    def __init__(self, encrypting_key, *args, **kwargs):
+        del encrypting_key  # We take this to match the Enrico public API, but we don't use it, because...
+
+        # ...we're going to use the DKG public key as the encrypting key, and ignore the key passed in.
+        encrypting_key_we_actually_want_to_use = self._dkg_insight.dkg.public_key
+        super().__init__(
+            # https://imgflip.com/i/7o0po4
```
derekpierre, line 148: 😂

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,290 @@
...
+            decryption_response = ThresholdDecryptionResponse(
+                ritual_id=55,  # TODO: Abstract this somewhere
+                decryption_share=bytes(decryption_share_bytes),
+            )
+
+            encrypted_decryptiopn_response = trdp.encrypt_decryption_response(
+                decryption_response=decryption_response,
+                requester_public_key=etdr.requester_public_key,
+            )
+            responses[validator.address] = encrypted_decryptiopn_response
```
derekpierre, line 220: Typo here:

```suggestion
            encrypted_decryption_response = trdp.encrypt_decryption_response(
                decryption_response=decryption_response,
                requester_public_key=etdr.requester_public_key,
            )
            responses[validator.address] = encrypted_decryption_response
```

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,290 @@
...
+            encrypting_key=encrypting_key_we_actually_want_to_use,
+            *args,
+            **kwargs,
+        )
+
+
+class DKGOmniscientDecryptionClient(ThresholdDecryptionClient):
+    def gather_encrypted_decryption_shares(
+        self,
+        encrypted_requests,
```
derekpierre, line 158: typehint would be nice here:

```suggestion
        encrypted_requests: Dict[ChecksumAddress, EncryptedThresholdDecryptionRequest],
```

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,290 @@
...
+
+
+class DoomedDecryptionClient(ThresholdDecryptionClient):
+    """
+    A decryption client that always fails, claiming that conditions are not satisfed.
+    """
+
+    def gather_encrypted_decryption_shares(
+        self,
+        encrypted_requests,
```
derekpierre, line 233: Typehint would be nice here:

```suggestion
        encrypted_requests: Dict[ChecksumAddress, EncryptedThresholdDecryptionRequest],
```

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,49 @@
...
+
+cohort = bob._dkg_insight.fake_ritual.fake_nodes
```
derekpierre, line 35: To keep inner workings hidden, can this be?

```suggestion
ritual = bob.get_ritual_from_id(ritual_id=ANYTHING_CAN_BE_PASSED_AS_RITUAL_DATA)
cohort = bob.resolve_cohort(ritual=ritual, timeout=<>)
```

Then the inner workings can remain hidden.

You can use `ritual` from above on L39 to get the participant public keys.

examples/tdec/simple_tdec_using_testnet.py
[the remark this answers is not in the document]
jMyles: I thought about that., but I think that `simple_tdec_testnet` sounds like it's a tool to launch a testnet.

examples/tdec/simple_tdec_local.py
jMyles: Yeah that's probably a good idea.

nucypher/characters/lawful.py
```diff
@@ -772,12 +787,40 @@ def threshold_decrypt(
...
+
+        addresses_and_dfrags = (
+            self.get_decryption_shares_using_existing_decryption_request(
+                decryption_request, participant_public_keys, cohort, threshold
+            )
+        )
+
+        # TODO: 3154
+        conditions_as_json_string = str(decryption_request.conditions)
+        conditions_as_list = json.loads(conditions_as_json_string)
```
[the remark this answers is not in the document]
jMyles: Righto - gotta fix these.

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,49 @@
+from nucypher.characters.chaotic import NiceGuyEddie, ThisBobAlwaysDecrypts
+from nucypher.policy.conditions.lingo import ConditionLingo
+
+plaintext = b"paz al amanecer"
+THIS_IS_NOT_A_TRINKET = 55
+
+enrico = NiceGuyEddie(encrypting_key=THIS_IS_NOT_A_TRINKET)
+bob = ThisBobAlwaysDecrypts(domain="lynx", eth_provider_uri="Nowhere")
+
+ANYTHING_CAN_BE_PASSED_AS_RITUAL_DATA = 55
```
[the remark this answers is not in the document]
jMyles: Yeah, I used this name because, at the time, I was faking a whole bunch of ritual data (not just the ID).

nucypher/characters/chaotic.py
```diff
@@ -0,0 +1,290 @@
...
+        # Everybody has the same public key at the moment.
+        fifty_fiver = (
+            self.threshold_request_decrypting_power._get_static_secret_from_ritual_id(
+                55
+            )
+        )
+        return fifty_fiver.public_key()
+
+
+class _Blasphemy:
```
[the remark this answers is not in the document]
jMyles: It was previously called `FakeRitual`, but @KPrasch and I both thought this was a misnomer - it's neither fake nor a ritual, so we thought that a more unrelated name might be less confusing.  I'm not sure `_Blasphemy` is it, but the internals are reasonably plain to see for anyone who is sufficiently interested.

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,49 @@
+from nucypher.characters.chaotic import NiceGuyEddie, ThisBobAlwaysDecrypts
```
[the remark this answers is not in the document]
jMyles: I've heard worse ideas.  :-)

tests/integration/characters/test_dkg_and_testnet_bypass.py
```diff
@@ -0,0 +1,68 @@
...
+        conditions=definitely_false_condition,
+        ritual_id=ANYTHING_CAN_BE_PASSED_AS_RITUAL_DATA,
+    )
+
+    decrypted_cleartext_from_ciphertext_list = bob.threshold_decrypt(
+        ciphertext=ciphertext,
+        ritual_id=ANYTHING_CAN_BE_PASSED_AS_RITUAL_DATA,
+        conditions=definitely_false_condition,
+    )
+
```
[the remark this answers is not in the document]
jMyles: Do you think it needs to be tested specifically offline?  The logic shows that it works internally.

But I mean, sure, not a terrible idea I guess.

examples/tdec/simple_tdec_local.py
derekpierre: Remind me why we need to use `decrypt_using_existing_decryption_request()` and not just `threshold_decrypt()`? Why save the ThresholdDecryptionRequest in this script? Also, if `decrypt_using_existing_decryption_request()` is a function on Bob, could this function just resolve the cohort internally instead of taking it as a parameter?

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,49 @@
...
+    ciphertext=ciphertext,
+    ritual_id=ANYTHING_CAN_BE_PASSED_AS_RITUAL_DATA,
+    conditions=before_the_beginning_of_time,
+)
+
+THESE_CAN_BE_FAKE_FOR_THE_PURPOSES_OF_THIS_DEMO = []
+
+cohort = bob._dkg_insight.fake_ritual.fake_nodes
+
+decrypted_cleartext_from_tdr_list = bob.decrypt_using_existing_decryption_request(
```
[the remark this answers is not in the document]
jMyles: It's a list of bytes.   But yeah, bad naming.  I'll fix.

nucypher/characters/lawful.py
```diff
@@ -438,7 +438,9 @@ def decrypt_message_kit(self, label: bytes, message_kit: MessageKit) -> List[byt
 
 class Bob(Character):
     banner = BOB_BANNER
+    default_dkg_variant = FerveoVariant.PRECOMPUTED
```
derekpierre, line 441: @piotr-roslaniec is `precomputed` supposed to be the default or `simple`. I thought `simple` was the one with an actual threshold < num_shares, and for `precomputed` threshold == num_shares.

I keep mixing up the two 😓 ...

piotr-roslaniec: Both should work just fine. It's [my humble opinion](https://discord.com/channels/866378471868727316/866378471868727319/1125366637592313916) that we should `precomputed` as a default, but if there is a preference for more robustness (threshold < num_shares) then there is nothing stopping us from using `simple` variant. We just need to make sure that the decrypted understands which variant is being used to collect the necessary amount of decryption shares.

derekpierre: > It's [my humble opinion](https://discord.com/channels/866378471868727316/866378471868727319/1125366637592313916)

The issue is not the ritual really (yes that requires all nodes to participate), but rather on the decryption side i.e. Bob. If I understand correctly, `precomputed` requires all `n` decryption fragments to be obtained before decryption vs `simple` which requires a threshold of them i.e. only `m`-of-`n` where `m` < `n`.

`precomputed` therefore, can lead to availability issues i.e. if any node wasn't properly up and running. Only requiring a threshold is one of the main tenets of our fundamental value prop for CBD. For example, the parameter package of 16-of-30 or whatever it was would not be possible with `precomputed`. 

If my understanding is correct, then I think `simple` should be the default, and if some adopter wants to risk availability by using `precomputed` they can do so if they like. (cc @cygnusv , @arjunhassard )

tests/integration/characters/test_dkg_and_testnet_bypass.py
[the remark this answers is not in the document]
derekpierre: That function has separate logic that hackers may use, and therefore should be CI tested. it's unclear to me to the function gets tested otherwise - although perhaps I'm missing something.

nucypher/characters/chaotic.py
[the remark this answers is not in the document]
KPrasch: Is this perhaps a stand-in for `Coordinator`?

KPrasch (approved):
There are still some changes that can be made for clarity but the tooling here is very useful for python-based hackathon projects as-is (particularly with bad wifi! 😆)

nucypher/characters/lawful.py
jMyles: I agree that `simple` makes more sense as the default.

We also need to bury it better - it needs to be modifiable only with great intent.

examples/tdec/simple_tdec_local.py
jMyles: > Why save the ThresholdDecryptionRequest in this script? 

One of the emerging interfaces is a file format, `.tdr`, which allows people to simply pass a file to the network as a request.

Imagine being able to right-click a file and click "Decrypt with Threshold".  Or pass a file to a discord bot and have it decrypt for you.

@KPrasch points out that it makes more sense to divide this into multiple demos.  Which yeah.  But for now, it's useful to just see both ways.

> Also, if decrypt_using_existing_decryption_request() is a function on Bob, could this function just resolve the cohort internally instead of taking it as a parameter?

Dude heck yeah.  I mean, better even then that is to just include the cohort in the tdr.

tests/integration/characters/test_dkg_and_testnet_bypass.py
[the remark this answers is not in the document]
jMyles: Oh you're right.  I was thinking of `decrypt_using_existing_decryption_request`.

nucypher/characters/chaotic.py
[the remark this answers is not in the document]
jMyles: Yeah - how about `Uncoordinated` ?

examples/tdec/simple_tdec_local.py
```diff
@@ -0,0 +1,48 @@
+from nucypher.characters.chaotic import NiceGuyEddie as _Enrico
+from nucypher.characters.chaotic import ThisBobAlwaysDecrypts
+from nucypher.policy.conditions.lingo import ConditionLingo
+
+plaintext = b"paz al amanecer"
+THIS_IS_NOT_A_TRINKET = 55
```
KPrasch, line 6: I know this is a critique - but perhaps include a comment?

```suggestion
THIS_IS_NOT_A_TRINKET = 55  # sometimes called "public key"
```

tests/integration/characters/test_dkg_and_testnet_bypass.py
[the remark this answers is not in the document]
jMyles: See: https://github.com/nucypher/nucypher/issues/3172

derekpierre (approved):
🎸

PR #3157, opened 2023-06-23
Removes `ValidatorMessage` usage from `dkg.py`

KPrasch:
**Type of PR:**
Bugfix

**Required reviews:** 
1

### DIFF
--- a/nucypher/crypto/ferveo/dkg.py
+++ b/nucypher/crypto/ferveo/dkg.py
@@ -54,7 +54,7 @@ def aggregate_transcripts(
 ) -> Tuple[AggregatedTranscript, DkgPublicKey]:
     validators = [t[0] for t in transcripts]
     _dkg = _make_dkg(nodes=validators, shares=shares, *args, **kwargs)
-    validator_msgs = [ValidatorMessage(v[0], v[1]) for v in transcripts]
+    validator_msgs = [(v[0], v[1]) for v in transcripts]
     pvss_aggregated = _dkg.aggregate_transcripts(validator_msgs)
     verify_aggregate(pvss_aggregated, shares, validator_msgs)
     LOGGER.debug(f"derived final DKG key {bytes(_dkg.public_key).hex()[:10]}")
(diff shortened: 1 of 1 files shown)


derekpierre (changes requested):
Confirmed locally that the failure you encountered was due to an old `nucypher-core` version (< 0.10.0) on your machine so this change is not needed. 

After undoing this change locally, the unit tests pass for me with `nucypher-core` 0.10.0:

```
derek @ MacBook-Pro ~/Documents/Github/repos/forks/kieran/nucypher
 [5] → git diff
diff --git a/nucypher/crypto/ferveo/dkg.py b/nucypher/crypto/ferveo/dkg.py
index 05442910b..19412438f 100644
--- a/nucypher/crypto/ferveo/dkg.py
+++ b/nucypher/crypto/ferveo/dkg.py
@@ -1,7 +1,6 @@
 from enum import Enum
 from typing import List, Tuple, Union

-from eth_utils import keccak
 from nucypher_core.ferveo import *

@@ -54,7 +53,7 @@ def aggregate_transcripts(
 ) -> Tuple[AggregatedTranscript, DkgPublicKey]:
     validators = [t[0] for t in transcripts]
     _dkg = _make_dkg(nodes=validators, shares=shares, *args, **kwargs)
-    validator_msgs = [(v[0], v[1]) for v in transcripts]
+    validator_msgs = [ValidatorMessage(v[0], v[1]) for v in transcripts]
     pvss_aggregated = _dkg.aggregate_transcripts(validator_msgs)
     verify_aggregate(pvss_aggregated, shares, validator_msgs)
     LOGGER.debug(f"derived final DKG key {bytes(_dkg.public_key).hex()[:10]}")
@@ -64,7 +63,7 @@ def aggregate_transcripts(
 def verify_aggregate(
     pvss_aggregated: AggregatedTranscript,
     shares: int,
-    transcripts: List[Tuple[Validator, Transcript]],
+    transcripts: List[ValidatorMessage],
 ):
     pvss_aggregated.verify(shares, transcripts)

```

We may need to rebuild `dkg-dev-7` tags / images if this change was incorporated.

KPrasch:
Thanks!  You're correct -I had the wrong version of nucypher-core installed.  Since the import at the top of the file uses `*` it was not obvious what the issue was.  I'll simply close this PR without merging.  Also - PR #3158 will help resolve this by using explicit imports (although we need to be careful about name-shadowing conventions emerging for objects like `Ciphertext`).

derekpierre:
No worries.

That being said, do we need to rebuild and retag `dkg-dev-7` for `nucypher` then (from last week)? I assume it included this change?

PR #3158, opened 2023-06-25
Linting

meatballs:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
Fixes several linting errors

**Notes for reviewers:**
The only two remaining linting error codes that persist in the codebase after this PR are:
> F401 - line length. That tends to fixed incrementally by darken
> F821 - undefined name. This one is a bug in several linters which give a false positive where string hints are used.


nucypher/crypto/ferveo/dkg.py
```diff
@@ -1,8 +1,18 @@
...
+    Ciphertext,
+    DecryptionSharePrecomputed,
+    DecryptionShareSimple,
+    Dkg,
+    DkgPublicKey,
+    Keypair,
+    Transcript,
+    Validator,
+    ValidatorMessage,
+)
```
KPrasch, line 15: This is a clear improvement, although I do have concerns about emerging name-shadowing conventions for common names like `Ciphertext`.   Alternate suggestion:

```py
from nucypher_core import ferveo
...
ferveo.Ciphertext
``` 

is better?

KPrasch (approved):

nucypher/crypto/ferveo/dkg.py
meatballs: That was my preference too, but it broke the unit tests, so I went with this for now.

meatballs: I could fix that up and go for the namespace solution instead, if you want, but that felt like something best done in a separate PR. I also wasn't sure how deep that rabbit hole might go...

nucypher/utilities/seednodes.py
```diff
@@ -1,14 +1,14 @@
...
-from json.decoder import JSONDecodeError
-
-import json
-
-import os
-
-from typing import Set, Optional, Dict, List
-
-from nucypher.config.constants import DEFAULT_CONFIG_ROOT
+# from json.decoder import JSONDecodeError
```
derekpierre, line 3: At this point, this entire file is commented out and therefore unused. If you can delete it, and the test suite passes, then let's just delete it.

nucypher/crypto/ferveo/dkg.py
derekpierre: 👍 . Given this module is specifically for `dkg.py` and therefore `ferveo`-specific I believe there shouldn't be any name shadowing with `umbral` etc. in here.

Definitely need to be explicit in other common spots though.

nucypher/utilities/seednodes.py
meatballs: Yep! I figured there must be a good reason for it still existing and didn't dare get rid!

derekpierre: Heh, no good reason as far as I know. You have the green light to remove🚦- just blame it on me if any issue subsequently comes up 😅

meatballs:
For the false +ve F821 errors, I'd suggest adding #noqa F821 to each line where a string hint is used.

The alternative is to ignore that error either for an entire file or the whole project, but that means any genuine errors would also be suppressed.

I'm happy to get that done and add it to this PR if we're happy that's the way to go.

derekpierre:
> For the false +ve F821 errors, I'd suggest adding #noqa F821 to each line where a string hint is used.
> 
> The alternative is to ignore that error either for an entire file or the whole project, but that means any genuine errors would also be suppressed.
> 
> I'm happy to get that done and add it to this PR if we're happy that's the way to go.

Great work on this PR! Thanks 🙏 

We can think about the false errors for string hints later on. I'm not sure how extensively we use string typehints, but adding `#noqa 821` to each relevant line seems excessive; perhaps it can be considered a flakes bug...? Maybe there are some discussions in their github about properly addressing/working around those without needing to add the comment to our code.

derekpierre (approved):
🎸

meatballs:
I'll have a proper read of https://peps.python.org/pep-0484/#forward-references and see if these actually are false +ves or not...

meatballs:
I'd like to get to the point where whichever linter we choose shows zero errors. Then, we can add a pre commit hook and gh action check.

Line length, we can ignore and just let darker do its thing. Which just leaves these F821 errors...

PR #3165, opened 2023-06-29
Add linter config

### LINKED ISSUE #3161: Undefined Name Errors
Assuming #3159 is merged, that will leave 5 F821 errors in the codebase. These can be seen using, for example, [ruff](https://github.com/astral-sh/ruff):
```bash
>>> ruff check --select F821 nucypher
nucypher/blockchain/eth/interfaces.py:523:23: F821 Undefined name `JSONRPCStdoutEmitter`
nucypher/cli/actions/select.py:98:22: F821 Undefined name `Staker`
nucypher/cli/config.py:56:23: F821 Undefined name `JSONRPCStdoutEmitter`
nucypher/cli/painting/deployment.py:119:72: F821 Undefined name `DispatcherDeployer`
nucypher/cli/painting/deployment.py:123:35: F821 Undefined name `DispatcherDeployer`
Found 5 errors.
```

I can't find those three classes defined anywhere either within this repo or nucypher-core.

meatballs:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Adds linting with ruff to pre-commit and gh actions workflow

**Issues fixed/closed:**
- Fixes #3161 

**Notes for reviewers:**
- I've gone with ruff as the linter and for ignoring line length altogether


meatballs:
@derekpierre I'm not sure why this is failing. My suspicion is that it's something to do with me adding ruff to the dev dependencies and rebuilding the dev-requirements file. Presumably, that mean the dependencies cache is now invalid for the first time in a while.

derekpierre:
> @derekpierre I'm not sure why this is failing. My suspicion is that it's something to do with me adding ruff to the dev dependencies and rebuilding the dev-requirements file. Presumably, that mean the dependencies cache is now invalid for the first time in a while.

Already answered on discord but putting this here. The dependency files aren't correct. Run `scripts/dependencies/relock_dependencies.sh` from the topmost `nucypher` folder.

derekpierre:
My questions about `ruff`:
- Seems like `ruff` is only in beta `0.0.275` - is there a major release planned? Any concerns about potential ongoing breaking changes before a major release is done.
- Does it allow for only modifying chunks that are changed as part of a commit, like darker (https://github.com/akaihola/darker),  instead of being all-encompassing?

meatballs:
I haven't set ruff to make any changes - only to report errors. I personally don't like linters making changes anyway - they don't always get it right and, often, the real solution is to a `noqa` instruction.

meatballs:
I've also pinned the version, so any changes to ruff would have to be done via PR here anyway

meatballs:
Seems we have a dependency resolution problem now. I'm pretty sure this isn't something I've caused by just adding a linter!

derekpierre:
> I haven't set ruff to make any changes - only to report errors. I personally don't like linters making changes anyway - they don't always get it right and, often, the real solution is to add a noqa instruction.

I must admit, I wouldn't like having #noqa instructions all over our code.

Currently, our `darker` pre-commit hook makes local changes when attempting to do `git commit` and those can either be accepted or rejected. I think I like that set-up, although perhaps others don't feel the same way. It ensures that our local pre-commit hook (actually makes changes) and the CI action (only reporting) are in lockstep, unless purposely ignored. Also because the CI action is only for modified chunks, the errors don't persist across subsequent PRs which I like a lot. I don't want to think too much about linting.

Any CI task that fails, should in most cases be fixed. Ideally, we don't want to have an over-aggressive linter CI action raising errors that don't get caught by the pre-commit hook, or else the CI action will be ignored by everyone over time.

derekpierre:
> Seems we have a dependency resolution problem now. I'm pretty sure this isn't something I've caused by just adding a linter!

Indeed. Welcome to [dependency hell](https://en.wikipedia.org/wiki/Dependency_hell) 🔥 .

It's just caused by the dependency tree updating. There probably needs to be some version restriction to package(s) to make everthing fine again.

meatballs:
It would be perfectly possible to add the `--fix` flag to the pre-commit hook but not to the github action. That way, any automated changes are under the control of the developer locally.

meatballs:
Sorry, but I don't have the time available to get stuck into fixing this dependency problem. I was expecting this to be a simple job to finish off clearing out the linting errors. I'm out of time for the moment.

meatballs:
I had a quick poke around. The problem in the gh action is:
```
The conflict is caused by:
    nucypher[dev] 6.2.0 depends on typing-extensions==4.7.0; python_version >= "3.7"
    nucypher[dev] 6.2.0 depends on typing-extensions==4.7.0; python_version >= "3.7" and extra == "dev"
    ape-solidity 0.6.6 depends on typing-extensions==4.5.0
 ```
 
 but I can't spot why there's a `typing-extensions` dependency directly on `nucypher[dev]`. It's not listed in `Pipfile`, so where's it coming from?

File: .github/workflows/python_tests.yml [hunk 1]
```diff
@@ -43,6 +43,9 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install .[dev]
+          
+      - name: Lint with ruff
```
derekpierre, line 47: Any reason this needs to be done as part of the python test workflow, what about doing it on its own like `darker`?

Also, do we still need `darker` if `ruff` is now being used?

meatballs: > Any reason this needs to be done as part of the python test workflow, what about doing it on its own like darker?

I added into here because this workflow already installs a heap of stuff and it seemed more efficient to add one rather than creating a whole new image build. Happy to split it out if that's what's preferred.

> Also, do we still need darker if ruff is now being used?

Yes, they do very different things. darker is a formatter, ruff is a linter. The former only cares about code layout; the latter cares about the code itself and looks for common anti-patterns.

derekpierre: > I added into here because this workflow already installs a heap of stuff and it seemed more efficient to add one rather than creating a whole new image build. Happy to split it out if that's what's preferred.

The path I'm going with this question is if we don't do it as part of python tests, is it even needed in the dependency requirements. Is this the only reason you need to add `ruff` to the dependency requirements? It's like `darker` - we don't need a dependency on `darker`, it's only used in pre-commit and CI, and not in the code itself.

derekpierre: 👍 I think if we split it out, then we don't need to update the Pipfile and requirements files, or `pyproject.toml` which is a lot simpler.

Taking a step back, `ruff` isn't really a `nucypher` dependency, but rather a tool used for builds/pre-commit hooks, so splitting it out seems cleaner to me.

meatballs: OK. I'll get that done at some point...

meatballs:
@derekpierre Is this now more how you had in mind?

KPrasch (approved):

theref (approved):

derekpierre:
> @derekpierre Is this now more how you had in mind?

👍 . Indeed

Just some questions for you regarding removing the options passed to `darker`.

meatballs:
Both darker and ruff can fix import sorting. However, ruff can also fix a lot more - imports is simply one of those.

It therefore made sense to move that job into ruff. The remaining two options that were being passed darker are the defaults anyway, so that entire options line could then be removed.

For now, ruff is set to report only. I figured that choosing what and if ruff should automatically fix is a topic for a whole other conversation later on!

derekpierre:
> Both darker and ruff can fix import sorting. However, ruff can also fix a lot more - imports is simply one of those.
> 
> It therefore made sense to move that job into ruff. The remaining two options that were being passed darker are the defaults anyway, so that entire options line could then be removed.
> 
> For now, ruff is set to report only. I figured that choosing what and if ruff should automatically fix is a topic for a whole other conversation later on!

Righto! Thanks for clarifying.

derekpierre (approved):
🎸

PR #3169, opened 2023-07-04
Update dkg script to use `click` options - exec args were getting complex

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Update `nucypher_dkg` script to use `click` instead of exec args.

Successfully created rituals # 14 and # 15 on `lynx` using the script.

```
$ python ./scripts/hooks/nucypher_dkg.py --eth-provider <> --coordinator-provider <> --ritual-id -1 --signer keystore://<>
...
--------- Initiating Ritual ---------
Using account 0xfDd99d58a7D5696c4f757fAFcAc401d5c9AB87A3 to initiate DKG Ritual
Enter your keystore password:
Commencing DKG Ritual on mumbai using 0xfDd99d58a7D5696c4f757fAFcAc401d5c9AB87A3
Using staking providers for DKG: ['0x210eeAC07542F815ebB6FD6689637D8cA2689392', '0xb15d5A4e2be34f4bE154A1b08a94Ab920FfD8A41']
Broadcasting INITIATERITUAL Legacy Transaction (0.00022690050453801 ETH @ 1.50000003 gwei)
TXHASH 0xeb8485c8d4bc2d4fb004aed4a88ee4e0eac5b27752f6020143c763a37f44c6e1
Waiting 600 seconds for receipt
DKG Ritual #15 initiated: 0xeb8485c8d4bc2d4fb004aed4a88ee4e0eac5b27752f6020143c763a37f44c6e1
Waiting for Ritual to complete; 0s elapsed thus far
Waiting for Ritual to complete; 20s elapsed thus far
Waiting for Ritual to complete; 40s elapsed thus far
Waiting for Ritual to complete; 60s elapsed thus far
Waiting for Ritual to complete; 80s elapsed thus far
DKG Ritual #15 completed after 101s
--------- Data Encryption ---------
...
```

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/scripts/hooks/nucypher_dkg.py
+++ b/scripts/hooks/nucypher_dkg.py
@@ -58,15 +58,16 @@ emitter = StdoutEmitter(verbosity=2)
     "--ritual-id",
     "ritual_id",
     "-r",
-    help="Ritual ID; existing id or -1 (to initiate a new ritual)",
+    help="Ritual ID; defaults to -1 to initiate a new ritual",
     type=click.INT,
+    default=-1,
     required=True,
 )
 @click.option(
     "--signer",
     "signer_uri",
     "-S",
-    help="Signer URI for initializing new ritual",
+    help="Signer URI for initiating a new ritual with Coordinator contract",
     default=None,
     type=click.STRING,
 )
@@ -82,7 +83,8 @@ def nucypher_dkg(
         raise click.BadOptionUsage(
             option_name="--ritual-id, --signer",
             message=click.style(
-                "--signer must be provided when --ritual-id is -1", fg="red"
+                "--signer must be provided to create new ritual when --ritual-id is not provided",
+                fg="red",
             ),
         )
 
--- a/scripts/hooks/nucypher_dkg.py
+++ b/scripts/hooks/nucypher_dkg.py
@@ -1,21 +1,9 @@
-# Run nucypher dkg on 'lynx'/'mumbai' by running:
-#
-#    python scripts/hooks/nucypher_dkg.py <ETH_PROVIDER_URI> <COORDINATOR_PROVIDER_URI> <RITUAL_ID> <SIGNER_URI_FOR_COORDINATOR_NETWORK>
-#
-# For example:
-#
-# Use existing ritual id (eg. id '0'):
-#    python ./scripts/hooks/nucypher_dkg.py <ETH_PROVIDER_URI> <COORDINATOR_PROVIDER_URI> 0
-#
-# Go through entire process of initiating a ritual, waiting for it to finish, then use for encryption/decryption:
-#    python ./scripts/hooks/nucypher_dkg.py <ETH_PROVIDER_URI> <COORDINATOR_PROVIDER_URI> -1 <SIGNER_URI_FOR_COORDINATOR_NETWORK>
-#
-
 import random
 import sys
 import time
 
 import click
+import maya
 from nucypher_core.ferveo import DkgPublicKey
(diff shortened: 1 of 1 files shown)


scripts/hooks/nucypher_dkg.py
```diff
@@ -36,191 +24,224 @@
...
+    type=click.Choice(["mumbai"]),
+    default="mumbai",
+)
+@click.option(
+    "--ritual-id",
+    "ritual_id",
+    "-r",
+    help="Ritual ID; existing id or -1 (to initiate a new ritual)",
+    type=click.INT,
+    required=True,
```
manumonti, line 63:
```suggestion
    default=-1,
```

This is just a suggestion. Could it make sense to have a default value of -1?

From UX perspective, it sounds weird to me introducing -1 value when actually I want a new ritual ID to be generated.

With this suggestion, user can use `--ritual-id` argument when there is an initiated ritual, or don't use this argument if there is not an initiated ritual (and therefore, a new ritual will be initiated).

WDYT?

scripts/hooks/nucypher_dkg.py
```diff
@@ -1,21 +1,9 @@
...
-# Go through entire process of initiating a ritual, waiting for it to finish, then use for encryption/decryption:
-#    python ./scripts/hooks/nucypher_dkg.py <ETH_PROVIDER_URI> <COORDINATOR_PROVIDER_URI> -1 <SIGNER_URI_FOR_COORDINATOR_NETWORK>
-#
-
 import random
 import sys
 import time
 
 import click
+import maya
```
manumonti, line 6: Just out of curiosity: What are the advantages of using maya instead of canonical time here?

manumonti (approved):
I really like how the code looks with CLICK, the code is pretty clean. Thanks.

LGTM 💥! I've just left some suggestions/questions.

piotr-roslaniec (approved):

scripts/hooks/nucypher_dkg.py
cygnusv: Agree with Manu here!

cygnusv (approved):
Looks good! I'm wondering if we should turn some of our scripts into ape scripts, there may be benefits, like account management.

scripts/hooks/nucypher_dkg.py
derekpierre: Good point! ✔️

scripts/hooks/nucypher_dkg.py
derekpierre: `maya` makes displaying the elapsed time in seconds easier - https://github.com/derekpierre/nucypher/blob/clickerize-script/scripts/hooks/nucypher_dkg.py#L176.

`time.time() - start_time` produces a float that needs rounding (eg. `18.45042395591736`) whereas `maya.now() - start_time)` produces a `datetime.timedelta`.

derekpierre:
> Looks good! I'm wondering if we should turn some of our scripts into ape scripts, there may be benefits, like account management.

@cygnusv interested to hear more about what you are thinking - don't know if I can picture it just yet.

scripts/hooks/nucypher_dkg.py
manumonti: Thanks!

PR #3174, opened 2023-07-08
Set default ferveo variant to `simple` instead of `precomputed`

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Until we get better logic for handling any failures during `precomputed` variant, the `simple` variant should be our default.

**Issues fixed/closed:**
Related to #3143 .

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/newsfragments/3174.feature.rst
+++ b/newsfragments/3174.feature.rst
@@ -1,0 +1,1 @@
+Reset default ferveo variant to be ``simple``. We can revisit whenever we add logic to properly deal with request failures for the ``precomputed`` variant.
--- a/nucypher/characters/chaotic.py
+++ b/nucypher/characters/chaotic.py
@@ -199,8 +199,7 @@ class DKGOmniscientDecryptionClient(ThresholdDecryptionClient):
             ciphertext = decrypted_encryption_request.ciphertext
             conditions_bytes = str(decrypted_encryption_request.conditions).encode()
 
-            # Presuming simple for now.  Is this OK?
-            decryption_share = aggregate.create_decryption_share_precomputed(
+            decryption_share = aggregate.create_decryption_share_simple(
                 dkg=dkg,
                 ciphertext=ciphertext,
                 aad=conditions_bytes,
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -438,7 +438,7 @@ class Alice(Character, actors.PolicyAuthor):
 
 class Bob(Character):
     banner = BOB_BANNER
-    default_dkg_variant = FerveoVariant.PRECOMPUTED
+    default_dkg_variant = FerveoVariant.SIMPLE
     _default_crypto_powerups = [SigningPower, DecryptingPower]
     _threshold_decryption_client_class = ThresholdDecryptionClient
 
@@ -745,7 +745,7 @@ class Bob(Character):
         conditions: Lingo,
         context: Optional[dict] = None,
         ursulas: Optional[List["Ursula"]] = None,
-        variant: str = "precomputed",
+        variant: str = "simple",
         peering_timeout: int = 60,
     ) -> bytes:
         ritual = self.get_ritual_from_id(ritual_id)
@@ -1510,7 +1510,7 @@ class Enrico:
     """A data source that encrypts data for some policy's public key"""
 
     banner = ENRICO_BANNER
-    default_dkg_variant = FerveoVariant.PRECOMPUTED
+    default_dkg_variant = FerveoVariant.SIMPLE
 
     def __init__(self, encrypting_key: Union[PublicKey, DkgPublicKey]):
         self.signing_power = SigningPower()
--- a/nucypher/crypto/powers.py
+++ b/nucypher/crypto/powers.py
@@ -1,5 +1,5 @@
 import inspect
-from typing import List, Optional, Tuple
+from typing import List, Optional, Tuple, Union
 
 from eth_account._utils.signing import to_standard_signature_bytes
 from eth_typing.evm import ChecksumAddress
@@ -17,6 +17,7 @@ from nucypher_core import (
 from nucypher_core.ferveo import (
     AggregatedTranscript,
     Ciphertext,
+    DecryptionSharePrecomputed,
     DecryptionShareSimple,
     DkgPublicKey,
     Transcript,
@@ -276,7 +277,7 @@ class RitualisticPower(KeyPairBasedPower):
             ciphertext: Ciphertext,
             conditions: bytes,
             variant: FerveoVariant
-    ) -> DecryptionShareSimple:
+    ) -> Union[DecryptionShareSimple, DecryptionSharePrecomputed]:
         decryption_share = dkg.derive_decryption_share(
             ritual_id=ritual_id,
             me=Validator(address=checksum_address, public_key=self.keypair.pubkey),


piotr-roslaniec (approved):

KPrasch (changes requested):
I'd like to make a request for changes before merging this PR:  let's include a documentation update (either in this repo or in the threshold docs).  We are often getting terms confused, and it's not obvious to adopters why this functionality exists at all.

derekpierre:
> I'd like to make a request for changes before merging this PR: let's include a documentation update (either in this repo or in the threshold docs). We are often getting terms confused, and it's not obvious to adopters why this functionality exists at all.

There will be a follow-up to this PR to remove the ability to specify the variant to use for decryption - this was one of the Barcelona conversations that occurred. Ultimately the variant used will be controlled internally, and any fallback logic will still be controlled internally by our code - https://github.com/nucypher/nucypher/issues/3181.

Users should not be able to specify ferveo variant to use for decryption - so we don't need to document it.

KPrasch (dismissed):
> Users should not be able to specify ferveo variant to use for decryption - so we don't need to document it.

I don't agree with you here - even if the variant functionality is not exposed, Having a canonical reference for the meaning of simple vs. precomputed is ~~beneficial~~ necessary.

derekpierre:
> > Users should not be able to specify ferveo variant to use for decryption - so we don't need to document it.
> 
> I don't agree with you here - even if the variant functionality is not exposed, Having a canonical reference for the meaning of simple vs. precomputed is beneficial.

The question is who is it beneficial for? If it's something that is not exposed to users then no point putting it in user-facing docs. And, if you mean for devs then that still shouldn't be in user-facing docs.

If you had your druthers, where would you put the documentation? If you were thinking `ferveo` crate / python docs - that I agree with, but that would be a separate issue for the `ferveo`/`nucypher-core` repos.

KPrasch:
Personally I mixup "simple" vs. "precomputed" very frequently, and I'm not the only one.  

We're lacking in many forms of documentation right now so perhaps it's hard to invent a new place, but that's what's required here imo.  Nothing is private here, and it's not related to contribution, so it's not exactly development documentation, closer to architecture/system documentation, which may not be *user-facing per-se* but can certainly be public.  Personally I enjoy when documentation goes deep anyways, and this is a good of a place to start as any. 

Where do you think is the best place?

derekpierre:
> Where do you think is the best place?
Not sure about "best", but at the moment my view is that we are using `ferveo` to do this internally, so the documentation about `simple/precomputed` should be in `ferveo`. Similar to any other "external" library that we use internally.

> Can you remind me what the difference between the two is (and furthermore what performance implications it has)?

cc @cygnusv , @piotr-roslaniec to keep me honest:

Assume the threshold is 6-of-10:

### Simple

- you can receive **any** 6 shares back eg. you concurrently contact all 10 and once any 6 shares (from any of the 10) are  returned you are done
    
#### _PROs_

- any (arbitrary) 6 shares (out of 10)
- less of an availability issue (1 failed/downed node out of 10) doesn't cause a failure
- just one round needed if done concurrently (hit 10 nodes, wait for 6 replies)

#### _CONs_

- since the 6 nodes are any (non-specific), the decryption calculation requires Bob to do more work locally when combining returned decryption fragments.
   
### Precomputed

- first decide ahead of time on the 6 nodes you want to reach out to for the 6 shares
- if any of those specific nodes don't return a decryption frag (some form of failure/outage) - the overall decryption fails and this round fails
- If first round failed, you then need logic for additional failover rounds eg.:
      - you need to recreate a different list of 6 nodes that doesn't include that downed node and try contacting those 6 again (you have to contact all 6 again because the calc changes based on the composition of the 6 specific nodes)
        OR
      - you then switch to doing `simple`  
        OR
        ...
        
#### _PROs_

- in the happy path, only contact 6 nodes (no need for the other 4) 
- faster operation in the happy path - 1 round and the 6 nodes all respond.
- since the 6 nodes are known (specific 6 nodes), the decryption calculation requires Bob to do less work locally when combining returned decryption fragments

#### _CONs_

- availability issue - needs failover logic if 1 of the specific 6 nodes is down/fails. If the happy case doesn't happen it ends up potentially being an overall slower operation than for `simple`.

KPrasch:
> the documentation about simple/precomputed should be in `ferveo`.

Alright, let's put it there for now. 👍🏻

derekpierre:
> Also, thanks for the detailed reply, this is the best written explanation we have so far!

Credit to @cygnusv who had a chat with me about it in Barcelona.

KPrasch:
@derekpierre Let's move the above comment to the ferveo codebase for now?  I'd like to get this PR merged, but don't want to move forward without a documented reference for simple vs. precomputed.

derekpierre:
> @derekpierre Let's move the above comment to the ferveo codebase for now? I'd like to get this PR merged, but don't want to move forward without a documented reference for simple vs. precomputed.

I beg to differ - that documentation is separate from this issue and delaying this PR delays other work like removing the option for specifying the variant - #3181 .

I have filed https://github.com/nucypher/ferveo/issues/140, in `ferveo` which is marked for CBD 7.0.0 in `On Calendar` - that seems fair to me. 

Does that change your opinion?

KPrasch:
We have not been including documentation in PRs as a matter of best practice for quite some time which has led to quite a bit of technical deb in the form of missing documentation.  This is starting to create slowdowns and undue cognitive load (like the time before the knowledge base).  There have been many instances of confusion around these terms and we can solve that once and for all with a quick documentation commit.

I also want to unblock the upstream PR but the effort of committing your (excellent) comment above and put in in a markdown file somewhere so we have an official reference is worth waiting for compared to more moments of confusion.

derekpierre:
> We have not been including documentation in PRs as a matter of best practice for quite some time which has led to quite a bit of technical deb in the form of missing documentation. This is starting to create slowdowns and undue cognitive load (like the time before the knowledge base). There have been many instances of confusion around these terms and we can solve that once and for all with a quick documentation commit.
> 
> I also want to unblock the upstream PR but the effort of committing your (excellent) comment above and put it in a markdown file somewhere so we have an official reference is worth waiting for compared to more moments of confusion.

I agree with you about the documentation - but you are specifically saddling me with that debt. This PR undoes a previous code change, not done by me, so that we change the default back to `simple`.

If you want someone to own producing the simple/pre-computed documentation for v7.0.0, then assign the filed issue to me and/or someone else to take care of - so that someone is responsible for addressing your concerns (which I share). It's fine if it is assigned to me, but what we differ on is the timing of addressing that / making this PR directly dependent on that effort.

Any text will need to be reviewed by @cygnusv.

> I am happy to help with that, shall I make the appropriate commit/PR and link it here?

Thanks for the offer. If you want to start that effort, go for it 👍 .

KPrasch:
> I agree with you about the documentation - but you are specifically saddling me with that debt. This PR undoes a previous code change, not done by me, so that we change the default back to simple.

I am in no way targeting you with the responsibility of documenting all of ferveo, I'm taking advantage of the attention this PR is creating around the simple vs precomputed notion.  

The way to combat the accumulation of tech debt is to include related docs and tests in PRs as a best practice, it's not unrelated.  I was unable to review this PR without your above clarification, which I think is a pretty good signal that we need a reference.  Plus, your comment is 85% of what's needed in this moment, and I genuinely think we will benefit from a simple markdown reference while this iron is hot, and the thoughts are fresh.

derekpierre:
> I am struggling to understand the pushback.

The pushback is not on "producing" the documentation, I agree with that hence the issue being filed, linking that issue to this one, marked for CBD 7.0.0 On Calendar (so it's doesn't fall by the wayside),  and advocating for ownership (I've now assigned it to myself).

It's a pushback on the "timing". I can commit to producing that documentation in the next few days but I would rather not shift gears to writing that at this moment (my explanation above needs cleaning up and a review from David), and this `+8, -7` PR should not have to wait for that. Especially if we have a PR comment already in place to use for understanding this PR.

KPrasch:
I understand it's a small change - but the importance is not the low line count, it's that you are demonstrating an understanding of these concepts and think it's a convenient time to make a 10 line note about which is which.  This isn't an undue burden, and it's great (best) timing considering the ideas are fresh, especially on a topic that has been subject to confusion.  

I'm the guilty one here, I changed the default to precomputed because @piotr-roslaniec and I both confused it with simple. (This is the third PR switching variants now). We can easily avoid more with a simple reference.  

I'll accept that your above comment is cannon for now, but I hope requests for documentation in the future are better received.

derekpierre:
> it's a convenient time to make a 10 line note about which is which

I want to take my explanation in this PR as a base and produce something of reasonable quality, and not just a copy-paste job. If I just copy-paste, I won't look at it again. The explanation that I provided in this PR, hinged on a 6-of-10 example, but perhaps there can be an actual explanation and then an example to illustrate.

So I don't view it as a 5 mins. thing, but rather at the very least a one-hour thing. Perhaps the original `ferveo` (non-nucypher version) has a good explanation somewhere that we can leverage, perhaps there's a cleaner way to express the differences. 

> I'll accept that your above comment is cannon for now, but I hope requests for documentation in the future are better received.

I took your docs suggestion to heart, and I intend to do it properly, but properly means more time - that's all.

derekpierre:
Variant characteristics are now documented in the `ferveo` repos - https://github.com/nucypher/ferveo/blob/main/book/src/variants.md.

PR #3170, opened 2023-07-06
DKG ceremony initiation script improvements

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Allow dkg size to be specified
- Allow multiple rituals to be created via parameter.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #3160, opened 2023-06-27
Update `nucypher_dkg` hook/script to use separate provider uris for staking and coordinator

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
Updates the dkg script to use separate provider uris for eth staking network and coordinator network.

The script can be tested using existing dkg ritual id `2`.

**Issues fixed/closed:**

**Why it's needed:**

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #3182, opened 2023-07-11
Promote DKG scripting to a full-blown CLI command

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**

- introduces `nucypher dkg init` and `nucypher dkg check` CLI commands
- Recognizes the growth of the DKG initiation script as a formal CLI:
    - #3170 
    - #3169
    - #3160 
    - #3103 

**Why it's needed:**

- Dehydration and modularization of previous work on DKG scripting
- Easily status check a ritual's availability
- Ritual creation is still limited to a dev script and not as part of the public `nucypher` command (it may be added later).

**Notes for reviewers:**

This PR has been used on the lynx testnet to generate more then 15 `Ritual`s.


nucypher/cli/commands/dkg.py
```diff
@@ -0,0 +1,343 @@
...
+    "-n",
+    help="Number of nodes to participate in ritual",
+    type=click.INT,
+    default=0,
+)
+@click.option(
+    "--num-rituals",
+    "num_rituals",
+    "-r",
+    help="The number of rituals to initiate",
```
cygnusv, line 105: This option in particular doesn't make much sense for a `nucypher` command. `nucypher` is still a user-facing application (to node operators), and we're exposing here pretty much development-only functionalities

cygnusv: The script was fine, no? An alternative is the `hidden=True` kwarg for click options.

KPrasch: The script outgrew itself, it evolved into a full click application.  This merely formalizes it's usage.  If your understanding of the `nucypher` CLI is that it is *only* meant for node operators, then why have the `ursula` namespace at all?

KPrasch: My point is that the CLI can potentially be used for operations other than node operation, zooming out even more, this codebase is not just about nodes.  It also has a full-blown python API for TACo.

jMyles: > Why indeed!

Correct me if I'm misreading here, but it feels to me like you're saying, without actually saying, that you don't think that the other Characters and utilities warrant having CLI tooling built for them.

If that is your position, let's have a larger (but please, not protracted) discussion about that.  I think it's very obvious that this utility is asking to be made into a CLI-command, given its trajectory.  So if we want to force it in a different direction, it needs to be part of a larger notion of what this interface represents and to whom.

derekpierre: I must admit, initially I was also pro on moving the command as done in this PR, but I am reconsidering now, and I think @cygnusv 's argument is a good one.

While it is slightly awkward to have it in `scripts/` because it uses `click`, it was clear that it was dev functionality like relocking dependencies and the other scripts in that folder.

Moving the dkg script to a `nucypher` command means that this happens:
```bash
$ nucypher --help
Usage: nucypher [OPTIONS] COMMAND [ARGS]...

  Top level command for all things nucypher.

Options:
  --version       Echo the CLI version
  --config-path   Echo the configuration root directory path
  --logging-path  Echo the logging root directory path
  --help          Show this message and exit.

Commands:
  dkg     DKG management commands
  ursula  "Ursula the Untrusted" PRE Re-encryption node management commands.
  ```
which is strange because the `dkg` command is just for testing, and can lead to confusion.
  
If the use of `click` is specifically the main onus for moving, originally I had both the creation of rituals,and the encryption/decryption testing of existing rituals as one command and used   `sys.argv[<index>` which got complicated because the parameters and the relationships between them became too complex (see #3169). Hence the use of  `click` to make this simpler - but I probably should have just split the script into two separate scripts anyway: one for ritual creation, one for testing existing rituals. This is evidenced by the work in this PR to split that functionality into two different commands (`init`,`check` respectively), which is a good thing.

One option is to have two separate scripts in the `scripts/` folder, and therefore we may not even need to use `click` anyway since `sys.argv[<index>]` would be sufficient because the complexities of the parameters and their relationships would be reduced. Of course, we could still use `click` or we could have something like:
- `dkg_init <eth-provider> <coordinator-provider> <num_rituals> <dkg_size> <signer> [optional<eth-staking-network>]
- `dkg_check <eth-provider> <coordinator-provider> <ritual_id> [optional<eth-staking-network>]

**Comment by derekpierre** [comment]
> > the dkg command is just for testing, and can lead to confusion.
> 
> Is DKG ceremony initiation only used in tests? I don't understand this counter point - there are many instances that DKGs will be created in a production setting.

"testing" may not have been the correct terminology, but more like "dev tooling".

There is a stronger argument for the `nucypher dkg check` command; although the coordinator network is currently limited to `mumbai`, so maybe change that if we end up merging this PR. But I don't know about the `init` (in the short term). Perhaps you can add to the discussion here - https://github.com/nucypher/nucypher/issues/3186#issuecomment-1634329210 for the CLI.

I'm not saying it should never be in `nucypher` command group. My hesitation at least for the short term is having a public `nucypher dkg init` command that isn't permissionless and no one else can use but us, which sounds like dev tooling at that point. One advantage of using the `scripts/` folder is that I don't think it is included in released versions of `nucypher`.

When it becomes permissionless, I can see it possibly being added to the CLI. It's possible that an analogous front-end tool may allow for simpler UX / fee adjustments based on adopter options etc....kind of like a menu to choose add-on features from...maybe... 🤷 ... but doesn't mean that the CLI would not also be useful.

Perhaps others feel differently, which is fine.

**Comment by cygnusv** [comment]
> Is DKG ceremony initiation only used in tests?

Of course not. If you scroll back, you'll see that this discussion started with my comment to exposing `--num-rituals` as a CLI option. This absolutely has no sense in production, and that was my original concern (hence my suggestion to use the `hidden` flag). 

Now, the question about the overall role of the CLI is definitely a much more open-ended discussion, and I think both sides (CLI vs script) are reasonable. I can be fine with either, as long as there's some sort of usage separation between what's meant for external vs internal consumption.

> To be fair, I did indeed remove alice, bob, bond, status and enrico commands myself

That's a decision which I totally supported (and still support). To my recollection, the main reason was to get rid of code bloat and maintenance overhead, particularly since, from the existing three user facing interfaces (nucypher-ts, python API, CLI), the CLI was IMO the less useful and clunkier of all.

File: nucypher/cli/commands/dkg.py [hunk 2]
```diff
@@ -0,0 +1,343 @@
...
+        agent_class=PREApplicationAgent,
+        registry=staking_network_registry,
+        provider_uri=eth_provider_uri,
+    )  # type: PREApplicationAgent
+
+    return coordinator_agent, application_agent, staking_network_registry
+
+
+@click.group()
+def dkg():
```
cygnusv, line 76: I think in general we should about exposing the "DKG" term, which is very obscure. What do you think about `rituals`?

KPrasch:
> this discussion started with my comment to exposing --num-rituals as a CLI option. This absolutely has no sense in production

Can you help me understand this better? Why is it senseless to create more than one DKG at once in production?  I can imagine wanting to create a handful of DKG ceremonies that will be sold off simultaneously.

KPrasch:
> some sort of usage separation between what's meant for external vs internal consumption

Creation of DKG ceremonies will ultimately be a public-facing operation.  I'm struggling to understand how either of these commands are considered testing or dev tooling.  They are both user-facing (but permissioned) operations.  Am I missing something?

nucypher/cli/commands/dkg.py
```diff
@@ -0,0 +1,343 @@
...
+            option_name="--ritual-id, --signer",
+            message=click.style(
+                "--signer must be provided to create new ritual when --ritual-id is not provided",
+                fg="red",
+            ),
+        )
+    if dkg_size <= 1 or dkg_size % 2 != 0:
+        raise click.BadOptionUsage(
+            option_name="--dkg-size",
+            message=click.style("DKG size must be > 1 and a power of 2", fg="red"),
```
cygnusv, line 130: The power-of-2 restriction was removed in https://github.com/nucypher/ferveo/pull/134. Unsure about its current support in nucypher, but I think it'd be better to remove the restriction here.

cygnusv:
> Can you help me understand this better? It's not obvious to me why it has "absolutely no sense" to create more than one DKG at once in production? I can imagine wanting to create a handful of DKG ceremonies that will be sold off simultaneously.

The fact that we may want to create something between 1 to 10 cohorts for the permissioned private beta doesn't justify doing this in batch and offer this as a public tool. There's many factors that will make it much better for everyone to do cohort creating one by one. Things like different timelines (it's very likely not all adopters will be ready on day 1 after launch), different authorizers, or even different cohort sampling strategies.

> Creation of DKG ceremonies will ultimately be a public-facing operation. I'm struggling to understand how either of these commands are considered testing or dev tooling. They are both user-facing (but permissioned) operations. Am I missing something?

In the private beta/v7 phase, cohort creation is permissioned, which means there's no need to define user-facing tooling yet. So let's say it's just premature.

cygnusv:
> The fact that we may want to create something between 1 to 10 cohorts for the permissioned private beta doesn't justify doing this in batch and offer this as a public tool. There's many factors that will make it much better for everyone to do cohort creating one by one. Things like different timelines (it's very likely not all adopters will be ready on day 1 after launch), different authorizers, or even different cohort sampling strategies.

As @derekpierre pointed out earlier, this comment https://github.com/nucypher/nucypher/issues/3186#issuecomment-1634329210 is very illustrative on the rationale for not exposing batch ritual creation.

KPrasch:
> The fact that we may want to create something between 1 to 10 cohorts for the permissioned private beta doesn't justify doing this in batch and offer this as a public tool

I'm not justifying this feature set only for the private beta,  it can be a fully-featured and flexible way of making multiple parametrized cohorts, in production.

KPrasch:
We have decided not to merge this PR.

KPrasch:
Hmm, perhaps I've misunderstood.  In any case, I'd prefer to withdraw these changes now - we can reopen it at a later time 👍🏻

derekpierre:
Re-opening PR

derekpierre:
Changes made:
- Moved `dkg init` command to standalone script in `/scripts/hooks` folder.
- Renamed `dkg` command to `ritual` command.
- Updated allowed values for `--eth-staking-network` and `--coordinator-network` to include `mainnet`/`polygon` respectively.
- Removed power of 2 restriction for dkg size.

nucypher/cli/commands/dkg.py
derekpierre: ✔️

nucypher/cli/commands/dkg.py
derekpierre: ✔️ - went with `ritual`.

KPrasch:
>  Moved `dkg init` command to standalone script in `/scripts/hooks` folder.

This is an awkward and confusing division of logic.  Before this PR was reopened, both init and check were presented as sister commands. Without them both here I think it's best we leave check in `scripts` as well (or potentially move check back to scripts and leave `init` as part of the CLI).  

We may have opposing views here, but `init` was the main functionality to expose - `check` is a bonus (which is really *decrypt*, and not *check*). Without `init` the "umph" has been taken out of the PR. Ultimately this puts the check command into a isolated/detached CLI in an unintuitive way.

> Renamed `dkg` command to `ritual` command.

This is more obscure than the `dkg` namespace IMO.

I appreciate the effort to make a compromise here, but in my opinion the outcome is strictly worse. Overall, I'd prefer to close this PR in favor of the `scripts` approach unless we decide to expose both the dkg initialization, status, and decryption functionality in a cohesive manner.

derekpierre:
Ok. Onward and upward.

PR #3183, opened 2023-07-12
Node DKG Ritual event handling robustness

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Ideally, the handling of Ritual events during ritual initiation to be more robust:
- Properly ignore/filter away ritual events for rituals not relevant to the node
- Easily rectify/continue its participation in rituals in case it restarts while a ritual it is participating in is still going on
- Use Coordinator events to gleam node's participation in ritual, and what the node has/has not done already for a ritual.
- If unsure of state, use the Coordinator contract to update what its view of its state is.
- Don't block waiting for a tx to post transcript/aggregated transcript to be fully completed before being able to continue processing events - use `fire_and_forget`, store the tx hash, and keep it moving. 
- If the node already has a pending tx for posting a transcript/aggregated transcript do *not* submit another tx for the same action.

This PR attempts to achieve that. 

This PR does not address speeding up already submitted txs for post transcript/aggregation in the case of a gas spike that may last. There may be additional functionality we can do for replacement transactions (slow txs when posting transcripts/aggregations), adjusting gas etc. but if necessary that should be addressed in a separate PR. There may also be tweaks we can make to the event scanner. These need more mindshare before tackling. (cc @KPrasch , @cygnusv ).

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
During ETH Barcelona we initiated 15 rituals one after the other, and the nodes showed a significant decrease in performance for a variety of reasons:
- Incorrect filtering of events
- Lots of exception handling in twisted threads (cost of handling errors then executing again)
- Repeated events
- Blocking while waiting for txs to complete.

**Notes for reviewers:**

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -396,7 +396,8 @@ class Ritualist(BaseActor):
         receipt = self.coordinator_agent.post_transcript(
             ritual_id=ritual_id,
             transcript=transcript,
-            transacting_power=self.transacting_power
+            transacting_power=self.transacting_power,
+            fire_and_forget=True,
         )
         return receipt
 
@@ -416,7 +417,8 @@ class Ritualist(BaseActor):
             aggregated_transcript=aggregated_transcript,
             public_key=public_key,
             participant_public_key=participant_public_key,
-            transacting_power=self.transacting_power
+            transacting_power=self.transacting_power,
+            fire_and_forget=True,
         )
         return receipt
 
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -339,13 +339,6 @@ class Ritualist(BaseActor):
             ThresholdRequestDecryptingPower
         )  # used for secure decryption request channel
 
-    def get_ritual(self, ritual_id: int) -> CoordinatorAgent.Ritual:
-        try:
-            ritual = self.ritual_tracker.rituals[ritual_id]
-        except KeyError:
-            raise self.ActorError(f"{ritual_id} is not in the local cache")
-        return ritual
-
     def _resolve_validators(
             self,
             ritual: CoordinatorAgent.Ritual,
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -423,7 +423,6 @@ class Ritualist(BaseActor):
         timestamp: int,
     ):
         """Perform round 1 of the DKG protocol for a given ritual ID on this node."""
-
         if self.checksum_address not in participants:
             # should never get here
             self.log.error(
@@ -437,27 +436,24 @@ class Ritualist(BaseActor):
         ritual = self.coordinator_agent.get_ritual(ritual_id, with_participants=True)
         status = self.coordinator_agent.get_ritual_status(ritual_id=ritual_id)
 
-        # FIXME: Rearrange checks - for the moment, obtain more information for logging purposes
+        # validate the status
+        if status != CoordinatorAgent.Ritual.Status.AWAITING_TRANSCRIPTS:
+            self.log.debug(
+                f"ritual #{ritual_id} is not waiting for transcripts; status={status}; skipping execution"
+            )
+            return
+
         # validate the active ritual tracker state
         participant = self.coordinator_agent.get_participant_from_provider(
             ritual_id=ritual_id, provider=self.checksum_address
         )
-
-        self.log.debug(
-            f"Ritual {ritual_id} with status {status}, timestamp {ritual.init_timestamp}. "
-            f"Stored transcript data is '{participant.transcript.hex()}'"
-        )
-
-        # validate the status
-        if status != CoordinatorAgent.Ritual.Status.AWAITING_TRANSCRIPTS:
-            raise self.RitualError(
-                f"ritual #{ritual_id} is not waiting for transcripts; status={status}."
-            )
-
         if participant.transcript:
-            raise self.RitualError(
-                f"Node {self.transacting_power.account} has already posted a transcript for ritual {ritual_id}"
+            self.log.debug(
+                f"Node {self.transacting_power.account} has already posted a transcript for ritual {ritual_id}; skipping execution"
             )
+            return
+
+        # TODO - process pending receipts before submitting a new tx (perhaps tx already submitted, but not finalized
         self.log.debug(f"performing round 1 of DKG ritual #{ritual_id} from blocktime {timestamp}")
 
         # gather the cohort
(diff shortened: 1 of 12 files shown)


derekpierre:
Moving this change back to draft status - need to rethink something. Skip on reviews for now.

theref (approved):
This is an excellent piece of work 🚀

manumonti (approved):
Wow, such a great work!

Tests are really thorough 👏

LGTM ✌️

I left a comment of no importance

nucypher/datastore/dkg.py
```diff
@@ -13,31 +15,41 @@ def __init__(self):
...
+        self, ritual_id: int, txhash_or_receipt: Union[TxReceipt, HexBytes]
+    ) -> None:
+        self.data["transcript_receipts"][ritual_id] = txhash_or_receipt
 
-    def get_transcript_receipt(self, ritual_id: int) -> TxReceipt:
-        return self.data["transcript_receipts"][ritual_id]
+    def get_transcript_receipt(
+        self, ritual_id: int
+    ) -> Optional[Union[TxReceipt, HexBytes]]:
+        return self.data["transcript_receipts"].get(ritual_id)
```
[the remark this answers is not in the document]
derekpierre: We use `.get()` because if the entry for that `ritual_id` is not present, we want it to return `None` and not raise a `KeyError`.

Note the typehint for what the method returns:
`Optional[Union[TxReceipt, HexBytes]]` which effectively means the method can return either `TxReceipt` OR `HexBytes` OR `None`.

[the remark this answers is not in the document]
derekpierre: > Should we use .get() for the rest of calls to ["transcript_receipts"][ritual_id] to add homogeneity to the code? (like in lines 19, 34, 44, and so on)

You are absolutely right - I'll take care of that. Nice catch!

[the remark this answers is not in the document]
derekpierre: ✔️

nucypher/blockchain/eth/actors.py
```diff
@@ -390,45 +382,46 @@ def _resolve_validators(
...
         )
-        receipt = self.coordinator_agent.post_aggregation(
+        tx_hash = self.coordinator_agent.post_aggregation(
             ritual_id=ritual_id,
             aggregated_transcript=aggregated_transcript,
             public_key=public_key,
             participant_public_key=participant_public_key,
-            transacting_power=self.transacting_power
+            transacting_power=self.transacting_power,
+            fire_and_forget=True,
```
piotr-roslaniec, line 413: I initially didn't get what `fire_and_forget` does, I call this mode "block until complete",

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -185,57 +183,191 @@ def stop(self):
         """Stop the event scanner task."""
         return self.task.stop()
 
-    def __action_required(self, event_type: Type[ContractEvent], block_number: int, ritual_id: int):
-        """Check if an action is required for a given event."""
-        if (event_type, ritual_id) in self.active_tasks:
-            self.log.debug(
-                f"Already tracking {event_type} for ritual {ritual_id} from block #{block_number}"
-            )
+    def _action_required(self, ritual_event: AttributeDict) -> bool:
```
piotr-roslaniec, line 186: We could intuit there is some sort of state machine here

piotr-roslaniec (approved):

nucypher/blockchain/eth/actors.py
cygnusv: I guess you mean block until complete is when fire and forget is false, right? Because fire and forget = true is the opposite of blocking

jMyles:
Without delving into the substance (for which I might still like a walkthrough), I notice that, when annotating the changes, a lot of the lines have as their most recent commit things like, "code cleanup" and "I want to kill the linter but I'm not saying so in a commit message".  

When practical to do so, I'd love to see these changes amended into the commits where the issue was introduced, so that reading 'blame style' gives a clearer indication of the intention of each line change.

(Another way of reiterating my snark in Barcelona on this topic :-) )

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -185,57 +183,191 @@ def stop(self):
...
+        # establish participation state first
+        participation_state = self._get_participation_state(ritual_event)
+
+        if not participation_state.participating:
             return False
+
+        # does event have an associated action
+        event_type = getattr(self.contract.events, ritual_event.event)
+
+        event_has_associated_action = event_type in self.actions
```
jMyles, line 197: Nice.

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -185,57 +183,191 @@ def stop(self):
...
-        name, args = event.event, event.args
+    def _get_ritual_participant_info(
+        self, ritual_id: int
+    ) -> Optional[CoordinatorAgent.Ritual.Participant]:
+        """
+        Returns node's participant information for the provided
+        ritual id; None if node is not participating in the ritual
+        """
+        participants = self.coordinator_agent.get_participants(ritual_id=ritual_id)
+        for p in participants:
```
jMyles, line 225: I'm just thankful you didn't try this in a comprehension ;-)

manumonti:
> Without delving into the substance (for which I might still like a walkthrough), I notice that, when annotating the changes, a lot of the lines have as their most recent commit things like, "code cleanup" and "I want to kill the linter but I'm not saying so in a commit message".
> 
> When practical to do so, I'd love to see these changes amended into the commits where the issue was introduced, so that reading 'blame style' gives a clearer indication of the intention of each line change.
> 
> (Another way of reiterating my snark in Barcelona on this topic :-) )

This is an interesting discussion. Another option could be to squash all the commits in a few ones with meaningful commit messages.

jMyles:
Yeah, more of a general critique - not specific to this PR.  But in the case of this PR, I'm talking about commits like:

* "Resolve darker complaints."
* "Fix Linter Issues"
* "Code cleanup for testing. "

Even though none of them are monsters, it's still easier (at least for me) to grok the history if we try to ensure that the annotations show the last logical (rather than syntactic) motivation.  Obviously that's not always possible, and even when it is, sometimes impractical because it requires long delves into interactive rebase, etc.

So yeah, not a huge complaint.  More of a soft directional reorientation.

As for squashing: I'm broadly cautious of squashing (though I do like @cygnusv's technique of adding amendments via rebase and then autosquashing `fixup` commits).   Especially if I'm catching up on a module I haven't seen in a while, it's great to see a more granular history of the changes to the line I'm investigating.  But yeah, sometimes it's good.

jMyles (approved):
Looks good to me, though I do wish things had lined up for a walkthrough.  Soon.

This is really a lot of fun code!

derekpierre:
> Looks good to me, though I do wish things had lined up for a walkthrough. Soon.
> 
> This is really a lot of fun code!

If you liked this one check out #3191  😄 .

PR #3043, opened 2022-12-24
Unfederate Configurations

KPrasch:
**Type of PR:**
Bugfix

**Required reviews:** 
1

**What this does:**
- fixes two bugs introduced in #3030 by:
  - bumps the configuration version to follow-up #3030 's changed to ursula config
  - use the operator address to identify and name ursula config files (not the "checksum address")
  
**Note**
- Staged on lynx testnet @ 1b277651a1cdff2fc0d4c6032cbf98b0d6cc7cf9

PR #3185, opened 2023-07-12
[WIP] Multichain Condition Evaluation

KPrasch:
**Type of PR:**
- Feature

**Required reviews:** 
3

**What this does:**

**Issues fixed/closed:**

**Why it's needed:**

**Notes for reviewers:**


nucypher/policy/conditions/evm.py
```diff
@@ -19,13 +23,27 @@
...
+    10: "optimism/mainnet",
+    69: "optimism/kovan",
+    420: "optimism/goerli",
+    42161: "arbitrum/mainnet",
+    421613: "arbitrum/goerli",
+    100: "gnosis/mainnet",
+    10200: "gnosis/chiado",
+    534352: "scroll/mainnet",
+    534351: "scroll/sepolia",
+}
```
cygnusv, line 46: While I totally agree with the direction of the work on this PR, I want to be cautious on this specific set of lines. Let's validate them before actually adding them to the allowlist. We need criteria to evaluate both:
(1) Demand-side: To have evidence adopters requesting some chain are "sponge-worthy" (i.e. not because someone claiming to use TACO needs some chain, we're going to add it). 
(2) Supply-side: make sure that stakers/node operators can handle the overhead of adding more chains to the list. In some cases this can be supported by "free plans" (e.g., from this list, Infura supports free add-ons for polygon, optimism and arbitrum, although threre's the lingering question about how much can we lean on Infura), but what about Gnosis and Scroll, for instance?

cygnusv: Original issue tracking this:
https://github.com/nucypher/tdec/issues/7

jMyles: Since these are for reads, I very much prefer for the default path not to be Infura / another centralized solution.  We can provide RPC endpoints as defaults.  If somebody wants to use their Infura, we can make that easy too.

But this is a _very_ easy place to provide a default that's more decentralized.

As for the specifics: Gnosis and Scroll seems like obvious defaults.  Gnosis is the minting ground for POAP, which I think we want to support out of the gate.  Scroll is a ZK insurgent which Electora used successfully in their hackathon win.

And although the loss of the Today Sponge is a tragedy, I don't think we're at risk of running out of multichain RPC endpoints.

It's more like "Top of the Multichain to you!"

(with the exclamation point)

KPrasch: I'm suggesting gnosis and scroll be allowed here since there are adopters actively using both, fortunately there are several officially maintained RPC endpoints for both networks.  An infura integration can ease some concerns here, but I'd also like to suggest we include public endpoint lists maintained by third parties like chainlist.network.

if there's more validation to do before adding to this list we can remove it, however I'd like to deploy a gnosis-supporting testnet asap for use with the revealer adopter asap.

KPrasch: A bigger issue this illuminates is if we'll require all nodes to support all endpoints or not.  If nodes can select only a subset, we will need to build out an endpoint for use during sampling to query which nides support which chains... Which itself raises questions about economic schemes, etc.

cygnusv: > A bigger issue this illuminates is if we'll require all nodes to support all endpoints or not. If nodes can select only a subset, we will need to build out an endpoint for use during sampling to query which nides support which chains... Which itself raises questions about economic schemes, etc.

No, I don't think we want pick-and-choose endpoints for v7. As for the future, we we would have to solve first the questions you mention. But it's definitely something worth exploring and I think we already mention this in some issue (I'll try to find it later).

I suggest that for the moment let's focus here on those chains that will be 100% required for all nodes, and that should satisfy both demand- and supply-side requirements. We know it is the case for Ethereum and Polygon. Is this the case for Gnosis chain and Scroll? Just to be clear, I'm not asking this rhetorically: I genuinely would like to see what's the evidence supporting both requirements.

> @jmyles: [...] But this is a very easy place to provide a default that's more decentralized.
> @KPrasch: [...] fortunately there are several officially maintained RPC endpoints for both networks.

These two comments are IMO contradictory.

jMyles: Yeah, I'm not sure how to think about "optional" chains.  Also, since nodes only stake on Ethereum mainnet, it's not immediately obvious how we'd slash for bad behavior in the future on other chains (though of course there are ways, just not immediately obvious ones).

That said, this list seems like a good list of defaults.

I think @KPrasch was thinking that my record release was using conditions on Gnosis when he wrote the above, but that's not so - only Ethereum mainnet (though it depends on merge of this logic just the same!).  

However, I can think of an imminent use for Gnosis: using POAPs at events where we're presenting for the purposes of live demos.  In fact, I can imagine doing just this at the Encode hackathon in a couple days.

theref:
@KPrasch would love to help out on this. So far I can see you're sorting EVM conditions, do you think this would be a good time to take a look at non evm (solana) conditions too? Requirements are very similar, put a request in to a json rpc

Or is that too much work? I guess there would be changes needed to the condition spec to handle solana contracts

nucypher/blockchain/eth/actors.py
```diff
@@ -336,6 +320,54 @@ def __init__(
...
+                # Prevents the Ursula/Ritualist from starting up if this happens.
+                raise NotImplementedError(
+                    f"Chain ID {chain_id} is not supported by the Ritualist."
+                )
+
+            providers = list()
+            for uri in condition_provider_uris:
+                provider = HTTPProvider(endpoint_uri=uri)
+                providers.append(provider)
+            condition_providers[chain_id].update(providers)
```
derekpierre, line 359: What if the payment_method and the condition_provider_uris used/had different endpoint URIs for the same chain? Des one get overwritten by the other? Not that this would typically happen of course 😅 .

Kind of related to my earlier comment in this file.

nucypher/blockchain/eth/actors.py
```diff
@@ -336,6 +320,54 @@ def __init__(
...
+
+        # Log the chains that the Ritualist is connected to.
+        humanized_chain_ids = ", ".join(
+            _CONDITION_CHAINS[chain_id] for chain_id in condition_providers
+        )
+        self.log.info(
+            f"Connected to {len(condition_providers)} blockchains: {humanized_chain_ids}"
```
derekpierre, line 366:
```suggestion
            f"Connected to {len(condition_providers)} blockchains for condition checking: {humanized_chain_ids}"
```

nucypher/config/base.py
```diff
@@ -304,7 +304,7 @@ class CharacterConfiguration(BaseConfiguration):
     'Sideways Engagement' of Character classes; a reflection of input parameters.
     """
 
-    VERSION = 5  # bump when static payload scheme changes
+    VERSION = 6  # bump when static payload scheme changes
```
derekpierre, line 307: Our current publicly released configuration version is 4 - https://github.com/nucypher/nucypher/blob/main/nucypher/config/base.py#L325C59-L325C59.
So:
1. Do we need to go from 5->6 or can we just stay at 5? 5 was only an internal unreleased dev change (https://github.com/nucypher/nucypher/pull/3043). Looks like we already have a script for v4 -> v5 (https://github.com/nucypher/nucypher/commit/2d0f02a287277171bb68c508d8cb43e5f6b72f06). Can that be piggy-backed upon?
2. Do we need a related migration script for Ursula as part of this change? Whatever version we end up using (5 or 6), we'll need a migration script from v4.

nucypher/policy/conditions/evm.py
```diff
@@ -187,19 +219,30 @@ def _execute_call(self, parameters: List[Any]) -> Any:
...
-        self._configure_provider(providers=providers)
-        parameters, return_value_test = _resolve_any_context_variables(
-            self.parameters, self.return_value_test, **context
-        )
-        try:
-            result = self._execute_call(parameters=parameters)
-        except Exception as e:
-            raise RPCExecutionFailed(f"Contract call '{self.method}' failed: {e}")
+        for provider in self._next_endpoint(providers=providers):
+            self._configure_provider(provider=provider)
```
derekpierre, line 230: Should we put the `_configure_provider()` call into `_next_endpoint()`, or is there another reason to keep it out of there? Seems like you'll always need the provider to be configured before using.

nucypher/blockchain/eth/actors.py
```diff
@@ -300,10 +283,11 @@ class RitualError(BaseActor.ActorError):
 
     def __init__(
         self,
-        provider_uri: str,  # this is a blockchain connection to the chain with the coordinator contract
+        coordinator_provider_uri: str,
         network: str,  # this must be the network where the coordinator lives
         crypto_power: CryptoPower,
         transacting_power: TransactingPower,
+        condition_provider_uris: Optional[Dict[int, List[str]]] = None,
```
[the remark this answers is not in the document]
jMyles: Can you help me understand your thinking here?

It seems to me that this might plausibly be a frequent occurrence - is there something special that we need to consider about it?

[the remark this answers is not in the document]
derekpierre: I assume it's ok, but I'm just confirming that if we have two separate URIs for the same chain (one for coordinator, and another for conditions), which is perfectly fine, which means that we create separate connections to the same chain, that we are still good. If we are, then all good.

Related to my question below:
> What if the payment_method and the condition_provider_uris used/had different endpoint URIs for the same chain?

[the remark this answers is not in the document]
jMyles: I'm not sure how these connections are handled under the hood by web3.  

I surmise it won't be an issue, though.

Since we index by chainID, we uhh... most likely won't have different uris being passed anyway, except when it's overridden?

KPrasch:
I still have at least one incoming commit for this PR, but I'd like to mark it as RFR now!

piotr-roslaniec (approved):

nucypher/policy/conditions/evm.py
KPrasch: ✔️

nucypher/config/base.py
KPrasch: I don't mind bumping the version here again as a safety measure, but I do agree that a migration script may be useful.  I'll add one.

nucypher/blockchain/eth/actors.py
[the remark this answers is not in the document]
KPrasch: No issue with using the same endpoint for both, and indeed they are deliberately separable.  This decoupling may be desirable if nodes will use a separate endpoint for transacting (dkg participation) vs. read-only (condition evaluation), particularly if the node is configured to use public RPC endpoints for conditions.

nucypher/config/base.py
KPrasch: ✔️

nucypher/blockchain/eth/actors.py
KPrasch: Nope they will not be overwritten, just maintained as separate endpoints.

manumonti (approved):
Great work 👏! LGTM 🙌

jMyles (approved):
Since I worked on many aspects of this PR, I'm not sure how much weight is properly attributed to my approval, but I give it nonetheless.

PR #3193, opened 2023-07-21
Don't expose Ferveo variant in python Bob public API

### LINKED ISSUE #3181: Remove the ability for users to specify Ferveo Variant
Users should not be able to specify the ferveo variant to use for threshold decryption. It is not worth the complexity and users will not understand the tradeoffs. Instead, we will control the variant used internally in the code and remove any parameterization.  Any fallback logic for switching ferveo variants will always ultimately be controlled internally by the code.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Remove FerveoVariant parameter from public python API for threshold decryption.

**Issues fixed/closed:**
Fixes #3181 .

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/crypto/ferveo/dkg.py
+++ b/nucypher/crypto/ferveo/dkg.py
@@ -78,6 +78,7 @@ def verify_aggregate(
 ):
     pvss_aggregated.verify(shares, transcripts)
 
+
 def derive_decryption_share(
     nodes: List[Validator],
     aggregated_transcript: AggregatedTranscript,
@@ -93,7 +94,7 @@ def derive_decryption_share(
     try:
         derive_share = _VARIANTS[variant]
     except KeyError:
-        raise Exception(f"invalid variant {variant}")
+        raise ValueError(f"Invalid variant {variant}")
     share = derive_share(
         # first arg here is intended to be "self" since the method is unbound
         aggregated_transcript,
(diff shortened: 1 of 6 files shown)


nucypher/crypto/ferveo/dkg.py
```diff
@@ -93,7 +94,7 @@ def derive_decryption_share(
     try:
         derive_share = _VARIANTS[variant]
     except KeyError:
-        raise Exception(f"invalid variant {variant}")
+        raise ValueError(f"Invalid variant {variant}")
```
piotr-roslaniec, line 97: Just a thought: I've noticed that there is a similar error message being thrown by `ferveo` Python bindings. I don't think it's that confusing, but if someone we're to search for `Invalid variant` in `nucypher` this is where they would end up, not in `ferveo` or `nucypher` core.

piotr-roslaniec (approved):

tests/integration/config/test_keystore_integration.py
```diff
@@ -191,7 +192,7 @@ def test_ritualist(temp_dir_path, testerchain, dkg_public_key):
     ciphertext = enrico.encrypt_for_dkg(plaintext=plaintext, conditions=CONDITIONS)
     decryption_request = ThresholdDecryptionRequest(
         ritual_id=ritual_id,
-        variant=0,
+        variant=FerveoVariant.SIMPLE.value,
```
cygnusv, line 195: Why is `.value` required here?

cygnusv (approved):

PR #3171, opened 2023-07-07
Tracking changes for `nucypher-core@0.11.0` update

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:** 
- 2

**What this does:**
- Ensures compatibility with an upcoming `nucypher-core@0.11.0` release

**Issues fixed/closed:**
- Related to: https://github.com/nucypher/nucypher-core/pull/69

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
- Awaits release of `nucypher-core@0.11.0` and replacing the git dependencies to an actual Python package

PR #3196, opened 2023-08-02
Replace git dependency with `nucypher-core@0.11.0` package

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:** 
- 1

**What this does:**
- Follow-up to https://github.com/nucypher/nucypher/pull/3171


derekpierre (approved):
🎸.

Let's try not to have git commit dependencies merged into `development` which then need to be updated to actual versions via another PR. The original code should not be merged before actual versions are used.

PR #3205, opened 2023-08-24
Multichain connection broadcasting endpoint for Ursula

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
- Defines a new JSON endpoint (GET @ `/condition_chains`) that returns a list of all chain IDs Ursula is connected to.
```python
{"version": 1.0, "evm": [1, 5, 80001]}
```
- Multichain fixture and utilities cleanup

**Why it's needed:**
- Assistance in evaluating network health and node operator compliance with multichain connection requirements

**Notes for reviewers:**
- The response is versioned so we can expand the number fields in a non-evm future feature
- Co-Authored by @theref

### DIFF
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -135,6 +135,18 @@ def _make_rest_app(this_node, log: Logger) -> Flask:
         # TODO: What's the right status code here?  202?  Different if we already knew about the node(s)?
         return all_known_nodes()
 
+    @rest_app.route("/condition_chains", methods=["GET"])
+    def condition_chains():
+        """
+        An endpoint that returns the condition evaluation blockchains
+        this node has connected to.
+        """
+        # TODO: When non-evm chains are supported, bump the version.
+        #  this can return a list of chain names or other verifiable identifiers.
+
+        payload = {"version": 1.0, "evm": list(this_node.condition_providers)}
+        return Response(json.dumps(payload), mimetype="application/json")
+
     @rest_app.route('/decrypt', methods=["POST"])
     def threshold_decrypt():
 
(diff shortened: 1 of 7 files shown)


theref (approved):
Looks good, i'm excited to see those chain ids on the lynx prometheus metrics soon

piotr-roslaniec (approved):
I can see there are relevant fixtures being added, etc., but would it be interesting to also have an acceptance test for the actual endpoint?

KPrasch:
> I can see there are relevant fixtures being added, etc., but would it be interesting to also have an acceptance test for the actual endpoint?

There is an integration test that sends a "live" request to the new endpoint included in this PR here: <https://github.com/nucypher/nucypher/blob/f44bccf55443d47c4dc860e5fffd1142fe13b7be/tests/integration/network/test_condition_blockchain_broadcast.py>

I believe this is sufficient because we don't need a running local blockchain to test the network endpoint.

The code/fixture reorg is a bit distracting and noisy, but it was required handle our fixtures layout in a way that appeases both ape and linting. :-)

nucypher/network/server.py
```diff
@@ -135,6 +135,18 @@ def node_metadata_exchange():
...
+    @rest_app.route("/condition_chains", methods=["GET"])
+    def condition_chains():
+        """
+        An endpoint that returns the condition evaluation blockchains
+        this node has connected to.
+        """
+        # TODO: When non-evm chains are supported, bump the version.
+        #  this can return a list of chain names or other verifiable identifiers.
+
+        payload = {"version": 1.0, "evm": list(this_node.condition_providers)}
```
derekpierre, line 147: Should the version value be a string?

PR #3208, opened 2023-08-25
Bump CI's Maximum Python Version to 3.11

### LINKED ISSUE #3184: Deprecate pysha3 once and for all
It appears that we have one remaining dependency on PySha3: `eip712-structs`.

We use this to make format the address that gets EIP712'd as evidence. 

Perhaps the easiest path is to adopt this project and swap in `hashlib`.

Note: this is one of the things - perhaps the only thing - holding us back from python 3.11 compatibility.

jMyles:
##### Requested Reviews 
1

##### What this does
- Indicates compatibility with python 3.11 in package metadata
- Bumps the upper limit bookend python version on CI to python 3.11
- Run Codecov alongside 3.11 test execution

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ "3.8", "3.10", "3.11" ]
+        python-version: [ "3.8", "3.11" ]
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -17,7 +17,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ "3.8", "3.10" ]
+        python-version: [ "3.8", "3.10", "3.11" ]
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -53,94 +53,94 @@ jobs:
 
       # Unit tests
       - name: Unit Tests (Coverage)
-        if: matrix.python-version == '3.10'
+        if: matrix.python-version == '3.11'
         run: |
           coverage run --data-file=unit_data -m pytest tests/unit
           coverage xml -i --data-file=unit_data -o unit-coverage.xml
 
       - name: Unit Tests
-        if: matrix.python-version != '3.10'
+        if: matrix.python-version != '3.11'
         run: python -m pytest tests/unit
 
       # Integration tests
       - name: Integration Tests (Coverage)
-        if: matrix.python-version == '3.10'
+        if: matrix.python-version == '3.11'
         run: |
           coverage run --data-file=integration_data -m pytest tests/integration
           coverage xml -i --data-file=integration_data -o integration-coverage.xml
 
       - name: Integration Tests
-        if: matrix.python-version != '3.10'
+        if: matrix.python-version != '3.11'
         run: python -m pytest tests/integration
 
       # Acceptance tests
       - name: Agents Tests (Coverage)
-        if: matrix.python-version == '3.10'
+        if: matrix.python-version == '3.11'
         working-directory: tests/acceptance
         run: |
           coverage run --data-file=acceptance_agent_data -m pytest agents
           coverage xml -i --data-file=acceptance_agent_data -o acceptance-agents-coverage.xml
 
       - name: Agents Tests
-        if: matrix.python-version != '3.10'
+        if: matrix.python-version != '3.11'
         working-directory: tests/acceptance
         run: python -m pytest agents
 
       - name: Actors Tests (Coverage)
-        if: matrix.python-version == '3.10'
+        if: matrix.python-version == '3.11'
         working-directory: tests/acceptance
         run: |
           coverage run --data-file=acceptance_actors_data -m pytest actors
           coverage xml -i --data-file=acceptance_actors_data -o acceptance-actors-coverage.xml
 
       - name: Actors Tests
-        if: matrix.python-version != '3.10'
+        if: matrix.python-version != '3.11'
         working-directory: tests/acceptance
         run: python -m pytest actors
 
 
       - name: Conditions Tests (Coverage)
-        if: matrix.python-version == '3.10'
+        if: matrix.python-version == '3.11'
         working-directory: tests/acceptance
         run: |
           coverage run --data-file=acceptance_conditions_data -m pytest conditions
           coverage xml -i --data-file=acceptance_conditions_data -o acceptance-conditions-coverage.xml
 
       - name: Conditions Tests
-        if: matrix.python-version != '3.10'
+        if: matrix.python-version != '3.11'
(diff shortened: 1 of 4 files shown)


KPrasch:
Is this PR RFR?

derekpierre:
#3222 removed the dependency on pysha3.

KPrasch:
> #3222 removed the dependency on pysha3.

Shall we close this PR then?

derekpierre:
> Shall we close this PR then?

I think @jMyles had some hopes about python 3.11 as part of this PR. He can better answer that question.

At the very least some of the PR is no longer needed.

derekpierre:
Few things:

1. Python 3.8 eol seems to be October, 2024, not October 2023 - https://devguide.python.org/versions/#supported-versions. If that's correct then let's not drop 3.8 support.

2. If we are including python 3.11 as supported, then let's update our github workflow to continue to bookend our python supported versions. Right now it tests using `[ "3.8", "3.10" ]` so let's change that to `[ "3.8", "3.11" ]` - https://github.com/nucypher/nucypher/blob/development/.github/workflows/python_tests.yml#L19C25-L19C42

3. Given 2) above, we need to update our code coverage in the python-tests workflow. Code coverage tests only run for "3.10", so we'll need to update those checks to be for "3.11" - https://github.com/nucypher/nucypher/blob/development/.github/workflows/python_tests.yml#L55 - there are multiple of these "if" checks to update from "3.10" to "3.11" (perhaps there was an easier way to achieve this as part of the workflow instead of having multiple of these "if" checks... 🤷 ).

KPrasch:
@jMyles Do you intend to complete this PR? The GH workflow still needs updating.

jMyles:
Oh, right you are!

Stand by.

jMyles:
>  there are multiple of these "if" checks to update from "3.10" to "3.11"

I'm a little puzzled on why we have those at all - is the intention to only run those steps (which represent all the test runs, right?) for a single version, even though we name two (now three) in the matrix spec?

jMyles:
Also, what of 3.9?  It is omitted in the actions workflow - is there a reason for that?

derekpierre:
We basically bookend the python versions supported for efficiency. The more versions we add to the matrix the longer CI builds take to run since it runs tests for each version. 

So instead we simply run tests on the min supported python version, and max python supported version.

Codecov tests are slower than non-codecov tests, and only need to be run for one python version (it's the same coverage report anyway 😄 ), so we only pick one version to run tests with codecov.

jMyles:
Ahh, it's just codecov that's per-version!  Got it.

jMyles:
So, I don't see a particular reason to fiddle with the version that runs codecov if it ain't broken.

I think we can call this RFR now (just pushed a rebase with changes from the above RFCs).

.github/workflows/python_tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ "3.8", "3.10" ]
+        python-version: [ "3.8", "3.10", "3.11" ]
```
derekpierre, line 19: > So, I don't see a particular reason to fiddle with the version that runs codecov if it ain't broken.

The reason to fiddle with codecov, is that this should just be:

```suggestion
        python-version: [ "3.8", "3.11" ]
```

to bookend our python support, since running with 3.10 is excessive. So if we aren't include 3.10 in the matrix, then we need to change the "if" statements for code cov since currently they are used for 3.10. It can either be used for 3.8 or 3.11 (doesn't really matter).

docs/source/references/installation.rst
```diff
@@ -24,7 +24,7 @@ Docker Installation and Update
 Local Installation
 ------------------
 
-``nucypher`` supports Python 3.7 and 3.8. If you don’t already have it, install `Python <https://www.python.org/downloads/>`_.
+``nucypher`` supports Python 3.9, 3.10, and 3.11. If you don’t already have it, install `Python <https://www.python.org/downloads/>`_.
```
derekpierre, line 27:
```suggestion
``nucypher`` supports Python 3.8, 3.9, 3.10, and 3.11. If you don’t already have it, install `Python <https://www.python.org/downloads/>`_.
```

setup.py
```diff
@@ -45,9 +43,9 @@
     "Operating System :: OS Independent",
     "Programming Language :: Python",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
```
derekpierre, line 46:
```suggestion
    "Programming Language :: Python :: 3.8",
    "Programming Language :: Python :: 3.9",
```

derekpierre (changes requested):

.github/workflows/python_tests.yml
derekpierre: What's the thinking behind needing tests to run on 3.10, if we already run tests for 3.8 and 3.11?

.github/workflows/python_tests.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [ "3.8", "3.10" ]
+        python-version: [ "3.8", "3.11" ]
```
derekpierre, line 19: Now that we are running tests for python 3.8 and 3.11, we now need to update the code cov "if statements" to use one of these python versions instead of 3.10. eg. https://github.com/nucypher/nucypher/pull/3208/files#diff-df62b370065991dee949f7e19f1e99442c52f7023d3a572a4166d1597b7f0c61L55 (and all others) could be:

```
if: matrix.python-version == '3.8'
...
if: matrix.python-version != '3.8'
```

so that code cov would run for the 3.8 tests and not for 3.11 respectively. We could use 3.11 instead of 3.8 for code cov but I figure that may change to 3.12 or higher before we change the min supported version from 3.8.

Also, there may be a way to define a job variable that we can then use for the check, so that changing it in the future is easier and more maintainable - https://docs.github.com/en/actions/learn-github-actions/variables#defining-environment-variables-for-a-single-workflow - something like:

```yml
jobs:
   python-tests:
     env:
       CODE_COV_PY_VERSION: 3.8
     steps:
       ...
       - name: Unit Tests (Coverage)
          if: matrix.python-version == "$CODE_COV_PY_VERSION"
            <unit tests with coverage>
       - name: Unit Tests
          if: matrix.python-version != "$CODE_COV_PY_VERSION"
            <unit test with no coverage>
       ...
                         
```

derekpierre (dismissed):
See comment about code cov if statements

KPrasch:
bump @jMyles

KPrasch:
> See comment about code cov if statements

Fixed in [5ff758c](https://github.com/nucypher/nucypher/pull/3208/commits/5ff758c5621f0947b561091d0903ac31f51c102a)

.github/workflows/python_tests.yml
KPrasch: ✔️ kept the existing approach for now, just bumped the version to run codecov along with 3.11.

KPrasch:
Related to https://github.com/nucypher/nucypher/issues/3273

KPrasch (approved):

derekpierre (approved):

PR #3194, opened 2023-07-25
AccessControlPolicy (ACP) and ThresholdMessageKit

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
> High-level idea of the changes introduced in this PR. 
> List relevant API changes (if any), as well as related PRs and issues.

**Issues fixed/closed:**
Related to:
- #3187 
- https://github.com/nucypher/ferveo/issues/147
- https://github.com/nucypher/ferveo/pull/149

Closes:
- #3176 

Depends on:
- https://github.com/nucypher/nucypher-core/pull/74

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #3216, opened 2023-08-28
Use released `nucypher-core@0.12.0` package

piotr-roslaniec:
**Type of PR:**
- Other

**Required reviews:** 
- 1

**What this does:**
- Replaces GH repository in project dependencies with a released `nucypher-core` package

- [ ] Relock deps

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -11,7 +11,7 @@ python_version = "3"
 constant-sorrow = ">=0.1.0a9"
 bytestring-splitter = ">=2.4.0"
 hendrix = ">=4.0"
-nucypher-core = "==0.11.0"
+nucypher-core = "==0.12.0"
 # Cryptography
 cryptography = ">=3.2"
 mnemonic = "*"
--- a/requirements.txt
+++ b/requirements.txt
@@ -55,7 +55,7 @@ msgpack==1.0.5
 msgpack-python==0.5.6
 multidict==5.2.0 ; python_version >= '3.6'
 mypy-extensions==0.4.4 ; python_version >= '2.7'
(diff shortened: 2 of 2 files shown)


piotr-roslaniec:
Those CI failures don't seem related to `nucypher-core`.

derekpierre:
> Those CI failures don't seem related to `nucypher-core`.

Seems to be changes to `nucypher-contracts` affecting the acceptance tests.
```
 AttributeError: Dependency project 'nucypher-contracts' has no contract 'ThresholdStakingForPREApplicationMock'.
 ```

Side note: we should consider doing a relocking of overall dependencies instead of just `nucypher-core`. i.e. we should just update the Pipfile to use `nucypher-core` 0.12.0, and then try running `scripts/dependencies/relock_dependencies.sh` to regenerate the [dev-]requirements.txt files.

derekpierre:
Took a deeper look at this.

While I would like a full relock of dependencies - there are some issues with that related to eip-structs 712, and eth-accounts versions causing dependency conflicts which I believe @jMyles seems to be looking into - it's good to get this github dependency out of our code in the interim.

I've tested this change locally using a modified ape-config for acceptance tests to use https://github.com/derekpierre/nucypher-contracts/tree/local-testing instead of `main` to exclude the latest Coordinator changes from https://github.com/nucypher/nucypher-contracts/pull/92; the acceptance tests pass. 

I'd like to have this merged anyway given that the update to the `nucypher-core` github dependency should have occurred before #3194 was merged, and is essentially the same code as `0.12.0` anyway. The change is pretty self-contained.

I'm approving for merge. (cc @KPrasch)

derekpierre (approved):
🎸

KPrasch (approved):

PR #3220, opened 2023-09-01
Add .ruff_cache folder to .gitignore file

manumonti:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Ruff linter generates a cache folder when it runs pre-commit checks. This folder is ignored since it has its own .gitignore file inside the folder. But for homogeneity purposes, this should be added to .gitignore file.

Also, adding to the general .gitignore file makes this folder have a different treatment by IDEs.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ __pycache__
 tests/acceptance/blockchain/.build/
 __local__.json
 /.cache
+.ruff_cache/
 *.so
 /MANIFEST
 /.tox


.gitignore
```diff
@@ -10,6 +10,7 @@ __pycache__
 tests/acceptance/blockchain/.build/
 __local__.json
 /.cache
+.ruff_cache
```
derekpierre, line 13: Technically, without the slash at the end, this ignores *any* file or directory called `.ruff_cache` and does work.
If the goal is just the directory in the root directory and its contents, we can use `.ruff_cache/` (similar to what we do for the `.pytest_cache` folder (L20).

derekpierre (approved):
🎸

piotr-roslaniec (approved):

.gitignore
manumonti: Right! Done ✅

PR #3223, opened 2023-09-11
Make Prometheus exporter always run with Ursula

manumonti:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Prometheus metrics are going to be a necessary part of the Threshold network. Through them, the network can know if the PRE nodes are meeting the node's requirements. This can be useful, for instance, to calculate the staking rewards.

So, the export of these metrics must be mandatory and not an option that the user can enable or disable using command flags.

### DIFF
--- a/nucypher/utilities/prometheus/metrics.py
+++ b/nucypher/utilities/prometheus/metrics.py
@@ -22,10 +22,11 @@ from nucypher.utilities.prometheus.collector import (
 class PrometheusMetricsConfig:
     """Prometheus configuration."""
     def __init__(self,
-                 port: int,
-                 listen_address: str = '',  # default to localhost ip
-                 collection_interval: int = 90,  # every 1.5 minutes
-                 start_now: bool = False):
+        port: int,
+        listen_address: str = "",  # default to localhost ip
+        collection_interval: int = 90,  # every 1.5 minutes
+        start_now: bool = False,
+    ):
 
         if not port:
             raise ValueError('port must be provided')
--- a/tests/integration/utilities/test_prometheus_collectors.py
+++ b/tests/integration/utilities/test_prometheus_collectors.py
@@ -14,7 +14,10 @@ from nucypher.utilities.prometheus.collector import (
     StakingProviderMetricsCollector,
     UrsulaInfoMetricsCollector,
 )
-from nucypher.utilities.prometheus.metrics import create_metrics_collectors
+from nucypher.utilities.prometheus.metrics import (
+    PrometheusMetricsConfig,
+    create_metrics_collectors,
+)
 from tests.constants import MOCK_ETH_PROVIDER_URI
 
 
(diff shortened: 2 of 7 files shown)


piotr-roslaniec (dismissed):

nucypher/characters/lawful.py
```diff
@@ -993,11 +989,13 @@ def run(
         ritualist: bool = True,
         hendrix: bool = True,
         start_reactor: bool = True,
-        prometheus_config: "PrometheusMetricsConfig" = None,
         preflight: bool = True,
         block_until_ready: bool = True,
         eager: bool = False,
     ) -> None:
+        # Locally scoped to prevent import without prometheus explicitly installed
```
derekpierre, line 996: If prometheus is installed, this should no longer need to be locally scoped

derekpierre (changes requested):
Prometheus is typically installed via the setup.py file for the `dev` or `ursula` extras (https://github.com/nucypher/nucypher/blob/development/setup.py#L135), eg. `pip install nucypher[dev]` or `pip install nucypher[ursula]`. 

If we want the `prometheus_client` to always be installed, you'll need to take it out of the setup.py, and add it to the Pipfile for `nucypher`, then relock dependencies (`scripts/dependencies/relock_dependencies.sh`)

nucypher/characters/lawful.py
manumonti: It makes sense, thanks. Done ✅

My only concern is if this could cause any error in Porter. If I understand correctly, adding prometheus_client to the requirements will cause Porter to also install this dependency when deploying.

Some checking should be done before merging this.

derekpierre: > My only concern is if this could cause any error in Porter. If I understand correctly, adding prometheus_client to the requirements will cause Porter to also install this dependency when deploying.

Fair. But if the goal is to have prometheus always be installed so that Ursula can always use it, then I'm not sure we can do anything about that.

You are right that while Porter would ultimately install the prometheus dependency (impliclty as part of the `nucypher` dependency), it won't use it - but that may also apply to other existing `nucypher` dependencies anyway.

manumonti:
Fixed tests and added some news

nucypher/characters/lawful.py
manumonti: Done!

manumonti:
Unrelated tests are failing in CI since we have new contracts in nucypher-contracts.

I don't expect the tests to be passed until #3226  is merged.

manumonti:
This PR is going to be split into several PRs.

The first of them is #3245.

TBD: Prometheus is run along Ursula (not as an option)

nucypher/utilities/prometheus/metrics.py
```diff
@@ -21,19 +20,13 @@
...
-
-        if not port:
-            raise ValueError('port must be provided')
 
-        self.port = port
-        self.listen_address = listen_address
-        self.collection_interval = collection_interval
-        self.start_now = start_now
+    def __init__(self):
+        self.port = 9101
```
derekpierre, line 25: Forcing 9101 is tricky because other services may be run on the same machine, including other things that use prometheus. In fact, we probably have this situation on lynx-2 at the moment. It has a Porter that uses prometheus and probably uses port 9101 for its metrics.

manumonti:
This PR is ready to be reviewed.

tests/integration/utilities/test_prometheus_collectors.py
```diff
@@ -29,6 +32,41 @@ def mock_operator_confirmation(random_address, mock_taco_application_agent):
     mock_taco_application_agent.get_staking_provider_info.return_value = info
 
 
+def test_start_prometheus_exporter_called(ursulas, mock_prometheus):
+    port = 9101
+
+    # Reset start_prometheus_exporter mock just in case it was previously called
+    mock_prometheus.reset_mock()
```
piotr-roslaniec, line 39: I'm not a Python scientist, but would it be possible to move this call into the `mock_prometheus` fixture or is there some reason not to do this? Just curious

piotr-roslaniec (approved):

KPrasch (approved):

tests/integration/utilities/test_prometheus_collectors.py
derekpierre: Is it necessary for the fixture to be "session"-scoped?

For some of these tests perhaps you can instead use a `spy`?

manumonti: > Is it necessary for the fixture to be "session"-scoped?
> 
> For some of these tests perhaps you can instead use a `spy`?

This was added because there are some tests that run several Ursula's instances. So, since Prometheus exporter now runs with each Ursula instance, those tests also try to run several Prometheus exporter instances. This causes test errors because each  Prometheus instance tries to lock the same resources (network port). 

Also, Prometheus exporter is not necessary for most tests, and running it can cause errors, a waste of resources, and tests running for more time. So I decided to just mock it.

Does it make sense?

More info in this commit: 
https://github.com/nucypher/nucypher/pull/3294/commits/9c1547ef5e507b5ba32c151ad83f4252a8982061

PR #3224, opened 2023-09-11
Disable default collector metrics

manumonti:
By default, prometheus exports 'process', 'gc' and 'platform' collector metrics. This information is not useful in our case and introduces some noise to the metrics, so they have been disabled.

**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
By default, prometheus exports 'process', 'gc' and 'platform' collector metrics. This information is not useful in our case and introduces some noise to the metrics, so they have been disabled.

Example of default metrics:

```
# HELP python_gc_objects_collected_total Objects collected during gc
# TYPE python_gc_objects_collected_total counter
python_gc_objects_collected_total{generation="0"} 10717.0
python_gc_objects_collected_total{generation="1"} 3776.0
python_gc_objects_collected_total{generation="2"} 75.0
# HELP python_gc_objects_uncollectable_total Uncollectable objects found during GC
# TYPE python_gc_objects_uncollectable_total counter
python_gc_objects_uncollectable_total{generation="0"} 0.0
python_gc_objects_uncollectable_total{generation="1"} 0.0
python_gc_objects_uncollectable_total{generation="2"} 0.0
# HELP python_gc_collections_total Number of times this generation was collected
# TYPE python_gc_collections_total counter
python_gc_collections_total{generation="0"} 549.0
python_gc_collections_total{generation="1"} 49.0
python_gc_collections_total{generation="2"} 4.0
```

**Issues fixed/closed:**
This is part of "Revisiting Prometheus work": https://github.com/nucypher/nucypher/issues/3240

**Note for reviewers:**

**This is the 2nd PR to review about Prometheus epic.**

### DIFF
--- a/nucypher/utilities/prometheus/metrics.py
+++ b/nucypher/utilities/prometheus/metrics.py
@@ -2,6 +2,7 @@
 import json
 from typing import List
 
+from prometheus_client import GC_COLLECTOR, PLATFORM_COLLECTOR, PROCESS_COLLECTOR
 from prometheus_client.core import Timestamp
 from prometheus_client.registry import REGISTRY, CollectorRegistry
 from prometheus_client.utils import floatToGoString
@@ -122,6 +123,11 @@ def start_prometheus_exporter(
     from prometheus_client.twisted import MetricsResource
     from twisted.web.resource import Resource
     from twisted.web.server import Site
+
+    # Disabling default collector metrics
+    REGISTRY.unregister(GC_COLLECTOR)
+    REGISTRY.unregister(PLATFORM_COLLECTOR)
+    REGISTRY.unregister(PROCESS_COLLECTOR)
 
     metrics_collectors = create_metrics_collectors(ursula)
     # initialize collectors


nucypher/utilities/prometheus/metrics.py
```diff
@@ -129,6 +129,11 @@ def start_prometheus_exporter(
     from twisted.web.resource import Resource
     from twisted.web.server import Site
 
+    # Disabling default collector metrics
+    REGISTRY.unregister(GC_COLLECTOR)
+    REGISTRY.unregister(PLATFORM_COLLECTOR)
+    REGISTRY.unregister(PROCESS_COLLECTOR)
```
piotr-roslaniec, line 135: Process collector (RAM & CPU usage, etc.) sounds generally useful to me. How did we decide that this is irrelevant to our goals?

piotr-roslaniec (approved):

nucypher/utilities/prometheus/metrics.py
manumonti: Fair question.

I'm focused on having client metrics (I mean, Ursula metrics) to have enough information to properly calculate Threshold staking rewards. At the same time, I would like to make a clean-up in the current metrics removing all unuseful or redundant information to have only relevant information as tBTCv2 nodes do:

```
# TYPE btc_connectivity gauge
btc_connectivity 1 1694426887268

client_info{version="v2.0.0-m4"} 1

# TYPE connected_bootstrap_count gauge
connected_bootstrap_count 7 1694426947164

# TYPE connected_peers_count gauge
connected_peers_count 140 1694426947166

# TYPE eth_connectivity gauge
eth_connectivity 1 1694426887162

# TYPE tbtc_pre_params_count gauge
tbtc_pre_params_count 1000 1694426948350
```

In any case, this PROCESS_COLLECTOR has not been collecting information, so no information about CPU or RAM is exposed at all. So I would prefer to disable this collector which is not working anyway.

That said, I could agree that process information (RAM or CPU) may be useful for troubleshoothing or to better understand the performance of the nodes. But IMHO we should address this later on.

piotr-roslaniec: I see how this metric is not relevant to the dashboard we're building and I agree it doesn't seem like a priority. Thanks for explaining!

manumonti:
Unrelated tests are failing in CI since we have new contracts in nucypher-contracts.

I don't expect the tests to be passed until #3226  is merged.

manumonti:
This PR is ready to review!

KPrasch (approved):

Issue #3229, opened 2023-09-15
Devise github workflow to have PRs into epic branches run CI tests


derekpierre:
It would be nice if pull requests into epic branches would run CI tests (python at least, maybe others).

Perhaps the branch filter for our CI builds can be something like:
```
on:
  pull_request:
    branches:
      - main
      - development
      - epic-*
```

So any PR into an epic branch that matches the epic glob pattern eg. `epic-coordinator-13` will then have CI tests run.

Open to other ideas.

PR #3230, opened 2023-09-15
[DO NOT MERGE] [WIP] PR created to see passing CI builds for #3227

derekpierre:
Since #3227 is going into an EPIC branch (see #3226) the CI builds don't run. I don't want to change the base branch for the PR to `development` for #3227 because then reviewers can't review just alpha-13 changes - instead they would get all changes from epic #3226.

Filed an issue to better handle this situation, #3229 , but this is the workaround for now.

This PR >>>>> **SHOULD NOT** <<<<< be merged. It will be closed once #3227 is merged after review.


derekpierre:
#3227 now merged into #3226. This PR is not needed for builds anymore.

PR #3231, opened 2023-09-18
Add return value to start prometheus function

manumonti:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Adding the Prometheus exporter configuration as a return value for the function start_prometheus_exporter allows the class calling the function to know how Prometheus has been deployed (port, metrics_prefix, connection interval, listen address).

**Notes for reviewers:**
This PR is based on https://github.com/nucypher/nucypher/pull/3223.

**Please, don't merge this PR until #3223 is merged.**


piotr-roslaniec (approved):
LGTM. I wouldn't mind if this change was a part of the parent PR.

manumonti:
This PR is RFR

tests/unit/test_prometheus.py
```diff
@@ -55,7 +54,7 @@ def test_prometheus_metrics_config():
                                                 start_now=True)
     assert prometheus_config.listen_address == listen_address
     assert prometheus_config.collection_interval == collection_interval
-    assert prometheus_config.start_now
+    assert prometheus_config.start_now is True
```
piotr-roslaniec, line 57: Just curious, why do we need to `is True` here? Wouldn't a plain `assert` work?

manumonti: Yes, the plain assertion was working perfectly. I just thought that a more explicit assertion could be verbose and, hence, more clear about what we are checking here.

But I don't have a strong opinion here, we can return to the original assertion if that works better.

piotr-roslaniec: I don't have a strong opinion either, `assert` is taking both values and expressions and a sprinkle of verbosity wouldn't hurt

KPrasch (approved):

PR #3237, opened 2023-09-20
Use of TACoChildApplicationAgent for staking information and accommodate syncing issues between Root and Child

### LINKED ISSUE #3235: TACo protocol should use TACoChildApplication as a reference for staking information
From the protocol point of view (i.e. Coordinator contract), the canonical information lives on the `TACoChildApplication`. This information in turn is sourced from the root TACoApplication on mainnet via the state sync bridge, but that's outside of the scope of nodes' view of the Taco protocol

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

Closes #3235 .

**Why it's needed:**

**Notes for reviewers:**


nucypher/blockchain/eth/actors.py
```diff
@@ -662,41 +680,58 @@ def block_until_ready(self, poll_rate: int = None, timeout: int = None):
...
+                            color="green",
+                        )
+                    else:
+                        child_bonded_address_info = (
+                            f"({truncate_checksum_address(taco_child_bonded_address)})"
+                            if taco_child_bonded_address != NULL_ADDRESS
+                            else ""
+                        )
+                        emitter.message(
+                            f"! Bonded staking provider address {truncate_checksum_address(taco_root_bonded_address)} on {taco_root_pretty_chain_name} not yet synced to child application on {taco_child_pretty_chain_name} {child_bonded_address_info}; waiting for sync",
```
theref, line 711: love this 🚀

nucypher/blockchain/eth/constants.py
```diff
@@ -10,6 +10,7 @@
 STAKING_ESCROW_STUB_CONTRACT_NAME = "StakingEscrowStub"
 ADJUDICATOR_CONTRACT_NAME = "Adjudicator"
 TACO_APPLICATION_CONTRACT_NAME = "TACoApplication"
+TACO_CHILD_APPLICATION_CONTRACT_NAME = "TACoChildApplication"
```
KPrasch, line 13: This PR is looking good overall!  I do have one comment related to this line:  To avoid needing to change this line depending on whether or not a node is running on testnet vs mainnet, I assert that we always use this contract name in the registry, which is the historical convention anyways.

derekpierre: 👍 I agree with that assertion. We'll need to modify some logic when deploying to lynx as part of work for https://github.com/nucypher/nucypher-contracts/pull/119

derekpierre:
An assumption I'm making:

`Teacher` nodes check two things about nodes when validating operator metadata:
1.  `is operator bonded` - https://github.com/derekpierre/nucypher/blob/taco-child/nucypher/network/nodes.py#L1042
2. `are they really staking` - https://github.com/derekpierre/nucypher/blob/taco-child/nucypher/network/nodes.py#L1058

I assume the canonical spot for this information is indeed TACo root and NOT the child. Both are set on the root application, which subsequently updates the child application, so presumably, the root is the best place for that information. So I've left the above functions unchanged, i.e. they continue to use the TACoApplicationAgent (root application agent).

cc @vicky.z , @dnunez

KPrasch:
> I assume the canonical spot for this information is indeed TACo root and NOT the child. Both are set on the root application, which subsequently updates the child application, so presumably, the root is the best place for that information. So I've left the above functions unchanged, i.e. they continue to use the `TACoApplicationAgent` (root application agent).

Given that nodes are not permitted to start up without bonding state being synchronized with the child, this is a fair assumption.  In this case, we want a node to become invalid as soon as it's unbonded, correct (no need to await state sync for unbonding)?  We don't want to sample unqualified nodes after all, even before the unbonding state is bridged.  Side question, does deauthorization of the TACo application on the staking-side result in a forced unbonding?

derekpierre:
> Given that nodes are not permitted to start up without bonding state being synchronized with the child, this is a fair assumption. In this case, we want a node to become invalid as soon as it's unbonded, correct (no need to await state sync for unbonding)? We don't want to sample unqualified nodes after all, even before the unbonding state is bridged. 

👍 . Same thing for our OperatorBondingTracker (it uses root application).

> Side question, does deauthorization of the TACo application on the staking-side result in a forced unbending?

Looks that way - https://github.com/nucypher/nucypher-contracts/blob/main/contracts/contracts/TACoApplication.sol#L445.

@vzotova / @cygnusv can confirm.

nucypher/blockchain/eth/constants.py
vzotova: I suppose even for Lynx still will be `TACoChildApplication` abi, right?

derekpierre: Yes, the ABI is the same, but we need some magic when generating the lynx registry since the contract names for root and child application contracts are different - https://github.com/derekpierre/nucypher-contracts/blob/lynx-deployment/contracts/contracts/testnet/LynxSet.sol.

KPrasch (approved):

PR #3238, opened 2023-09-20
Rename integration tests

manumonti:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
These tests have the same name as other tests in the `unit` and `acceptance` folders.

VS Code IDE is unable to do the discovery process of the tests when:

1. There are different sources of tests (different folders)
2. Some names are duplicated in tests of different folders.

This causes the discovery process to crash, so tests are not shown in Test Explorer.

Renaming the tests with duplicated names works and lets the developers using VS Code keep working with this IDE.

**Note for reviewers**

**Do not merge before #3239**


tests/integration/blockchain/test_integration_block_confirmations.py
```diff
@@ -6,7 +6,7 @@
 
 
 @pytest.mark.skip("This test need to be refactored to use some other transaction than deployment")
-def test_block_confirmations(testerchain, test_registry, mocker):
```
derekpierre, line 9: Any reason the test itself needs to be renamed? Are there other tests with the same name?

newsfragments/3238.misc.rst
derekpierre, line 1: Lets do `3238.dev.rst` since this is purely dev related - and doesn't need to be in release notes.

manumonti: You're right. Thanks 👍
✅ Done

tests/integration/blockchain/test_integration_block_confirmations.py
manumonti: Not at all. I thought it was necessary, but I have just checked and renaming the file is enough.

✅ Done

tests/integration/blockchain/test_block_confirmations.py
```diff
@@ -1,43 +0,0 @@
-import pytest
-
-from nucypher.blockchain.eth.clients import EthereumClient
-from nucypher.blockchain.eth.signers.software import Web3Signer
-from nucypher.crypto.powers import TransactingPower
-
-
-@pytest.mark.skip("This test need to be refactored to use some other transaction than deployment")
-def test_block_confirmations(testerchain, test_registry, mocker):
```
KPrasch, line 9: 👍🏻 good removal

derekpierre (approved):
🎸

PR #3239, opened 2023-09-21
Format tests code to meet Ruff and Darker code style 

manumonti:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

This reformats the code of the following tests to meet Ruff and Darker code style guides.

This is a necessary step for PR #3238 . The mentioned PR consists of a renaming of these tests, so Ruff and Darker checks are executed in CI, resulting in failed tests since currently, these tests don't meet the mentioned code style guidelines.

### DIFF
--- a/tests/integration/blockchain/test_block_confirmations.py
+++ b/tests/integration/blockchain/test_block_confirmations.py
@@ -5,39 +5,53 @@ from nucypher.blockchain.eth.signers.software import Web3Signer
 from nucypher.crypto.powers import TransactingPower
 
 
-@pytest.mark.skip("This test need to be refactored to use some other transaction than deployment")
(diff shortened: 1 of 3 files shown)


piotr-roslaniec (approved):

manumonti:
For some reason the "Check Release Note" CI step is failing because is trying to find a 3238 newsfragment file instead of 3239, which is the correct PR number.

In any case, tests are passing for #3238, which is a continuation of this PR. So this PR can be merged after positive reviews.

KPrasch (approved):

PR #3241, opened 2023-09-21
EconomicsFactory/Economics Removal

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Removal of obsolete `EconomicsFactory`/`Economics` classes.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


vzotova (approved):

tests/fixtures.py
```diff
@@ -366,8 +359,10 @@ def policy_rate():
 
 
 @pytest.fixture(scope='module')
-def policy_value(application_economics, policy_rate):
-    value = policy_rate * application_economics.min_operator_seconds
+def policy_value(policy_rate):
+    # TODO constant?
+    min_operator_seconds = 60 * 60 * 24  # one day in seconds
```
KPrasch, line 364: Good thought, in lieu of a full-blown class it will still be useful to ensure the same value is used during deployment either by fixturization, or use of a constant.

derekpierre: ✔️

piotr-roslaniec (approved):

nucypher/cli/types.py
```diff
@@ -128,7 +127,8 @@ def convert(self, value, param, ctx):
 WEI = click.IntRange(min=1, clamp=False)  # TODO: Better validation for ether and wei values?
 GWEI = DecimalRange(min=0)
 
-__min_authorization = TToken.from_units(Economics._default_min_authorization).to_tokens()
+
+__min_authorization = TToken(40_000, "T").to_tokens()  # TODO right spot for this?
```
cygnusv, line 131: What's the use for this anyway? The min authorization is whatever is in the contract, and we don't use the `nucypher` CLI for deployment or staking.

tests/fixtures.py
```diff
@@ -366,8 +360,8 @@ def policy_rate():
 
 
 @pytest.fixture(scope='module')
-def policy_value(application_economics, policy_rate):
-    value = policy_rate * application_economics.min_operator_seconds
+def policy_value(policy_rate):
+    value = policy_rate * MIN_OPERATOR_SECONDS
```
cygnusv, line 364: What does this constant represent?

cygnusv (approved):

KPrasch (approved):
You know how much I love read lines 🤠

nucypher/cli/types.py
derekpierre: It's used in a CLI option as a default value - https://github.com/nucypher/nucypher/blob/development/nucypher/cli/options.py#L65 - which is used as an Ursula parameter for specify the seed node minimum stake.

tests/fixtures.py
derekpierre: It's the minimum amount of time before an operator can be changed. It's the same value used in the TACoApplication contract.

nucypher/cli/types.py
cygnusv: huh, but I'm wondering why is that a CLI parameter. The only thing that should matter is what's in the contract, so I don't see how a different value can be used

derekpierre: The CLI parameter allows nodes to specify a higher stake requirement for its seed node teacher than the minimum (trust mitigation I guess... 🤷), but you are right that perhaps the logic can be reworked so that if a value is not specified, the min value is obtained from the contract.

Issue #3240, opened 2023-09-21
Revisiting Prometheus metrics - roadmap


manumonti:
Prometheus can be a good first approximation to check and track the status of TACo nodes on the Threshold Network.

Having enough metrics can be useful for tracking appropriately if the nodes are complying with staking requirements and, hence, if the associated stakers are eligible for staking rewards.

Also, these metrics can be useful to monitor nodes' status, quickly discovering possible malfunctions or inefficiencies.

The roadmap can be something similar to:

1. prometheus-client package is no longer a dev dependency, but a Ursula dependency: now, it is included in pipfile so it's installed for Ursula.
2. Prometheus execution is optional: users can decide if they want Prometheus running or not. Also, we don't want to have a deprecated flag (--prometheus) in nucypher ursula run command for release 7.1, so this flag is removed, and enabling Prometheus is controlled by an environment variable or config file.
3. Current metrics pruning and renaming: https://github.com/nucypher/nucypher/pull/3224 and https://github.com/nucypher/nucypher/pull/3232.
4. Revisit current metrics and remove those that are not working or are no longer useful (like operator ETH balance).
5. Add useful metrics: https://github.com/nucypher/nucypher/issues/3236
6. Add return value to start Prometheus function: https://github.com/nucypher/nucypher/pull/3231
7. Enabling support for multiple intervals to collect metrics (large interval for those metrics that requires a connection with a web3 provider as Infura and short interval for those metrics only dependent on the node status like RAM, CPU, etc.)

~~8. Prometheus running along Ursula is mandatory. No option to disable it. This is a necessary mechanism to calculate rewards and to collect statistics of the network.~~

9. Add tests: https://github.com/nucypher/nucypher/issues/3218

derekpierre:
The foundation of prometheus has now been completed.

Of course additional metrics can always be added as desired.  Closing this issue now.

PR #3245, opened 2023-09-22
Add prometheus-client package to Pipfile

manumonti:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
- This PR adds `prometheus-client` to Pipfile, so Prometheus will be installed for Ursula in any case. Up to now, this package was only installed when running a `dev` installation of nucypher.

- Nucypher dependencies have been relocked using the `relock_dependencies` script.

- Prometheus imports are no longer conditional on the installation of it.

**Issues fixed/closed:**
This is part of "Revisiting Prometheus work": https://github.com/nucypher/nucypher/issues/3240

**Notes for reviewers:**
This is the 1st PR to review about Prometheus epic. 

Also, this work had been made in https://github.com/nucypher/nucypher/pull/3223, but this is been split into more atomic PRs.


manumonti:
This PR is ready to review!

docs-requirements.txt
```diff
@@ -1,33 +1,33 @@
...
-click==8.1.7 ; python_version >= '3.7'
-click-default-group==1.2.4 ; python_version >= '2.7'
-docutils==0.17.1
-idna==3.4 ; python_version >= '3.5'
-imagesize==1.4.1 ; python_version >= '2.7' and python_version not in '3.0, 3.1, 3.2, 3.3'
-importlib-resources==6.0.1 ; python_version < '3.10'
+alabaster==0.7.13; python_version >= '3.6'
+babel==2.12.1; python_version >= '3.7'
+certifi==2023.7.22; python_version >= '3.6'
+charset-normalizer==2.1.1; python_full_version >= '3.6.0'
```
piotr-roslaniec, line 5: Just a question: IIRC, `nucypher` supports Python 3.8+. The contents or `requirements.txt` files etc., are supposed to reflect the minimum supporter Python version, right?

piotr-roslaniec (approved):

KPrasch (approved):

docs-requirements.txt
manumonti: I think so.

In any case, requirements.txt is generated by `relock_dependencies` script. Not sure what is the logic behind all that `python_version` tags. Maybe is the `python_version` that all those packages need?

derekpierre: These are autogenerated by pip when relocking and doesn't take into account `nucypher`'s specific minimum supported version. Some of the restrictions are applicable while others are redundant. For example, https://github.com/nucypher/nucypher/blob/development/requirements.txt#L9, since we support 3.8, 3.9, 3.10. That dependency only applies to < 3.9.

Issue #3242, opened 2023-09-21
Review CLI functionality and Dev Scripts


derekpierre:
Lots of changes (including refactoring) have been made to the codebase. 

Once we do the latest test deployment to Lynx (probably `alpha-13`), we need to review our CLI commands eg. `nucypher status events` and dev scripts eg. `nucypher_agents.py`, `nucypher_dkg.py` to ensure that they still work properly.

Related to #3241 , #3243 .

PR #3254, opened 2023-09-27
Update DKG dev script to work on Lynx (`dkg-dev-13`)

derekpierre:
**Type of PR:**
Other

**Required reviews:** 
2

**What this does:**
- Introduces `InMemorySigner` with tests
- Updates DKG script to account for enrico authorization.

**Why it's needed:**
- DKG scripting had fallen out of compatibility with the alpha 13 contract suite, this PR restores compatibility
- After redeploying Lynx with Alpha 13, new rituals and threshold decryption live testing is needed

**Notes for reviewers:**

- Related to #3242 .

- We've successfully initiated ritual id `#1`, authorized the default Enrico ('0x070a85eD1Ddb44ecD07e746235bE0B959ff5b30A') to use it, and can encrypt/decrypt data using the ritual.
- You can observe Coordinator txs here - https://mumbai.polygonscan.com/address/0x8e49989f9d3ad89c8ab0de21fba2e00c67ca872f
- You can run the script yourself to test encryption/decryption with ritual 1 by running:
```bash
$ python scripts/hooks/nucypher_dkg.py --eth-provider <GOERLI PROVIDER URI> --coordinator-provider <MUMBAI PROVIDER URI> --ritual-id 1
```

### DIFF
--- a/tests/unit/test_memory_signer.py
+++ b/tests/unit/test_memory_signer.py
@@ -1,0 +1,57 @@
+import pytest
+from cytoolz import assoc
+from eth_account._utils.legacy_transactions import Transaction
+from eth_utils import to_checksum_address
+from hexbytes import HexBytes
+
+from nucypher.blockchain.eth.constants import LENGTH_ECDSA_SIGNATURE_WITH_RECOVERY
+from nucypher.blockchain.eth.signers import InMemorySigner, Signer
+from tests.unit.test_web3_signers import TRANSACTION_DICT
+
+
+@pytest.fixture(scope="function")
+def signer():
+    _signer = InMemorySigner()
+    return _signer
+
+
+@pytest.fixture(scope="function")
+def account(signer):
+    _account = signer.accounts[0]
+    return _account
+
+
+def test_memory_signer_from_signer_uri():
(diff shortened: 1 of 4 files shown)


KPrasch (approved):

tests/unit/test_memory_signer.py
```diff
@@ -0,0 +1,57 @@
...
+
+from nucypher.blockchain.eth.constants import LENGTH_ECDSA_SIGNATURE_WITH_RECOVERY
+from nucypher.blockchain.eth.signers import InMemorySigner, Signer
+from tests.unit.test_web3_signers import TRANSACTION_DICT
+
+
+@pytest.fixture(scope="function")
+def signer():
+    _signer = InMemorySigner()
+    return _signer
```
piotr-roslaniec, line 15: Why not just `return InMemorySigner()`?

piotr-roslaniec (approved):

tests/unit/test_memory_signer.py
KPrasch: It's a function of stylistic habit - I prefer block vs inline `return` usage so that a there is a line that can be used in the debugger inside the function body.

PR #3261, opened 2023-10-01
Rework Contract Registry Data Format and Internal API

KPrasch:
**Type of PR:**
Rework

**Required reviews:** 
2

**What this does:**

##### Rework
- Reworks contract registry data into JSON format supporting multi-chain deployments
- Reworks contract types into `RegistrySource`s
- Use nucypher domain names as registry filenames
- Refactor `registry_from_ape_deployments` -> `ApeRegistrySource`
- Expands test coverage of `RegistrySourceManager` and `RegistrySource`s
- Cleanup overusage of `test_registry_source_manager`; rename to `mock_registry_sources`
- Implements `MockRegistrySource` for unit and integration tests
- Converts mainnet, tapir, and lynx registries to the new format

##### Removals
- Support for empty or no-content contract registries
- `VersionedContract` and searching of the contract registry by version
- Registry writing utilities (previously required for `nucypher-deploy`)
- `nucypher/blockchain/eth/contract_registry/historical`
- Retires legacy mainnet contracts and oryx testnet

##### Renames
- `CanonicalRegistrySource` -> `RegistrySource`
- `registry.network` -> `registry.domain`

**Why it's needed:**
- Unblocks multi-chain testnet deployments to domains that include polygon mumbai
- Removal of outdated/dead code related to legacy deployment tools that has migrated to nucypher/nuypher-contracts
- Fixes deployment namespace and blockchain network name shadowing
- Corrects the domain/registry data model to reflect multi-chain deployments
- Relocation of contracts to nucypher/nucypher-contracts means contract versioning falls out-of-scope of nucypher/nucypher 

**Notes for reviewers:**
- PR #3262 is a required follow-up before this PR is fully integrated
- Based over #3259 
- Related to https://github.com/nucypher/nucypher/issues/2918
- Many of the diff lines are renames and ABIs
- Reminder: JSON keys must be strings
- nucypher/nucypher-contracts registry merge utilities will require an update for compatibility
- Example of new registry data format
```python
# nucypher/nucypher/blockchain/eth/contract_registry/lynx.json

{
  "5": {
    "TACoApplication": {
      "address": "0xA59820B4C01A3c3869554DB0243a02bcA7b15760",
      "abi": [...]
    },
    ...
  },
  "80001": {
    "TACoChildApplication": {
      "address": "0xdC078916291e1ef683A4f469fa32025c872194df",
      "abi": [...]
    },
    "Coordinator": {
[...]

PR #3262, opened 2023-10-02
TACo Domain Entrenchment/Assocation

derekpierre:
**Type of PR:**
Refactor

**Required reviews:** 
2

**What this does:**
- Fixes incorrect data model for selection of domains and networks via character APIs
- Lean into the fact that TACo `domain` (mainnet, tapir, lynx etc.) always implies mapping of eth network -> poly network, so specifying separate parameters for the polygon network is redundant.
- Refactors `nucypher.blockchain.eth.domains`
    - Deprecate `NetworksInventory`
    - Introduces `ChainInfo`, `DomainInfo` and `TACoDomain` 
- Deprecate cli/config parameters:
     - `pre-payment-network`
     - `coordinator-network`
- Rename cli/config/method parameters:
     - `pre-payment-provider` -> `polygon_endpoint`
     - `eth_provider_uri` -> `eth_endpoint`
     - `coordinator_provider_uri` -> `polygon_endpoint`
     - `condition_provider_uris` -> `condition_blockchain_endpoints`
     - `provider_uri` -> `(eth|polygon|blockchain)_endpoint[s]`


nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,83 @@
...
+
+
+class PolyNetwork(ChainInfo, Enum):
+    POLYGON = ChainInfo(137, "polygon")
+    MUMBAI = ChainInfo(80001, "mumbai")
+    # testing
+    TESTERCHAIN = ChainInfo(131277322940537, TEMPORARY_DOMAIN)
+
+
+class TACoNetwork(NamedTuple):
```
KPrasch, line 32: To help disambiguate between "networks" I suggest using "domain"

```suggestion
class TACoDomain(NamedTuple):
```

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,83 @@
...
+    eth_network: EthNetwork
+    poly_network: PolyNetwork
+
+    def is_testnet(self) -> bool:
+        return self.eth_network != EthNetwork.MAINNET
+
+
+class UnrecognizedNetwork(RuntimeError):
+    """Raised when a provided network name is not recognized."""
+    pass
```
KPrasch, line 43: `pass` can be removed here - the docstring is enough to indicate a block.

```suggestion
```

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,83 @@
...
-    @classmethod
-    def get_ethereum_chain_id(cls, network):
-        try:
-            return cls.__to_chain_id_eth[network]
-        except KeyError:
-            raise cls.UnrecognizedNetwork(network)
+from nucypher.config.constants import TEMPORARY_DOMAIN
+
+
+class ChainInfo(NamedTuple):
```
KPrasch, line 7: 👍🏻 Excellent rename

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,83 @@
...
+    chain_id: int
+    chain_name: str
```
KPrasch, line 9: Since you're already in a "chain" namespace here, you don't need to smurf.  Consider `chain_id` -> `id` and `chain_name` -> `name`.  In implementation this will be:

```python
chain_info = ChainInfo(name='lynx', id=5)
_id, name = chain_info.id, chain_info.name
```

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,83 @@
...
+    @property
+    def name(self) -> str:
+        return self.chain_name
+
+
+class EthNetwork(ChainInfo, Enum):
+    MAINNET = ChainInfo(1, "mainnet")
+    GOERLI = ChainInfo(5, "goerli")
+    SEPOLIA = ChainInfo(11155111, "sepolia")
+    # testing
```
KPrasch, line 21: redundant comment IMO

```suggestion
```

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,83 @@
...
+    TESTERCHAIN = ChainInfo(131277322940537, TEMPORARY_DOMAIN)
+
+
+class PolyNetwork(ChainInfo, Enum):
+    POLYGON = ChainInfo(137, "polygon")
+    MUMBAI = ChainInfo(80001, "mumbai")
+    # testing
```
KPrasch, line 28:
```suggestion
```

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,83 @@
...
 
     @classmethod
-    def validate_network_name(cls, network_name: str):
-        if network_name not in cls.NETWORKS:
-            raise cls.UnrecognizedNetwork(
-                f"{network_name} is not a recognized network."
-            )
+    def get_network_names(cls) -> List[str]:
+        networks = [network.name for network in cls.SUPPORTED_NETWORKS]
+        return networks
```
KPrasch, line 83: 👍🏻

nucypher/characters/lawful.py
```diff
@@ -825,8 +817,8 @@ def __init__(
         operator_address: Optional[ChecksumAddress] = None,
         client_password: Optional[str] = None,
         transacting_power: Optional[TransactingPower] = None,
-        operator_signature_from_metadata=NOT_SIGNED,
-        eth_provider_uri: Optional[str] = None,
+        eth_endpoint: Optional[str] = None,
+        polygon_endpoint: Optional[str] = None,
```
KPrasch, line 821: 👍🏻  These names look great here, it's obvious in retrospect that this is more understandable and accurate.

KPrasch (commented):
Overall looking good!  Here's a preliminary review.

nucypher/blockchain/eth/networks.py
derekpierre: ✔️

nucypher/blockchain/eth/networks.py
derekpierre: ✔️

nucypher/blockchain/eth/networks.py
derekpierre: ✔️

nucypher/blockchain/eth/networks.py
derekpierre: ✔️

nucypher/blockchain/eth/networks.py
derekpierre: I renamed [Eth/Poly]Network -> [Eth/Polygon]Chain, so not sure if that changes your mind,.

I think we need to lean in to either "domain" or "network". We've used both across a variety of areas. We've used "domain" for characters / config /  registry (your PR), but "network" for other areas such as: ContractPayment, CLI options, BaseActor, CLI tools.

I am partial to "network" (but this involves updating the config from domain), so I'm game to change it to "domain", but just want to confirm that everything in our codebase related to names like "lynx", "mainnet" etc. will be considered "domain" moving forward.

KPrasch: I suggest that we use `domain` as the name for the abstraction representing a set of nucypher contracts deployed across two blockchains and `network` in the industry-conventional manner to indicate a blockchain network (mainnet, sepolia, goerli, polygon-mumbi, etc.).

The `domain` abstraction used in this manner is intentionally the same `domain` used in P2P.  This is sensible since P2P verification of node stakes, bonds, and other on-chain state must be represented in the correct contract set (and thus blockchains) in the `domain`.

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,75 @@
...
+    MAINNET = ChainInfo(1, "mainnet")
+    GOERLI = ChainInfo(5, "goerli")
+    SEPOLIA = ChainInfo(11155111, "sepolia")
+    TESTERCHAIN = ChainInfo(131277322940537, TEMPORARY_DOMAIN)
+
+
+class PolygonChain(ChainInfo, Enum):
+    POLYGON = ChainInfo(137, "polygon")
+    MUMBAI = ChainInfo(80001, "mumbai")
+    TESTERCHAIN = ChainInfo(131277322940537, TEMPORARY_DOMAIN)
```
KPrasch, line 22: What's the advantage of using classes vs. module-level constants for these `ChainInfo` instances?  Less indirection will be an improvement IMO.

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,75 @@
...
+class NetworksInventory:
+    MAINNET = TACoNetwork("mainnet", EthChain.MAINNET, PolygonChain.POLYGON)
+    # Testnets
+    ORYX = TACoNetwork("oryx", EthChain.GOERLI, PolygonChain.POLYGON)
+    LYNX = TACoNetwork("lynx", EthChain.GOERLI, PolygonChain.MUMBAI)
+    TAPIR = TACoNetwork("tapir", EthChain.GOERLI, PolygonChain.MUMBAI)
+    # TODO did Ibex even use a PolyNetwork?
+    IBEX = TACoNetwork(
+        "ibex", EthChain.GOERLI, PolygonChain.MUMBAI
+    )  # this is required for configuration file migrations (backwards compatibility)
```
KPrasch, line 47: No, it did not.  Perhaps use a sentinel value instead.  This could be as simple as `None` but you could use some kind of constant.

KPrasch (commented):
Since the data model of registry has changed from one-to-one to one-to-many in PR #3261 there are going to be conflicts in code blocks pertaining to `network`, `domain`, `coordinator_network` and `pre_payment_network`.

derekpierre:
> Since the data model of registry has changed from one-to-one to one-to-many in PR #3261 there are going to be conflicts in code blocks pertaining to `network`, `domain`, `coordinator_network` and `pre_payment_network`.

For sure! The rebase is going to be a doozy!

nucypher/blockchain/eth/networks.py
derekpierre: To me, it better fortifies associations/linking and naming; constants can be related to each other, but are not directly linked.

nucypher/blockchain/eth/actors.py
```diff
@@ -131,23 +131,23 @@ def __init__(self, registry: BaseContractRegistry, **kwargs):
...
+    #         registry=self.registry,
+    #     )
+    #     return self.__token_agent
+    #
+    # @property
+    # def token_balance(self) -> NU:
+    #     """Return this actor's current token balance"""
+    #     balance = int(self.token_agent.get_balance(address=self.checksum_address))
+    #     nu_balance = NU(balance, 'NuNit')
+    #     return nu_balance
```
KPrasch, line 150: I need to have a closer look, but this is most likely an artifact from legacy staking tools.

nucypher/blockchain/eth/networks.py
```diff
@@ -1,57 +1,68 @@
...
+
+    DEFAULT_DOMAIN_NAME: str = MAINNET.name
 
     @classmethod
-    def validate_network_name(cls, network_name: str):
-        if network_name not in cls.NETWORKS:
-            raise cls.UnrecognizedNetwork(
-                f"{network_name} is not a recognized network."
-            )
+    def from_domain_name(cls, domain: str) -> TACoDomain:
```
KPrasch, line 63: This method name implies that it's an alternate constructor, but it's not (it's a resolver).  I think this can this be simplified by removing `NetworksInventory` all together - There are too many namespaces and levels of indirection here IMO.  I'd like to encourage simple module-level constants instead.

nucypher/blockchain/eth/providers.py
```diff
@@ -14,25 +14,33 @@ class ProviderError(Exception):
     pass
 
 
-def _get_IPC_provider(eth_provider_uri) -> BaseProvider:
-    uri_breakdown = urlparse(eth_provider_uri)
+def _get_IPC_provider(blockchain_endpoint) -> BaseProvider:
```
KPrasch, line 17: Can be simplified to just `endpoint: str` IMO.  There are several more usages below as well.

```suggestion
def _get_IPC_provider(endpoint: str) -> BaseProvider:
```

nucypher/blockchain/eth/registry.py
```diff
@@ -233,11 +237,13 @@ def read(self) -> Union[list, dict]:
...
-    def from_latest_publication(cls,
-                                *args,
-                                source_manager=None,
-                                network: str = NetworksInventory.DEFAULT,
-                                **kwargs) -> 'BaseContractRegistry':
+    def from_latest_publication(
+        cls,
+        *args,
+        source_manager=None,
+        network: str = NetworksInventory.DEFAULT_DOMAIN_NAME,
```
KPrasch, line 244: I suggest removing the default here and always explicitly passing the correct value.  Is the default even used anywhere?

nucypher/characters/base.py
```diff
@@ -35,7 +35,8 @@ class Character(Learner):
     def __init__(
         self,
         domain: str,
-        eth_provider_uri: str = None,
+        eth_endpoint: str = None,
+        polygon_endpoint: str = None,
```
KPrasch, line 39: 👍🏻 looking good

nucypher/blockchain/eth/networks.py
derekpierre: Oh. I think I misunderstood what you originally meant. Yes, this change is indeed being made - it's on my list. Changes incoming.

derekpierre: ✔️

nucypher/blockchain/eth/registry.py
derekpierre: 👍 . This will get removed as part of the rebase over your changes.

nucypher/blockchain/eth/networks.py
derekpierre: ✔️  - used `None`. It's unclear to me that we need an ibex entry at all. I'm just adhering to the comment which was there from before.

nucypher/blockchain/eth/networks.py
derekpierre: ✔️ Resolved - misunderstood you're message.

nucypher/blockchain/eth/providers.py
derekpierre: Do you mean just for this module or in general?

derekpierre: Some spots like `providers.py`/`interfaces.py` (perhaps others...) it is probably redundant to use the `blockchain_` prefix. In some other areas, I think it does provide clarity.

Made an initial change to `providers.py`/`interfaces.py`.

nucypher/blockchain/eth/networks.py
derekpierre: ✔️

nucypher/blockchain/eth/actors.py
derekpierre: ✔️

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,62 @@
...
+
+SUPPORTED_DOMAIN_NAMES = [domain.name for domain in SUPPORTED_DOMAINS]
+
+
+def from_domain_name(domain: str) -> TACoDomain:
+    for taco_domain in SUPPORTED_DOMAINS:
+        if taco_domain.name == domain:
+            return taco_domain
+
+    raise UnrecognizedDomain(f"{domain} is not a recognized domain.")
```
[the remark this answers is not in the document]
derekpierre: A bit of a meld of your idea and mine.

Adding test domains is a bit simpler with the meld. Let me know what you think.

examples/pre/finnegans_wake_demo/finnegans-wake-demo-l2.py
```diff
@@ -40,22 +40,21 @@
 
 print("\n************** Setup **************\n")
 
-###########
-# Network #
-###########
+##########
+# DOMAIN #
```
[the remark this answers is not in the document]
derekpierre: ✔️

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,70 @@
...
+
+class EthChain(ChainInfo, Enum):
+    MAINNET = (1, "mainnet")
+    GOERLI = (5, "goerli")
+    SEPOLIA = (11155111, "sepolia")
+    TESTERCHAIN = (131277322940537, "eth-tester")
+
+
+class PolygonChain(ChainInfo, Enum):
+    POLYGON = (137, "polygon")
```
KPrasch, line 18: Since you're already in a "polygon" namespace, you can further disambiguate and just call this "mainnet" (`PolygonChain.MAINNET`)

```suggestion
    MAINNET = (137, "polygon")
```

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,70 @@
...
+class ChainInfo(NamedTuple):
+    id: int
+    name: str
+
+
+class EthChain(ChainInfo, Enum):
+    MAINNET = (1, "mainnet")
+    GOERLI = (5, "goerli")
+    SEPOLIA = (11155111, "sepolia")
+    TESTERCHAIN = (131277322940537, "eth-tester")
```
KPrasch, line 14: Can `TESTERCHAIN` be relocated to mocking / testing facilities?

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,70 @@
...
+class DomainInfo(NamedTuple):
+    name: str
+    eth_chain: EthChain
+    polygon_chain: PolygonChain
+
+    def __str__(self) -> str:
+        return self.name
+
+    def __bytes__(self) -> bytes:
+        return self.name.encode()
```
KPrasch, line 32: If we're going to access `.name` directly, these two methods are not beneficial.

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,70 @@
...
+    ORYX = DomainInfo("oryx", EthChain.GOERLI, PolygonChain.POLYGON)
+    IBEX = DomainInfo(
+        "ibex", EthChain.GOERLI, None
+    )  # this is required for configuration file migrations (backwards compatibility)
+
+    DEFAULT_DOMAIN_NAME: str = MAINNET.name
+
+    SUPPORTED_DOMAINS = [
+        MAINNET,
+        ORYX,
```
KPrasch, line 57:
```suggestion
```

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,70 @@
...
+    def is_testnet(self) -> bool:
+        return self.eth_chain != EthChain.MAINNET
+
+
+class TACoDomain:
+    class Unrecognized(RuntimeError):
+        """Raised when a provided domain name is not recognized."""
+
+    MAINNET = DomainInfo("mainnet", EthChain.MAINNET, PolygonChain.POLYGON)
+    # Testnets
```
KPrasch, line 44:
```suggestion
```

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,70 @@
...
+    LYNX = DomainInfo("lynx", EthChain.GOERLI, PolygonChain.MUMBAI)
+    TAPIR = DomainInfo("tapir", EthChain.SEPOLIA, PolygonChain.MUMBAI)
+    # TODO remove these (oryx, ibex) when appropriate
+    ORYX = DomainInfo("oryx", EthChain.GOERLI, PolygonChain.POLYGON)
+    IBEX = DomainInfo(
+        "ibex", EthChain.GOERLI, None
+    )  # this is required for configuration file migrations (backwards compatibility)
```
KPrasch, line 51: I think these can be removed - we're not going to support either of these networks post-merge.

```suggestion
```

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,70 @@
...
+    TESTERCHAIN = (131277322940537, "eth-tester")
+
+
+class PolygonChain(ChainInfo, Enum):
+    POLYGON = (137, "polygon")
+    MUMBAI = (80001, "mumbai")
+    TESTERCHAIN = (131277322940537, "eth-tester")
+
+
+class DomainInfo(NamedTuple):
```
KPrasch, line 23: This namespace seems superfluous to me and reintroduces the same kind of repetition and indirection in previous iterations.  I do think a dictionary will reduce the line count and better-suited for the way you are accessing these values.  That being said It's not a blocker since this does work... just a bit bulky IMO.

KPrasch (approved):

KPrasch (approved):
Great work - it's excellent to see the code collapse into a more natural registry/domains data model.  Lots of technical debt resolution and cleanup just in time for the release candidate 👏🏻 🤠 

~~These is still a test failure in `test_learner_learns_about_domains_separately` and perhaps we will make additional iterations the `domains.py` module, but we can address those in a follow-up into PR #3271~~

nucypher/blockchain/eth/domains.py
derekpierre: ✔️

nucypher/blockchain/eth/domains.py
derekpierre: ✔️

nucypher/blockchain/eth/domains.py
derekpierre: It can be, and I've now done that. My initial hesitation for doing that was that type wise it's not technically correct - we would only be able to create a domain with `ChainInfo` objects, and not `EthChain`/ `PolygonChain` objects because they are enums. 

In any case, the move works because of the `ChainInfo` base class.

nucypher/blockchain/eth/domains.py
derekpierre: ✔️

nucypher/blockchain/eth/domains.py
KPrasch: 👍🏻 Ultimately you've got it down to 55 lines with much better semantics then before.  I can also appreciate *explicit is better than implicit*!  LGTM

nucypher/blockchain/eth/domains.py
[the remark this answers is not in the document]
KPrasch: ✔️

nucypher/blockchain/eth/domains.py
KPrasch: ✔️

nucypher/blockchain/eth/domains.py
KPrasch: ✔️

KPrasch:
> It's strange. This test passes for me locally...I'll dig into it, but yeah it can be ignored for this PR for now.

It appears to be flaky, sometimes it passes.  Perhaps it has something to do with mocking side effects only evident with the test group executed on CI. 🤔

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,55 @@
+from enum import Enum
+from typing import NamedTuple
+
+
+class ChainInfo(NamedTuple):
+    id: int
+    name: str
+
+
+class EthChain(ChainInfo, Enum):
```
theref, line 10: how come this is a class not a variable? There'll only be one instance right? or is it useful to have multiple `EthChain`s around?

derekpierre: One instance _per ethereum chain_ . 

It's an enum - so it can only be specific values: MAINNET, GOERLI, SEPOLIA.

KPrasch: It also adds some semantic value ie `EthChain.MAINNET` vs. `PolygonChain.MAINNET`

theref: sounds good! Thank you :)

theref (approved):

examples/pre/finnegans_wake_demo/README.md
```diff
@@ -20,8 +20,7 @@ and adding your provider and wallet details.  To set the variables in your curre
 Optionally, you can change the network the demo is running on by changing the value of `L1_NETWORK` and `L2_NETWORK`.
 If you change these values be sure to also change `L1_PROVIDER_URI` and `L2_PROVIDER_URI` accordingly.
 
-Available options for `L1_NETWORK` are `tapir` or `mainnet`.
-Available options for `L2_NETWORK` are `mumbai` or `mainnet`
+Available options for `TACO_NETWORK` are `lynx`, `tapir` or `mainnet`.
```
cygnusv, line 23:
```suggestion
Available options for `TACO_DOMAIN` are `lynx`, `tapir` or `mainnet`.
```

newsfragments/3262.misc.txt
```diff
@@ -0,0 +1,3 @@
+Since the L2 network is always implied based on the TACo network connected to, we no longer need those config/parameters across the codebase, it can be inferred.
```
cygnusv, line 1: just curious why this is a .txt file instead of .rst

nucypher/blockchain/eth/domains.py
```diff
@@ -0,0 +1,55 @@
...
+        return self.eth_chain != EthChain.MAINNET
+
+
+class TACoDomain:
+    class Unrecognized(RuntimeError):
+        """Raised when a provided domain name is not recognized."""
+
+    MAINNET = DomainInfo("mainnet", EthChain.MAINNET, PolygonChain.MAINNET)
+    LYNX = DomainInfo("lynx", EthChain.GOERLI, PolygonChain.MUMBAI)
+    TAPIR = DomainInfo("tapir", EthChain.SEPOLIA, PolygonChain.MUMBAI)
```
cygnusv, line 37: Ah, so refreshing!

cygnusv (approved):
Awesome 🎸  This was a lot more work than I expected, though. Thanks for taking care of it.

KPrasch:
> Awesome 🎸 This was a lot more work than I expected, though. Thanks for taking care of it.

I agree it was more work than expected.  also reaffirms the tight coupling between `domain` and registry selection.  Of all of the technical debt areas we could address, I'm glad we decided to tackle this one now.

newsfragments/3262.misc.txt
KPrasch: Whoops!  Thanks for noticing.  Renamed it ✔️

derekpierre: Nice catch @cygnusv , thanks for fixing @KPrasch .

PR #3272, opened 2023-10-05
[WIP] Inconsistent cache usage when code updates occur

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

We've had some issues with needing to clear caches when PRs make updates to classes/module locations/names. I think it is because we don't do a pip install with our updated code whenever there is a cache hit.

This PR attempts to address/test out this theory. I don't think it's more than just the one line change, but want a CI build to test it out.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

It is annoying to manually have to clear the cache for the PR's CI build.

**Notes for reviewers:**


KPrasch (commented):
Has the recent domains intermittent failure prompted this change?  If so what makes you think this will fix it?

derekpierre:
> Has the recent domains intermittent failure prompted this change? If so what makes you think this will fix it?

Not the "domains" failure specifically. I have had some PRs recently where whenever I do some refactoring i.e. lots of moving of packages/classes around, the CI tests fail, and once I clear the github cache (https://github.com/nucypher/nucypher/actions/caches) for the corresponding PR, with no additional commits, then the CI tests pass. 

Cached dependencies only update when there are changes to the [dev-]requirements files, but not when the code itself is updated i.e. our `nucypher` code. So basically we use a stale cache from a prior test run without updating the environment with the latest code changes made in the latest commit, and then the tests complain about not finding a package/class. 

Just haven't been able to address it until now.

KPrasch (approved):
Let's give this a try

derekpierre:
@KPrasch This is an example of how this issue manifests itself - https://github.com/nucypher/nucypher/actions/runs/6430151376/job/17460583120?pr=3271. Going to clear the cache and run the tests again.

cygnusv:
> > Has the recent domains intermittent failure prompted this change? If so what makes you think this will fix it?
> 
> Not the "domains" failure specifically. I have had some PRs recently where whenever I do some refactoring i.e. lots of moving of packages/classes around, the CI tests fail, and once I clear the github cache (https://github.com/nucypher/nucypher/actions/caches) for the corresponding PR, with no additional commits, then the CI tests pass.
> 
> Cached dependencies only update when there are changes to the [dev-]requirements files, but not when the code itself is updated i.e. our `nucypher` code. So basically we use a stale cache from a prior test run without updating the environment with the latest code changes made in the latest commit, and then the tests complain about not finding a package/class.
> 
> Just haven't been able to address it until now.

But this means that we're caching `nucypher` code? Isn't that something we _don't_ want to happen anyway?

.github/workflows/python_tests.yml
```diff
@@ -39,11 +39,12 @@ jobs:
...
-      - name: Install Dependencies
+      - name: Upgrade pip
         if: steps.python_cache.outputs.cache-hit != 'true'
-        run: |
-          python -m pip install --upgrade pip
-          pip install .[dev]
+        run: python -m pip install --upgrade pip
+
+      - name: Install packages (incl. updated code)
+        run: pip install .[dev]
```
cygnusv, line 47: But I don't understand, what's the point of using caches then? Should we perhaps try to install `requirements.txt` and `dev-requirements.txt`, force caching these, and only then install `nucypher`?

manumonti (approved):
LGTM! Nice troubleshooting 🤘

cygnusv (approved):

PR #3274, opened 2023-10-06
 Running Lynx simple example as a GH action

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Refactor
- [ ] Other

**Required reviews:** 
> How many reviews does the PR author need?
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Creates a GH action that runs the testnet script on Lynx
   * Action is triggered by pushing PRs, manually, and every 20 minutes.
   * It will give us valuable information about state of Lynx testnet and API stability

**Notes for reviewers:**
Sister PR for nucypher-ts: https://github.com/nucypher/nucypher-ts/pull/316

Improvements for later:
   * The 20 minutes schedule can be reconsidered soon, once we gain a bit of confidence in both Lynx and the GH action workflow.
   * The workflow can be improved with more informative diagnosis in case of failures
   * For the moment, it uses many hard-coded values (ritual ID, encryptor signing key, RPC provider, etc). There's lot of room for improvement there too.

Waiting for #3272 to see what to do with caching of python dependencies

### DIFF
--- a/.github/workflows/lynx.yml
+++ b/.github/workflows/lynx.yml
@@ -10,6 +10,10 @@ on:
       - epic-*
   workflow_dispatch:
 
+env:
+  DEMO_L1_PROVIDER_URI: ${{ vars.DEMO_L1_PROVIDER_URI }}
+  DEMO_L2_PROVIDER_URI: ${{ vars.DEMO_L2_PROVIDER_URI }}
+
 jobs:
   python-tests:
 
@@ -51,7 +55,4 @@ jobs:
         run: python -c "import nucypher; print(nucypher.__version__)"
 
       - name: Lynx testnet script
-        env:
-          DEMO_L1_PROVIDER_URI: ${{ secrets.DEMO_L1_PROVIDER_URI }}
-          DEMO_L2_PROVIDER_URI: ${{ secrets.DEMO_L2_PROVIDER_URI }}
         run: python examples/testnet_simple_taco.py
--- a/.github/workflows/lynx.yml
+++ b/.github/workflows/lynx.yml
@@ -1,0 +1,56 @@
+name: '🔎 Lynx Testnet Example'
+
+on:
+  schedule:
+    - cron: "0/20 * * * *"  # Every 20 minutes
+  pull_request:
+    branches:
+      - main
+      - development
+      - epic-*
+  workflow_dispatch:
+
+jobs:
+  python-tests:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ "3.10" ]
+
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install latest Rust stable
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+
+      - id: setup_python
+        name: Set up Python ${{ matrix.python-version }} Environment
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - id: python_cache
+        name: Retrieve Cached Python Dependencies
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.pythonLocation }}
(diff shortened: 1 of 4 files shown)


.github/workflows/lynx.yml
```diff
@@ -0,0 +1,58 @@
...
+        uses: actions/cache@v3
+        with:
+          path: ${{ env.pythonLocation }}
+          key: ${{ runner.os }}-pip-${{ steps.setup_python.outputs.python-version }}-${{ hashFiles('dev-requirements.txt', 'requirements.txt') }}
+
+      - name: Install Dependencies
+        if: steps.python_cache.outputs.cache-hit != 'true'
+        run: |
+          python -m pip install --upgrade pip
+          pip install .[dev]
```
derekpierre, line 49: @cygnusv, FYI this step will be modified - see #3272 .

cygnusv: ACK. Let's wait for 3272 then, because I have a question there. Whatever we agree, I'll make the proper change here.

manumonti (changes requested):
Same concerns about exposing a private key that in https://github.com/nucypher/nucypher-ts/pull/316.

The rest of the code looks fine for me.

cygnusv:
> Same concerns about exposing a private key that in [nucypher/nucypher-ts#316](https://github.com/nucypher/nucypher-ts/pull/316).
> 
> The rest of the code looks fine for me.

See https://github.com/nucypher/nucypher-ts/pull/316#discussion_r1352353679

derekpierre (approved):
🎸  - looks great!

> The 20-minute schedule can be reconsidered soon, once we gain a bit of confidence in both Lynx and the GH action workflow.

Agree here - with it already running for PRs etc., once/twice a day down the road is probably sufficient.

> For the moment, it uses many hard-coded values (ritual ID, encryptor signing key, RPC provider, etc). There's lot of room for improvement there too.

The hard-coded ritual id is something we should think about further. Would be nice to perhaps use an env variable that we can update independently without needing to update the code.

.github/workflows/lynx.yml
derekpierre: Example of a failure that #3272 should resolve is the current failure from this PR python-tests run. https://github.com/nucypher/nucypher/actions/runs/6466815165/job/17555592260?pr=3274.

Without #3272 , you would need to delete the cache for PR 3274 from https://github.com/nucypher/nucypher/actions/caches and rerun the python-tests job.

manumonti (approved):
LGTM! ✌️

.github/workflows/lynx.yml
```diff
@@ -0,0 +1,58 @@
...
+        run: pip install .[dev]
+
+      - name: Check CLI Entrypoint
+        run: nucypher --help
+
+      - name: Check Python Entrypoint
+        run: python -c "import nucypher; print(nucypher.__version__)"
+
+      - name: Lynx testnet script
+        run: python examples/testnet_simple_taco.py
```
derekpierre, line 58: Side note: we could think about using `testnet_compound_multichain_taco` instead. A bit of a heavier script because of the conditions. I'm currently noticing some inconsistent behaviour with it, see #3281.

KPrasch (approved):

.github/workflows/lynx.yml
```diff
@@ -0,0 +1,58 @@
+name: '🔎 Lynx Testnet Example'
+
+on:
+  schedule:
+    - cron: "0/20 * * * *"  # Every 20 minutes
```
KPrasch, line 5: This trigger will have access to GH secrets, correct?

.github/workflows/lynx.yml
cygnusv: Yes, instead or in addition, after all is free!

.github/workflows/lynx.yml
cygnusv: I think so, but GH actions docs are sometimes not great so I'm not 100% on this.

PR #3279, opened 2023-10-10
Nodes should reject decryption requests for expired rituals

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Expired rituals should not be allowed by nodes.

Technically this was implemented at the smart contract level for defining duration, but the nodes never enforced it. It's a bug from a testnet perspective, but a feature for 7.0.0 that wasn't completed previously, hence the newsfragment is a "feature" and not "bug".

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
Otherwise, what's the point of paying for longer periods of time 😅 .

**Notes for reviewers:**

### DIFF
--- a/examples/testnet_simple_taco.py
+++ b/examples/testnet_simple_taco.py
@@ -41,7 +41,7 @@ coordinator_agent = CoordinatorAgent(
 ritual_id = 3  # got this from a side channel
 ritual = coordinator_agent.get_ritual(ritual_id)
 
-# known already authorized signer for ritual 1
+# known authorized encryptor for ritual 3
 signer = InMemorySigner(private_key=DEFAULT_TEST_ENRICO_PRIVATE_KEY)
 enrico = Enrico(
     encrypting_key=DkgPublicKey.from_bytes(bytes(ritual.public_key)), signer=signer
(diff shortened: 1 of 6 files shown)


manumonti (approved):
LGTM! Good catch! 🎣

examples/testnet_simple_taco.py
```diff
@@ -38,7 +38,7 @@
     blockchain_endpoint=polygon_endpoint,
     registry=registry,
 )
-ritual_id = 1  # got this from a side channel
+ritual_id = 3  # got this from a side channel
 ritual = coordinator_agent.get_ritual(ritual_id)
 
 # known already authorized signer for ritual 1
```
cygnusv, line 44: This mentions ritual 1. But more importantly, is this hardcoded Enrico also authorized for ritual 3?

cygnusv (approved):
Minor RFC, but apart from that, all good!

examples/testnet_simple_taco.py
derekpierre: ✔️ - updated.

> is this hardcoded Enrico also authorized for ritual 3?

It is. I used the `nucypher_dkg` script to authorize the hardcoded Enrico while testing ritual 3 - https://github.com/nucypher/nucypher/blob/development/scripts/hooks/nucypher_dkg.py#L336

Issue #3283, opened 2023-10-13
Better measure decryption time - python and taco-web


cygnusv:
- [ ] Devise proper monitoring tooling to measure decryption time with different settings (e.g., with/without Porter, different cohort sizes, etc)
- [ ] Identify major time sinks (this probably requires some sort of profiling in Bob, Ursula and Porter sides)
- [ ] ?
- [ ] Profit

cygnusv:
From current CI (https://github.com/nucypher/nucypher/actions/runs/6508418168/job/17677737409):
`Decryption time: 7.473639011383057 seconds`

derekpierre:
We expect improvements due to #3428

KPrasch:
This issue is too broad - but instead of closing it let's set a hard goal.

derekpierre:
> From current CI (https://github.com/nucypher/nucypher/actions/runs/6508418168/job/17677737409): `Decryption time: 7.473639011383057 seconds`

Updated baseline - https://github.com/nucypher/nucypher/actions/runs/7974718034/job/21771271169
```
39412 function calls (36444 primitive calls) in 4.364 seconds
```

derekpierre:
Let's come up with a better testing strategy than lynx. (`mainnet`, our own `mainnet` ritual, `taco-web` decryption time, porter efficiency etc.).

Issue #3281, opened 2023-10-12
`testnet_compound_multichain_taco` seems to be inconsistent


derekpierre:
Sometimes the script successfully executes, other times it fails due to a timeout (currently 10s).

```
Traceback (most recent call last):
  File "./examples/testnet_compound_multichain_taco.py", line 113, in <module>
    cleartext = bob.threshold_decrypt(threshold_message_kit=threshold_message_kit)
  File "/Users/derek/Documents/Github/repos/forks/derek/nucypher/nucypher/characters/lawful.py", line 753, in threshold_decrypt
    decryption_shares = self._get_decryption_shares(
  File "/Users/derek/Documents/Github/repos/forks/derek/nucypher/nucypher/characters/lawful.py", line 675, in _get_decryption_shares
    raise Ursula.NotEnoughUrsulas(
nucypher.characters.lawful.NotEnoughUrsulas: Threshold of Ursulas unable to decrypt: {}
```

Seems to be a timeout issue but more investigation needed here. Consider value of timeout used, are we losing efficiency in our logic somewhere, etc.

Somewhat related to #3155 (though not related to PRE).

derekpierre:
Related to #3283.

derekpierre:
Lots of improvements related to decryption optimization. There is also a configurable timeout. This is now done.

PR #3285, opened 2023-10-16
Reduce blockchain calls to get chain id

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

- 2 Blockchain calls were being made whenever chain id received was an integer instead of hex; this is unnecessary. Simply get the returned result of the blockchain call, then determine if it is hex or int.
- Cache chain id value once obtained for the lifetime of the `EthereumClient` since it will not change - no need to continuously make the call every time.

**Issues fixed/closed:**
> - Fixes #...
Related to #3281 , #3283 from a more general efficiency perspective.

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/blockchain/eth/clients.py
+++ b/nucypher/blockchain/eth/clients.py
@@ -139,6 +139,8 @@ class EthereumClient:
 
         self._add_default_middleware()
 
+        self.__chain_id = None
+
     def _add_default_middleware(self):
         # default retry functionality
         self.log.debug('Adding RPC retry middleware to client')
@@ -242,12 +244,17 @@ class EthereumClient:
 
     @property
     def chain_id(self) -> int:
-        try:
-            # from hex-str
-            return int(self.w3.eth.chain_id, 16)
-        except TypeError:
-            # from str
-            return int(self.w3.eth.chain_id)
+        if not self.__chain_id:
+            try:
+                # from hex-str
+                chain_id = int(self.w3.eth.chain_id, 16)
+            except TypeError:
+                # from str
+                chain_id = int(self.w3.eth.chain_id)
+
+            self.__chain_id = chain_id
+
+        return self.__chain_id
 
     @property
     def net_version(self) -> int:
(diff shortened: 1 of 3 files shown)


nucypher/blockchain/eth/clients.py
```diff
@@ -242,12 +244,18 @@ def set_gas_strategy(self, gas_strategy):
...
+            try:
+                # from hex-str
+                chain_id = int(result, 16)
+            except TypeError:
+                # from str
+                chain_id = int(result)
+
+            self.__chain_id = chain_id
+
+        return self.__chain_id
```
KPrasch, line 258: Use `functools`?

```suggestion
    @functools.cached_property
    def chain_id(self) -> int:
        result = self.w3.eth.chain_id
        try:
            chain_id = int(result, 16)
        except TypeError:
            chain_id = int(result)
        return chain_id
```

derekpierre: ✔️  - 🎸

cygnusv:
I think this is fine, but maybe we can reintroduce the simple cache middleware, since chain id is one of the endpoints that are cached there. See 
https://github.com/ethereum/web3.py/blob/716054896d421ff3355fa8daf6089ea3ec509bcf/web3/middleware/cache.py#L40

This cache was removed from our client due to a bug in web3py that we found (https://github.com/ethereum/web3.py/issues/2977) but this was fixed in v6.5 and we're currently running v6.9, so it should be fine.

derekpierre:
> I think this is fine, but maybe we can reintroduce the simple cache middleware, since chain id is one of the endpoints that are cached there. See https://github.com/ethereum/web3.py/blob/716054896d421ff3355fa8daf6089ea3ec509bcf/web3/middleware/cache.py#L40
> 
> This cache was removed from our client due to a bug in web3py that we found ([ethereum/web3.py#2977](https://github.com/ethereum/web3.py/issues/2977)) but this was fixed in v6.5 and we're currently running v6.9, so it should be fine.

Fair, we can look into that as a follow-up - #3286

KPrasch (approved):

piotr-roslaniec (approved):

PR #3288, opened 2023-10-17
Automation: docker publication triggered by git tagging

KPrasch:


.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
...
+      with:
+        username: ${{ secrets.DOCKER_USERNAME }}
+        password: ${{ secrets.DOCKER_PASSWORD }}
+
+    - name: Extract Tag Name
+      run: echo "VERSION=${GITHUB_REF#refs/tags/}" >> $GITHUB_ENV
+
+    - name: Build and push Docker image
+      run: |
+        docker-compose build --tag nucypher/nucypher:$VERSION .
```
derekpierre, line 32: Instead of VERSION, I think you can just use `github.ref_name` assuming the filter is only for tags - https://docs.github.com/en/actions/learn-github-actions/contexts#github-context.

derekpierre (approved):
🎸

PR #3298, opened 2023-10-23
Decouples Decryption from Server Logic

### LINKED ISSUE #3299: Move threshold_decrypt logic from server.py to Ursula

KPrasch:
**Type of PR:**
Bugfix

**Required reviews:** 
3

**What this does:**
- Fixes #3299 
- Catch errors while requesting tdec from nodes (during `derive_decryption_share`) and return them as an HTTP response.
- Ensures that an HTTP response is always produced even in the case of unexpected errors
- Improves the ability to create effective unit tests of core business logic

**Why it's needed:**
- Nodes are not explicitly handling decryption errors which makes them noisy and feedback less useful.
- It's too difficult to unit test logic that must be executed inside a running flask server

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1,3 +1,4 @@
+import json
 import time
 from collections import defaultdict
 from decimal import Decimal
@@ -51,6 +52,7 @@ from nucypher.crypto.powers import (
 )
 from nucypher.datastore.dkg import DKGStorage
 from nucypher.policy.conditions.evm import _CONDITION_CHAINS
+from nucypher.policy.conditions.utils import evaluate_condition_lingo
 from nucypher.policy.payment import ContractPayment
 from nucypher.utilities.emitters import StdoutEmitter
 from nucypher.utilities.logging import Logger
@@ -595,7 +597,7 @@ class Operator(BaseActor):
         )
 
 
-    def _authorize_ritual(self, decryption_request: ThresholdDecryptionRequest):
+    def _verify_active_ritual(self, decryption_request: ThresholdDecryptionRequest):
         # TODO: #3052 consider using the DKGStorage cache instead of the coordinator agent
         # dkg_public_key = this_node.dkg_storage.get_public_key(decryption_request.ritual_id)
         ritual = self.coordinator_agent.get_ritual(
@@ -616,10 +618,10 @@ class Operator(BaseActor):
                 f"Ritual {decryption_request.ritual_id} is expired",
             )
 
-    def _authorize_ciphertext(
+    def _verify_ciphertext_authorization(
         self, decryption_request: ThresholdDecryptionRequest
     ) -> None:
-        # check whether enrico is authorized
+        """check that the ciphertext is authorized for this ritual"""
         ciphertext_header = decryption_request.ciphertext_header
         authorization = decryption_request.acp.authorization
         if not self.coordinator_agent.is_encryption_authorized(
@@ -631,17 +633,52 @@ class Operator(BaseActor):
                 f"Encrypted data not authorized for ritual {decryption_request.ritual_id}",
             )
 
-    def _authorize_decryption_request(
+    def _evaluate_conditions(
         self, decryption_request: ThresholdDecryptionRequest
     ) -> None:
-        self._authorize_ritual(decryption_request)
-        self._authorize_ciphertext(decryption_request)
+        # requester-supplied condition eval context
+        context = None
+        if decryption_request.context:
+            # nucypher_core.Context -> str -> dict
+            context = json.loads(str(decryption_request.context)) or dict()
+
+        # obtain condition from request
+        condition_lingo = json.loads(
+            str(decryption_request.acp.conditions)
+        )  # nucypher_core.Conditions -> str -> Lingo
+        if not condition_lingo:
+            # this should never happen for CBD - defeats the purpose
+            raise self.UnauthorizedRequest(
+                "No conditions present for ciphertext - invalid for CBD functionality",
+            )
+
+        # evaluate the conditions for this ciphertext
+        error = evaluate_condition_lingo(
+            condition_lingo=condition_lingo,
+            context=context,
+            providers=self.condition_providers,
+        )
+        if error:
+            raise self.UnauthorizedRequest(
+                f"Condition evaluation failed: {error}",
+            )
+
+    def _verify_decryption_request_authorization(
+        self, decryption_request: ThresholdDecryptionRequest
+    ) -> None:
+        """check that the decryption request is authorized for this ritual"""
+        self._verify_active_ritual(decryption_request)
+        self._verify_ciphertext_authorization(decryption_request)
+        self._evaluate_conditions(decryption_request)
 
     def _derive_decryption_share_for_request(
         self,
         decryption_request: ThresholdDecryptionRequest,
     ) -> Union[DecryptionShareSimple, DecryptionSharePrecomputed]:
-        self._authorize_decryption_request(decryption_request=decryption_request)
+        """Derive a decryption share for a given request"""
+        self._verify_decryption_request_authorization(
+            decryption_request=decryption_request
+        )
         try:
             decryption_share = self.derive_decryption_share(
                 ritual_id=decryption_request.ritual_id,
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -135,6 +135,15 @@ class Operator(BaseActor):
     class OperatorError(BaseActor.ActorError):
         """Operator-specific errors."""
 
+    class RitualNotFoundException(Exception):
+        """Ritual is not found."""
+
+    class UnauthorizedRequest(Exception):
+        """Request is not authorized."""
+
+    class DecryptionFailure(Exception):
+        """Decryption failed."""
+
     def __init__(
         self,
         eth_endpoint: str,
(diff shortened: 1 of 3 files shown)


cygnusv:
How about addressing https://github.com/nucypher/nucypher/issues/3299 here?

KPrasch:
> How about addressing #3299 here?

🖖🏻 see 2fb66472e7a8dc865c0f023ed76f26a9bee71426

jMyles (approved):
My qualm is that we have a long, proud tradition of burying the quintessence of our innovation inside the logic of HTTP routing, and I'm just not sure now is the time for clarity and good sense.

cygnusv:
> My qualm is that we have a long, proud tradition of burying the quintessence of our innovation inside the logic of HTTP routing, and I'm just not sure now is the time for clarity and good sense.

Make `server.py` great again

nucypher/blockchain/eth/actors.py
```diff
@@ -585,6 +594,83 @@ def encrypt_threshold_decryption_response(
...
+        # TODO: #3052 consider using the DKGStorage cache instead of the coordinator agent
+        # dkg_public_key = this_node.dkg_storage.get_public_key(decryption_request.ritual_id)
+        ritual = self.coordinator_agent.get_ritual(
+            decryption_request.ritual_id, with_participants=True
+        )
+
+        # enforces that the node is part of the ritual
+        participants = [p.provider for p in ritual.participants]
+        if self.checksum_address not in participants:
+            raise self.UnauthorizedRequest(
```
derekpierre, line 608: Perhaps we separate this case from the case when the encryptor is not authorized by using a different exception?

Same for expired ritual?

nucypher/blockchain/eth/actors.py
```diff
@@ -585,6 +594,83 @@ def encrypt_threshold_decryption_response(
...
+            )
+
+        # check that ritual not timed out
+        current_timestamp = self.coordinator_agent.blockchain.get_blocktime()
+        if current_timestamp > ritual.end_timestamp:
+            raise self.UnauthorizedRequest(
+                f"Ritual {decryption_request.ritual_id} is expired",
+            )
+
+    def _authorize_ciphertext(
```
derekpierre, line 619: This sounds like an action vs a check. Perhaps something like `_check_ciphertext_authorization(...)`?

nucypher/blockchain/eth/actors.py
```diff
@@ -585,6 +594,83 @@ def encrypt_threshold_decryption_response(
...
+        if not self.coordinator_agent.is_encryption_authorized(
+            ritual_id=decryption_request.ritual_id,
+            evidence=authorization,
+            ciphertext_header=bytes(ciphertext_header),
+        ):
+            raise self.UnauthorizedRequest(
+                f"Encrypted data not authorized for ritual {decryption_request.ritual_id}",
+            )
+
+    def _authorize_decryption_request(
```
derekpierre, line 634: Similar comment as above. Is this more like a check/validation than "authorize" action?

nucypher/characters/lawful.py
```diff
@@ -1364,6 +1367,55 @@ def as_external_validator(self) -> Validator:
...
+            raise self.UnauthorizedRequest(
+                f"Condition evaluation failed: {error}",
+            )
+        return condition_lingo
+
+    def _authorize_decryption_request(
+        self, decryption_request: ThresholdDecryptionRequest
+    ) -> None:
+        super()._authorize_decryption_request(decryption_request)
+        self._authorize_lingo(decryption_request)
```
derekpierre, line 1403: What about calling `_authorize_lingo(...)` before calling `self._derive_decryption_share_request(...)` below, then you don't need to override `_authorize_decryption_request`?

nucypher/characters/lawful.py
```diff
@@ -1364,6 +1367,55 @@ def as_external_validator(self) -> Validator:
...
+        error = evaluate_condition_lingo(
+            condition_lingo=condition_lingo,
+            context=context,
+            providers=self.condition_providers,
+        )
+        if error:
+            raise self.UnauthorizedRequest(
+                f"Condition evaluation failed: {error}",
+            )
+        return condition_lingo
```
derekpierre, line 1397: Is this return needed?

nucypher/network/server.py
```diff
@@ -148,106 +147,25 @@ def condition_chains():
...
-        participants = [p.provider for p in ritual.participants]
-
-        # enforces that the node is part of the ritual
-        if this_node.checksum_address not in participants:
-            return Response(
-                f"Node not part of ritual {decryption_request.ritual_id}",
-                status=HTTPStatus.FORBIDDEN,
+            response = this_node.handle_threshold_decryption_request(encrypted_request)
+            response = Response(
+                response=bytes(response),
```
derekpierre, line 156:
```suggestion
            decryption_response = this_node.handle_threshold_decryption_request(encrypted_request)
            response = Response(
                response=bytes(decryption_response),
```

nucypher/network/server.py
```diff
@@ -148,106 +147,25 @@ def condition_chains():
...
-            bytes(encrypted_response),
-            headers={"Content-Type": "application/octet-stream"},
-        )
+            return response
+        except this_node.RitualNotFoundException:
+            return Response("Ritual not found", status=HTTPStatus.NOT_FOUND)
+        except this_node.UnauthorizedRequest as e:
+            return Response(str(e), status=HTTPStatus.UNAUTHORIZED)
+        except this_node.DecryptionFailure as e:
+            return Response(str(e), status=HTTPStatus.INTERNAL_SERVER_ERROR)
```
derekpierre, line 166: Previously, you had this as BAD_REQUEST, did you change your mind?

derekpierre (commented):
Love the direction.

Some comments around the use of `_authorize` in method names.

nucypher/blockchain/eth/actors.py
KPrasch: Unless we're going to raise a different HTTP status codes for those situations I don't think expanding the number of exception classes will have any benefit.  Expired rituals and threshold decryption requests sent to the wrong node are correctly understood as  *Unauthorized*, also the error message will be passed along with the HTTP response to hint to the requester what went wrong (ie 403 `Encrypted data not authorized for ritual x` )

nucypher/blockchain/eth/actors.py
KPrasch: Naming things... 😅 I can see how it might sound confusing since we use "authorize" to mean granting access via the public API. In a sense this the moment when enrico's authorization is actually enforced: ursula is guarding it's private materials and grants the request to use the cryptographic backend.  To me that sounds like an "authorization"... but perhaps "verify"?  it invokes cryptographic ideas.

```python
def _verify_ciphertext_authorization(...)
```

nucypher/characters/lawful.py
KPrasch: no, I'll remove it

nucypher/characters/lawful.py
KPrasch: I attempted to integrate this logic into the existing layers by keeping each block as local to it's dependencies as possible.   This essentially resulted in blockchain interactions on `Operator` and high-level `ThresholdDecryptionRequest|Response` handling on `Ursula`.  Also, by embedding the authorization call inside the derivation it's more misuse resistant -- `derive_decryption_share_for_request(decryption_request)` handles everything (I suppose we can consider this a public API since it's directly exposed by the webserver).

nucypher/network/server.py
KPrasch: agreed - I'll make an update.

nucypher/network/server.py
KPrasch: I'm unsure about it.  At least the server routing is focused here and we can deliberate about it :-)  Opinions?

nucypher/blockchain/eth/actors.py
derekpierre: 👌

nucypher/network/server.py
derekpierre: IMHO, I liked the change to `BAD_REQUEST` - a 500 seemed harsh.

There's always 418 - https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/418 (kidding 😅 ).

KPrasch: So by switching to a 4xx status code we're telling the client that it's their problem.  Will this hold true?  It's always bad input that causes `DecryptionFailure` to be raised?

nucypher/characters/lawful.py
derekpierre: Ah Interesting. Given that philosophy, since evaluating condition lingo requires blockchain calls, would it make sense to move `authorize_lingo` functionality into actor.py as well?

KPrasch: to be honest I'm not sure it's a valid approach it turned out imperfect and had abstraction leaks that I cannot identify yet.  The main reason for moving lingo here is that is that condition providers is local to Ursula. 🤔

nucypher/network/server.py
derekpierre: It's tough to know why the decryption failed unless there are specific exceptions returned from `nucypher-core`. Perhaps @piotr-roslaniec / @cygnusv can shed some light here? In the vast majority of cases for a decryption failure, would the reason be a bad request vs some other factor?

KPrasch: By the time the `DecryptionFailure` can be raised, protocol objects are already de serialized, validated, the ritual and the ciphertext authorization are already checked.

derekpierre: Right. Then the 500 seems more fitting.

nucypher/characters/lawful.py
KPrasch: Ah yes, you're correct! I'll attempt a slight refactor

nucypher/blockchain/eth/actors.py
KPrasch: ✔️

nucypher/characters/lawful.py
KPrasch: ✔️

nucypher/blockchain/eth/actors.py
```diff
@@ -585,6 +596,118 @@ def encrypt_threshold_decryption_response(
...
+        self._verify_active_ritual(decryption_request)
+        self._verify_ciphertext_authorization(decryption_request)
+        self._evaluate_conditions(decryption_request)
+
+    def _derive_decryption_share_for_request(
+        self,
+        decryption_request: ThresholdDecryptionRequest,
+    ) -> Union[DecryptionShareSimple, DecryptionSharePrecomputed]:
+        """Derive a decryption share for a given request"""
+        self._verify_decryption_request_authorization(
```
derekpierre, line 679: 💥 - I love how this reads.

nucypher/blockchain/eth/actors.py
```diff
@@ -585,6 +596,118 @@ def encrypt_threshold_decryption_response(
...
+                "No conditions present for ciphertext - invalid for CBD functionality",
+            )
+
+        # evaluate the conditions for this ciphertext
+        error = evaluate_condition_lingo(
+            condition_lingo=condition_lingo,
+            context=context,
+            providers=self.condition_providers,
+        )
+        if error:
```
derekpierre, line 661: We need to be careful here. There is some rich information in this `error`. It is an `EvalError` type that includes a recommended status code. Many are BAD_REQUEST, but some are FORBIDDEN, NOT_IMPLEMENTED eg. https://github.com/nucypher/nucypher/blob/v7.0.0/nucypher/policy/conditions/utils.py#L85.

Perhaps we need to also update what `evaluate_condition_lingo` does/returns in the case of a problem...

nucypher/network/server.py
```diff
@@ -148,106 +147,27 @@ def condition_chains():
...
-        )
+            return response
+        except this_node.RitualNotFoundException:
+            return Response("Ritual not found", status=HTTPStatus.NOT_FOUND)
+        except this_node.UnauthorizedRequest as e:
+            return Response(str(e), status=HTTPStatus.UNAUTHORIZED)
+        except this_node.DecryptionFailure as e:
+            return Response(str(e), status=HTTPStatus.BAD_REQUEST)
+        except Exception as e:
+            return Response(str(e), status=HTTPStatus.INTERNAL_SERVER_ERROR)
```
derekpierre, line 170: You've convinced me that `DecryptionFailure` should be a 500.

```suggestion
        except this_node.DecryptionFailure as e:
            return Response(str(e), status=HTTPStatus.INTERNAL_SERVER_ERROR)
        except Exception as e:
            return Response(str(e), status=HTTPStatus.INTERNAL_SERVER_ERROR)
```

nucypher/blockchain/eth/actors.py
KPrasch: Pushed an update that allows `EvalError` to bubble up the the server routing.

nucypher/network/server.py
KPrasch: ✔️

nucypher/network/server.py
KPrasch: ✔️

derekpierre (approved):
🎸

PR #3304, opened 2023-10-23
Use `v*.*.*` branches to trigger github actions instead of `development`

KPrasch:


.github/workflows/darker.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches:
       - main
-      - development
+      - v*
```
derekpierre, line 7: Did you want to do `v*.*.*`? Otherwise any branch starting with `v` will match...

KPrasch: tbh I like the simplicity and we can likely avoid other types of upstream branches that start with "v".

KPrasch: updated to use `v*.*.*` ✔️

KPrasch:
> Does the branch therefore symbolize the major/minor version, but tags would now only be used for updates to the patch number...?

We can still tag both minor and patch releases as they get merged.  Patch (bugfix) commits can use a PR based directly over the version they are patching.

derekpierre (approved):
🎸

PR #3308, opened 2023-10-24
Relock deps

manumonti:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Relock dependencies


Pipfile
```diff
@@ -26,9 +26,10 @@ msgpack = "*"
 aiohttp = "==3.8.2"
 flask = "*"
 requests = "*"
+urllib3 = ">=1.26.16,<2" # because of eth-ape
```
derekpierre, line 29: Was this change and the one below necessary? Or could the version of `eth-ape` just be limited to "<=0.6.19".

I'd much rather limit a dev-dependency that a regular dependency.

manumonti: Yes, I agree. We limited eth-ape to <=0.6.19 and it worked for urllib3 limitation. But we still needed to limit py-evm to <0.8 because of evm-trace

derekpierre: Ah ok 👍

KPrasch (approved):
🤠

PR #3311, opened 2023-10-24
Sunset Legacy Docs

KPrasch:
Documentation has been relocated to https://docs.threshold.network (https://github.com/threshold-network/threshold).  

This PR remove the legacy documentation source code, and the package build configurations.  Release notes and an image are preserved in `nucypher/docs`.

**Requested Reviews**
3


piotr-roslaniec:
WDYT about turning some of these files into Markdown files in the repo, for example `CONTRIBUTING.md`

KPrasch:
> WDYT about turning some of these files into Markdown files in the repo, for example `CONTRIBUTING.md`

Indeed, I think there's a treasure trove of good inspiration in these docs... think of this as a fresh start. we'll extract what we need while relocating it to the appropriate home :-)

README.md
```diff
@@ -6,64 +6,10 @@
 [![pyversions](https://img.shields.io/pypi/pyversions/nucypher.svg)](https://pypi.org/project/nucypher/)
 [![codecov](https://codecov.io/gh/nucypher/nucypher/branch/development/graph/badge.svg?token=9w1dAFqog5)](https://codecov.io/gh/nucypher/nucypher)
 [![discord](https://img.shields.io/discord/411401661714792449.svg?logo=discord)](https://discord.gg/7rmXa3S)
-[![Documentation Status](https://readthedocs.org/projects/nucypher/badge/?version=latest)](https://nucypher.readthedocs.io/en/latest/)
 [![license](https://img.shields.io/pypi/l/nucypher.svg)](https://www.gnu.org/licenses/gpl-3.0.html)
 
 ----
 
-The NuCypher network provides accessible, intuitive, and extensible runtimes and interfaces
```
derekpierre, line 14: If removing this then you might as well remove the rest of the README i.e. Whitepapers, Network, Economics, Cryptography.

That being said instead of leaving it empty you can either link to the Threshold Docs eg. https://docs.threshold.network/applications/threshold-access-control, AND/OR copy the same blurb from the link.

My preference would be to copy the same blurb form the link here, and link to the docs for "more information" or something like that.

derekpierre (commented):
Where did the release notes rst from docs end up? I think we can keep that in the top-level directory for now.

README.md
KPrasch: ✔️

CONTRIBUTING.rst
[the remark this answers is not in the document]
KPrasch: ✔️

KPrasch:
> Where did the release notes rst from docs end up? I think we can keep that in the top-level directory for now.

I left them in place at `docs/release_notes` as to not disturb the towncrier configuration.

README.md
```diff
@@ -1,112 +1,46 @@
...
 NuCypher is a community-driven project and we're very open to outside contributions.
 
-All our development discussions happen in our [Discord server](https://discord.gg/7rmXa3S), where we're happy to answer technical questions, discuss feature requests,
+All our development discussions happen in our [Discord server](https://discord.gg/7rmXa3S), where we're happy to answer
+technical questions, discuss feature requests,
 and accept bug reports.
 
-If you're interested in contributing code, please check out our [Contribution Guide](https://docs.nucypher.com/en/latest/support/contribution.html)
+If you're interested in contributing code, please check out
+our [Contribution Guide](https://docs.nucypher.com/en/latest/support/contribution.html)
```
derekpierre, line 30: I guess some of these links need to be updated. This and some below.

derekpierre:
> > Where did the release notes rst from docs end up? I think we can keep that in the top-level directory for now.
> 
> I left them in place at `docs/release_notes` as to not disturb the towncrier configuration.

We could move it to the top-level directory and just modify the path here - https://github.com/nucypher/nucypher/blob/v7.0.0/pyproject.toml#L4.

Dealer's choice 👍

KPrasch:
> > > Where did the release notes rst from docs end up? I think we can keep that in the top-level directory for now.
> > 
> > 
> > I left them in place at `docs/release_notes` as to not disturb the towncrier configuration.
> 
> We could move it to the top-level directory and just modify the path here - https://github.com/nucypher/nucypher/blob/v7.0.0/pyproject.toml#L4.
> 
> Dealer's choice 👍

✔️

piotr-roslaniec (approved):

README.md
```diff
@@ -1,119 +1,52 @@
-![](./docs/source/.static/img/nucypher.png)
+![](nucypher.png)
 
-*A decentralized cryptological network offering accessible, intuitive, and extensible runtimes and interfaces for secrets management and dynamic access control.*
+*A decentralized cryptological network offering accessible, intuitive, and extensible runtimes and interfaces for
```
arjunhassard, line 3: Nit-picking, but do we still want to call ourselves a 'cryptological network'? Since NuCypher is a contributor to a bigger network with services that aren't exclusively cryptographic primitives (even if they're based on primitives, that's not how they're consumed), and also 'cryptological' implies that there's both a cryptographic and cryptanalysis component to the service (which would make more sense if we were white hats proving vulnerabilities in SGX or something, rather than just loosely that we analyze the security of our own offerings).

README.md
```diff
@@ -1,119 +1,52 @@
...
+secrets management and dynamic access control.*
```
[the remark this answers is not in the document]
KPrasch: Agree - to be clear this is not new/original.  It must have been reformatted in the shuffle.  I welcome an overhaul of this README I do feel that it needs refreshing.  Main aim here is to retire the stale docs ans start fresh :-)

PR #3271, opened 2023-10-04
[Epic] Remodel Domains

KPrasch:
- #3261 
- #3262 
- #3275

PR #3315, opened 2023-10-30
Fully Integrate `TACoDomain`

KPrasch:
**Type of PR:**
Rework

**Required reviews:** 
2

**What this does:**
- Imagine this as the 2nd submission of a 3-part PR: 
    1. follows #3271
    2. this PR
    3. continued in #3318 
- integrates `TACoDomain` with internal/external APIs 
- Maintains backwards compatibility for string domains for characters (strings are eagerly converted to instances for character creation)
- collapses `DomainInfo` into `TACoDomain`
- simplifies `domains.py` for use as module-level constant (new conventional usage `from nucypher.blockchain.eth import domains`)
- renames `<type>.taco_domain_info` -> `<type>.domain` for `Ursula`, `Character`, `Actor`, `ContractRegistry`, etc.
- renames `TEMPORARY_DOMAIN` -> `TEMPORARY_DOMAIN_NAME`
- resolves several codebase `TODO`s

Example usage of `domains`:
```python
from nucypher.blockchain.eth import domains

# any of these will work
domain = 'lynx'
domain = domains.LYNX
domain = TACoDomain(...)

# duck typing with str()
domain: TACoDomain = domains.get_domain(str(domain))
```

Note: Based over #3317

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1,8 +1,9 @@
+import time
 from collections import defaultdict
+from decimal import Decimal
+from typing import DefaultDict, Dict, List, Optional, Set, Tuple, Union
 
 import maya
-import time
-from decimal import Decimal
 from eth_typing import ChecksumAddress
 from hexbytes import HexBytes
 from nucypher_core import (
@@ -22,12 +23,10 @@ from nucypher_core.ferveo import (
     Transcript,
     Validator,
 )
-from typing import DefaultDict, Dict, List, Optional, Set, Tuple, Union
 from web3 import HTTPProvider, Web3
 from web3.types import TxReceipt
 
 from nucypher.acumen.nicknames import Nickname
-from nucypher.blockchain.eth import domains
 from nucypher.blockchain.eth.agents import (
     ContractAgency,
     CoordinatorAgent,
@@ -39,9 +38,7 @@ from nucypher.blockchain.eth.constants import NULL_ADDRESS
 from nucypher.blockchain.eth.decorators import validate_checksum_address
 from nucypher.blockchain.eth.domains import TACoDomain
 from nucypher.blockchain.eth.interfaces import BlockchainInterfaceFactory
-from nucypher.blockchain.eth.registry import (
-    ContractRegistry,
-)
+from nucypher.blockchain.eth.registry import ContractRegistry
 from nucypher.blockchain.eth.signers import Signer
 from nucypher.blockchain.eth.trackers import dkg
 from nucypher.blockchain.eth.trackers.bonding import OperatorBondedTracker
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -248,22 +248,14 @@ class Operator(BaseActor):
     ) -> DefaultDict[int, Set[HTTPProvider]]:
         """Multi-provider support"""
 
-        # If condition_blockchain_endpoints is None the node operator
-        # did not configure any additional condition providers.
         condition_blockchain_endpoints = condition_blockchain_endpoints or dict()
-
-        # These are the chains that the Operator will connect to for conditions evaluation (read-only).
         condition_providers = defaultdict(set)
 
-        # Now, add any additional providers that were passed in.
         for (
             chain_id,
             condition_blockchain_endpoints,
         ) in condition_blockchain_endpoints.items():
             if not self._is_permitted_condition_chain(chain_id):
-                # this is a safety check to prevent the Operator from connecting to
-                # chains that are not supported by ursulas on the network;
-                # Prevents the Ursula/Operator from starting up if this happens.
                 raise NotImplementedError(
                     f"Chain ID {chain_id} is not supported for condition evaluation by this Operator."
                 )
@@ -272,10 +264,8 @@ class Operator(BaseActor):
             for uri in condition_blockchain_endpoints:
                 provider = self._make_condition_provider(uri)
                 providers.add(provider)
-
             condition_providers[int(chain_id)] = providers
 
-        # Log the chains that the Operator is connected to.
         humanized_chain_ids = ", ".join(
             _CONDITION_CHAINS[chain_id] for chain_id in condition_providers
         )
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -37,6 +37,7 @@ from nucypher.blockchain.eth.agents import (
 from nucypher.blockchain.eth.clients import PUBLIC_CHAINS
 from nucypher.blockchain.eth.constants import NULL_ADDRESS
 from nucypher.blockchain.eth.decorators import validate_checksum_address
+from nucypher.blockchain.eth.domains import TACoDomain
 from nucypher.blockchain.eth.interfaces import BlockchainInterfaceFactory
 from nucypher.blockchain.eth.registry import (
     ContractRegistry,
@@ -69,7 +70,7 @@ class BaseActor:
     @validate_checksum_address
     def __init__(
         self,
-        domain: str,
+        domain: TACoDomain,
         registry: ContractRegistry,
         transacting_power: Optional[TransactingPower] = None,
         checksum_address: Optional[ChecksumAddress] = None,
@@ -92,7 +93,7 @@ class BaseActor:
 
         self.transacting_power = transacting_power
         self.registry = registry
-        self.domain = domains.get_domain(str(domain))
+        self.domain = domain
         self._saved_receipts = list()  # track receipts of transmitted transactions
 
     def __repr__(self):
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1,9 +1,8 @@
-import time
 from collections import defaultdict
-from decimal import Decimal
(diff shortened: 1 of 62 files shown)


nucypher/network/nodes.py
```diff
@@ -970,13 +968,10 @@ class Teacher:
     __DEFAULT_MIN_SEED_STAKE = 0
 
     def __init__(self,
-                 domain: str,  # TODO: Consider using a Domain type
```
KPrasch, line 973: ✔️

nucypher/characters/base.py
```diff
@@ -33,7 +34,7 @@ class Character(Learner):
 
     def __init__(
         self,
-        domain: str,
+        domain: Union[str, TACoDomain],
```
derekpierre, line 37: Any reason we don't just take a `TACoDomain` instead of a Union that includes str?

nucypher/config/base.py
```diff
@@ -834,7 +832,7 @@ def configure_pre_payment_method(self):
         if pre_payment_class.ONCHAIN:
             # on-chain payment strategies require a blockchain connection
             pre_payment_strategy = pre_payment_class(
-                domain=self.taco_domain_info.name,
+                domain=str(self.domain),
```
derekpierre, line 835: Any reason we don't want to just pass the `TACoDomain` object instead of the str?

nucypher/network/nodes.py
```diff
@@ -281,7 +278,9 @@ def __init__(self,
 
         self._abort_on_learning_error = abort_on_learning_error
 
-        self.__known_nodes = self.tracker_class(domain=domain, this_node=self if include_self_in_the_state else None)
+        self.__known_nodes = self.tracker_class(
+            domain=self.domain, this_node=self if include_self_in_the_state else None
```
derekpierre, line 282: Where is `self.domain` set? If not set above, this may also have effects on Porter which is a `Learner`.

nucypher/blockchain/eth/actors.py
```diff
@@ -248,22 +248,14 @@ def connect_condition_providers(
...
-        # Now, add any additional providers that were passed in.
         for (
             chain_id,
             condition_blockchain_endpoints,
         ) in condition_blockchain_endpoints.items():
             if not self._is_permitted_condition_chain(chain_id):
-                # this is a safety check to prevent the Operator from connecting to
-                # chains that are not supported by ursulas on the network;
-                # Prevents the Ursula/Operator from starting up if this happens.
                 raise NotImplementedError(
```
piotr-roslaniec, line 259: Functionally nothing changed, it's just that we've removed verbose comments, correct?

nucypher/network/nodes.py
```diff
@@ -46,13 +46,13 @@
 from nucypher.utilities.logging import Logger
 
 TEACHER_NODES = {
-    TACoDomain.MAINNET.name: (
+    domains.MAINNET: (
```
derekpierre, line 49: There are a few spots in the code that expect the key for the dictionary to be a string - eg. https://github.com/nucypher/nucypher/blob/b9e1a2705ac592cd423c815ce8c250865d0d9b8e/nucypher/characters/lawful.py#L1181.

Is that still valid?

nucypher/characters/base.py
KPrasch: That will work, doing so will reduce the usage of domains as strings even more, perhaps only exposed by config.   The initial thought was backwards-compatibility but perhaps that's misplaced here.

nucypher/config/base.py
KPrasch: I must have missed this one in the refactor -- I'll take a look!

nucypher/network/nodes.py
KPrasch: Great question == right now it's set in `Character.__init__`. I'm trying to be strategic about where to do the str -> domain lookup, perhaps porter can handle this on it's own?

derekpierre: Ignoring Porter, logically it feels like the `domain` is appropriately a part of `Learner`...i.e. the existential question is, *what is a learner without a domain*?

nucypher/characters/base.py
derekpierre: Just wondering how "integrated" the object can be...if you look at the code and it doesn't make sense in some areas, that's fair.

nucypher/config/base.py
KPrasch: ✔️

nucypher/network/nodes.py
KPrasch: ✔️

nucypher/characters/base.py
KPrasch: ✔️

nucypher/blockchain/eth/domains.py
```diff
@@ -18,38 +24,69 @@ class PolygonChain(ChainInfo, Enum):
...
+
+    def __hash__(self) -> int:
+        return hash(self.name)
+
+    def __bytes__(self) -> bytes:
+        return self.name.encode()
+
+    def __eq__(self, other: Any) -> bool:
+        try:
+            return self.name == other.name
```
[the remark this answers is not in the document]
KPrasch: As long as we're avoiding name collisions what's the benefit of a more complex equality check?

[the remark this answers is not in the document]
derekpierre: I guess the question I have is, if `TACoDomain` is a public class, what is stopping someone from creating their own `TACoDomain` object by doing the following (perhaps by mistake):
```python
myDomain = TACoDomain("mainnet", EthChain.GOERLI, PolygonChain.MUMBAI)
...
```
and passing it around as the domain? If there is indeed something that prevents that, then great!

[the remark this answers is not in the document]
KPrasch: Nothing is preventing that however that seems like acceptable behavior to me...?  (technically any object with a matching name attribute will eval truthy) It's pretty hard misuse by mistake imo, as you've demonstrated you'd need to be interacting with domain construction/mutation to begin with, then set the name incorrectly.  Changing it may have subtle side-effects... I need to think about it a bit

manumonti (changes requested):
Great work! I left some questions

tests/conftest.py
```diff
@@ -149,13 +143,6 @@ def mock_condition_blockchains(module_mocker):
         "nucypher.policy.conditions.evm._CONDITION_CHAINS",
         {TESTERCHAIN_CHAIN_ID: "eth-tester/pyevm"},
     )
```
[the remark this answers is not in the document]
KPrasch: good eye -- this is intentional design to support running a subset of tests while still loading specific fixtures in different execution environments.

nucypher/blockchain/eth/domains.py
```diff
@@ -18,38 +24,72 @@ class PolygonChain(ChainInfo, Enum):
...
 
-    @classmethod
-    def get_domain_info(cls, domain: str) -> DomainInfo:
-        for taco_domain in cls.SUPPORTED_DOMAINS:
-            if taco_domain.name == domain:
-                return taco_domain
 
-        raise cls.Unrecognized(f"{domain} is not a recognized domain.")
+@memoize
+def get_domain(d: Any) -> TACoDomain:
```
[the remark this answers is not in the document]
KPrasch: since I'm type checking on the first line of the function body I'm entertaining the idea that this public function may receive an invalid input that is not a string.

nucypher/blockchain/eth/domains.py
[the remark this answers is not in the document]
KPrasch: ✔️  after carefully checking it seems a more complex check has desirable effects!  This is a more correct and rigid design especially since we're able to compare enum instances directly.  Need to think about expanding this check in #3318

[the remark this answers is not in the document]
derekpierre: 👍

derekpierre (approved):
🎸

nucypher/blockchain/eth/domains.py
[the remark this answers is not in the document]
KPrasch: (also updated `__hash__` for consistency)

tests/integration/config/test_storages.py
```diff
@@ -2,11 +2,11 @@
 
 from nucypher.blockchain.eth.signers.software import Web3Signer
 from nucypher.characters.lawful import Ursula
-from nucypher.config.constants import TEMPORARY_DOMAIN
+from nucypher.config.constants import TEMPORARY_DOMAIN_NAME
 from nucypher.config.storages import ForgetfulNodeStorage, TemporaryFileBasedNodeStorage
 from nucypher.policy.payment import SubscriptionManagerPayment
 from nucypher.utilities.networking import LOOPBACK_ADDRESS
-from tests.constants import MOCK_ETH_PROVIDER_URI
+from tests.constants import MOCK_ETH_PROVIDER_URI, TESTERCHAIN_CHAIN_ID
```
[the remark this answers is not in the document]
KPrasch: ✔️

tests/conftest.py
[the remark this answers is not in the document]
manumonti: > good eye -- this is intentional design to support running a subset of tests while still loading specific fixtures in different execution environments.

Can you elaborate on this? I didn't catch the part of loading specific fixtures in different execution environments.

PR #3351, opened 2023-11-24
Refactor EvalError as an exception class

KPrasch:
**Type of PR:**
- Bugfix

**Required reviews:** 
2


jMyles:
Wait... where does this even get raised?  I see the handling in server.py.

KPrasch:
> Wait... where does this even get raised? I see the handling in server.py.

it's not obvious at all. https://github.com/nucypher/nucypher/blob/v7.0.0/nucypher/blockchain/eth/actors.py#L659

PR #3368, opened 2023-12-15
Don't use gas strategies

cygnusv:
Gas strategies are meant only for legacy transactions (pre-EIP1559). See https://web3py.readthedocs.io/en/stable/gas_price.html

Just by using a gas strategy, web3py understands we're switching to legacy mode. See how TX defaults change after we configure the gas strategy: 
```
(Pdb) fill_transaction_defaults(self.w3, {})
{'value': 0, 'data': b'', 'gas': 53000, 'maxFeePerGas': 1694505782, 'maxPriorityFeePerGas': 1694505750, 'chainId': 80001}
(Pdb) n
--Return--
> /root/gh/nucypher/nucypher/blockchain/eth/interfaces.py(264)attach_middleware()->None
-> self.configure_gas_strategy()
(Pdb) fill_transaction_defaults(self.w3, {})
{'value': 0, 'data': b'', 'gas': 53000, 'gasPrice': 1694505766, 'chainId': 80001}
```

**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/newsfragments/3368.bugfix.rst
+++ b/newsfragments/3368.bugfix.rst
@@ -1,0 +1,1 @@
+Don't use web3.py gas strategies, since that switches TX mode to legacy.
--- a/nucypher/blockchain/eth/interfaces.py
+++ b/nucypher/blockchain/eth/interfaces.py
@@ -259,7 +259,8 @@ class BlockchainInterface:
             self.log.debug('Injecting POA middleware at layer 0')
             self.client.inject_middleware(geth_poa_middleware, layer=0)
 
-        self.configure_gas_strategy()
+        # TODO:  See #2770
+        # self.configure_gas_strategy()
 
     def configure_gas_strategy(self, gas_strategy: Optional[Callable] = None) -> None:
 


cygnusv:
From web3py (https://github.com/ethereum/web3.py/blob/ecc45a4b9aaecff926bcd4c707bdaa8c05ce5d3f/web3/_utils/transactions.py#L69C1-L80C2): 
```py
TRANSACTION_DEFAULTS = {
    ...
    "gasPrice": lambda w3, tx: w3.eth.generate_gas_price(tx),
    "maxFeePerGas": (
        lambda w3, tx: w3.eth.max_priority_fee
        + (2 * w3.eth.get_block("latest")["baseFeePerGas"])
    ),
    "maxPriorityFeePerGas": lambda w3, tx: w3.eth.max_priority_fee,
    ...
}
```
By default, `generate_gas_price` will call `eth_gasPrice`. I've been doing some tests with Infura on Polygon mainnet and this is what I got. For example, at a given time, I got these outputs:

* `eth_gasPrice` = 148.63 gwei
* `baseFeePerGas` = 133.6 gwei
* `max_priority_fee` = 30 gwei

This means that for legacy TXs, gas price was 148.63 gwei at a time where baseFee was around 133, which in practice means that we're capping the gas price of the TX to 148 gwei. I'm not sure if such TX would actually go through until baseFee gets lower than 148 - 30 = 118 gwei, and in any case, we'd be spending 148 gwei regardless if the base fee goes significantly lower.

For normal TXs (i.e., post-EIP-1559), maxPriorityFee would be 30 and max fee would be 30 + 2*133.6 = 297.2 gwei, which means that, again assuming a baseFee around 133, we'd be spending 133+30=163 gwei per gas, but TX would go through if base fee spikes up to 297 gwei.

With this in mind, I think we should transition away from legacy TXs, considering perhaps a mechanism to make sure that priority fees are generous enough, possibly through some gas price oracle (e.g, https://docs.polygon.technology/tools/gas/polygon-gas-station/)

KPrasch (approved):

PR #3396, opened 2024-01-18
set MAX_UPLOAD_CONTENT_LENGTH to 250kb

piotr-roslaniec:
**Type of PR:**
- Bugfix

**Required reviews:** 
- 1

**What this does:**
- Increases `MAX_UPLOAD_CONTENT_LENGTH` to 250kb

**Why it's needed:**
- In TACo, as the threshold increases, so does the number of decryption requests. Those requests are sent in bulk to Porter, who's currently rejecting any request larger than 50kb. So for any threshold larger than ~12, this will result in HTTP `413 Request Entity Too Large`.
- Setting `MAX_UPLOAD_CONTENT_LENGTH` to 250kb should accommodate thresholds up to about 50 
(I eyeballed this one)

**Notes for reviewers:**
- Is 250kb too big/small? Could we increase it to, for example, 500kb? I don't think we're going to run rituals larger than 50-of-M anytime soon?

### DIFF
--- a/newsfragments/3396.bugfix.rst
+++ b/newsfragments/3396.bugfix.rst
@@ -1,0 +1,1 @@
+Fix `MAX_UPLOAD_CONTENT_LENGTH` too small for mainnet TACo rituals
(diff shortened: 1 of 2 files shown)


cygnusv (approved):
> Is 250kb too big/small? Could we increase it to, for example, 500kb? I don't think we're going to run rituals larger than 50-of-M anytime soon?

250kb seems about right to me for now, it can potentially accommodate up to 100 requests. In any case, I don't think we're going to run cohorts bigger than 30 in the short term (particularly since Coordinator has a cap of 32 nodes for DKGs).

KPrasch (approved):
Thanks!

PR #3398, opened 2024-01-19
Handle incoming [proxied] request IPv4 resolution

### LINKED ISSUE #3378: Endpoint /ping reveals sensitive information
**Describe the Bug**
The `/ping` endpoint exposed by each node returns the IP address of the requester. It's not working as intended when a node is behind a proxy. Then it will return the IP address of the proxy. I think it is not intended and not secure as it reveals the internal address range of the infrastructure where the node is running.

**To Reproduce**
Open https://<PUBLIC_IP_OF_THE_NODE:<PORT>/ping

**Problematic Code**

https://github.com/nucypher/nucypher/blob/v7.0.4/nucypher/network/server.py#L276

KPrasch:
**Type of PR:**
Bugfix

**Required reviews:** 
2

**What this does:**
- Handles the possibility of incoming request headers that have been mutated by intermediate network proxies, like nginx.
- Attempts to resolve IPv6 to IPv4 when possible

**Issues fixed/closed:**
- Fixes #3378 

**Why it's needed:**
- Incoming request headers were being improperly parsed and handled when a request is forwarded by a proxy, leading to a remote node's internal/private IP address being returned to the caller.
- IPv6 addresses are unsupported by the protocol.  Handles unexpected IPv6 values in /ping.

**Notes for reviewers:**
- Given that `/ping` is used both for sampling and configuration assist - what is the desired behavior when the address is invalid, private, or missing?   Currently responding with HTTP 503.
- XFF header reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#selecting_an_ip_address

### DIFF
--- a/nucypher/utilities/networking.py
+++ b/nucypher/utilities/networking.py
@@ -1,7 +1,7 @@
 import random
 from http import HTTPStatus
 from ipaddress import AddressValueError, IPv4Address, IPv6Address, ip_address
-from typing import List, Optional, Union
+from typing import Optional, Union
 
 import requests
 from flask import Request
@@ -209,6 +209,7 @@ def determine_external_ip_address(
 
 
 def _is_global_ipv4(ip: str) -> bool:
+    """Check if an IP address is a global IPv4 address according to RFC 1918."""
     try:
         ip = ip_address(ip.strip())
         return isinstance(ip, IPv4Address) and ip.is_global
@@ -217,6 +218,7 @@ def _is_global_ipv4(ip: str) -> bool:
 
 
 def _resolve_ipv4(ip: str) -> Optional[str]:
+    """Resolve an IP address to IPv4 if required and possible."""
     try:
         ip = ip_address(ip.strip())
     except AddressValueError:
@@ -227,21 +229,21 @@ def _resolve_ipv4(ip: str) -> Optional[str]:
         return str(ip)
 
 
-def _ip_sources(request: Request, trusted_proxies: Optional[List[str]] = None) -> str:
-    if not trusted_proxies:
-        trusted_proxies = []
+def _extract_ip(header: str, request: Request) -> Optional[str]:
+    """Extract the first IP address from a request header."""
+    if header in request.headers:
+        for ip in request.headers[header].split(","):
+            yield ip
+
+
+def _ip_sources(request: Request) -> str:
+    """Iterate over all possible sources of IP addresses in a request's headers."""
     for header in ["X-Forwarded-For", "X-Real-IP"]:
-        if header in request.headers:
-            if trusted_proxies and (request.remote_addr not in trusted_proxies):
-                yield None  # Do not trust this request
-            for ip in request.headers[header].split(","):
-                yield ip
+        yield from _extract_ip(header, request)
     yield request.remote_addr
 
 
-def get_request_global_ipv4(
-    request: Request, trusted_proxies: Optional[List[str]] = None
-) -> Optional[str]:
+def get_request_global_ipv4(request: Request) -> Optional[str]:
     """
     If the request is forwarded from a proxy, the first global IP address in the chain is returned.
     'X-Forwarded-For' (XFF) https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For#selecting_an_ip_address
@@ -256,9 +258,8 @@ def get_request_global_ipv4(
     address ranges must not be returned.
     https://www.ietf.org/rfc/rfc1918.txt
 
-    Optionally, a list of trusted proxies can be provided to help mitigate spoofing attacks.
     """
-    for ip_str in _ip_sources(request=request, trusted_proxies=trusted_proxies):
+    for ip_str in _ip_sources(request=request):
         ipv4_address = _resolve_ipv4(ip_str)
         if ipv4_address and _is_global_ipv4(ipv4_address):
             return ipv4_address
--- a/nucypher/utilities/networking.py
+++ b/nucypher/utilities/networking.py
@@ -179,6 +179,9 @@ def determine_external_ip_address(
     3. A centralized IP address service
 
     If the IP address cannot be determined for any reason UnknownIPAddress is raised.
+
+    This function is intended to be used by nodes operators running `nucypher ursula init`
+    to assist determine their global IPv4 address for configuration purposes only.
     """
     rest_host = None
 
@@ -205,7 +208,7 @@ def determine_external_ip_address(
     return rest_host
 
 
-def _is_valid_ipv4(ip: str) -> bool:
+def _is_global_ipv4(ip: str) -> bool:
     try:
         ip = ip_address(ip.strip())
         return isinstance(ip, IPv4Address) and ip.is_global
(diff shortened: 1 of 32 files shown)


nucypher/utilities/networking.py
```diff
@@ -200,4 +201,60 @@ def determine_external_ip_address(
...
+
+def _is_valid_ipv4(ip: str) -> bool:
+    try:
+        ip = ip_address(ip.strip())
+        return isinstance(ip, IPv4Address) and ip.is_global
+    except AddressValueError:
+        return False
+
+
+def _ipv4_to_ipv6(ip: str) -> Optional[str]:
```
cygnusv, line 216: shouldn't this be called `_ipv4_from_ipv6` or  `_ipv6_to_ipv4`?

KPrasch: Yes! fixed.

nucypher/utilities/networking.py
```diff
@@ -200,4 +204,62 @@ def determine_external_ip_address(
...
+def _extract_ip(header: str, request: Request) -> Optional[str]:
+    """Extract the first IP address from a request header."""
+    if header in request.headers:
+        for ip in request.headers[header].split(","):
+            yield ip
+
+
+def _ip_sources(request: Request) -> str:
+    """Iterate over all possible sources of IP addresses in a request's headers."""
+    for header in ["X-Forwarded-For", "X-Real-IP"]:
```
piotr-roslaniec, line 241: How did we select those headers? Are these industry standards for proxies/forwarding?

nucypher/utilities/networking.py
```diff
@@ -200,4 +204,62 @@ def determine_external_ip_address(
...
+def _resolve_ipv4(ip: str) -> Optional[str]:
+    """Resolve an IP address to IPv4 if required and possible."""
+    try:
+        ip = ip_address(ip.strip())
+    except AddressValueError:
+        return None
+    if isinstance(ip, IPv6Address) and ip.ipv4_mapped:
+        return str(ip.ipv4_mapped)
+    elif isinstance(ip, IPv4Address):
+        return str(ip)
```
derekpierre, line 229: Seems like `ip_address(...)` returns either IPv6 or IPv4...can this `elif` go away or be changed to an `else`?

nucypher/utilities/networking.py
```diff
@@ -200,4 +204,62 @@ def determine_external_ip_address(
...
+
+    In all cases, tests that the IPv4 range is global. RFC 1918 privately designated
+    address ranges must not be returned.
+    https://www.ietf.org/rfc/rfc1918.txt
+
+    """
+    for ip_str in _ip_sources(request=request):
+        ipv4_address = _resolve_ipv4(ip_str)
+        if ipv4_address and _is_global_ipv4(ipv4_address):
+            return ipv4_address
```
derekpierre, line 265: Does it make sense for `_resolve_ipv4` to instead return an `IPv4Address` object instead of a string? Then you don't really need to do the extra call to `ip_address(...)` in `_is_global_ipv4` which is duplicated and `_is_global_ipv4` is no longer needed. For example, you could end up just doing this:

```suggestion
    for ip_str in _ip_sources(request=request):
        ipv4_address = _resolve_ipv4(ip_str)
        if ipv4_address and ipv4_address.is_global:
            return str(ipv4_address)
```

wdyt?

tests/integration/cli/actions/test_select_config_file.py
```diff
@@ -62,6 +57,7 @@ def test_auto_select_config_file(
                                               config_file=str(config_path)) in captured.out
 
 
+@pytest.mark.skip("planned for removal in PR #3382")
```
derekpierre, line 60: Are the changes for this test needed for v7.1.x? 3382 is set for 7.2.x

tests/unit/test_ping_utils.py
```diff
@@ -0,0 +1,104 @@
...
+        ValueError, match="'not_an_ip' does not appear to be an IPv4 or IPv6 address"
+    ):
+        _is_global_ipv4("not_an_ip")
+
+
+def test_resolve_ipv4_with_non_mapped_ipv6():
+    assert _resolve_ipv4("2001:0db8::") is None
+
+
+def test_ip_sources_with_both_headers(mock_request_factory):
```
derekpierre, line 62: Nice test!

derekpierre:
Did you run `darker` separately? Just curious since there seem to be unrelated updates to imports etc.

derekpierre:
> > Did you run darker separately? Just curious since there seems to be unrelated updates to imports etc.
> 
> I did run `ruff --fix` separately perhaps that's the cause? Trying to pass pre-commit checks.

👍 - probably. I guess `ruff --fix` ran on all files as opposed to files specific to your changes; usually I run `ruff --fix <filepath>`.

tests/integration/cli/actions/test_select_config_file.py
KPrasch: There appeared to be an unrelated test failure in this test.  I can investigate the issue, but this test is in support of a very specific configuration behavior when the host system has multiple configuration files. Suggestions?

nucypher/utilities/networking.py
KPrasch: ✔️ nice idea - I also think it's more pythonic to use higher level abstractions here.

nucypher/utilities/networking.py
KPrasch: ✔️ (sort of!)

KPrasch: Actually that will have a slightly different result.  This function is designed to return `None` if `not isinstance(ip, IPv6Address) and ip.ipv4_mapped`

derekpierre: True., I guess return` None` if you get passed the elif then...

tests/integration/cli/actions/test_select_config_file.py
derekpierre: Weird that it fails now. In any case, it makes sense not to spend time on it given the specific area of the code, related to multiple configurations which doesn't really apply to ursula anyway, and if the fix isn't obvious.

KPrasch: > Weird that it fails now. 

💯 I suspect that it might be a dependency change?  I agree that this test failure is highly contained and unlikely to indicate any real issues running taco nodes.

nucypher/utilities/networking.py
```diff
@@ -200,4 +204,53 @@ def determine_external_ip_address(
...
+def _resolve_ipv4(ip: str) -> Optional[IPv4Address]:
+    """Resolve an IP address to IPv4 if required and possible."""
+    try:
+        ip = ip_address(ip.strip())
+    except AddressValueError:
+        return None
+    if isinstance(ip, IPv6Address) and ip.ipv4_mapped:
+        return ip.ipv4_mapped
+    elif isinstance(ip, IPv4Address):
+        return ip
```
derekpierre, line 220: Looks like `ipv4_mapped` can return `None` so should we just do this (I should have noticed this earlier), then all our cases are handled.

```suggestion
    if isinstance(ip, IPv6Address):
        return ip.ipv4_mapped
    # if not ipv6 then it is ipv4 already
    return ip
```

KPrasch: ✔️

derekpierre (approved):
🎸

PR #3408, opened 2024-01-23
Include latest event scanner block in `/status` endpoint

KPrasch:
##### Requested Reviews
1

### DIFF
--- a/nucypher/characters/lawful.py
+++ b/nucypher/characters/lawful.py
@@ -1343,6 +1343,7 @@ class Ursula(Teacher, Character, Operator):
             previous_fleet_states=previous_fleet_states,
             known_nodes=known_nodes_info,
             balance_eth=balance_eth,
+            block_height=self.ritual_tracker.scanner.get_last_scanned_block(),
         )
 
     def as_external_validator(self) -> Validator:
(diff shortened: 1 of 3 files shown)


nucypher/characters/lawful.py
```diff
@@ -1343,6 +1343,7 @@ def status_info(self, omit_known_nodes: bool = False) -> "LocalUrsulaStatus":
             previous_fleet_states=previous_fleet_states,
             known_nodes=known_nodes_info,
             balance_eth=balance_eth,
+            block_height=self.ritual_tracker.scanner.get_last_scanned_block(),
```
cygnusv, line 1346: 👨‍🍳 💋

derekpierre:
> Great! My only comment: can we make the status variable name for this called `last_scanned_block` or something like that?

+1 on the renaming.

nucypher/network/templates/basic_status.mako
```diff
@@ -183,6 +183,10 @@ def character_span(character):
                 %endfor
             </td>
         </tr>
+        <tr>
+            <td><i>Latest Scanned Block:</i></td>
+            <td>${ status_info.block_height }</td>
+        </tr>
```
derekpierre, line 189: Is it necessary to include this information in the actual html status page?

PR #3409, opened 2024-01-23
removes the `/revoke` endpoint

KPrasch:


derekpierre:
What do we do about the following?
- https://github.com/nucypher/nucypher/blob/v7.3.x/nucypher/network/middleware.py#L226 (middleware call)
- https://github.com/nucypher/nucypher/blob/v7.3.x/nucypher/characters/lawful.py#L411 (Alice.revoke)

KPrasch:
Thanks for noticing those areas I overlooked.  I took another pas using find and replace and removed more usage. ✔️

cygnusv (approved):
Loving the red lines 🪓

PR #3419, opened 2024-01-26
Optimize participant and transcript reads

### LINKED ISSUE #3394: Node 0x3070f20f86fDa706Ac380F5060D256028a46eC29 raised {'code': -32000, 'message': 'call returned result of length 122944 exceeding limit 100000'}
**Describe the Bug**
```
Node 0x3070f20f86fDa706Ac380F5060D256028a46eC29 raised {'code': -32000, 'message': 'call returned result of length 122944 exceeding limit 100000'}
```

**To Reproduce**
During TDec decryption, on mainnet ritual 1

**Traceback or Screenshots (Optional)**
If applicable, add the full error message / Traceback or upload screenshots to help explain your problem.

 - OS Version:
 - Nucypher Version:
 - Ethereum Node Version:

### LINKED ISSUE #3417: Don't get all participants in Ritual.Participant[] since it can be large due to transcripts bytes being in each entry
Consider pagination.

Concrete example: instead of pulling 30 Pariticipant structs in one call (`getParticipants` contract call), get them in chunks.

Pariticipant struct is data heavy because of the bytes for the transcript.

KPrasch:
###### RPC Payload Volume Optimization
- Optimizes `Coordinator` reads for transcript data
- Implements updated `Coordinator` participant `views`:
  - participant data with optional transcripts attached
  - paginated query results for participant data
- Implements automated page size and integration with optimized coordinator functions
- Updates `CoordinatorAgent` to sync with `Coordinator` abi changes 

###### Code Quality
- Isolates ritual data fetch and validation into discrete methods
- Improves code reuse of ritual I/O between `Bob` and `Ursula`
- extracts `Coordinator` struct models to a dedicated module `nucypher/blockchain/eth/models.py`

###### Bugfixes
- Clear pending txhashes when they are included in a block
- Properly detect pending ritual transactions

###### More
- Updates lynx contract registry to sync with nucypher/nucypher-contracts `v0.19.0` (upgrade)
- Prepares a site for integration with #3424 

##### Issues
- Fixes #3394 
- Fixes #3417 

##### Notes
- depends on https://github.com/nucypher/nucypher-contracts/pull/231
- Requires an upgrade of `Coordinator`.
- 184 line registry diff makes the true green line count +810 (-538) = +272 additions


tests/acceptance/ape-config.yaml
```diff
@@ -5,7 +5,7 @@ plugins:
 
 dependencies:
   - name: nucypher-contracts
-    github: nucypher/nucypher-contracts
+    github: kprasch/nucypher-contracts
```
derekpierre, line 8: Just putting a note here to change this before we merge.

nucypher/blockchain/eth/actors.py
```diff
@@ -358,10 +347,78 @@ def publish_aggregated_transcript(
...
+        except TransactionNotFound:
+            return txhash, None
+        status = receipt.get("status")
+        if status == 1:
+            # If status in response equals 1 the transaction was successful.
+            # If it is equals 0 the transaction was reverted by EVM.
+            # https://web3py.readthedocs.io/en/stable/web3.eth.html#web3.eth.Eth.get_transaction_receipt
+            # TODO: What follow-up actions can be taken if the transaction was reverted?
+            return txhash, receipt
+        raise self.ActorError(
```
derekpierre, line 376: @KPrasch given that we are now going to separate retry/replacement from the act of `perform_round_1`/`perform_round_2`, I don't think (?) we should raise here. 

In the Actor's case, it only cares about whether there is a pending tx or not. In this case, if the original tx was reverted, the tx is not pending (and the  **eventual** dkg tracker (#3424 ) presumably can't do anything about a reverted tx) so should this return `None, None` so that the Actor tries another post transcript/post aggregation tx?

Basically, would raising here potentially cause the node to never get out of a reverted tx situation?

derekpierre: We may need to separate `get_phase_receipt` logic to check only for a tx hash (used by Actor), and txhash & receipt (used by DKG Tracker).

derekpierre: One potential musing to "fix" this - https://github.com/derekpierre/nucypher/commit/5881403d0a0fbebdf31444be4755b1d8b4e17901. Let me know what you think.

KPrasch: lgtm 👍🏻 even with the little amount of direction finding we've done in #3424 it's clear we're not going to fire retry phases in the same async task as `perform_round1|2`.

derekpierre: 👍 - I'll just add some tests for it.

derekpierre: ✔️ - done!

tests/acceptance/ape-config.yaml
KPrasch:
```suggestion
    github: nucypher/nucypher-contracts
```

derekpierre (approved):
Implicit approval since I'm a co-author of the PR (🎸 )

nucypher/blockchain/eth/contract_registry/lynx.json
```diff
@@ -662,6 +662,30 @@
                         }
                     ]
                 },
+                {
```
theref, line 665: can we use the published `nucypher-contracts` instead?

nucypher/blockchain/eth/actors.py
```diff
@@ -358,10 +347,80 @@ def publish_aggregated_transcript(
             public_key=public_key,
             participant_public_key=participant_public_key,
             transacting_power=self.transacting_power,
-            fire_and_forget=True,
         )
         return tx_hash
 
+    def get_phase_receipt(
```
theref, line 353: What are the phases?

KPrasch: It's another name for the dkg "rounds" while trying to avoid name shadowing `round` in several places.

nucypher/blockchain/eth/agents.py
```diff
@@ -761,41 +663,70 @@ def get_ritual(self, ritual_id: int, with_participants: bool = True) -> Ritual:
...
-        if with_participants:
-            participants = self.get_participants(ritual_id)
-            ritual.participants = participants
+    def get_ritual(
+        self,
+        ritual_id: int,
+        transcripts: bool = False,
+    ) -> Coordinator.Ritual:
+        """
+        Exposes three views of Coordinator.Rituals:
```
piotr-roslaniec, line 677:
```suggestion
        Exposes two views of Coordinator.Rituals:
```

piotr-roslaniec (approved):

nucypher/blockchain/eth/actors.py
```diff
@@ -375,175 +434,152 @@ def perform_round_1(
...
             # TODO: Handle this better #3096
-            self.log.debug(f"Failed to generate a transcript for ritual #{ritual_id}: {str(e)}")
+            self.log.critical(
+                f"Failed to generate a transcript for ritual #{ritual.id}: {str(e)}"
+            )
             raise e
 
         # store the transcript in the local cache
-        self.dkg_storage.store_transcript(ritual_id=ritual_id, transcript=transcript)
+        self.dkg_storage.store_transcript(ritual_id=ritual.id, transcript=transcript)
```
derekpierre, line 474: @KPrasch I just realized that we don't use the stored transcript at all (Note that this is different from the hash for the transaction, it is the actual transcript). I'd like to remove this code (and analogous code for the aggregated transcript in round 2) - however, I'd prefer to do it in a follow-up PR.

derekpierre: Filed https://github.com/nucypher/nucypher/issues/3426

KPrasch: followed up: #3428

Issue #1942, opened 2020-04-30
CLI Flag to use local registry without specifying filepath


KPrasch:
This is useful for backward compatibility scenarios (especially testnets) as a convenience to use the locally embedded contract registry.

derekpierre:
Nice to have - at the moment you can specify the filepath.

Marking as won't fix.

PR #3420, opened 2024-01-26
Updates remote registry source url

### LINKED ISSUE #3336: Github registry source and the use of `development` branch
The GithubRegistry source for the contract registry expects the registry to be located in the `development` branch of the repos, which currently does not exist.

Either we reinstate the `development` branch or come up with an alternative.

KPrasch:
**Type of PR:**
Bugfix 

**Required reviews:** 
2

**Issues fixed/closed:**
- Fixes #3336 

**Why it's needed:**
- Only embedded registry sources are being used by v7.0.x (via fallback from remote sources) which reduces update agility.


derekpierre:
Just thinking this out loud here:

Now that we are doing deployments from `nucypher-contracts` instead of `nucypher/nucypher` does it make sense to point the URL there instead?

One slight concern I have with using `nucpyher/nucypher` is that when we update contracts there will be some delay between when contracts are deployed via `nucypher-contracts` scripts (PR, review, merge), and when the updated registry gets copied over the `nucypher/nucypher` and into `main`. 

Basically pointing to `nucypher-contracts` requires no coordination, but pointing to `nucypher/nucypher` requires coordination - which I like as a pro.

The flip-side of using `nucypher-contracts` is that if we do an incompatible ABI change to a contract on `mainnet`, then nodes will be affected immediately after updating the registry in `nucypher-contracts`, **before** a `nucypher/nucypher` release is available (changes merged into `nucypher/nucypher:main` as part of a release). That being said, the contract has changed on the blockchain so nodes are affected anyway ... 🤔

KPrasch:
Perhaps we can include a fix for https://github.com/nucypher/nucypher/issues/1942 in this PR to alleviate testnet issues?

derekpierre:
> Perhaps we can include a fix for #1942 in this PR to alleviate testnet issues?

So the idea here is to have a CLI parameter to force the use of the embedded registry over the github one? 

I guess the node would still have issues if the change is incompatible since the most important contracts for the protocol are all proxied (`TACoApplication`, `TACoChildApplication`, `Coordinator`), so calling them based on an older incompatible ABI definition may not help...?

That being said, I don't know if I have a grasp of the full scope; perhaps this is still a good backup in some situations? 

(cc @vzotova , @cygnusv )

KPrasch:
One issue I see right now is that testnet nodes need to reference a different branch (not `main`), or just always use an embedded registry.  Normally the mainnet registry in the main branch will be up to date with the latest version and we'll handle development registries separately.

 This means that for a launching node's client code to be incompatible with the official mainnet regsitry, it must be out of date anyways, right?

derekpierre:
> One issue I see right now is that testnet nodes need to reference a different branch (not `main`), or just always use an embedded registry. Normally the mainnet registry in the main branch will be up to date with the latest version and we'll handle development registries separately.

I think using `nucypher-contracts` as the source solves the testnet issue because then they always get the latest. Our deployment scripts always ensure that we deploy (testnet) contracts from `nucypher-contracts` and we subsequently always create a PR to merge the updated registry into `nucypher-contracts:main`.

(As a side note, our `nucypher/nucypher` code runs acceptance tests using `nucypher-contracts:main`)

cygnusv:
I totally support the idea of retrieving from `nucypher-contracts`, ideally to a stable branch or release tag. wdyt?

derekpierre:
> I totally support the idea of retrieving from `nucypher-contracts`, ideally to a stable branch or release tag. wdyt?

Interesting. I was wondering if `nucypher-contracts:main` was sufficient but I guess there is a delay between deployment of contracts and an actual upgrade by the Council...is that a concern? A fixed name instead of a release number would be ideal, because that's the goal of the github registry source - that the registry nodes use can be updated by simply restarting without needing a code update. Maybe `stable` as you suggested, if `main` indeed should not be used.

vzotova:
I definitely like idea with tags and `nucypher-contracts:main` as one source because right after upgrade anyway old ABIs are stale

piotr-roslaniec (approved):

PR #3424, opened 2024-01-30
Fault Tolerance: Async Transactions and Retries

### LINKED ISSUE #1433: Race conditions should be explored when broadcasting multiple transactions
In ` nucypher/blockchain/eth/interfaces.py`, `send_transaction` we get nonce from geth transaction count and use it later in the method. If the same method is called in a different thread, it may happen that both methods get the same nonce (and one of those transactions will fail).
We need to either create a lock until we broadcast the transaction, or make a geth-independent counter for nonces (or both).

### LINKED ISSUE #2370: Speedups and cancellations for nucypher contract transactions

### LINKED ISSUE #3101: Retry slow or failed transcript posting transactions
consider repurposing `WorkTracker`.

### LINKED ISSUE #3391: DKG finalization runs out of gas sometimes
Occasionally, the final node in the cohort to post an aggregation runs out of gas causing the transaction to fail, and thus the entire DKG to fail.    This is most likely caused by a race condition:  Since the last node to post an aggregation must also execute the DKG finalization block on-chain a bit more gas expenditure is needed compared to the other nodes in the cohort.  If all nodes generate and broadcast transaction posting transactions at the same time, none of them will generate an accurate gas estimate since none of them "believes" they will be the last one to post.

https://mumbai.polygonscan.com/tx/0xc4667d599e07678ed710e11bd43cc5202d6b6a191c5ad7dd689fc85de6744ba4

KPrasch:
##### What this does
- automatically retry failed transactions

##### Issues
- Fixes #3101 
- Closes #2370 
- Closes #1433 

##### Notes
- long history but small turnover
- depends on https://github.com/nucypher/ATxM/pull/1
(pinned to `atxm@ git+https://github.com/KPrasch/atxm.git@7c3d210c13042b259ae76866be619b8f659bb661`)
- requires a full dependency update and relock post-merge
- co-authored by @derekpierre 
- related (partially fixes #3391)


KPrasch:
subsumed by #3437

PR #3432, opened 2024-02-05
Auto migration of outdated configuration files

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

After an upgrade that included changing the configuration file, lynx nodes were failing:
```
No Ursula configurations found. Run 'nucypher ursula init' then try again.

Ignoring configuration file '/root/.local/share/nucypher/ursula.json' - version is too old,
Run `nucypher ursula config migrate --config-file '/root/.local/share/nucypher/ursula.json'` to update it.

No Ursula configurations found. Run 'nucypher ursula init' then try again.
Aborted!
```

The original configuration file for lynx nodes used an old version, and the node would not startup until the configuration file was manually migrated using `nucypher ursula config migrate...`. Instead the node should detect that it has an outdated config file, and automatically migrate it so that it can start up.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/cli/actions/migrate.py
+++ b/nucypher/cli/actions/migrate.py
@@ -1,0 +1,44 @@
+from pathlib import Path
+
+import click
+
+from nucypher.config.migrations import MIGRATIONS
+from nucypher.config.migrations.common import (
+    InvalidMigration,
+    WrongConfigurationVersion,
+)
+from nucypher.utilities.emitters import StdoutEmitter
+
+
+def migrate(emitter: StdoutEmitter, config_file: Path):
+    for jump, migration in MIGRATIONS.items():
+        old, new = jump
+        emitter.message(f"Checking migration {old} -> {new}")
+        if not migration:
+            emitter.echo(
+                f"Migration {old} -> {new} not found.",
+                color="yellow",
+                verbosity=1,
+            )
+            continue  # no migration script
+        try:
+            migration(config_file)
+            emitter.echo(
+                f"Successfully ran migration {old} -> {new}",
+                color="green",
+                verbosity=1,
+            )
+
+        except WrongConfigurationVersion:
+            emitter.echo(
+                f"Migration {old} -> {new} not required.",
+                color="yellow",
+                verbosity=1,
+            )
+            continue  # already migrated
+
+        except InvalidMigration as e:
+            emitter.error(f"Migration {old} -> {new} failed: {str(e)}")
+            raise click.Abort()
+
+    emitter.echo("Done! ✨", color="green", verbosity=1)
--- a/nucypher/cli/actions/select.py
+++ b/nucypher/cli/actions/select.py
@@ -14,9 +14,11 @@ from nucypher.cli.actions.migrate import migrate
 from nucypher.cli.literature import (
     DEFAULT_TO_LONE_CONFIG_FILE,
     GENERIC_SELECT_ACCOUNT,
+    IGNORE_OLD_CONFIGURATION,
     MIGRATE_OLD_CONFIGURATION,
     NO_ACCOUNTS,
     NO_CONFIGURATIONS_ON_DISK,
+    PROMPT_TO_MIGRATE,
     SELECT_DOMAIN,
     SELECTED_ACCOUNT,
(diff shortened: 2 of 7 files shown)


nucypher/cli/actions/select.py
```diff
@@ -160,10 +164,17 @@ def select_config_file(emitter: StdoutEmitter,
 
             parsed_config_files.append(fp)
             parsed_addresses_and_filenames.append([config_checksum_address, Path(fp).name])  # store checksum & filename
-        except config_class.OldVersion:
-            # no use causing entire usage to crash if file can't be used anyway - inform the user; they can
-            # decide for themself
-            emitter.echo(IGNORE_OLD_CONFIGURATION.format(config_file=fp), color='yellow')
+        except config_class.OldVersion as e:
+            # attempt migration
```
derekpierre, line 168: Note: this implies that any config file encountered with an older version is migrated - even if we are only reading the file eg. `nucypher ursula config` to show the contents of the file.  Instead of failing and telling the user to migrate the configuration using the `migrate` function, we automatically migrate the file, and then can continue to show the contents of the file.

Just double-checking if we are comfortable with this.

nucypher/cli/actions/migrate.py
```diff
@@ -0,0 +1,44 @@
...
+    for jump, migration in MIGRATIONS.items():
+        old, new = jump
+        emitter.message(f"Checking migration {old} -> {new}")
+        if not migration:
+            emitter.echo(
+                f"Migration {old} -> {new} not found.",
+                color="yellow",
+                verbosity=1,
+            )
+            continue  # no migration script
```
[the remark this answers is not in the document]
derekpierre: > Do we assume that not having a migration is ok (because it may not be needed) or are we going to throw an exception somewhere else?

Fair question - essentially both. `no-ops` shouldn't really happen because then what was the point of updating the configuration version. There is no migration for v2->v3; I think (?) it was because we didn't do the migration in conjunction with it at the time, and we may have lost track of what we changed 😞 , but that was a long time ago. 

The migration is done in substeps - https://github.com/nucypher/nucypher/blob/v7.2.x/nucypher/config/migrations/__init__.py. If the migration fails at any point in that chain, then an exception is raised. In any case, if any single step of migration fails, the entire thing fails anyway. So the fact that v2->v3 is a no-op means that **if** we started with a v1 or v2 file and the resulting v3 file can't be migrated to v4 then the whole migration fails. If it can be migrated from v3 -> v4 then no problem and we continue with the remaining migrations to latest.

That being said I don't believe there are any actual config versions lower than v4 (`nucypher v6.0.0`) out there - we can handle migrations from v4 all the way to latest - there's a test for this, https://github.com/nucypher/nucypher/blob/v7.2.x/tests/integration/config/test_configuration_migrations.py#L42.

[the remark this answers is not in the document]
piotr-roslaniec: What do you think about the alternative, i.e. filling the missing migrations with `no-ops` and throwing when no migration is found?

nucypher/cli/actions/select.py
KPrasch: If someone is running the `config` CLI commands it's safe to assume it's interactive-safe.  It's totally fine to prompt here if you choose to.

derekpierre: Yeah this is the tricky part indeed. We can do this. We would need an extra parameter like `do_auto_migrate` for `select_config_file` which defaults to `False`.

Then, any reading calls don't do the migration, and then only set `do_auto_migrate` to True, when Ursula run is making the call. We don't want to prompt during `nucypher ursula run`.

derekpierre: ✔️ - resolved. User is prompted by default unless `do_auto_migrate` is set to True. It is currently only set to try for `nucypher ursula run`.

KPrasch (approved):

PR #3433, opened 2024-02-06
Build and push package to pypi whenever a release is published

theref:
**Type of PR:**
- [ ] Feature

**Required reviews:** 
- [x] 1

When a release is published, this workflow will be triggered to build `nucypher` and push to https://pypi.org/project/nucypher/

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -1,0 +1,51 @@
+name: Publish Python 🐍 distribution 📦 to PyPI
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  build:
+    name: Build distribution 📦
+    runs-on: ubuntu-latest
+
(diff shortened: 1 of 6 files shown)


KPrasch (commented):
Thanks for picking up this issue :-) 

I ran the build command in the action and the output looks mostly good, but I think we may be able to reduce the amount of source code included in the sdist/bdist (for example, it looks like we're shipping tests since it's not properly handled in `MANIFEST`).  

Also there are some build warnings that will be good to resolve: https://gist.github.com/KPrasch/71feefcc53032a372ee2a2e04715ec11

KPrasch (commented):
Another note - If we're going to use `python -m build` instead of calling `setup.py` it will be good to cleanup `Makefile` to disambiguate between the different build strategies.

PR #3437, opened 2024-02-08
Fault Tolerance: Async Transactions

### LINKED ISSUE #1433: Race conditions should be explored when broadcasting multiple transactions

### LINKED ISSUE #2370: Speedups and cancellations for nucypher contract transactions

### LINKED ISSUE #3101: Retry slow or failed transcript posting transactions

### LINKED ISSUE #3391: DKG finalization runs out of gas sometimes

KPrasch:
##### What this does
- automatically retry failed transactions

##### User Story 
- https://github.com/nucypher/tdec/issues/151

##### Issues
- Fixes #3101 
- Closes #2370 
- Closes #1433 

##### Notes
- Based over #3438 
- long history but small turnover - original history in https://github.com/nucypher/nucypher/pull/3424
- depends on https://github.com/nucypher/ATxM/pull/1
(pinned to `atxm@ git+https://github.com/KPrasch/atxm.git@7c3d210c13042b259ae76866be619b8f659bb661`)
- requires a full dependency update and relock post-merge
- co-authored by @derekpierre 
- related (partially fixes #3391)

Tests are currently blocked by https://github.com/ApeWorX/ape/issues/1933 (tracked in https://github.com/nucypher/nucypher/issues/3455)


manumonti (approved):
LGTM! 👏

piotr-roslaniec (approved):

piotr-roslaniec (approved):

tests/mock/interfaces.py
```diff
@@ -13,6 +14,19 @@ class MockBlockchain(TesterBlockchain):
 
     FAKE_TX_HASH = HexBytes(b"FAKE29890FAKE8349804")
 
+    FAKE_ASYNX_TX = FutureTx(
+        id=1,
+        params={
+            "to": HexBytes(b"FAKEFAKEFAKEFAKEFAKEFAKEFAKEFAKEFAKEFAKE"),
```
derekpierre, line 20: fake

tests/unit/test_operator_bonded_tracker.py
```diff
@@ -20,9 +20,6 @@ def test_operator_never_bonded(mocker, get_random_checksum_address):
 
     tracker = OperatorBondedTracker(ursula=ursula)
     try:
-        d = threads.deferToThread(tracker.start)
```
derekpierre, line 23: How come we no longer start the tracker anymore?

nucypher/blockchain/eth/actors.py
```diff
@@ -440,27 +407,74 @@ def _is_phase_1_action_required(self, ritual_id: int) -> bool:
...
             return
 
+        # check if there is  already pending tx for this ritual + round combination
+        async_tx = self.ritual_tracker.active_rituals.get(
+            self._phase_id(ritual_id, PHASE1)
+        )
+        if async_tx:
+            self.log.info(
+                f"Active ritual in progress: {self.transacting_power.account} has submitted tx"
+                f"for ritual #{ritual_id}, phase #{PHASE1} (final: {async_tx.final})"
```
derekpierre, line 470: what is async_tx.final? I sthat the final tx hash for completion, or something else?

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -164,6 +167,10 @@ def web3(self):
     def contract(self):
         return self.coordinator_agent.contract
 
+    @property
+    def active_rituals(self) -> Dict[Tuple[RitualId, PhaseId], AsyncTx]:
+        return self.__phase_txs
```
derekpierre, line 172: When do stale entries get cleared? Eg. ritual completed?

derekpierre:
Nice work!

nucypher/blockchain/eth/actors.py
```diff
@@ -440,27 +407,74 @@ def _is_phase_1_action_required(self, ritual_id: int) -> bool:
 
         return True
 
+    @staticmethod
+    def _phase_id(ritual_id: int, phase: int) -> Tuple[RitualId, PhaseId]:
+        return RitualId(ritual_id), PhaseId(phase)
```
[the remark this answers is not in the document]
KPrasch: It's strictly for type annotations (there many `int`s flying around and in particular I'm helping devs not accidentally get them swapped in future iterations).  Internally I tried to use a this type of tuple everywhere possible and this is it's "constructor".  I suppose it can be promoted to a `NamedTuple` and it will clean this up even more.

nucypher/blockchain/eth/actors.py
```diff
@@ -440,27 +407,74 @@ def _is_phase_1_action_required(self, ritual_id: int) -> bool:
...
+        Perform phase 1 of the DKG protocol for a given ritual ID on this node.
+
+        This method is idempotent and will not submit a transcript if one has
+        already been submitted. It is dispatched by the EventActuator when it
+        receives a StartRitual event from the blockchain.  Since the EventActuator
+        scans overlapping blocks, it is possible that this method will be called
+        multiple times for the same ritual.  This method will check the state of
+        the ritual and participant on the blockchain before submitting a transcript.
+
+        If a there is a tracked AsyncTx for the given ritual and round
```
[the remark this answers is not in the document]
KPrasch: ✔️

nucypher/blockchain/eth/actors.py
```diff
@@ -440,27 +407,74 @@ def _is_phase_1_action_required(self, ritual_id: int) -> bool:
...
+            #
+            # This is a *nearly* a critical error.  It's possible that the
+            # log it as an error and return None to avoid crashing upstack
+            # async tasks and drawing unnecessary amounts of attention to the issue.
+            message = (
+                f"{self.checksum_address}|{self.wallet_address} "
+                f"is not a member of ritual {ritual_id}"
             )
+            self.log.error(message)
+            return
```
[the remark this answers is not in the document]
KPrasch: While I do agree that this is a true erorr condition - if it theoretically did happen do you want to crash the async task?

nucypher/blockchain/eth/actors.py
```diff
@@ -440,27 +407,74 @@ def _is_phase_1_action_required(self, ritual_id: int) -> bool:
...
+            )
             return
 
+        # check if there is  already pending tx for this ritual + round combination
+        async_tx = self.ritual_tracker.active_rituals.get(
+            self._phase_id(ritual_id, PHASE1)
+        )
+        if async_tx:
+            self.log.info(
+                f"Active ritual in progress: {self.transacting_power.account} has submitted tx"
```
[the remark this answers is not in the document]
KPrasch: ✔️

nucypher/blockchain/eth/actors.py
KPrasch: Excellent question - it's a boolean. This is part of the `atxm` API that allows callers to block until an async transaction is final.

nucypher/blockchain/eth/actors.py
```diff
@@ -535,12 +545,23 @@ def _is_phase_2_action_required(self, ritual_id: int) -> bool:
...
             return
 
+        # check if there is a pending tx for this ritual + round combination
+        async_tx = self.ritual_tracker.active_rituals.get(
+            self._phase_id(ritual_id=ritual_id, phase=PHASE2)
+        )
+        if async_tx:
+            self.log.info(
+                f"Active ritual in progress Node {self.transacting_power.account} has submitted tx"
+                f"for ritual #{ritual_id}, phase #{PHASE1} (final: {async_tx.final})."
```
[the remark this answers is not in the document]
KPrasch: thanks! ✔️

nucypher/blockchain/eth/trackers/bonding.py
```diff
@@ -14,7 +14,7 @@ class OperatorNoLongerBonded(RuntimeError):
 
     def __init__(self, ursula):
         self._ursula = ursula
-        super().__init__()
+        super().__init__(interval=self.INTERVAL)
```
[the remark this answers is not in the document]
KPrasch: I'm not sure if this is still true with other recent changes on these lines, however the interval was not being set as expected on the super-instance and this was a way to ensure it is.

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -50,15 +51,17 @@ class EventScannerTask(SimpleTask):
 
     INTERVAL = 120  # seconds
 
-    def __init__(self, scanner: Callable, *args, **kwargs):
+    def __init__(self, scanner: Callable):
         self.scanner = scanner
-        super().__init__(*args, **kwargs)
+        super().__init__(interval=self.INTERVAL)
```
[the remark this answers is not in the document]
KPrasch: same answer as above 👍🏻

nucypher/blockchain/eth/trackers/dkg.py
KPrasch: As of the latest commit in this PR, these are not cleared.  I think the most straight-forward solution is to prune final asynx transactions on each round performance.  I'd elect to accept those changes in a subsequent PR, if that's acceptable.

nucypher/characters/lawful.py
```diff
@@ -967,12 +967,19 @@ def run(
...
         if discovery and not self.lonely:
             self.start_learning_loop(now=eager)
             if emitter:
-                emitter.message(f"✓ Node Discovery ({self.domain})", color="green")
+                emitter.message(f"✓ P2P Networking ({self.domain})", color="green")
+
+        if transaction_tracking:
+            # Uncomment to enable tracking for both chains.
+            # mainnet.tracker.start(now=False)
+            polygon.tx_machine.start(now=False)
```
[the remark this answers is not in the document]
KPrasch: The logical grouping of initializing all the async tasks for the lifecycle of the program is good, and I'd hesitate before making an exception for this one.

nucypher/utilities/task.py
```diff
@@ -28,7 +29,7 @@ def start(self, now: bool = False):
         if not self.running:
             d = self._task.start(interval=self.interval, now=now)
             d.addErrback(self.handle_errors)
-            # return d
+            return d
```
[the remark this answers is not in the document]
KPrasch: ✔️

tests/acceptance/actors/test_dkg_ritual.py
```diff
@@ -94,282 +109,238 @@ def condition(test_registry):
...
+        yield clock.advance(interval)
+        yield testerchain.time_travel(seconds=1)
+
+    testerchain.tx_machine.stop()
+    assert not testerchain.tx_machine.running
+
+    status = coordinator_agent.get_ritual_status(ritual_id)
+    assert status == Coordinator.RitualStatus.ACTIVE
+    status = coordinator_agent.get_ritual_status(ritual_id)
+    assert status == Coordinator.RitualStatus.ACTIVE
```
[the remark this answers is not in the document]
KPrasch: rebase artifact ✔️

tests/mock/coordinator.py
```diff
@@ -141,7 +141,7 @@ def post_transcript(
                     p.provider for p in ritual.participants
                 ],  # TODO This should not be
             )
-        return self.blockchain.FAKE_TX_HASH
+        return self.blockchain.FAKE_ASYNX_TX
```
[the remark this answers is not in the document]
KPrasch: ✔️ (and other instances)

tests/unit/test_operator_bonded_tracker.py
KPrasch: it's already running at the time this test is executed so this call had no impact. Also, this test awaits the deferred from `.run` directly in order to make it's assertions, which pass.

nucypher/blockchain/eth/interfaces.py
```diff
@@ -521,141 +568,167 @@ def sign_and_broadcast_transaction(
...
-            use_pending_nonce = replace  # TODO: #2385
+    def send_async_transaction(
+        self,
+        contract_function: ContractFunction,
+        transacting_power: TransactingPower,
+        transaction_gas_limit: Optional[int] = None,
+        gas_estimation_multiplier: float = 1.15,
+        info: Optional[Dict] = None,
+        payload: dict = None,
+    ) -> TxReceipt:
```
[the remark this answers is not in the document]
KPrasch: ✔️

nucypher/blockchain/eth/actors.py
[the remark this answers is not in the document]
derekpierre: > I suppose it can be promoted to a NamedTuple and it will clean this up even more.

That's what I was thinking. That being said, I trust your judgement, so your call.

nucypher/blockchain/eth/actors.py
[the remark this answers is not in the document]
derekpierre: Truthfully I do. I want it to fail spectacularly to make it noticeable 😅. 

It is something that should/would be easily encountered on testnets.

nucypher/blockchain/eth/trackers/bonding.py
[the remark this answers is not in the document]
derekpierre: That was the original bug, but that was fixed via https://github.com/nucypher/nucypher/pull/3435. Post that change, were you still observing a problem?

tests/unit/test_operator_bonded_tracker.py
derekpierre: > t's already running at the time this test is executed

Why is it already running? Is something else calling `start()`?

> Also, this test awaits the deferred from .run directly in order to make it's assertions, which pass.

It does but one additional part of the test is also ensuring that stopping the tracker works (L32)

nucypher/blockchain/eth/actors.py
derekpierre: One reason for the question is, is there an edge case we need to be worried about...maybe...? 

Presumably final being True means that the transaction has completed in some way (successful or not, but completed). If we get here then the contract indicates that action is needed by the node, but if the async tx is final yet the contract still claims that the node needs to act, what do we do then? I may be misunderstanding the asynctx library...

nucypher/characters/lawful.py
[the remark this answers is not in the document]
derekpierre: Not sure I follow...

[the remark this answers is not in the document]
KPrasch: > Does the tx_machine need to be stopped as part of the Ursula.stop(...) method?

Not sure about this one but I think it's low impact since the runtime lifecycle normally does not require `stop` to be invoked.  In tests we have control over individual async tasks which is sufficient.  that being said we can include programmatic stopping functionality for completeness

> `.start`

There benefit to keeping calls to `.start` in this function is that we're explicitly keeping all async task initiation for production runtimes inside one scope.  Since `BlockchainInterfaceFactory` operates with an internal cache my initial hesitation is that by moving `.start` to a lower layer we'd be implicitly firing async tasks in tests that we may loose control over.

nucypher/blockchain/eth/trackers/bonding.py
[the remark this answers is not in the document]
KPrasch: I'll try removing it, re-running the tests, and let you know

nucypher/blockchain/eth/actors.py
KPrasch: Another excellent question - these are the types of questions I am trying to answer in atxm.  final will only be true here if the transaction has actually been included in a block, however this object may have a faulted state instead of a final state.  These faults are strategy-specific and can be handled here in nucypher at this layer.  For example, what shall we do with a reverted transaction?  I am not trying to resolve that question in this PR, rather, establish a platform where can can build a proper response.

There are indeed edge cases!

nucypher/blockchain/eth/trackers/bonding.py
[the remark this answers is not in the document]
KPrasch: ✔️

nucypher/blockchain/eth/trackers/dkg.py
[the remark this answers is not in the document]
KPrasch: ✔️

tests/mock/interfaces.py
cygnusv: Well played @derekpierre

newsfragments/3424.feature.rst
```diff
@@ -0,0 +1 @@
+Introduces automated protocol transaction retries
```
cygnusv, line 1: This newsfragment should be removed

nucypher/blockchain/eth/actors.py
```diff
@@ -341,78 +347,40 @@ def _resolve_validators(
 
         return result
 
-    def publish_transcript(self, ritual_id: int, transcript: Transcript) -> HexBytes:
-        tx_hash = self.coordinator_agent.post_transcript(
+    def publish_transcript(self, ritual_id: int, transcript: Transcript) -> AsyncTx:
+        async_tx = self.coordinator_agent.post_transcript(
```
cygnusv, line 351: 👨‍🍳 💋

tests/unit/test_operator_bonded_tracker.py
KPrasch: > Why is it already running? Is something else calling start()?

Yes, but I'm not sure what (too out-of-scope to investigate) In general there some background noise in the form of uncontrolled async tasks while running the integration and acceptance suites.  Cleaning them up will speed up test performance and accuracy but i snot causing any issues I can see right now.

> It does but one additional part of the test is also ensuring that stopping the tracker works (L32)

I don't think this test is useful since stop is not really used for normal node operation and no assertions are made about the call to stop.  I kind of see this line as test tear-down not related to "test_operator_never_bonded".

In any case I restored the diff to it's original state.

nucypher/blockchain/eth/actors.py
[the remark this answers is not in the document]
KPrasch: Spectacular, eh?  Have another look.

nucypher/blockchain/eth/actors.py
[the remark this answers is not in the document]
KPrasch: ✔️  Good call - it's way more formal as a `NamedTuple` (`PhaseId` can be shorthanded as `pid`)

nucypher/blockchain/eth/agents.py
```diff
@@ -186,11 +191,17 @@ def approve_transfer(
...
         self._validate_zero_allowance(amount, spender_address, transacting_power)
 
-        payload: TxParams = {'gas': Wei(500_000)}  # TODO #842: gas needed for use with geth! <<<< Is this still open?
-        contract_function: ContractFunction = self.contract.functions.approve(spender_address, amount)
-        receipt: TxReceipt = self.blockchain.send_transaction(contract_function=contract_function,
-                                                              payload=payload,
-                                                              transacting_power=transacting_power)
+        payload: TxParams = {
+            "gas": Wei(500_000)
+        }  # TODO #842: gas needed for use with geth! <<<< Is this still open?
```
cygnusv, line 196: This whole method can be removed (although not necessarily in this PR)

nucypher/blockchain/eth/agents.py
```diff
@@ -816,23 +844,22 @@ def post_aggregation(
         public_key: DkgPublicKey,
         participant_public_key: SessionStaticKey,
         transacting_power: TransactingPower,
-        fire_and_forget: bool = True,
```
cygnusv, line 819: Ah, so refreshing!

nucypher/blockchain/eth/agents.py
```diff
@@ -796,17 +825,16 @@ def post_transcript(
...
         transcript: Transcript,
         transacting_power: TransactingPower,
-        fire_and_forget: bool = True,
-    ) -> Union[TxReceipt, HexBytes]:
+    ) -> AsyncTx:
         contract_function: ContractFunction = self.contract.functions.postTranscript(
             ritualId=ritual_id, transcript=bytes(transcript)
         )
-        receipt = self.blockchain.send_transaction(
+        atx = self.blockchain.send_async_transaction(
```
cygnusv, line 832: Not very important, but I feel that `async_tx` is more descriptive (and also it has more occurences)

nucypher/blockchain/eth/actors.py
```diff
@@ -446,8 +447,12 @@ def perform_round_1(
...
                 f"is not a member of ritual {ritual_id}"
             )
-            self.log.error(message)
-            return
+            try:
+                raise RuntimeError(message)
+            finally:
+                # log and stop reactor; this will crash the application.
+                self.log.critical(message)
+                return reactor.stop()
```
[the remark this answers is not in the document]
KPrasch: Hmm, that seems like strictly worse behavior than just logging it loudly and moving on without needed to reschedule the task.  If the goal is to catch this early and loudly, then imo crashing its the best way to achieve that.  Personally, I still prefer a critical log and return, it's the safest option but still produces a use error mode.

nucypher/blockchain/eth/agents.py
```diff
@@ -816,23 +844,22 @@ def post_aggregation(
...
             dkgPublicKey=Ferveo.G1Point.from_dkg_public_key(public_key),
             decryptionRequestStaticKey=bytes(participant_public_key),
         )
-        receipt = self.blockchain.send_transaction(
+        atx = self.blockchain.send_async_transaction(
             contract_function=contract_function,
             gas_estimation_multiplier=1.4,
             transacting_power=transacting_power,
-            fire_and_forget=fire_and_forget,
+            info={"ritual_id": ritual_id, "phase": PHASE2},
```
cygnusv, line 858: Should't this be a PhaseId? Same comment for the other method.

nucypher/blockchain/eth/trackers/dkg.py
KPrasch: opened https://github.com/nucypher/nucypher/issues/3454

nucypher/blockchain/eth/interfaces.py
```diff
@@ -311,6 +328,7 @@ def connect(self):
         # Connect if not connected
         try:
             self.w3 = self.Web3(provider=self._provider)
+            self.tx_machine.w3 = self.w3  # share this web3 instance with the tracker
```
cygnusv, line 331: This seems a bit hacky. Also, wasn't this already done in L222?

tests/unit/test_operator_bonded_tracker.py
KPrasch: ✔️

nucypher/blockchain/eth/agents.py
KPrasch: sure, it can be.  The only way this dict is used currently is for async logging.  it has not predefined format other than expecting that every values it string-representable.

nucypher/blockchain/eth/interfaces.py
KPrasch: Good eye 😅 

> wasn't this already done in L222?

The `Configuration` object lifecycle includes a stage where `self.w3 is NO_BLOCKCHAIN_CONNECTION` (L219) until `.connect` is called, where the w3 instance is actually resolved.

This line is mimicking the same "connection" behavior as the interface itself.

newsfragments/3424.feature.rst
KPrasch: ✔️

nucypher/blockchain/eth/agents.py
KPrasch: thanks! ✔️

tests/unit/test_operator_bonded_tracker.py
```diff
@@ -46,10 +46,6 @@ def test_operator_bonded_but_becomes_unbonded(mocker, get_random_checksum_addres
...
-        yield d
```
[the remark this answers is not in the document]
KPrasch: ✔️

nucypher/blockchain/eth/agents.py
KPrasch: You see this is an example of the linter getting me in trouble at work! 😆 I hope you don't mind if we leave this for another PR, it's out of scope.

nucypher/blockchain/eth/actors.py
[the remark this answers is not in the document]
derekpierre: It is indeed spectacular in terms of how it fails. It's not something unrecoverable so crashing Ursula seems heavy-handed to me.

>  I still prefer a critical log and return,
In that case, if I have to choose between crashing vs critical log and return, I'd take the critical log and return. Although perhaps include a stack trace in the critical log...

Just logging a statement as critical can be easily missed in logs. wdyt? I don't want to error to go unnoticed during a casual look through the logs; this is what I'm aiming for.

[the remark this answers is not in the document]
KPrasch: ✔️

theref:
@KPrasch first three commits aren't verified. I think you can just do a rebase to fix it

nucypher/characters/lawful.py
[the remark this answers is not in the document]
KPrasch: Pushes cc22ed134bb5eeef1091c73cfa087c78615b0e41 which allows unexposes eager atxm startup functionality from ursula ans allows atxm to manage it's own async task internally.

Also Included commit d75e20f87c62349a8b04030199a30c1be742afd2 which explicitly returns `None` from all `SimpleTask` subclass start and stop calls.

nucypher/blockchain/eth/agents.py
KPrasch: addressed in 8ddd452137cd37f3b0d99a04ea9b7eda65af63d1

KPrasch:
Tracking the test failure cause here https://github.com/nucypher/nucypher/issues/3455

nucypher/blockchain/eth/interfaces.py
```diff
@@ -107,28 +111,32 @@ def __init__(self,
...
             else:
-                cost = transaction_fee + self.payload.get('value', 0)
-                message = f'{self.name} from {self.payload["from"][:8]} - {self.base_message}.' \
-                          f'Calculated cost is {prettify_eth_amount(cost)},' \
-                          f'but sender only has {prettify_eth_amount(self.get_balance())}.'
+                cost = transaction_fee + self.payload.get("value", 0)
+                message = (
+                    f'{self.name} from {self.payload["from"][:8]} - {self.base_message}.'
+                    f"Calculated cost is {prettify_eth_amount(cost)},"
+                    f"but sender only has {prettify_eth_amount(self.get_balance())}."
```
derekpierre, line 138:
```suggestion
                    f'{self.name} from {self.payload["from"][:8]} - {self.base_message}. '
                    f"Calculated cost is {prettify_eth_amount(cost)}, "
                    f"but sender only has {prettify_eth_amount(self.get_balance())}."
```

nucypher/blockchain/eth/interfaces.py
```diff
@@ -521,141 +569,167 @@ def sign_and_broadcast_transaction(
...
+            )
 
             # Secondary check
             tx = self.client.get_transaction(txhash)
             if tx["gas"] == receipt["gasUsed"]:
-                raise self.InterfaceError(f"Transaction consumed 100% of transaction gas."
-                                          f"Full receipt: \n {pprint.pformat(receipt, indent=2)}")
+                raise self.InterfaceError(
+                    f"Transaction consumed 100% of transaction gas."
+                    f"Full receipt: \n {pprint.pformat(receipt, indent=2)}"
```
derekpierre, line 660:
```suggestion
                    f"Transaction consumed 100% of transaction gas. "
                    f"Full receipt: \n {pprint.pformat(receipt, indent=2)}"
```

derekpierre (approved):
Well done! 🎸

nucypher/blockchain/eth/interfaces.py
KPrasch: ✔️

nucypher/blockchain/eth/interfaces.py
KPrasch: ✔️

KPrasch:
Thanks for the details reviews everyone! 🤠

PR #3438, opened 2024-02-08
Relock Dependencies

KPrasch:


cygnusv:
wdyt if we remove `bytestring-splitter = ">=2.4.0"` here and we just close #3430 ? It's probably the safest approach.

Pipfile
```diff
@@ -9,38 +9,38 @@ python_version = "3"
...
 requests = "*"
 mako = "*"
 # CLI
 click = ">=7.0"
 colorama = "*"
 tabulate = "*"
 # Serialization
-bytestring-splitter = ">=2.4.0"
 marshmallow = "*"
-msgpack = "*"
```
cygnusv, line 34: Note that existing PRE examples use msgpack (although those are broken anyway)

Pipfile
```diff
@@ -9,38 +9,38 @@ python_version = "3"
...
 appdirs = "*"
 constant-sorrow = ">=0.1.0a9"
 maya = "*"
 pendulum = ">=3.0.0b1"
 prometheus-client = "*"
-setuptools = "*"  # for distutils
-urllib3 = "<2,>=1.26.16"  # eth-ape
+setuptools = "*"
+urllib3 = "*"
+watchdog = "<4,>=3"
```
cygnusv, line 43: Where are we using this now?

KPrasch: it's a sub-dependency of both eth-ape and hendrix.  don't ask me how I know 😅

cygnusv: If it's a subdependency, do we need to include it in our Pipfile? Unless the `"<4,>=3"` is particularly relevant...

KPrasch: Unfortunately pipenv is unable to correctly resolve a complete version freeze in many situations, and this is a way to enforce a restriction at the top of the dependency tree that satisfies the resolver.  In other words, without this pin it's not possible to create a version lock without also fiddling around with some of the other dependency versions.  This is essentially true for most of the pins in `Pipfile` 's history that are not top-level dependencies (directly imported).

There are alternative build systems we may consider, or just simply loosen up and accept some amount of version incompatibility.

Pipfile
KPrasch: `msgpack` is still a project dependency since another dependency depends on it, but it does not need to be pinned.

KPrasch: shipping

PR #3439, opened 2024-02-08
skip tests using live network requests

KPrasch:


tests/unit/test_memory_certs.py
```diff
@@ -78,6 +78,7 @@ def test_send_request(session, mocker):
     assert response == "response"
 
 
+@pytest.mark.skip(reason="This test uses a real network connection")
```
cygnusv, line 81: wdy about using `xfail` instead of `skip`?

KPrasch, line 81:
```suggestion
@pytest.mark.xfail(reason="This test uses a real network connection")
```

tests/unit/test_memory_certs.py
```diff
@@ -91,6 +92,7 @@ def test_https_request_with_cert_caching():
     assert response.status_code == 200
 
 
+@pytest.mark.skip(reason="This test uses a real network connection")
```
KPrasch, line 95:
```suggestion
@pytest.mark.xfail(reason="This test uses a real network connection")
```

PR #3441, opened 2024-02-09
Push "latest" tag to Docker hub when published release is latest

theref:


KPrasch (commented):
Thanks for picking up this issue - looking forward to having a fully automated (but also safe) way to publish mainnet releases.  

This is a pretty elaborate approach to tagging! :-) I suggest giving this another iteration using the publish trigger instead of co-mingling with lynx tagging on merge.  Also let's introduce a manual approval step similar to our friends at keep network.

theref:
what sort of manual step? the publishing of the latest release is already manual right?

and yeah sure, i can separate it out into it's own flow

KPrasch:
> what sort of manual step? the publishing of the latest release is already manual right?
> 
> and yeah sure, i can separate it out into it's own flow

It will look so nice next to the PyPI publication! ( https://github.com/nucypher/nucypher/actions/runs/7891730114, well done this works perfectly)

KPrasch (approved):
lgtm - let's give it a try.

PR #3444, opened 2024-02-12
Repetitive calls - get events for multiple event types in one getLogs call

### LINKED ISSUE #3443: Repetitive RPC calls by EventScanner
From Alchemy stats of a node running v7.2.0-dev.11, filtering for `eth_getLogs` exclusively:
image

Observations:
* 24 calls each 5 minutes (1 each 12.5 seconds). For prior version, 8640 calls to eth_getLogs during last 24h (1 call each 10 seconds). So it seems there's not much improvement yet.
* For some reason, there are several simultaneous calls scanning the same range, normally 3 times. For example, look at the block ranges in the first 6 `eth_getLogs` calls: the first 3 have the same range, the next 3 the same, etc. I say simultaneous because, at least according to Alchemy, these calls happen in blocks of 6 that happen roughly at the same time (at least, within a second). The next block of 6 happened one minute before, etc.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Closes #3443 

Previously we were call eth_getLogs individually for each event type: StartRitual, StartAggregationRound, EndRitual - https://github.com/nucypher/nucypher/blob/v7.2.x/nucypher/utilities/events.py#L239 . 

Instead we can make one eth_getLogs call for all 3 event types. Our volume of rpc calls should now be 1/3 of what it was previously.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/utilities/events.py
+++ b/nucypher/utilities/events.py
@@ -11,7 +11,6 @@ where events are added wherever the scanner left off.
 import csv
 import datetime
 import json
-import logging
 import time
 from abc import ABC, abstractmethod
 from collections import OrderedDict
@@ -30,6 +29,7 @@ from web3.types import BlockIdentifier
 
 from nucypher.blockchain.eth.agents import EthereumContractAgent
 from nucypher.blockchain.eth.events import EventRecord
+from nucypher.utilities.logging import Logger
 
 
 def generate_events_csv_filepath(contract_name: str, event_name: str) -> Path:
@@ -68,7 +68,7 @@ def write_events_to_csv_file(csv_file: Path,
     return True
 
 
-logger = logging.getLogger(__name__)
+logger = Logger("events")
 
 
 class EventScannerState(ABC):
@@ -155,7 +155,7 @@ class EventScanner:
         :param chain_reorg_rescan_window: Number of blocks to rescan in case of chain reorganization (to prevent missed blocks)
         """
 
-        self.logger = logger
+        self.logger = Logger(self.__class__.__name__)
         self.contract = contract
         self.web3 = web3
         self.state = state
@@ -268,7 +268,9 @@ class EventScanner:
         # from our in-memory cache
         block_when = get_block_when(block_number)
 
-        logger.debug("Processing event %s, block:%d count:%d", event["event"], event["blockNumber"])
+        self.logger.debug(
+            f"Processing event {event['event']}, block: {block_number}",
+        )
         processed = self.state.process_event(block_when, event)
         return processed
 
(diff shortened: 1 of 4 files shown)


nucypher/utilities/events.py
```diff
@@ -378,45 +374,54 @@ def _fetch_events_for_all_contracts(
...
+    event_topics_to_abis = {}
+    for event_type in events:
+        event_abi = event_type._get_event_abi()
+        event_topic = encode_hex(event_abi_to_log_topic(event_abi))  # type: ignore
+        topics.add(event_topic)
+        event_topics_to_abis[event_topic] = event_abi
+
+    event_filter_params = {
+        "address": contract.address,
+        "topics": [list(topics)],
```
cygnusv, line 394:
```suggestion
        "topics": list(topics),
```
This should be a list of topics, right?

nucypher/utilities/events.py
```diff
@@ -378,45 +374,54 @@ def _fetch_events_for_all_contracts(
...
+        "fromBlock": from_block,
+    }
+    if to_block is not None:
+        event_filter_params["toBlock"] = to_block
+
+    logger.debug(
+        f"Querying eth_get_logs with the following parameters: {event_filter_params}"
```
cygnusv, line 401:
```suggestion
        f"Querying eth_getLogs with the following parameters: {event_filter_params}"
```

nucypher/utilities/events.py
```diff
@@ -361,12 +360,9 @@ def scan(
...
-def _fetch_events_for_all_contracts(
-        web3,
-        event,
-        argument_filters: dict,
-        from_block: int,
-        to_block: int) -> Iterable:
+def _fetch_events_for_contract(
+    web3, contract, events, from_block: int, to_block: int
+) -> Iterable:
     """Get events using eth_get_logs API.
```
cygnusv, line 366:
```suggestion
    """Get events using eth_getLogs API.
```

manumonti (approved):
Good improvement! LGTM 👌

KPrasch:
Looking mostly good! This issues appears to be strictly an optimization, did you mean to base it over v7.2.x (frozen)?

nucypher/utilities/events.py
derekpierre: No. It is a list of list on purpose. See https://web3py.readthedocs.io/en/stable/filters.html#event-log-filters.

For searching for events using topics the first element is the set of topics, and subsequent entries are arguments to filter the events by.

In our case we want multiple topics (events) to be searched for at the same time with no arguments (essentially don't filter on any arguments) so:
`[[A, B, C]]` means search for events with topics A OR B OR C

`[A, B, C]` would mean search for events with topic A, and arguments B, C.

derekpierre:
> Looking mostly good! This issues appears to be strictly an optimization, did you mean to base it over v7.2.x (frozen)?

Yes, I did mean to base it over `v7.2.x` for a couple reasons:
1. Of all of the RPC call reduction work we have done on v7.2.x thus far, this will have the biggest impact. It should reduce the number of rpc calls to 1/3 of what it was before which is very significant, and was one of the goals of 7.2.
2. It is a very specific change in a limited area (single file) of the code base.

derekpierre:
Updated screenshot of alchemy endpoint being used by lynx-3 which is currently running the fix.
Screenshot 2024-02-13 at 8 19 18 AM

KPrasch (approved):

PR #3457, opened 2024-02-16
Poetry

KPrasch:


KPrasch:
Overall the build time is quite a bit faster, however for some reason solidity compile time is much longer (perhaps an less optimized sub-dependency was resolved for eth-ape?).  In any case I think a future relock may resolve the slowness since we need to pin eth-ape <7 until the remapping issue is resolved.

piotr-roslaniec (approved):

manumonti (approved):
LGTM!

We should have done this long time ago.

setup.py
```diff
@@ -57,90 +31,16 @@
...
-
-URSULA_REQUIRES = ["sentry-sdk"]
-
 EXTRAS = {
-
-    # Admin
-    "dev": DEV_REQUIRES + URSULA_REQUIRES,
-    "benchmark": DEV_REQUIRES + BENCHMARK_REQUIRES,
-    "deploy": DEPLOY_REQUIRES,
-    "ursula": URSULA_REQUIRES,
```
derekpierre, line 143: This "ursula" extra is used by the Dockerfile for building `nucypher/nucypher`...if we are getting rid of it, we should update the Dockerfile.

setup.py
```diff
@@ -181,8 +75,9 @@ def read_requirements(path):
...
     description=ABOUT['__summary__'],
     license=ABOUT['__license__'],
-    long_description_content_type="text/markdown",
-    long_description_markdown_filename='README.md',
-    keywords="nucypher, proxy re-encryption",
-    classifiers=PYPI_CLASSIFIERS
+    keywords="nucypher, threshold, taco, "
+    "threshold cryptography, "
+    "distributed key generation, "
+    "proxy re-encryption",
```
derekpierre, line 81: should we get rid of "proxy re-encryption" and use "threshold access contract" / "taco"?

.bumpversion.cfg
```diff
@@ -1,7 +1,7 @@
 [bumpversion]
 current_version = 7.3.0
 commit = True
-tag = True
+tag = False
```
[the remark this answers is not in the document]
KPrasch: I admit it's a bit out of scope but I don't think we're using bumpversion auto-tagging anymore?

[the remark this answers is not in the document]
derekpierre: I guess you mean, now whenever we run bumpversion we always use `--no-tag` anyway. Ok, that makes sense.

setup.py
KPrasch: ✔️ thinned it out a bit

setup.py
KPrasch: ✔️

nucypher/blockchain/eth/interfaces.py
```diff
@@ -50,7 +43,7 @@
 from nucypher.utilities.logging import Logger
 
 Web3Providers = Union[
```
derekpierre, line 45: Seems like this type is not even used...and can potentially be removed.

derekpierre (approved):
🎸

PR #3476, opened 2024-04-11
DKG Fault Tolerance Improvements

### LINKED ISSUE #3453: Handle Reverted/Faulted DKG Transactions
As hard as Ursula might try - sometimes a transaction does not succeeded after a while.  What responses and actions can nodes take to handle transactions that faulted due to reverted, timeout, hitting the spending cap, etc.?

### LINKED ISSUE #3454: Prune `PhaseId` from `ActiveRitualTracker`
The `RitualTracker` maintains a dictionary of `AsyncTx` objects that  grows slowly but is unbounded.   These can be cleaned up when dkg transactions are finalized or otherwise fully resolved.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
- Utilize ATxM asynchronous callbacks to facilitate more dkg fault tolerance. The callbacks are all mandatory by ATxM except for `on_broadcast` since it is simply a success scenario:
  - `on_broadcast`
      - This means that tx was successfully broadcasted to an rpc provider - we simply log in this case, similar to what is done for synchronous txs
  - `on_broadcast_failure`
      - This means that the tx was not successfully broadcasted to an rpc provider - resubmitting wouldn't really matter, so we simply allow the tx to remain in atxm and get retried periodically 
  - `on_fault`
      - This means that the tx had some unrecoverable failure, or spend too long waiting to be executed by atxm, so atxm will no longer track it. This will trigger a resubmission of the original tx by the node
  - `on_finalized`
      - This has two cases: 1) tx was finalized and successful, 2) tx was reverted and unsuccessful. in 1)'s case the node simply stops tracking the async tx because the changes are now reflected on the blockchain. In 2)'s case the node will resubmit the tx. 
  - `on_insufficient_funds`
      - Not much we can do about this at the moment, other than simply log. Expansion of this functionality can be iterated over in future changes.    
 
_Txs are first broadcasted by ATxM and then finalization implies the availability of a blockchain receipt_

 - Nodes utilize the speedup strategy provided by ATxM to speed up txs that are taking too long. The default time period before waiting to speed up a tx is 2 mins., but this is parameterizable. If you have strong feelings about this value, please voice them.
 
**Issues fixed/closed:**

- Related to https://github.com/nucypher/tdec/issues/151
- Depends on https://github.com/nucypher/ATxM/pull/31
- Fixes #3453 
- Fixes #3454 

**Why it's needed:**

This change facilitates the ability for nodes to be more fault tolerant when the initial tx for posting a transcript or posting an aggregated transcript fail for some reason.

**Notes for reviewers:**

View the changes here first - https://github.com/nucypher/ATxM/pull/31. It is an intricate system.


nucypher/blockchain/eth/actors.py
```diff
@@ -347,14 +351,84 @@ def _resolve_validators(
 
         return result
 
+    def _setup_async_hooks(
```
manumonti, line 354: I like the clarity with which the different scenarios are set up here.

manumonti (approved):
Wow, it has been a great work. LGTM! 🙌

piotr-roslaniec (approved):

vzotova (approved):

PR #3478, opened 2024-04-15
Add --metrics-listen-address CLI Prometheus option

### LINKED ISSUE #3451: `--metrics-listen-address` disappeared
**Describe the Bug**
I can't choose an interface for exposing Prometheus metrics.

The flag `--metrics-listen-address` has disappeared from the list in v7.2.0.

The metrics endpoint is exposed using the same interface as the REST endpoint. This behavior may lead to security issues, especially when the Node Operator doesn't use a firewall with proper rules.

**System:**
 - Nucypher Version: v7.2.0

manumonti:
**Type of PR:**
- [X] Bugfix

**Required reviews:** 
- [X] 2

**What this does:**
It enables a CLI option to specify which of the IP network interfaces available in the machine will be used to expose the metrics (Prometheus exporter).

**Issues fixed/closed:**
Fixes #3451.

### DIFF
--- a/tests/integration/cli/test_ursula_cli_prometheus.py
+++ b/tests/integration/cli/test_ursula_cli_prometheus.py
@@ -77,7 +77,7 @@ def test_ursula_cli_prometheus(
     ), "Wrong value for start_now in prometheus_config"
 
 
-def test_ursula_cli_prometheus_metrics_non_default_port_and_interval(
+def test_ursula_cli_prometheus_metrics_non_default_config(
     click_runner,
     mocker,
     ursulas,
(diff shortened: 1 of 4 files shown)


piotr-roslaniec (approved):

tests/integration/cli/test_ursula_cli_prometheus.py
```diff
@@ -88,6 +88,7 @@ def test_ursula_cli_prometheus_metrics_non_default_port_and_interval(
 ):
     port = 6666
     interval = 30
+    listen_address = "192.0.2.101"
```
derekpierre, line 91: Nice test!

derekpierre (commented):
🎸

derekpierre (approved):

PR #3479, opened 2024-04-15
Node/Contract Compatibility

### LINKED ISSUE #3466: Restarting prometheus metrics task
**Describe the Bug**
The application prints the following error every few seconds:

```
	Error during prometheus metrics collection: Traceback (most recent call last):	
  File "/usr/local/lib/python3.12/site-packages/twisted/internet/base.py", line 695, in run	
    self.mainLoop()	
  File "/usr/local/lib/python3.12/site-packages/twisted/internet/base.py", line 705, in mainLoop	
    self.runUntilCurrent()	
  File "/usr/local/lib/python3.12/site-packages/twisted/internet/base.py", line 1090, in runUntilCurrent	
  File "/usr/local/lib/python3.12/site-packages/twisted/internet/task.py", line 251, in __call__	
    d = maybeDeferred(self.f, *self.a, **self.kw)	
  File "/usr/local/lib/python3.12/site-packages/twisted/internet/defer.py", line 209, in maybeDeferred	
    result = f(*args, **kwargs)	
  File "/usr/local/lib/python3.12/site-packages/nucypher/blockchain/eth/trackers/prometheus.py", line 14, in run	
    collector.collect()	
  File "/usr/local/lib/python3.12/site-packages/nucypher/utilities/prometheus/collector.py", line 55, in collect	
    self._collect_internal()	
  File "/usr/local/lib/python3.12/site-packages/nucypher/utilities/prometheus/collector.py", line 233, in _collect_internal	
    child_application_agent.is_operator_confirmed(	
  File "/usr/local/lib/python3.12/site-packages/nucypher/blockchain/eth/decorators.py", line 94, in wrapped	
    return func(*args, **kwargs)	
  File "/usr/local/lib/python3.12/site-packages/nucypher/blockchain/eth/agents.py", line 483, in is_operator_confirmed	
    staking_provider_info = self.staking_provider_info(staking_provider)	
  File "/usr/local/lib/python3.12/site-packages/nucypher/blockchain/eth/decorators.py", line 94, in wrapped	
    return func(*args, **kwargs)	
  File "/usr/local/lib/python3.12/site-packages/nucypher/blockchain/eth/agents.py", line 475, in staking_provider_info	
    return TACoChildApplicationAgent.StakingProviderInfo(*result)	
builtins.TypeError: StakingProviderInfo.__new__() takes 5 positional arguments but 7 were given

Restarting prometheus metrics task!
```

I also noticed that the metric `operator_matic_balance` is `0.0`.
I don't know when this has been happening. I just noticed the above today.

 - Nucypher Version: v7.2.0 (Docker image)

derekpierre:
Based over #3476 .

**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

Fixes #3466 
Related to https://github.com/nucypher/nucypher-contracts/pull/256

**Why it's needed:**
- Only retrieve struct values from contracts based on a number of expected fields in the matching Python struct. This better maintains compatibility whenever the struct in the contract adds fields. This manifests as #3466 .
- Remove old and no longer used code associated with operator information.
- Update the embedded registry based on the updated contract registry from `nucypher-contracts`. This was missed in the last update, **BUT** the mainnet nodes were fine because they pulled the registry from GitHub as a priority over the embedded registry.
- Agents need to better handle overloaded functions. Web3py does not resolve the function based on the number of args provided, so we need to resolve the function ourselves.  There are two such methods `getActiveStakingProviders` in `TACoChildApplication`, and `getParticipants` in `Coordinator`.  In some cases ("getParticipants") we've been lucky that the function we wanted to use happened to be one used instead of the other one, otherwise the call would fail due to abi <-> parameter mismatch. This change makes the method to use more explicit by using its signature to find the method first. Eventually, the old function that was overloaded will be removed from the contract and is therefore no longer a problem - however, until then, we need to properly/deterministically handle the overload case. 

For example, the issue with `getActiveStakingProviders` manifests itself whenever sampling is done from python:
```
E   web3.exceptions.Web3ValidationError:
E   Could not identify the intended function with name `getActiveStakingProviders`, positional arguments with type(s) 'int,int' and keyword arguments with type(s) '{}'.
E   Found 1 function(s) with the name `getActiveStakingProviders`: ['getActiveStakingProviders(uint256,uint256,uint32)']
E   Function invocation failed due to improper number of arguments.
```
(^ 7.2.0 code running with latest contract updates)

When nodes upgrade from older code to the latest code the original (older) overloaded functions can be removed from the contract, and then the use of a signature to differentiate the two will no longer be necessary.

We solve these specific cases for now since the use of overloaded functions is only really for backwards compatibility which can be broken once the update version of a node has been released for long enough for everyone to have upgraded i.e. the "overloading" is only needed temporarily, and using the signature for all agent functions is pretty ugly.

- Added a test to confirm that ContractCondition do not suffer the same issues as agents -> it does not.

**Notes for reviewers:**

Lynx nodes are already running this code.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@ secrets management and dynamic access control.*
 
 ----
 
-# TACo Access Control
+# Threshold Access Control (TACo)
 
-TACo (Threshold Access Control) is end-to-end encrypted data sharing and communication, without the requirement of
+TACo is end-to-end encrypted data sharing and communication, without the requirement of
 trusting a centralized authority, who might unilaterally deny service or even decrypt private user data. It is the only
 access control layer available to Web3 developers that can offer a decentralized service, through a live,
 well-collateralized and battle-tested network.  See more here: [https://docs.threshold.network/applications/threshold-access-control](https://docs.threshold.network/applications/threshold-access-control)
@@ -22,7 +22,7 @@ well-collateralized and battle-tested network.  See more here: [https://docs.thr
 
 NuCypher is a community-driven project and we're very open to outside contributions.
 
-All our development discussions happen in our [Discord server](https://discord.gg/7rmXa3S), where we're happy to answer
+All our development discussions happen in our [Discord server](https://discord.gg/threshold), where we're happy to answer
 technical questions, discuss feature requests,
 and accept bug reports.
 
--- a/newsfragments/3479.misc.rst
+++ b/newsfragments/3479.misc.rst
@@ -1,0 +1,1 @@
+Nodes will better handle the effects of compatible contract changes such as additions to structs and overloaded functions.
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -962,13 +962,6 @@ class Operator(BaseActor):
                 color="green",
             )
 
-    def get_work_is_needed_check(self):
-        def func(self):
-            # we have not confirmed yet
-            return not self.is_confirmed
-
-        return func
-
 
 class PolicyAuthor(NucypherTokenActor):
     """Alice base class for blockchain operations, mocking up new policies!"""
--- a/nucypher/blockchain/eth/agents.py
+++ b/nucypher/blockchain/eth/agents.py
@@ -430,11 +430,14 @@ class TACoChildApplicationAgent(StakerSamplingApplicationAgent):
     def _get_active_staking_providers_raw(
         self, start_index: int, max_results: int
     ) -> Tuple[int, List[bytes]]:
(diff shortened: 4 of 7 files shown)


manumonti (approved):
LGTM! 👌

theref (approved):

piotr-roslaniec (approved):

PR #3487, opened 2024-04-23
One more thing: Individual logging observers need to adhere to log levels

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
> High-level idea of the changes introduced in this PR. 
> List relevant API changes (if any), as well as related PRs and issues.

If a separate library, say `ATxM` (which actually does this), creates their own TwistedLogger then that logger does not filter by log level, so the individual observers need to filter by log level themselves.

As a result, each global observer used by `nucypher` is now wrapped to ensure that messages are filtered by log level.

**Issues fixed/closed:**
> - Fixes #...

Related to #3483 
Fixes #1712 

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**
> What should reviewers focus on? 
> Is there a particular commit/function/section of your PR that requires more attention from reviewers?

PR #3489, opened 2024-04-23
Ritual tx starvation protection

### LINKED ISSUE #3488: For `broadcast_failure` it would be good to remove and resubmit tx to prevent starvation of other rituals
remove & resubmit would ensure that other ritual txs won't get starved, and ensures that modified txs don't persist if the modifications are not working.

derekpierre:
Based over #3487 .

**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**

The `on_broadcast_failure` callback gets called from ATxM because of one of 2 scenarios:
1. recoverable error occurred 3 times
2. non-recoverable error

but the tx is not removed from the atxm queue and is continuously retried until a timeout, which means that this problematic ritual tx can cause starvation for other ritual txs in the queue waiting to be executed. (shout out to @KPrasch for thinking about this).

Instead when `on_broadcast_failure` is called for a specific ritual tx, have it remove the tx from the queue, and queue a fresh tx.

**Issues fixed/closed:**
> - Fixes #...

Fixes #3488 .

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/newsfragments/3489.misc.rst
+++ b/newsfragments/3489.misc.rst
@@ -1,0 +1,1 @@
+Prevent ritual tx starvation when broadcast failures occur by proactively removing problematic tx from the ATxM queue, and resubmitting a fresh tx.
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -385,6 +385,13 @@ class Operator(BaseActor):
                 f"{tx_type} async tx {tx.id} for DKG ritual# {phase_id.ritual_id} "
                 f"failed to broadcast {e}; the same tx will be retried"
             )
+            # either multiple retries already completed for recoverable error,
+            # or simply a non-recoverable error - remove and resubmit
+            # (analogous action to a node restart of old)
+            self.coordinator_agent.blockchain.tx_machine.remove_queued_transaction(tx)
+
+            # submit a new one
+            resubmit_tx()
 
         def on_fault(tx: FaultedTx):
             # fault means that tx was removed from atxm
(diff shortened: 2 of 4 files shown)


manumonti (approved):
Good catch! 🙌 
LGTM

manumonti:
Should we consider this strategy as the default one for ATxM?

If I understand correctly, when broadcast_failure because of `unrecoverable_error` (malformed tx, for instance), the TX is kept in the queue unless you explicitly remove it, which doesn't seem to make sense much sense.

WDYT?

piotr-roslaniec (approved):

derekpierre:
> Should we consider this strategy as the default one for ATxM?
> 
> If I understand correctly, when broadcast_failure because of `unrecoverable_error` (malformed tx, for instance), the TX is kept in the queue unless you explicitly remove it, which doesn't seem to make sense much sense.
> 
> WDYT?

That's a really good observation/question. 

It's possible but the tough part is that ATxM (as a standalone library and not specific to `nucypher`) maintains a FIFO queue, which could contain queued transactions that are dependent on transactions in the queue ahead of it being executed first. So proactively removing and requeueing a transaction can throw the order out of whack. It's one of the reasons the failure callables (`on_fault`, `on_broadcast_failure` etc.) are mandatory (https://github.com/nucypher/ATxM/pull/31). That way instead of ATxM guessing what to do, the user with more context/insight into the transactions to execute can proactively do something about it.

In `nucypher`'s case, the added ritual phase transactions are independent and self-contained, so re-ordering them is fine, but perhaps for other use cases, it isn't that clear-cut/easy.

That being said, more mindshare in this area could yield a reasonable compromise/solution here - so I'll file an issue in ATxM and we can continue this discussion there.

PR #3495, opened 2024-05-06
WIP:Address match condition

AndreaRettaroli:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
> High-level idea of the changes introduced in this PR: The idea is to introduce a native condition that allow to verify a match between two addresses without pass via a smart contract with access list it means no fee required. The address match will be between the first one passed from the Encryptor as condition value. The second address is taken from the parameters, so the address of the signer during the decryption. 
> List relevant API changes (if any), as well as related PRs and issues: Conditions

**Issues fixed/closed:**
> - Fixes: temporary suppressed ruff warnings to do the PR.

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network: With this feature create use cases like: 
- enable only one address to unlock ciphertext will be really easy to set up, no extra costs for users that wants to use this condition. This will provide a lot of effort for the development of [TACo Snap](https://taikai.network/cryptocanal/hackathons/ethdam2024/projects/clux1ntjg00m5z301xb0bdehj/idea), a MetaMask snap that use TACo as a secret storage. Where a secret can be a seed phrase, or passwords.

**Notes for reviewers:**
> What should reviewers focus on? For now just verify if this feature is acceptable for you project, in terms of behaviour.
> Is there a particular commit/function/section of your PR that requires more attention from reviewers? _

@theref


KPrasch:
We have decided to close this PR without accepting the changes since it's already possible to define a condition that only matches a specific address without the need for a higher-level abstraction.    Nonetheless, thanks for taking the time to learn about our codebase and make a contribution :-)  Feel free to join us in the TACo Discord for further discussions

Issue #3493, opened 2024-04-30
Discussion: Expansion of Condition Evaluation RPC providers


cygnusv:
This issue is just trying to start a discussion on the inclusion of fallback RPC providers in the nucypher client. We need to carefully consider the problems and solutions to the tension between adding more chains for TACo conditions and ensuring RPC provider diversity. In a worst case scenario, node operators could not add their own RPC providers for a new supported chain and requests are routed through the same default RPC providers, introducing single points of failure. Regardless of these issues, here's a potential approach to introduce default RPC providers:

1. Introduce in the client code a list of default RPC providers (e.g., from https://chainlist.org/) for each new supported chain. Additionally, keep a "default slack time" for each chain.

2. RPC Provider configuration:
    * Recommended: node operator supplies their own RPC provider (not from the default list, obviously)
    * Fallback option: client chooses one from the default list, in which case a WARNING should be emitted in the logs/console.
 
3. Health check (at least at startup time):
   * Get latest block timestamp T and current local timestamp C
   * If `0 <= T - C < DEFAULT_SLACK_TIME[chain_id]`, continue.
   * Otherwise, blockchain seems stale. Some potential actions:
       * Emit WARNING
       * Don't allow node to continue
       * Enter https://github.com/nucypher/nucypher/issues/3404
    * Ideally, re-check periodically (e.g., hourly?)
    
    
 

5. Consider using #3372 to introduce penalties for stale/unresponsive blockchain. This should disincentivize the use of default providers, which would be left as a last resort and risky configuration.

KPrasch:
I am performing a survey of the codebse to advance the discussion.   I have two initial thoughts

(1) Note that this issue may me slightly mis-scoped because an RPC endpoint fallback solution is already deployed to mainnet as of https://github.com/nucypher/nucypher/pull/3185.

(2) Regardless of how e decide to rollout new condition RPC endpoint requirements to mainnet nodes, we will need to engage the (presently unused, but purpose-built) `/condition_chains` HTTPS endpoint on nodes as part of a negotiation for DKG initialization.

cygnusv:
> (1) Note that this issue may be slightly mis-scoped because an RPC endpoint fallback solution is already deployed to mainnet as of #3185.

Right, naming wasn't great, thanks for taking care of that. What I meant was "hardcoded default fallbacks". 

[...]

PR #3496, opened 2024-05-07
Default RPC endpoints for Condition Evaluation

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
Introduces default red-only RPC endpoints at node startup for use with condition evaluation only.  Performs rudimentary health checks for each endpoint.

**Issues fixed/closed:**
- #3493 

**Why it's needed:**
- Increases the availability and robustness of TACo node's ability to deliver on service requests that involve on-chain reads.

**Notes for reviewers:**
I'm unsure if this PR needs more work.  Running testnet experiments now.

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -275,13 +275,7 @@ class Operator(BaseActor):
     def connect_condition_providers(
         self, endpoints: Dict[int, List[str]]
     ) -> DefaultDict[int, List[HTTPProvider]]:
-        providers = defaultdict(list)
-
-        # de-duplicate, preserving order
-        deduplicated = {}
-        for chain_id, uris in endpoints.items():
-            deduplicated[chain_id] = list(dict.fromkeys(uris))
-        endpoints = deduplicated
+        providers = defaultdict(list)  # use list to maintain order
 
         # check that we have endpoints for all condition chains
         if set(self.domain.condition_chain_ids) != set(endpoints):
@@ -290,7 +284,10 @@ class Operator(BaseActor):
                 f"{set(self.domain.condition_chain_ids) - set(endpoints)}"
             )
 
-        # check that each chain id is supported
+        # ensure that no endpoint uri for a specific chain is repeated
+        duplicated_endpoint_check = defaultdict(set)
+
+        # User-defined endpoints for chains
         for chain_id, endpoints in endpoints.items():
             if not self._is_permitted_condition_chain(chain_id):
                 raise NotImplementedError(
@@ -299,6 +296,12 @@ class Operator(BaseActor):
 
             # connect to each endpoint and check that they are on the correct chain
             for uri in endpoints:
+                if uri in duplicated_endpoint_check[chain_id]:
+                    self.log.warn(
+                        f"Duplicated user-supplied blockchain uri, {uri}, for condition evaluation on chain {chain_id}; skipping"
+                    )
+                    continue
+
                 provider = self._make_condition_provider(uri)
                 if int(Web3(provider).eth.chain_id) != int(chain_id):
                     raise self.ActorError(
@@ -310,13 +313,20 @@ class Operator(BaseActor):
                         f"user-supplied condition RPC endpoint {uri} is unhealthy"
                     )
                 providers[int(chain_id)].append(provider)
+                duplicated_endpoint_check[chain_id].add(uri)
 
         # Ingest default/fallback RPC providers for each chain
         for chain_id in self.domain.condition_chain_ids:
             default_endpoints = get_healthy_default_rpc_endpoints(chain_id)
             for uri in default_endpoints:
+                if uri in duplicated_endpoint_check[chain_id]:
+                    self.log.warn(
+                        f"Duplicated fallback blockchain uri, {uri}, for condition evaluation on chain {chain_id}; skipping"
+                    )
+                    continue
                 provider = self._make_condition_provider(uri)
                 providers[chain_id].append(provider)
+                duplicated_endpoint_check[chain_id].add(uri)
 
         humanized_chain_ids = ", ".join(
             _CONDITION_CHAINS[chain_id] for chain_id in providers
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -37,8 +37,7 @@ from nucypher.blockchain.eth.agents import (
     TACoApplicationAgent,
     TACoChildApplicationAgent,
 )
-from nucypher.blockchain.eth.clients import PUBLIC_CHAINS
-from nucypher.blockchain.eth.constants import NULL_ADDRESS
+from nucypher.blockchain.eth.constants import NULL_ADDRESS, DEFAULT_RPC_ENDPOINTS, PUBLIC_CHAINS
 from nucypher.blockchain.eth.decorators import validate_checksum_address
 from nucypher.blockchain.eth.domains import TACoDomain
 from nucypher.blockchain.eth.interfaces import (
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -54,8 +54,8 @@ from nucypher.blockchain.eth.trackers import dkg
 from nucypher.blockchain.eth.trackers.bonding import OperatorBondedTracker
 from nucypher.blockchain.eth.utils import (
     get_healthy_default_rpc_endpoints,
-    truncate_checksum_address,
     rpc_endpoint_health_check,
+    truncate_checksum_address,
 )
 from nucypher.crypto.powers import (
     CryptoPower,
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -4,7 +4,7 @@ import time
 import traceback
 from collections import defaultdict
 from decimal import Decimal
-from typing import DefaultDict, Dict, List, Optional, Set, Union
+from typing import DefaultDict, Dict, List, Optional, Union
 
 import maya
 from atxm.exceptions import InsufficientFunds
@@ -271,14 +271,13 @@ class Operator(BaseActor):
         provider = HTTPProvider(endpoint_uri=uri)
         return provider
 
-
     def connect_condition_providers(
         self, endpoints: Dict[int, List[str]]
-    ) -> DefaultDict[int, Set[HTTPProvider]]:
-        providers = defaultdict(set)
+    ) -> DefaultDict[int, List[HTTPProvider]]:
+        providers = defaultdict(list)
 
         # check that we have endpoints for all condition chains
-        if self.domain.condition_chain_ids != set(endpoints):
+        if set(self.domain.condition_chain_ids) != set(endpoints):
             raise self.ActorError(
                 f"Missing blockchain endpoints for chains: "
                 f"{self.domain.condition_chain_ids - set(endpoints)}"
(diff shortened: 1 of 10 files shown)


nucypher/blockchain/eth/utils.py
```diff
@@ -87,45 +91,59 @@ def rpc_endpoint_health_check(endpoint: str, max_drift_seconds: int = 60) -> boo
...
     if response.status_code != 200:
+        LOGGER.debug(
+            f"RPC endpoint {endpoint} is unhealthy: {response.status_code} | {response.text}"
+        )
         return False
 
     try:
         data = response.json()
         if "result" not in data:
+            LOGGER.debug(f"RPC endpoint {endpoint} is unhealthy: no response data")
```
theref, line 106: I think this should be a higher level warning, possibly `warn`?

KPrasch: Yeah, that's fair. Maybe I can make this a bit more concise.

nucypher/blockchain/eth/utils.py
```diff
@@ -62,3 +69,97 @@ def get_tx_cost_data(transaction_dict: TxParams):
...
+        LOGGER.debug(f"RPC endpoint {endpoint} is unhealthy: network error")
+        return False
+
+    if response.status_code != 200:
+        LOGGER.debug(
+            f"RPC endpoint {endpoint} is unhealthy: {response.status_code} | {response.text}"
+        )
+        return False
+
+    try:
```
derekpierre, line 103: Seems like we can squash some of these try except blocks:
```python
try:
    ...
    if drift > max:
       return False
except RequestException:
   ...
except KeyError:
    ...
except TypeError:
   ...
except <other>:
   ...
return True
```

nucypher/blockchain/eth/utils.py
derekpierre: We may not want to do `warn` on an individual basis - but perhaps rather based on a proportion of overall endpoints that failed, for example if > 50%  of the 10 total available endpoints are unhealthy, then print a warn message, as opposed to printing a warn message for each endpoint every time it is checked and is unhealthy.

nucypher/blockchain/eth/actors.py
```diff
@@ -294,6 +300,13 @@ def connect_condition_providers(
                     )
                 providers[int(chain_id)].add(provider)
 
+        # Ingest default RPC providers for each chain
+        for chain_id in self.domain.condition_chain_ids:
+            default_endpoints = get_healthy_default_rpc_endpoints(chain_id)
+            for uri in default_endpoints:
+                provider = self._make_condition_provider(uri)
+                providers[chain_id].add(provider)
```
derekpierre, line 308: `providers` is a `set`, but do we want a bias to user-defined endpoints first, before using default endpoints ...? i.e. only if the user-defined endpoints are problematic do we then fallback to use the default ones.

nucypher/blockchain/eth/actors.py
```diff
@@ -266,6 +271,7 @@ def _make_condition_provider(uri: str) -> HTTPProvider:
         provider = HTTPProvider(endpoint_uri=uri)
         return provider
 
+
     def connect_condition_providers(
         self, endpoints: Dict[int, List[str]]
     ) -> DefaultDict[int, Set[HTTPProvider]]:
```
derekpierre, line 277: The creation of this set of providers is only done once on startup. Would checking health need to be ongoing/periodic? When evaluating conditions, I know the node iterates through the set of endpoints if there are multiple available and if prior ones experience execution issues/failures, but should we adjust the order based on problematic endpoints?

For example, say on startup we have 10 endpoints that are healthy, so we always use the 1st endpoint - no problem. However, 2 days later, the first 5 endpoints are now unhealthy; then, for subsequent condition evaluation, we will always iterate through and try the first 5, with each experiencing failure before we move to the next and eventually use the valid 6th endpoint. This will continue until either:
1. The earlier endpoints in the first 5 spots resolve their issues
2. The node restarts

In this way, there may be an efficiency hit if the first 5 always have to fail before we use the 6th. This is probably more pronounced if default RPC endpoints are earlier in the set (e.g., overloading of public endpoints)—see my prior comment about whether we want to bias towards user-defined endpoints.

KPrasch: This is something that has crossed my mind but I have been uncertain about when to perform a refresh.  In fact this can be further broken down into smaller pieces. (1) Refreshing the chainlist/endpoints URLs and (2) performing an additional round of health checks.  It will be good to take an incremental approach, but ideally it will be good if the user-defined endpoints can be always attempted first.

nucypher/blockchain/eth/actors.py
KPrasch: In other words, use a list.  Yes that is sensible, I'll make an update.

nucypher/blockchain/eth/actors.py
```diff
@@ -292,13 +297,22 @@ def connect_condition_providers(
...
+                providers[chain_id].append(provider)
+
+        self.log.info(f"Connected to {len(providers)} default RPC endpoints")
 
         humanized_chain_ids = ", ".join(
             _CONDITION_CHAINS[chain_id] for chain_id in providers
         )
         self.log.info(
-            f"Connected to {len(providers)} blockchains for condition checking: {humanized_chain_ids}"
+            f"Connected to {len(providers)} total endpoints for condition checking: {humanized_chain_ids}"
```
derekpierre, line 315: `len(providers)` is really the number of chain ids, instead of the number of providers/endpoints - there are multiple endpoints per chain. 

You would have to sum the length of each value in the dictionary to get the total endpoints. Since you are going through the endpoints in the `for` loops above, you can keep a running count as you go, perhaps...?

nucypher/blockchain/eth/utils.py
KPrasch: ✔️

nucypher/blockchain/eth/actors.py
KPrasch: Taking these points into consideration, resolved it in a slightly different way.

theref (approved):

nucypher/blockchain/eth/utils.py
```diff
@@ -62,3 +69,107 @@ def get_tx_cost_data(transaction_dict: TxParams):
...
+        "id": 1,
+    }
+    LOGGER.debug(f"Checking health of RPC endpoint {endpoint}")
+    try:
+        response = requests.post(
+            endpoint,
+            json=query,
+            headers={"Content-Type": "application/json"},
+            timeout=5,
+        )
```
cygnusv, line 92: Any particular reason for doing this instead of creating a one-off web3 instance?

nucypher/blockchain/eth/utils.py
```diff
@@ -62,3 +69,107 @@ def get_tx_cost_data(transaction_dict: TxParams):
...
+        LOGGER.error(f"No default RPC endpoints found for chain ID {chain_id}")
+        return healthy
+
+    for endpoint in chain_endpoints:
+        if rpc_endpoint_health_check(endpoint=endpoint):
+            healthy.append(endpoint)
+
+    LOGGER.info(f"Healthy RPC endpoints for chain ID {chain_id}: {healthy}")
+
+    if len(healthy) == 0:
```
cygnusv, line 170:
```suggestion
    if not healthy:
```

nucypher/blockchain/eth/utils.py
```diff
@@ -62,3 +69,107 @@ def get_tx_cost_data(transaction_dict: TxParams):
...
+    endpoints = get_default_rpc_endpoints()
+    chain_endpoints = endpoints.get(chain_id)
+
+    if not chain_endpoints:
+        LOGGER.error(f"No default RPC endpoints found for chain ID {chain_id}")
+        return healthy
+
+    for endpoint in chain_endpoints:
+        if rpc_endpoint_health_check(endpoint=endpoint):
+            healthy.append(endpoint)
```
cygnusv, line 166: Can be done with list comprehension, but you can take it or leave it.

nucypher/blockchain/eth/utils.py
```diff
@@ -62,3 +69,107 @@ def get_tx_cost_data(transaction_dict: TxParams):
...
+    return True  # finally!
+
+
+def get_default_rpc_endpoints() -> Dict[int, List[str]]:
+    """
+    Fetches the default RPC endpoints for various chains
+    from the nucypher/chainlist repository.
+    """
+    LOGGER.debug("Fetching default RPC endpoints from remote chainlist")
+    response = requests.get(CHAINLIST_URL)
```
cygnusv, line 141: We should catch `requests.exceptions.RequestException` (see https://requests.readthedocs.io/en/latest/user/quickstart/#errors-and-exceptions)

KPrasch: Good catch ✔️

nucypher/blockchain/eth/utils.py
KPrasch: ✔️

nucypher/blockchain/eth/utils.py
cygnusv: No, not really. It just got my attention when I saw a line of code parsing hexstrings into ints to get the block timestamp, which is something that web3py handles automatically via the default pythonic middleware.

nucypher/blockchain/eth/actors.py
```diff
@@ -292,13 +298,26 @@ def connect_condition_providers(
...
+            for uri in default_endpoints:
+                provider = self._make_condition_provider(uri)
+                providers[chain_id].append(provider)
 
         humanized_chain_ids = ", ".join(
             _CONDITION_CHAINS[chain_id] for chain_id in providers
         )
         self.log.info(
-            f"Connected to {len(providers)} blockchains for condition checking: {humanized_chain_ids}"
+            f"Connected to {len(providers.values())} RPC endpoints for condition "
```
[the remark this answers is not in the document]
derekpierre: ^ Seems like I missed a closing parentheses in my suggestion above which is causing the CI build to fail.

nucypher/blockchain/eth/actors.py
```diff
@@ -292,13 +298,26 @@ def connect_condition_providers(
...
+                        f"user-supplied condition RPC endpoint {uri} is unhealthy"
+                    )
+                providers[int(chain_id)].append(provider)
+
+        # Ingest default/fallback RPC providers for each chain
+        for chain_id in self.domain.condition_chain_ids:
+            default_endpoints = get_healthy_default_rpc_endpoints(chain_id)
+            for uri in default_endpoints:
+                provider = self._make_condition_provider(uri)
+                providers[chain_id].append(provider)
```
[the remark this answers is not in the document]
KPrasch: After considering this, I don't think it poses any concern.  I will however, cast this list to an intermediate set to de-duplicate.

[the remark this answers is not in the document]
KPrasch: Ended up implementing a strategy that preserves order while still de-duplicating ✔️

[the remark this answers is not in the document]
derekpierre: You are correct that duplicates can exist within the user-supplied endpoints, but duplicates can also exist across user-supplied endpoints *and* fallback endpoints, e.g., if the user-supplied endpoint value was a public RPC endpoint that is also in the fallback list.

For example, supposed in the ursula.json file the user's endpoint for sepolia is `https://rpc.sepolia.org`, that is the same endpoint that is included in the fallback list, hence the duplication issue.

Here is a commit that addresses what I meant - https://github.com/derekpierre/nucypher/commit/9ca2a2a042eda7e39a39c629991ab7055236e64e. Take a look. You can cherry-pick it or do your own implementation as deemed necessary.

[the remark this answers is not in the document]
KPrasch: Ah, Gotcha.  Well, changing datatypes from a set to a list removed duplicate protection for user-supplied endpoints but the case you describe, where a user has supplied an endpoint that is already in the fallback list is a pretty unusual/rare corner case in my opinion.  In fact, it's not completely clear what the ideal behavior or the user's intentions is in this situation 🤔 perhaps we just issue a logging warning? Is this problematic somehow?  Perhaps they intend to prioritize this endpoint?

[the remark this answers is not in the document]
derekpierre: > it's not completely clear what the ideal behavior or the user's intentions is in this situation 🤔 perhaps we just issue a logging warning? 

I think we prioritize the user-supplied endpoint, as we already do with our logic; whatever endpoint they provide. Agreed on the warning - the commit logs a warning for that situation.

> Is this problematic somehow?
It's more of a potential efficiency issue. If the endpoint ends up not being healthy down the road, we may try it twice (and it fail twice) during condition evaluation because of the dupe. This is definitely an edge case, though, but it is a possibility.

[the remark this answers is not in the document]
KPrasch: Your implementation looks workable - but I'm unsure that ignoring a duplicate endpoint is the desired behavior here?  What do you think about crashing instead?  or just permitting the duplicate but issuing a warning?

derekpierre (approved):
🎸

PR #3499, opened 2024-05-13
Compatibility with nucypher/nucypher-contracts #254

KPrasch:


tests/acceptance/conftest.py
```diff
@@ -210,13 +218,27 @@ def taco_child_application(
     return proxy_contract
 
 
+@pytest.fixture(scope="module")
+def adjudicator(nucypher_dependency, deployer_account, taco_application):
```
vzotova, line 222: Adjudicator for taco child is an address not contract (for now), in the future it will be peer snitching protocol. Probably not best name from my side initially

KPrasch: Oh, I see.   So it's okay to just use any random address here for now?

vzotova: yep

vzotova (approved):

PR #3500, opened 2024-05-13
Sequential Conditions

### LINKED ISSUE #3148: Limit `CompoundAccessControlCondition` to prevent overloading Ursulas
Related to discussion in #3140 raised by @piotr-roslaniec - https://github.com/nucypher/nucypher/pull/3140#discussion_r1230923557.

What sort of limits should we put on `CompoundAccessControlConditions`.

```
@[piotr-roslaniec](https://github.com/piotr-roslaniec) piotr-roslaniec [21 minutes ago]
Not specific to this PR:

Should we put a ceiling on the number of elements in conditions? Something in the order of ~100. We could short-circuit it too: evaluate the first 100, and if there exists another then return False.
Seems like a possible DoS vector. WDYT?

@[KPrasch](https://github.com/KPrasch) KPrasch [16 minutes ago]
That's a very sensible restriction, in fact I'd make the ceiling much lower, perhaps 5-10 elements since each call adds o(n) runtime at minimum, and maps directly onto node operational costs. Even beyond that, we may consider some form of rate limiting. All of which needs to be addressed in another PR IMO.

@[derekpierre](https://github.com/derekpierre) derekpierre [13 minutes ago]
Nice thought!

We can potentially limit the length of the list of conditions, and raise an InvalidCondition exception if too long during object creation/schema validation.

I'd rather do that than allow the object to be created then fail during execution because it is too long - could be too confusing to the user. That being said, technically you can work around it by having repeated nested compound conditions each with operands under the limit.

So maybe some kind of overall count needs to be calculated during creation...?

Some good thoughts here. Let's tackle it in another PR.


@[piotr-roslaniec](https://github.com/piotr-roslaniec) piotr-roslaniec [6 minutes ago]
There's definitely both client and server validation involved there, I think this is often the case with the condition module as a whole.

I agree that's material for another contribution. I'll be interested in scoping this, also for the purpose of implementation in nucypher-ts.
```

### LINKED ISSUE #3494: Discussion: Sequential condition resolution
Study supporting sequential resolution of conditions to support multistep condition resolution. For example, a condition may involve returning the result of a function call and feeding the result to a second function call (this becomes more interesting when the function calls are in different chains)

derekpierre:
**Very early and unfinished draft - still not sure if this is the direction we want to go**

**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

### Sequential condition functionality:

Current grammar for sequential conditions:
```python
CONDITION_VARIABLE = {
    "varName": STR,
    "condition": {
        CONDITION
    }
}

SEQUENTIAL_CONDITION = {
    "name": ...  (Optional),
    "conditionType": "sequential",
    "conditionVariables": [CONDITION_VARIABLE*]
}
```

1. Execution a list of conditions in order
2. Allow results of one condition to be passed to any subsequent condition(s).
3. If any condition fails (execution or return value check) then the entire condition fails.
4. All conditions return two values from `verify(...)`:
    - `condition_successful` - True/False i.e. whether the condition passed or not
    - `execution_value` - the value used for evaluating whether the condition was met or not

    For a sequential condition the `execution_value` is a list or each `execution_value` returned by the individual conditions. The index of the value matches the condition index in the list of conditions used.

### Multi-condition Limits
- Concept of Multi-Condition condition type. Basically any condition that uses a list of conditions.
- A Multi-Condition looks like the following:
  ```python
  class MultiCondition:
     self.conditions = List[...].  # list of conditions
  ```
  This currently encompasses compound conditions and sequential conditions
- Limit on number of conditions for a multi-condition type. For now the list size is capped at length 5
- Limit nesting of multi-condition types including combinations of nesting of each other. The limit is set to 2 levels of nesting i.e. you can have a multi condition type that includes a sub multi-condition type but that sub multi-condition cannot then include another multi-condition type.

Basically in the worst case you have a multi-condition, that has all 5 conditions in the list be multi-conditions, meaning that max number of conditions to evaluate for any multi-condition (based on current limits) will be 5 * 5 = 25 conditions

See tests for more information.

### Separation of execution calls from condition schema / return value check
Internally, an `ExecCondition` object (i.e. non-logical condition) is now composed of an `ExecutionCall` and a `ReturnValueTest`. 

For example:
```
RPCCondition:
   - RPCCall
   - ReturnValueTest

ContractCondition:
   - ConditionCall
   - ReturnValueTest
   
TimeCondition:
   - ConditionCall
   - ReturnValueTest

JsonApiCondition:
   - JsonApiCall
   - ReturnValueTest
...
```

The `ExecutionCall` is focused on the obtaining of a value to use for the condition eg. rpc call, contract call etc., and then that result is passed to the `ReturnValueTest`.

This split separates condition calls (i.e. execution) from condition schema & return value checking.

These conditions all follow the same internal paradigm:
```python
class Condition:
    def __init__(...):
        self.execution_call = ...
        self.return_value_test = ...
        
    def verify(...):
        execution_result = self.execution_call.execute().  # the on-chain method to execute
        condition_check_success = self.return_value_test.eval(execution_result).  # check whether result passes the condition check i.e. True/False
        return condition_check_success, execution_result.  # return tuple of ("condition successful", "execution result")
}
```

Therefore condition execution can be easily tested separately from schema/lingo.

**Issues fixed/closed:**

- Closes #3494 
- Closes #3148 
- Closes https://github.com/nucypher/sprints/issues/56

**Why it's needed:**

**Notes for reviewers:**

This is an initial design to build upon, and I'm trying to prevent this code from getting too large and from getting stale over time. Follow-up refinements/improvements can be outlined/discussed here - https://github.com/nucypher/nucypher/issues/3555.

### DIFF
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -232,7 +232,7 @@ _COMPARATOR_FUNCTIONS = {
 #
 # CONDITION = BASE_CONDITION | COMPOUND_CONDITION
 #
-# CALL = RPC_CALL | TIME_CALL | CONTRACT_CALL | ...
+# CALL = RPC_CALL | TIME_CALL | CONTRACT_CALL | JSON_API_CALL ...
 #
 # VARIABLE = {
 #     "varName": STR,
@@ -272,15 +272,12 @@ class _ExecutionCallField(fields.Dict):
             ContractCall,
             RPCCall,
         )
+        from nucypher.policy.conditions.offchain import JsonApiCall
         from nucypher.policy.conditions.time import TimeRPCCall
 
         call_type = execution_call_dict.get("callType")
 
-        for execution_call_type in (
-            RPCCall,
-            TimeRPCCall,
-            ContractCall,
-        ):
+        for execution_call_type in (RPCCall, TimeRPCCall, ContractCall, JsonApiCall):
             if execution_call_type.CALL_TYPE == call_type:
                 return execution_call_type
 
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -75,6 +75,7 @@ class ConditionType(Enum):
     RPC = "rpc"
     JSONAPI = "json-api"
     COMPOUND = "compound"
+    SEQUENTIAL_CONTRACTS = "sequentialContracts"
 
     @classmethod
     def values(cls) -> List[str]:
@@ -425,7 +426,7 @@ class ConditionLingo(_Serializable):
         Inspects a given bloc of JSON and attempts to resolve it's intended  datatype within the
         conditions expression framework.
         """
-        from nucypher.policy.conditions.evm import ContractCondition, RPCCondition
+        from nucypher.policy.conditions.evm import ContractCondition, RPCCondition, SequentialContractCondition
         from nucypher.policy.conditions.offchain import JsonApiCondition
         from nucypher.policy.conditions.time import TimeCondition
 
@@ -438,6 +439,7 @@ class ConditionLingo(_Serializable):
             RPCCondition,
             CompoundAccessControlCondition,
             JsonApiCondition,
+            SequentialContractCondition,
         ):
             if condition.CONDITION_TYPE == condition_type:
                 return condition
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -4,7 +4,7 @@ import json
 import operator as pyoperator
 from enum import Enum
 from hashlib import md5
-from typing import Any, List, Optional, Tuple, Type, Union
+from typing import Any, Dict, List, Optional, Set, Tuple, Type, Union
 
 from hexbytes import HexBytes
 from marshmallow import (
@@ -19,8 +19,13 @@ from marshmallow import (
 )
 from marshmallow.validate import OneOf, Range
 from packaging.version import parse as parse_version
+from web3 import HTTPProvider
 
-from nucypher.policy.conditions.base import AccessControlCondition, _Serializable
+from nucypher.policy.conditions.base import (
+    AccessControlCondition,
+    ExecutionVariable,
+    _Serializable,
+)
 from nucypher.policy.conditions.context import (
     _resolve_context_variable,
     is_context_variable,
@@ -60,6 +65,8 @@ class _ConditionField(fields.Dict):
 # }
 #
 # COMPOUND_CONDITION = {
+#     "name": ...  (Optional)
+#     "conditionType": "compound",
 #     "operator": OPERATOR,
 #     "operands": [CONDITION*]
 # }
@@ -75,7 +82,7 @@ class ConditionType(Enum):
     RPC = "rpc"
     JSONAPI = "json-api"
     COMPOUND = "compound"
-    SEQUENTIAL_CONTRACTS = "sequentialContracts"
+    SEQUENTIAL = "sequential"
 
     @classmethod
     def values(cls) -> List[str]:
@@ -164,7 +171,6 @@ class CompoundAccessControlCondition(AccessControlCondition):
 
         self._validate_operator_and_operands(operator, operands, InvalidCondition)
 
-        self.condition_type = condition_type
         self.operator = operator
         self.operands = operands
         self.condition_type = condition_type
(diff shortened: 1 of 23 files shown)


nucypher/policy/conditions/lingo.py
```diff
@@ -212,6 +224,121 @@ def __init__(self, operand: AccessControlCondition):
...
+
+    def __init__(self, var_name: str, condition: AccessControlCondition):
+        self.var_name = var_name
+        self.condition = condition
+
+
+class SequentialAccessControlCondition(MultiConditionAccessControl):
+    CONDITION_TYPE = ConditionType.SEQUENTIAL.value
+    MAX_NUM_CONDITIONS = 5
+    MAX_MULTI_CONDITION_NESTED_LEVEL = 2
```
arjunhassard, line 256: We can re-parametrize these limits as we encounter more use cases involving sequential calls/conditions, and weigh the implied fee revenue against risk of misuse

nucypher/policy/conditions/lingo.py
```diff
@@ -443,3 +560,46 @@ def check_version_compatibility(cls, version: str):
...
+            self.execution_call = self._create_execution_call(*args, **kwargs)
+        except ValueError as e:
+            raise InvalidCondition(str(e))
+
+        super().__init__(condition_type=condition_type, name=name)
+
+    @abstractmethod
+    def _create_execution_call(self, *args, **kwargs) -> ExecutionCall:
+        """
+        Returns the execution call that the condition executes.
```
arjunhassard, line 590: I guess we can regard an `ExecutionCall` condition as being conditioned on the call returning something that can be processed by the next condition? Or is there not actually a condition here – i.e. whatever is returned, the sequence continues? 

e.g. developer creates an `ExecutionCall` to call a weather API and get today's temperature, which should then be fed into the next `Condition`. If an error is returned rather than an integer – will that still be fed into the `Condition`? Would the developer have to explicitly add a `ReturnValueTest` to the first `ExecutionCall` to avoid that happening?

nucypher/policy/conditions/offchain.py
```diff
@@ -148,6 +121,55 @@ def query_response(self, data: Any) -> Any:
 
         return result
 
+
+class JsonApiCondition(BaseExecAccessControlCondition):
+    """
+    A JSON API condition is a condition that can be evaluated by reading from a JSON
+    HTTPS endpoint. The response must return an HTTP 200 with valid JSON in the response body.
+    The response will be deserialized as JSON and parsed using jsonpath.
+    """
```
arjunhassard, line 130: You can use the logic of sequential conditions to mitigate the issue with unpredictably deprecated APIs. You probably still need updatable conditions eventually

arjunhassard (approved):
This one PR makes TACo's condition lingo twice as powerful! Excellent work

nucypher/policy/conditions/lingo.py
derekpierre: For now, SequentialCondition only uses Conditions (not `ExecutionCalls`) which mandate the use of a `ReturnValueTest`.

`ExecutionCall`, for now, is purely an internal low-level abstraction of the execution that a condition does before any condition is checked. Currently, developers are not exposed to calls at all, only conditions i.e. same as before.

For example, for an RPCCondition, it uses an `RPCCall` object and calls `execute_call.execute()` which simply calls `get_balance(...)` and returns the result. The RPCCondition takes this result and then performs the actual condition check using its ReturnValueTest. 

This repeats i.e. TimeCondition uses TimeRPCCall, ContractCondition uses a ContractCall etc.

So all conditions that involve execution (i.e. not logical calls) follow the same internal paradigm:

```python
class Condition:
    def __init__(...):
        self.execution_call = ...
        self.return_value_test = ...
        
    def verify(...):
        execution_result = self.execution_call.execute().  # the on-chain method to execute
        condition_check_success = self.return_value_test.eval(execution_result).  # check whether result passes the condition check i.e. True/False
        return condition_check_success, execution_result.  # return tuple of ("condition successful", "execution result")
}
```

The returned `condition_check_success` is used to check whether the condition passed or not i.e `True`/`False`

For SequentialConditions if any condition does not pass i.e. the `condition_check_success` value is `False` or the execution raised an exception (eg. ConnectionError connecting to a json web server), no subsequent condition is executed/checked and the overall condition fails. If that condition passes, then the `execution_result` is stored as the `varName` for the ConditionVariable for that condition and subsequent conditions can use `:<varName>` to use the returned value in their execution/return value check.

---

It is possible that based on discussions or feedback, we could decide that SequentialConditions should allow for any combination of Condition and ExecutionCalls so that devs don't always have to provide `ReturnValueTest` if they don't need any sanity checking of values before they are used. This was explored earlier in the PR when I initially mused on it. See 2nd bullet point in https://github.com/nucypher/nucypher/issues/3555.

nucypher/policy/conditions/offchain.py
derekpierre: Interesting thought. Could you expand on that for me?

theref (approved):

KPrasch (approved):
Overall looks good.  There's some heavy lifting going on in the internal reorganization which adds new layers of abstraction.  This does make the internal API more modular and extensible.  I'm unsure about some of the naming conventions for both the internal and public APIs but this PR does a great job at implementing the new functionality without affecting taco-web in a big way.  Issue https://github.com/nucypher/nucypher/issues/3555 covers follow-up work sufficiently. Nice work!

manumonti (approved):
This is a great work! I left some thoughts / doubts, but I don't see anything relevant enough to block the merge.

By the way, this probably allow us to continue ENS condition work: https://github.com/nucypher/taco-web/issues/104

PR #3502, opened 2024-05-14
Sign-in With Ethereum (SIWE) verification for `:userAddress` (Bob decryption request)

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

Adds Sign-in With Ethereum as a valid signing authentication protocol for proving wallet ownership[ when `:userAddress` is required by a condition.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


tests/unit/conditions/test_auth.py
```diff
@@ -0,0 +1,112 @@
+import pytest
+
+from nucypher.policy.conditions.auth import Auth, EIP712Auth, SIWEAuth
```
KPrasch, line 3: Help me to build a mental picture or system metaphor for this new API:  what is the relationship between the `Auth` ABC tree and say, Ciphertext headers?

dev-requirements.txt
```diff
@@ -178,13 +180,13 @@ tzdata==2024.1 ; python_version >= "3.8" and python_version < "4"
...
 watchdog==3.0.0 ; python_version >= "3.8" and python_version < "4"
 wcwidth==0.2.13 ; python_version >= "3.8" and python_version < "4"
 web3==6.15.1 ; python_version >= "3.8" and python_version < "4"
 web3[tester]==6.15.1 ; python_version >= "3.8" and python_version < "4"
 websockets==12.0 ; python_version >= "3.8" and python_version < "4"
-werkzeug==3.0.2 ; python_version >= "3.8" and python_version < "4"
+werkzeug==3.0.3 ; python_version >= "3.8" and python_version < "4"
 wrapt==1.16.0 ; python_version >= "3.8" and python_version < "4"
 yarl==1.9.4 ; python_version >= "3.8" and python_version < "4"
 zipp==3.18.1 ; python_version >= "3.8" and python_version < "4"
```
KPrasch, line 192: Is this relock due to an update or a new dependency?

tests/unit/conditions/test_auth.py
derekpierre: Nothing - this PR is Bob-specific at the moment. the `auth` is related to the authentication/verification of what is provided in `:userAddress`. One potential follow-on thought is that we could add additional auth checks like "confirming the domain allowed via SIWE" eg. what lit does - https://developer.litprotocol.com/v2/accesscontrol/evm/siwe#domain.

This type of auth condition could be made more generally as something like, "check an additional auth property" as an additional condition on the ciphertext:

```json
{
    "conditionType": "auth",
    "scheme": "SIWE",
    "authMessagePropertyToVerify": "domain"  // not a real name just being overly wordy for clarity
    "returnValueTest": {
      comparator: "==",
      value: "localhost:3050",
    },
}
```
but this is still up in the air and needs adopter validation.

nucypher/policy/conditions/auth.py
```diff
@@ -0,0 +1,86 @@
...
+        if address_for_signature != expected_address:
+            # verification failed - addresses don't match
+            raise cls.AuthenticationFailed(
+                f"Invalid EIP712 signature; does not match expected address, {expected_address}"
+            )
+
+
+class SIWEAuth(Auth):
+    @classmethod
+    def authenticate(cls, data, signature, expected_address):
```
piotr-roslaniec, line 67: I see this method is similar to this `validate` method in SIWE example from login.xyz: https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.ts#L195

A more complete example that actually checks the data in `typedData`, not just signature, can be seen here: https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.ts#L211

derekpierre: If I'm understanding what you mean, the answer depends on what is in the message - is the SIWE template message constant (eg. the domain is `taco`), or can it vary.  If it varies then would checking those things be more of add-on auth conditions that would need to be used - that's what I was referring to here - https://github.com/nucypher/nucypher/pull/3502#discussion_r1603258091. Is checking `domain`, `resources` etc. something that is done by default, or are add-on authentication conditions - eg. the following from lit - https://developer.litprotocol.com/v2/accesscontrol/evm/siwe#domain. 

That depends on what is done in the client - do you have thoughts there?

derekpierre: > A more complete example that actually checks the data in typedData, not just signature, can be seen here:
> https://github.com/spruceid/siwe/blob/main/packages/siwe/lib/client.ts#L211

Just to clarify, `verify` is used in siwe-py (not the equivalent of `validate`)- https://github.com/spruceid/siwe-py/blob/main/siwe/siwe.py#L283 - which has the same options options, but aren't currently used.

piotr-roslaniec: Thanks for the context, I was just making a general observation. But I think I'm convinced now that we don't need to do any other steps for authentication on the backend side. It should be enough that we require users to provide a way to validate `domain` and `resources` using the contract conditions.

I think this should hold as long our adopters don't try doing anything funny like reusing existing DIDs issues for something else and not for TACo specifically.

derekpierre: > I think this should hold as long our adopters don't try doing anything funny like reusing existing DIDs issues for something else and not for TACo specifically.

Yeah, in this case do we/should we force the "domain" (or some other individual property) value in the SIWE message to be TACo-specific... that we can check by default.

derekpierre: Another option, if we want to allow SIWE messages originally signed for other uses that are being reused (not sure if we want to do that but providing options ?) while having some guardrails, is that we could consistently enforce a freshness check relative to the `issued-at` value in the SIWE message - <https://docs.login.xyz/general-information/siwe-overview/eip-4361#message-field-descriptions>.

So, if `current time` > the `issued-at` time + `<some default constant offset>` (eg. 2 hours), then the authentication of the provided SIWE message fails.

derekpierre: See [6421aed](https://github.com/nucypher/nucypher/pull/3502/commits/6421aedb8c2e42f3e60a2382321cfda5b29bf78d)

piotr-roslaniec: > Yeah, in this case do we/should we force the "domain" (or some other individual property) value in the SIWE message to be TACo-specific (?)... that we can check by default. Basically SIWE messages must be TACo-specific.

I think it makes a lot of sense. Because I imagine SIWE users already doing some checks and balances to ensure that the DID relates to their application specifically. But issuing auth for said app is not the same as issuing auth for TACo within that dapp, and I think that could leak to some issues.

manumonti (approved):
Good feature! I left a couple of comments/suggestions, but this looks good to me. 👍

theref (approved):

nucypher/policy/conditions/auth.py
```diff
@@ -0,0 +1,99 @@
...
+
+        if address_for_signature != expected_address:
+            # verification failed - addresses don't match
+            raise cls.AuthenticationFailed(
+                f"EIP712 verification failed; signature does not match expected address, {expected_address}"
+            )
+
+
+class SIWEAuth(Auth):
+    # TODO; this is a safety precaution - is this what we want and is this the correct value?
```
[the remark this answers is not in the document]
arjunhassard: ID

tests/unit/conditions/test_auth.py
arjunhassard: Agreed, let's leave the allowed domain confirmation until after we have some adopters of the joint SDK, since we don't know exactly what other auth properties might be requested, and we may still want to combine these more closely with conditions.

nucypher/policy/conditions/auth.py
arjunhassard: > Anyone can sign a message with a TACo-specific property in it (eg. not using taco-web)

What exactly can 'anyone' do here, with regard to accessing data? If they also fulfill the conditions specified, are they spoofing anything?

nucypher/policy/conditions/auth.py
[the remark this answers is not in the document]
arjunhassard: I understand this as the duration after which the signed authentication message is no longer valid. For an end-user consuming a stream of data, would this mean they would be prompted for a signature again every 2 hours?

arjunhassard (commented):
Nice work, added some questions

nucypher/policy/conditions/auth.py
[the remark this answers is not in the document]
manumonti: >I understand this as the duration after which the signed authentication message is no longer valid. For an end-user consuming a stream of data, would this mean they would be prompted for a signature again every 2 hours?

This duration only refers to the time the app will wait since the signature is requested.

I.e. if the user receives on Metamask (or whatever) the signature requests, and the user just leaves the request and after that two-hours-timeout the user does the signature, this signature will not be considered valid.

nucypher/policy/conditions/auth.py
manumonti: > What exactly can 'anyone' do here, with regard to accessing data? If they also fulfill the conditions specified, are they spoofing anything?

If I'm understanding correctly, "signing a message with a TACo-specific property" is not bad per se. It's more like a feature that we don't know if we want for TACo.

The use case is more like a user signing a message with some other application, and reusing this signature for TACo. There is no spoofing, simply they are using the same signature for two different products.

KPrasch (approved):
Integrating SIWE took a surprisingly low diff, the biggest portion here is from the relock.  You already addressed my comments, looking good to me 👍🏻

nucypher/policy/conditions/auth.py
derekpierre: @arjunhassard , 

what @manumonti said above is spot on. There's no spoofing of accounts to gain access, it's more about whether we need a specific SIWE message for TACo or not.

The question is, is TACo its own application OR a service used by an application as it pertains to SIWE? SIWE messages are seemingly for application-level interactions, but TACo doesn't work that way - it has a client-side and a server-side that are "disconnected" (different memory and execution space).

If TACo is its own application, sure, we can add a taco-specific value to the message, but SIWE is just a message template. Because the client-side and server-side are disconnected, you can't guarantee that the message came from using the `taco-web`/`taco-auth` library - any software can generate the same template message, so what are we gaining here?

In the latter case (TACo is just a service and not an app itself) from a UX perspective, signing in to an application using SIWE and then having to sign again using SIWE for TACo decryption could be annoying, in that the user has to sign a different message in addition to to what they already signed for the application? As a service used by the app, decryption using TACo (base functionality for `:userAddress`) just cares about the user is who they say they are with a reasonable message (not stale/expired). That could be the same application sign-in message. Any add-on conditions on the encrypted data may care about more than that, e.g. that the resource being accessed is allowed based on the "resources" value in SIWE message etc., but base TACo doesn't actually care or have the awareness to care. That awareness seems more like an additional condition on the data.

With all of that said, I'm not sure about the best practice for SIWE on the client side. But that is something to think about.

(cc @cygnusv )

nucypher/policy/conditions/auth.py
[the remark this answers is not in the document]
derekpierre: > For an end-user consuming a stream of data, would this mean they would be prompted for a signature again every 2 hours?

If there is a current decryption request that uses a SIWE signature more than 2 hours old, then decryption would fail. Ideally `taco-web`/`taco-auth` would ensure that the signature is refreshed for a decryption operation if the original signature is cutting it close/stale.

PR #3508, opened 2024-05-28
SIWE (EIP4361) Authentication Part 2

### LINKED ISSUE #3506: How does `Not Before` in SIWE affect our 2-hour freshness window
#3502 enacted a 2-hour freshness window based on the `issued-at` value in the SIWE message. However, what about the optional `not-before` value, https://docs.login.xyz/general-information/siwe-overview/eip-4361#message-field-descriptions;  how should that impact the 2-hour freshness check.

Since the `not-before` date could be some timeframe in the future after `issued-at`, what does that mean for our 2-hour freshness check.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
- Add directives for `:userAddressEIP712` and `:userAddressEIP4361` as user address context variables that will be verified based on the relevant scheme
- `:userAddress` will default to using the EIP712 authentication scheme
- Don't do the additional _2 hour from "issued-at" stale check_ when `not-before` is used in SIWE message (#3506)

**Issues fixed/closed:**

- Fixes #3506; somewhat related although not the main focus - really a remnant from #3502.
- Follow-up from #3502 .

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -20,7 +20,7 @@ CONTEXT_PREFIX = ":"
 CONTEXT_REGEX = re.compile(":[a-zA-Z_][a-zA-Z0-9_]*")
 
 USER_ADDRESS_SCHEMES = {
-    USER_ADDRESS_CONTEXT: Auth.AuthScheme.EIP712.value,  # default to EIP712
+    USER_ADDRESS_CONTEXT: None,  # any of the available auth types
     USER_ADDRESS_EIP712_CONTEXT: Auth.AuthScheme.EIP712.value,
     USER_ADDRESS_EIP4361_CONTEXT: Auth.AuthScheme.EIP4361.value,
 }
@@ -53,7 +53,7 @@ def _resolve_user_address(user_address_context_variable, **context) -> ChecksumA
 
         scheme = user_address_info.get("scheme", Auth.AuthScheme.EIP712.value)
         expected_scheme = USER_ADDRESS_SCHEMES[user_address_context_variable]
-        if scheme != expected_scheme:
+        if expected_scheme and scheme != expected_scheme:
             raise UnexpectedScheme(
                 f"Expected {expected_scheme} authentication scheme, but received {scheme}"
             )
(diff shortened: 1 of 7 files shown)


nucypher/policy/conditions/context.py
```diff
@@ -12,18 +13,30 @@
...
+USER_ADDRESS_EIP4361_CONTEXT = ":userAddressEIP4361"
 
 CONTEXT_PREFIX = ":"
 CONTEXT_REGEX = re.compile(":[a-zA-Z_][a-zA-Z0-9_]*")
 
+USER_ADDRESS_SCHEMES = {
+    USER_ADDRESS_CONTEXT: Auth.AuthScheme.EIP712.value,  # default to EIP712
+    USER_ADDRESS_EIP712_CONTEXT: Auth.AuthScheme.EIP712.value,
+    USER_ADDRESS_EIP4361_CONTEXT: Auth.AuthScheme.EIP4361.value,
+}
```
piotr-roslaniec, line 26: I wonder if we could do this substitution on the client side only? I guess not, since it wouldn't be backwards-compatible on the server-side (with existing lingos).

derekpierre: Correct. We can't because we have to be backwards compatible with any condition that already uses `:userAddress`.

derekpierre: If we want it, we can allow *any* auth scheme for `:userAddress`, i.e. either EIP4361 or EIP712 could be allowed.

piotr-roslaniec (approved):

nucypher/policy/conditions/context.py
derekpierre: ✔️  - https://github.com/nucypher/nucypher/pull/3508/commits/0214471462e03529d236ec009a2a333f63112ca9

KPrasch (approved):

PR #3511, opened 2024-06-19
Offchain Condition Evaluation

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
3

**What this does**
- Expands Lingo to accommodate the expression and processing of HTTPS GET requests returning an 200 status code and JSON response body 
- Ursulas evaluate "offchain"/"web2" conditions from HTTPS endpoints
- Introduces`JsonApiCondition` type

**What this does not do**
- Does not expand condition authorization model
- Does not modify the behavior of `:userAddress` or any other context variables

Example lingo:

```json
{
    "conditionType": "json-api",
    "endpoint": "https://api.coingecko.com/api/v3/simple/price",
    "parameters": {
        "ids": "ethereum",
        "vs_currencies": "usd",
    },
    "query": "ethereum.usd",
    "returnValueTest": {">=", "8000.0"}
}
```


tests/acceptance/conditions/test_offchain_evaluation.py
```diff
@@ -0,0 +1,206 @@
+import json
```
derekpierre, line 1: Is it possible to move this module to `tests/unit/conditions`? Is there something in the `acceptance` fixtures that's needed? 

If possible to move, it'll improve the test running time by skipping the excessive setup that acceptance test fixtures perform.

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,142 @@
...
+        @post_load
+        def make(self, data, **kwargs):
+            return OffchainCondition(**data)
+
+    def __init__(
+        self,
+        endpoint: str,
+        query: Optional[str],
+        return_value_test: ReturnValueTest,
+        headers: Optional[dict] = None,
```
derekpierre, line 71: Obviously not related to this PR but rather the ones that build on it, I'm intrigued by the use of headers to specify the "Authentication" header. It's an interesting idea.

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,142 @@
...
+        self.return_value_test = return_value_test
+        self.logger = self.LOGGER
+
+    def fetch(self) -> requests.Response:
+        """Fetches data from the endpoint."""
+        try:
+            response = requests.get(
+                self.endpoint, params=self.parameters, headers=self.headers
+            )
+            response.raise_for_status()
```
derekpierre, line 93: Cool - didn't know about this.

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,142 @@
...
+
+    class Schema(CamelCaseSchema):
+
+        name = fields.Str(required=False)
+        condition_type = fields.Str(
+            validate=validate.Equal(ConditionType.OFFCHAIN.value), required=True
+        )
+        headers = fields.Dict(required=False)
+        parameters = fields.Dict(required=False)
+        endpoint = fields.Str(required=True)
```
piotr-roslaniec, line 50: What do you think about adding a facility (in this PR or just in general) to restrict access to certain offchain elements. I think less permissive is better (an allowlist). One example of restricted-by-default element could be an `http://` request. We could restrict all headers by default too. 

This is something we should discuss in respect to the security of offchain evaluation.

piotr-roslaniec (approved):

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,142 @@
...
+        return result
+
+    def verify(self, **context) -> Tuple[bool, Any]:
+        """
+        Verifies the offchain condition is met by performing a read operation on the endpoint
+        and evaluating the return value test with the result. Parses the endpoint's JSON response using
+        JSONPath.
+        """
+        response = self.fetch()
+        data = self.deserialize_response(response)
```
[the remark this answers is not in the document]
KPrasch: At least up to this point, I tried to enforce status 200 only.  I made an update to explicitly enforce this.

tests/acceptance/conditions/test_offchain_evaluation.py
KPrasch: ✔️

nucypher/policy/conditions/offchain.py
piotr-roslaniec: > Perhaps we just don't expose headers at all?

Maybe not until it's functionally necessary?

> With regard to HTTP requests - let's just block them altogether?

I think so

KPrasch: ✔️

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,142 @@
...
+    ):
+        if condition_type != self.CONDITION_TYPE:
+            raise InvalidCondition(
+                f"{self.__class__.__name__} must be instantiated with the {self.CONDITION_TYPE} type."
+            )
+
+        self.endpoint = endpoint
+        self.headers = headers
+        self.parameters = parameters
+        self.query = query
```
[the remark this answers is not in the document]
derekpierre: > Or are you thinking that of the conditions API as a top-level python API for developer consumption?

Yes - more related to the top-level python API i.e. someone creating the object from Python. If we can reuse the Schema for validation, all the more better.

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,142 @@
...
+    def query_response(self, data: Any) -> Any:
+        try:
+            expression = parse(self.query)
+            matches = expression.find(data)
+            if not matches:
+                self.logger.info("No matches found for the JSONPath query.")
+                raise ConditionEvaluationFailed(
+                    "No matches found for the JSONPath query."
+                )
+            result = matches[0].value
```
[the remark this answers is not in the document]
derekpierre: > In my experimentation with the coingecko API, it's possible that even JSON responses that have length of one can still return multiple results

Interested to hear more about this example. 

A list/set of values included in 1 match is not a problem (`matches[0]`) and fine (kind of like when a contract call returns multiple values as a result of the call), it's more the number of matches (`len(matches)`). i.e. a match can contain multiple values, but the number of matches should be 1.

nucypher/policy/conditions/offchain.py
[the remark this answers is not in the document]
derekpierre: Definitely doesn't have to be as part of this PR (this is the base foundation after all 😅 ), but something we should file an issue for and ensure that we think through whether it makes sense or not to do.

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,168 @@
...
+
+class JSONPathField(Field):
+    default_error_messages = {
+        "invalidType": "Expression of type '{type(value)}' is not valid for JSONPath",
+        "invalid": "'{value}' is not a valid JSONPath expression",
+    }
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if not isinstance(value, str):
+            raise self.make_error("invalidType", input=value)
```
cygnusv, line 28:
```suggestion
            raise self.make_error("invalidType", value=value)
```
Do you mean `value` here? otherwise, not sure how this format is supposed to work.

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,168 @@
...
+        return value
+
+
+class HTTPSField(Field):
+    default_error_messages = {
+        "invalid": "'{value}' is not a valid HTTPS endpoint",
+    }
+
+    def _deserialize(self, value, attr, data, **kwargs):
+        if not value.startswith("https://"):
```
cygnusv, line 42: Do we need a more complete validation here? E.g. `https://https://https://foofoofoo` satisfies current validation. It seems there's packages for this (e.g. https://yozachar.github.io/pyvalidators/stable/api/url/)

KPrasch: If more robust validation is needed, I'd prefer to make a simple regex here vs. introducing a new dependency.  Alternatively I might be able to use some standlib url parsing tools.  In any case, I don't think it's a serious concern since only valid URLs will be usable by requests.

nucypher/policy/conditions/offchain.py
KPrasch: Oh, I'll take a look at the docs.  Maybe I misunderstood `make_error`.

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,142 @@
...
+    class Schema(CamelCaseSchema):
+
+        name = fields.Str(required=False)
+        condition_type = fields.Str(
+            validate=validate.Equal(ConditionType.OFFCHAIN.value), required=True
+        )
+        headers = fields.Dict(required=False)
+        parameters = fields.Dict(required=False)
+        endpoint = fields.Str(required=True)
+        query = JSONPathField(required=True)
```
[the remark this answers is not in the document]
cygnusv: > (1) empty queries

Single, primitive values are valid JSON and therefore don't require a query path. E.g.: `42` is a valid JSON string.

> My general impression is that if the response body is empty then it's not a JSON API. 

Agreed, as `''` is not valid JSON.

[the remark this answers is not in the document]
derekpierre: > Single, primitive values are valid JSON and therefore don't require a query path. E.g.: 42 is a valid JSON string.

The crux of the blank query question was more along these lines i.e. whether the "query" field should be required or not. If it is a required field we'd need to allow a non-existent query (blank or otherwise) for an endpoint that simply returns a value eg. 42.

Seems like "query" should *not* be a required field.

nucypher/policy/conditions/offchain.py
```diff
@@ -0,0 +1,168 @@
...
+        except (requests.exceptions.RequestException, ValueError) as json_error:
+            self.logger.error(f"JSON parsing error occurred: {json_error}")
+            raise ConditionEvaluationFailed(
+                f"Failed to parse JSON response: {json_error}"
+            )
+        return data
+
+    def query_response(self, data: Any) -> Any:
+        try:
+            expression = parse(self.query)
```
cygnusv, line 146: Should this be done earlier? I mean, when constructing the JSONApiCondition object.

nucypher/policy/conditions/offchain.py
cygnusv: Oh that's a nice find!

nucypher/policy/conditions/offchain.py
KPrasch: It's already validated by the marshmallow schema to handle errors - does that cover your concern?  Parsing it into an object in the constructor implies reworking most of the tests and init-time validation.

nucypher/policy/conditions/offchain.py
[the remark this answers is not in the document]
KPrasch: > Interested to hear more about this example.

Here's an example:

```python
response = requests.get('https://api.coingecko.com/api/v3/simple/price?ids=ethereum&vs_currencies=usd')
# Check if the request was successful
if response.status_code == 200:
    data = response.json()
    # Parse the JSON response using jsonpath-ng
    jsonpath_expression = parse('$.ethereum.usd')
    match = jsonpath_expression.find(data)

...

>>> match
Out[14]: [DatumInContext(value=3335.29, path=Fields('usd'), context=DatumInContext(value={'usd': 3335.29}, path=Fields('ethereum'), context=DatumInContext(value={'ethereum': {'usd': 3335.29}}, path=Root(), context=None)))]
```

nucypher/policy/conditions/offchain.py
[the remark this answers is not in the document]
KPrasch: Opened https://github.com/nucypher/nucypher/issues/3517

nucypher/policy/conditions/offchain.py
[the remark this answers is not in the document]
derekpierre: In your example above `len(match)` is 1 so that's fine, there is only one match value to use i.e.  it is clear that `match[0].value` is what we use.

```python
In [41]: response = requests.get('https://api.coingecko.com/api/v3/simple/price?ids=ethereum&vs_currencies=usd')

In [42]: data = response.json()

In [43]: jsonpath_expr = parse('$.ethereum.usd')

In [44]: match = jsonpath_expr.find(data)

In [45]: len(match)
Out[45]: 1

In [46]: match[0].value
Out[46]: 3293.78
```

A match will have a value, and that value can be a number, list, dict etc. , and that's fine as long as the number of matches based on the expression (i.e. `len(match)`) is 1. So for example, the following has a dict of values for a single match and is still ok:
```python
In [51]: response = requests.get("https://api.weather.gov/gridpoints/LWX/97,71/forecast")

In [52]: data = response.json()

In [53]: jsonpath_expr = parse("$.properties.periods[?(@.name=='Tonight')]")

In [54]: match = jsonpath_expr.find(data)

In [55]: len(match)
Out[55]: 1

In [57]: match[0].value
Out[57]:
{'number': 2,
 'name': 'Tonight',
 'startTime': '2024-07-03T18:00:00-04:00',
 'endTime': '2024-07-04T06:00:00-04:00',
 'isDaytime': False,
 'temperature': 73,
 'temperatureUnit': 'F',
 'temperatureTrend': '',
 'probabilityOfPrecipitation': {'unitCode': 'wmoUnit:percent', 'value': None},
 'windSpeed': '6 to 9 mph',
 'windDirection': 'S',
 'icon': '/icons/land/night/sct?size=medium',
 'shortForecast': 'Partly Cloudy',
 'detailedForecast': 'Partly cloudy, with a low around 73. South wind 6 to 9 mph.'}
 ```

The case that I'm concerned about, unless I'm misunderstanding is the following, where the jsonpath query is not specific enough and returns multiple matches as a result of `find(...)`:
```python
In [60]: response = requests.get("https://api.weather.gov/gridpoints/LWX/97,71/forecast")

In [61]: data = response.json()

In [62]: jsonpath_expr = parse("$.properties.periods..temperature")

In [63]: match = jsonpath_expr.find(data)

In [64]: len(match)
Out[64]: 14
```

Basically, if the `jsonpath.find(...)` returns more than 1 match ( i.e.`len(match) > 1`) then which match value do we check? 

- Is simply checking `match[0].value` correct? Doesn't feel that way, because in the example above there are 13 other values. 

- Do we iterate through the match values and include them in a list? For example,
```python
In [65]: [m.value for m in match]
Out[65]: [88, 73, 94, 75, 95, 77, 93, 72, 91, 73, 93, 75, 94, 74]
```

- Should we just raise an exception if `len(match) > 1`? Basically, when `len(match) > 1`, the JSON path is perhaps deemed insufficiently specific?

In any case, we can discuss/consider this in a separate follow-up issue, outside of this PR.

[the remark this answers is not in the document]
derekpierre: Filed #3518 .

derekpierre (approved):
🎸  

Approving, but it is unclear if we still want to merge into 7.4.x or merge into 7.5.x (non existent branch). Related to https://github.com/nucypher/sprints/issues/9.

derekpierre (approved):
🎸

PR #3513, opened 2024-06-21
Accept/Process EIP-4361 passthrough context variable used for single sign-on capabilities

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [X] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Add capability for node to process SIWE passthrough messages/signatures that are provided via the `:userAddressExternalEIP4361` context variable.

Related to https://github.com/nucypher/taco-web/pull/529.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -21,7 +21,7 @@ CONTEXT_PREFIX = ":"
 CONTEXT_REGEX = re.compile(":[a-zA-Z_][a-zA-Z0-9_]*")
 
 USER_ADDRESS_SCHEMES = {
-    USER_ADDRESS_CONTEXT: None,  # TODO either EIP712 or EIP4361 for now, but should use the default that is eventually decided (likely EIP4361)
+    USER_ADDRESS_CONTEXT: None,  # TODO either EIP712 or EIP4361 for now, but should use the default that is eventually decided (likely EIP4361) - #tdec/178
     USER_ADDRESS_EIP712_CONTEXT: EvmAuth.AuthScheme.EIP712.value,
     USER_ADDRESS_EIP4361_CONTEXT: EvmAuth.AuthScheme.EIP4361.value,
     USER_ADDRESS_EIP4361_EXTERNAL_CONTEXT: EvmAuth.AuthScheme.EIP4361.value,
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -15,14 +15,16 @@ from nucypher.policy.conditions.exceptions import (
 USER_ADDRESS_CONTEXT = ":userAddress"
 USER_ADDRESS_EIP712_CONTEXT = ":userAddressEIP712"
 USER_ADDRESS_EIP4361_CONTEXT = ":userAddressEIP4361"
+USER_ADDRESS_EIP4361_EXTERNAL_CONTEXT = ":userAddressExternalEIP4361"
 
 CONTEXT_PREFIX = ":"
 CONTEXT_REGEX = re.compile(":[a-zA-Z_][a-zA-Z0-9_]*")
 
 USER_ADDRESS_SCHEMES = {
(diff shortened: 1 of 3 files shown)


nucypher/policy/conditions/context.py
```diff
@@ -15,14 +15,16 @@
 USER_ADDRESS_CONTEXT = ":userAddress"
 USER_ADDRESS_EIP712_CONTEXT = ":userAddressEIP712"
 USER_ADDRESS_EIP4361_CONTEXT = ":userAddressEIP4361"
+USER_ADDRESS_EIP4361_EXTERNAL_CONTEXT = ":userAddressExternalEIP4361"
```
KPrasch, line 18: Just to clarify - Is this intended for eventual deprecation in favor of `:userAddress`?

piotr-roslaniec (changes requested):

nucypher/policy/conditions/context.py
```diff
@@ -15,14 +15,16 @@
...
 
 CONTEXT_PREFIX = ":"
 CONTEXT_REGEX = re.compile(":[a-zA-Z_][a-zA-Z0-9_]*")
 
 USER_ADDRESS_SCHEMES = {
-    USER_ADDRESS_CONTEXT: None,  # any of the available auth types
+    USER_ADDRESS_CONTEXT: None,  # TODO either EIP712 or EIP4361 for now, but should use the default that is eventually decided (likely EIP4361)
```
cygnusv, line 24: We should track deciding on a default.

tests/unit/conditions/test_context.py
```diff
@@ -136,10 +137,12 @@ def test_user_address_context_invalid_typed_data(
             [
                 USER_ADDRESS_EIP712_CONTEXT,
                 USER_ADDRESS_EIP4361_CONTEXT,
+                USER_ADDRESS_EIP4361_EXTERNAL_CONTEXT,
             ],
             [
                 EvmAuth.AuthScheme.EIP4361.value,
                 EvmAuth.AuthScheme.EIP712.value,
+                EvmAuth.AuthScheme.EIP712.value,
```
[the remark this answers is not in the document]
cygnusv: I think the goal is to test for incorrect auth scheme, so `EIP712.value` would be an acceptable choice (currently the only other alternative). Perhaps the test variable `valid_user_address_auth_message` can be renamed to something else that makes clear that is an unexpected auth message type.

cygnusv (approved):

nucypher/policy/conditions/context.py
derekpierre: Assuming we end up deprecating EIP712, I think we will simply end up with 2 context variables:
- `:userAddress` used for taco-specific (non single-sign on) EIP4361
- `:userAddressExternalEIP4361` (or whatever we end up calling it, if a different name is desired) used for single sign-on with EIP4361.

For the encryptor it will be the separation between relying on TACo for EIP4361 vs relying on their app's existing EIP4361 capabilities.

The other 2 context variables (`:userAddressEIP712`, `:userAddressEIP4361`) can be deprecated; more like a remove since they were never publicly released.

tests/unit/conditions/test_context.py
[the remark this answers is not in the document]
derekpierre: The goal is to test using a valid auth message generated by the `valid_user_address_auth_message` fixture, but to use it with a mismatched/incompatible context variable. The message is indeed valid, but not correct for the corresponding context variable. For example, the `:userAddressExternalEIP4361` context key whose value contains an EIP712 auth message.

✔️ - updated the name of the test to make that clearer.

nucypher/policy/conditions/context.py
derekpierre: ✔️ - https://github.com/nucypher/tdec/issues/178

KPrasch (approved):

PR #3515, opened 2024-06-27
[WIP] Remove EIP712 Authentication

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

- Removal of the use of EIP712
- Removal of two no longer necessary context variables (`:userAddressEIP712`, `:userAddressEIP712`) - these were never released publicly so it's a removal not a deprecation.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

There will be a `taco-web` and node incompatibility. Suppose an app continued to use the old version of `taco-web` (which used EIP712 as the default), then an EIP712 auth signature would be used for `:userAddress`, and sent to nodes as proof. A node with these changes would reject it because the scheme is not supported.

If that app updated to use a newer version of `taco-web` that uses EIP4361 by default, then that problem would immediately go away.

Are we ok with forcing the `taco-web` upgrade otherwise apps that utilize `:userAddress` won't work, or should we still have nodes accept EIP712 used from older `taco-web` clients for some period of time (until a subsequent release eg. `7.5.0`). Nodes can detect older `taco-client` usages because the auth signatures would not include a "scheme" entry since "scheme" was never specified in older `taco-web` clients.

Thoughts?

### DIFF
--- a/newsfragments/3515.removal.rst
+++ b/newsfragments/3515.removal.rst
@@ -1,0 +1,3 @@
+Deprecate the use of EIP712 authentication for proving wallet address ownership
+for the `:userAddress` context variable for conditions. Instead EIP4361 authentication
+messages are expected to be used.
--- a/nucypher/policy/conditions/auth/evm.py
+++ b/nucypher/policy/conditions/auth/evm.py
@@ -2,14 +2,11 @@ from enum import Enum
 from typing import List
 
 import maya
-from eth_account.account import Account
-from eth_account.messages import HexBytes, encode_typed_data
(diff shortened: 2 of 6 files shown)


piotr-roslaniec (approved):
Pending decision on deprecation

derekpierre:
Full removal should not yet occur. Closing PR in favour of an incoming one.

PR #3524, opened 2024-07-24
Do not retry unrecoverable ferveo errors

KPrasch:
Required Reviews
2

**What this does**
Do not attempt to retry transcript generation if a cryptographic error bubbles up from ferveo itself.  This is a unrecoverable situation so there is no point in retrying.  Currently, nodes re-attempt transcript generation even in unrecoverable circumstances which temporarily blocks the node from participating in other DKG ceremonies.

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -633,11 +633,10 @@ class Operator(BaseActor):
                 ritual_id=ritual.id,
             )
         except Exception as e:
-            # TODO: Handle this better #3096
             self.log.critical(
                 f"Failed to generate a transcript for ritual #{ritual.id}: {str(e)}"
             )
-            raise e
+            return
 
         # publish the transcript and store the receipt
         self.dkg_storage.store_validators(ritual_id=ritual.id, validators=validators)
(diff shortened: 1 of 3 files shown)


nucypher/blockchain/eth/actors.py
```diff
@@ -623,11 +623,10 @@ def perform_round_1(
                 ritual_id=ritual.id,
             )
         except Exception as e:
-            # TODO: Handle this better #3096
             self.log.critical(
                 f"Failed to generate a transcript for ritual #{ritual.id}: {str(e)}"
             )
-            raise e
+            return
```
KPrasch, line 629: Note for reviewers - It may not be obvious what's happening here.  If this line raises, it ultimately crashes the async looping call that dispatches this task without removing it from the queue which results in a re-attempt on the next iteration. 

By returning here instead, we allow the task to complete without errors which removes this task from the queue.

derekpierre: > Question: is it better to log a full stack trace here for log visibility (same below)?

Given that it is critical - we should log the stack trace. Same for below. wdyt?

derekpierre: Side note: could be good to add a simple unit/integration test that ensures `perform_round_1`/`perform_round_2` return safely, without raising, for a cryptographic exception generated in the try-block.

newsfragments/3524.dev.rst
derekpierre, line 1: Should we regard this as a bugfix instead?

manumonti (approved):
LGTM! 🙏

PR #3538, opened 2024-08-12
Key Recovery and Auditing Tools

### LINKED ISSUE #3536: CLI command to output Ferveo keys
After v7.4.0, we need a command/option that outputs the Ferveo keys without starting up the node and connecting to P2P and contracts. This can be necessary for operators that lose access to their original keystore and have to manually set Ferveo keys in the contract. Alternatively, instead of following a manual approach, the node can start up and set the new keys to the contract.
Regardless whether it's manual or not, a penalty may be considered, since resetting Ferveo keys, effectively means that previous rituals that included a given node won't be serviced by it anymore.

derekpierre:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
- Introduces new CLI commands:
  - `nucypher ursula public-keys [--from-mnemonic|--keystore-fiepath <path>]`
    - View local public keys from keystores or mnemonic phrases
  - `nucypher ursula audit [--view-mnemonic]`
    - Verify that a keystore password can decrypt local keys
    - Verify that a mnemonic phrase matches a local keystore file
    - View mnemonic phrase
- Expands `nucypher ursula recover [--keystore-filepath]`
  - automatically updates pre-existing ursula configuration file with recovered key filepath from mnemonic or keystore
- Expands `nucypher ursula init ... --with-mnemonic`
  - for node host relocation: creates a new configuration importing keys from the mnemonic

**Issues fixed/closed:**
- Fixes #3536 
- Closes https://github.com/nucypher/sprints/issues/47

**Why it's needed:**
- Assist in mitigation of key loss or recovery scenarios.

**Notes for reviewers:**

This branch has been published to a pre-release on docker (`nucypher/nucypher:recovery`)

##### CLI Usage Examples:

Recover using a mnemonic (with existing config):
```
nucypher ursula recover
```

Recover/Relocate by creating a new configuration and mnemonic:
```
nucypher ursula init ... --with-mnemonic
```

View public keys for a given mnemonic:     
```
nucypher ursula public-keys --from-mnemonic
```

Audit password and mnemonic:
```
nucypher ursula audit
```
(uses default configuration file to obtain keystore path)

```
nucypher ursula audit --config-file <>
```
(uses specified configuration file to obtain keystore path)

```
nucypher ursula audit --keystore-filepath <>
```
(uses specified keystore file to)

View mnemonic:
```
nucypher ursula audit ... --view-mnemonic
```

##### Docker Usage Examples:

Recover using a mnemonic (with existing config):
```
docker run -it -v ~/.local/share/nucypher:/root/.local/share/nucypher:rw -v ~/.ethereum/:/root/.ethereum:ro nucypher/nucypher:recovery nucypher ursula recover
```

Recover/Relocate by creating a new configuration and mnemonic:
```
docker run -it -v ~/.local/share/nucypher:/root/.local/share/nucypher:rw -v ~/.ethereum/:/root/.ethereum:ro nucypher/nucypher:recovery nucypher ursula init ... --with-mnemonic
```

View public keys for a given mnemonic:     
```
docker run -it nucypher/nucypher:recovery nucypher ursula public-keys --from-mnemonic
```

Audit password and mnemonic:
```
docker run -it -v ~/.local/share/nucypher:/root/.local/share/nucypher:rw -v ~/.ethereum/:/root/.ethereum:ro nucypher/nucypher:recovery nucypher ursula audit
```

View mnemonic:
```
docker run -it -v ~/.local/share/nucypher:/root/.local/share/nucypher:rw -v ~/.ethereum/:/root/.ethereum:ro nucypher/nucypher:recovery nucypher ursula audit --view-mnemonic
```

### DIFF
--- a/nucypher/cli/actions/configure.py
+++ b/nucypher/cli/actions/configure.py
@@ -196,4 +196,5 @@ def update_config_keystore_path(keystore_path: Path, config_file: Path = None) -
         ursula_config = json.load(f)
         ursula_config["keystore_path"] = keystore_path
         f.seek(0)
-        json.dump(ursula_config, f, indent=4)
+        json.dump(ursula_config, f, indent=2)
+        f.truncate()  # rest of the file truncated
--- a/nucypher/cli/actions/configure.py
+++ b/nucypher/cli/actions/configure.py
@@ -187,3 +187,13 @@ def perform_startup_ip_check(emitter: StdoutEmitter, ursula: Ursula, force: bool
         raise click.Abort()
     else:
         emitter.message('✓ External IP matches configuration', 'green')
+
+
+def update_config_keystore_path(keystore_path: Path, config_file: Path = None) -> None:
+    """Update the ursula.json configuration file to use the provided keystore path."""
+    keystore_path = str(keystore_path.resolve())
+    with open(config_file, "r+") as f:
+        ursula_config = json.load(f)
+        ursula_config["keystore_path"] = keystore_path
+        f.seek(0)
+        json.dump(ursula_config, f, indent=4)
--- a/nucypher/cli/commands/ursula.py
+++ b/nucypher/cli/commands/ursula.py
@@ -164,7 +164,7 @@ class UrsulaConfigOptions:
                 # TODO: Exit codes (not only for this, but for other exceptions)
                 return click.get_current_context().exit(1)
 
-    def generate_config(self, emitter, config_root, force, key_material):
+    def generate_config(self, emitter, config_root, force, key_material, with_mnemonic):
 
         if self.dev:
             raise RuntimeError(
@@ -193,6 +193,7 @@ class UrsulaConfigOptions:
         return UrsulaConfiguration.generate(
             password=get_nucypher_password(emitter=emitter, confirm=True),
             key_material=bytes.fromhex(key_material) if key_material else None,
+            with_mnemonic=with_mnemonic,
             config_root=config_root,
             rest_host=self.rest_host,
             rest_port=self.rest_port,
@@ -323,7 +324,14 @@ def ursula():
 @option_config_root
 @group_general_config
 @option_key_material
-def init(general_config, config_options, force, config_root, key_material):
+@click.option(
+    "--with-mnemonic",
+    help="Initialize with a mnemonic phrase instead of generating a new keypair from scratch",
+    is_flag=True,
+)
+def init(
+    general_config, config_options, force, config_root, key_material, with_mnemonic
+):
     """Create a new Ursula node configuration."""
     emitter = setup_emitter(general_config, config_options.operator_address)
     _pre_launch_warnings(emitter, dev=None, force=force)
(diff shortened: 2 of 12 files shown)


nucypher/cli/actions/configure.py
```diff
@@ -187,3 +187,13 @@ def perform_startup_ip_check(emitter: StdoutEmitter, ursula: Ursula, force: bool
...
+
+
+def update_config_keystore_path(keystore_path: Path, config_file: Path = None) -> None:
+    """Update the ursula.json configuration file to use the provided keystore path."""
+    keystore_path = str(keystore_path.resolve())
+    with open(config_file, "r+") as f:
+        ursula_config = json.load(f)
+        ursula_config["keystore_path"] = keystore_path
+        f.seek(0)
+        json.dump(ursula_config, f, indent=4)
```
derekpierre, line 199: This didn't work the way we expected, which is why I had to change to it to two separate open blocks. See https://stackoverflow.com/questions/42297465/does-json-dump-append-to-file

derekpierre: Resolved using `truncate()` after writing the data.

derekpierre:
Updated the `nucypher/nucypher:recovery` docker image to be based on commit [b2772be](https://github.com/nucypher/nucypher/pull/3538/commits/b2772bebaf03934cab28d7c80e3767ef23b38f40).

nucypher/config/base.py
```diff
@@ -824,13 +836,20 @@ def write_keystore(
         password: str,
         key_material: Optional[bytes] = None,
         interactive: bool = True,
+        with_mnemonic: bool = False,
```
derekpierre, line 839: Is there a conflict between `with_mnemonic` and  `key_material` i.e. if `key_material` is not `None` then `with_mnemoic` should be `False`, and the opposite i.e. if `with_mnemonic` is True then `key_material` should be `None`. Do we check that this is the case for the parameters somewhere in the call stack? I may have missed it.

nucypher/cli/commands/ursula.py
```diff
@@ -350,15 +358,24 @@ def init(general_config, config_options, force, config_root, key_material):
...
-        "NOTE: Initializing a new Ursula node configuration is a one-time operation\n"
-        "for the lifetime of your node.  This is a two-step process:\n\n"
-        "1. Creating a password to encrypt your operator keys\n"
-        "2. Securing a taco node seed phase\n\n"
-        "Please follow the prompts.",
-        color="cyan",
-    )
+    if with_mnemonic:
+        emitter.echo(
+            "Hello Operator, welcome back :-) \n\n"
```
derekpierre, line 363: Is it possible for the init to be called with `--with-mnemonic` the first time ever or is there something preventing that? In which case it may not be "welcome back" necessarily.

KPrasch: I'm working under the assumption that they are in a possession of a mnemonic that they received on first `init`.  I suppose it's possible for someone to generate their own mnemonic and bring it along.  What do you think?

nucypher/config/base.py
KPrasch: Yes - it's a good point.  I'll make an update.

nucypher/cli/commands/ursula.py
derekpierre: Yeah. I'm wondering if `--with-mnemonic` my be used similar to how `--key-material` would be if you have your own source of randomness.

KPrasch: ✔️

nucypher/config/base.py
KPrasch: ✔️

KPrasch (approved):

derekpierre:
Updated `nucypher/nucypher:recovery` docker image to be based on commit https://github.com/nucypher/nucypher/pull/3538/commits/cf1d9e06c5c299b13a64b101c9701f8ec6bc3213

nucypher/cli/commands/ursula.py
```diff
@@ -402,6 +538,50 @@ def destroy(general_config, config_options, config_file, force):
...
+@click.option(
+    "--from-mnemonic",
+    help="View TACo public keys from mnemonic seed words",
+    is_flag=True,
+)
+def public_keys(config_file, keystore_filepath, from_mnemonic):
+    """Display the public keys of a keystore."""
+    emitter = StdoutEmitter()
+
+    if sum(1 for i in (keystore_filepath, config_file, from_mnemonic) if i) > 1:
```
cygnusv, line 557:
```suggestion
    if len(i for i in (keystore_filepath, config_file, from_mnemonic) if i is not None) != 1:
```

cygnusv (approved):

nucypher/cli/commands/ursula.py
derekpierre: - We have to be careful since `from_mnemonic` is a boolean (`False` by default), so we can't just check `is not None`.
- It is valid usage if none of the 3 arguments are set (with `from_mnemonic` being `False`), which is why `> 1` was used. In this case we use the default config file location which has a keystone path value in it.

@cygnusv , do you have a strong preference here between counting arguments and checking the length of the resulting list? If you indeed have a strong opinion, we could do the following
```python
len([i for i in (keystore_filepath, config_file, from_mnemonic) if bool(i)]) > 1
```

(open to any alternatives as well)

theref (approved):

nucypher/cli/commands/ursula.py
cygnusv: > * it is valid usage if none of the 3 arguments are set (with `from_mnemonic` being `False`), which is why `> 1` was used. In this case we use the default config file location which has a keystore path value in it.

In that case, the wording of the error message above should be changed to `"At most one of --keystore-filepath, --config-file, or --from-mnemonic must be specified"`.

derekpierre: 👍 - addressed in https://github.com/nucypher/nucypher/pull/3545.

PR #3553, opened 2024-09-13
Condition result that is a string needs to be quoted

derekpierre:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

A condition execution result that is an unquoted string and will be used for return value test comparison will raise an exception. Any string that is passed to `ast.literal_eval()` must be quoted.

This affects `JsonApiCondition` in particular because JSON strings are not usually quoted. For example, see test in PR here - https://github.com/nucypher/nucypher/pull/3553/files#diff-f13d425b7bdc36d0f95ef2a4f4d356124fb4af34f11f0194302522cb881c25e3R126-R153. 

Since the book title is what is retrieved using the jsonpath query, which is "Test Title" or "Title", if you pass these to ast.literal_eval(...) it will raise an exception:
```python
In [2]: ast.literal_eval(str("Test Title"))

  File ~/.local/share/virtualenvs/nucypher-z-zdnWw9/lib/python3.10/site-packages/IPython/core/interactiveshell.py:3550 in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)

  Cell In[2], line 1
    ast.literal_eval(str("Test Title"))

  File /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py:64 in literal_eval
    node_or_string = parse(node_or_string.lstrip(" \t"), mode='eval')

  File /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py:50 in parse
    return compile(source, filename, mode, flags,

  File <unknown>:1
    Test Title
         ^
SyntaxError: invalid syntax


In [3]: ast.literal_eval(str("Title"))
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 ast.literal_eval(str("Title"))

File /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py:110, in literal_eval(node_or_string)
    108                 return left - right
    109     return _convert_signed_num(node)
--> 110 return _convert(node_or_string)

File /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py:109, in literal_eval.<locals>._convert(node)
    107         else:
    108             return left - right
--> 109 return _convert_signed_num(node)

File /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py:83, in literal_eval.<locals>._convert_signed_num(node)
     81     else:
     82         return - operand
---> 83 return _convert_num(node)

File /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py:74, in literal_eval.<locals>._convert_num(node)
     72 def _convert_num(node):
     73     if not isinstance(node, Constant) or type(node.value) not in (int, float, complex):
---> 74         _raise_malformed_node(node)
     75     return node.value

File /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/ast.py:71, in literal_eval.<locals>._raise_malformed_node(node)
     69 if lno := getattr(node, 'lineno', None):
     70     msg += f' on line {lno}'
---> 71 raise ValueError(msg + f': {node!r}')

ValueError: malformed node or string on line 1: <ast.Name object at 0x104ea6e30>
```

It's the same reason we already force users to quote strings used for `value` in ReturnValueTest, but we have to do the same for execution results such as these.

Instead "Test Title" needs to be made into "'Test Title'".
```python
In [4]: ast.literal_eval(str("'Test Title'"))
Out[4]: 'Test Title'

In [5]: ast.literal_eval(str("'Title'"))
Out[5]: 'Title'
```

We never ran into this problem before because on-chain conditions do not return strings, but JSON endpoints can.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

Used a dev newsfragment because it fixes a bug for Json API conditions, which hasn't yet been publicly released - so not really a bug fix from a release note point of view.

### DIFF
--- a/nucypher/policy/conditions/offchain.py
+++ b/nucypher/policy/conditions/offchain.py
@@ -158,7 +158,12 @@ class JsonApiCondition(AccessControlCondition):
                 or (result.startswith('"') and result.endswith('"'))
             ):
                 # quote the string
-                result = f"'{result}'"
+                if "'" in result:
+                    # use double quotes
+                    result = f'"{result}"'
+                else:
+                    # use single quotes
+                    result = f"'{result}'"
 
         return result
 
--- a/nucypher/policy/conditions/offchain.py
+++ b/nucypher/policy/conditions/offchain.py
@@ -157,4 +157,9 @@ class JsonApiCondition(AccessControlCondition):
         response = self.fetch()
         data = self.deserialize_response(response)
         result = self.query_response(data)
-        return self.return_value_test.eval(result), result
+
+        resolved_return_value_test = self.return_value_test.with_resolved_context(
+            **context
+        )
+        eval_result = resolved_return_value_test.eval(result)  # test
+        return eval_result, result
--- a/nucypher/policy/conditions/offchain.py
+++ b/nucypher/policy/conditions/offchain.py
@@ -151,19 +151,16 @@ class JsonApiCondition(AccessControlCondition):
     @staticmethod
     def _process_result_for_eval(result: Any):
         # strings that are not already quoted will cause a problem for literal_eval
-        if isinstance(result, str):
-            # check if already quoted; if not, quote it
-            if not (
-                (result.startswith("'") and result.endswith("'"))
-                or (result.startswith('"') and result.endswith('"'))
-            ):
-                # quote the string
-                if "'" in result:
-                    # use double quotes
-                    result = f'"{result}"'
-                else:
-                    # use single quotes
-                    result = f"'{result}'"
+        if not isinstance(result, str):
+            return result
+
+        # check if already quoted; if not, quote it
+        if not (
+            (result.startswith("'") and result.endswith("'"))
+            or (result.startswith('"') and result.endswith('"'))
+        ):
+            quote_type_to_use = '"' if "'" in result else "'"
+            result = f"{quote_type_to_use}{result}{quote_type_to_use}"
 
         return result
 
--- a/nucypher/policy/conditions/offchain.py
+++ b/nucypher/policy/conditions/offchain.py
@@ -148,6 +148,20 @@ class JsonApiCondition(AccessControlCondition):
 
         return result
 
+    @staticmethod
+    def _process_result_for_eval(result: Any):
+        # strings that are not already quoted will cause a problem for literal_eval
+        if isinstance(result, str):
+            # check if already quoted; if not, quote it
+            if not (
+                (result.startswith("'") and result.endswith("'"))
+                or (result.startswith('"') and result.endswith('"'))
+            ):
+                # quote the string
+                result = f"'{result}'"
+
+        return result
+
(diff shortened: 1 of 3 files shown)


nucypher/policy/conditions/offchain.py
```diff
@@ -148,6 +148,20 @@ def query_response(self, data: Any) -> Any:
...
 
+    @staticmethod
+    def _process_result_for_eval(result: Any):
+        # strings that are not already quoted will cause a problem for literal_eval
+        if isinstance(result, str):
+            # check if already quoted; if not, quote it
+            if not (
+                (result.startswith("'") and result.endswith("'"))
+                or (result.startswith('"') and result.endswith('"'))
+            ):
```
cygnusv, line 159: There's some edge cases here. e.g., `'` or `"` would both be valid quoted strings. 

Regarding edge cases I don't know how we should treat examples like `ab"c` or, being more realistic, `O'Sullivan`. The latter, for example, would be quoted as `'O'Sullivan'`, which I figure will break literal eval, unless it is double quoted, instead of single quoted.

Perhaps, the best option would be to disallow quotes to some degree, but unsure how.

derekpierre: ✔️ Nice catch! Made an update to properly handle quotations, and improved the testing. That being said, if there are combinations of both types on quotations in the string, I'm not sure how to handle that - that type of situation feels like an extreme edge-case which I'm ok with saying we don't support.

nucypher/policy/conditions/offchain.py
```diff
@@ -148,6 +148,25 @@ def query_response(self, data: Any) -> Any:
...
         return result
 
+    @staticmethod
+    def _process_result_for_eval(result: Any):
+        # strings that are not already quoted will cause a problem for literal_eval
+        if isinstance(result, str):
+            # check if already quoted; if not, quote it
+            if not (
+                (result.startswith("'") and result.endswith("'"))
+                or (result.startswith('"') and result.endswith('"'))
```
KPrasch, line 158: `startswith` and `endswith` can accept tuples which can simplify this slightly.  Also, if you like to reduce further you can use ternary operation to determine the quote type.

```python
    if isinstance(result, str) and not (result.startswith(("'", '"')) and result.endswith(("'", '"'))):
        quote = '"' if "'" in result else "'"
```

KPrasch (approved):

nucypher/policy/conditions/offchain.py
derekpierre: ✔️ - https://github.com/nucypher/nucypher/pull/3553/commits/963c1312dd1de15da28d54e5434d59e83f7b8985

theref (approved):

Issue #3555, opened 2024-09-18
Additional SequentialConditions considerations


derekpierre:
Additional considerations based on initial SequentialConditiion design work done in #3500:

```python
CONDITION_VARIABLE = {
    "varName": STR,
    "condition": {
        CONDITION
    }
}

SEQUENTIAL_CONDITION = {
    "name": ...  (Optional),
    "conditionType": "sequential",
    "conditionVariables": [CONDITION_VARIABLE*]
}
```

- For a `ConditionVariable`, `CONDITION` has an optional `name` value that was intended for labelling. Perhaps the already existing `name` value can be used instead of `varName`, and then the `ConditionVariable` type will no longer be needed.

- `CONDITION`  has a required `returnValueTest` property. Therefore each condition in the sequential condition will require a check i.e. return value test. The need for check may apply in some cases, but in others, perhaps the result of a condition execution call does not need to be checked - i.e. we only need the result of the underlying call and don't need to check the value. Since returnValueTest is required devs could do a trivial check like != "" or != -1 or ... if they really don't care about the value of the result, but just need it for a subsequent condition.

Instead a SequentialCondition may use some combination of CALL(s) and CONDITION(s) where some items don't need to check the result while others do respectively. Alternatively, if we continue to just use CONDITIONS then perhaps a noop return value test could be good instead of always having the condition do some trivial check as part of the required return value test.

KPrasch:
Ah - beat me to it!  One more suggestion is to repurpose the existing `name` field in favor of `varName`.  Currently it only serves as an optional human-readable description but may as well give it some purpose.  This may also reduce the impact on taco-web.  Alternately if we decide to keep `varName` let's rename `name` to description to disambiguate.

KPrasch:
Somehow that bullet point was in my blind spot.  Thanks!

derekpierre:
> Somehow that bullet point was in my blind spot. Thanks!

All good. I like your wording.

derekpierre:
> Alternately if we decide to keep varName let's rename name -> description to disambiguate.

We do have to be cautious about property renames of the existing lingo. If we want to do a rename, we'll likely have to version the lingo since existing conditions that actually used `name` would still need to work. In any case, we can cross that bridge if we get there.

PR #3556, opened 2024-09-18
Clean up/Simplify condition validation

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Simplify condition validation via inheritance.

To validate initialization values provided during Condition initialization we use a combination of variable checks in the python constructor `__init__()`, and condition schema validation via marshmallow. 

These checks are typically done using separate functions/methods. The reason being that there is a Python API for conditions (i.e. condition objects instantiated directly via constructor), and a condition lingo received from web apps (i.e. objects instantiated from json strings) which go through the marshmallow schema mechanism.

Since this validation is duplicated, all checks performed by schemas needed to be done via python objects - but that wasn't fully the case. Since there are separate checks via Python API and others done via Marshmallow schema, this can be confusing and will eventually become out of sync (it is already 😬 ).

The marshmallow schema mechanism is already feature-rich, so let's have all validation be performed by the schema for us.

Therefore:
- The base `AccessControlCondition` now performs a schema validation in its constructor via `_validate()` call. 
- Because conditions are validated in the base class constructor and via their schemas, condition sub-classes should set all member variables before calling super().__init__() since the schema values used are obtained from `.to_dict()` which creates a dict based on member variables for a condition class.
- All necessary validation should be performed via Schema classes i.e. `@validates` and `@validates_schema` decorators. 
    - This created a problem where underlying ExecutionCalls were instantiated in the constructor before the condition schema was validated. Therefore, execution calls should have their own schema which can be validated before instantiation by a condition. The condition schema extends the execution call schema so that fields don't have to be duplicated for condition and execution call, but the fields for underlying execution calls are implicitly used as part of the condition schema itself. (There is no change in overall condition schema, just a reorganization between execution call and condition. This also paves the way for execution calls to possibly be top-level/higher-order objects that can be used separately from conditions.

**Issues fixed/closed:**
> - Fixes #...

- Follow-up to #3500 
- Somewhat related to #3555 - "This also paves the way for execution calls to possibly be top-level/higher-order objects that can be used separately from conditions."

**Why it's needed:**

**Notes for reviewers:**

This is a tricky code re-org but happy to walk anyone through the change. Let me know.

### DIFF
--- a/nucypher/policy/conditions/base.py
+++ b/nucypher/policy/conditions/base.py
@@ -1,7 +1,7 @@
 import json
 from abc import ABC, abstractmethod
 from base64 import b64decode, b64encode
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, List, Optional, Tuple
 
 from marshmallow import Schema, ValidationError, fields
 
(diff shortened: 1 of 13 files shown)


manumonti (approved):
LGTM 👌

manumonti (approved):

KPrasch (approved):

PR #3557, opened 2024-09-30
Update ritual ID used in Lynx examples

manumonti:
**Type of PR:**
- [X] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
This change will update the ritual ID used in the examples. Note that thess examples are used on CI. This is needed because the Lynx ritual #0 has expired, so a new ritual has been initiated (id #`27`).

### DIFF
--- a/examples/testnet_compound_multichain_taco.py
+++ b/examples/testnet_compound_multichain_taco.py
@@ -39,7 +39,7 @@ coordinator_agent = CoordinatorAgent(
     blockchain_endpoint=polygon_endpoint,
     registry=registry,
 )
-ritual_id = 0  # got this from a side channel
+ritual_id = 27  # got this from a side channel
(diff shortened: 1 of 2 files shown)


manumonti:
Actually, I'm not sure if pointing to `main` is the right move. What do you think?

derekpierre:
> Actually, I'm not sure if pointing to `main` is the right move. What do you think?

Good intuition. This should be set to merge into `v7.5.x` instead of `main`.
Currently, we only merge versioned branches into `main` i.e. `vA.B.x`

theref:
looks like the newsfragment needs updating :)

theref (approved):

derekpierre (approved):
🎸

PR #3560, opened 2024-10-23
JSON API Authorized Requests (OAuth and JWT Support)

### LINKED ISSUE #3559: Potential OAuth Support
A typical OAuth workflow involves:
1. Authorization Request: the application/client requests authorization from the resource owner (user).
2. Authorization Grant: the user gives consent, and the authorization server (e.g. Google, Facebook) issues an authorization grant.
3. Access Token Request: the client exchanges the authorization grant for an access token with the authorization server
4. Resource Access: the client uses the access token to request data from the resource server

```
     +--------+                               +---------------+
     |        |--(1)- Authorization Request ->|   Resource    |
     |        |                               |     Owner     |
     |        |<-(2)-- Authorization Grant ---|               |
     |        |                               +---------------+
     |        |
     |        |                               +---------------+
     |        |--(3)-- Authorization Grant -->| Authorization |
     | Client |                               |     Server    |
     |        |<-(3)----- Access Token -------|               |
     |        |                               +---------------+
     |        |
     |        |                               +---------------+
     |        |--(4)----- Access Token ------>|    Resource   |
     |        |                               |     Server    |
     |        |<-(4)--- Protected Resource ---|               |
     +--------+                               +---------------+
```

TACo could function as part of a client system, using an already obtained access token to verify off-chain conditions. A possible use case for OAuth within TACo would involve scenarios where off-chain API endpoints, such as those used for `JsonApiCondition` , require user authentication via access tokens.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Allow "authorizationToken" to optionally be provided via the `JsonApiCondition`. The token should always be provided as a context variable since hardcoding it doesn't make much sense:
   - It would expose a token in plaintext
   - Tokens have expiry dates
   - Tokens are associated with specific users, so the use of the hardcoded token would be associated with a specific user and not the requester
   Instead, the token should be provided at decryption time. How the token is obtained (OAuth dance or JWT) is up to the application and not `taco-web`.
- The value for the "authorizationToken" context variable can either be OAuth or JWT tokens.
- Example lingo:
```json
{
  "conditionType":"json-api",
  "endpoint":"https://api.example.com/data",
  "parameters":{
    "ids":"ethereum",
    "vs_currencies":"usd"
  },
  "authorizationToken":":authToken",   <---- addition
  "query":"$.store.book[0].price",
  "returnValueTest":{
    "comparator":"==",
    "value":1.0
  }
}
```
- The context variable used for `authorizationToken` value can be anything and doesn't have to be named ":authToken". This way, adopters can use whatever context variable they want, including across multiple conditions in something like a `SequentialCondition` instead of us being prescriptive about what the name should be. Our only stipulation is that the value should be a context variable.
- The value for the context variable should be provided at decryption time by the requester. Once provided the value is used within the `Authorization` header for the HTTPS GET request i.e.
`Authorization: Bearer <authToken>`
- Expands the use of context variables since they could be included within the "endpoint", "parameters", "query" as substrings and not just an individual value.

**Issues fixed/closed:**
- Closes #3559 
- Related to https://github.com/nucypher/sprints/issues/85 - the node side of things.
- Related to #3561 

**Why it's needed:**

**Notes for reviewers:**
I performed local testing of the `JsonApiCondition` that used the GitHub API and my GitHub fine-grained access token. I did not commit those tests because my access token would be exposed.

### DIFF
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -93,6 +93,13 @@ def is_context_variable(variable) -> bool:
     return isinstance(variable, str) and CONTEXT_REGEX.fullmatch(variable)
 
 
+def string_contains_context_variable(variable: str) -> bool:
+    matches = re.findall(CONTEXT_REGEX, variable)
+    if not matches:
+        return False
+    return True
+
+
 def get_context_value(context_variable: str, **context) -> Any:
     try:
         # DIRECTIVES are special context vars that will pre-processed by ursula
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -95,9 +95,7 @@ def is_context_variable(variable) -> bool:
 
 def string_contains_context_variable(variable: str) -> bool:
     matches = re.findall(CONTEXT_REGEX, variable)
-    if not matches:
-        return False
-    return True
+    return bool(matches)
 
 
 def get_context_value(context_variable: str, **context) -> Any:
@@ -124,10 +122,9 @@ def resolve_any_context_variables(
     if isinstance(param, list):
         return [resolve_any_context_variables(item, **context) for item in param]
     elif isinstance(param, dict):
-        result = {}
-        for k, v in param.items():
-            result[k] = resolve_any_context_variables(v, **context)
-        return result
+        return {
+            k: resolve_any_context_variables(v, **context) for k, v in param.items()
+        }
     elif isinstance(param, str):
         # either it is a context variable OR contains a context variable within it
         # TODO separating the two cases for now out of concern of regex searching
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -116,9 +116,9 @@ def get_context_value(context_variable: str, **context) -> Any:
     return value
 
 
-def _resolve_context_variable(param: Union[Any, List[Any]], **context):
+def resolve_context_variable(param: Union[Any, List[Any]], **context):
     if isinstance(param, list):
-        return [_resolve_context_variable(item, **context) for item in param]
+        return [resolve_context_variable(item, **context) for item in param]
     elif is_context_variable(param):
         return get_context_value(context_variable=param, **context)
     else:
@@ -130,6 +130,6 @@ def resolve_parameter_context_variables(parameters: Optional[List[Any]], **conte
         processed_parameters = []  # produce empty list
     else:
         processed_parameters = [
-            _resolve_context_variable(param, **context) for param in parameters
+            resolve_context_variable(param, **context) for param in parameters
         ]
     return processed_parameters
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -1,6 +1,6 @@
 import re
 from functools import partial
-from typing import Any, List, Optional, Union
+from typing import Any, Dict, List, Union
 
 from eth_typing import ChecksumAddress
 from eth_utils import to_checksum_address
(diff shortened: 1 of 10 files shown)


theref (approved):

manumonti (approved):
LGTM!

nucypher/policy/conditions/context.py
```diff
@@ -90,12 +90,14 @@ def _resolve_user_address(user_address_context_variable, **context) -> ChecksumA
 
 
 def is_context_variable(variable) -> bool:
-    if isinstance(variable, str) and variable.startswith(CONTEXT_PREFIX):
-        if CONTEXT_REGEX.fullmatch(variable):
-            return True
-        else:
-            raise ValueError(f"Context variable name '{variable}' is not valid.")
-    return False
+    return isinstance(variable, str) and CONTEXT_REGEX.fullmatch(variable)
```
KPrasch, line 93: very clean 👍🏻

nucypher/policy/conditions/context.py
```diff
@@ -116,20 +118,31 @@ def get_context_value(context_variable: str, **context) -> Any:
...
     if isinstance(param, list):
-        return [_resolve_context_variable(item, **context) for item in param]
-    elif is_context_variable(param):
-        return get_context_value(context_variable=param, **context)
+        return [resolve_any_context_variables(item, **context) for item in param]
+    elif isinstance(param, dict):
+        result = {}
+        for k, v in param.items():
+            result[k] = resolve_any_context_variables(v, **context)
+        return result
```
KPrasch, line 130: It's only stylistic but while we are code golfing here :-)

```suggestion
        return {k, resolve_any_context_variables(v, **context) for k, v in param.items()}
```

nucypher/policy/conditions/context.py
```diff
@@ -116,20 +118,31 @@ def get_context_value(context_variable: str, **context) -> Any:
...
+            return get_context_value(context_variable=param, **context)
+        else:
+            matches = re.findall(CONTEXT_REGEX, param)
+            for context_var in matches:
+                # checking out of concern for faulty regex search within string
+                if context_var in context:
+                    resolved_var = get_context_value(
+                        context_variable=context_var, **context
+                    )
+                    param = param.replace(context_var, str(resolved_var))
```
KPrasch, line 145: This is very defensive - don't fully trust your regex skill? :-)

nucypher/policy/conditions/context.py
```diff
@@ -90,12 +90,14 @@ def _resolve_user_address(user_address_context_variable, **context) -> ChecksumA
...
-            raise ValueError(f"Context variable name '{variable}' is not valid.")
-    return False
+    return isinstance(variable, str) and CONTEXT_REGEX.fullmatch(variable)
+
+
+def string_contains_context_variable(variable: str) -> bool:
+    matches = re.findall(CONTEXT_REGEX, variable)
+    if not matches:
+        return False
+    return True
```
cygnusv, line 100:
```suggestion
    return bool(matches)
```

nucypher/policy/conditions/offchain.py
```diff
@@ -42,56 +48,80 @@ class Schema(ExecutionCall.Schema):
...
+            )
+            headers = {"Authorization": f"Bearer {resolved_authorization_token}"}
+
         try:
             response = requests.get(
-                self.endpoint, params=self.parameters, timeout=self.timeout
+                resolved_endpoint,
+                params=resolved_parameters,
+                timeout=self.timeout,
+                headers=headers,
```
KPrasch, line 105: looks good 👍🏻

KPrasch (approved):

nucypher/policy/conditions/context.py
cygnusv, line 130:
```suggestion
        return {k: resolve_any_context_variables(v, **context) for k, v in param.items()}
```

cygnusv (approved):

nucypher/policy/conditions/context.py
derekpierre: Haha that and I'm worried about legit `:<>` expressions within a string field.

Not applicable because I think it is not allowed for JsonPath, but a ridiculous example being some expression syntax analogous to the JsonPath query using a legit ":<>" combination.

nucypher/policy/conditions/context.py
derekpierre: ✔️

nucypher/policy/conditions/context.py
derekpierre: ✔️

Issue #3395, opened 2024-01-17
Find a better order for TDec requests, perhaps based on latency data obtained from discovery loop


cygnusv:
If we can keep track of the recent average latency with peers, we can use that to prioritize faster peers when making TDec requests in the `ThresholdDecryptionClient`

PR #3562, opened 2024-10-31
Learner node latency tracking

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Updates a `Learner` to track average node connection latency and then uses it to order threshold decryption requests. Only the `/node_metadata` endpoint for all nodes is used for collecting stats. This is called whenever the Learner learns about other nodes from a node (teacher).

The list of ursulas representing the ritual cohort is ordered from smallest average latency to largest, and that list is used for sending decryption requests via the `ThresholdDecryptionRequestFactory`.

Porter uses the same code and will directly benefit from this change. Porter can use this code before an official `nucypher/nucypher` by simply using a github dependency. Consequently, we can release an updated Porter as soon as we like.

**Issues fixed/closed:**
Closes https://github.com/nucypher/nucypher/issues/3395.

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -1,9 +1,8 @@
 import time
-from collections import defaultdict, deque
+from collections import deque
 from contextlib import suppress
 from queue import Queue
-from threading import Lock
-from typing import List, Optional, Set, Tuple
+from typing import Optional, Set, Tuple
 
 import maya
 import requests
@@ -41,6 +40,7 @@ from nucypher.crypto.signing import InvalidSignature, SignatureStamp
 from nucypher.network.exceptions import NodeSeemsToBeDown
 from nucypher.network.middleware import RestMiddleware
 from nucypher.network.protocols import InterfaceInfo, SuspiciousActivity
+from nucypher.utilities.latency import NodeLatencyStatsCollector
 from nucypher.utilities.logging import Logger
 
 TEACHER_NODES = {
@@ -52,72 +52,6 @@ TEACHER_NODES = {
     domains.LYNX: ("https://lynx.nucypher.network:9151",),
     domains.TAPIR: ("https://tapir.nucypher.network:9151",),
 }
-
-
-class NodeLatencyStatsCollector:
-    TOTAL_TIME = "total_time"
-    COUNT = "count"
-    MAX_LATENCY = 2**32 - 1  # just need a large number
-
-    class NodeExecutionContextManager:
-        def __init__(
-            self,
-            stats_collector: "NodeLatencyStatsCollector",
-            staker_address: str,
-        ):
-            self._stats_collector = stats_collector
-            self.staker_address = staker_address
-
-        def __enter__(self):
-            self.start_time = time.perf_counter()
-            return self
-
-        def __exit__(self, exc_type, exc_val, exc_tb):
-            if exc_type:
-                # exception occurred - reset stats since connectivity was compromised
-                self._stats_collector.reset_stats(self.staker_address)
-            else:
-                # no exception
-                end_time = time.perf_counter()
-                execution_time = end_time - self.start_time
-                self._stats_collector.update_stats(self.staker_address, execution_time)
-
-    def __init__(self):
-        # staker_address -> { "total_time": <float>, "count": <integer> }
-        self._node_stats = defaultdict(lambda: {self.TOTAL_TIME: 0.0, self.COUNT: 0})
-        self._lock = Lock()
-
-    def update_stats(self, staking_address: str, latest_time_taken: float):
-        with self._lock:
-            self._node_stats[staking_address][self.TOTAL_TIME] += latest_time_taken
-            self._node_stats[staking_address][self.COUNT] += 1
-
-    def reset_stats(self, staking_address: str):
-        with self._lock:
-            self._node_stats[staking_address][self.TOTAL_TIME] = 0
-            self._node_stats[staking_address][self.COUNT] = 0
-
-    def get_latency_tracker(self, staker_address: str) -> NodeExecutionContextManager:
-        return self.NodeExecutionContextManager(
-            stats_collector=self, staker_address=staker_address
-        )
-
-    def get_average_latency_time(self, staking_address: str):
-        with self._lock:
-            count = self._node_stats[staking_address][self.COUNT]
-            # just need a large number > 0
-            return (
-                self.MAX_LATENCY
-                if count == 0
-                else self._node_stats[staking_address][self.TOTAL_TIME] / count
-            )
-
-    def order_addresses_by_latency(self, staking_addresses: List[str]):
-        result = sorted(
-            staking_addresses,
-            key=lambda staking_address: self.get_average_latency_time(staking_address),
-        )
-        return result
 
 
 class NodeSprout:
--- a/nucypher/network/nodes.py
+++ b/nucypher/network/nodes.py
@@ -1,8 +1,9 @@
 import time
-from collections import deque
+from collections import defaultdict, deque
 from contextlib import suppress
 from queue import Queue
-from typing import Optional, Set, Tuple
+from threading import Lock
+from typing import List, Optional, Set, Tuple
 
 import maya
 import requests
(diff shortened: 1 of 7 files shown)


derekpierre:
Something we could do is parameterize whether stats are collected or not in the `Learner` constructor - default being `False`. For e.g. `Ursula` objects do not need to collect stats since they doesn't send threshold decryption requests. However, the `Bob` and `Porter` objects do, so they would set that value to `True` by default. Any thoughts about this? (cc @cygnusv , @KPrasch ).

vzotova (approved):

manumonti (approved):
This looks great! 🙌

nucypher/network/nodes.py
```diff
@@ -827,11 +830,14 @@ def learn_from_teacher_node(self, eager=False, canceller=None):
...
 
         try:
-            response = self.network_middleware.get_nodes_via_rest(
-                node=current_teacher,
-                announce_nodes=announce_nodes,
-                fleet_state_checksum=self.known_nodes.checksum,
-            )
+            with self.node_latency_collector.get_latency_tracker(
+                current_teacher.checksum_address
+            ):
```
cygnusv, line 835: A context manager...fancy!

nucypher/utilities/latency.py
```diff
@@ -0,0 +1,84 @@
...
+        )
+        self._lock = Lock()
+
+    def _update_stats(self, staking_address: ChecksumAddress, latest_time_taken: float):
+        with self._lock:
+            old_avg = self._node_stats[staking_address][self.CURRENT_AVERAGE]
+            old_count = self._node_stats[staking_address][self.COUNT]
+
+            updated_count = old_count + 1
+            updated_avg = ((old_avg * old_count) + latest_time_taken) / updated_count
```
cygnusv, line 54: I think this is fine for this PR, but at a certain point, historical time measures are not relevant. I think a potential next iteration of this context manager can focus on collecting statistics for the last X hours or X usages. WDYT?

nucypher/utilities/latency.py
```diff
@@ -0,0 +1,84 @@
...
+            self,
+            stats_collector: "NodeLatencyStatsCollector",
+            staker_address: ChecksumAddress,
+        ):
+            self._stats_collector = stats_collector
+            self.staker_address = staker_address
+
+        def __enter__(self):
+            self.start_time = time.perf_counter()
+            return self
```
cygnusv, line 29: Is it necessary to use to lock in here (and `__exit__`) too? Or in other words, is it possible for `self.start_time` to become a shared resource?

nucypher/utilities/latency.py
derekpierre: Yep. We could do a moving average of `X` values eg. we could always store the last `X` values e.g. the last 5 latency times and do an ongoing average of that. Every new value would overwrite the oldest value.

nucypher/utilities/latency.py
derekpierre: > is it possible for self.start_time to become a shared resource?

It's not possible in the way we currently use it. 

However, it's generally possible, but seems more of a misuse of the context manager instance.

i.e. the expected usage is generally
```python
with node_latency_collector.get_latency_tracker(<>):
    ...
```
i.e. a fresh instance of the context manager is used when needed. Fresh instances prevent the the shared resource since they are self-contained.

The scenarios I can think of for `start_time` to become a shared resource would be:
1. nesting of the context manager. e.g.
```python
# get and store instance
tracker = node_latency_collector.get_latency_tracker(<>)

with tracker:
     ....
     with tracker:
         ...
```

so the nesting causes the original start time from the outside block to be changed when the internal block is hit. This is particularly ugly and shouldn't be done.

OR 

2. the use of the same instance in multiple threads:
```python
# get and store instance
tracker = node_latency_collector.get_latency_tracker(<>)

# use tracker instance within thread_1
# use same tracker instance within thread_2
...
```
So, you would need to have a common context manager instance and then use that instance concurrently in separate threads.

nucypher/utilities/latency.py
derekpierre: Fixed via https://github.com/nucypher/nucypher/pull/3562/commits/fb4152d8e20004d738f469c87bcfb12f971c88d7.

Used a window of 5. If you (or anyone else) have thoughts on a better value to use, let me know.

nucypher/utilities/latency.py
cygnusv: Right, I think I was confused by the use of locks to handle stats, but then I realized this is due to the stats collector being the potentially shared resource.

tests/unit/test_node_latency_collector.py
```diff
@@ -0,0 +1,303 @@
...
+    ]  # test of the test - "that's sooo meta"
+
+    return executions, sorted_order
+
+
+def floats_sufficiently_equal(a: float, b: float):
+    if abs(a - b) < 1e-9:
+        return True
+
+    return False
```
cygnusv, line 50:
```suggestion
    return abs(a - b) < 1e-9
```
But perhaps a better approach is to use `math.isclose(a, b)`:
https://docs.python.org/3/library/math.html#math.isclose

nucypher/utilities/latency.py
```diff
@@ -0,0 +1,78 @@
...
+                self._stats_collector.reset_stats(self.staker_address)
+            else:
+                # no exception
+                end_time = time.perf_counter()
+                execution_time = end_time - self.start_time
+                self._stats_collector._update_stats(self.staker_address, execution_time)
+
+    def __init__(self, max_moving_average_window: int = MAX_MOVING_AVERAGE_WINDOW):
+        self._node_stats = defaultdict(
+            lambda: collections.deque([], maxlen=max_moving_average_window)
```
cygnusv, line 43: 💯

nucypher/utilities/latency.py
```diff
@@ -0,0 +1,78 @@
...
+            stats_collector=self, staker_address=staker_address
+        )
+
+    def get_average_latency_time(self, staking_address: ChecksumAddress) -> float:
+        with self._lock:
+            readings = list(self._node_stats[staking_address])
+            num_readings = len(readings)
+            # just need a large number > 0
+            return (
+                self.MAX_LATENCY if num_readings == 0 else sum(readings) / num_readings
```
cygnusv, line 68: A potential discussion is to evaluate the difference between returning the average or the median value, or in other words, to assess if the returned value should be affected or not by outliers. From the latency analysis I did, we can see nodes that do very well on average have mediocre medians, and the other way around.

cygnusv (approved):

tests/unit/test_node_latency_collector.py
derekpierre: ✔️  - nice find!

tests/unit/test_node_latency_collector.py
```diff
@@ -0,0 +1,297 @@
...
+    sorted_order = sorted(
+        list(executions.keys()),
+        key=lambda x: sum(executions.get(x)) / len(executions.get(x)),
+    )
+    assert sorted_order == [
+        node_4,
+        node_3,
+        node_2,
+        node_1,
+    ]  # test of the test - "that's sooo meta"
```
KPrasch, line 42: 😆

nucypher/network/nodes.py
```diff
@@ -827,11 +835,19 @@ def learn_from_teacher_node(self, eager=False, canceller=None):
...
+                    current_teacher.checksum_address
+                )
+                if self.node_latency_collector
+                else contextlib.nullcontext()
             )
+            with optional_latency_context_manager:
+                response = self.network_middleware.get_nodes_via_rest(
+                    node=current_teacher,
+                    announce_nodes=announce_nodes,
+                    fleet_state_checksum=self.known_nodes.checksum,
```
KPrasch, line 849: This is a reasonable place to gather latency samples - that being said are there any other options  we can consider?  Side note: based on the construction of the collector below it's important to only measure one request on a regular basis or else we'll get skewed results.  Am I understanding this correctly?

nucypher/network/nodes.py
```diff
@@ -827,11 +835,19 @@ def learn_from_teacher_node(self, eager=False, canceller=None):
...
-            response = self.network_middleware.get_nodes_via_rest(
-                node=current_teacher,
-                announce_nodes=announce_nodes,
-                fleet_state_checksum=self.known_nodes.checksum,
+            optional_latency_context_manager = (
+                self.node_latency_collector.get_latency_tracker(
+                    current_teacher.checksum_address
+                )
+                if self.node_latency_collector
+                else contextlib.nullcontext()
```
KPrasch, line 843: This basically just makes the context manager... do nothing... in case of it being switched off, correct?  What's your motivation for installing a feature switch?  Do you plan to expose it via ursula config or CLI?

nucypher/utilities/latency.py
```diff
@@ -0,0 +1,78 @@
...
+
+        def __exit__(self, exc_type, exc_val, exc_tb):
+            if exc_type:
+                # exception occurred - reset stats since connectivity was compromised
+                self._stats_collector.reset_stats(self.staker_address)
+            else:
+                # no exception
+                end_time = time.perf_counter()
+                execution_time = end_time - self.start_time
+                self._stats_collector._update_stats(self.staker_address, execution_time)
```
KPrasch, line 39: Nice usage of a context manager here - the API is very portable this way.   👍🏻

nucypher/utilities/latency.py
```diff
@@ -0,0 +1,78 @@
...
+
+from eth_typing import ChecksumAddress
+
+
+class NodeLatencyStatsCollector:
+    """
+    Thread-safe utility that tracks latency statistics related to P2P connections with other nodes.
+    """
+
+    MAX_MOVING_AVERAGE_WINDOW = 5
```
KPrasch, line 15: This is 5 deque values?

nucypher/utilities/latency.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def get_latency_tracker(
+        self, staker_address: ChecksumAddress
+    ) -> NodeLatencyContextManager:
+        return self.NodeLatencyContextManager(
+            stats_collector=self, staker_address=staker_address
+        )
+
+    def get_average_latency_time(self, staking_address: ChecksumAddress) -> float:
+        with self._lock:
```
KPrasch, line 63: Where/when do you anticipate this being a shared resource?

nucypher/network/nodes.py
derekpierre: Yep, you got it. In general, we need to be consistent when measuring "node latency", i.e. use the same measurement for all nodes.

nucypher/network/nodes.py
derekpierre: > This basically just makes the context manager... do nothing... in case of it being switched off, correct?

Yep. 

> What's your motivation for installing a feature switch? Do you plan to expose it via ursula config or CLI?

Related to this - https://github.com/nucypher/nucypher/pull/3562#discussion_r1830745901.

Ursula has no *current* use for the statistics, but Bob and Porter do.

KPrasch: Ah right, makes sense. 👍🏻

nucypher/utilities/latency.py
derekpierre: Yep, only the last 5 values will be used for judging latency. At the moment we calculate average - but that calc can change if we find it not sufficiently representative of actual latency. @cygnusv 's comment here - https://github.com/nucypher/nucypher/pull/3562#discussion_r1830745901

nucypher/utilities/latency.py
derekpierre: In Porter/Bob's case it's if the stats values are being updated (via learning loop task), while decryption requests are being sent (the thread factory in `ThresholdDecryptionClient`).

nucypher/network/nodes.py
derekpierre: I linked to the wrong comment above - updated.

nucypher/utilities/latency.py
KPrasch: My intuition is that this is a pretty small window, but since we're mainly using this calculation for instant consumption, it's a reasonable size. 👍🏻

KPrasch (approved):
Nice implementation!  

Feature request (for this PR or a future PR): Publish current latency results in a JSON status endpoint or prometheus collector for continued optimization research and performance monitoring. 🤠

derekpierre:
> Feature request (for this PR or a future PR): Publish current latency results in a JSON status endpoint or prometheus collector for continued optimization research and performance monitoring. 🤠

💯 - perhaps we could add it to Porter's prometheus metrics.

cygnusv (approved):

PR #3564, opened 2024-11-06
Json Condition :userAddress context variable

theref:
- [x] failing test
- [x] fix

### DIFF
--- a/tests/unit/conditions/test_json_api_condition.py
+++ b/tests/unit/conditions/test_json_api_condition.py
@@ -258,7 +258,9 @@ def test_json_api_condition_evaluation_with_auth_token(mocker):
     )
 
 
-def test_json_api_condition_evaluation_with_user_address_context_variable(mocker):
+def test_json_api_condition_evaluation_with_user_address_context_variable(
+    mocker, valid_eip712_auth_message
+):
     mocked_get = mocker.patch(
         "requests.get",
         return_value=mocker.Mock(status_code=200, json=lambda: 0.0),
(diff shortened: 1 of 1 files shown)


tests/unit/conditions/test_json_api_condition.py
```diff
@@ -258,6 +258,27 @@ def test_json_api_condition_evaluation_with_auth_token(mocker):
     )
 
 
+def test_json_api_condition_evaluation_with_user_address_context_variable(
+    mocker, valid_eip712_auth_message
```
derekpierre, line 262: Let's use `valid_eip4361_auth_message` instead. EIP712 support will be deprecated in the future. It hasn't yet because older versions of `@nucypher/taco` (`taco-web`) used it.

derekpierre (commented):
Is the goal of the PR to add this test?

tests/unit/conditions/test_json_api_condition.py
theref: I thought I'd found a bug, but this test proves that I haven't.
More than happy to make the switch to 4361

derekpierre: > I thought I'd found a bug, but this test proves that I haven't.

😌 - that's a relief! Thanks for the additional testing. I'm always pro more tests.

Side note: you can add a dev newsfragment for this PR to make the "Check Release Note" CI job pass.

derekpierre (approved):
🎸  - looks good. Just needs a rebase.

PR #3566, opened 2024-12-02
Allow RPC urls to be used for RPC and Contract Conditions

theref:
**Type of PR:**
- Feature

**Required reviews:** 
- 2

**What this does:**
- update schema for RPC/Contract conditions
- if an rpc url is provided, use it. Otherwise default to current chainID based logic

### DIFF
--- a/nucypher/policy/conditions/evm.py
+++ b/nucypher/policy/conditions/evm.py
@@ -83,12 +83,8 @@ class RPCCall(ExecutionCall):
                 "null": "Undefined method name",
             },
         )
-        parameters = fields.List(
-            fields.Field, attribute="parameters", required=False, allow_none=True
-        )
-        rpc_endpoint = fields.Url(
-            attribute="rpc_endpoint", required=False, relative=False, allow_none=True
-        )
+        parameters = fields.List(fields.Field, required=False, allow_none=True)
+        rpc_endpoint = fields.Url(required=False, relative=False, allow_none=True)
 
         @validates_schema
         def validate_chain(self, data, **kwargs):
--- a/nucypher/policy/conditions/evm.py
+++ b/nucypher/policy/conditions/evm.py
@@ -173,11 +173,9 @@ class RPCCall(ExecutionCall):
         self,
         provider: HTTPProvider,
         resolved_parameters: List[Any],
-        check_chain: bool = True,
     ) -> Any:
         w3 = self._configure_w3(provider)
-        if check_chain:
-            self._check_chain_id(w3)
+        self._check_chain_id(w3)
         return self._execute(w3, resolved_parameters)
 
     def execute(self, providers: Dict[int, Set[HTTPProvider]], **context) -> Any:
@@ -204,7 +202,7 @@ class RPCCall(ExecutionCall):
 
         if self.rpc_endpoint:
             result, error = self._execute_with_provider(
-                HTTPProvider(self.rpc_endpoint), resolved_parameters, check_chain=False
+                HTTPProvider(self.rpc_endpoint), resolved_parameters
             )
             if error:
                 raise RPCExecutionFailed(error)
--- a/nucypher/policy/conditions/evm.py
+++ b/nucypher/policy/conditions/evm.py
@@ -183,6 +183,21 @@ class RPCCall(ExecutionCall):
                 self.parameters, **context
             )
 
+        # use local rpc endpoint if available
+        if self.chain in providers:
+            latest_error = ""
+            for provider in self._next_endpoint(providers=providers):
+                result, error = self._execute_with_provider(
+                    provider, resolved_parameters
+                )
+                if result:
+                    return result
+                latest_error = error
+
+            raise RPCExecutionFailed(
+                f"RPC call '{self.method}' failed; latest error - {latest_error}"
+            )
+
         if self.rpc_endpoint:
             result, error = self._execute_with_provider(
                 HTTPProvider(self.rpc_endpoint), resolved_parameters, check_chain=False
@@ -191,16 +206,7 @@ class RPCCall(ExecutionCall):
                 raise RPCExecutionFailed(error)
             return result
 
-        latest_error = ""
-        for provider in self._next_endpoint(providers=providers):
-            result, error = self._execute_with_provider(provider, resolved_parameters)
-            if result:
-                return result
-            latest_error = error
-
-        raise RPCExecutionFailed(
-            f"RPC call '{self.method}' failed; latest error - {latest_error}"
-        )
+        raise NoConnectionToChain(chain=self.chain)
 
     def _execute(
         self, w3: Web3, resolved_parameters: List[Any]
--- a/nucypher/policy/conditions/evm.py
+++ b/nucypher/policy/conditions/evm.py
@@ -35,7 +35,6 @@ from nucypher.policy.conditions.context import (
 )
 from nucypher.policy.conditions.exceptions import (
     NoConnectionToChain,
-    RequiredContextVariable,
     RPCExecutionFailed,
 )
 from nucypher.policy.conditions.lingo import (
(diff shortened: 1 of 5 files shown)


nucypher/policy/conditions/evm.py
```diff
@@ -168,41 +169,64 @@ def _next_endpoint(
...
+    def _execute_with_provider(
+        self,
+        provider: HTTPProvider,
+        resolved_parameters: List[Any],
+        check_chain: bool = True,
+    ) -> Any:
+        w3 = self._configure_w3(provider)
+        if check_chain:
+            self._check_chain_id(w3)
+        return self._execute(w3, resolved_parameters)
```
derekpierre, line 181: The checking of the chain should always be done. It ensures that the provider's chain id and condition chain id always match up - this is irrespective of whether the customized rpc_endpoint is provided or not.

```suggestion
    def _execute_with_provider(
        self,
        provider: HTTPProvider,
        resolved_parameters: List[Any],
    ) -> Any:
        w3 = self._configure_w3(provider)
        self._check_chain_id(w3)
        return self._execute(w3, resolved_parameters)
```

^ callers will need to be updated.

nucypher/policy/conditions/evm.py
```diff
@@ -168,41 +169,64 @@ def _next_endpoint(
...
+            rpc_endpoint_, rpc_method = self.method.split("_", 1)
+            rpc_function = self._get_web3_py_function(w3, rpc_method)
+            rpc_result = rpc_function(*resolved_parameters)  # RPC read
+            return rpc_result, None
+        except RequiredContextVariable:
+            raise
+        except Exception as e:
+            error = f"RPC call '{self.method}' failed: {e}"
+            self.LOG.warn(error)
+            return None, error
```
derekpierre, line 229: How come you decided to move this here, and not continue to allow this method raise the exception, and it be handled at the top level? Maybe I'm missing something, but `_execute` should propagate the exception, and let the code at the higher level handle it appropriately. It feels like the use of multiple endpoints is now creeping into a specific method for execution a rpc call, and any other methods in the call stack before the code to try a different endpoint.

Same comment in other spots (contract, time) where the same thing is done.

tests/unit/conditions/test_rpc_condition.py
```diff
@@ -133,3 +136,98 @@ def test_rpc_condition_invalid_comparator_value_type(invalid_value, rpc_conditio
...
+        chain=unsupported_chain,
+        return_value_test=ReturnValueTest("==", 0),
+        parameters=["0xaDD9D957170dF6F33982001E4c22eCCdd5539118"],
+        rpc_endpoint="https://fallback.example.com",
+    )
+
+    condition.verify(providers={})
+    mock_http_provider_spy.assert_called_once_with(ANY, "https://fallback.example.com")
+
+    # Test Case 3: Unsupported chain with no rpc_endpoint - should raise error
```
derekpierre, line 226: 👍

tests/unit/conditions/test_time_condition.py
```diff
@@ -86,3 +87,29 @@ def test_time_condition_invalid_comparator_value_type(invalid_value, time_condit
...
             ),
         )
+
+
+def test_time_condition_verify(mocker):
+    # Set up Web3 with mocked responses
+    mock_block = mocker.Mock()
+    mock_block.timestamp = 1000  # Set a fixed timestamp
+
+    w3 = mocker.Mock()
```
derekpierre, line 97: I believe using `mocker.MagicMock()` here will save you from having to always set the middleware onion as Mock (L100) - that property would automatically return a mock.

tests/unit/conditions/test_time_condition.py
```diff
@@ -86,3 +87,29 @@ def test_time_condition_invalid_comparator_value_type(invalid_value, time_condit
...
+        "nucypher.policy.conditions.time.TimeRPCCall._configure_w3", return_value=w3
+    )
+
+    # Test case 1: Current time is greater than condition time
+    condition = TimeCondition(
+        return_value_test=ReturnValueTest(">", 500),  # Should pass as 1000 > 500
+        chain=TESTERCHAIN_CHAIN_ID,
+        method=TimeRPCCall.METHOD,
+    )
+    local_provider = HTTPProvider("https://local-provider.example.com")
```
derekpierre, line 113: Although the line is effectively the same you have the option to save yourself from always needing to use a string:
```python
local_provider = Mock(spec=HTTPProvider)
```

derekpierre (changes requested):
For testing, you should also add an additional rpc condition that uses the rpc_endpoint property to the condition lingo test - https://github.com/nucypher/nucypher/blob/v7.5.x/tests/unit/conditions/test_condition_lingo.py#L27.

Once added to the compound condition, it will be tested to ensure that serialization/deserialization of the condition (with the rpc endpoint value) works.

tests/unit/conditions/test_rpc_condition.py
```diff
@@ -133,3 +136,98 @@ def test_rpc_condition_invalid_comparator_value_type(invalid_value, rpc_conditio
...
+
+    # Mock eth module
+    mock_eth = mocker.Mock()
+    mock_eth.get_balance.return_value = 0
+    mock_eth.chain_id = 8453
+
+    # Create Web3 mock with required attributes
+    mock_w3 = mocker.Mock()
+    mock_w3.eth = mock_eth
+    mock_w3.middleware_onion = mocker.Mock()
```
derekpierre, line 153: You can collapse this into:
```suggestion
    # Mock eth module
    mock_w3 = mocker.Mock()
    mock_w3.eth.get_balance.return_value = 0
    mock_w3.eth.chain_id = 8453
```

tests/unit/conditions/test_rpc_condition.py
```diff
@@ -133,3 +136,98 @@ def test_rpc_condition_invalid_comparator_value_type(invalid_value, rpc_conditio
                 value=invalid_value,
             ),
         )
+
+
+def test_rpc_condition_uses_provided_endpoint(mocker):
+    # Mock HTTPProvider
```
derekpierre, line 142:
```suggestion
    # Spy HTTPProvider
```

^ some other similar spots

tests/unit/conditions/test_time_condition.py
```diff
@@ -86,3 +87,29 @@ def test_time_condition_invalid_comparator_value_type(invalid_value, time_condit
...
+
+def test_time_condition_verify(mocker):
+    # Set up Web3 with mocked responses
+    mock_block = mocker.Mock()
+    mock_block.timestamp = 1000  # Set a fixed timestamp
+
+    w3 = mocker.Mock()
+    w3.eth.get_block.return_value = mock_block
+    w3.eth.chain_id = TESTERCHAIN_CHAIN_ID
+    w3.middleware_onion = mocker.Mock()
```
derekpierre, line 100: I don't think this line is necessary:
```suggestion
```

nucypher/policy/conditions/evm.py
theref: fixed :)

nucypher/policy/conditions/evm.py
theref: at the time i thought i was removing duplicated code, but i hadn't realised that i'd need to repeat this for several conditions. You're right, it's a downgrade - I've reset the behaviour

nucypher/policy/conditions/evm.py
```diff
@@ -87,12 +87,17 @@ class Schema(ExecutionCall.Schema):
         parameters = fields.List(
             fields.Field, attribute="parameters", required=False, allow_none=True
         )
+        rpc_endpoint = fields.Url(
+            attribute="rpc_endpoint", required=False, relative=False, allow_none=True
```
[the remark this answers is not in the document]
theref: yep, works fine without :)

derekpierre:
Subsumed by #3569.

PR #3569, opened 2024-12-06
EVM Condition on "Any" (Major) EVM Chain (only `lynx` for now)

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Very rough initial work as a POC for `lynx` to support any EVM chain for EVM (RPC, Contract, Time) conditions.

A barebones version as of [d414553](https://github.com/nucypher/nucypher/pull/3569/commits/d414553e5306f4de6b4789a7c734a75caa98c52c) is actually running on `lynx` currently via the `nucypher/nucypher:experimental` docker image.

Some outstanding questions/discussion points:
- [x] How do we limit use to `lynx`. Conditions don't have a concept of domain so even using the mapping, the condition needs to have a way to determine what domain the node is running in

**A: We obtain default rpc endpoints for supported chains based on domain. Each domain has their own rpc endpoint mapping in https://github.com/nucypher/chainlist**

- [x] Should we continue to use `https://chainid.network` URL or use our own more robust storage (eg. github + local, IPFS etc.). Centralization concerns using chainid.network - that being said, if this is only for lynx maybe we dont' care

**A: We will use https://github.com/nucypher/chainlist (github solution) for now.**

- [x] Fetching RPC endpoints is currently `memoize` i.e. only one call is made and then cached. While caching is a good thing perhaps we should use a timed cache, i.e. fetch a fresh list every hour / or some other timeframe? However, do we expect the list to change or not; if not, then time-based cache doesn't achieve anything, and perhaps we can just proactively make the request on startup.

**A: public RPC endpoints are obtained on startup from https://github.com/nucypher/chainlist. Health checks are performed to `mainnet` but not testnets (for now). `lynx` (not `tapir`) will now support many chains with many rpc endpoints and performing health checks will cause start up to be very slow. Since this is just for testnet we will skip this for now. In any case, the fallback nature of the using the RPC endpoints allow for redundancy if any RPC endpoint is down.**

- [x] Do we need health checks for public rpc URLs. Currently the public RPC URLs are obtained at decrytion time, so any health check slows this down - can we be more proactive about health checks in some way

**A: Health checks are only performed for `mainnet` domain, and done on startup and not at decryption time**

- [x] Does this functionality also allow for replacement of the data used for default RPC endpoints done in https://github.com/nucypher/nucypher/pull/3496 for officially supported chains.

**A: Not a replacement, but we ended up utilizing it since we continue to obtain default public rpc endpoints on startup. The only thing that really changed is that https://github.com/nucypher/chainlist now contains separate endpoint mapping files for each domain.**

^ All of these don't need to be addressed in this PR but they are for facilitating thoughts around whether a) should they be done or not, and b) should they be done sooner rather than later. The `lynx` restriction is necessary, however.

**Issues fixed/closed:**
> - Fixes #...

- Subsumes #3566 .
- Related work done in `nucypher/chainlist`:
    - https://github.com/nucypher/chainlist/pull/2
    - https://github.com/nucypher/chainlist/pull/3

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/newsfragments/3569.feature.rst
+++ b/newsfragments/3569.feature.rst
@@ -1,0 +1,1 @@
+Added plumbing to support EVM condition evaluation on "any" (major) EVM chain outside of Ethereum and Polygon - only enabled on `lynx` testnet for now.
--- a/newsfragments/3569.feature.rst
+++ b/newsfragments/3569.feature.rst
@@ -1,1 +1,1 @@
-Added plumbing to support EVM condition evaluation on "any" (major) EVM chain outside of Ethereum and Polygon - only enabled on `lynx` testnet for now.
+Added plumbing to support EVM condition evaluation on "any" (major) EVM chain outside of Ethereum and Polygon - only enabled on ``lynx`` testnet for now.
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -65,7 +65,6 @@ from nucypher.crypto.powers import (
     TransactingPower,
 )
 from nucypher.datastore.dkg import DKGStorage
-from nucypher.policy.conditions.evm import _CONDITION_CHAINS
 from nucypher.policy.conditions.utils import evaluate_condition_lingo
 from nucypher.policy.payment import ContractPayment
 from nucypher.types import PhaseId
@@ -266,38 +265,38 @@ class Operator(BaseActor):
             return receipt
 
     @staticmethod
-    def _is_permitted_condition_chain(chain_id: int) -> bool:
-        return int(chain_id) in [int(cid) for cid in _CONDITION_CHAINS.keys()]
-
-    @staticmethod
     def _make_condition_provider(uri: str) -> HTTPProvider:
         provider = HTTPProvider(endpoint_uri=uri)
         return provider
 
     def connect_condition_providers(
-        self, endpoints: Dict[int, List[str]]
+        self, user_defined_endpoints: Dict[int, List[str]]
     ) -> DefaultDict[int, List[HTTPProvider]]:
-        providers = defaultdict(list)  # use list to maintain order
 
-        # check that we have endpoints for all condition chains
-        if set(self.domain.condition_chain_ids) != set(endpoints):
+        # check that we have mandatory user configured endpoints
+        mandatory_user_configured_chains = {
+            self.domain.eth_chain.id,
+            self.domain.polygon_chain.id,
+        }
+        if mandatory_user_configured_chains != set(user_defined_endpoints):
             raise self.ActorError(
-                f"Missing blockchain endpoints for chains: "
-                f"{set(self.domain.condition_chain_ids) - set(endpoints)}"
+                f"User supplied endpoints for chains don't match mandatory chains: "
+                f"{set(user_defined_endpoints)} vs expected {mandatory_user_configured_chains}"
             )
+
+        providers = defaultdict(list)  # use list to maintain order
 
         # ensure that no endpoint uri for a specific chain is repeated
         duplicated_endpoint_check = defaultdict(set)
 
         # User-defined endpoints for chains
-        for chain_id, endpoints in endpoints.items():
-            if not self._is_permitted_condition_chain(chain_id):
-                raise NotImplementedError(
-                    f"Chain ID {chain_id} is not supported for condition evaluation by this operator."
+        for chain_id, endpoint_list in user_defined_endpoints.items():
+            if not endpoint_list:
+                raise self.ActorError(
+                    f"No mandatory user endpoint configured for chain {chain_id}"
                 )
-
             # connect to each endpoint and check that they are on the correct chain
-            for uri in endpoints:
+            for uri in endpoint_list:
                 if uri in duplicated_endpoint_check[chain_id]:
                     self.log.warn(
                         f"Duplicated user-supplied blockchain uri, {uri}, for condition evaluation on chain {chain_id}; skipping"
@@ -318,9 +317,12 @@ class Operator(BaseActor):
                 duplicated_endpoint_check[chain_id].add(uri)
 
         # Ingest default/fallback RPC providers for each chain
-        for chain_id in self.domain.condition_chain_ids:
-            default_endpoints = get_healthy_default_rpc_endpoints(chain_id)
-            for uri in default_endpoints:
+        # for chain_id in self.domain.condition_chain_ids:
+        default_endpoints = get_healthy_default_rpc_endpoints(self.domain)
+        for chain_id, chain_endpoints in default_endpoints.items():
+            # randomize list so that the same fallback RPC endpoints aren't always used by all nodes
+            random.shuffle(chain_endpoints)
+            for uri in chain_endpoints:
                 if uri in duplicated_endpoint_check[chain_id]:
                     self.log.warn(
                         f"Duplicated fallback blockchain uri, {uri}, for condition evaluation on chain {chain_id}; skipping"
@@ -330,12 +332,9 @@ class Operator(BaseActor):
                 providers[chain_id].append(provider)
                 duplicated_endpoint_check[chain_id].add(uri)
 
-        humanized_chain_ids = ", ".join(
-            _CONDITION_CHAINS[chain_id] for chain_id in providers
-        )
         self.log.info(
             f"Connected to {sum(len(v) for v in providers.values())} RPC endpoints for condition "
-            f"checking on chain IDs {humanized_chain_ids}"
+            f"checking on chain IDs {providers.keys()}"
         )
 
         return providers
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -317,7 +317,6 @@ class Operator(BaseActor):
                 duplicated_endpoint_check[chain_id].add(uri)
 
         # Ingest default/fallback RPC providers for each chain
-        # for chain_id in self.domain.condition_chain_ids:
         default_endpoints = get_healthy_default_rpc_endpoints(self.domain)
         for chain_id, chain_endpoints in default_endpoints.items():
             # randomize list so that the same fallback RPC endpoints aren't always used by all nodes
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -270,18 +270,18 @@ class Operator(BaseActor):
         return provider
 
     def connect_condition_providers(
-        self, user_defined_endpoints: Dict[int, List[str]]
+        self, operator_configured_endpoints: Dict[int, List[str]]
     ) -> DefaultDict[int, List[HTTPProvider]]:
 
         # check that we have mandatory user configured endpoints
-        mandatory_user_configured_chains = {
+        mandatory_configured_chains = {
             self.domain.eth_chain.id,
             self.domain.polygon_chain.id,
         }
-        if mandatory_user_configured_chains != set(user_defined_endpoints):
+        if mandatory_configured_chains != set(operator_configured_endpoints):
             raise self.ActorError(
-                f"User supplied endpoints for chains don't match mandatory chains: "
-                f"{set(user_defined_endpoints)} vs expected {mandatory_user_configured_chains}"
+                f"Operator-configured condition endpoints for chains don't match mandatory chains: "
+                f"{set(operator_configured_endpoints)} vs expected {mandatory_configured_chains}"
             )
 
         providers = defaultdict(list)  # use list to maintain order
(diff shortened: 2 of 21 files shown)


nucypher/policy/conditions/evm.py
```diff
@@ -154,15 +155,17 @@ def _next_endpoint(
...
-
-        # if there are no entries for the chain ID, there
-        # is no connection to that chain available.
-        except KeyError:
-            raise NoConnectionToChain(chain=self.chain)
+        rpc_providers = providers.get(self.chain, None)
         if not rpc_providers:
-            raise NoConnectionToChain(chain=self.chain)  # TODO: unreachable?
+            # only use when no connection available
+            public_rpc_providers = fetch_public_rpc_endpoints_for_chain(
```
theref, line 161: i think we should call this on node startup, rather than every time a condition is evaluated

theref (commented):
This is great, I think we need to decide how we're going to distinguish between mainnet/tapir/lynx. My suggestion is to have a hardcoded mapping that is read on startup. Each domain will point at a different url and we can manage the contents of those urls (including potentially encrypting the mainnet file if it contains a 'master' dRPC key)

nucypher/blockchain/eth/decorators.py
```diff
@@ -111,7 +111,7 @@ def save_receipt(actor_method) -> Callable:  # TODO: rename to "save_result"?
     @functools.wraps(actor_method)
     def wrapped(self, *args, **kwargs) -> dict:
         receipt_or_txhash = actor_method(self, *args, **kwargs)
-        self._saved_receipts.append((datetime.utcnow(), receipt_or_txhash))
+        self._saved_receipts.append((datetime.now(timezone.utc), receipt_or_txhash))
```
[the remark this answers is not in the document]
derekpierre: `datetime.utcnow()` has been deprecated.

nucypher/policy/conditions/evm.py
derekpierre: 👍  - That's reasonable. It may play in with whether we want to address the following, which I added in the PR description.
```
Does this functionality also allow for replacement of the data used for default RPC endpoints done in https://github.com/nucypher/nucypher/pull/3496 for officially supported chains.
```

Currently `fetch_public_rpc_endponts_for_chain` (`_fetch_public_rpc_endpoints_from_chainlist`) uses `memoize` so only one HTTP request call is ever made and the result is cached - so the request is not actually made every condition evaluation but rather only the first one. 

Also, we could think about possibly using a time-based cache instead, if we expect the list is refreshed every so often (eg. one per hour). If not, then caching is one time, and to your point, can happen once at startup.

Updated the PR description to have this discussion point.

derekpierre:
> This is great, I think we need to decide how we're going to distinguish between mainnet/tapir/lynx. My suggestion is to have a hardcoded mapping that is read on startup. Each domain will point at a different url and we can manage the contents of those urls (including potentially encrypting the mainnet file if it contains a 'master' dRPC key)

The mapping could be a good addition, but If I'm understanding (?), it may not completely solve our problem. The tricky part is deciding which domain entry to use in the mapping from within the condition if we customise behaviour based on the domain. Condition objects have no clue about the domain of the overarching Ursula node, so this is a mechanism we need to come up with a solution for - either a way for Conditions to determine domain, OR, a solution to restrict "any chain" to `lynx` outside of conditions that allows conditions to continue not to have to care about domain.

nucypher/blockchain/eth/decorators.py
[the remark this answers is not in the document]
theref: awesome, thanks

nucypher/policy/conditions/evm.py
theref: ah nice, love that it's using `memoize` that's perfect

theref:
> > This is great, I think we need to decide how we're going to distinguish between mainnet/tapir/lynx. My suggestion is to have a hardcoded mapping that is read on startup. Each domain will point at a different url and we can manage the contents of those urls (including potentially encrypting the mainnet file if it contains a 'master' dRPC key)
> 
> The mapping could be a good addition, but If I'm understanding (?), it may not completely solve our problem. The tricky part is deciding which domain entry to use in the mapping from within the condition if we customise behaviour based on the domain. Condition objects have no clue about the domain of the overarching Ursula node, so this is a mechanism we need to come up with a solution for - either a way for Conditions to determine domain, OR, a solution to restrict "any chain" to `lynx` outside of conditions that allows conditions to continue not to have to care about domain.

the condition doesn't need to have any knowledge of the domain, for the same reason that it doesn't need to at the moment. You just send a decryption request to lynx/mainnet etc and the nodes handle it. A mainnet node will only have access to mainnet chains, a tesnet node will only have access to tesnet chains

nucypher/policy/conditions/evm.py
theref: Completely agree, I think caching is much better than just a simple call on startup

derekpierre:
> > the condition doesn't need to have any knowledge of the domain, for the same reason that it doesn't need to at the moment. You just send a decryption request to lynx/mainnet etc and the nodes handle it. A mainnet node will only have access to mainnet chains, a tesnet node will only have access to tesnet chains
> 
> Oh I think I see what you mean. You mean have the node configure the condition chains / providers ahead of time, so that the providers passed into the condition evaluation call are already there. That's interesting! 🎸

Although, that's a lot of chains.

derekpierre:
Merged work from @theref 's PR - https://github.com/derekpierre/nucypher/pull/8.

tests/unit/conditions/test_rpc_condition.py
```diff
@@ -47,15 +47,6 @@ def test_invalid_rpc_condition():
             parameters=["0xaDD9D957170dF6F33982001E4c22eCCdd5539118"],
         )
 
-    # unsupported chain id
```
theref, line 50: I think we should keep this test

tests/unit/conditions/test_time_condition.py
```diff
@@ -27,14 +27,6 @@ def test_invalid_time_condition():
             method="time_after_time",
         )
 
-    # chain id not permitted
```
theref, line 30: and again - the difference is that rather than it being "not permitted" it's just "not supported"

tests/unit/conditions/test_time_condition.py
```diff
@@ -60,19 +52,16 @@ def test_time_condition_schema_validation(time_condition):
...
         condition_dict["method"] = "my_blocktime"
         TimeCondition.from_dict(condition_dict)
 
     with pytest.raises(InvalidConditionLingo):
         # chain id not an integer
+        condition_dict = time_condition.to_dict()
         condition_dict["chain"] = str(TESTERCHAIN_CHAIN_ID)
         TimeCondition.from_dict(condition_dict)
 
-    with pytest.raises(InvalidConditionLingo):
```
theref, line 71: and another - i think this was me :D sorry!

tests/unit/conditions/test_rpc_condition.py
```diff
@@ -101,14 +92,10 @@ def test_rpc_condition_schema_validation(rpc_condition):
 
     with pytest.raises(InvalidConditionLingo):
         # chain id not an integer
+        condition_dict = rpc_condition.to_dict()
         condition_dict["chain"] = str(TESTERCHAIN_CHAIN_ID)
         RPCCondition.from_dict(condition_dict)
 
-    with pytest.raises(InvalidConditionLingo):
```
theref, line 107: and here

theref (commented):
Approved :)

theref (approved):

nucypher/policy/conditions/evm.py
```diff
@@ -154,15 +137,10 @@ def _next_endpoint(
...
     ) -> Iterator[HTTPProvider]:
         """Yields the next web3 provider to try for a given chain ID"""
-        try:
-            rpc_providers = providers[self.chain]
-
-        # if there are no entries for the chain ID, there
-        # is no connection to that chain available.
-        except KeyError:
-            raise NoConnectionToChain(chain=self.chain)
+        rpc_providers = providers.get(self.chain, None)
```
KPrasch, line 140: `None` is the default non-matching return value for `.get`.  However, i want to clarify how we got here - is this meant to evaluate against an empty list because all health checks failed for that chain id?

nucypher/policy/conditions/evm.py
```diff
@@ -56,16 +56,6 @@
...
-_CONDITION_CHAINS = {
-    1: "ethereum/mainnet",
-    11155111: "ethereum/sepolia",
-    137: "polygon/mainnet",
-    80002: "polygon/amoy",
-    # TODO: Permit support for these chains
-    # 100: "gnosis/mainnet",
-    # 10200: "gnosis/chiado",
-}
-
```
KPrasch, line 68: Right.  This was the first iteration of limiting which chains are permitted for mainnet condition evaluation - which has now been replaced partly by nucpher/chainlist domain-named files.  Does this also imply that there are no longer any restrictions placed on user-defined endpoints as well?

nucypher/blockchain/eth/domains.py
```diff
@@ -107,7 +89,7 @@ def condition_chain_ids(self) -> set:
 }
 
 
-@memoize
+@cache
```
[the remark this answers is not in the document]
KPrasch: ah - I see now... this is part of rework to permit refreshing of endpoints based on health checks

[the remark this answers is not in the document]
KPrasch: ah - I see now... this is part of rework to permit refreshing of endpoints based on health checks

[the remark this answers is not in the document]
theref: same functionality, except that `memoize` doesn't allow you to reset the cache, which caused issues in testsing

nucypher/blockchain/eth/utils.py
```diff
@@ -159,23 +161,22 @@ def get_default_rpc_endpoints() -> Dict[int, List[str]]:
...
-
-    healthy = [
-        endpoint for endpoint in chain_endpoints if rpc_endpoint_health_check(endpoint)
-    ]
-    LOGGER.info(f"Healthy default RPC endpoints for chain ID {chain_id}: {healthy}")
-    if not healthy:
-        LOGGER.warn(
-            f"No healthy default RPC endpoints available for chain ID {chain_id}"
-        )
+    if not domain.is_testnet:
```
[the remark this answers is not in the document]
theref: because there are soooo many endpoints that it will be checking. This will take a very long time on testnet

nucypher/policy/conditions/evm.py
theref: this PR doesn't add functionality for user defined endpoints

nucypher/crypto/tls.py
```diff
@@ -42,7 +42,7 @@ def generate_self_signed_certificate(
         private_key = ec.generate_private_key(curve(), default_backend())
     public_key = private_key.public_key()
 
-    now = datetime.datetime.utcnow()
+    now = datetime.datetime.now(datetime.timezone.utc)
```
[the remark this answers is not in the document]
derekpierre: It is unrelated. `datetime.utcnow()` and `datetime.utcfromtimestamp()` are deprecated and we were getting warnings about it - https://github.com/nucypher/nucypher/pull/3569/commits/007fdeb9fca7516d31255e54c071dee047ae89a5

nucypher/policy/conditions/evm.py
derekpierre: Clarifying what you mean by "user".

If you mean restrictions on "node providers", that still holds for `mainnet`/`tapir` - see https://github.com/nucypher/nucypher/pull/3569/commits/da23d687bef0d23064b16aaae64c8865dfc3ea6b although `user_configured_endpoints` may be a misnomer there, perhaps it should be `node_configured_endpoints`... cc @theref .

If you mean the user that defines the conditions, the `nucypher/chainlist domain-named files` is what replaces user-defined endpoints in conditions. Instead, public rpc endpoints are used for whatever chain ID outside of ethereum/polygon is specified. It's part of the reason we are only allowing on `lynx` so that we can understand if this is sufficient or not. Developer prototyping can happen without the condition itself needing to hardcode an RPC URL.

nucypher/policy/conditions/evm.py
derekpierre: It's possible because we no longer validate chain id against a known list of chain IDs for a domain i.e. any chain id value is valid within the definition of a condition at condition creation time. Conditions have no concept of domains, and so can't validate a chain id for `mainnet` vs `lynx`.

Remember that chains -> provider urls are obtained from the `chainlist` repos on startup and can technically contain "any" chain id mapping entry. However, if in a condition, the user provides a chain id that the domain doesn't support (eg. `chainId: 12345678987654321`) on `lynx`, then we won't have a provider for that, but we won't know until condition evaluation time when we try to get the providers for the corresponding chain ID.

nucypher/blockchain/eth/utils.py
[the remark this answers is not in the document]
derekpierre: ^ Just to add to what @theref said, it would involve doing health checks on all of these URLs - https://github.com/nucypher/chainlist/blob/main/lynx.json, which seems tedious and excessive for `lynx` which seems excessive.

If we expand `mainnet` to support "any" evm chain, it's something we have to consider possibly not doing. Given that we fallback to the next RPC URL in the list of URLs of a previous endpoint fails, the health check may be excessive at that point. We can adjust the list in realtime as certain URLs work so that the list is biased over time...?

nucypher/blockchain/eth/actors.py
```diff
@@ -265,39 +264,39 @@ def set_provider_public_key(self) -> Union[TxReceipt, None]:
...
-                    f"Chain ID {chain_id} is not supported for condition evaluation by this operator."
+        for chain_id, endpoint_list in user_defined_endpoints.items():
+            if not endpoint_list:
+                raise self.ActorError(
+                    f"No mandatory user endpoint configured for chain {chain_id}"
                 )
-
             # connect to each endpoint and check that they are on the correct chain
-            for uri in endpoints:
+            for uri in endpoint_list:
```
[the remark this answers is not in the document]
derekpierre: ✔️

nucypher/policy/conditions/evm.py
derekpierre: Fixed name of `user_configured_endpoints` variable, and associated log/exception messages - https://github.com/nucypher/nucypher/pull/3569/commits/9494ede1716686e3def775d069d66e82c93c8e64

KPrasch (approved):

nucypher/blockchain/eth/utils.py
```diff
@@ -133,17 +135,17 @@ def rpc_endpoint_health_check(endpoint: str, max_drift_seconds: int = 60) -> boo
     return True  # finally!
 
 
-def get_default_rpc_endpoints() -> Dict[int, List[str]]:
+@cache
+def get_default_rpc_endpoints(domain: TACoDomain) -> Dict[int, List[str]]:
     """
     Fetches the default RPC endpoints for various chains
```
cygnusv, line 141:
```suggestion
    For a given domain, fetches the default RPC endpoints for various chains
```

cygnusv (approved):

newsfragments/3569.feature.rst
```diff
@@ -0,0 +1 @@
+Added plumbing to support EVM condition evaluation on "any" (major) EVM chain outside of Ethereum and Polygon - only enabled on `lynx` testnet for now.
```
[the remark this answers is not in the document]
derekpierre: ✔️

nucypher/blockchain/eth/utils.py
derekpierre: ✔️

Issue #3567, opened 2024-12-03
JWT condition type


cygnusv:
This is just a draft idea for a new type of conditions, `JWTCondition`, aimed at describing a set of constrains that must be met by a JWT signature presented by the requester. How this signature/token is produced or obtained is outside of the scope of TACo, but usually they will be obtained by the requester as part of the business logic of their use case, normally through a web2 service. 

Here are some high-level examples of constrains on the JWT token that TACo can use :
* Issuer must be `0xabcd`.
* Signature must have been issued at most `24 hours` ago.
* Expiration time must be at least `30 minutes` away.
* Subject must be `:userAddress`.

An example of a JWT condition object:

```js
{
    conditionType: 'jwt',
    issuer: '0xacbd',
    subject: ':userAddress',
    expiration_window: 1800,   // wrt to blocktime
    issued_window: 86400    // at most 86400 seconds before now
}
```

I initially started drafting this example JWT condition with `ReturnValueTests` for each field, but then I realized this could be misused easily. In practice:
* We want equality checks for identities, like "`Issuer`" or "`Subject`", so we should model them as simple key-value fields, where the value is a string. Note that these can be context parameters, e.g. "`:userAddress`".
* For time-related fields, we want to be able to express validity constrains for JWT tokens, not arbitrary time relations. In the case of "`expiration`", I think it only makes sense to compare the JWT expiration field wrt to current local/block time, so the only thing left to express is how much of a leeway you want to allow as part of the condition, hence why I named the field `expiration_window` and only allow a numeric value.
* For the "`issued` at" field in JWTs, things are a bit more tricky. I see at least two potential constrain expressions:
    * `issued_after`: Issued after a specific timestamp (the value).
    * `issued_window`: Issued at most `X` seconds before current time, where `X` is the value.

It's very likely that, depending on the use case, there will be domain-specific claim fields that will be part of the JWTCondition, so we should be able to allow such fields. For example, an e-book distribution platform can use JWTs with standard fields (like issuer, expiration, etc), plus ad-hoc fields like `ebook_id`, `device_id`, `region`, etc.

Some references on JWT and JWT claims:
* https://jwt.io/introduction
* JWT spec: https://datatracker.ietf.org/doc/html/rfc7519
* JWT Claims reference: https://www.iana.org/assignments/jwt/jwt.xhtml

derekpierre:
[...]

PR #3570, opened 2024-12-12
Support for JWTConditions

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
This PR introduces an experimental new type of conditions that can be satisfied by the requester providing an acceptable JWT token. What's "acceptable" depends on what's defined on the condition, and for the moment is restricted to:
(1) the issuer identity,
(2) its corresponding public key,
(3) and the token must not be expired.

**Issues fixed/closed:**
- Closes https://github.com/nucypher/nucypher/issues/3567

**Why it's needed:**
This is a simple type of condition that can facilitate integrations with existing authentication and authorization web2 services. Given that the JWT standard is very flexible and supports defining custom types of claims, there's lots of use cases that can be created with JWT conditions, e.g. DRM platforms, content distribution, identity management, etc. 

A small side note regarding centralization vs decentralization: in the web2 world, the entity issuing JWT tokens will be centralized is most cases; this is of course not a limitation for TACo, but it's worth noting. Conversely, in a "pure" web3 setting, TACo can accept JWT tokens issued by a decentralized entity, assuming they use a threshold digital signature, e.g, threshold ECDSA. This is to say that, from the verification perspective, TACo is purely agnostic to the token issuing setting.

**Notes for reviewers:**
Initially, the original issue https://github.com/nucypher/nucypher/issues/3567 mentioned potential additional restrictions that could be defined in the condition. Along the way, I decided to drop them to simplify the PR, given that we don't actually have any validation on this, and to be honest, adding this is very simple once we have clarity on them.

Note there's a sister PR in taco-web: https://github.com/nucypher/taco-web/pull/604


nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,168 @@
...
+            public_key=public_key,
+            expected_issuer=expected_issuer,
+            # subject=subject,
+            # expiration=expiration_window,
+            # issued_window=issued_window,
+            condition_type=condition_type,
+            name=name,
+            return_value_test=ReturnValueTest(
+                comparator="==", value=True
+            ),  # TODO: Workaround for now
```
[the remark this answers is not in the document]
cygnusv: Not necessarily the RTV evaluation logic itself, but the fact that RTVs are mandatory for ExecutionCalls is what I wanted to challenge, particularly for the case of execution calls that directly return a boolean value (this can also happen for offchain calls, for example). Maybe it's a very niche case, so I'm fine with RTVs even for this case.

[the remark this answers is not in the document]
derekpierre: Just to clarify one thing. 

An `ExecutionCall` does not have an RTV but `ExecutionCallAccessControlCondition` does i.e. the wrapping of the call into a condition. Remember, a call is execution, a condition is value checking. 

If the evaluation of the underlying execution call is always a boolean as opposed to some value that you need to check, then `JWTCondition` can simply extend `AccessControlCondition` instead and still use the execution call internally. Alternatively, we could:
- add a further abstraction e.g. split the `ExecutionCallAccessControlCondtion` into two, one that uses RTV (underlying call returns a value that needs to be compared) and another that doesn't (the underlying call returns a boolean) 
OR 
- make `return_value_test` optional for the `ExecutionCallAccessControlCondition` (although this may involve needing to override the schema when RTV is required)...need to think more on this.

Side Note: Originally, the thought was that `ExecutionCall`s could be serialized and used on its own without a condition hence they had their own schema, and we ended up not doing that for SequentialConditions, but kept the absraction. One thought I have had more recently is that `ExecutionCall` should not define a Schema at all and just be functional; then subclassing of calls is easier, and conditions can still use them. This is coming up in the Non-Evm JSON RPC work...happy to chat more about it in voice.

dev-requirements.txt
```diff
@@ -48,46 +48,46 @@ eth-rlp==1.0.1 ; python_version >= "3.9" and python_version < "4"
...
-eval-type-backport==0.2.0 ; python_version >= "3.9" and python_version < "3.10"
-evm-trace==0.2.3 ; python_version >= "3.9" and python_version < "4"
+ethpm-types==0.6.19 ; python_version >= "3.9" and python_version < "4"
+eval-type-backport==0.2.2 ; python_version >= "3.9" and python_version < "3.10"
+evm-trace==0.2.4 ; python_version >= "3.9" and python_version < "4"
 evmchains==0.0.13 ; python_version >= "3.9" and python_version < "4"
 exceptiongroup==1.2.2 ; python_version >= "3.9" and python_version < "3.11"
 executing==2.1.0 ; python_version >= "3.9" and python_version < "4"
 filelock==3.16.1 ; python_version >= "3.9" and python_version < "4"
-flask==3.0.3 ; python_version >= "3.9" and python_version < "4"
```
theref, line 58: should all of these packages be changing version?

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,143 @@
...
+
+    _valid_jwt_algorithms = (
+        "ES256",
+        "RS256",
+    )  # https://datatracker.ietf.org/doc/html/rfc7518#section-3.1
+
+    class Schema(ExecutionCall.Schema):
+        jwt_token = fields.Str(required=True)
+        # TODO: See #3572 for a discussion about deprecating this in favour of the expected issuer
+        public_key = fields.Str(
```
theref, line 29: let's add some validation:

```python
        @validates("public_key")
        def validate_public_key(self, value):
            if not value:
                raise ValidationError("Public key cannot be empty")

            # Skip validation if it's a context variable
            if is_context_variable(value):
                return value

            # Check PEM format
            if not (value.startswith("-----BEGIN PUBLIC KEY-----") and 
                   value.endswith("-----END PUBLIC KEY-----")):
                raise ValidationError("Public key must be in PEM format with proper header/footer")

            try:
                from cryptography.hazmat.primitives.serialization import load_pem_public_key
                from cryptography.hazmat.backends import default_backend
                load_pem_public_key(value.encode(), backend=default_backend())
            except Exception as e:
                raise ValidationError(f"Invalid public key format: {str(e)}")

            return value
  ```

File: tests/unit/conditions/test_jwt_condition.py [hunk 4]
```diff
@@ -0,0 +1,146 @@
...
+    condition = JWTCondition(
+        jwt_token=":aContextVariableForJWTs",
+        public_key=TEST_ECDSA_PUBLIC_KEY,
+    )
+
+    assert condition.jwt_token == ":aContextVariableForJWTs"
+    assert condition.public_key == TEST_ECDSA_PUBLIC_KEY
+    assert condition.condition_type == JWTCondition.CONDITION_TYPE
+
+
```
theref, line 94: maybe add a test for an invalid public key:
```python
    # Invalid public key format
    malformed_key = "NOT-A-VALID-KEY"
    with pytest.raises(jwt.exceptions.InvalidKeyError):
        jwt.decode(token, malformed_key, algorithms=["ES256"])
```

tests/unit/conditions/test_jwt_condition.py
```diff
@@ -0,0 +1,146 @@
...
+TEST_ECDSA_PUBLIC_KEY = (
+    "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEXHVxB7s5SR7I9cWwry"
+    "/JkECIReka\nCwG3uOLCYbw5gVzn4dRmwMyYUJFcQWuFSfECRK+uQOOXD0YSEucBq0p5tA==\n-----END PUBLIC "
+    "KEY-----"
+)
+
+ISSUED_AT = datetime.now(tz=timezone.utc)
+
+TEST_JWT_TOKEN = jwt.encode(
```
theref, line 31: should we test multiple algorithms?

theref (dismissed):

dev-requirements.txt
cygnusv: 🤷 that's the output of the requirements locking script. To be honest, I had to tweak the result since it removed a dependency from atxm, so at this point, I'm not sure we can fully trust the script. Regardless, I think that's something we can revisit outside this PR.

tests/unit/conditions/test_jwt_condition.py
cygnusv: Unsure. I mean, we can, but at the same time, I'm not sure to what extent we're just testing the underlying library.

theref: yeah that's very fair

tests/unit/conditions/test_jwt_condition.py
```diff
@@ -0,0 +1,146 @@
...
+        jwt_token=":anotherContextVariableForJWTs",
+        public_key=TEST_ECDSA_PUBLIC_KEY,
+        expected_issuer="Isabel",
+    )
+
+    context = {":anotherContextVariableForJWTs": token}
+    success, result = condition.verify(**context)
+    assert success
+    assert result == {"iss": "Isabel"}
+
```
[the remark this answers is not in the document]
cygnusv: Same as above, the main reason I didn't do it is that it felt to me we'd be mostly testing the underlying library. But I recognize I was 50/50 on this, tbh, so let's go with it 👍

nucypher/policy/conditions/jwt.py
cygnusv: Fair enough, let me see what does the library actually requires, and I'll add a relevant check.

tests/unit/conditions/test_jwt_condition.py
```diff
@@ -0,0 +1,146 @@
...
+def jwt_token(with_iat: bool = True, claims: dict = None):
+    claims = claims or dict()
+    if with_iat:
+        claims["iat"] = ISSUED_AT
+
+    return jwt.encode(claims, TEST_ECDSA_PRIVATE_KEY, algorithm="ES256")
+
+
+class TestJWTVerificationCall(JWTVerificationCall):
```
derekpierre, line 44: How come we need this test call class? Is that because you are directly testing the call as opposed to having the condition execute the call?

tests/unit/conditions/test_jwt_condition.py
```diff
@@ -0,0 +1,146 @@
...
+        jwt_token=token, public_key=TEST_ECDSA_PUBLIC_KEY, expected_issuer="Isabel"
+    )
+    payload = call.execute()
+    assert payload == {"iss": "Isabel"}
+
+
+def test_jwt_condition_initialization():
+    condition = JWTCondition(
+        jwt_token=":aContextVariableForJWTs",
+        public_key=TEST_ECDSA_PUBLIC_KEY,
```
derekpierre, line 87: Comment from the `taco-web` side that was meant for here - https://github.com/nucypher/taco-web/pull/604#discussion_r1887082118

```
Is it possible to provide the public key without the PEM header and 
footer i.e. just use the base64, or hex if that is more familiar/acceptable? 
On the server-side we can do the relevant format adjustment if 
needed: convert it to PEM, OR not bother if the 
library allows base64/hex format, or ...

(cc @cygnusv )
```

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,143 @@
...
+
+    def execute(self, **context) -> Any:
+        jwt_token = resolve_any_context_variables(self.jwt_token, **context)
+
+        require = []
+        if self.expected_issuer:
+            require.append("iss")
+
+        try:
+            payload = jwt.decode(
```
derekpierre, line 68: Nice!

tests/unit/conditions/test_jwt_condition.py
cygnusv: ✅

tests/unit/conditions/test_jwt_condition.py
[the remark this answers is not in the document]
cygnusv: ✅

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,143 @@
...
+        def validate_jwt_token(self, value):
+            if value and not is_context_variable(value):
+                raise ValidationError(
+                    f"Invalid value for JWT token; expected a context variable, but got '{value}'"
+                )
+
+    def __init__(
+        self,
+        jwt_token: Optional[str] = None,
+        public_key: Optional[str] = None,
```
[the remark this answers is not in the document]
cygnusv: ✅

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,143 @@
...
+        def make(self, data, **kwargs):
+            return JWTCondition(**data)
+
+    def __init__(
+        self,
+        condition_type: str = ConditionType.JWT.value,
+        name: Optional[str] = None,
+        jwt_token: Optional[str] = None,
+        public_key: Optional[str] = None,
+        expected_issuer: Optional[str] = None,
```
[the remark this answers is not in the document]
cygnusv: ✅

tests/unit/conditions/test_jwt_condition.py
cygnusv: Yes, I guess a different way could have been to mock the validation method. At that time, I just found this more natural.

tests/unit/conditions/test_jwt_condition.py
cygnusv: What's the concern here? Space?

tests/unit/conditions/test_jwt_condition.py
derekpierre: Defintely not an issue; just curious. 

Basically, the condition executes the underlying call, and then compares the result with a return value test. So calling `verify()` on the condition implicitly tests `execute()` on the call. So, you may end up with repeated tests if you test the call and condition separately.

tests/unit/conditions/test_jwt_condition.py
derekpierre: Conciseness, yes, and also a question of format. 

The PEM header and footer are constants and add bulk. Can we make it more concise?  Something like:
```json
{
    "conditionType": "jwt",
    "jwtToken":":jwtToken",
    "publicKey":"MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEXHVxB7s5SR7I9cWwry/JkECIRekaCwG3uOLCYbw5gVzn4dRmwMyYUJFcQWuFSfECRK+uQOOXD0YSEucBq0p5tA=="
}
```
instead of
```json
{
    "conditionType": "jwt",
    "jwtToken":":jwtToken",
    "publicKey":"-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEXHVxB7s5SR7I9cWwry/JkECIRekaCwG3uOLCYbw5gVzn4dRmwMyYUJFcQWuFSfECRK+uQOOXD0YSEucBq0p5tA==\n-----END PUBLIC KEY-----"
}
```

Also, is the full PEM format the most common format adopters would have for the public key in their applications when creating the condition i.e. is PEM the "best"/most-widely-used format to accept in the client-side API?  Presumably so?

derekpierre:
Let's also add a jwt condition to the lingo tests as well for testing going from json -> JWTCondition - see https://github.com/nucypher/nucypher/blob/v7.5.x/tests/unit/conditions/test_condition_lingo.py#L27. You can include it in the sequential condition or if then else condition.

tests/unit/conditions/test_jwt_condition.py
cygnusv: PEM headers don't always have the exact `"-----BEGIN PUBLIC KEY-----..."` starting substring. There's some variations depending on the algorithm and backend used to generate the public key. The content is already base64, so if we remove the header and tail we'd be saving up ~50 bytes, at the expense of losing the explicitness that the PEM header and tail already provides.

re: is PEM the most common format, I don't have a definite answer here, since it's very domain specific (i.e., blockchain projects would never use PEM, but consumer-oriented software normally uses it). There's other formats like JWKs (see https://github.com/nucypher/nucypher/issues/3572), which can be added along the road.

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,156 @@
...
+        try:
+            self.execution_call = JWTVerificationCall(
+                jwt_token=jwt_token,
+                public_key=public_key,
+                expected_issuer=expected_issuer,
+            )
+        except ExecutionCall.InvalidExecutionCall as e:
+            raise InvalidCondition(str(e)) from e
+
+        super().__init__(condition_type=condition_type, name=name)
```
derekpierre, line 139: Nice adjustment.

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,156 @@
...
+    @property
+    def public_key(self):
+        return self.execution_call.public_key
+
+    @property
+    def expected_issuer(self):
+        return self.execution_call.expected_issuer
+
+    def verify(self, **context) -> Tuple[bool, Any]:
+        payload = self.execution_call.execute(**context)
```
derekpierre, line 154: Seems like at the moment we only have 2 outcomes:
- a non-None payload which causes verify to return `True`
OR
- raise an exception in the failure case eg. expired jwt, invalid issuer, invalid audience etc. (https://github.com/jpadilla/pyjwt/blob/master/jwt/exceptions.py#L9) i.e. we never actually return `False` from verify. 

Of course, clearly we want to raise for unsupported algorithm, but do we want to always raise for a failed validation instead of possibly returning (`False`, `None`)? Just wondering if, aside from unsupported algorithm, for an `InvalidTokenError` (or subsets of it) we would ever want to return (`False`, `None`) from `verify`?

nucypher/policy/conditions/lingo.py
```diff
@@ -59,7 +59,7 @@ def _deserialize(self, value, attr, data, **kwargs):
         return instance
 
 
-# CONDITION = TIME | CONTRACT | RPC | JSON_API | COMPOUND | SEQUENTIAL | IF_THEN_ELSE_CONDITION
+# CONDITION = TIME | CONTRACT | RPC | JSON_API | JSON_RPC | JWT | COMPOUND | SEQUENTIAL | IF_THEN_ELSE_CONDITION
```
derekpierre, line 62: Thanks!

nucypher/policy/conditions/jwt.py
derekpierre: Thinking about this more, we **need** it to return `False` instead of exceptions for cases when the JWT verification fails for non-error cases eg. invalid issuer, invalid audience, expired etc., because logical conditions (`CompoundCondition`, `SequentialCondition`, `IfThenElseCondition`) would not function properly because exceptions are considered error cases, not verification failures. So exceptions cause the logical condition to immediately stop its execution and not continue based on logical operations that can move on despite an outcome of `False`. 

For example, think of an `IfThenElseCondition` where the JWT condition is used for the `if` part. If the `if` part fails, then the `else` should be executed, but because an exception is raised for failure, the entire condition stops and fails because the exception is interpreted as an underlying error case within the overall condition.

tests/unit/conditions/test_jwt_condition.py
derekpierre: Fair point!

nucypher/policy/conditions/jwt.py
derekpierre: Chatted with @cygnusv some more, and instead filed https://github.com/nucypher/nucypher/issues/3582, where the exception could optionally be considered as a condition failure as part of the logical condition itself.

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,156 @@
...
+                )
+
+        @validates("public_key")
+        def validate_public_key(self, value):
+            try:
+                public_key = load_pem_public_key(
+                    value.encode(), backend=default_backend()
+                )
+                if isinstance(public_key, rsa.RSAPublicKey):
+                    return value
```
derekpierre, line 55: This function is "void", so you don't actually have to return anything. Either it raises an exception (failed validation) or not (all is well).

```suggestion
                    return
```

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,156 @@
...
+                elif isinstance(public_key, ec.EllipticCurvePublicKey):
+                    curve = public_key.curve
+                    if curve.name != JWTVerificationCall.SECP_CURVE_FOR_ES256:
+                        raise ValidationError(
+                            f"Invalid EC public key curve: {curve.name}"
+                        )
+            except Exception as e:
+                raise ValidationError(f"Invalid public key format: {str(e)}")
+
+            return value
```
derekpierre, line 65: The return can be removed (see comment above)
```suggestion
                raise ValidationError(f"Invalid public key format: {str(e)}")

```

derekpierre (commented):
While python ignores types, we maintain condition typehints here - https://github.com/nucypher/nucypher/blob/v7.5.x/nucypher/policy/conditions/types.py#L126. Let's add `JWTCondition` as well - see `JsonApiCondition`/`JsonRpcCondition` as an example.

nucypher/policy/conditions/jwt.py
cygnusv: 2b048ea3ee847bb6a15494dae71a025ec744c3c2

nucypher/policy/conditions/jwt.py
cygnusv: 2b048ea3ee847bb6a15494dae71a025ec744c3c2

vzotova (approved):

tests/unit/conditions/test_jwt_condition.py
```diff
@@ -0,0 +1,236 @@
...
+
+TEST_ECDSA_PUBLIC_KEY = (
+    "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEXHVxB7s5SR7I9cWwry"
+    "/JkECIReka\nCwG3uOLCYbw5gVzn4dRmwMyYUJFcQWuFSfECRK+uQOOXD0YSEucBq0p5tA==\n-----END PUBLIC "
+    "KEY-----"
+)
+
+ISSUED_AT = calendar.timegm(datetime.now(tz=timezone.utc).utctimetuple())
+
+TEST_JWT_TOKEN = jwt.encode(
```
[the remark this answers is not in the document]
cygnusv: https://github.com/nucypher/nucypher/issues/3584

nucypher/policy/conditions/jwt.py
```diff
@@ -0,0 +1,154 @@
...
+class JWTVerificationCall(ExecutionCall):
+
+    _valid_jwt_algorithms = (
+        "ES256",
+        "RS256",
+    )  # https://datatracker.ietf.org/doc/html/rfc7518#section-3.1
+
+    SECP_CURVE_FOR_ES256 = "secp256r1"
+
+    class Schema(ExecutionCall.Schema):
```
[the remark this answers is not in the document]
cygnusv: https://github.com/nucypher/nucypher/issues/3584

nucypher/policy/conditions/types.py
```diff
@@ -75,6 +75,12 @@ class JsonRpcConditionDict(BaseExecConditionDict):
     authorizationToken: NotRequired[str]
 
 
+class JWTConditionDict(_AccessControlCondition):
+    jwt_token: str
+    public_key: str  # TODO: See #3572 for a discussion about deprecating this in favour of the expected issuer
+    expected_issuer: NotRequired[str]
```
derekpierre, line 81: This is for the JSON itself (before it gets converted to snake case) so it should be camelcase instead.

```suggestion
    jwtToken: str
    publicKey: str  # TODO: See #3572 for a discussion about deprecating this in favour of the expected issuer
    expectedIssuer: NotRequired[str]
```

derekpierre (approved):
@cygnusv See added suggestion about changing to camel case for the type.

Other than that, looks great! 🎸

PR #3571, opened 2024-12-16
JsonRpcCondition (JSON RPC 2.0 specification)

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

Based over #3569 .

- Allows condition based on JSON 2.0 RPC calls. It conforms to JSON RPC 2.0 spec - <https://www.jsonrpc.org/specification>.
- Refactors JSON API and RPC conditions into common package with common underlying code.

Screenshot 2025-01-02 at 9 29 28 AM

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


nucypher/policy/conditions/lingo.py
```diff
@@ -697,7 +698,8 @@ def resolve_condition_class(
         conditions expression framework.
         """
         from nucypher.policy.conditions.evm import ContractCondition, RPCCondition
-        from nucypher.policy.conditions.offchain import JsonApiCondition
+        from nucypher.policy.conditions.json.api import JsonApiCondition
+        from nucypher.policy.conditions.json.rpc import JsonRpcCondition
```
cygnusv, line 702: I'm wondering if we should have a higher-level module that re-exports all conditions, something like `from nucypher.policy.conditions import FooCondition, ...`. Speaking of which, I think at some moment we should drop the `policy` part from the import path. 
cc @KPrasch

nucypher/policy/conditions/json/rpc.py
```diff
@@ -0,0 +1,177 @@
...
+
+    def verify(self, **context) -> Tuple[bool, Any]:
+        result = self.execution_call.execute(**context)
+        result_for_eval = process_result_for_condition_eval(result)
+
+        resolved_return_value_test = self.return_value_test.with_resolved_context(
+            **context
+        )
+        eval_result = resolved_return_value_test.eval(result_for_eval)  # test
+        return eval_result, result
```
cygnusv, line 177: Code for `verify()` method seems identical to `JsonApiCondition.verify()`. I'm wondering if there's a way to generalize both. If I understand it correctly, on an abstract level, `JsonRpcCondition` is just a subcase of `JsonApiCondition`, even if it's not explicitly inheriting from it now.

nucypher/policy/conditions/json/rpc.py
```diff
@@ -0,0 +1,177 @@
...
+
+        # response contains a value for either "result" or "error"
+        error = data.get("error", None)
+        if error:
+            raise JsonRequestException(
+                f"JSON RPC Request failed with error in response: code={error['code']}, msg={error['message']}"
+            )
+
+        # obtain result first then perform query
+        result = data["result"]
```
cygnusv, line 74: Should we catch here the case for a malformed JSON-RPC response that doesn't have a `result` field? Otherwise, we're left with a mere `KeyError`.

nucypher/policy/conditions/json/rpc.py
```diff
@@ -0,0 +1,177 @@
...
+    ExecutionCallAccessControlCondition,
+    ReturnValueTest,
+)
+
+
+class BaseJsonRPCCall(JsonRequestCall, ABC):
+    class Schema(JsonRequestCall.Schema):
+        method = fields.Str(required=True)
+        params = fields.Field(required=False, allow_none=True)
+        query = JSONPathField(required=False, allow_none=True)
```
cygnusv, line 29: If I understand it correctly, queries for `JsonRpcCalls` will be parsed not from the json root, but from the `result` field, right? If so, this should be explicit somewhere.

cygnusv (approved):
Great work, just some minor comments.

nucypher/policy/conditions/json/utils.py
```diff
@@ -0,0 +1,17 @@
+from typing import Any
+
+
+def process_result_for_condition_eval(result: Any):
```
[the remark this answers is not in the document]
derekpierre: Currently, JSON conditions are the only ones that can possibly return strings from execution.

nucypher/policy/conditions/lingo.py
derekpierre: We can, but this is the Python API which we don't document / broadcast to adopters - so this would really be for our own benefit. Of course, we can reconfigure the packages in a different PR if you like.

nucypher/policy/conditions/json/rpc.py
derekpierre: Correct. When you say "explicit", what do you mean exactly? Something like `query_from_result` instead of `query` for the field name or something like that? I was hoping documentation would be sufficient.

nucypher/policy/conditions/json/rpc.py
derekpierre: Good catch - ✔️

nucypher/policy/conditions/json/rpc.py
derekpierre: ✔️ - it is indeed identical.

It's subtle, but JsonRpcCondition can't subclass JsonApiCondition. JsonApiCondition only allows HTTP GET, and JsonRpcCondition only allows HTTP POST. Also `JsonRpcCondition` takes `method` which `JsonApiCondition` doesn't.

Created a common base superclass instead (similar to what I did for the `JsonApiCall` and `JsonRpcCall`).

nucypher/policy/conditions/lingo.py
cygnusv: Don't worry, it's more an open question regarding our own modules structure.

KPrasch (approved):

theref (approved):

Issue #3503, opened 2024-05-16
Programatic usage of interface contract ABIs


KPrasch:
Adapt the contract agents internal API so it can be used with known interface ABIs that may not have a fixed or otherwise predefined contract address.

cygnusv:
This will be useful for instantiating templated contracts like `Subscriptions` and `IEncryptionAuthorizer`.

PR #3576, opened 2025-01-27
EIP1271 Support

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Support for [EIP-1271](https://ethereum.org/en/developers/tutorials/eip-1271-smart-contract-signatures/)
- Will require a corresponding `AuthProvider` implementation in `taco-web` - https://github.com/nucypher/taco-web/pull/617
- Allows EIP1271 auth signature to be used with the `:userAddress` context variable. This way either EIP4361 (for EOA wallets) or EIP1271 (for smart contract wallets) can be used with the `:userAddress` context variable. *(The initial implementation in this PR created a `:userAddressEIP1271` context variable but this was not correct, it would mean that only EIP1271 wallets could be used for condition, which doesn't really make sense - this was fixed and rebased over)*.

**Issues fixed/closed:**
> - Fixes #...

- Related to https://github.com/nucypher/sprints/issues/116

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -155,7 +155,7 @@ def _make_rest_app(this_node, log: Logger) -> Flask:
         # TODO: When non-evm chains are supported, bump the version.
         #  this can return a list of chain names or other verifiable identifiers.
 
-        payload = {"version": 1.0, "evm": list(this_node.condition_providers)}
+        payload = {"version": 1.0, "evm": list(this_node.condition_providers.providers)}
         return Response(json.dumps(payload), mimetype="application/json")
 
     @rest_app.route('/decrypt', methods=["POST"])
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -154,7 +154,7 @@ def _make_rest_app(this_node, log: Logger) -> Flask:
         """
         # TODO: When non-evm chains are supported, bump the version.
         #  this can return a list of chain names or other verifiable identifiers.
-        providers = this_node.condition_providers.providers
+        providers = this_node.condition_provider_manager.providers
         sorted_chain_ids = sorted(list(providers))
         payload = {"version": 1.0, "evm": sorted_chain_ids}
         return Response(json.dumps(payload), mimetype="application/json")
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -260,7 +260,7 @@ def _make_rest_app(this_node, log: Logger) -> Flask:
                 try:
                     evaluate_condition_lingo(
                         condition_lingo=condition_lingo,
-                        providers=this_node.condition_providers,
+                        providers=this_node.condition_provider_manager,
                         context=context,
                     )
                 except ConditionEvalError as error:
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -154,8 +154,9 @@ def _make_rest_app(this_node, log: Logger) -> Flask:
         """
         # TODO: When non-evm chains are supported, bump the version.
         #  this can return a list of chain names or other verifiable identifiers.
-
-        payload = {"version": 1.0, "evm": list(this_node.condition_providers.providers)}
+        providers = this_node.condition_providers.providers
+        sorted_chain_ids = sorted(list(providers))
+        payload = {"version": 1.0, "evm": sorted_chain_ids}
         return Response(json.dumps(payload), mimetype="application/json")
 
     @rest_app.route('/decrypt', methods=["POST"])
--- a/nucypher/policy/conditions/auth/evm.py
+++ b/nucypher/policy/conditions/auth/evm.py
@@ -1,16 +1,20 @@
 from enum import Enum
-from typing import List
+from typing import List, Optional
 
 import maya
 from eth_account.account import Account
 from eth_account.messages import HexBytes, encode_typed_data
+from eth_typing import ChecksumAddress
 from siwe import SiweMessage, VerificationError
+
+from nucypher.policy.conditions.utils import ConditionProviderManager
 
 
 class EvmAuth:
     class AuthScheme(Enum):
         EIP712 = "EIP712"
         EIP4361 = "EIP4361"
+        EIP1271 = "EIP1271"
 
         @classmethod
         def values(cls) -> List[str]:
@@ -26,7 +30,13 @@ class EvmAuth:
         """The message is too old."""
 
     @classmethod
-    def authenticate(cls, data, signature, expected_address):
+    def authenticate(
+        cls,
+        data,
+        signature: str,
+        expected_address: str,
+        providers: Optional[ConditionProviderManager] = None,
+    ):
         raise NotImplementedError
 
     @classmethod
@@ -35,13 +45,21 @@ class EvmAuth:
             return EIP712Auth
         elif scheme == cls.AuthScheme.EIP4361.value:
             return EIP4361Auth
+        elif scheme == cls.AuthScheme.EIP1271.value:
+            return EIP1271Auth
 
         raise ValueError(f"Invalid authentication scheme: {scheme}")
 
 
 class EIP712Auth(EvmAuth):
     @classmethod
-    def authenticate(cls, data, signature, expected_address):
+    def authenticate(
+        cls,
+        data,
+        signature: str,
+        expected_address: str,
+        providers: Optional[ConditionProviderManager] = None,
+    ):
         try:
             # convert hex data for byte fields - bytes are expected by underlying library
             # 1. salt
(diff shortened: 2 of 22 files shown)


nucypher/network/server.py
```diff
@@ -154,8 +154,9 @@ def condition_chains():
         """
         # TODO: When non-evm chains are supported, bump the version.
         #  this can return a list of chain names or other verifiable identifiers.
-
-        payload = {"version": 1.0, "evm": list(this_node.condition_providers)}
+        providers = this_node.condition_providers.providers
```
vzotova, line 157: probably dumb question, should it be `condition_provider_manager` instead of `condition_providers`?

vzotova (approved):
Looks good to me, just one small question:

nucypher/network/server.py
derekpierre: ✔️  - Not dumb at all. In fact, great catch 🎸 . That was a bug from a rename, but the mocked integration test still passed. Fixed now.

theref (dismissed):

vzotova (approved):

derekpierre:
Q: In the auth-signature format outlined here - <https://github.com/nucypher/sprints/issues/116#issuecomment-2610990171>. Should we also include `signerAddress` in anticipation of an improvement to accommodate ERC-6492 - https://github.com/nucypher/sprints/issues/116#issuecomment-2640530560? Perhaps something to consider in a follow-up.

theref:
> Q: In the auth-signature format outlined here - [nucypher/sprints#116 (comment)](https://github.com/nucypher/sprints/issues/116#issuecomment-2610990171). Should we also include `signerAddress` in anticipation of an improvement to accommodate ERC-6492 - [nucypher/sprints#116 (comment)](https://github.com/nucypher/sprints/issues/116#issuecomment-2640530560)?
> 
> Likely better to consider in a follow-up.

I can't think of any reason not to. Are there any times when providing the `signerAddress` wouldn't be possible?

nucypher/policy/conditions/auth/evm.py
```diff
@@ -1,16 +1,22 @@
...
+from nucypher.policy.conditions.exceptions import NoConnectionToChain
+from nucypher.policy.conditions.utils import ConditionProviderManager
+from nucypher.utilities.logging import Logger
+
 
 class EvmAuth:
     class AuthScheme(Enum):
         EIP712 = "EIP712"
         EIP4361 = "EIP4361"
+        EIP1271 = "EIP1271"
```
cygnusv, line 19: Keep 'em coming

nucypher/policy/conditions/auth/evm.py
```diff
@@ -106,3 +132,113 @@ def authenticate(cls, data, signature, expected_address):
             raise cls.AuthenticationFailed(
                 f"Invalid EIP4361 signature; signature not valid for expected address, {expected_address}"
             )
+
+
+class EIP1271Auth(EvmAuth):
+    EIP1271_ABI = """[
```
cygnusv, line 138: Not por this PR, but at some moment we should create a more general organization to (1) store ABIs and (2) instantiate `Contract` entities from them (see https://github.com/nucypher/nucypher/issues/3503)

nucypher/policy/conditions/auth/evm.py
```diff
@@ -106,3 +132,113 @@ def authenticate(cls, data, signature, expected_address):
...
+                # Interact with the EIP1271 contract
+                eip1271_contract = web3_instance.eth.contract(
+                    address=expected_address, abi=cls.EIP1271_ABI
+                )
+                result = eip1271_contract.functions.isValidSignature(
+                    data_hash,
+                    signature_bytes,
+                ).call()
+                if result == cls.MAGIC_VALUE_BYTES:
+                    return  # Successful authentication
```
cygnusv, line 196: I think you should consider also the negative case (`result != MAGIC_BYTES`) as equally important. Current behavior is implicitly considering this case less important, and keeps "looking for a positive" in other endpoints. Authentication is only considered negative if _all_ endpoints don't consider the signature valid, while for positive authentication, the _any_ positive result is enough. 

I think the ´else´ clause should be added here, so the first negative case should raise `AuthenticationFailed`.

cygnusv (commented):
Great work, looking forward to see it merged. There's one comment regarding the behavior on negative authentications that I think should be addressed, though.

nucypher/policy/conditions/auth/evm.py
derekpierre: It's a bit nuanced.

The big issue is that we have to continue trying other providers if we weren't able to make the contract function call. However, if we do make the contract call, and get a result (positive or negative) then we shouldn't try any subsequent providers. In the positive case, we return, but in the negative case we `break` out of the for loop (L198) and then raise an exception on L209. This `break` ensures that we don't try any subsequent providers assuming we were indeed able to call the contract function. Consequently, the first negative case does indeed raise an `AuthenticationFailed` exception.

You may ask well, why not just raise the `AuthenticationFailed` exception instead of breaking out of the for-loop? Well, then we would need an addition except in the for loop specifically for the `AuthenticationFailed` exception so that it can be re-raised and not caught by the catch-all exception on L199, which IMHO was more complex.

nucypher/policy/conditions/utils.py
```diff
@@ -22,6 +25,59 @@
...
+            try:
+                w3 = self._configure_w3(provider=provider)
+                self._check_chain_id(chain_id, w3)
+                yield w3
+                iterator_returned_at_least_one = True
+            except InvalidConnectionToChain as e:
+                # don't expect to happen but must account
+                # for any misconfigurations of public endpoints
+                self.logger.warn(
+                    f"Invalid blockchain connection; expected chain ID {e.expected_chain}, but detected {e.actual_chain}"
```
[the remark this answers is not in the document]
derekpierre: Totally - ✔️

derekpierre:
> > Q: In the auth-signature format outlined here - [nucypher/sprints#116 (comment)](https://github.com/nucypher/sprints/issues/116#issuecomment-2610990171). Should we also include `signerAddress` in anticipation of an improvement to accommodate ERC-6492 - [nucypher/sprints#116 (comment)](https://github.com/nucypher/sprints/issues/116#issuecomment-2640530560)?
> > Likely better to consider in a follow-up.
> 
> I can't think of any reason not to. Are there any times when providing the `signerAddress` wouldn't be possible?

Truthfully, it's less about whether it's possible to add (which I agree I think it would be easy to), but rather about whether we should. 

It adds an entry to the AuthSignature `typedData`, that isn't necessary for EIP1271, but the user will need to provide. The `signerAddress` isn't needed for anything since all the verification does `isValidSignature(dataHash, signature)` on the `address` in the AuthSignature (i.e. contract address) and we replace `:userAddress` with the contract address. The logic, for `nucypher` at the moment, doesn't really care about what EOA etc. actually generated the signature to send to the contract for EIP1271 verification - the contract cares but we don't. 

For the potential use of EIP-6492, I don't know enough about it as yet to understand if:
1. we should support it
2. it can fit into the same case for EIP1271, and from a logic standpoint it might be better for us to just create an additional authentication scheme for EIP6492 instead of retrofitting it into EIP1271 i.e. a separate EIP6492 AuthSignature (`taco-web`) and authentication object for verification (`nucypher`).

I'm all ears if you are more familiar with EIP6492 and we can chat more about it, but without more research/information it feels like it could be good to punt for now and include in our subsequent work for Account Abstraction.

(cc @arjunhassard , @cygnusv )

manumonti (approved):
Wow, great work here 🚀

nucypher/policy/conditions/auth/evm.py
cygnusv: Fair enough! The logic is a bit tricky because of the potential exceptions flow but you sold it well. I agree that other approaches can be more verbose and potentially more complex.

cygnusv (approved):

tests/acceptance/conditions/test_conditions.py
```diff
@@ -926,3 +932,56 @@ def test_json_rpc_condition_non_evm_prototyping_example():
     )
     success, _ = condition.verify()
     assert success
+
+
+def test_rpc_condition_using_eip1271(
+    testerchain, deployer_account, eip1271_contract_wallet, condition_providers
```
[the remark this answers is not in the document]
manumonti: 👆 not relevant at all. Just in case we want a cleaner code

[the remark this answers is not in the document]
derekpierre: ✔️

PR #3585, opened 2025-03-21
Handle BigInts provided in JSON from `taco-web`

derekpierre:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [x] 3

**What this does:**

Based over #3586 .

Original issue - https://github.com/nucypher/taco-web/issues/634.

`taco-web` does not handle large numbers, and as such must use `BigInt`. However `BigInt` cannot be serialized to a number in JSON, so instead it is passed as a string `"12312312....1231233132n"` (format established in https://github.com/nucypher/taco-web/pull/637), and as such nodes must handle receiving a number like that. Python does not have the same issue, but must accomodate numbers received from `taco-web` in that format.

Moving forward we have to be careful about any conditions that accept integer values that can be large, since `taco-web` could pass those values as serialized BigInts. I don't think this is an issue for any existing conditions (eg. chain-id is an int, but chain ids are not that large), but something we need to keep an eye on moving forward.

**Issues fixed/closed:**
> - Fixes #...

Related to https://github.com/nucypher/taco-web/pull/637.

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/newsfragments/3585.feature.rst
+++ b/newsfragments/3585.feature.rst
@@ -1,0 +1,1 @@
+Allow BigInt values from ``taco-web`` typescript library to be provided as strings.
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -11,7 +11,10 @@ from nucypher.policy.conditions.exceptions import (
     InvalidContextVariableData,
     RequiredContextVariable,
 )
-from nucypher.policy.conditions.utils import ConditionProviderManager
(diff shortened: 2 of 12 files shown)


vzotova (approved):

KPrasch (approved):

manumonti (approved):
LGTM! ✌️

PR #3591, opened 2025-04-24
Initial work for POC

derekpierre:
**Type of PR:**
Feature

**Required reviews:** 
2

**What this does:**
- `SigningCoordinatorAgent`
- `SigningRitualTracker`
- Signing ritual participation
- `/sign` endpoint

**Issues fixed/closed:**
...

**Why it's needed:**
...

**Notes for reviewers:**
...


theref (commented):
Ok, so the idea is to have all nodes post an initial signature upon Signing Cohort creation. This acts on-chain verifiability of success/acknowledgement. Can be used in rewards/slashing.

Does failure to submit a tx make the multisig invalid?

nucypher/characters/lawful.py
```diff
@@ -1284,6 +1288,10 @@ def handle_threshold_decryption_request(
         )
         return encrypted_response
 
+    def handle_threshold_signature_request(
```
theref:
```suggestion
    def handle_1271_signature_request(
```
We may implement true threshold signatures in the future

nucypher/crypto/powers.py
```diff
@@ -192,12 +192,16 @@ def unlock(self, password: str = None, duration: int = None) -> bool:
                                              duration=duration)
         return result
 
-    def sign_message(self, message: bytes) -> bytes:
+    # TODO: this is only a workaround - what are we going to do with this? standardize vs not standardize?
```
theref, line 195: not sure I understand the problem, could you explain?

nucypher/network/server.py
```diff
@@ -311,4 +312,46 @@ def status():
             return Response(response=html_error, headers=headers, status=HTTPStatus.INTERNAL_SERVER_ERROR)
         return Response(response=content, headers=headers)
 
+    @rest_app.route("/sign", methods=["POST"])
+    def threshold_sign():
```
theref, line 318:
```suggestion
    def 1271_sign():
```

tests/acceptance/conditions/test_signing_cohort.py
```diff
@@ -0,0 +1,513 @@
...
+        data=encoded_deposit_function,
+    )
+
+    return _multisig_contract_wallet
+
+
+def test_simple_data_message_signing(
+    multisig_contract_wallet, accounts, signing_cohort, deployer_account
+):
+    data = "Labor omnia vincit improbus."
```
theref, line 123: "Labor conquers all, wicked man." - of course.

tests/acceptance/conditions/test_signing_cohort.py
```diff
@@ -0,0 +1,513 @@
...
+
+    aggregated_signature = b"".join(received_signatures)
+
+    # should fail
+    with pytest.raises(ContractLogicError):
+        multisig_contract_wallet.execute(
+            receiver.address,
+            eth_amount,
+            b"",
+            os.urandom(len(aggregated_signature)),
```
theref, line 191: it should fail because of this line?

tests/acceptance/conditions/test_signing_cohort.py
```diff
@@ -0,0 +1,513 @@
...
+        == EIP1271Auth.MAGIC_VALUE_BYTES
+    )
+    assert multisig_contract_wallet.getSigners() == [
+        ursula.operator_address for ursula in signing_cohort
+    ]
+
+    # handover to new cohort
+    new_cohort = ursulas[COHORT_SIZE : COHORT_SIZE * 2]
+    for i, new_signer in enumerate(new_cohort):
+        multisig_contract_wallet.replaceSigner(
```
theref, line 308: love this

nucypher/characters/lawful.py
derekpierre: I suspect that the relevant logic will be dependent on the signing request blob received. We can always differentiate the two using, the contract metadata (info based on cohortId),  the `SigningRequest` object that is provided OR have a separate endpoint.

In any case, this was just starter code that I chatted with @KPrasch about, he's focused on this area and probably has his own changes. (ps @KPrasch you are welcome to add your changes to this branch, up to you)

nucypher/network/server.py
derekpierre: See comment in lawful.py

nucypher/crypto/powers.py
derekpierre: When a signature is produced using `eth_account`, the recovery ID (`v` component) includes a chain ID offset, due to EIP-155 which basically introduces a replay protection mechanism by incorporating the chain ID into the transaction signing process.

This is fine for Ethereum, but if you're passing the signature to another environment (eg. Rust) it needs to be modified to:
- 0 or 1 (if using ECDSA_recover-style APIs)
- 27 or 28 (if using Solidity `ecrecover` directly)

By default our `sign_message` modified the signature to use standard (0 or 1) because the only thing being signed until now was protocol stuff (not sent to contracts), but since we are now sending this signature to a smart contract it needs to not be.

This change ensures that the default standardization continues to occur unless standardize is set to `False` (eg. for posting the signature). I left the TODO because I wasn't sure if this was the right/best way to work around this.

tests/acceptance/conditions/test_signing_cohort.py
derekpierre: Yep, the signature is invalid for the contract wallet since not produced by the signers. It just produces garbage of the same aggregated length.

To make the test even more potent I can add a case where (t-1)signatures in the aggregated are correct, but the last one is wrong - maybe I'll add that later.

derekpierre:
> Ok, so the idea is to have all nodes post an initial signature upon Signing Cohort creation. This acts on-chain verifiability of success/acknowledgement. Can be used in rewards/slashing.

Correct - it's very analogous to the post transcript for DKGs. Felt like we can't just assign nodes to a signing cohort, without some acknowledgement by them and them proving that they can actually sign with their operator address first. Of course "what" they sign can be modified - just needed something to sign.

> Does failure to submit a tx make the multisig invalid?

The multisig will not be deployed without all members of the cohort posting a signature - https://github.com/nucypher/nucypher-contracts/pull/366/files#diff-f52286060c24f211bcca37ac36fc0fad47b4b8b5680999d955b2ec9fca892742R216

KPrasch:
rebased @ 983d6bcfc7607b59b8529900670d2b4616fac6d4

derekpierre:
@KPrasch since this PR is getting rather large `+8,809 −4,775`, now that tests are passing, wdyt about merging this PR into the `signing` epic branch (changed the base branch of this PR)?

I think we have a good base of code to merge, and then we can have subsequent iterative PRs on top of it.

derekpierre:
@KPrasch , @theref let's review/merge this initial work. The PR merges into the `signing` epic. We can merge subsequent changes into the `signing` epic for improvements/adjustments.

This PR alone is `+8,809 −4,775`

Issue #3600, opened 2025-06-03
feat: Implement Wallet Allowlist Condition for Node-Side Validation


Muhammad-Altabba:
#### Description
This feature introduces a new access control condition, [WalletAllowlistCondition](cci:2://file:///Users/funcy/repos/taco/taco-web/packages/taco/src/conditions/schemas/wallet-allowlist.ts:33:0-35:2), enabling decryption permissions based on an allowlist of Ethereum wallet addresses. Users prove ownership of an address in the allowlist by signing a specific message. This condition is evaluated by NuCypher nodes during decryption requests.

#### Motivation
The [WalletAllowlistCondition](cci:2://file:///Users/funcy/repos/taco/taco-web/packages/taco/src/conditions/schemas/wallet-allowlist.ts:33:0-35:2) provides a gas-free and lightweight method for managing decryption permissions. It leverages standard wallet signature verification, offering an alternative to on-chain mechanisms like NFT ownership or smart contract interactions. This allows for more flexible and cost-effective access control scenarios.

#### Node Responsibilities
- NuCypher nodes will receive policies containing this condition type.
- Upon a decryption request, nodes will evaluate the [WalletAllowlistCondition](cci:2://file:///Users/funcy/repos/taco/taco-web/packages/taco/src/conditions/schemas/wallet-allowlist.ts:33:0-35:2) by performing the signature verification against the provided context and the stored allowlist.
- If the condition is met, the node will provide its share of decryption material (re-encryption).

#### Benefits
- **Gas-Free:** No blockchain transactions are required for authentication or permission updates (beyond initial policy creation).
- **Off-Chain Management:** Permissions are managed within the policy definition itself.
- **Standard Cryptography:** Similar to other wallet-related conditions, it relies on well-established Ethereum message signing (EIP-191).

#### Related Client-Side Implementation
[...]

PR #3599, opened 2025-05-30
[EPIC] Signing Phase 1

derekpierre:
**Type of PR:**
Feature

**What this does:**
- [x] #3591 
- [x] #3596 
- [x] #3595 
- [x] #3597 
- [x] #3598 
- [x] #3604 
- [x] #3602 
- [x] #3606 
- [x] #3609 
- [x] #3614 
- [x] #3617 
- [x] #3620 
- [x] #3618
- [x] #3622 
- [x] #3624

PR #3601, opened 2025-06-04
Implement AddressAllowlistCondition for access control

Muhammad-Altabba:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Closes https://github.com/nucypher/nucypher/issues/3600

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/.github/workflows/darker.yml
+++ b/.github/workflows/darker.yml
@@ -17,7 +17,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.12'
+          python-version: "3.12"
       - uses: akaihola/darker@1.7.2
         with:
-          version: "1.7.2"
+          version: "2.1.1" # This will use the 2.1.1 darker Python package while using the 1.7.2 darker github action
--- a/.github/workflows/darker.yml
+++ b/.github/workflows/darker.yml
@@ -18,6 +18,6 @@ jobs:
       - uses: actions/setup-python@v4
         with:
           python-version: "3.12"
-      - uses: akaihola/darker@1.7.2
+      - uses: akaihola/darker@v2.1.1
         with:
           version: "2.1.1" # This will use the 2.1.1 darker Python package while using the 1.7.2 darker github action
--- a/.github/workflows/darker.yml
+++ b/.github/workflows/darker.yml
@@ -20,4 +20,4 @@ jobs:
           python-version: "3.12"
       - uses: akaihola/darker@v2.1.1
         with:
-          version: "2.1.1" # This will use the 2.1.1 darker Python package while using the 1.7.2 darker github action
+          version: "2.1.1"
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,20 +1,17 @@
-exclude: '^nucypher/blockchain/eth/contract_registry/*'
+exclude: "^nucypher/blockchain/eth/contract_registry/*"
 repos:
-
   - repo: local
     hooks:
-
       - id: tests
         name: Run Nucypher Unit Tests
         entry: scripts/run_unit_tests.sh
         language: system
         types: [python]
-        stages: [push]  # required additional setup: pre-commit install && pre-commit install -t pre-push
+        stages: [push] # required additional setup: pre-commit install && pre-commit install -t pre-push
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
     rev: v4.5.0
     hooks:
-
       # Git
       - id: forbid-new-submodules
 
@@ -38,13 +35,13 @@ repos:
   - repo: https://github.com/akaihola/darker
     rev: v2.1.1
     hooks:
-    -   id: darker
-        args: ["--check"]
+      - id: darker
+        args: ["--check", "--diff", "--color"]
         stages: [push]
-    -   id: darker
+      - id: darker
         stages: [commit]
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.4.5
     hooks:
-    - id: ruff
+      - id: ruff
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -14,12 +14,9 @@ from nucypher.policy.conditions.context import (
     USER_ADDRESS_CONTEXT,
     resolve_any_context_variables,
 )
-from nucypher.policy.conditions.exceptions import (
-    InvalidConditionContext,
-)
-from nucypher.policy.conditions.lingo import (
-    ConditionType,
-)
+from nucypher.policy.conditions.exceptions import InvalidConditionContext
+from nucypher.policy.conditions.lingo import ConditionType
+
 
 class AddressAllowlistCondition(AccessControlCondition):
     """
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -21,10 +21,6 @@ from nucypher.policy.conditions.lingo import (
     ConditionType,
 )
 
-# Maximum number of addresses allowed in the address allowlist
-MAX_ALLOWLIST_ADDRESSES = 25
-
-
 class AddressAllowlistCondition(AccessControlCondition):
     """
     A condition that checks if a user's wallet address is in a list of allowed addresses.
@@ -34,6 +30,10 @@ class AddressAllowlistCondition(AccessControlCondition):
     CONDITION_TYPE = ConditionType.ADDRESS_ALLOWLIST.value
 
     class Schema(AccessControlCondition.Schema):
+
+        # Maximum number of addresses allowed in the address allowlist
+        MAX_ALLOWLIST_ADDRESSES = 25
+
         condition_type = fields.Str(
             validate=validate.Equal(ConditionType.ADDRESS_ALLOWLIST.value),
             required=True,
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -15,7 +15,7 @@ from nucypher.policy.conditions.context import (
     resolve_any_context_variables,
 )
 from nucypher.policy.conditions.exceptions import (
-    RequiredContextVariable,
+    InvalidConditionContext,
 )
 from nucypher.policy.conditions.lingo import (
     ConditionType,
@@ -106,7 +106,7 @@ class AddressAllowlistCondition(AccessControlCondition):
                 - None (no additional data returned)
         """
         if not context:
-            raise RequiredContextVariable("No value provided for context variable")
+            raise InvalidConditionContext("No value provided for context variable")
 
         # Get user's address using resolve_any_context_variables
         user_address = resolve_any_context_variables(USER_ADDRESS_CONTEXT, **context)
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -1,7 +1,13 @@
 from typing import Dict, List, Optional, Tuple, Union
 
 from eth_utils import to_checksum_address
-from marshmallow import fields, validate
+from marshmallow import (
+    ValidationError,
+    fields,
+    post_load,
+    validate,
+    validates,
+)
 
 from nucypher.policy.conditions.base import AccessControlCondition
 from nucypher.policy.conditions.context import (
@@ -9,12 +15,14 @@ from nucypher.policy.conditions.context import (
     resolve_any_context_variables,
 )
 from nucypher.policy.conditions.exceptions import (
-    InvalidCondition,
     RequiredContextVariable,
 )
 from nucypher.policy.conditions.lingo import (
     ConditionType,
 )
+
+# Maximum number of addresses allowed in the address allowlist
+MAX_ALLOWLIST_ADDRESSES = 25
 
 
 class AddressAllowlistCondition(AccessControlCondition):
@@ -26,17 +34,46 @@ class AddressAllowlistCondition(AccessControlCondition):
     CONDITION_TYPE = ConditionType.ADDRESS_ALLOWLIST.value
 
     class Schema(AccessControlCondition.Schema):
-        condition_type = fields.Constant(ConditionType.ADDRESS_ALLOWLIST.value)
+        condition_type = fields.Str(
+            validate=validate.Equal(ConditionType.ADDRESS_ALLOWLIST.value),
+            required=True,
+        )
+
         addresses = fields.List(
             fields.String(required=True),
             required=True,
-            validate=validate.Length(min=1, max=25),
+            validate=validate.Length(min=1, max=MAX_ALLOWLIST_ADDRESSES),
         )
+
+        @validates("addresses")
+        def validate_addresses(self, addresses):
+            # Check for duplicates
+            if len(set(addresses)) != len(addresses):
+                raise ValidationError("Duplicate addresses are not allowed")
+
+            # Validate all addresses
+            for address in addresses:
+                try:
+                    # Ensure addresses have proper checksum
+                    normalized_address = to_checksum_address(address)
+                except ValueError as e:
+                    raise ValidationError(f"Invalid Ethereum address: {address}") from e
+                if normalized_address != address:
+                    raise ValidationError(
+                        f"Address {address} is not a checksummed address"
+                    )
+
+        @post_load
+        def make(self, data, **kwargs):
+            return AddressAllowlistCondition(**data)
 
     def __init__(
         self,
         addresses: List[str],
         name: Optional[str] = None,
+        condition_type: str = ConditionType.ADDRESS_ALLOWLIST.value,
+        *args,
+        **kwargs,
     ):
         """
         Initialize a AddressAllowlistCondition.
@@ -45,28 +82,15 @@ class AddressAllowlistCondition(AccessControlCondition):
             addresses: List of checksummed Ethereum addresses that are allowed to decrypt
             name: Optional name for the condition
         """
-        # Check for duplicates
-        if len(set(addresses)) != len(addresses):
-            raise InvalidCondition("Duplicate addresses are not allowed")
-
-        # Validate all addresses
-        for address in addresses:
-            try:
-                # Ensure addresses have proper checksum
-                normalized_address = to_checksum_address(address)
-            except ValueError as e:
-                raise InvalidCondition(f"Invalid Ethereum address: {address}") from e
-            if normalized_address != address:
-                raise InvalidCondition(
-                    f"Address {address} is not a checksummed address"
-                )
 
         # Store the validated addresses
         self.addresses = addresses
 
         super().__init__(
-            condition_type=self.CONDITION_TYPE,
+            condition_type=condition_type,
             name=name,
+            *args,
+            **kwargs,
         )
 
     def verify(self, **context) -> Tuple[bool, Union[None, Dict]]:
@@ -85,11 +109,9 @@ class AddressAllowlistCondition(AccessControlCondition):
             raise RequiredContextVariable("No value provided for context variable")
 
         # Get user's address using resolve_any_context_variables
-        user_address = resolve_any_context_variables(context[USER_ADDRESS_CONTEXT])[
-            "address"
-        ]
+        user_address = resolve_any_context_variables(USER_ADDRESS_CONTEXT, **context)
         # Simply check if the normalized address is in the allowlist
-        is_allowed = to_checksum_address(user_address) in self.addresses
+        is_allowed = user_address in self.addresses
 
         return is_allowed, None
 
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -81,18 +81,8 @@ class AddressAllowlistCondition(AccessControlCondition):
         *args,
         **kwargs,
     ):
-        """
-        Initialize a AddressAllowlistCondition.
-
-        Args:
-            addresses: List of checksummed Ethereum addresses that are allowed to decrypt
-            name: Optional name for the condition
-        """
-
-        # Store the validated user address
         self.user_address = user_address
 
-        # Store the validated addresses
         self.addresses = addresses
 
         super().__init__(
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -43,8 +43,7 @@ class AddressAllowlistCondition(AccessControlCondition):
         Initialize a AddressAllowlistCondition.
 
         Args:
-            addresses: List of checksummed wallet addresses that are allowed to decrypt (must be
-             properly checksummed Ethereum addresses or an exception will be raised)
+            addresses: List of checksummed Ethereum addresses that are allowed to decrypt
             name: Optional name for the condition
         """
         # Check for duplicates
@@ -84,9 +83,7 @@ class AddressAllowlistCondition(AccessControlCondition):
                 - None (no additional data returned)
         """
         if not context:
-            raise InvalidConditionContext(
-                "Context is required for address-allowlist condition"
-            )
+            raise RequiredContextVariable("No value provided for context variable")
 
         # Get user's address using resolve_any_context_variables
         user_address = resolve_any_context_variables(context[USER_ADDRESS_CONTEXT])[
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -106,8 +106,12 @@ class AddressAllowlistCondition(AccessControlCondition):
 
         # Get user's address using resolve_any_context_variables
         user_address = resolve_any_context_variables(self.user_address, **context)
+
+        user_address_bytes = bytes.fromhex(user_address[2:])
+        # the addresses are supposed to be already validated, in taco-web, to have a valid checksum
+        addresses_bytes = [bytes.fromhex(address[2:]) for address in self.addresses]
         # Simply check if the user address is in the allowlist
-        is_allowed = user_address in self.addresses
+        is_allowed = user_address_bytes in addresses_bytes
 
         return is_allowed, None
 
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -74,10 +74,10 @@ class AddressAllowlistCondition(AccessControlCondition):
 
     def __init__(
         self,
-        user_address: str,
         addresses: List[str],
-        name: Optional[str] = None,
+        user_address: str = USER_ADDRESS_CONTEXT,
         condition_type: str = ConditionType.ADDRESS_ALLOWLIST.value,
+        name: Optional[str] = None,
         *args,
         **kwargs,
     ):
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -58,16 +58,12 @@ class AddressAllowlistCondition(AccessControlCondition):
             # Validate all addresses
             for address in addresses:
                 try:
-                    # Ensure addresses have proper checksum
-                    normalized_address = to_checksum_address(address)
+                    # Ensure addresses are valid
+                    to_checksum_address(address)
                 except ValueError as e:
                     raise ValidationError(
                         f"Invalid Ethereum address: '{address}', provided in the list of addresses"
                     ) from e
-                if normalized_address != address:
-                    raise ValidationError(
-                        f"Address '{address}', provided in the list of addresses, is not a checksummed address"
-                    )
 
         @post_load
         def make(self, data, **kwargs):
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -14,8 +14,12 @@ from nucypher.policy.conditions.context import (
     USER_ADDRESS_CONTEXT,
     resolve_any_context_variables,
 )
-from nucypher.policy.conditions.exceptions import InvalidConditionContext
-from nucypher.policy.conditions.lingo import ConditionType
+from nucypher.policy.conditions.exceptions import (
+    InvalidConditionContext,
+)
+from nucypher.policy.conditions.lingo import (
+    ConditionType,
+)
 
 
 class AddressAllowlistCondition(AccessControlCondition):
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -18,16 +18,16 @@ from nucypher.policy.conditions.lingo import (
 )
 
 
-class WalletAllowlistCondition(AccessControlCondition):
+class AddressAllowlistCondition(AccessControlCondition):
     """
     A condition that checks if a user's wallet address is in a list of allowed addresses.
     The user must provide a signed message to prove ownership of the wallet.
     """
 
-    CONDITION_TYPE = ConditionType.WALLET_ALLOWLIST.value
+    CONDITION_TYPE = ConditionType.ADDRESS_ALLOWLIST.value
 
     class Schema(AccessControlCondition.Schema):
-        condition_type = fields.Constant(ConditionType.WALLET_ALLOWLIST.value)
+        condition_type = fields.Constant(ConditionType.ADDRESS_ALLOWLIST.value)
         addresses = fields.List(
             fields.String(required=True),
             required=True,
@@ -40,7 +40,7 @@ class WalletAllowlistCondition(AccessControlCondition):
         name: Optional[str] = None,
     ):
         """
-        Initialize a WalletAllowlistCondition.
+        Initialize a AddressAllowlistCondition.
 
         Args:
             addresses: List of checksummed wallet addresses that are allowed to decrypt (must be
@@ -85,7 +85,7 @@ class WalletAllowlistCondition(AccessControlCondition):
         """
         if not context:
             raise InvalidConditionContext(
-                "Context is required for wallet-allowlist condition"
+                "Context is required for address-allowlist condition"
             )
 
         # Get user's address using resolve_any_context_variables
@@ -99,7 +99,7 @@ class WalletAllowlistCondition(AccessControlCondition):
 
     def __eq__(self, other):
         return (
-            isinstance(other, WalletAllowlistCondition)
+            isinstance(other, AddressAllowlistCondition)
             and self.condition_type == other.condition_type
             and set(self.addresses) == set(other.addresses)
             and self.name == other.name
@@ -114,13 +114,13 @@ class WalletAllowlistCondition(AccessControlCondition):
     @classmethod
     def from_dict(cls, data):
         """
-        Create a WalletAllowlistCondition from a dictionary.
+        Create a AddressAllowlistCondition from a dictionary.
 
         Args:
             data: Dictionary containing the condition data
 
         Returns:
-            WalletAllowlistCondition instance
+            AddressAllowlistCondition instance
         """
         # Extract values from the dict (camelCase keys are converted to snake_case by marshmallow)
         addresses = data.get("addresses", [])
@@ -132,13 +132,13 @@ class WalletAllowlistCondition(AccessControlCondition):
     @classmethod
     def from_json(cls, data):
         """
-        Create a WalletAllowlistCondition from a JSON string.
+        Create a AddressAllowlistCondition from a JSON string.
 
         Args:
             data: JSON string containing the condition data
 
         Returns:
-            WalletAllowlistCondition instance
+            AddressAllowlistCondition instance
         """
         # Parse JSON to dict
         if isinstance(data, str):
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -1,4 +1,3 @@
-import json
 from typing import Dict, List, Optional, Tuple, Union
 
 from eth_utils import to_checksum_address
@@ -11,7 +10,7 @@ from nucypher.policy.conditions.context import (
 )
 from nucypher.policy.conditions.exceptions import (
     InvalidCondition,
-    InvalidConditionContext,
+    RequiredContextVariable,
 )
 from nucypher.policy.conditions.lingo import (
     ConditionType,
@@ -94,51 +93,5 @@ class AddressAllowlistCondition(AccessControlCondition):
 
         return is_allowed, None
 
-    def __eq__(self, other):
-        return (
-            isinstance(other, AddressAllowlistCondition)
-            and self.condition_type == other.condition_type
-            and set(self.addresses) == set(other.addresses)
-            and self.name == other.name
-        )
-
     def __repr__(self):
-        addresses_str = ", ".join(self.addresses[:3])
-        if len(self.addresses) > 3:
-            addresses_str += f"... (+{len(self.addresses) - 3} more)"
-        return f"{self.__class__.__name__}(addresses=[{addresses_str}])"
-
-    @classmethod
-    def from_dict(cls, data):
-        """
-        Create a AddressAllowlistCondition from a dictionary.
-
-        Args:
-            data: Dictionary containing the condition data
-
-        Returns:
-            AddressAllowlistCondition instance
-        """
-        # Extract values from the dict (camelCase keys are converted to snake_case by marshmallow)
-        addresses = data.get("addresses", [])
-        name = data.get("name")
-
-        # Create a new instance
-        return cls(addresses=addresses, name=name)
-
-    @classmethod
-    def from_json(cls, data):
-        """
-        Create a AddressAllowlistCondition from a JSON string.
-
-        Args:
-            data: JSON string containing the condition data
-
-        Returns:
-            AddressAllowlistCondition instance
-        """
-        # Parse JSON to dict
-        if isinstance(data, str):
-            data = json.loads(data)
-
-        return cls.from_dict(data)
+        return f"{self.__class__.__name__}(addresses_count={len(self.addresses)})"
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -1,6 +1,7 @@
 from typing import Dict, List, Optional, Tuple, Union
 
 from eth_utils import to_checksum_address
+from hexbytes import HexBytes
 from marshmallow import (
     ValidationError,
     fields,
@@ -107,9 +108,9 @@ class AddressAllowlistCondition(AccessControlCondition):
         # Get user's address using resolve_any_context_variables
         user_address = resolve_any_context_variables(self.user_address, **context)
 
-        user_address_bytes = bytes.fromhex(user_address[2:])
+        user_address_bytes = bytes(HexBytes(user_address))
         # the addresses are supposed to be already validated, in taco-web, to have a valid checksum
-        addresses_bytes = [bytes.fromhex(address[2:]) for address in self.addresses]
+        addresses_bytes = [bytes(HexBytes(address)) for address in self.addresses]
         # Simply check if the user address is in the allowlist
         is_allowed = user_address_bytes in addresses_bytes
 
--- a/nucypher/policy/conditions/address.py
+++ b/nucypher/policy/conditions/address.py
@@ -38,7 +38,10 @@ class AddressAllowlistCondition(AccessControlCondition):
             validate=validate.Equal(ConditionType.ADDRESS_ALLOWLIST.value),
             required=True,
         )
-
+        user_address = fields.Str(
+            validate=validate.Equal(USER_ADDRESS_CONTEXT),
+            required=True,
+        )
         addresses = fields.List(
             fields.String(required=True),
             required=True,
@@ -57,10 +60,12 @@ class AddressAllowlistCondition(AccessControlCondition):
                     # Ensure addresses have proper checksum
                     normalized_address = to_checksum_address(address)
                 except ValueError as e:
-                    raise ValidationError(f"Invalid Ethereum address: {address}") from e
+                    raise ValidationError(
+                        f"Invalid Ethereum address: '{address}', provided in the list of addresses"
+                    ) from e
                 if normalized_address != address:
                     raise ValidationError(
-                        f"Address {address} is not a checksummed address"
+                        f"Address '{address}', provided in the list of addresses, is not a checksummed address"
                     )
 
         @post_load
@@ -69,6 +74,7 @@ class AddressAllowlistCondition(AccessControlCondition):
 
     def __init__(
         self,
+        user_address: str,
         addresses: List[str],
         name: Optional[str] = None,
         condition_type: str = ConditionType.ADDRESS_ALLOWLIST.value,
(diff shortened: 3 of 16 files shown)


nucypher/policy/conditions/wallet.py
```diff
@@ -0,0 +1,143 @@
...
+                - Boolean indicating if verification passed
+                - None (no additional data returned)
+        """
+        if not context:
+            raise InvalidConditionContext(
+                "Context is required for wallet-allowlist condition"
+            )
+
+        # Extract the user's address from the context
+        user_address = extract_user_address_from_context(context)
```
cygnusv: Have a look at how this is handled in other conditions: https://github.com/nucypher/nucypher/blob/1dff7d830c2bdb3f35475e2ec5866e71c373d63e/nucypher/policy/conditions/evm.py#L106-L133

This is a generic EVM condition, so let's forget about the RPC execution part, and how `resolved_parameters` is populated. As you can see this implies calling `resolve_any_context_variables`, which automatically handles (1) resolution of `:userAddress` to an ETH address, and, even more importantly, (2) cryptographic validation that the requester owns that address. I'd change your custom context resolution logic to use `resolve_any_context_variables`, similarly as the code above.

nucypher/policy/conditions/context.py
```diff
@@ -133,6 +133,32 @@ def get_context_value(
     return value
 
 
+def extract_user_address_from_context(context: Dict[str, Any]) -> ChecksumAddress:
```
cygnusv: I'd remove this function and use `resolve_any_context_variables` below. See my other comment for the rationale.

cygnusv (dismissed):
This is a good start, but there's a few changes I think are important:
- There's existing logic to resolve `:userAddress`, also in a more secure way (i.e. including actual validation of a signature). I suggest to use that instead of the code proposed in `context.py`
- Also, the proposed context resolution code also allows non-standard user addresses keywords; let's stick to the standard.

As a separate comment, I'd rename this type of condition to `AddressAllowlist` instead of `Wallet...`. The latter has a deeper meaning, which in this case I don't think it's necessary to allude for a mere inclusion check of an address in a list.

nucypher/policy/conditions/wallet.py
Muhammad-Altabba: Sorry I thought the cryptographic validation happen already at some pervious steps. I will modify to use `resolve_any_context_variables`.

Muhammad-Altabba: I updated to use `resolve_any_context_variables` in the commit: https://github.com/nucypher/nucypher/pull/3601/commits/53688d63d16559e6a51bc0c9a290ef3044e5a906
However, what do you think about using `_resolve_user_address` directly after renaming it to `resolve_user_address` or after wrapping it inside a function that provide `USER_ADDRESS_CONTEXT` as the first param?
The new call would look like something similar to:
```py
        user_address = resolve_user_address(**context)["address"] , 
        # or: (and would choose more proper name later)
        user_address = resolve_user_public_address(**context)
```

tests/unit/conditions/test_address_allowlist_condition.py
```diff
@@ -0,0 +1,108 @@
...
+    assert original_condition == deserialized_condition
+
+    # Serialize to JSON
+    json_str = original_condition.to_json()
+
+    # Deserialize from JSON
+    deserialized_condition = AddressAllowlistCondition.from_json(json_str)
+
+    # Check equality
+    assert original_condition == deserialized_condition
```
manumonti: Maybe it is worth to add some schema validation? like

https://github.com/nucypher/nucypher/blob/c1a66a4cb9fc9b4b49751492f329e5053601acdd/tests/unit/conditions/test_time_condition.py#L30-L62

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,147 @@
...
+from nucypher.policy.conditions.lingo import (
+    ConditionType,
+)
+
+
+class AddressAllowlistCondition(AccessControlCondition):
+    """
+    A condition that checks if a user's wallet address is in a list of allowed addresses.
+    The user must provide a signed message to prove ownership of the wallet.
+    """
```
manumonti, line 30: WDYT? to avoid using the word wallet as @cygnusv suggested.

```suggestion
    """
    A condition that checks if a user's EOA address is in a list of allowed addresses.
    The user must provide a signed message to prove ownership of the EOA.
    """
```

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,147 @@
...
+        self,
+        addresses: List[str],
+        name: Optional[str] = None,
+    ):
+        """
+        Initialize a AddressAllowlistCondition.
+
+        Args:
+            addresses: List of checksummed wallet addresses that are allowed to decrypt (must be
+             properly checksummed Ethereum addresses or an exception will be raised)
```
manumonti: WDYT about this wording? Just trying to not use the word wallet and to not give redundant explanations.
```suggestion
            addresses: List of checksummed Ethereum addresses that are allowed to decrypt
```

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,147 @@
...
+
+        Returns:
+            Tuple containing:
+                - Boolean indicating if verification passed
+                - None (no additional data returned)
+        """
+        if not context:
+            raise InvalidConditionContext(
+                "Context is required for address-allowlist condition"
+            )
```
manumonti: What about using `RequiredContextVariable`?
```suggestion
            raise RequiredContextVariable("No value provided for context variable")
```

manumonti (changes requested):
Great! 🚀

I have some suggestions tho

nucypher/policy/conditions/address.py
Muhammad-Altabba: I guess here as we use `wallet "address"` it is descriptive enough, especially that it is for both EOA and Smart Contract Accounts as the authentication supports smart contract signature as well. Or what do you think?

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,144 @@
...
+    """
+
+    CONDITION_TYPE = ConditionType.ADDRESS_ALLOWLIST.value
+
+    class Schema(AccessControlCondition.Schema):
+        condition_type = fields.Constant(ConditionType.ADDRESS_ALLOWLIST.value)
+        addresses = fields.List(
+            fields.String(required=True),
+            required=True,
+            validate=validate.Length(min=1, max=25),
```
derekpierre: Let's make the maximum value of 25 a constant.

derekpierre: You could create a `ChecksumAddressField` object and take a list of those - analogous to what was done for Ursula addresses in Porter - see https://github.com/nucypher/nucypher-porter/blob/development/porter/fields/ursula.py#L8. This way, it would handle any invalid values AND your  `__init__` can take `addresses: List[ChecksumAddress]`, and you won't have to normalize them.

Alternatively, we did some simple validation in ContractCondition for the `contract_address` field; you could take a look at what was done there.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        if not context:
+            raise RequiredContextVariable("No value provided for context variable")
+
+        # Get user's address using resolve_any_context_variables
+        user_address = resolve_any_context_variables(context[USER_ADDRESS_CONTEXT])[
+            "address"
+        ]
+        # Simply check if the normalized address is in the allowlist
+        is_allowed = to_checksum_address(user_address) in self.addresses
```
derekpierre: You don't need to normalize the provided checksum address for `:userAddress` - it is properly derived.

```suggestion
        is_allowed = user_address in self.addresses
```

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,144 @@
...
+                - Boolean indicating if verification passed
+                - None (no additional data returned)
+        """
+        if not context:
+            raise RequiredContextVariable("No value provided for context variable")
+
+        # Get user's address using resolve_any_context_variables
+        user_address = resolve_any_context_variables(context[USER_ADDRESS_CONTEXT])[
+            "address"
+        ]
```
derekpierre: For this call you simply need to provide the context variable to resolve, and the context; you don't need to resolve anything else yourself.

```suggestion
        user_address = resolve_any_context_variables(USER_ADDRESS_CONTEXT, **context)
```

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,147 @@
...
+                normalized_address = to_checksum_address(address)
+            except ValueError as e:
+                raise InvalidCondition(f"Invalid Ethereum address: {address}") from e
+            if normalized_address != address:
+                raise InvalidCondition(
+                    f"Address {address} is not a checksummed address"
+                )
+
+        # Store the validated addresses
+        self.addresses = addresses
```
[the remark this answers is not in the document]
derekpierre: You don't need this if you implement the field suggestion I mentioned above.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,144 @@
...
+
+        return is_allowed, None
+
+    def __eq__(self, other):
+        return (
+            isinstance(other, AddressAllowlistCondition)
+            and self.condition_type == other.condition_type
+            and set(self.addresses) == set(other.addresses)
+            and self.name == other.name
+        )
```
derekpierre: What's the reason for implementing this? We don't usually need to compare conditions, so unless there is a strong reason, I'd remove this function.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,144 @@
...
+
+    def __repr__(self):
+        addresses_str = ", ".join(self.addresses[:3])
+        if len(self.addresses) > 3:
+            addresses_str += f"... (+{len(self.addresses) - 3} more)"
```
derekpierre: I would just print the length of addresses, rather than printing any specific addresses. Also helps with privacy concerns of logs.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,144 @@
...
+            data: JSON string containing the condition data
+
+        Returns:
+            AddressAllowlistCondition instance
+        """
+        # Parse JSON to dict
+        if isinstance(data, str):
+            data = json.loads(data)
+
+        return cls.from_dict(data)
```
derekpierre: These functions `from_dict(...)`, `from_json(...)` are already handled for you by the base class, and the use of the Schema. You'll notice that other conditions don't have to do this. Take a look at the `_Serializable` superclass for more details.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,144 @@
...
+    ):
+        """
+        Initialize a AddressAllowlistCondition.
+
+        Args:
+            addresses: List of checksummed Ethereum addresses that are allowed to decrypt
+            name: Optional name for the condition
+        """
+        # Check for duplicates
+        if len(set(addresses)) != len(addresses):
```
derekpierre: Nice!

tests/unit/conditions/test_address_allowlist_condition.py
```diff
@@ -0,0 +1,108 @@
...
+    allowed_account1 = Account.create()
+    allowed_account2 = Account.create()
+    not_allowed_account = Account.create()
+
+    # Create condition with allowed accounts
+    addresses = [allowed_account1.address, allowed_account2.address]
+    condition = AddressAllowlistCondition(addresses=addresses)
+
+    # Test successful verification with allowed account
+    context = {USER_ADDRESS_CONTEXT: {"address": allowed_account1.address}}
```
derekpierre: This won't work if you are properly resolving the `:userAddress`. You can skip the implicit validation during resolution by doing something like the following - https://github.com/nucypher/nucypher/blob/main/tests/acceptance/conditions/test_conditions.py#L86 - see the patch at the top of the test where the resolving skips the validation and simply returns the expected address.

tests/unit/conditions/test_address_allowlist_condition.py
derekpierre: You can compare the to_dict() outcomes and then you don't need the `__eq__` check. Computing equality on the dictionary version of the original and deserialized object is probably sufficient. If `__eq__` is only added for testing, then it's additional code we don't need.

derekpierre (changes requested):
Nice start! - left some suggestions.

tests/unit/conditions/test_address_allowlist_condition.py
Muhammad-Altabba: I replaced the serialization test by validation test.

nucypher/policy/conditions/address.py
Muhammad-Altabba: Thanks,

nucypher/policy/conditions/address.py
manumonti: > I guess here as we use `wallet "address"` it is descriptive enough, especially that it is for both EOA and Smart Contract Accounts as the authentication supports smart contract signature as well. Or what do you think?

it sounds good to me!

tests/unit/conditions/test_address_allowlist_condition.py
Muhammad-Altabba: Good catch, thanks,
However, I fixed it by passing a full correct data. As I think it is better to keep the validation since this specific test is about testing `condition.verify(**context)` so now a correctly constructed context is passed.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,118 @@
...
+                - Boolean indicating if verification passed
+                - None (no additional data returned)
+        """
+        if not context:
+            raise RequiredContextVariable("No value provided for context variable")
+
+        # Get user's address using resolve_any_context_variables
+        user_address = resolve_any_context_variables(USER_ADDRESS_CONTEXT, **context)
+        # Simply check if the normalized address is in the allowlist
+        is_allowed = user_address in self.addresses
```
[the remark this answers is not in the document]
Muhammad-Altabba: Using `user_address` and not `normalized_user_address` is intentional in order to ensure every address is checksummed and throw otherwise.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,119 @@
...
```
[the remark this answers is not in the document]
Muhammad-Altabba: It is intentional to not use a normalized address. The comparison is case sensitive.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    CONDITION_TYPE = ConditionType.ADDRESS_ALLOWLIST.value
+
+    class Schema(AccessControlCondition.Schema):
+        condition_type = fields.Str(
+            validate=validate.Equal(ConditionType.ADDRESS_ALLOWLIST.value),
+            required=True,
+        )
+
```
derekpierre: We need to define an additional field that has the `:userAddress` value in the condition itself, in order to be explicit about this condition needing a signed message for the user to prove they are who they say they are - see comment here https://github.com/nucypher/taco-web/pull/665#discussion_r2136597559.

(Somewhat similar prior conversation about explicitness - https://github.com/nucypher/nucypher/issues/3567#issuecomment-2515521314)

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,119 @@
...
+        Returns:
+            Tuple containing:
+                - Boolean indicating if verification passed
+                - None (no additional data returned)
+        """
+        if not context:
+            raise InvalidConditionContext("No value provided for context variable")
+
+        # Get user's address using resolve_any_context_variables
+        user_address = resolve_any_context_variables(USER_ADDRESS_CONTEXT, **context)
```
derekpierre: This is where the field I mentioned above would be used. Instead it would be something like

```suggestion
        user_address = resolve_any_context_variables(self.requester_address, **context)
```
(or whatever you end up calling the field)

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,119 @@
...
+        # Simply check if the normalized address is in the allowlist
```
derekpierre: No longer normalized
```suggestion
        # Simply check if the user address is in the allowlist
```

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,119 @@
...
+        is_allowed = user_address in self.addresses
+
+        return is_allowed, None
+
+    def __repr__(self):
+        return f"{self.__class__.__name__}(addresses_count={len(self.addresses)})"
```
derekpierre, line 120: 👌

tests/unit/conditions/test_address_allowlist_condition.py
```diff
@@ -0,0 +1,169 @@
...
+    allowed_account1 = Account.create()
+    allowed_account2 = Account.create()
+    not_allowed_account = Account.create()
+
+    # Create condition with allowed accounts
+    addresses = [allowed_account1.address, allowed_account2.address]
+    condition = AddressAllowlistCondition(addresses=addresses)
+
+    # Create proper EIP712 typed data structures for each account
+    def create_auth_message_for_account(account):
```
derekpierre: I can totally understand why you implemented this `create_auth_message_for_account()` function, but we can leverage an existing fixutre to achieve the same thing (with a small tweak). A couple things:
1. We have deprecated EIP712 auth messages in favour of EIP4361 (Sign-In with Ethereum) signatures in `taco-web`. We just haven't deprecated it yet in `nucypher` to continue to support older versions of `taco-web`. However, we will eventually deprecate and remove EIP712 support, so ideally, we only use EIP4361 auth messages for our tests.

2. You can leverage the existing `valid_eip4361_auth_message` fixture that we have to achieve this same thing and save yourself the extra code. The fixture does need to be tweaked a little since it only generates one but you need it to produce multiple different accounts/auth messages for the same test. In this case, you can use a fixture factory pattern.

See commit https://github.com/derekpierre/nucypher/commit/4d5bec8aa35fdc2200aaf1ee8920d5aab4b54a61 (based on top of your existing changes) which I drafted as a fix based on the points above - take a look. I didn't want to push a commit on top of your PR without your permission - so If you agree you can incorporate the commit.

tests/unit/conditions/test_address_allowlist_condition.py
derekpierre, line 1: Nice tests.

derekpierre (dismissed):

manumonti (approved):
No more suggestions from my side

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,128 @@
...
+                if normalized_address != address:
+                    raise ValidationError(
+                        f"Address '{address}', provided in the list of addresses, is not a checksummed address"
+                    )
+
+        @post_load
+        def make(self, data, **kwargs):
+            return AddressAllowlistCondition(**data)
+
+    def __init__(
```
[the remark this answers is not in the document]
Muhammad-Altabba: Validation should be happening through the superclass.

tests/fixtures.py
```diff
@@ -700,31 +700,39 @@ def valid_eip712_auth_message():
...
+            "typedData": f"{siwe_message}",
+        }
 
-    return auth_message
+        return auth_message
+
+    return _valid_eip4361_auth_message
+
+
+@pytest.fixture
```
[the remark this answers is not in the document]
Muhammad-Altabba: It is there to be used if needed in the future.

manumonti (approved):

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,128 @@
...
+)
+from nucypher.policy.conditions.exceptions import (
+    InvalidConditionContext,
+)
+from nucypher.policy.conditions.lingo import (
+    ConditionType,
+)
+
+# Maximum number of addresses allowed in the address allowlist
+MAX_ALLOWLIST_ADDRESSES = 25
```
derekpierre: Can we move  this constant within the AddressAllowlistCondition class.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,128 @@
...
+        condition_type: str = ConditionType.ADDRESS_ALLOWLIST.value,
+        *args,
+        **kwargs,
+    ):
+        """
+        Initialize a AddressAllowlistCondition.
+
+        Args:
+            addresses: List of checksummed Ethereum addresses that are allowed to decrypt
+            name: Optional name for the condition
```
derekpierre: This args list is incomplete - let's just remove the lines

```suggestion
```

nucypher/policy/conditions/types.py
```diff
@@ -81,6 +81,17 @@ class JWTConditionDict(_AccessControlCondition):
...
+# AddressAllowlistCondition represents:
+# {
+#     "conditionType": "address-allowlist",
+#     "addresses": List[str] (Ethereum addresses)
+# }
+#
+class AddressAllowlistConditionDict(_AccessControlCondition):
+    addresses: List[str]
+
+
```
derekpierre: Missing `userAddress`:

```suggestion
#
# AddressAllowlistCondition represents:
# {
#     "conditionType": "address-allowlist",
#     "addresses": List[str] (Ethereum addresses)
#.    "userAddress": str
# }
#
class AddressAllowlistConditionDict(_AccessControlCondition):
    addresses: List[str]
    userAddress: str


```

tests/unit/conditions/test_address_allowlist_condition.py
```diff
@@ -0,0 +1,135 @@
...
+    condition_dict = condition.to_dict()
+
+    # No issues here
+    AddressAllowlistCondition.from_dict(condition_dict)
+
+    # No issues with optional name
+    condition_dict["name"] = "my_address_allowlist"
+    AddressAllowlistCondition.from_dict(condition_dict)
+
+    with pytest.raises(InvalidConditionLingo):
```
derekpierre: Something to think about: when many of these different scenarios produce the same exception, it's good to also use `match="<unique substring of exception message>"` to ensure that each case is actually being tested. There are some examples of that in other tests.

Up to you if you decide to do it or not.

derekpierre (commented):
Looking good - just left some clean up suggestions.

tests/unit/conditions/test_address_allowlist_condition.py
```diff
@@ -0,0 +1,135 @@
...
+            addresses=["not-an-ethereum-address"],
+        )
+
+    # Test with duplicate addresses
+    with pytest.raises(InvalidCondition):
+        AddressAllowlistCondition(
+            user_address=USER_ADDRESS_CONTEXT,
+            addresses=[account1.address, account1.address],
+        )
+
```
derekpierre, line 53: Could also test a couple other scenarios:
1. invalid `user_address` value (i.e. not the correct context variable value)
2. non-checksummed address in addresses.

tests/unit/conditions/test_address_allowlist_condition.py
```diff
@@ -0,0 +1,135 @@
...
+        # Empty addresses list
+        condition_dict = condition.to_dict()
+        condition_dict["addresses"] = []
+        AddressAllowlistCondition.from_dict(condition_dict)
+
+    with pytest.raises(InvalidConditionLingo):
+        # Invalid address format
+        condition_dict = condition.to_dict()
+        condition_dict["addresses"] = ["not-an-ethereum-address"]
+        AddressAllowlistCondition.from_dict(condition_dict)
```
derekpierre, line 154: Could also test analogous cases for `user_address`

nucypher/policy/conditions/address.py
Muhammad-Altabba: That would be nice. But then I tried the code
```py
        addresses = fields.List(
            fields.String(required=True),
            required=True,
            validate=validate.Length(min=1, max=MAX_ALLOWLIST_ADDRESSES),
        )
```
It could not be able to reference `MAX_ALLOWLIST_ADDRESSES` and not even by `AddressAllowlistCondition.MAX_ALLOWLIST_ADDRESSES` because while defining the inner class `Schema`, the outer class properties are not accessible to the compiler yet because it sees the outer class as not fully available yet. 


And so the validation would become:

```py
        addresses = fields.List(
            fields.String(required=True),
            required=True,
            validate=validate.Length(min=1),
        )
```
And would then need to add something like the following to `validate_addresses(...)`:
```py
            if len(addresses) > AddressAllowlistCondition.MAX_ALLOWLIST_ADDRESSES:
                raise ValidationError(f"Maximum of {AddressAllowlistCondition.MAX_ALLOWLIST_ADDRESSES} addresses are allowed")
```

So, if you still think this is a better approach. Or if you have a suggestion. Let me know please, so I go with checking the length inside `validate_addresses(...)`.

nucypher/policy/conditions/lingo.py
```diff
@@ -119,6 +119,7 @@ class ConditionType(Enum):
     COMPOUND = "compound"
     SEQUENTIAL = "sequential"
     IF_THEN_ELSE = "if-then-else"
+    ADDRESS_ALLOWLIST = "address-allowlist"
```
cygnusv, line 122: Needs to be added to the comment in L107 above.

nucypher/policy/conditions/address.py
cygnusv: I think it should be possible. This is valid Python:
```python
>>> class Outer:
...     FOO = 1
...     class Inner:
...         BAR = Outer.FOO
...         
>>> Outer.Inner.BAR
1
```

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,118 @@
...
+    def __init__(
+        self,
+        user_address: str,
+        addresses: List[str],
+        name: Optional[str] = None,
+        condition_type: str = ConditionType.ADDRESS_ALLOWLIST.value,
+        *args,
+        **kwargs,
+    ):
+        self.user_address = user_address
```
cygnusv, line 82: If `user_address` is supposed to be equal than `USER_ADDRESS_CONTEXT`, shouldn't we initialize it directly in the constructor, by default? We do something similar in other condition schemas.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,118 @@
...
+                    # Ensure addresses have proper checksum
+                    normalized_address = to_checksum_address(address)
+                except ValueError as e:
+                    raise ValidationError(
+                        f"Invalid Ethereum address: '{address}', provided in the list of addresses"
+                    ) from e
+                if normalized_address != address:
+                    raise ValidationError(
+                        f"Address '{address}', provided in the list of addresses, is not a checksummed address"
+                    )
```
cygnusv: This approach may be too strict. For reference, we don't require checksum addresses in other condition schemas (e.g. `contract_address` in `ContractCall.Schema`). This is not a blocker, but we should strive for consistency. We can either relax this code, or make the one in `ContractCall.Schema` stricter, although tbh, I lean to the former.

tests/unit/conditions/test_address_allowlist_condition.py
```diff
@@ -0,0 +1,165 @@
...
+    # Create test addresses
+    account1 = Account.create()
+    account2 = Account.create()
+
+    addresses = [account1.address, account2.address]
+
+    # Test successful initialization
+    condition = AddressAllowlistCondition(
+        user_address=USER_ADDRESS_CONTEXT, addresses=addresses
+    )
```
cygnusv: See my comment above regarding initialization of `user_address`, because it may be avoidable. That would be better UX for the API.

nucypher/policy/conditions/address.py
derekpierre: Is the `MAX_ALLOWLIST_ADDRESSES ` constant used anywhere other than within the Schema class? If not, you can put the constant in the Schema class itself.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,118 @@
...
+        @post_load
+        def make(self, data, **kwargs):
+            return AddressAllowlistCondition(**data)
+
+    def __init__(
+        self,
+        user_address: str,
+        addresses: List[str],
+        name: Optional[str] = None,
+        condition_type: str = ConditionType.ADDRESS_ALLOWLIST.value,
```
derekpierre:
```suggestion
        addresses: List[str],
        user_address: str = USER_ADDRESS_CONTEXT,
        condition_type: str = ConditionType.ADDRESS_ALLOWLIST.value,
        name: Optional[str] = None,
```

nucypher/policy/conditions/address.py
derekpierre: The former would be better i.e. relaxing this code.

The others aren't strict because of the use of return value test, which uses bytes comparison for addresses, and not string comparison.

In the signing branch there is the addition of the "in" operator for return value test - https://github.com/nucypher/nucypher/pull/3599/commits/d3072608daf3937ee3d5edca27780ee3412862b1 - so we could think of using that instead. Something to think about in the future.

nucypher/policy/conditions/address.py
Muhammad-Altabba: @cygnusv regarding  https://github.com/nucypher/nucypher/pull/3601#discussion_r2163704998
Actually when I use

```py
class AddressAllowlistCondition(AccessControlCondition):

    MAX_ALLOWLIST_ADDRESSES = 25

    class Schema(AccessControlCondition.Schema):

        addresses = fields.List(
            fields.String(required=True),
            required=True,
            validate=validate.Length(min=1, max=AddressAllowlistCondition.MAX_ALLOWLIST_ADDRESSES),
        )

...
```
The following interesting error comes:
```py
nucypher/policy/conditions/address.py:25: in <module>
    class AddressAllowlistCondition(AccessControlCondition):
nucypher/policy/conditions/address.py:37: in AddressAllowlistCondition
    class Schema(AccessControlCondition.Schema):
nucypher/policy/conditions/address.py:49: in Schema
    validate=validate.Length(min=1, max=AddressAllowlistCondition.MAX_ALLOWLIST_ADDRESSES),
E   NameError: name 'AddressAllowlistCondition' is not defined
```

And when checking further. The error comes if python version was less than 3.7. So, let me know if it is still advised to change the code. And in this case we will need to update `python_requires='>=3',` to be `python_requires='>=3.7',` at our `setup.py` file.

nucypher/policy/conditions/address.py
Muhammad-Altabba: Actually the usage of this variable as an explicit provided one was done according to: https://github.com/nucypher/nucypher/pull/3601#discussion_r2138442235. And I took `jwt.py` as example. Please, let me know if I did not get it right.

nucypher/policy/conditions/address.py
Muhammad-Altabba: I hope to make the other schemas stricter, but I afraid of introducing breaking changes that would require upgrading the major version (some context discussion exists here: https://github.com/nucypher/taco-web/pull/665#discussion_r2138468856). 

At the same time I believe that checksum should be enforced when applicable.
As from my point of view checksum is a must that is typically oversight in the Web3 space. Every international, national or enterprise-level applications that I come to know, that use a number or a string as an identifier, uses a checksum. So, not just IBANs and Credit and Debit card numbers uses additional numbers or characters for checksum, but also National Identification Numbers, Passports and Drivers Driver's License uses Checksum. 

So, I think it is better to start the practice of enforcing the checksum. And for the old code we can revisit some time when we would raise the major version number. Or what do you think?

nucypher/policy/conditions/address.py
derekpierre: We only support python version >=3.9 and < 4 - https://github.com/nucypher/nucypher/blob/main/pyproject.toml#L8.

nucypher/policy/conditions/address.py
Muhammad-Altabba: Actually, I believe that not enforcing a checksum is like not enforcing that an email address should be having `@`. And normalizing the ethereum addresses by converting all to lower or uppercase is like stripping the last 1 or 2 digits of the passports, national ids, IBANs and Visa/Master Card numbers when saving or processing them because they are just checksums.
Maybe I am unnecessarily very opinionated about this 😬. But this my humble opinion 😊 .

nucypher/policy/conditions/address.py
derekpierre: That case is different since the `jwtToken` value can be any user-defined context variable (`:jwtToken` or `:token` or ...) - what it is providing is just a key-value pair.

In this case you have to use a specific context variable `:userAddress` so that the logic for ensuring the user provides a signed message to prove ownership of the wallet address is adhered to.

nucypher/policy/conditions/address.py
Muhammad-Altabba: So, I guess the version at setup.py need to be updated? https://github.com/nucypher/nucypher/blob/main/setup.py#L50. I did it. Let me please know it it was meant to be something else.

nucypher/policy/conditions/address.py
Muhammad-Altabba: Thanks @derekpierre 
I meat I took the approach of jwtToken by having an explicit variable as it was suggested to use explicit and not implicit. The variable in user_address is always `:userAddress` so I enforced that by throwing if it was not.  What is the exact point to be done in this regard?
I think @cygnusv was recommending using implicit value by not passing a value it to the constructor. Or is this something else meant?

nucypher/policy/conditions/address.py
Muhammad-Altabba: I fixed the in consistency in python version. However, this specific error does not seem to be related to python but to pytest. It gives an error if the inner class referenced the out class attributes. Let me know, please, if you could manage to do it on your machine without an error.

derekpierre: > So, I guess the version at setup.py need to be updated? https://github.com/nucypher/nucypher/blob/main/setup.py#L50. I did it. Let me please know it it was meant to be something else.

I believe so (good catch), but @KPrasch would know better than me.

> However, this specific error does not seem to be related to python but to pytest. It gives an error if the inner class referenced the out class attributes. Let me know, please, if you could manage to do it on your machine without an error.

Can you just add the constant to the Schema class instead?

nucypher/policy/conditions/address.py
derekpierre: What @cygnusv meant (or at least what I took from it), is using `USER_ADDRESS_CONTEXT` as the default value for `user_address` in the python constructor - my suggestion here https://github.com/nucypher/nucypher/pull/3601#discussion_r2163762093 does that. That way you don't have to always pass the `user_address` parameter in your tests. It's the same philosophy we use for `condition_type` in constructors.

derekpierre: Seems to be resolved via - https://github.com/nucypher/nucypher/pull/3601/commits/d53b641353d5e791bc9d1151f8676af42a7effb6.

tests/unit/conditions/test_address_allowlist_condition.py
derekpierre: Resolved via https://github.com/nucypher/nucypher/pull/3601/commits/d53b641353d5e791bc9d1151f8676af42a7effb6

nucypher/policy/conditions/address.py
derekpierre: Ethereum wallet addresses are 20 bytes long and represented in hexadecimal, which is case-insensitive. As a result, the EVM treats checksummed and non-checksummed addresses as equivalent. For backend equality comparisons, a consistent case-insensitive strategy (e.g., converting to lowercase) is sufficient and functionally identical.

The primary purpose of EIP-55 checksums is to catch human data entry errors - same for your other examples with credit cards etc. This makes them valuable at the user-interface/application-level, but less relevant for backend processing logic.

So, while I get your point from an overall safety perspective, I think we can loosen the restriction here to be consistent for now, and revisit the checksumming issue separately.

.pre-commit-config.yaml
```diff
@@ -38,13 +35,13 @@ repos:
   - repo: https://github.com/akaihola/darker
     rev: v2.1.1
     hooks:
-    -   id: darker
-        args: ["--check"]
+      - id: darker
+        args: ["--check", "--diff", "--color"]
```
derekpierre, line 39: We don't want "--diff" for the pre-commit, because `--diff` only shows the recommended update but does not update the file, and we want the pre-commit to update the file.

```suggestion
        args: ["--check", "--color"]
```

.github/workflows/darker.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.12'
+          python-version: "3.12"
       - uses: akaihola/darker@1.7.2
         with:
-          version: "1.7.2"
+          version: "2.1.1" # This will use the 2.1.1 darker Python package while using the 1.7.2 darker github action
```
derekpierre: I'm surprised that this works, I would have assumed the version has to be < the tag in the "uses" entry. In any case why do we not change both spots?

```suggestion
      - uses: akaihola/darker@2.1.1
        with:
          version: "2.1.1"
```

.pre-commit-config.yaml
Muhammad-Altabba: According to https://github.com/akaihola/darker the `--check` already does not update the file, right?
And, in that case, if writing to the files directly was intended. Both `--check` and. `--diff` needs to be removed, I guess.

.github/workflows/darker.yml
Muhammad-Altabba: I was also expecting this `      - uses: akaihola/darker@2.1.1` to work as I tried it earlier. But darker did not publish the github actions for akaihola/darker@2.1.1 and the last one is akaihola/darker@1.7.2. Although the last python package they published was 2.1.1.

.pre-commit-config.yaml
derekpierre: Seems so indeed. I assumed since `--check` was there already, that it wrote to files. Perhaps as a pre-commit hook it ignores the `--check` arg. In any case, if both need to be removed to be correct, then we should do that.

.github/workflows/darker.yml
derekpierre: What is the latest gh action version? It's unclear to me then that the usage of `version: "2.1.1"` is having any effect. 

We can update to whatever the latest gh action is instead of 2.1.1 for both the pre-commit and the gh action. It's best for those to be using the same version.

Muhammad-Altabba: Digging more into this. The github actions is reading from github branches. And the naming convention was updated to include `v` at the start of the version number.
I update now.

nucypher/policy/conditions/address.py
derekpierre: ^ Once the strictness is relaxed I can approve - unless there is still more to discuss.

.pre-commit-config.yaml
Muhammad-Altabba: Now Looking at the full darker hooks:
```yaml
  - repo: https://github.com/akaihola/darker
    rev: v2.1.1
    hooks:
      - id: darker
        args: ["--check", "--diff", "--color"]
        stages: [push]
      - id: darker
        stages: [commit]
```

 In the current configuration:

- When `commit`, darker will automatically format the Python files
- When `push`, darker will only check if files need formatting (using --check) and show the diffs that would be applied (--diff), but won't actually modify them. And this case of finding issues at push but not at commit, I think happens on cases like merging & rebasing... So, I think, it make sense to give a notification and prevent pushing in those cases so the pusher would have a look and modify manually. 

This was the old behaviour and still work as it was. The ` "--diff", "--color"` just makes the output nicer and more informative.

And so, all seems good to me. Let me know if you have some suggestions please.

.github/workflows/darker.yml
```diff
@@ -17,7 +17,7 @@ jobs:
           fetch-depth: 0
       - uses: actions/setup-python@v4
         with:
-          python-version: '3.12'
-      - uses: akaihola/darker@1.7.2
+          python-version: "3.12"
+      - uses: akaihola/darker@v2.1.1
         with:
-          version: "1.7.2"
+          version: "2.1.1" # This will use the 2.1.1 darker Python package while using the 1.7.2 darker github action
```
derekpierre: This comment is not needed anymore

```suggestion
          version: "2.1.1"
```

nucypher/policy/conditions/address.py
Muhammad-Altabba: @derekpierre @cygnusv 
Let me try convincing on this for the last time 😉. As I still could not push a commit to relax a validation that I see as very primitive to have 😊.

For us we do not know if the addresses will be entered manually by the use and then passed to us, or if it will be read from an automated system. This is why exactly why ids like the IBAN and even the driving licence has a checksum. So, I still cannot find a reason why we would encrypt a secret without validating the passed addresses. And I believe when a system allows for errors they will happen especially when counting on human-entered data.

Unfortunately Ethereum like Bitcoin was initially developed without any checksum in its protocol. And this led and still leading to people losing money as they do mistakes when entering the addresses. And there is lots of reported and unreported incidents. Even though the EIP-55 is not sufficient, because for example it fails to enforce an efficient checksum when the address consists of only numerical characters, but at least this is the least we have.

I cannot think or a way for us to sacrifice user experience allowing to ask for encryption to possibly wrong addresses. And delegating this to the UI (the users of taco-web), does not seem good enough either. Because unfortunately there is no guarantee that there will be no dapps which will just convert the address to lower case, or just pass it as is.

I hope the above clarify why I see this as an essential practice. But if not, let me know so I remove the codes that do the validation, and keep the discussion to some another opportunity in future...

Thanks for reading 😄

Muhammad-Altabba: Thanks @cygnusv @derekpierre,

I removed the checksum validation as per our discussion that this validation will be done by taco-web only. 

However, the rest of the validations are still here in addition to being in the taco-web: 
- Ensure no duplication (It is currently only here but I will add it to taco-web).
- Ensure the string is a valid ethereum address. Here just validating that the string could represent an ethereum address. And at taco-web this validation will happen in addition to the checksum.

The validation here now looks like:
```py
  @validates("addresses")
  def validate_addresses(self, addresses):
      # Check for duplicates
      if len(set(addresses)) != len(addresses):
          raise ValidationError("Duplicate addresses are not allowed")

      # Validate all addresses
      for address in addresses:
          try:
              # Ensure addresses are valid
              to_checksum_address(address)
          except ValueError as e:
              raise ValidationError(
                  f"Invalid Ethereum address: '{address}', provided in the list of addresses"
              ) from e
```

tests/unit/conditions/test_address_allowlist_condition.py
derekpierre: FYI this can now be the following:

```suggestion
    condition = AddressAllowlistCondition(
        addresses=addresses
    )
```

Same for other locations with the same code.

nucypher/policy/conditions/address.py
derekpierre: Makes sense - was this update pushed up?

Muhammad-Altabba: Sorry, they are now pushed.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,115 @@
...
+                - Boolean indicating if verification passed
+                - None (no additional data returned)
+        """
+        if not context:
+            raise InvalidConditionContext("No value provided for context variable")
+
+        # Get user's address using resolve_any_context_variables
+        user_address = resolve_any_context_variables(self.user_address, **context)
+        # Simply check if the user address is in the allowlist
+        is_allowed = user_address in self.addresses
```
derekpierre: This is a case-sensitive check. We should make this check case-insensitive, otherwise you are beholden to the case of the list of addresses provided. I know we talked about validating on the user/application-side, but this is more about completeness of the functionality on the backend.

For example, you could internally convert the list of addresses to be a list of lowercase addresses then you could do something like:
```suggestion
        is_allowed = user_address.lower() in self.addresses
```
OR

```suggestion
        is_allowed = False
        for address in self.addresses:
             if user_address.lower() == address.lower():
                 is_allowed = True
```
There are alternative strategies for case insensitivity you could consider as well. You just need to make sure that whatever strategy is used is used for both the `user_address` to check, is also used for populating the list it is being checked against.

If you haven't already, let's also add unit test(s) to ensure case-insensitivity.

Muhammad-Altabba: As we agreed that the taco-web will pass only checksummed address, right? Which means the case should be sensitive. And converting to lowercase here defeats the purpose of a checksum that is delegate to be on the taco-web, as it is the one that will compose the condition and pass to the nodes.

As we are relaxing the check of a checksum in this repo, we sill require the address to be checksummed and hence case sensitive.

derekpierre: > As we agreed that the taco-web will pass only checksummed address, right?

That's correct, but that agreement was for `taco-web`, and not `nucypher/nucypher`. It's the same reason you aren't validating that the address is checksummed in the `addresses` field in `nucypher/nucypher` rather only checking that it's a valid ethereum address. From a functional perspective the backend (and the EVM) doesn't care about case. 

For the backend i.e. `nucypher/nucypher`,  if you compare an address, the checksummed address and non-checksummed versions should be equal.

> we sill require the address to be checksummed 

The backend i.e. `nucypher/nucypher` doesn't *require* it, only `taco-web`. Remember that ethereum addresses are case insensitive.

Muhammad-Altabba: Sorry, think if taco-web will do the job of ensuring the addresses are checksumed, then it needs to be case sensitive here. Otherwise if there is any doupt that the addresses are not passed early as checksummed, then I think we need to keep back the checksum here. As per my understanding that the only reason for not enforcing a checksum here, is that it is already enforced at taco web.
Let us please have a discussion again next week, if you still think we should not keep the checksum but also convert to lower case.
Thanks,

derekpierre: Happy to chat about it on Monday.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,119 @@
...
+                - Boolean indicating if verification passed
+                - None (no additional data returned)
+        """
+        if not context:
+            raise InvalidConditionContext("No value provided for context variable")
+
+        # Get user's address using resolve_any_context_variables
+        user_address = resolve_any_context_variables(self.user_address, **context)
+
+        user_address_bytes = bytes.fromhex(user_address[2:])
```
derekpierre: If you use the `HexBytes` class it'll save you from needing to skip the '0x' in the beginning of the string. Example:

```suggestion
        user_address_bytes = bytes(HexBytes(user_address))
```
same for the address list conversion to list of bytes below.

nucypher/policy/conditions/address.py
```diff
@@ -0,0 +1,119 @@
...
+        # the addresses are supposed to be already validated, in taco-web, to have a valid checksum
+        addresses_bytes = [bytes.fromhex(address[2:]) for address in self.addresses]
```
derekpierre:
```suggestion
        addresses_bytes = [bytes(HexBytes(address)) for address in self.addresses]
```

derekpierre (approved):
🎸

PR #3602, opened 2025-06-06
Support Multiple Curves for ECDSA Conditions

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
1

**What this does:**
- Supports All elliptic curves in `ecdsa` library for ECDSA verification
- Standardizes public keys and signatures to hex encoding for ECDSA conditions
- Expands testing of ECDSA conditions to support multicurve

**Why it's needed:**
Some applications (like discord bots) use other curves, namely, ED25519.

**Notes**
Based over #3605

### DIFF
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -104,6 +104,10 @@ def is_context_variable(variable) -> bool:
     return isinstance(variable, str) and CONTEXT_REGEX.fullmatch(variable)
 
 
+def is_bytes_context_variable(variable) -> bool:
+    return isinstance(variable, str) and variable.startswith(":bytes")
+
+
 def string_contains_context_variable(variable: str) -> bool:
     matches = re.findall(CONTEXT_REGEX, variable)
     return bool(matches)
@@ -126,6 +130,14 @@ def get_context_value(
             raise RequiredContextVariable(
                 f'No value provided for unrecognized context variable "{context_variable}"'
             )
+        elif is_bytes_context_variable(context_variable):
+            # if it is a bytes context variable, convert the string to bytes
+            try:
+                value = bytes.fromhex(value)
+            except ValueError:
+                raise InvalidContextVariableData(
+                    f'Invalid bytes context variable "{context_variable}"; expected a hex string'
+                )
         elif isinstance(value, str):
             # possible big int value
             value = check_and_convert_big_int_string_to_int(value)
@@ -149,16 +161,14 @@ def resolve_any_context_variables(
         }
     elif isinstance(param, str):
         # either it is a context variable OR contains a context variable within it
-        # TODO separating the two cases for now out of concern of regex searching
-        #  within strings (case 2)
-        if is_context_variable(param):
+        if is_context_variable(param) or is_bytes_context_variable(param):
             return get_context_value(
                 context_variable=param, providers=providers, **context
             )
         else:
+            # Handles multiple context variables within a string (ie 'https://api.github.com/user/:foo/:bar')
             matches = re.findall(CONTEXT_REGEX, param)
             for context_var in matches:
-                # checking out of concern for faulty regex search within string
                 if context_var in context:
                     resolved_var = get_context_value(
                         context_variable=context_var, providers=providers, **context
--- a/nucypher/policy/conditions/ecdsa.py
+++ b/nucypher/policy/conditions/ecdsa.py
@@ -24,6 +24,9 @@ from nucypher.policy.conditions.exceptions import (
 )
 from nucypher.utilities.logging import Logger
 
+SUPPORTED_ECDSA_CONDITION_CURVES = {c.name: c for c in curves}
+DEFAULT_ECDSA_CONDITION_CURVE = NIST192p
+
 
 class ECDSAVerificationCall(ExecutionCall):
     # Use SHA-256 for hashing
@@ -35,7 +38,7 @@ class ECDSAVerificationCall(ExecutionCall):
         verifying_key = fields.Str(required=True)
         curve = fields.Str(
             required=False,
-            validate=validate.OneOf([curve.name for curve in curves]),
+            validate=validate.OneOf(list(SUPPORTED_ECDSA_CONDITION_CURVES)),
         )
 
         @post_load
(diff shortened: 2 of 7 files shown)


nucypher/policy/conditions/ecdsa.py
```diff
@@ -41,30 +45,27 @@ def validate_message(self, value):
...
                     raise ValidationError(
-                        f"Invalid signature format, must be base64 encoded: {str(e)}"
+                        f"Invalid signature format, must be hex encoded: {str(e)}"
                     )
 
         @validates("verifying_key")
         def validate_verifying_key(self, value):
             try:
-                VerifyingKey.from_pem(value.encode())
+                bytes.fromhex(value)
```
derekpierre: Was the reason for not trying to create the VerifyingKey the fact that you now need to specify the curve to create the VerifyingKey?

If so, then you can check multiple schema values at the same time using `@validates_schema` instead of `@validates` - see something analogous for `ContractCall` https://github.com/nucypher/nucypher/blob/main/nucypher/policy/conditions/evm.py#L273.

nucypher/policy/conditions/ecdsa.py
```diff
@@ -143,14 +148,21 @@ def __init__(
...
         verifying_key: str,
+        curve: Optional[str] = NIST192p.name,
         condition_type: str = CONDITION_TYPE,
         name: Optional[str] = None,
     ):
+        if curve not in self.SUPPORTED_CURVES:
+            raise InvalidCondition(
+                f"Unsupported curve: {curve}. Supported curves are: {list(self.SUPPORTED_CURVES.keys())}"
+            )
+
```
derekpierre: I don't think you need this validation check because the validation of ExecutionCall schema already does this check for you since the underlying curve field in the schema has a validation associated with it.

nucypher/policy/conditions/ecdsa.py
```diff
@@ -143,14 +148,21 @@ def __init__(
...
         try:
             self.execution_call = ECDSAVerificationCall(
                 message=message,
                 signature=signature,
                 verifying_key=verifying_key,
+                curve=self.SUPPORTED_CURVES[curve],
```
derekpierre: I think you can just pass the curve value here - see comment above. It will be validated by the Schema.

nucypher/policy/conditions/ecdsa.py
```diff
@@ -26,6 +26,10 @@ class Schema(ExecutionCall.Schema):
         message = fields.Raw(required=True)
         signature = fields.Str(required=True)
         verifying_key = fields.Str(required=True)
+        curve = fields.Str(
```
derekpierre, line 39: 👍

nucypher/policy/conditions/ecdsa.py
```diff
@@ -130,6 +134,7 @@ class ECDSACondition(AccessControlCondition):
     """
 
     CONDITION_TYPE = "ecdsa"  # Add this to ConditionType enum
+    SUPPORTED_CURVES = {c.name: c for c in curves}
```
derekpierre: If my comments below are accepted, then you may not need this constant, or it can be moved.

tests/unit/conditions/test_ecdsa_condition.py
```diff
@@ -134,20 +139,142 @@ def test_ecdsa_condition_verify_invalid_signature():
...
         ":message_variable": TEST_MESSAGE,
-        ":signature_variable": invalid_signature_b64,
+        ":signature_variable": invalid_signature,
     }
     success, result = condition.verify(**context)
     assert not success
     assert result is False
+
+
+def test_ecdsa_condition_different_curves():
```
derekpierre, line 163: Is it possible to parameterize this test to take (curve name, <other things>) tuples to simplify the test?

Similar comment for some of the integration tests?

tests/integration/conditions/test_ecdsa_condition_integration.py
```diff
@@ -123,3 +129,32 @@ def test_ecdsa_in_compound_condition():
 
     # Now both conditions are valid, so AND should pass
     assert and_lingo.eval(**context) is True
+
+
+def test_discord_ed25519_with_ecdsa_condition():
```
derekpierre, line 132: 🔥

nucypher/policy/conditions/ecdsa.py
KPrasch: ✔️

nucypher/policy/conditions/ecdsa.py
```diff
@@ -41,30 +55,46 @@ def validate_message(self, value):
...
+                curve = SUPPORTED_ECDSA_CONDITION_CURVES[curve_name]
+            else:
+                curve = DEFAULT_ECDSA_CONDITION_CURVE
             try:
-                VerifyingKey.from_pem(value.encode())
+                verifying_key_bytes = bytes.fromhex(value)
+            except ValueError:
+                raise ValidationError(
+                    "Invalid verifying key format, must be hex encoded"
+                )
```
derekpierre, line 82: I don't believe you need this because `@validates` functions get called before `@validates_schema` functions, so this validation gets taken care of by `validate_signature` - https://marshmallow.readthedocs.io/en/latest/upgrading.html?utm_source=chatgpt.com#schema-level-validators-are-skipped-when-field-validation-fails.

nucypher/policy/conditions/ecdsa.py
```diff
@@ -41,30 +55,46 @@ def validate_message(self, value):
         @validates("signature")
         def validate_signature(self, value):
             if not is_context_variable(value):
-                # Try to decode it to ensure it's valid base64
                 try:
-                    base64.b64decode(value)
+                    bytes.fromhex(value)
```
derekpierre, line 59: Does this imply that we **only** accept hex that is **not** `0x` prefixed?

If so, we need to make sure that the `taco-web` side adheres to that.

nucypher/policy/conditions/ecdsa.py
```diff
@@ -26,6 +36,10 @@ class Schema(ExecutionCall.Schema):
         message = fields.Raw(required=True)
         signature = fields.Str(required=True)
```
derekpierre, line 37: Should `message` and `signature` basically always be context variables given that we don't know these until request time? The only thing set for the condition at encryption time would be the verifying key and the curve..., is that right?

nucypher/policy/conditions/context.py
```diff
@@ -104,6 +104,10 @@ def is_context_variable(variable) -> bool:
     return isinstance(variable, str) and CONTEXT_REGEX.fullmatch(variable)
 
 
+def is_bytes_context_variable(variable) -> bool:
```
derekpierre, line 107: Thought about it some more, and given that you would accept signature as a hex string in the context var, then it makes sense to also accept message as hex as well instead of doing the `:bytes:message:` context variable. Not really much of a difference and feels like you can always to `bytes.fromHex()` for the `message` like you do with the signature anyway.

Happy to chat more about it tomorrow in case I'm misunderstanding.

nucypher/policy/conditions/ecdsa.py
KPrasch: It's a good observation.  Do you think we should go with 0x prefix here as an additional measure to ensure bytes deserialization/usage?

nucypher/policy/conditions/ecdsa.py
KPrasch: Yes - that's correct.

nucypher/policy/conditions/context.py
KPrasch: You’ve got the idea. Another scenario to consider is when the condition is meant to verify a hex string itself (not its decoded bytes). If we always assume hex strings should be deserialized into bytes, then we lose the ability to verify plain strings.

One potential solution is to give the `ECDSACondition` (or perhaps `SignatureRequest`?) object an additional field that explicitly indicates the expected message type (e.g. raw, utf8, hex, base64, etc.).

This implementation takes a broader approach to allow any context variable to be case into bytes in Ursula's runtime.

nucypher/policy/conditions/ecdsa.py
KPrasch: I think this is resolved?   Have another look and let me know if I've addressed your suggestion here.

nucypher/policy/conditions/ecdsa.py
derekpierre: `message` and `signature` should just context variables and no other value - let's follow-up in another PR (keep momentum).

tests/unit/conditions/test_ecdsa_condition.py
derekpierre: Follow-up in another PR to keep momentum.

derekpierre (approved):
Nice work. Some follow-up changes that we agreed to do in a follow-up PR.

nucypher/policy/conditions/ecdsa.py
derekpierre: I think the `0x` prefix is good. It would be clearer that it is hex, and consistent with our other uses of hex that already use the `0x` prefix.

We can discuss more in the follow-up PR.

PR #3606, opened 2025-06-11
Attribute-based Signing Object Condition

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

- A condition that can introspect various values on an object being signed.
- Add "in" operator to return value test for checking membership in a list.

**Issues fixed/closed:**

Related to https://github.com/nucypher/sprints/issues/176

**Notes for reviewers:**

Good milestone for a review/merge into the epic and then do follow-up PRs. I don't want these PRs to get too big.

**NOTE** There are some TODO comments in the PR - take a look and provide any opinions.

### DIFF
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -39,6 +39,7 @@ from nucypher.policy.conditions.types import ConditionDict, Lingo
 from nucypher.policy.conditions.utils import (
     CamelCaseSchema,
     ConditionProviderManager,
+    check_and_convert_any_big_ints,
     check_and_convert_big_int_string_to_int,
 )
 
@@ -50,23 +51,11 @@ class AnyField(fields.Field):
     numbers as integers, so those need converting to integers.
     """
 
-    def _convert_any_big_ints_from_string(self, value):
-        if isinstance(value, list):
-            return [self._convert_any_big_ints_from_string(item) for item in value]
-        elif isinstance(value, dict):
-            return {
-                k: self._convert_any_big_ints_from_string(v) for k, v in value.items()
-            }
-        elif isinstance(value, str):
-            return check_and_convert_big_int_string_to_int(value)
-
-        return value
-
     def _serialize(self, value, attr, obj, **kwargs):
         return value
 
     def _deserialize(self, value, attr, data, **kwargs):
-        return self._convert_any_big_ints_from_string(value)
+        return check_and_convert_any_big_ints(value)
 
 
 class AnyLargeIntegerField(fields.Int):
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -21,9 +21,9 @@ from marshmallow.validate import OneOf, Range
 from packaging.version import parse as parse_version
 
 from nucypher.policy.conditions.base import (
-    AccessControlCondition,
+    Condition,
     ExecutionCall,
-    MultiConditionAccessControl,
+    MultiCondition,
     _Serializable,
 )
 from nucypher.policy.conditions.context import (
@@ -140,7 +140,7 @@ class Operator(Enum):
         return [op.value for op in cls]
 
 
-class CompoundAccessControlCondition(MultiConditionAccessControl):
+class CompoundCondition(MultiCondition):
     """
     A combination of two or more conditions connected by logical operators such as AND, OR, NOT.
 
@@ -166,7 +166,7 @@ class CompoundAccessControlCondition(MultiConditionAccessControl):
     def _validate_operator_and_operands(
         cls,
         operator: str,
-        operands: List[AccessControlCondition],
+        operands: List[Condition],
     ):
         if operator not in cls.OPERATORS:
             raise ValidationError(
@@ -191,7 +191,7 @@ class CompoundAccessControlCondition(MultiConditionAccessControl):
                 message="Maximum of {cls.MAX_NUM_CONDITIONS} operands allowed for '{operator}' compound condition",
             )
 
-    class Schema(AccessControlCondition.Schema):
+    class Schema(Condition.Schema):
         condition_type = fields.Str(
             validate=validate.Equal(ConditionType.COMPOUND.value), required=True
         )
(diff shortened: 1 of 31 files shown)


derekpierre:
Good milestone for a review/merge into the epic and then do follow-up PRs.

nucypher/policy/conditions/lingo.py
```diff
@@ -120,6 +109,8 @@ class ConditionType(Enum):
     SEQUENTIAL = "sequential"
     IF_THEN_ELSE = "if-then-else"
     ECDSA = "ecdsa"
+    ATTRIBUTE = "attribute"
+    ABI_ATTRIBUTE = "abi-attribute"
```
KPrasch, line 113: Help me understand:  does this imply abi decoding?  What necessitates a variant from the standard attribute condition?

nucypher/policy/conditions/lingo.py
```diff
@@ -668,9 +674,22 @@ def eval(self, data) -> bool:
...
+        comparator_function = _COMPARATOR_FUNCTIONS.get(self.comparator)
+        if comparator_function == pyoperator.contains:
+            # for 'contains' operator, the left operand is the value and
+            # right operand is the data to check
+
+            # first need to sanitize all values in list
+            sanitized_list = [self._sanitize_value(v) for v in self.value]
+            left_operand = sanitized_list
+
+            right_operand = self._sanitize_value(processed_data)
```
KPrasch, line 686: ⚡

tests/acceptance/actors/test_signing_ritual.py
```diff
@@ -353,19 +387,240 @@ def test_user_op_signing_request_fulfilment(
...
+    )
+
+    # set condition
+    signing_object_condition = SigningObjectAttributeCondition(
+        attribute_name="callData",
+        return_value_test=ReturnValueTest(
+            comparator="==",
+            value=HexBytes(erc20_transfer_op.call_data).hex(),
+        ),
+    )
```
KPrasch, line 526: Do I understand correctly that this condition only permits a 1 token erc20 transfer?  Would love to see an example of higher-order helpers as the python API for userop guardrails in plain language (perhaps a follow up PR that is API-focused that will be used as a reference for taco-web).

nucypher/policy/conditions/lingo.py
derekpierre: It does imply ABI decoding.

The original attirbute condition is just looking at the raw attribute value on a python object and the value check is based on the attribute's value as whole.

The idea here is that for attributes like "callData", it is made up of bytes that represent specific execution data.  Consequently, if you want to check some value within the calldata you will need the expected abi for the callData to understand it's representation, and the breakdown of the values that comprise the call data, for further filtering. An example being, I only expect transfer function calls to be signed, but the value of each transfer should be < 5 ETH.

This is still something to be refined - may need to basically mimic what is done in some of the caveat enforcers defined eg. https://github.com/MetaMask/delegation-framework/blob/main/src/enforcers/AllowedMethodsEnforcer.sol OR https://github.com/zerodevapp/kernel-7579-plugins/tree/master/policies

tests/acceptance/actors/test_signing_ritual.py
derekpierre: Yes, It permits a very specific `callData` bytes value.

Not sure whether higher order helpers will be in `nucypher` or in `taco-web`. We have more freedom with `taco-web` in terms of upgrade process for adding more. That being said, I just don't know what the base fully looks like and therefore what the helpers could look like in the moment - open to hearing about some if you have thoughts.

derekpierre:
I am thinking of reworking the `abi_attribute` condition in this PR a bit. Some of its shortcomings are:
1. Whenever the `calldata` doesn't match the provided `abi_decode_string`, an exception is thrown. Of course we can catch the exception and then return False for the condition, but catching a condition exception and then simply return False feels like it is masking underlying error cases - this was something that also came up in JWTCondition. In any case, with the current code  it is hard to say "allow these methods but not this other method"
2. It's a bit clunky to add multiple checks for the same call data. You need multiple `SigningObjectAbiAttributeCondition` objects for checking different values for the same signing object
3. There is some functional overlap between `abi_decode_value_index` and the `index` value in return value. I didn't want to use the `return_value_test`'s `index` because it seemed disconnected, but now that I'm rethinking this condition, I should have still avoided this conflict.

To that end, I'm thinking of reworking the abi condition, but I plan to do so in a separate PR as a follow-up to this one. One possible structure for the updated condition could be:

```json
{
    "condition_type": "allowed_abi_calls",
    "attribute_name": "call_data",
    "allowed_abi_calls": {
        "transfer(address, uint256)": [{
            "parameter_index": 0,            # address value
            "returnValueTest": {
                "comparator": "in",
                "value": [
                    "0xabcdef",
                    "0xfedcba"
                ]
            }
        },
        {
            ...
        }],
        "approve(address, uint256)": [],  # empty i.e. no additional parameter value validation
        "execute(address)": [
            ...
        ]
    }
}
```

In this way, allowed calls can be easily specified (keys in the dictionary), and then any additional parameter value validations for the function parameters can be done as part of the same condition (dictionary values). Note that the corresponding list of additional validations for a specified abi signature can be empty for the corresponding abi method which implies that the abi method is allowed and there are no additional validations to perform on the parameter values.

The condition would basically check that the call data matches one of the keys in the "allowed_abi_calls" dictionary and perform any corresponding additional validations on parameter values if provided.

Still thinking on it. If anyone has ideas, or wants to jam on it let me know (cc @arjunhassard ).

derekpierre:
> To that end, I'm thinking of reworking the abi condition, but I plan to do so in a separate PR as a follow-up to this one
...
> Still thinking on it. If anyone has ideas, or wants to jam on it let me know (cc @arjunhassard ).

Working through the above (with some further tweaks) in https://github.com/nucypher/nucypher/pull/3609.

theref (approved):

nucypher/policy/conditions/signing/base.py
```diff
@@ -0,0 +1,236 @@
...
+        self,
+        attribute_name: str,
+        return_value_test: ReturnValueTest,
+        signing_object_context_var: str = SIGNING_CONDITION_OBJECT_CONTEXT_VAR,
+        condition_type: str = ConditionType.ATTRIBUTE.value,
+        name: Optional[str] = None,
+    ):
+        # TODO what about camel case (from library) vs snake case (python) for
+        #  attribute names (is this sufficient?)
+        #  At the moment since there will be a python object, can we assume snake case conversion always?
```
manumonti, line 85: I think this conversion to snake case is fine. We are already using snake case for :userAddress context variable, so it makes sense to me to use the same convention for attributes.

manumonti (approved):
LGTM! 🙌

PR #3609, opened 2025-06-19
Signing Object ABI Attribute Validation follow up

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Follow-up to #3606 .
Based over #3606 .

The big issue being that callData can be nested within itself, so how do we structure the condition to handle that - looking at a recursive solution for now. It is very likely that this is a foundational base for which specific helpers would need to be built on top of - not sure if the helpers will be in python or typescript currently.

Structure:

```python
# AbiParameterValidation represents:
# {
#     "parameterIndex": str
#     "indexWithinTuple": int
#     "returnValueTest: <>
#     "nestedAbiValidation: <>
# }
```
(_either "returnValueTest" or "nestedAbiValidation" i.e. check the value itself or its an underlying calldata set of bytes that needs its own validation_)

```python
# AbiCallValidation
# {
#    "allowedAbiCalls": {
#        <human_abi_signature>: ["AbiParameterValidation"]
#    }
# }

# SigningObjectAbiAttributeCondition represents:
# {
#     "attributeName": str
#     "signingObjectContextVar": ":signingConditionObject"
#     "abiValidation": <abi_call_validation>
# }
```

- Also added "!in" operator for return value test to balance our "in" operator.

_`lynx` is already running this code_

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

The other possible aspect of abi validation would be to actually simulate the transaction and then have additional checks on events, balances etc. - this was raised by @cygnusv . However, to simulate tx execution and check subsequent state is a much more involved process. It requires either local simulation eg. anvil (https://medium.com/@maria.magdalena.makeup/foundry-anvil-a-local-ethereum-node-for-development-642ca28f7892) , hardhat etc. or remote simulation eg. tenderly API (https://docs.tenderly.co/simulations/single-simulations#simulate-via-api), alchemy API (https://www.alchemy.com/docs/reference/simulation-examples) etc.

Local simulation requires:
- an additional process for running `anvil`
- additional time for launching the process then running the simulation which increases response time by the node
- existing fork in process needs to be restarted to update based on newer blocks i.e. we can't just have a background process that periodically refreshes blockchain state to latest block, but rather it would need to be restarted periodically.

Remote simulation requires:
- API keys and associated payment
- Can there be one key that all nodes use
- Centralizing force inflicted on nodes performing signing checks

In both cases, additional condition configuration options will need to be provided for any post-simulation checking of state.  Not sure where to go with this at the moment, so will look into it as part of a separate PR.

### DIFF
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -281,13 +281,17 @@ _COMPARATOR_FUNCTIONS = {
     "<": pyoperator.lt,
     "<=": pyoperator.le,
     ">=": pyoperator.ge,
-    "in": pyoperator.contains,  # currently only supports checking value in list (not sub-string/bytes comparisons)
+    # currently only supports checking value in list (not sub-string/bytes comparisons)
+    "in": pyoperator.contains,
+    "!in": lambda container, item: pyoperator.not_(
+        pyoperator.contains(container, item)
+    ),
 }
 
 
 class ConditionVariable(_Serializable):
     class Schema(CamelCaseSchema):
-        var_name = fields.Str(required=True)  # TODO: should this be required?
+        var_name = fields.Str(required=True)
         condition = _ConditionField(required=True)
 
         @post_load
@@ -585,10 +589,10 @@ class ReturnValueTest(_Serializable):
                 return
 
             comparator = data.get("comparator")
-            if comparator == "in" and not isinstance(value, list):
+            if comparator in ["in", "!in"] and not isinstance(value, list):
                 raise ValidationError(
                     field_name="value",
-                    message=f'"{type(value)}" is not a valid type for the "in" comparator; only list is allowed.',
+                    message=f'"{type(value)}" is not a valid type for the "in"/"!in" comparators; only list is allowed.',
                 )
 
         @post_load
@@ -675,9 +679,9 @@ class ReturnValueTest(_Serializable):
 
         processed_data = self._process_data(data, self.index)
         comparator_function = _COMPARATOR_FUNCTIONS.get(self.comparator)
-        if comparator_function == pyoperator.contains:
-            # for 'contains' operator, the left operand is the value and
-            # right operand is the data to check
+        if self.comparator in ["in", "!in"]:
+            # for 'contains'/`not_(contains)` operators, the left operand is
+            # the value (container) and the right operand is the data to check
 
             # first need to sanitize all values in list
             sanitized_list = [self._sanitize_value(v) for v in self.value]
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -281,9 +281,9 @@ _COMPARATOR_FUNCTIONS = {
     "<": pyoperator.lt,
     "<=": pyoperator.le,
     ">=": pyoperator.ge,
-    # currently only supports checking value in list (not sub-string/bytes comparisons)
-    "in": pyoperator.contains,
-    "!in": lambda container, item: pyoperator.not_(
+    # currently only supports checking value in list and not sub-string/bytes comparisons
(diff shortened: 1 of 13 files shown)


nucypher/utilities/abi.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            # account for nested structures i.e. tuples
+            if char == "(":
+                depth += 1
+            elif char == ")":
+                depth -= 1
+            current.append(char)
+    if current:
+        arg_types.append("".join(current).strip())
+
```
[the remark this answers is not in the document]
derekpierre: ✔️

derekpierre:
FYI the commit - https://github.com/nucypher/nucypher/pull/3609/commits/0eada39851c6fe9090717f938903f92b09419224 in this PR is related to the update here - https://github.com/nucypher/nucypher-contracts/pull/385/commits/32263a21c7a2cf5c4e9f33d89eea14aefccf97e5 made to `nucypher-contracts` via https://github.com/nucypher/nucypher-contracts/pull/385

vzotova (approved):

manumonti (approved):
Looks great! And thanks for the walktrough 🙏

nucypher/policy/conditions/lingo.py
```diff
@@ -675,9 +679,9 @@ def eval(self, data) -> bool:
 
         processed_data = self._process_data(data, self.index)
         comparator_function = _COMPARATOR_FUNCTIONS.get(self.comparator)
-        if comparator_function == pyoperator.contains:
-            # for 'contains' operator, the left operand is the value and
-            # right operand is the data to check
+        if self.comparator in ["in", "!in"]:
+            # for 'contains'/`not_(contains)` operators, the left operand is
```
```suggestion
            # for `in`/`!in` operators, the left operand is
```

nucypher/policy/conditions/lingo.py
```diff
@@ -281,13 +281,17 @@ def __init__(self, operand: Condition):
     "<": pyoperator.lt,
     "<=": pyoperator.le,
     ">=": pyoperator.ge,
-    "in": pyoperator.contains,  # currently only supports checking value in list (not sub-string/bytes comparisons)
+    # currently only supports checking value in list (not sub-string/bytes comparisons)
+    "in": pyoperator.contains,
+    "!in": lambda container, item: pyoperator.not_(
+        pyoperator.contains(container, item)
```
cygnusv: Not that it matters too much, since this is not used as an infix operator, but the order of parameters between `in` and `contains` perhaps should be the opposite. In other words, `A contains B` is equivalent to `B in A`.

nucypher/policy/conditions/lingo.py
```diff
@@ -675,9 +679,9 @@ def eval(self, data) -> bool:
 
         processed_data = self._process_data(data, self.index)
         comparator_function = _COMPARATOR_FUNCTIONS.get(self.comparator)
-        if comparator_function == pyoperator.contains:
-            # for 'contains' operator, the left operand is the value and
-            # right operand is the data to check
+        if self.comparator in ["in", "!in"]:
+            # for 'contains'/`not_(contains)` operators, the left operand is
+            # the value (container) and the right operand is the data to check
```
cygnusv: See comment above regarding order of operands.

nucypher/policy/conditions/lingo.py
derekpierre: ✔️

derekpierre:
Merging now - if there are any additional comments/suggestions/RFCs, still speak up and we can do them in a follow-up PR.

PR #3615, opened 2025-07-02
`lynx` lingo validation endpoint

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**
Based over #3616 

Adds a `/validate_condition_lingo` to *only* `lynx` nodes or nodes during testing (`temporary-domain`).

This endpoint will be used for `taco-web` integration tests to validate that the `zod` schema defined in typescript is consistent with the python `marshmallow` schema defined in `nucypher`.

Until now, we've been able to do that test by attaching conditions to encrypted data, then asking the node to decrypt the data. But when nodes sign data, the conditions are obtained from the `SigningCoordinator` contract and not from the signing request, so we can't just test signing with different conditions easily

**Notes for reviewers:**
Review whether there are alternative better strategies.

### DIFF
--- a/nucypher/network/server.py
+++ b/nucypher/network/server.py
@@ -18,7 +18,7 @@ from nucypher_core import (
 from prometheus_client import REGISTRY, Counter, Summary
 
 from nucypher.blockchain.eth import domains
-from nucypher.config.constants import MAX_UPLOAD_CONTENT_LENGTH
+from nucypher.config.constants import MAX_UPLOAD_CONTENT_LENGTH, TEMPORARY_DOMAIN_NAME
 from nucypher.crypto.keypairs import DecryptingKeypair
 from nucypher.crypto.signing import InvalidSignature
 from nucypher.network.nodes import NodeSprout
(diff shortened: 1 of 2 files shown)


nucypher/network/server.py
```diff
@@ -311,4 +313,20 @@ def status():
...
+        # only available on Lynx or for testing
+        @rest_app.route("/validate_condition_lingo", methods=["POST"])
+        def validate_condition_lingo():
+            """
+            An endpoint that validates a condition lingo
+            """
+            try:
+                _ = ConditionLingo.from_json(request.get_json())
+                return Response(response="valid", status=HTTPStatus.OK)
+            except Exception as e:
```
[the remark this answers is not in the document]
derekpierre: ✔️

nucypher/network/server.py
```diff
@@ -311,4 +314,20 @@ def status():
             return Response(response=html_error, headers=headers, status=HTTPStatus.INTERNAL_SERVER_ERROR)
         return Response(response=content, headers=headers)
 
+    if (
+        this_node.domain == domains.LYNX
```
theref:
```suggestion
        this_node.domain in [domains.LYNX, TEMPORARY_DOMAIN_NAME]
```

theref (approved):

nucypher/network/server.py
derekpierre: Good eye - a bit of nuance here.

For `lynx` I'm comparing with the domain struct as expected, but with temporary domain I'm comparing the name of the domain (`TEMPORARY_DOMAIN_NAME`) instead of the struct itself on purpose so that I don't import the `TEMPORARY_DOMAIN` constant which is located in a test module (`tests.constants`). I don't want to import a test module from within production code.

derekpierre: Best of both worlds 😄 - https://github.com/nucypher/nucypher/pull/3615/commits/b341b8ad4d162157aa249b89c574806030ba56c8

PR #3618, opened 2025-07-09
Default to byte-casting ECDSA hex `:message` values

KPrasch:
**Type of PR:**
Other

Closes https://github.com/nucypher/sprints/issues/184

**Required reviews:** 
2

### DIFF
--- a/nucypher/policy/conditions/ecdsa.py
+++ b/nucypher/policy/conditions/ecdsa.py
@@ -47,20 +47,17 @@ class ECDSAVerificationCall(ExecutionCall):
 
         @validates("message")
         def validate_message(self, value):
-            if not is_context_variable(value) and not isinstance(value, (str, bytes)):
+            if not is_context_variable(value) and not isinstance(value, str):
                 raise ValidationError(
-                    f"Invalid value for message; expected a context variable, string, or bytes but got '{value}'"
+                    f"Invalid value for message; expected a context variable, or string, but got '{value}'"
                 )
 
         @validates("signature")
         def validate_signature(self, value):
             if not is_context_variable(value):
-                try:
-                    bytes.fromhex(value)
-                except Exception as e:
-                    raise ValidationError(
-                        f"Invalid signature format, must be hex encoded: {str(e)}"
-                    )
+                raise ValidationError(
+                    f"Invalid value for signature; expected a context variable, but got '{value}'"
+                )
 
         @validates_schema
         def validate_verifying_key(self, data, **kwargs):
@@ -104,10 +101,7 @@ class ECDSAVerificationCall(ExecutionCall):
             message_value = resolve_any_context_variables(self.message, **context)
 
             # Handle message encoding with 0x prefix logic using HexBytes
-            if isinstance(message_value, bytes):
-                # Already bytes, use as-is
-                message = message_value
-            elif isinstance(message_value, str):
+            if isinstance(message_value, str):
                 # Only use HexBytes for 0x-prefixed strings
                 if message_value.startswith("0x"):
                     # 0x prefix indicates hex - use HexBytes for intelligent parsing
@@ -122,7 +116,7 @@ class ECDSAVerificationCall(ExecutionCall):
                     message = message_value.encode("utf-8")
             else:
                 raise ExecutionCall.InvalidExecutionCall(
-                    f"Message must be bytes or string, got {type(message_value)}"
+                    f"Message must be string, got {type(message_value)}"
                 )
 
             signature_hex = resolve_any_context_variables(self.signature, **context)
--- a/nucypher/policy/conditions/ecdsa.py
+++ b/nucypher/policy/conditions/ecdsa.py
@@ -4,6 +4,7 @@ from typing import Any, Optional, Tuple
 from ecdsa import BadSignatureError, NIST192p, VerifyingKey
 from ecdsa.curves import Curve, curves
 from ecdsa.util import sigdecode_string
+from hexbytes import HexBytes
 from marshmallow import (
     ValidationError,
     fields,
@@ -15,7 +16,6 @@ from marshmallow import (
 
 from nucypher.policy.conditions.base import Condition, ExecutionCall
 from nucypher.policy.conditions.context import (
-    USER_ADDRESS_CONTEXT,
     is_context_variable,
     resolve_any_context_variables,
 )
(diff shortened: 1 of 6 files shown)


nucypher/policy/conditions/ecdsa.py
```diff
@@ -114,12 +114,24 @@ def execute(self, **context) -> bool:
                 # Normal resolution for other cases
                 message = resolve_any_context_variables(self.message, **context)
 
-            signature_hex = resolve_any_context_variables(self.signature, **context)
-
-            # Ensure message is bytes
-            if isinstance(message, str):
+            # Special handling for :message context variable - treat as hex-encoded bytes
+            if self.message == ":message" and ":message" in context:
```
derekpierre: Is the case now that message is either a context variable or not?

so something like:
```python
message_value = self.message
if is_context_variable(self.message) {
   message_value = ...
} 

// process message_value
...
```

nucypher/policy/conditions/ecdsa.py
```diff
@@ -114,12 +114,24 @@ def execute(self, **context) -> bool:
...
+                message_value = context[":message"]
```
derekpierre: Let's use `resolve_any_context_variable` which is the proper way to resolve context variables instead of using the context dictionary directly i.e.

```python
message_value = resolve_any_context_variables(param=self.message, **context)
```

nucypher/policy/conditions/ecdsa.py
```diff
@@ -114,12 +114,24 @@ def execute(self, **context) -> bool:
...
+                if isinstance(message_value, str):
+                    try:
+                        # Try to decode as hex first
+                        message = bytes.fromhex(message_value)
+                    except ValueError:
+                        # If hex decoding fails, treat as regular string
+                        message = message_value.encode("utf-8")
```
derekpierre: If we want to accept either string or hex, does it make sense for the hex to be 0x prefixed then. That way it is much more explicit the difference between a hex string message, and a string message...?

Without the 0x prefix it can be quite ambiguous depending on the message, and we would encounter edge case issues eg. signing "deadbeef" which can be either a valid hex string or a raw string.

KPrasch: At the moment, I'm just making the assumption that `:message` values are always meant to be interpreted as bytes.  Just so I understand correctly, your suggestion here is to use the `0x` prefix to indicate bytes and the absence of the prefix to indicate a hexstring?

nucypher/policy/conditions/ecdsa.py
KPrasch: I'm glad you pointed this out.  This is not only ambiguous since message is a required field for `ECDSAVarificationCall` but also it's a bit inflexible.   Have a look at the new version. 👍🏻

nucypher/policy/conditions/ecdsa.py
KPrasch: refactored this handling ✔️

nucypher/policy/conditions/ecdsa.py
```diff
@@ -114,11 +114,20 @@ def execute(self, **context) -> bool:
                 # Normal resolution for other cases
                 message = resolve_any_context_variables(self.message, **context)
 
-            signature_hex = resolve_any_context_variables(self.signature, **context)
+            # Always resolve message and apply hex decoding for ECDSA conditions
```
derekpierre: Seems like lines 103-115 can now be deleted.

nucypher/policy/conditions/ecdsa.py
derekpierre: 👌

nucypher/policy/conditions/ecdsa.py
derekpierre: I guess I'm wondering about the logistics of the value for "message".  In the corresponding `taco-web` PR, "message" can be a string, or a context variable so it isn't limited to just a hex string (unless I'm misunderstanding). Also, given that it can be a context variable, we are at the mercy of what is provided to the context by the requester - was the custom context variable value provided as bytes, string, hex-string etc., and we probably want to make all options consistent.

Side Note: in `taco-web` if someone provides bytes for the custom context variable value (`addCustomContextParameterValues(...)`), it automatically gets converted to hex (`0x` prefixed) , so the user doesn't actually have to provide hex themselves, but it does mean that you can get a `0x` hex-string if the user provided bytes as a value for the context. See https://github.com/nucypher/taco-web/blob/signing-epic/packages/taco/src/utils.ts#L3.

This means you can currently receive, non-0x-prefixed hex string, 0x-prefixed hex string, string.

--

In the current code the logic is that "I assume the string is hex, try to convert it to bytes, and if that conversion fails I assume the string is just a string and encode it as bytes"

A couple things become tricky:
1. If the requester provided the context var value as bytes (`Uint8Array` in typescript) - then that becomes a 0x-prefixed hex string. And that fails the `bytes.fromhex(...)` check, and gets used as a raw string, in which case the signature check will fail.

2. While that strategy is clear for something like "1a2b3c" (this is hex) vs "Hello, I'm a string" (this is not hex), there are cases where the string is technically hex, but was meant to be a string eg. "deadbeef" so the resulting bytes would be different. It depends on what the user sets as the value for the context variable.

Suppose the data that was signed was `"deadbeef"` (so not hex string but the raw string). Of course, when signed, it probably got converted to bytes. But what is important is what is set in the context for ":message"? 

If the context var is set as "deadbeef" because that's the raw string that was provided, does the node interpret that as hex or a string? Well in this case the node would interpret it as hex since it can perform the conversion, but the resulting bytes would be different, and therefore the signature check would fail because the bytes wouldn't match what was signed.

```python
In [1]: hex_conversion = bytes.fromhex("deadbeef")

In [2]: bytes_conversion = "deadbeef".encode('utf-8')

In [3]: hex_conversion == bytes_conversion
Out[3]: False
```

Whereas if we use the '0x' prefix, we can more clearly assume hex string vs regular string i.e. "0xdeadbeef" vs "deadbeef".

--

There are two things to consider:
1. Should we allow raw strings as values for "message" property of the condition - or should it always be hex; either provided directly  by the condition creator or via a context variable whose value is to be provided by the user. Perhaps message should always be hex string derived from bytes and we should raise if it isn't valid hex? Then we would never have to deal with non-hex strings at all. 

2. Should we use 0x-prefix hex strings all the time, so that either the user provides 0x-prefixed hex strings for the message value directly, or for setting the context variable value they provide "bytes" (which gets converted to 0x-prefixed hex string) or their own 0x-prefixed hex string. Feels like this provides consistency.

manumonti (approved):
apart from Derek's comment, it looks good to me! 👌

derekpierre (approved):
🎸

PR #3620, opened 2025-07-17
Fix SequentialCondition edge case

derekpierre:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

Fix edge case where variable name collisions could occur in nested sequential conditions. Previously, duplicate detection only applied to a single level and did not account for nested `SequentialCondition`s. This update adds recursive checks to ensure uniqueness across all nested levels.

See added test case, which previously failed to raise an exception before this fix.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

Take a look at the unit test for clarity on what the problem was. The test did cause an exception before the code was updated.

### DIFF
--- a/newsfragments/3620.bugfix.rst
+++ b/newsfragments/3620.bugfix.rst
@@ -1,0 +1,1 @@
+Added recursive duplicate detection across nested sequential conditions.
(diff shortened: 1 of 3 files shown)


manumonti (approved):
Good catch!

KPrasch (approved):

derekpierre:
Updated the logic slightly.

manumonti (approved):
Even better. 🎩 solution. LGTM!

PR #3622, opened 2025-07-29
Makes `curve` a required parameter for ECDSA condition API

KPrasch:
**Rationale**

As the old adage goes: "explicit is better than implicit".   Particularly for cross-language interoperability and testing ease, and misuse resistance, using an explicit elliptic curve for ecdsa conditions is advantageous.


derekpierre (approved):
🎸

PR #3608, opened 2025-06-17
Handover: crypto, powers, agents and actors

cygnusv:
This PR introduces support for handover operations for Ursulas.

Issues opened by this PR:
* https://github.com/nucypher/nucypher/issues/3623
* https://github.com/nucypher/ferveo/issues/209
* https://github.com/nucypher/ferveo/issues/210

TODO: Still needs to specify dependencies on the new ferveo and nucypher-core

**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**Notes for reviewers:**
* Commit history was prepared with a lot of care, so going over it sequentially will be very useful.
* This integration has been done in layers: cryptographic layer -> powers -> agents -> actors
* As usual, looking at tests is the best way to start. In particular:
  * `tests/acceptance/actors/test_dkg_ritual.py`
  * `tests/acceptance/agents/test_coordinator_agent.py`
  * `tests/unit/crypto/test_ferveo.py`

PR #3628, opened 2025-08-01
Handover Updates

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Based over #3608 

Additional clean up changes to #3608, but nothing major.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


nucypher/blockchain/eth/actors.py
```diff
@@ -711,19 +760,20 @@ def perform_round_2(self, ritual_id: int, timestamp: int) -> Optional[AsyncTx]:
...
                 threshold=ritual.threshold,
                 shares=ritual.shares,
                 checksum_address=self.checksum_address,
                 ritual_id=ritual.id,
-                transcripts=messages,
+                validator_messages=messages,
             )
+            dkg_public_key = aggregated_transcript.public_key
+            # FIXME: Workaround: remove the public key (last 8 + 48 bytes of the aggregated transcript)
+            # to pass size validation check on contract publish. See ferveo#209
```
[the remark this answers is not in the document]
derekpierre: Ignore

nucypher/blockchain/eth/actors.py
```diff
@@ -750,6 +800,315 @@ def perform_round_2(self, ritual_id: int, timestamp: int) -> Optional[AsyncTx]:
...
+        self, ritual_id: int, departing_validator: ChecksumAddress
+    ) -> HandoverTranscript:
+        ritual = self._resolve_ritual(ritual_id)
+        validators = self._resolve_validators(ritual)
+
+        # Raises ValueError if departing_validator is not in the providers list
+        handover_slot_index = ritual.providers.index(departing_validator)
+        # FIXME: Workaround: add serialized public key to aggregated transcript.
+        # Since we use serde/bincode in rust, we need a metadata field for the public key, which is the field size,
+        # as 8 bytes in little-endian. See ferveo#209
```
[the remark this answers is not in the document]
derekpierre: Ignore

nucypher/blockchain/eth/actors.py
```diff
@@ -750,6 +800,315 @@ def perform_round_2(self, ritual_id: int, timestamp: int) -> Optional[AsyncTx]:
...
+        if handover_request.departing_validator != self.checksum_address:
+            self.log.debug(
+                f"Handover request for ritual #{ritual_id} is not for this node."
+            )
+            return False
+
+        handover_status = self.coordinator_agent.get_handover_status(
+            ritual_id=ritual_id, departing_validator=self.checksum_address
+        )
+        if (
```
vzotova: Maybe extract most of this method? looks and works very similar to `_is_handover_transcript_required`

vzotova (approved):

nucypher/blockchain/eth/actors.py
derekpierre: ✔️ - good catch.

nucypher/crypto/ferveo/dkg.py
```diff
@@ -99,3 +124,45 @@ def produce_decryption_share(
...
+    keypair: Keypair,
+    ritual_id: int,
+    shares: int,
+    threshold: int,
+) -> HandoverTranscript:
+    if not all((nodes, aggregated_transcript, keypair)):
+        raise Exception("missing arguments")  # sanity check
+
+    dkg = _make_dkg(
+        # TODO: is fixed 0-index fine here? I don't believe it matters
```
theref, line 142: could you explain more about this?

derekpierre: Sure. This relates to my comment here: https://github.com/nucypher/nucypher/pull/3608/files/#r2246248823.

Normally, to instantiate a `ferveo.Dkg` object, you pass in a list of validators (i.e. node metadata for all cohort participants) and the specific validator representing the current node (the `me` parameter), which should be one of the validators in the list of validators.

However, during handover, the _incoming validator_ needs to create a `ferveo.Dkg` object to generate the handover transcript, but it's not yet part of the cohort, so it isn’t in the validators list. This begs the question: what should be passed as `me` in this case?

In @cygnusv 's original PR, my understanding was that he simply used `validators[0]` as a placeholder because it doesn't really matter who `me` is in this case as long as it is a validator from the list. I’m following the same approach, but I’ve made a small change: since `me` was always `validators[0]`, I removed it as an explicit parameter from `ritual_power.initiate_handover(...)` (see: https://github.com/nucypher/nucypher/pull/3608/files/#diff-723cc2356b4ad693dd96ce95e1ab9e5f38bc3d01224c6e37ef6dd1dfa23b76cbR864) and instead deferred setting it to a default (i.e. `validators[0]`) deeper in the call stack, specifically where the underlying `ferveo` call is made, which is this line.

@cygnusv had a `FIXME` comment about it when he used `validators[0]` so I left a TODO here.

theref: 🚀 thank you

theref (approved):

PR #3630, opened 2025-08-07
Re-enable DKG storage caching with Handover

### LINKED ISSUE #3623: Potential outdated cache issue in DKG tracker after finalized handover
Until now, `Operator` caches ritual and validators information for future use. This was fine, but with handover, this information can change (i.e. aggregate transcripts and validator sets slightly change after a successful handover). We should reset caches on all ursulas in a cohort after a finalized handover. For the moment in #3608, we simply invalidate caching of these temporarily, which has a minor performace hit, but in the future this should be handled properly. The way to do it would be that all ursulas in a cohort listen for HandoverFinalized events, even if they were not involved, and update their caches.

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

Based over #3628 

- PR #3608 removed caching for validators and ritual metadata. This PR reintroduces that caching, with the motivation explained in #3623. To mitigate previous concerns:
   - The cache for the validator list and Ritual object will now be cleared during any handover event, regardless of whether the node is actively participating in the handover. This is necessary because the handover process modifies the values within these objects, and nodes should re-fetch accurate data from the contract when needed.
   - Cache invalidation is triggered on any handover event (not just HandoverFinalization) to account for the timing gap between when a handover completes and when the node receives the event. By clearing the cache proactively, we ensure consistency.
   - Once Handover is completed and `HandoverFinalization` event is received, the cache is cleared one more time so the values will be re-fetched the next time, and then the cache can be utilized.
   
- Handover phases were previously using the same caching keys as DKG ritual phase 2 in DKG storage. This PR assigns them their own distinct keys.

**Issues fixed/closed:**

Closes #3623 

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -937,6 +937,9 @@ class Operator(BaseActor):
         departing_participant: ChecksumAddress,
         **kwargs,
     ) -> Optional[AsyncTx]:
+        # clear ritual object and validators since handover modifies ritual
+        self.dkg_storage.clear_active_ritual_object(ritual_id)
+        self.dkg_storage.clear_validators(ritual_id)
 
         # check if there is a pending tx for this phase
         async_tx = self.dkg_storage.get_ritual_phase_async_tx(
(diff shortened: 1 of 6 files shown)


theref:
can we point this PR at #3608 instead?

derekpierre:
> can we point this PR at #3608 instead?

Do you mean point it to PR #3628 (which it is based over)?

derekpierre:
> can we point this PR at #3608 instead?

It's now rebased over v7.6.x, so should be easier to review, if that was the issue.

theref:
> > can we point this PR at #3608 instead?
> 
> It's now rebased over v7.6.x, so should be easier to review, if that was the issue.

thanks!

theref (approved):

vzotova (approved):

PR #3635, opened 2025-08-11
Handover fixes after testing

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Based over #3634 .

Fixes issues encountered while testing handover on `tapir`.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -142,7 +142,7 @@ jobs:
       # Only upload coverage files after all tests have passed
       - name: Upload unit tests coverage to Codecov
         if: matrix.python-version == '3.12'
-        uses: codecov/codecov-action@v4
+        uses: codecov/codecov-action@v5
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           files: unit-coverage.xml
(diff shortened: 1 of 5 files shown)


newsfragments/3635.dev.rst
theref, line 1: is this supposed to be empty?

theref (approved):
ok, so you're removing the cache stuff from a few different handover functions and re-factoring such that it's run whenever handover events are received?

newsfragments/3635.dev.rst
derekpierre: Handover feature work was documented here (#3608) - https://github.com/nucypher/nucypher/pull/3608/files#diff-23c996ea6762d11ef44a3ab417c503f9ccfc3ce340502f0707628bd5e2d0f258. The thinking here was that these are really fixes on top of not already released feature work that we've already done.

derekpierre:
> ok, so you're removing the cache stuff from a few different handover functions and re-factoring such that it's run whenever handover events are received?

Correct. Because handover eventually changes the composition of the cohort, the cache is cleared whenever any handover event is received to limit the situation where the cohort is updated but the local cache is not. There is some delay between handover happening and the node receiving the event, which is the reason why I clear the cache for all handover events and not just `HandoverFinalized`.

vzotova (approved):

PR #3638, opened 2025-08-21
Fix v7.7.x Ritual Tracking Post v7.6.0 Rebase

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [X] 2
- [ ] 3

**What this does:**
Updates needed post rebase of v7.7.x over `main` (just released v7.6.0) that included handover changes. The ritual tracker was significantly refactored in v7.7.x before the rebase as part of #3599 , and so there were some conflicts/issues with rebasing over v7.6.0. 

This PR aims to solve that problem after a basic rebase that caused ritual tracking to no longer work properly.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

Key changes include:
- Refactored participation state management to handle both DKG and handover events separately
- Fixed participation state storage to use string keys consistently
- Updated test methods to use new participation state value retrieval methods

### DIFF
--- a/nucypher/blockchain/eth/trackers/dkg.py
+++ b/nucypher/blockchain/eth/trackers/dkg.py
@@ -1,4 +1,4 @@
-from typing import Optional, Tuple
+from typing import Optional, Union
 
 from prometheus_client import REGISTRY, Gauge
 from web3.datastructures import AttributeDict
@@ -15,6 +15,9 @@ class DkgRitualTracker(RitualTracker):
     )
 
     class DkgParticipationState(RitualTracker.ParticipationState):
+        """
+        Participation state for handover rituals.
+        """
         def __init__(
             self,
             participating=False,
@@ -25,10 +28,15 @@ class DkgRitualTracker(RitualTracker):
             self.already_posted_transcript = already_posted_transcript
             self.already_posted_aggregate = already_posted_aggregate
 
-        def update(self, updated_state: "DkgParticipationState"):
-            self.participating = updated_state.participating
-            self.already_posted_transcript = updated_state.already_posted_transcript
-            self.already_posted_aggregate = updated_state.already_posted_aggregate
+    class HandoverParticipationState(RitualTracker.ParticipationState):
+        """
+        Participation state for handover rituals.
+        """
+
+        PREFIX = "handover-"
+
+        def __init__(self, participating=False):
+            super().__init__(participating)
 
     def __init__(
         self,
@@ -38,9 +46,10 @@ class DkgRitualTracker(RitualTracker):
         contract = operator.coordinator_agent.contract
 
         self.handover_event_types = [
-            contract.events.HandoverTranscriptPosted,
             contract.events.HandoverRequest,
+            contract.events.HandoverTranscriptPosted,
             contract.events.HandoverFinalized,
+            contract.events.HandoverCanceled,
         ]
 
         actions = {
@@ -49,7 +58,8 @@ class DkgRitualTracker(RitualTracker):
             contract.events.HandoverRequest: operator.perform_handover_transcript_phase,
             contract.events.HandoverTranscriptPosted: operator.perform_handover_blinded_share_phase,
         }
-        events = [
+
+        all_events = [
             contract.events.StartRitual,
             contract.events.StartAggregationRound,
             contract.events.EndRitual,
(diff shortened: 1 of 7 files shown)


nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -25,10 +28,15 @@ def __init__(
...
-        def update(self, updated_state: "DkgParticipationState"):
-            self.participating = updated_state.participating
-            self.already_posted_transcript = updated_state.already_posted_transcript
-            self.already_posted_aggregate = updated_state.already_posted_aggregate
+    class HandoverParticipationState(RitualTracker.ParticipationState):
+        """
+        Participation state for Handover rituals.
+        """
+
+        PREFIX = "handover-"
```
KPrasch, line 39: so any event with this prefix will be processed?

derekpierre: It's part of the key for state tracking.

We use the same tracker for tracking DKG events and Handover events. If we tried to track state using the same state object, it just complicates the logic. So instead we track state for those sets pf events separately. See _particiaptionState (TTL cache).

For state checking regarding DKG events (StartRitual, StartAggregationRound, EndRitual) just use ritualID as the key to store/get the state eg. "12", but for handover state checking regarding Handover events (HandoverRequested, HandoverTranscriptPosted, HandoverFinalized, HandoverCanceled) the key is `handover-rituaID` eg. "handover-12".

It's basically a way of making sure that DKG state is always separate from Handover state even for the same ritual.

The states are needed to prevent us from making a bunch of RPC calls during DKG and Handover Rituals.

tests/acceptance/actors/test_signing_ritual_tracker.py
```diff
@@ -0,0 +1,808 @@
...
+    # SigningCohortDeployed received for cohort id 3
+    #
+    event_data = AttributeDict(
+        {
+            "event": "SigningCohortDeployed",
+            "blockNumber": block_number,
+            "args": AttributeDict(
+                {
+                    "cohortId": cohort_id_3,
+                    "chainIid": chain_id,
```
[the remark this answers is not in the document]
derekpierre: ✔️

tests/acceptance/actors/test_signing_ritual_tracker.py
```diff
@@ -0,0 +1,808 @@
...
+    with patch.object(agent, "get_timeout", return_value=faked_ritual_timeout):
+        # fake timeout only needed for initialization
+        active_ritual_tracker = SigningRitualTracker(operator=ursula)
+
+    now = maya.now()
+
+    initiate_signing_cohort = agent.contract.events.InitiateSigningCohort()
+    args_dict = {
+        "cohortId": cohort_id_1,
+        "chain": chain_id,
```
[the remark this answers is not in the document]
derekpierre: ✔️

vzotova (approved):

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -62,47 +72,64 @@ def __init__(
...
+                    event_type == self.contract.events.HandoverFinalized
+                    and event.args.departingParticipant
+                    == self.operator.checksum_address
+                ):
+                    # if handover is canceled/finalized we need to reset the participation state
+                    participation_state.participating = False
+
+                # pruning metadata/cleanup is sufficient when Handover is finalized or canceled;
                 # no further action required
                 return False
```
manumonti: Pretty sure I'm misunderstanding the handover logic here, but is it actually possible to have a HandoverCanceled or HandoverFinalized but with the node still participating of it?

In other words: is this the same that just


```suggestion
            if event_type in [
                self.contract.events.HandoverFinalized,
                self.contract.events.HandoverCanceled,
            ]:
                # pruning metadata/cleanup is sufficient when Handover is finalized or canceled;
                # no further action required
                return False
```

manumonti (approved):
LGTM! although I have a suggestion

nucypher/blockchain/eth/trackers/dkg.py
derekpierre: > is it actually possible to have a HandoverCanceled or HandoverFinalized but with the node still participating of it?

Yes. This is the case for nodes that are in the cohort and not part of handover (neither departing or incoming participant). Participation pertains to the ~~signing~~ cohort not the handover.

> In other words: is this the same that just

Not quite.

The underlying logic for participation after the handover is finished is
1. If handover was canceled then:
a) if the node was the incoming participant, then it is not participating in the cohort
b) if the node was the departing participant, then it is still participating in the cohort (stays the same since set before)
c) if the node was neither then its participation remains the same (no change)
2. If handover was finalized then:
a) if the node was the departing participant, then it is no longer participating in that cohort
b) if the node was the incoming participant, then it is now participating in that cohort (stays the same since set before)
c) if the node was neither then its participation remains the same (no change)

You are right for cases b) and c) in both 1) and 2), but it is the a) case for 1) and 2) what this extra check in the code is performing.

manumonti: oooh, got it. Yes, as you are pointing, I thought participant meant that you are a participant of the handover instead of the ritual. Now it's clear, thanks.

derekpierre: I'll update some of the class documentation to help clarify that. Based on the naming I can see where the confusion came from.

derekpierre: @manumonti, added https://github.com/nucypher/nucypher/pull/3638/commits/550dcda3909e2576379a6c8c51e2b62d1c5f3919.

manumonti (approved):

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -62,47 +72,66 @@ def __init__(
...
             timeout=self.coordinator_agent.get_dkg_timeout(),
         )
 
     def _get_identifier(self, event: AttributeDict) -> str:
-        return event.args.ritualId
+        event_type = getattr(self.contract.events, event.event)
+        if event_type in self.handover_event_types:
+            return f"{self.CohortParticipationStateDuringHandover.PREFIX}{event.args.ritualId}"
+        else:
+            return str(event.args.ritualId)
```
cygnusv: Shouldn't DKG participation state have a prefix going forward too (e.g, "DKG")? For consistency purposes, mainly. Not a hard comment, though.

nucypher/blockchain/eth/trackers/dkg.py
```diff
@@ -62,47 +72,66 @@ def __init__(
...
+                self.contract.events.HandoverCanceled,
+            ]:
+                if (
+                    event_type == self.contract.events.HandoverCanceled
+                    and event.args.incomingParticipant == self.operator.checksum_address
+                ) or (
+                    event_type == self.contract.events.HandoverFinalized
+                    and event.args.departingParticipant
+                    == self.operator.checksum_address
+                ):
```
cygnusv: For increased readability, you could use `is_incoming_participant_in_handover` and `is_departing_participant_in_handover` from below here.

cygnusv (approved):

nucypher/blockchain/eth/trackers/dkg.py
derekpierre: It can if we want. I just didn't since the dkg state was the first and default from before.

✔️ good to remain consistent, so added commit - https://github.com/nucypher/nucypher/pull/3638/commits/3587494e33d0975cffd84d3a2bf181951f8ecde2.

nucypher/blockchain/eth/trackers/dkg.py
derekpierre: ✔️  - great catch

PR #3640, opened 2025-09-02
Post handover: logging improvements

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -373,11 +373,7 @@ class Operator(BaseActor):
         for i, staking_provider_address in enumerate(ritual.providers):
             if self.checksum_address == staking_provider_address:
                 # Local
-                external_validator = Validator(
-                    address=self.checksum_address,
-                    public_key=self.ritual_power.public_key(),
-                    share_index=i,
-                )
+                public_key = self.ritual_power.public_key()
             else:
                 # Remote
                 # TODO optimize rpc calls by obtaining public keys altogether
@@ -385,15 +381,15 @@ class Operator(BaseActor):
                 public_key = self.coordinator_agent.get_provider_public_key(
                     provider=staking_provider_address, ritual_id=ritual.id
                 )
-                self.log.debug(
-                    f"Ferveo public key for {staking_provider_address} is {bytes(public_key).hex()[:-8:-1]}"
-                )
-                external_validator = Validator(
-                    address=staking_provider_address,
-                    public_key=public_key,
-                    share_index=i,
-                )
-            result.append(external_validator)
+            self.log.debug(
+                f"Validator #{i}: {staking_provider_address}, public key: {bytes(public_key).hex()[:10]}"
+            )
+            validator = Validator(
+                address=staking_provider_address,
+                public_key=public_key,
+                share_index=i,
+            )
+            result.append(validator)
 
         self.dkg_storage.store_validators(ritual.id, result)
 
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -815,7 +815,7 @@ class Operator(BaseActor):
         if status != Coordinator.RitualStatus.ACTIVE:
             # This is a normal state when replaying/syncing historical
             # blocks that contain StartRitual events of pending or completed rituals.
-            self.log.debug(
+            self.log.info(
                 f"Ritual #{ritual_id} is not active so handover is not possible; dkg status={status}."
             )
             return False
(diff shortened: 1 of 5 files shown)


nucypher/blockchain/eth/actors.py
```diff
@@ -900,6 +907,9 @@ def _produce_handover_transcript(
             shares=ritual.dkg_size,
             threshold=ritual.threshold,
         )
+        self.log.info(
+            f"Resulting handover transcript (first 10 bytes): {bytes(handover_transcript)[:10].hex()}"
+        )
```
derekpierre, line 912: Is there something significant about the first 10 bytes...e.g if debugging it proves something, or just confirmation that it was created?

Do we need this log given that we log the same before publishing the transcript - feels like a dupe. wdyt?

tests/unit/crypto/test_ferveo.py
```diff
@@ -56,13 +59,38 @@ def test_handover_with_encrypt_and_decrypt(
     )
 
     # Server can aggregate the transcripts
+    messages = [
+        ValidatorMessage(a, Transcript.from_bytes(t))
```
derekpierre, line 63: Does `Transcript.from_bytes(t)` perform a validation? Is that why we serialize as bytes then deserialize from bytes?

derekpierre (approved):
🎸 - Just had some clarifying questions.

manumonti (approved):
LGTM! 👍

nucypher/blockchain/eth/actors.py
cygnusv: > Is there something significant about the first 10 bytes...e.g if debugging it proves something, or just confirmation that it was created?

Confirmation, mostly, and I simply chose 10 bytes to avoid printing the whole thing. But actually, there's an interesting value printed in the first 4 bytes, which is the share index used and that has some debugging interest.

> Do we need this log given that we log the same before publishing the transcript - feels like a dupe. wdyt?

Yes, that's true, we can revisit once we gain more certainty on the handover process.

tests/unit/crypto/test_ferveo.py
cygnusv: There's a "syntactic" validation, in the sense that you can't just deserialized random garbage, but it may be random elements following the expected Transcript format.

The reason for the serialization/deserialization was to model the test a bit more realistic, since nodes will serialize transcripts for publication and then deserialize them before aggregation.

cygnusv: Forgot to mention that validation is performed during aggregation, but there's a lingering issue in Ferveo that we need to resolve. I just opened https://github.com/nucypher/ferveo/issues/215 to track this. I have some partial improvements in a local Ferveo branch, but we need to iterate on that a bit more, since it has implications on how ferveo is consumed by upper layers.

tests/unit/crypto/test_ferveo.py
```diff
@@ -56,13 +59,38 @@ def test_handover_with_encrypt_and_decrypt(
...
+    )
+    bad_messages = [
+        ValidatorMessage(a, Transcript.from_bytes(t)) for a, t in bad_messages
+    ]
+
+    with pytest.raises(InvalidTranscriptAggregate):
+        bad_aggregate = dkg.aggregate_transcripts(
+            bad_messages
+        )  # TODO: should raise here
+        bad_aggregate.verify(validators_num, bad_messages)
```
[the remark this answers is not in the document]
derekpierre: Ignoring.

tests/unit/crypto/test_ferveo.py
```diff
@@ -56,13 +59,38 @@ def test_handover_with_encrypt_and_decrypt(
...
     # And the client can also aggregate and verify the transcripts
     client_aggregate = AggregatedTranscript(messages)
     assert client_aggregate.verify(validators_num, messages)
 
+    # If transcripts are ill-formed, aggregation fails
+    bad_messages = validator_messages_bytes.copy()
+    valid_but_random_coefficient = bytes.fromhex(
+        "92b00f7796596e1790cb573e3c3c106d16882e3f688462800b7403e22d89feb7fe4784481baff3fb85698124d32a7e9d"
+    )
+    position = 16 + 48  # position of transcript.coefficients[1]
```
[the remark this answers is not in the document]
derekpierre: Ignoring Copilot suggestion.

nucypher/blockchain/eth/actors.py
```diff
@@ -373,27 +373,23 @@ def _resolve_validators(
...
-                    f"Ferveo public key for {staking_provider_address} is {bytes(public_key).hex()[:-8:-1]}"
-                )
-                external_validator = Validator(
-                    address=staking_provider_address,
-                    public_key=public_key,
-                    share_index=i,
-                )
-            result.append(external_validator)
+            self.log.debug(
+                f"Validator #{i}: {staking_provider_address}, public key: {bytes(public_key).hex()[:10]}"
```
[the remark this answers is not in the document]
derekpierre: Ignoring Copilot suggestion.

PR #3641, opened 2025-09-02
Optional condition fields wiithout a value set (i.e. set to `None`) should not be included in the serialized json

derekpierre:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Based over #3638 .

Don't serialize `None` values when the field is optional.

`SKIP_VALUES` was really a means to an end to not serialize `None` values, but it was incomplete - we shouldn't serialize `None` values IF-AND-ONLY-IF the field is optional. Therefore, we don't really need `SKIP_VALUES` to be defined since the only values being skipped was `None` historically and instead we can add the logic for optional fields and just make this capability global to all Schemas.

Without this change the serialization of conditions taco-web would be different than python.

This is purely a serialization issue.  Given the focus of `taco-web` for client-side functionality, typically python does not do the serialization of conditions, only the deserialization; so this is a relatively minor issue, but caused a problem setting conditions on the contract since python was used for serializing the condition to bytes.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/newsfragments/3641.bugfix.rst
+++ b/newsfragments/3641.bugfix.rst
@@ -1,0 +1,1 @@
+Condition fields that are optional and have no value assigned (i.e., set to None) should be excluded from the serialized JSON.
--- a/nucypher/policy/conditions/base.py
+++ b/nucypher/policy/conditions/base.py
@@ -64,7 +64,6 @@ class Condition(_Serializable, ABC):
     CONDITION_TYPE = NotImplemented
 
     class Schema(CamelCaseSchema):
-        SKIP_VALUES = (None,)
         name = fields.Str(required=False, allow_none=True)
         condition_type = NotImplemented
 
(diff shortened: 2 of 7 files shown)


theref:
Most recent 3 commits look great! But this seems to have stuff from another branch in it?

derekpierre:
> Most recent 3 commits look great! But this seems to have stuff from another branch in it?

This PR is based over #3638; only the last 3 commits are applicable.

theref (approved):

vzotova (approved):

manumonti (approved):
LGTM! 👍

PR #3647, opened 2025-09-16
Event scanner robustness

derekpierre:
**Type of PR:**
- [x] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Closes https://github.com/nucypher/sprints/issues/243.

Related to issues we've noticed with Alchemy free tier.

_(From Copilot)_
- Adds retry logic with configurable chunk size reduction for failed RPC requests
- Implements special detection and handling for Alchemy free tier limitations
- Introduces environment variable configuration for chunk size limits

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**
> Explain how this PR fits in the greater context of the NuCypher Network.
> E.g., if this PR address a `nucypher/productdev` issue, let reviewers know!

**Notes for reviewers:**

~~Don't merge PR until `tapir` testing is completed. See https://github.com/nucypher/nucypher/pull/3647#issuecomment-3307643722~~

PR #3643, opened 2025-09-04
Use latest debian trixie release as the base for our docker builds

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Switch from older debian bullseye to newer debian trixie, which reduces vulnerabilities for our docker images. Further process improvements can be made but this is a quick and relatively impactful win.

Improvements:
- rust-python image improvement:

Screenshot 2025-09-04 at 12 35 24 PM

- corresponding nucypher image (`scout`) over the latest 7.6.0 image:

Screenshot 2025-09-04 at 1 28 29 PM

Updates:
- Use python 3.12.11
- `adduser` must be explicitly installed on debian trixie.

**Notes for reviewers:**

The failing lint and python-test jobs are unrelated and will be resolved by #3638

PR #3648, opened 2025-09-18
Use debian trixie release for base node images

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**

Based over #3647 .

Moves up changes from #3643 from `7.7.x` to `7.6.1.y`.

Switch from older debian bullseye to newer debian trixie, which reduces vulnerabilities for our docker images. Further process improvements can be made but this is a quick and relatively impactful win.

Improvements:
- rust-python image improvement:

Screenshot 2025-09-04 at 12 35 24 PM

- corresponding nucypher image (`scout`) over the latest 7.6.0 image:

Screenshot 2025-09-04 at 1 28 29 PM

Updates:
- Use python 3.12.11
- `adduser` must be explicitly installed on debian trixie.


vzotova (approved):

manumonti (approved):
LGTM! 👌

derekpierre:
FYI Latest commit in this PR is running on `tapir`.

nucypher/utilities/events.py
```diff
@@ -69,6 +92,29 @@ def write_events_to_csv_file(csv_file: Path,
     return True
 
 
+def is_alchemy_free_tier(web3: Web3, http_error: HTTPError) -> bool:
+    try:
+        # very specific error case
+        rpc_response = http_error.response.json()
+        is_http_400 = http_error.response.status_code == HTTPStatus.BAD_REQUEST
```
Muhammad-Altabba, line 99: Was this error code verified to be the one sent back by Aclchemy for this specific scenario?

In the documentation it seems the status code for such a situation would be 403.
In their docs the HTTP Code 400 Meaning is: `Bad Request: The request is invalid (e.g incorrect format)` . And the Example Solution is: `Verify your request body and format is correct.`.
While HTTP Code 403 Meaning is: `Forbidden: the server has received the request successfully, but is intentionally refusing to process it due to some missing criteria.`. And the Example Solution is: `Upgrade tiers to increase capacity - Create a new API key - Update allowlist to permit request origin.`.

https://www.alchemy.com/docs/reference/error-reference#http-status-codes

Muhammad-Altabba (dismissed):
Can this have integration test that at least can run locally, if not at the CD/CI?
Also added a comment as the status code does not seem to go along the docs.

nucypher/utilities/events.py
Muhammad-Altabba: I saw that 400 seems to be confirmed working as in https://github.com/nucypher/nucypher/pull/3647#issuecomment-3309683879.

However, having integration test is still helpful. If Alchemy later changed something in this regards, the integration test will fail and we will get to know about the issue faster, I think. Especially that the error message to contain the string `"Free tier"` is not mentioned in their docs: https://www.alchemy.com/docs/reference/error-reference#http-status-codes. And this may indicate a sudden change later similar to how they changed it now without an announcement.
The integration test just need to pass call Alchemy free tier url with the parameters to cause it to return the same `Free tier` error.

nucypher/utilities/events.py
```diff
@@ -360,9 +446,89 @@ def scan(
...
+            event_filter_params["toBlock"] = to_block_to_use
+            logger.debug(
+                f"Querying eth_getLogs with the following parameters: {event_filter_params}"
+            )
+            logs = web3.eth.get_logs(event_filter_params)
+            return logs, to_block_to_use
+        except HTTPError as http_error:
+            logger.warn(
+                f"eth_getLogs API call failed for range {from_block} - {to_block_to_use} ({to_block_to_use - from_block} blocks) on attempt {attempt + 1}/{max_retries}: {http_error}"
+            )
```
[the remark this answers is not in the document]
derekpierre: The logger is a `twisted` logger not a Python logger. The `warn` call in `twisted` is not deprecated.

nucypher/utilities/events.py
derekpierre: ✔️  - addressed in #3647 via https://github.com/nucypher/nucypher/pull/3647/commits/babcbf755c50d739cf505a1d6351b11f96ac10e0.

PR #3656, opened 2025-09-30
feat(conditions): Add JsonCondition for direct JSON evaluation

theref:
## Summary

This PR introduces **JsonCondition**, a new condition type that evaluates JSON data directly without requiring an HTTP request. This provides a simpler, more general alternative to `JsonApiCondition` for cases where JSON data is already available.

## Motivation

Currently, `JsonApiCondition` is designed for fetching JSON from HTTPS endpoints and then querying it. However, there are use cases where:
- JSON data is already available via Context
- Data comes from Sequential Condition

`JsonCondition` fills this gap by providing direct JSON evaluation with optional JSONPath queries.

### Core Implementation
1. **New `JsonCondition` class** (`nucypher/policy/conditions/json/json.py`)
   - Accepts JSON as dict, list, primitive, or JSON string
   - Optional JSONPath query support
   - Context variable resolution
   - Simpler architecture (inherits from `Condition` directly, not `ExecutionCallCondition`)
   - Uses standard `AnyField` for serialization

2. **Shared query logic** (`nucypher/policy/conditions/json/utils.py`)
   - Extracted `query_json_data()` utility function
   - Eliminates code duplication between `JsonCondition` and `JsonApiCondition`
   - Both condition types now use the same JSONPath query logic

3. **Type system integration**
   - Added `ConditionType.JSON` enum value
   - Added `JsonConditionDict` type definition
   - Registered in `ConditionLingo.resolve_condition_class()`

### Testing
- 11 comprehensive unit tests covering all functionality
- Tests include: initialization, validation, query execution, context variables, serialization, edge cases
- All existing tests still pass (351/351 condition tests passing)

## Usage Example

```python
from nucypher.policy.conditions.json.json import JsonCondition

# With dict and JSONPath query
condition = JsonCondition(
    data={"store": {"book": [{"price": 10.5}]}},
    query="$.store.book[0].price",
    return_value_test=ReturnValueTest("==", 10.5)
)

# With JSON string
condition = JsonCondition(
    data='{"value": 42}',
    query="$.value",
    return_value_test=ReturnValueTest("==", 42)
)

# With primitive (no query needed)
condition = JsonCondition(
    data=100,
    return_value_test=ReturnValueTest(">", 50)
)

# Verify
success, value = condition.verify()
```

### Atomic Commits
The changes are organized into 4 logical commits:
1. **Refactor**: Extract shared JSON query logic
2. **Feature**: Add JsonCondition implementation
3. **Chore**: Register in type system
4. **Test**: Add comprehensive tests

Each commit is self-contained and can be reviewed independently.

### Code Quality
- ✅ All pre-commit hooks passing (ruff, darker, etc.)
- ✅ No regressions - all 351 existing condition tests pass
- ✅ Full test coverage for new functionality
- ✅ Follows existing patterns and conventions

### Backward Compatibility
- ✅ No breaking changes to existing APIs
- ✅ `JsonApiCondition` behavior unchanged (just uses shared utility)
- ✅ All existing tests pass without modification

## Related Issues

This implements a more basic, general JSON condition type as discussed in the context of making TACo conditions more flexible and composable.

## Checklist

- [x] Code follows project style guidelines
- [x] Tests added and passing
- [x] No breaking changes
- [x] Documentation updated (newsfragment included)
- [x] Pre-commit hooks passing

### DIFF
--- a/nucypher/policy/conditions/json/json.py
+++ b/nucypher/policy/conditions/json/json.py
@@ -1,36 +1,28 @@
 import json
 from typing import Any, Optional, Tuple
 
-from marshmallow import fields, post_load, validate
+from marshmallow import ValidationError, fields, post_load, validate, validates
 
 from nucypher.policy.conditions.base import Condition
-from nucypher.policy.conditions.context import resolve_any_context_variables
+from nucypher.policy.conditions.context import (
+    is_context_variable,
+    resolve_any_context_variables,
+)
 from nucypher.policy.conditions.exceptions import InvalidCondition
 from nucypher.policy.conditions.json.base import JSONPathField
 from nucypher.policy.conditions.json.utils import (
     process_result_for_condition_eval,
     query_json_data,
 )
-from nucypher.policy.conditions.lingo import AnyField, ConditionType, ReturnValueTest
-
-
-def _parse_json_data(data: Any) -> Any:
-    """
-    Parse JSON data if it's a string, otherwise return as-is.
-    """
-    if not isinstance(data, str):
-        return data
-
-    try:
-        return json.loads(data)
-    except ValueError as e:
-        raise InvalidCondition(f"Invalid JSON string: {e}") from e
+from nucypher.policy.conditions.lingo import ConditionType, ReturnValueTest
 
 
 class JsonCondition(Condition):
     """
-    A JSON condition evaluates data that is directly provided as JSON.
-    The data can be a dict, list, primitive value, or a JSON string.
+    A JSON condition evaluates JSON-compatible data from a context variable.
+
+    The data must be provided as a context variable (e.g., ':previousResult')
+    which typically comes from a previous condition in a Sequential workflow.
     An optional JSONPath query can be applied to extract a specific value.
     """
 
@@ -40,9 +32,16 @@ class JsonCondition(Condition):
         condition_type = fields.Str(
             validate=validate.Equal(ConditionType.JSON.value), required=True
         )
-        data = AnyField(required=True)
+        data = fields.Str(required=True)
         query = JSONPathField(required=False, allow_none=True)
         return_value_test = fields.Nested(ReturnValueTest.Schema(), required=True)
+
+        @validates("data")
+        def validate_data(self, value):
+            if not is_context_variable(value):
+                raise ValidationError(
+                    f"Invalid value for data; expected a context variable, but got '{value}'"
+                )
 
         @post_load
         def make(self, data, **kwargs):
@@ -50,14 +49,13 @@ class JsonCondition(Condition):
 
     def __init__(
         self,
-        data: Any,
+        data: str,
         return_value_test: ReturnValueTest,
         query: Optional[str] = None,
         condition_type: Optional[str] = ConditionType.JSON.value,
         name: Optional[str] = None,
     ):
-        # Parse JSON string if needed
-        self.data = _parse_json_data(data)
+        self.data = data
         self.query = query
         self.return_value_test = return_value_test
 
@@ -69,6 +67,15 @@ class JsonCondition(Condition):
         """
         # Resolve context variables in data if needed
         resolved_data = resolve_any_context_variables(self.data, **context)
+
+        # If resolved data is a JSON string, parse it
+        if isinstance(resolved_data, str):
+            try:
+                resolved_data = json.loads(resolved_data)
+            except (json.JSONDecodeError, ValueError) as e:
+                raise InvalidCondition(
+                    f"Context variable '{self.data}' contains invalid JSON string: {e}"
+                ) from e
 
         # Apply JSONPath query
         result = query_json_data(resolved_data, self.query, **context)
--- a/nucypher/policy/conditions/json/json.py
+++ b/nucypher/policy/conditions/json/json.py
@@ -23,7 +23,7 @@ def _parse_json_data(data: Any) -> Any:
 
     try:
         return json.loads(data)
-    except (json.JSONDecodeError, ValueError) as e:
+    except ValueError as e:
         raise InvalidCondition(f"Invalid JSON string: {e}") from e
 
 
--- a/nucypher/policy/conditions/json/json.py
+++ b/nucypher/policy/conditions/json/json.py
@@ -1,0 +1,85 @@
+import json
+from typing import Any, Optional, Tuple
+
+from marshmallow import fields, post_load, validate
+
+from nucypher.policy.conditions.base import Condition
+from nucypher.policy.conditions.context import resolve_any_context_variables
+from nucypher.policy.conditions.exceptions import InvalidCondition
+from nucypher.policy.conditions.json.base import JSONPathField
+from nucypher.policy.conditions.json.utils import (
+    process_result_for_condition_eval,
+    query_json_data,
+)
+from nucypher.policy.conditions.lingo import AnyField, ConditionType, ReturnValueTest
+
+
+def _parse_json_data(data: Any) -> Any:
+    """
+    Parse JSON data if it's a string, otherwise return as-is.
+    """
+    if not isinstance(data, str):
+        return data
+
+    try:
+        return json.loads(data)
+    except (json.JSONDecodeError, ValueError) as e:
+        raise InvalidCondition(f"Invalid JSON string: {e}") from e
+
+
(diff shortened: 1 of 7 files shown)


nucypher/policy/conditions/json/json.py
```diff
@@ -0,0 +1,85 @@
...
+def _parse_json_data(data: Any) -> Any:
+    """
+    Parse JSON data if it's a string, otherwise return as-is.
+    """
+    if not isinstance(data, str):
+        return data
+
+    try:
+        return json.loads(data)
+    except (json.JSONDecodeError, ValueError) as e:
```
derekpierre: `JSONDecodeError` is already a subclass of `ValueError`

```suggestion
    except (ValueError) as e:
```

nucypher/policy/conditions/json/json.py
```diff
@@ -0,0 +1,85 @@
...
+    An optional JSONPath query can be applied to extract a specific value.
+    """
+
+    CONDITION_TYPE = ConditionType.JSON.value
+
+    class Schema(Condition.Schema):
+        condition_type = fields.Str(
+            validate=validate.Equal(ConditionType.JSON.value), required=True
+        )
+        data = AnyField(required=True)
```
derekpierre: Would this simply always be a context variable? 🤔 

Analogous to `authorization_token` for `JsonApiCall` - https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/policy/conditions/json/api.py#L49

nucypher/policy/conditions/json/json.py
```diff
@@ -0,0 +1,85 @@
...
+        query = JSONPathField(required=False, allow_none=True)
```
derekpierre, line 36: Should this be required? Is there a case for it not needing to be used?

It wasn't required for JSON Rest endpoints because those can return values and not JSON.

If you deem it required, then update the constructor type to not be Optional.

nucypher/policy/conditions/json/json.py
```diff
@@ -0,0 +1,85 @@
...
+        result = query_json_data(resolved_data, self.query, **context)
+
+        # Process result for evaluation (handles string quoting)
+        result_for_eval = process_result_for_condition_eval(result)
+
+        # Evaluate against return value test
+        resolved_return_value_test = self.return_value_test.with_resolved_context(
+            **context
+        )
+        eval_result = resolved_return_value_test.eval(result_for_eval)
```
derekpierre, line 90: 🚀

nucypher/policy/conditions/types.py
```diff
@@ -60,6 +60,11 @@ class ContractConditionDict(RPCConditionDict):
     functionAbi: NotRequired[ABIFunction]
 
 
+class JsonConditionDict(BaseExecConditionDict):
+    data: Any
+    query: NotRequired[str]
```
derekpierre, line 65: Update if `query` deemed required.

derekpierre (commented):
Nice work - some questions/suggestions.

derekpierre:
Let's see if we can cover most/all of the missing 6 lines for code coverage in json/utils.py 😉 :

Screenshot 2025-09-30 at 1 57 30 PM

(Assuming the coverage report is correct, I'm surprised those lines aren't covered by other tests).

nucypher/policy/conditions/json/json.py
```diff
@@ -0,0 +1,85 @@
...
+    process_result_for_condition_eval,
+    query_json_data,
+)
+from nucypher.policy.conditions.lingo import AnyField, ConditionType, ReturnValueTest
+
+
+def _parse_json_data(data: Any) -> Any:
+    """
+    Parse JSON data if it's a string, otherwise return as-is.
+    """
```
manumonti: I'm wondering if we could manage data as a string (not a JSON string). For example, data is an address `data = "0xc0ffee254729296a45a3885639AC7E10F9d54979"`, or `data = "enabled"`

manumonti (approved):
Looks great to me. I have a suggestion though. 

BTW, I'm super fan of how this PR looks. Commits are clear and concise, with well-formed messages. The PR message adds all the necessary context and makes clear the purpose of the changes and how they were made. Awesome.

nucypher/policy/conditions/json/json.py
theref: you're right, I can't think of any situation where this wouldn't be populated by context in some way. For some reason I don't want to enforce that though

nucypher/policy/conditions/json/json.py
theref: The JSON could also just return a value, I think there's a test for that...if there's not I'll add one

nucypher/policy/conditions/json/json.py
derekpierre: > For some reason I don't want to enforce that though

Could you expand on your apprehension? If not a context variable, it means that it could be hardcoded - is there a legit case for that?

nucypher/policy/conditions/json/json.py
derekpierre: Fair point. I guess in the case of it just forwarding along a JSON response it could just be value.

There's an prior test for JsonApiCondition for just a value - see https://github.com/nucypher/nucypher/blob/main/tests/unit/conditions/test_json_api_condition.py#L57

nucypher/policy/conditions/json/json.py
```diff
@@ -0,0 +1,85 @@
...
+    try:
+        return json.loads(data)
+    except ValueError as e:
+        raise InvalidCondition(f"Invalid JSON string: {e}") from e
+
+
+class JsonCondition(Condition):
+    """
+    A JSON condition evaluates data that is directly provided as JSON.
+    The data can be a dict, list, primitive value, or a JSON string.
```
cygnusv: I don't understand the motivation for this. Shouldn't we restrict this to strings?

tests/unit/conditions/test_json_condition.py
```diff
@@ -0,0 +1,194 @@
...
+    assert condition.data == {"value": 42}
+
+
+def test_json_condition_with_primitive():
+    condition = JsonCondition(
+        data=42,
+        return_value_test=ReturnValueTest("==", 42),
+    )
+    assert condition.data == 42
+    assert condition.query is None
```
cygnusv: Related to my comment above. I know that `"42"` is a valid JSON string, but `42` is just an integer.

theref: I thnk `42` is also valid json, it's certainly accepted by a lot of json parsers, including the one we use.

nucypher/policy/conditions/json/json.py
theref: those are both valid json, so would be handled fine

nucypher/policy/conditions/json/json.py
theref: it's so that we can handle sequential conditions, where the json comes in already parsed as an object

cygnusv: Oh I see. Thanks for the explanation 👍

tests/unit/conditions/test_json_condition.py
cygnusv: Right, my point was more about acceptable types (str vs integers), but since you mentioned before that with sequential conditions JSON data may be already parsed as objects, then I have no further comment 👍

cygnusv (approved):

nucypher/policy/conditions/json/json.py
theref: i think it's just overly restrictive - what do we gain by restricting it to only context variables?

cygnusv: As a reference, the `jwt_token` field in `JwtCondition` is required to be a context variable.

theref: ok, didn't know that :) I don't feel overly strongly on this, so I'm happy to defer and follow you both

derekpierre: > i think it's just overly restrictive - what do we gain by restricting it to only context variables?

One gain is avoiding backwards compatibility issues.

It's easier to go from allowing only context variable to allowing anything, vs allowing anything (and potential user misuse) and then trying to restrict to only context variables.

derekpierre (approved):
🎸

PR #3658, opened 2025-09-30
Variable/Value operations before evaluation

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Allow capability to perform transformations on variables/values for `SequentialCondition` and `ReturnValueTest` before evaluation.
- This capability is provided via the `VariableOperation` object associated with `ConditionVariable` (`SequentialCondition`) and `ReturnValueTest` 

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/policy/conditions/json/utils.py
+++ b/nucypher/policy/conditions/json/utils.py
@@ -49,11 +49,6 @@ def query_json_data(data: Any, query: Optional[str], **context) -> Any:
             f"JSONPath error: {jsonpath_err}"
         ) from jsonpath_err
 
-    # if len(matches) > 1:
-    #     message = f"Ambiguous JSONPath query - multiple matches found for: {resolved_query}"
-    #     self.logger.info(message)
-    #     raise JsonRequestException(message)
-
     if len(matches) > 1:
         result = [match.value for match in matches]
     else:
--- a/nucypher/policy/conditions/json/utils.py
+++ b/nucypher/policy/conditions/json/utils.py
@@ -31,10 +31,7 @@ def query_json_data(data: Any, query: Optional[str], **context) -> Any:
     from jsonpath_ng.ext import parse
 
     from nucypher.policy.conditions.context import resolve_any_context_variables
-    from nucypher.policy.conditions.exceptions import (
-        ConditionEvaluationFailed,
-        JsonRequestException,
-    )
+    from nucypher.policy.conditions.exceptions import ConditionEvaluationFailed
 
     resolved_query = resolve_any_context_variables(query, **context)
     logger = Logger(__name__)
@@ -52,12 +49,14 @@ def query_json_data(data: Any, query: Optional[str], **context) -> Any:
             f"JSONPath error: {jsonpath_err}"
         ) from jsonpath_err
 
-    if len(matches) > 1:
-        message = (
-            f"Ambiguous JSONPath query - multiple matches found for: {resolved_query}"
-        )
-        logger.info(message)
-        raise JsonRequestException(message)
+    # if len(matches) > 1:
+    #     message = f"Ambiguous JSONPath query - multiple matches found for: {resolved_query}"
+    #     self.logger.info(message)
+    #     raise JsonRequestException(message)
 
-    result = matches[0].value
+    if len(matches) > 1:
+        result = [match.value for match in matches]
+    else:
+        result = matches[0].value
+
     return result
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -1,5 +1,6 @@
 import ast
 import base64
+import decimal
 import json
 import math
 import operator as pyoperator
@@ -8,6 +9,7 @@ from enum import Enum
 from hashlib import md5
 from typing import Any, List, Optional, Tuple, Type, Union
 
+from eth_utils import currency
 from hexbytes import HexBytes
 from marshmallow import (
     Schema,
@@ -294,6 +296,21 @@ _COMPARATOR_FUNCTIONS = {
 }
 
 
+def _eth_to_wei(a, _):
+    try:
+        return currency.to_wei(a, "ether")
+    except decimal.InvalidOperation as e:
+        raise TypeError(f"Invalid value for ethToWei conversion: {a}") from e
+
+
+def _wei_to_eth(a, _):
+    try:
+        return currency.from_wei(a, "ether")
+    except decimal.InvalidOperation as e:
+        raise TypeError(f"Invalid value for weiToEth conversion: {a}") from e
+
+
+# should raise TypeError for invalid inputs
 _OPERATOR_FUNCTIONS = {
     "+=": pyoperator.add,
     "-=": pyoperator.sub,
@@ -307,6 +324,7 @@ _OPERATOR_FUNCTIONS = {
     "abs": lambda a, _: abs(a),
     "avg": lambda a, _: statistics.mean(a),
     "ceil": lambda a, _: math.ceil(a),
+    "ethToWei": _eth_to_wei,
     "floor": lambda a, _: math.floor(a),
     "len": lambda a, _: len(a),
     "max": lambda a, _: max(a),
@@ -314,6 +332,7 @@ _OPERATOR_FUNCTIONS = {
     "min": lambda a, _: min(a),
     "mode": lambda a, _: statistics.mode(a),
     "sum": lambda a, _: sum(a),
+    "weiToEth": _wei_to_eth,
     # casting
     "str": lambda a, _: str(a),
     "int": lambda a, _: int(a),
@@ -325,6 +344,7 @@ _OPERATORS_WITH_NO_VALUE = {
     "abs",
     "avg",
     "ceil",
+    "ethToWei",
     "floor",
     "len",
     "max",
@@ -336,6 +356,7 @@ _OPERATORS_WITH_NO_VALUE = {
     "int",
     "float",
     "bool",
+    "weiToEth",
 }
 
 
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -2,6 +2,7 @@ import ast
 import base64
 import json
 import operator as pyoperator
+import statistics
 from enum import Enum
 from hashlib import md5
 from typing import Any, List, Optional, Tuple, Type, Union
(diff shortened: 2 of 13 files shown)


Muhammad-Altabba (commented):
I am not much into python. But this sounds great!

nucypher/policy/conditions/lingo.py
```diff
@@ -1,11 +1,15 @@
 import ast
```
Muhammad-Altabba, line 1: I think what Copilot suppressed due to low confidence is valid

> <summary>Comments suppressed due to low confidence (1)</summary>
> 
> **nucypher/policy/conditions/lingo.py:1**
> * The import of `RequiredContextVariable` is unused in this file. Consider removing it to keep imports clean.
> ```
> import ast
> ```

derekpierre: Perhaps I'm misunderstanding, but it's unclear what the issue is. `lingo.py` does not have an import for `RequiredContextVariable`. If you have an idea, I'm all ears.

nucypher/policy/conditions/json/utils.py
```diff
@@ -53,11 +50,8 @@ def query_json_data(data: Any, query: Optional[str], **context) -> Any:
         ) from jsonpath_err
 
     if len(matches) > 1:
-        message = (
-            f"Ambiguous JSONPath query - multiple matches found for: {resolved_query}"
-        )
-        logger.info(message)
-        raise JsonRequestException(message)
+        result = [match.value for match in matches]
```
theref, line 53: how come we're changing this?

nucypher/policy/conditions/lingo.py
```diff
@@ -291,18 +296,171 @@ def __init__(self, operand: Condition):
...
+    # We can add all kinds of operators over time, this is just a base start - given that
+    # we need ethToWei and weiToEth.
+    # Whether this set is too aggressive or not remains to be seen. We can always pare
+    # back on the operators we want to start with.
+    "+=": pyoperator.add,
+    "-=": pyoperator.sub,
+    "*=": pyoperator.mul,
+    "/=": pyoperator.truediv,
+    "%=": pyoperator.mod,
+    "^=": pyoperator.pow,
```
theref: are we doing anything about overly "expensive" conditions?

nucypher/policy/conditions/lingo.py
```diff
@@ -291,18 +296,171 @@ def __init__(self, operand: Condition):
...
+    "sum": lambda a, _: sum(a),
+    "weiToEth": _wei_to_eth,
+    # casting
+    "bool": lambda a, _: bool(a),
+    "float": lambda a, _: float(a),
+    "int": lambda a, _: int(a),
+    "str": lambda a, _: str(a),
+}
+
+_OPERATORS_WITH_NO_VALUE = {
```
theref: what do you mean "no value"?

nucypher/policy/conditions/lingo.py
```diff
@@ -709,7 +879,29 @@ def eval(self, data) -> bool:
...
+        # perform any additional operations before comparison
+        if self.operations:
+            try:
+                data = VariableOperation.calc_from_list(self.operations, data)
+            except Exception as e:
+                raise ReturnValueEvaluationError(
+                    f"Error performing operations on returned data: {e}"
+                )
+
+        processed_data = self._process_data(data)
```
theref, line 924: i'm unsure about processing after operations. Don't we need to convert types and then do operations?

nucypher/policy/conditions/lingo.py
```diff
@@ -671,31 +856,16 @@ def _sanitize_value(cls, value):
     def __handle_potential_bytes(data: Any) -> Any:
         return HexBytes(data).hex() if isinstance(data, bytes) else data
 
-    def _process_data(self, data: Any, index: Optional[int]) -> Any:
+    def _process_data(self, data: Any) -> Any:
```
theref, line 880: Can you explain why index is removed?

nucypher/policy/conditions/json/utils.py
derekpierre: Previously, we allowed only 1 result to be returned from a Json endpoint, because the condition could only check the value returned.  Now, because of operations, we can actually check a value based on multiple outputs from the JSON because we can perform operations on it (eg. `sum`). 

The cleanest example is this test -  https://github.com/derekpierre/nucypher/blob/variable-operations/tests/unit/conditions/test_signing_condition.py#L862 - which mimics the Engages API Rest endpoint. This is case where the user has points attributed to the different guilds they belong to, and the amount of tokens they can claim in the TGE is a function of the total points they have (in this test it is total points == allowed tokens in wei in the UserOp)

nucypher/policy/conditions/lingo.py
derekpierre: ✔️ - good question. It is now limited to a max of 5 operations.

nucypher/policy/conditions/lingo.py
derekpierre: It doesn't take an additional parameter i.e. no `b` value for the lambda function.

✔️ Renamed to `UNARY_OPERATOR_FUNCTIONS`.

nucypher/policy/conditions/lingo.py
derekpierre: We need to pull out the value at index earlier before the processing since it may need to undergo operations first. This is done on L823.

This `_process_data` is for modifying the value (bytes to hex, perhaps others in the future) before the comparison is performed.

Updated the function description.

nucypher/policy/conditions/lingo.py
derekpierre: The `_process_data` is the final step before evaluation/comparison. This step is for internal use and is specifically to facilitate `ast.literal_eval` when comparing byte values, and is invisible to the caller (we don't document this). That's why all operations are performed on the raw values returned from the condition (which the caller is aware of) before this internal processing takes place.

derekpierre (commented):
Great questions.

nucypher/policy/conditions/json/utils.py
theref: love it, makes complete sense

nucypher/policy/conditions/lingo.py
theref: it's actually something like this that worries me `[{"operation": "^=", "value": 999999}` - might be worth testing to see if it errors gracefully or crashes

theref (approved):

nucypher/policy/conditions/lingo.py
```diff
@@ -291,18 +296,180 @@ def __init__(self, operand: Condition):
...
 
 
+def _eth_to_wei(a, _):
+    try:
+        return currency.to_wei(a, "ether")
+    except decimal.InvalidOperation as e:
+        raise TypeError(f"Invalid value for ethToWei conversion: {a}") from e
+
+
+def _wei_to_eth(a, _):
```
```suggestion
def _wei_to_eth(a, _) -> decimal.Decimal:
```

nucypher/policy/conditions/lingo.py
```diff
@@ -291,18 +296,180 @@ def __init__(self, operand: Condition):
 }
 
 
+def _eth_to_wei(a, _):
```
```suggestion
def _eth_to_wei(a, _) -> int:
```

cygnusv:
We have to be very careful with what operations we allow here. There's many potential traps (e.g. floating point precision issues, overloaded operators, etc) that we need to weigh very carefully. Please see my branch https://github.com/cygnusv/nucypher/tree/variable-operations, and in particular the last 2 commits with some examples.

nucypher/policy/conditions/lingo.py
KPrasch: ☝🏻 Valid concern - even if memory holds, Python will keep crunching a ** b until timeout. CPU time grows super-linearly with the result size.  some options to consider:

1) `pow(a, b, m)` (specifically the three-arg usage) is the safe form (O(log b) time, constant memory), but not usable for exact equality comparison 
2) Add a pre-check to estimate digits and bail if it’s too large
3) Simplest fix: disable exponent transforms for now until we define safe limits or a modular form

theref: another option is a timeout:
```python
@contextmanager
def timeout(seconds=1):
    """Context manager that raises OperationTimeout if code takes too long."""
    def timeout_handler(signum, frame):
        raise OperationTimeout(f"Operation exceeded {seconds} second timeout")
    
    # Set the signal handler
    old_handler = signal.signal(signal.SIGALRM, timeout_handler)
    signal.alarm(seconds)
    try:
        yield
    finally:
        signal.alarm(0)
        signal.signal(signal.SIGALRM, old_handler)
```
expensive operations could then be wrapped:
```python
class VariableOperation(_Serializable):
    OPERATION_TIMEOUT_SECONDS = 1  # 1 second max per operation
    
    def calc(self, variable_value: Any):
        """Calculate the result of the operation on the variable value."""
        operation_function = _OPERATOR_FUNCTIONS[self.operation]
        
        try:
            with timeout(self.OPERATION_TIMEOUT_SECONDS):
                return operation_function(variable_value, self.value)
        except OperationTimeout as e:
            raise ConditionEvaluationFailed(
                f"Operation '{self.operation}' with value '{self.value}' exceeded timeout. "
                f"Result would be too expensive to compute."
            ) from e
```
the timeout is only be applied to a single operation function. And any single maths operation that's taking >1-2 secs is sus

cygnusv:
Created a PR to Derek's branch with some ideas, but I think this requires discussion with Derek:
https://github.com/derekpierre/nucypher/pull/9

nucypher/policy/conditions/lingo.py
Muhammad-Altabba: Sorry, I was confused with something else. And there is no such issue.

nucypher/policy/conditions/lingo.py
```diff
@@ -291,18 +296,180 @@ def __init__(self, operand: Condition):
...
+        @validates_schema
+        def validate_operation_and_value(self, data, **kwargs):
+            operation = data["operation"]
+            value = data.get("value")
+            if operation in _UNARY_OPERATOR_FUNCTIONS:
+                if value is not None:
+                    raise ValidationError(
+                        field_name="value",
+                        message=f'No value should be provided for operation "{operation}"',
+                    )
```
manumonti: I have a suggestion here. Instead of having a list of operators-without-2nd-value, which sounds to me like code-redundant, we could do something like this. (I have not tested it, though).

```suggestion
from inspect import signature  # at the begginning of the file 

class VariableOperation(_Serializable):
    class Schema(CamelCaseSchema):
        operation = fields.Str(
            required=True,
            validate=OneOf(_OPERATOR_FUNCTIONS, error="Not a permitted operation"),
        )
        value = AnyField(required=False, allow_none=True)

        @validates_schema
        def validate_operation_and_value(self, data, **kwargs):
            operation = data["operation"]
            value = data.get("value")
            if len(signature(operation).parameters) == 1:
                if value is not None:
                    raise ValidationError(
                        field_name="value",
                        message=f'No value should be provided for operation "{operation}"',
                    )
```

nucypher/policy/conditions/lingo.py
```diff
@@ -291,18 +296,180 @@ def __init__(self, operand: Condition):
...
+
+    @classmethod
+    def calc_from_list(cls, operations: List["VariableOperation"], variable_value: Any):
+        """
+        Calculate the result of a list of operations on the variable value.
+        """
+        result = variable_value
+        for operation in operations:
+            result = operation.calc(result)
+        return result
```
manumonti: I'm wondering if we can find a more descriptive name for this function.

- calc_sequentially_form_list
- apply_sequential_operations
- calc_operation_list
- apply_operations_sequentially

WDYT?

tests/unit/conditions/test_variable_operation.py
```diff
@@ -0,0 +1,161 @@
...
+    ("floor", None, 3.9, 3),
+    ("index", 1, [10, 20, 30], 20),
+    ("index", 0, [10, 20, 30], 10),
+    ("index", 2, [10, 20, 30], 30),
+    (
+        "index",
+        4,
+        ["Proper", "preparation", "prevents", "poor", "performance"],
+        "performance",
+    ),  # -- Ray Lewis
```
manumonti, line 34: 🤓

tests/unit/conditions/test_signing_condition.py
```diff
@@ -848,3 +857,160 @@ def test_signing_object_abi_attribute_condition_lingo_json_serialization(
...
+
+    sequential_condition = SequentialCondition(
+        condition_variables=[
+            ConditionVariable(var_name="ecdsa", condition=ecdsa_condition),
+            ConditionVariable(
+                var_name="points",
+                condition=json_condition,
+                operations=[
+                    VariableOperation(operation="sum"),
+                    VariableOperation(operation="*=", value=10**18),  # convert to wei
```
manumonti:
```suggestion
                    VariableOperation(operation="ethToWei"),
```

manumonti (dismissed):
I had some suggestions 🤏

nucypher/policy/conditions/lingo.py
Muhammad-Altabba: If putting a timeout, I think it is better to set it on the root condition evaluation.

derekpierre: Great convo.

✔️ - Removed for now since no one has really asked for this functionality, and unclear of its usefulness at the moment given the expressed concern.

nucypher/policy/conditions/lingo.py
derekpierre: Fixed via https://github.com/derekpierre/nucypher/pull/9

nucypher/policy/conditions/lingo.py
derekpierre: Fixed via https://github.com/derekpierre/nucypher/pull/9

nucypher/policy/conditions/lingo.py
derekpierre: ✔️ - nice! Addressed in https://github.com/nucypher/nucypher/pull/3658/commits/91e373af2ca000748ed766b7bc07a341622753c0

nucypher/policy/conditions/lingo.py
derekpierre: ✔️  - Ended up using `evaluate_operations` - https://github.com/nucypher/nucypher/pull/3658/commits/5b927a7ffc404c43508dde8937bf7384dc799b81

tests/unit/conditions/test_signing_condition.py
derekpierre: ✔️

derekpierre:
> We have to be very careful with what operations we allow here. There's many potential traps (e.g. floating point precision issues, overloaded operators, etc) that we need to weigh very carefully. Please see my branch https://github.com/cygnusv/nucypher/tree/variable-operations, and in particular the last 2 commits with some examples.

Good catch, and thanks for the thoughtful changes.

It's good that we now use `Decimal` for calculations, but we have to be careful because it is really an internal structure. We can't totally get rid of precision issues, but we can limit them. The concern with returning `Decimal` values from operations as a result is that:
1. Decimal is really for internal calculations
2. Decimal cannot be used with `ast.literal_eval()` which is used for value comparison
3. Decimal cannot be specified as a comparator value in ReturnValueTest
4. Decimal is not JSON serializable so someone using the floating point result in a JSON would not be able do so if left as Decimal and not float

So I made some updates on top of the PR. Take a look and let me know what you think.  There is some loss of precision when converting from Decimal back to float at the very end of all of the operations, but I don't see that being an issue for the most part given the level of precision that can be achieved. There is also the option for the user to perform the `round` operation on results to given precision if necessary.

KPrasch (approved):
👍🏻 looking good after the removal of exponentiation!

manumonti (approved):
Looking good! 🔝

cygnusv (approved):

PR #3659, opened 2025-10-10
Signing powers

vzotova:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [X] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Separates operator key and signing key

**Issues fixed/closed:**
> - Fixes #...

Depends on https://github.com/nucypher/nucypher-contracts/pull/436

**Why it's needed:**

**Notes for reviewers:**
Open question (@KPrasch @derekpierre ): is ThresholdSIgningPower default power for Ursula, and do we want to have default constructor (no parameters, random generation) for it?

### DIFF
--- a/newsfragments/3659.dev.rst
+++ b/newsfragments/3659.dev.rst
@@ -1,0 +1,1 @@
+Dedicated signing key
--- a/newsfragments/3659.dev.rst
+++ b/newsfragments/3659.dev.rst
@@ -1,1 +1,1 @@
-Dedicated signing key
+Dedicated signing key for Action Control multisig messages.
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -279,14 +279,6 @@ class Operator(BaseActor):
             blockchain_endpoint=eth_endpoint,
         )
 
-        # track active onchain rituals
-        self.ritual_tracker = dkg.DkgRitualTracker(
-            operator=self,
-        )
-        self.signing_ritual_tracker = signing.SigningRitualTracker(
-            operator=self,
-        )
-
         self.publish_finalization = (
             publish_finalization  # publish the DKG final key if True
         )
@@ -304,6 +296,14 @@ class Operator(BaseActor):
 
         self.dkg_storage = DKGRitualStorage()
         self.signing_storage = SigningRitualStorage()
+
+        # track active onchain rituals
+        self.ritual_tracker = dkg.DkgRitualTracker(
+            operator=self,
+        )
+        self.signing_ritual_tracker = signing.SigningRitualTracker(
+            operator=self,
+        )
 
     def set_provider_public_key(self) -> Union[TxReceipt, None]:
         # TODO: Here we're assuming there is one global key per node. See nucypher/#3167
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1215,7 +1215,7 @@ class Operator(BaseActor):
                 )
 
             self.log.info(
-                f"{self.threshold_signing_power.account[:8]} resubmitted a new async tx {async_tx.id} "
+                f"{self.transacting_power.account[:8]} resubmitted a new async tx {async_tx.id} "
                 f"of type {tx_type} for signing ritual #{phase_id.ritual_id}."
             )
 
@@ -1279,7 +1279,7 @@ class Operator(BaseActor):
         )
         if async_tx:
             self.log.info(
-                f"Active signing ritual in progress: {self.threshold_signing_power.account} has submitted tx "
+                f"Active signing ritual in progress: {self.transacting_power.account} has submitted tx "
                 f"for cohort #{cohort_id}, post signature phase, (final: {async_tx.final})."
             )
             return async_tx
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -1241,7 +1241,7 @@ class Operator(BaseActor):
         participant = self.signing_coordinator_agent.get_signer(
             cohort_id=cohort_id, provider=self.staking_provider_address
         )
-        if participant.signature:
+        if participant.signerAddress != NULL_ADDRESS:
             # This is a normal state, as the node may have already submitted a signature
             # for this cohort, and it's not necessary to submit another one. Carry on.
             self.log.debug(
@@ -1288,7 +1288,7 @@ class Operator(BaseActor):
 
     def _post_signature(self, cohort_id: int) -> AsyncTx:
         data_hash = self.signing_coordinator_agent.get_signing_cohort_data_hash(
-            cohort_id
+            cohort_id=cohort_id, operator_address=self.transacting_power.account
         )
         _hash, signature = self.threshold_signing_power.sign_message_eip191(
             data_hash, standardize=False
@@ -1301,7 +1301,6 @@ class Operator(BaseActor):
             cohort_id=cohort_id,
             signature=signature,
             transacting_power=self.transacting_power,
-            threshold_signing_power=self.threshold_signing_power,
             async_tx_hooks=async_tx_hooks,
         )
         self.signing_storage.store_ritual_phase_async_tx(
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -75,6 +75,7 @@ from nucypher.crypto.powers import (
     CryptoPower,
     RitualisticPower,
     ThresholdRequestDecryptingPower,
+    ThresholdSigningPower,
     TransactingPower,
 )
 from nucypher.datastore.ritual import (
@@ -211,6 +212,7 @@ class Operator(BaseActor):
         polygon_endpoint: str,
         pre_payment_method: ContractPayment,
         transacting_power: TransactingPower,
+        threshold_signing_power: ThresholdSigningPower,
         signer: Signer = None,
         crypto_power: CryptoPower = None,
         client_password: str = None,
@@ -235,6 +237,8 @@ class Operator(BaseActor):
                 signer=signer,
                 cache=True,
             )
+
+        self.threshold_signing_power = threshold_signing_power
 
         # We pass the newly instantiated TransactingPower into consume_power_up here, even though it's accessible
         # on the instance itself (being composed in the __init__ of the base class, which we will call shortly)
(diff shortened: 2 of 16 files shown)


nucypher/crypto/powers.py
```diff
@@ -236,6 +237,33 @@ def sign_transaction(self, transaction_dict: dict) -> bytes:
         return self._signer.sign_transaction(transaction_dict=transaction_dict)
 
 
+class ThresholdSigningPower(TransactingPower):
```
derekpierre, line 244: @vzotova , @KPrasch can this just be a subclass of `SigningPower` instead of `TransactingPower` and added as a default crypto power up for Ursula.

KPrasch: Due to the legacy of powers in which "signing" was imagined to be a strictly P2P notion, SigningPower is built around a different cryptographic paradigm.  `TransactingPower` is already nearly purpose-built for the purpose of this PR (an ethereum key wrapper), although I can understand the naming confusion.  In terms of default powers, since TransactingPower (and TLSHostingPower) require runtime-context, we currently do not have a design that supports them being defaults.  I'd regard this as a design imperfection, that being said, TransactingPower has enjoyed special treatment that can be repeated for the `ThresholdSigningPower`.

TLDR; IMO it does not make sense to make `ThresholdSigningPower` a subclass of `SigningPower` and in order to avoid design upheveal `ThresholdSigningPower` can be implemented as a special case, not a default power.

derekpierre:
@vzotova, just pushed the commits we chatted about. The earlier commits in the PR are unsigned, and they will need to be signed for the PR to be merged.

Also, does it make sense to look into merging https://github.com/nucypher/nucypher-contracts/pull/436 before this PR since it depends on those contract changes? Not an issue, but if the contracts PR gets merged, then you don't need the updates to this `ape-config.yml` file in this PR to update the contracts repos/branch to use. Of course, if you prefer to merge this PR first, we can always make that update at a later time.

vzotova:
> @vzotova, just pushed the commits we chatted about. The earlier commits in the PR are unsigned, and they will need to be signed for the PR to be merged.
> 
> Also, does it make sense to look into merging [nucypher/nucypher-contracts#436](https://github.com/nucypher/nucypher-contracts/pull/436) before this PR since it depends on those contract changes? Not an issue, but if the contracts PR gets merged, then you don't need the updates to this `ape-config.yml` file in this PR to update the contracts repos/branch to use. Of course, if you prefer to merge this PR first, we can always make that update at a later time.

good point, I'd merge PR in nucypher-contracts first

nucypher/network/concurrency.py
```diff
@@ -205,8 +205,8 @@ def worker(
             stagger_timeout=stagger_timeout,
         )
 
-        # sort successes by operator address
-        # successes is of type Dict[ChecksumAddress, Tuple[ChecksumAddress, ThresholdSignatureResponse]]
+        # sort successes by signer address
+        # successes is of type Dict[ChecksumAddress, Tuple[ChecksumAddress, SignatureResponse]]
         successes = OrderedDict(sorted(successes.items(), key=lambda item: item[1][0]))
```
cygnusv: Just to point out that this sorts by signer address _as strings_. It doesn't seem that this has any effect, but in case it matters for something else.

nucypher/crypto/keystore.py
```diff
@@ -46,6 +48,7 @@
 _RITUALISTIC_INFO = __INFO_BASE + b"ritualistic"
 _THRESHOLD_REQUEST_DECRYPTING_INFO = __INFO_BASE + b"threshold_request_decrypting"
 _TLS_INFO = __INFO_BASE + b"tls"
+_THRESHOLD_SIGNING_INFO = __INFO_BASE + b"threshold_signing"
```
cygnusv: Maybe let's suffix this with ECDSA: `b"threshold_signing_ecdsa"`. See my other comments in `powers.py`

nucypher/crypto/powers.py
```diff
@@ -236,6 +241,21 @@ def sign_transaction(self, transaction_dict: dict) -> bytes:
...
+    """A power that is dedicated to sign threshold cryptography requests."""
+
+    KEY_SIZE = 32  # must be 32 bytes long
```
cygnusv: Noting that this is ECDSA-specific.

tests/acceptance/ape-config.yaml
```diff
@@ -5,8 +5,8 @@ plugins:
 
 dependencies:
   - name: nucypher-contracts
-    github: nucypher/nucypher-contracts
-    ref: signing
+    github: vzotova/nucypher-contracts
+    ref: operator-signer
```
cygnusv: We should change this, right?

tests/acceptance/agents/test_signing_coordinator_agent.py
```diff
@@ -201,6 +211,9 @@ def test_post_signature(
     deployed_multisig = nucypher_dependency.ThresholdSigningMultisig.at(
         threshold_signing_multisig_clone_factory.getCloneAddress(cohort_id)
     )
-    assert deployed_multisig.getSigners() == cohort_operators
+    assert (
+        deployed_multisig.getSigners()
+        == accounts.unassigned_accounts[: len(cohort_operators)]
```
cygnusv: Perhaps `accounts.unassigned_accounts[: len(cohort_operators)]` should be made into a fixture. Same for L49.

nucypher/blockchain/eth/models.py
```diff
@@ -224,12 +224,14 @@ def make_signers(
...
         signature: bytes = bytes()
 
         @classmethod
         def from_data(cls, data: list):
             return cls(
                 provider=ChecksumAddress(data[0]),
                 operator=ChecksumAddress(data[1]),
-                signature=bytes(data[2]),
+                signerAddress=ChecksumAddress(data[2]),
+                signature=bytes(data[3]),
```
cygnusv: Does it make sense to recover the address from the signature? Since this is ECDSA anyway

tests/acceptance/ape-config.yaml
derekpierre: This will be changed back once https://github.com/nucypher/nucypher-contracts/pull/436 is merged.

nucypher/blockchain/eth/models.py
vzotova: it mimics what happens in contracts, so after finishing discussion about what we store and what relationship between signer<->operator - we can return to this one. If everything stays as now - I'd remove signature from the struct in contract

tests/acceptance/agents/test_signing_coordinator_agent.py
vzotova: ✔️

nucypher/crypto/powers.py
vzotova: ✔️

nucypher/crypto/keystore.py
vzotova: ✔️

newsfragments/3659.dev.rst
```diff
@@ -0,0 +1 @@
+Dedicated signing key
```
cygnusv:
```suggestion
Dedicated signing key for Action Control multisig messages.
```

nucypher/blockchain/eth/actors.py
```diff
@@ -1275,7 +1279,7 @@ def perform_post_signature(
         )
         if async_tx:
             self.log.info(
-                f"Active signing ritual in progress: {self.transacting_power.account} has submitted tx "
+                f"Active signing ritual in progress: {self.threshold_signing_power.account} has submitted tx "
```
cygnusv: The logging intention here is ambiguous: Do we want to log that an operator account submitted a TX or that a signer produced a signature? As far as I understand, the signer doesn't submit TXs, but operators do.

cygnusv (approved):

nucypher/network/concurrency.py
derekpierre: Good catch - @vzotova, that was my prior mistake separate from this PR. Can you update the code to mimic the ordering of addresses here - https://github.com/nucypher/nucypher/blob/v7.7.x/tests/acceptance/actors/test_dkg_ritual.py#L173

nucypher/blockchain/eth/actors.py
```diff
@@ -1211,7 +1215,7 @@ def resubmit_tx():
                 )
 
             self.log.info(
-                f"{self.transacting_power.account[:8]} resubmitted a new async tx {async_tx.id} "
+                f"{self.threshold_signing_power.account[:8]} resubmitted a new async tx {async_tx.id} "
```
derekpierre: Careful - this is still technically the operator (`transacting_power`) submitting the tx. Should we change this back to `transacting_power`?

nucypher/network/concurrency.py
```diff
@@ -185,8 +185,8 @@ def worker(
                 )
                 if response.status_code == HTTPStatus.OK:
                     response = SignatureResponse.from_bytes(response.content)
-                    operator_address = node_or_sprout.operator_address
-                    return operator_address, response
+                    signer_address = response.signer
+                    return signer_address, response
```
derekpierre: Signer address is already in the response (`SignatureResponse` object), so let's just return the response instead of the tuple.

vzotova: ✔️

nucypher/blockchain/eth/actors.py

nucypher/blockchain/eth/models.py
vzotova: ✔️

nucypher/network/concurrency.py
vzotova: ✔️

nucypher/blockchain/eth/actors.py
derekpierre: ^ one more spot here @vzotova .

tests/acceptance/ape-config.yaml
derekpierre: ^ Reminder about this once the `nucypher-contracts` PR is merged. It has sufficient approvals already, so feel free to merge.

nucypher/blockchain/eth/actors.py

newsfragments/3659.dev.rst
derekpierre: ^ Did this commit get lost?

derekpierre (approved):
🎸 - nice work!

Looks like the newsfragment commit may have gotten lost in the shuffle - all good otherwise,

newsfragments/3659.dev.rst
vzotova: thanks!
returned changes back

PR #3661, opened 2025-10-15
Reduce log level for condition evaluation

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [x] 1
- [ ] 2
- [ ] 3

**What this does:**

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/policy/conditions/json/utils.py
+++ b/nucypher/policy/conditions/json/utils.py
@@ -1,7 +1,5 @@
 from typing import Any, Optional
 
-from nucypher.utilities.logging import Logger
-
 
 def process_result_for_condition_eval(result: Any):
     # strings that are not already quoted will cause a problem for literal_eval
@@ -34,17 +32,14 @@ def query_json_data(data: Any, query: Optional[str], **context) -> Any:
     from nucypher.policy.conditions.exceptions import ConditionEvaluationFailed
 
     resolved_query = resolve_any_context_variables(query, **context)
-    logger = Logger(__name__)
 
     try:
         expression = parse(resolved_query)
         matches = expression.find(data)
         if not matches:
             message = f"No matches found for the JSONPath query: {resolved_query}"
-            logger.debug(message)
             raise ConditionEvaluationFailed(message)
     except (JsonPathLexerError, JsonPathParserError) as jsonpath_err:
-        logger.error(f"JSONPath error occurred: {jsonpath_err}")
         raise ConditionEvaluationFailed(
             f"JSONPath error: {jsonpath_err}"
         ) from jsonpath_err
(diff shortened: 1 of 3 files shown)


nucypher/policy/conditions/json/utils.py
```diff
@@ -41,7 +41,7 @@ def query_json_data(data: Any, query: Optional[str], **context) -> Any:
         matches = expression.find(data)
         if not matches:
             message = f"No matches found for the JSONPath query: {resolved_query}"
-            logger.info(message)
+            logger.debug(message)
```
derekpierre: Should this be a `warn` instead of `debug`?

This error case shouldn't happen often because the condition will fail and be useless otherwise.

cygnusv: This is not an error, but a falsy condition evaluation, and that's something that shouldn't be logged by default. Otherwise, there's privacy concerns for requesters since nodes would be logging their access patterns.

derekpierre: `falsy` in that the condition will return `False`?

I should have clarified, by "fail" I mean that the condition doesn't return `False` but rather the exception raised on the next line (L45) is propagated and a HTTP 400 is returned for the decryption/signing request.

That being said, perhaps we don't even need to log here, since it gets logged here - https://github.com/cygnusv/nucypher/blob/arepa/nucypher/policy/conditions/utils.py#L258.
^ Perhaps we should change that log statement to a `warn`? Unless you still think that it shouldn't be logged at all...?

cygnusv: Ok, maybe I misunderstood how this part works. I thought no matches here implied that the condition is falsy, but it seems this is an error. In that case, a logged message is OK. But if the message is logged anyway in L258 in the other utils module, then the change I introduced (from info to debug) doesn't change much?

derekpierre: >  I thought no matches here implied that the condition is falsy, but it seems this is an error

Yeah, it's subtle. Because there is no match for a result from the JSON, there's nothing to use for the return value test check, so it's really an error.

> But if the message is logged anyway in L258 in the other utils module, then the change I introduced (from info to debug) doesn't change much?

It does reduce the one instance of the logging since it now goes to debug.

Since it seems we have duplicate logging, wondering what you think of possibly going even further and removing this duplicated logging (L44) since it is already logged in L258 and whether the spot on L258 should log to `warn` instead of `info`?

derekpierre (approved):
🎸

PR #3662, opened 2025-10-17
Follow-up on Variable Operation where strings returned from JSON queries were not handled properly by return value test

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
- Increases max number of conditions within a SequentialCondition from 5 to 10 to allow for more complex logic.
- Handle strings within return value test instead of in JSON conditions, so that operations can be performed on the raw strings before needing to process them for use with `ast.literal_eval`
- Add test that utilizes parsing of values from Discord message payload to reproduce original bug/issue observed.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/newsfragments/3662.feature.rst
+++ b/newsfragments/3662.feature.rst
@@ -1,0 +1,1 @@
+Increased the maximum number of conditions within a ``SequentialCondition`` from 5 to 10 to support more complex logic.
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -480,6 +480,8 @@ class ConditionVariable(_Serializable):
 
 
 class SequentialCondition(MultiCondition):
+    MAX_NUM_CONDITIONS = 10
+
     """
     A series of conditions that are evaluated in a specific order, where the result of one
     condition can be used in subsequent conditions.
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -916,7 +916,7 @@ class ReturnValueTest(_Serializable):
         # leave as is
         return data
 
-    def _process_data(self, data: Any, call_level: int = 0) -> Any:
+    def _process_data(self, data: Any, top_level_call: bool = True) -> Any:
         """
         Process data for use with literal eval. Recursively processes data as needed.
         Steps:
@@ -929,24 +929,21 @@ class ReturnValueTest(_Serializable):
         if isinstance(data, (list, tuple)):
             # convert any bytes in list to hex (include nested lists/tuples); no additional indexing
             processed_data = [
-                self._process_data(data=item, call_level=call_level + 1)
-                for item in data
+                self._process_data(data=item, top_level_call=False) for item in data
             ]
             return processed_data
-
-        # convert bytes to hex if necessary
-        if isinstance(data, bytes):
+        elif isinstance(data, bytes):
+            # convert bytes to hex if necessary
             processed_data = self.__process_bytes(data)
             return processed_data
-
-        if isinstance(data, str):
+        elif isinstance(data, str):
             # only process strings at the top level call (i.e. don't need to process strings within lists/tuples)
-            if call_level > 0:
+            if not top_level_call:
                 return data
 
             return self.__process_string(data)
-
-        return data
+        else:
+            return data
 
     def eval(self, data) -> bool:
         if is_context_variable(self.value):
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -4,12 +4,13 @@ import json
 import math
 import operator as pyoperator
 import statistics
-from decimal import localcontext
+from decimal import Decimal, InvalidOperation, localcontext
 from enum import Enum
 from hashlib import md5
 from inspect import signature
 from typing import Any, List, Optional, Tuple, Type, Union
 
+from eth_utils import is_hexstr
 from hexbytes import HexBytes
 from marshmallow import (
     Schema,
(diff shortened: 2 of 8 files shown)


tests/unit/conditions/test_sequential_condition.py
```diff
@@ -397,6 +404,108 @@ def test_sequential_condition_variable_with_failed_operation(mock_condition_vari
...
+            value="0xA87722643685B38D37ecc7637ACA9C1E09c8C5e1",
+        ),
+    )
+    sequential_condition = SequentialCondition(
+        condition_variables=[
+            ConditionVariable(
+                var_name="amount1",
+                condition=amount_json_condition,
+                operations=[
+                    # ethToWei can covert value from string
```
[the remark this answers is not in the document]
derekpierre: ✔️

nucypher/policy/conditions/lingo.py
```diff
@@ -874,22 +877,76 @@ def _sanitize_value(cls, value) -> Any:
...
+            return True
+
+        # check if int or float as string
+        try:
+            _ = Decimal(data)
+            return True
+        except InvalidOperation:
+            pass
+
+        return False
```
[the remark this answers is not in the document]
derekpierre: Only applies to python native types, lower case "true"/"false" does not work with `ast.literal_eval()`

```python
In [1]: import ast

In [2]: ast.literal_eval("true")
ValueError                                Traceback (most recent call last)
Cell In[2], line 1
----> 1 ast.literal_eval("true")
...
ValueError: malformed node or string on line 1: <ast.Name object at 0x104f25810>

In [3]: ast.literal_eval("false")
ValueError                                Traceback (most recent call last)
Cell In[3], line 1
----> 1 ast.literal_eval("false")
...
ValueError: malformed node or string on line 1: <ast.Name object at 0x104ed7550>
```

as opposed to:

```python

In [4]: ast.literal_eval("False")
Out[4]: False

In [5]: ast.literal_eval("True")
Out[5]: True
```

vzotova (approved):

theref (approved):

newsfragments/3662.feature.rst
```diff
@@ -0,0 +1 @@
+Increased the maximum number of conditions within a ``SequentialCondition` from 5 to 10 to support more complex logic.
```
cygnusv:
```suggestion
Increased the maximum number of conditions within a ``SequentialCondition`` from 5 to 10 to support more complex logic.
```

nucypher/policy/conditions/lingo.py
```diff
@@ -874,22 +877,76 @@ def _sanitize_value(cls, value) -> Any:
...
 
         # convert bytes to hex if necessary
-        return self.__handle_potential_bytes(processed_data)
+        if isinstance(data, bytes):
+            processed_data = self.__process_bytes(data)
+            return processed_data
+
+        if isinstance(data, str):
+            # only process strings at the top level call (i.e. don't need to process strings within lists/tuples)
+            if call_level > 0:
```
cygnusv: If we're interested only in discerning top-level calls, then `call_level` can be substituted by a bool, right? Another stylistic comment: instead of a chain `if isinstance()`, how about `if/elif`? Feels less dependent on the use of `return`. Not strong comments anyway, you can take it or leave it.

nucypher/policy/conditions/lingo.py
```diff
@@ -874,22 +877,76 @@ def _sanitize_value(cls, value) -> Any:
...
+        # check if string is already primitive type, leave as is
+        if ReturnValueTest.__string_already_primitive_type(data):
+            return data
 
-    def _process_data(self, data: Any) -> Any:
+        # check if already quoted; if not, quote it
+        if not (
+            (data.startswith("'") and data.endswith("'"))
+            or (data.startswith('"') and data.endswith('"'))
+        ):
```
cygnusv: How about 1-letter strings like `'` or `"`? Maybe we should enforce length here, or perhaps a regex?

nucypher/policy/conditions/lingo.py
```diff
@@ -874,22 +877,76 @@ def _sanitize_value(cls, value) -> Any:
...
+            quote_type_to_use = '"' if "'" in data else "'"
```
cygnusv, line 911: What should happen if `data` is `foo"bar`? Result would be `"foo"bar"`.

newsfragments/3662.feature.rst
derekpierre: ✔️

nucypher/policy/conditions/lingo.py
derekpierre: ✔️  Fixed now. Both 0-length and 1-length strings are handled.

nucypher/policy/conditions/lingo.py
derekpierre: That becomes `'foo"bar'`. See added tests in `test_return_value`.

nucypher/policy/conditions/lingo.py
derekpierre: ✔️

nucypher/policy/conditions/lingo.py
```diff
@@ -479,6 +480,8 @@ def __init__(
 
 
 class SequentialCondition(MultiCondition):
+    MAX_NUM_CONDITIONS = 10
```
manumonti, line 483: Just to confirm: se we are increasing the maximum number of conditions from 5 to 10 but only for Sequential Conditions, right? I'm asking because on MultiCondition this number is still 5.

https://github.com/derekpierre/nucypher/blob/8122bbf157ee3dd6495ea9ec9a3c411c8e0352ee/nucypher/policy/conditions/base.py#L102

manumonti (approved):
LGTM! I left a question tho

nucypher/policy/conditions/lingo.py
derekpierre: Correct, only for SequentialConditions. Compound conditions are still limited to 5. The reason being complex logic like the tipping workflow probably requires at least 1 condition per property eg. `amount`, `recipient address` PLUS additional condition(s) for checking those variable value against other data like values in UserOp.

cygnusv (approved):

PR #3664, opened 2025-10-28
Add JSON, hex, and keccak operators for variable operations

theref:
## Summary

Adds new operators to the condition lingo system for enhanced variable operations:
- **JSON operators**: `fromJson`, `toJson` for parsing/serializing JSON data
- **Hex operators**: `fromHex`, `toHex` for hex encoding/decoding
- **Hash operator**: `keccak` for keccak256 hashing

## Changes

- Added 5 new operators to `VARIABLE_OPERATIONS` in `nucypher/policy/conditions/lingo.py`
- Comprehensive unit tests for JSON, hex, and keccak operators
- JSON-API integration tests for variable operations
- ECDSA condition integration tests demonstrating real-world usage with context parameter concatenation

## Testing

- ✅ Unit tests for all new operators with various input types
- ✅ JSON-API tests validating variable operations in API conditions
- ✅ Integration tests showing practical usage with ECDSA conditions and Discord payload simulation

### DIFF
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -342,6 +342,28 @@ _COMPARATOR_FUNCTIONS = {
 }
 
 
+def _to_hex(value):
+    """
+    Convert value to hex string.
+
+    Supports bytes, bytearray, int, and str types.
+    Strings are encoded to UTF-8 before conversion.
+
+    :param value: Value to convert to hex
+    :return: Hex string with 0x prefix
+    :raises TypeError: If value type cannot be converted to hex
+    """
+    try:
+        if isinstance(value, str):
+            # Encode regular strings to UTF-8
+            value = value.encode("utf-8")
+        # HexBytes handles bytes, bytearray, and int
+        h = HexBytes(value)
+        return h.hex()
+    except (TypeError, ValueError) as e:
+        raise TypeError(f"Invalid value for hex conversion: {e}")
+
+
 # should raise TypeError for invalid inputs
 _OPERATOR_FUNCTIONS = {
     # We can add all kinds of operators over time, this is just a base start - given that
@@ -376,11 +398,7 @@ _OPERATOR_FUNCTIONS = {
     "toJson": lambda a: json.dumps(a),
     # hex conversion
     "fromHex": lambda a: bytes(HexBytes(a)),
-    "toHex": lambda a: (
-        HexBytes(a).hex()
-        if isinstance(a, (bytes, bytearray))
-        else HexBytes(a.encode() if isinstance(a, str) else str(a).encode()).hex()
-    ),
+    "toHex": _to_hex,
     # hashing
     "keccak": lambda a: HexBytes(keccak(a.encode() if isinstance(a, str) else a)).hex(),
 }
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -394,13 +394,17 @@ _OPERATOR_FUNCTIONS = {
     "int": lambda a: int(a),
     "str": lambda a: str(a),
     # JSON conversion
-    "fromJson": lambda a: json.loads(a if isinstance(a, str) else a.decode()),
+    "fromJson": lambda a: json.loads(a),
     "toJson": lambda a: json.dumps(a),
     # hex conversion
     "fromHex": lambda a: bytes(HexBytes(a)),
     "toHex": _to_hex,
     # hashing
-    "keccak": lambda a: HexBytes(keccak(a.encode() if isinstance(a, str) else a)).hex(),
+    "keccak": lambda a: keccak(
+        a.encode()
+        if isinstance(a, str)
+        else (a if isinstance(a, (bytes, bytearray)) else str(a).encode())
+    ),
 }
 
 MAX_VARIABLE_OPERATIONS = 5
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -400,11 +400,7 @@ _OPERATOR_FUNCTIONS = {
     "fromHex": lambda a: bytes(HexBytes(a)),
     "toHex": _to_hex,
     # hashing
-    "keccak": lambda a: keccak(
-        a.encode()
-        if isinstance(a, str)
-        else (a if isinstance(a, (bytes, bytearray)) else str(a).encode())
-    ),
+    "keccak": lambda a: keccak(a.encode() if isinstance(a, str) else a),
 }
 
 MAX_VARIABLE_OPERATIONS = 5
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -10,7 +10,7 @@ from hashlib import md5
 from inspect import signature
 from typing import Any, List, Optional, Tuple, Type, Union
 
-from eth_utils import is_hexstr
+from eth_utils import is_hexstr, keccak
 from hexbytes import HexBytes
 from marshmallow import (
     Schema,
@@ -371,6 +371,18 @@ _OPERATOR_FUNCTIONS = {
     "float": lambda a: float(a),
     "int": lambda a: int(a),
     "str": lambda a: str(a),
+    # JSON conversion
+    "fromJson": lambda a: json.loads(a if isinstance(a, str) else a.decode()),
+    "toJson": lambda a: json.dumps(a),
+    # hex conversion
+    "fromHex": lambda a: bytes(HexBytes(a)),
+    "toHex": lambda a: (
+        HexBytes(a).hex()
+        if isinstance(a, (bytes, bytearray))
+        else HexBytes(a.encode() if isinstance(a, str) else str(a).encode()).hex()
+    ),
+    # hashing
+    "keccak": lambda a: HexBytes(keccak(a.encode() if isinstance(a, str) else a)).hex(),
 }
 
 MAX_VARIABLE_OPERATIONS = 5
--- a/tests/integration/conditions/test_ecdsa_condition_integration.py
+++ b/tests/integration/conditions/test_ecdsa_condition_integration.py
@@ -25,6 +25,14 @@ TEST_MESSAGE = (
     "There is a road, no simple highway, between the dawn and the dark of night. -JG"
 )
 
+# Discord Ed25519 test vectors (shared across multiple tests)
+DISCORD_PUBLIC_KEY_HEX = (
+    "b853dd9f496723daf64bf2f5a886086f790df66e61d7b6f7f98a50c9e5ede8f3"
+)
+DISCORD_SIGNATURE_HEX = "0a12acb96843700b724f1c9dba3075a7fc482677e0c713eb6be63bfea406fb33b4715975f1062b2dff95127bd559ee23758911bd217760727dc44e7880bc6e04"
+DISCORD_TIMESTAMP = "1749368683"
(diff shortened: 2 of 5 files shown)


nucypher/policy/conditions/lingo.py
```diff
@@ -330,6 +330,18 @@ def __init__(self, operand: Condition):
...
     "str": lambda a: str(a),
+    # JSON conversion
+    "fromJson": lambda a: json.loads(a if isinstance(a, str) else a.decode()),
+    "toJson": lambda a: json.dumps(a),
+    # hex conversion
+    "fromHex": lambda a: bytes(HexBytes(a)),
+    "toHex": lambda a: (
+        HexBytes(a).hex()
+        if isinstance(a, (bytes, bytearray))
+        else HexBytes(a.encode() if isinstance(a, str) else str(a).encode()).hex()
```
```suggestion
        else HexBytes(str(a).encode()).hex()
```

tests/integration/conditions/test_ecdsa_condition_integration.py
```diff
@@ -129,30 +136,257 @@ def test_ecdsa_in_compound_condition():
...
..._user":0,"theme_color":null,"topic":null,"type":0},"channel_id":"1380488052169769113","context":0,"data":{"id":"1380515955146358918","name":"sign","options":[{"name":"message","type":3,"value":"llamas"}],"type":1},"entitlement_sku_ids":[],"entitlements":[],"guild":{"features":[],"id":"1380488052169769110","locale":"en-US"},"guild_id":"1380488052169769110","guild_locale":"en-US","id":"1381177107555680327","locale":"en-US","member":{"avatar":null,"banner":null,"communication_disabled_until":null,"deaf":false,"flags":0,"joined_at":"2025-06-06T10:06:39.888000+00:00","mute":false,"nick":null,"pending":false,"permissions":"2251799813685247","premium_since":null,"roles":[],"unusual_dm_activity_until":null,"user":{"avatar":"9c5483a989a10edc8b831b6c8f284724","avatar_decoration_data":null,"clan":null,"collectibles":null,"discriminator":"0","global_name":"kprasch","id":"410212090289192960","primary_guild":null,"public_flags":0,"username":"kprasch"}},"token":"aW50ZXJhY3Rpb246MTM4MTE3NzEwNzU1NTY4MDMyNzo2ZlVlTkdIQVFIdkdhNUN2bXlYZ1RCRkZ5YkpuVm9WUG5Bbjl0TTMyUkFTenZGYXNXclBCMjZWUWJlalczcllRak9sc0JZR3Q4WW9sODBRcDg2c2hrYmRkYzlWcjF3TjdlRFVDMVBVTkZ3Z2VmRUp4VkI1MkZGNmVvM3hkWXd1Qg","type":2,"version":1}'
```
cygnusv: If these values are the same than in the previous test, maybe they can be defined as test constants.

tests/unit/conditions/test_variable_operation.py
```diff
@@ -244,6 +266,128 @@ def test_overloaded_operators():
     assert result == "TACoTACoTACo!"
 
 
+def test_string_concatenation():
```
cygnusv, line 285: There's already a similar test above, `test_overloaded_operators()`.

tests/unit/conditions/test_variable_operation.py
```diff
@@ -244,6 +266,128 @@ def test_overloaded_operators():
...
+    # Test JSON to hex and back
+    initial = {"address": "0x123", "amount": 100}
+    operations = [
+        VariableOperation(operation="toJson"),  # '{"address": "0x123", "amount": 100}'
+        VariableOperation(operation="toHex"),  # hex representation
+        VariableOperation(operation="fromHex"),  # back to JSON string bytes
+        VariableOperation(operation="fromJson"),  # back to original dict
+    ]
+    result = VariableOperation.evaluate_operations(operations, initial)
+    assert result == initial
```
cygnusv, line 327: 👏

cygnusv (approved):

nucypher/policy/conditions/lingo.py
derekpierre: We have to be really careful here. I don't think we can just cast to string. Not everything can/should be turned into hex, for example floats, so someone passing a float to this function should raise a `TypeError`.

```python
In [7]: h = HexBytes(17)

In [8]: h.hex()
Out[8]: '0x11'.   # correct

In [9]: h = HexBytes(str(17).encode())

In [10]: h.hex()
Out[10]: '0x3137'.   # not correct
```

`HexBytes` already takes care of bytes, bytearray, and integer cases for us, *EXCEPT* the case when a string is passed that is non-hex (it expects any string to already be hex):

```python
In [24]: HexBytes("dead").hex()
Out[24]: '0xdead'

In [25]: HexBytes("deadbeat").hex()
Error                                     Traceback (most recent call last)
Cell In[25], line 1
----> 1 HexBytes("deadbeat").hex()

...
File ~/.local/share/virtualenvs/nucypher-porter-wF43iFf2/lib/python3.12/site-packages/hexbytes/_utils.py:54, in hexstr_to_bytes(hexstr)
     50     raise ValueError(
     51         f"hex string {padded_hex} may only contain [0-9a-fA-F] characters"
     52     )
     53 else:
---> 54     return binascii.unhexlify(ascii_hex)

Error: Non-hexadecimal digit found
```

Do we care to differentiate between an already provided hex string, and a regular string...my view is no differentiation i.e. any string provided should be converted to bytes

Note that these operator functions raise `TypeError` if invalid types are provided.

What about a function that does something like:
```python
def to_hex(value):
    try:
         if isinstance(value, str):
             value = value.encode("utf-8")
         h = HexBytes(value)
         return h.hex()
    except Exception as e:
         raise TypeError(f"Invalid value for hex conversion: {e}")
```

^ (Perhaps `Exception` can be `(TypeError, ValueError)` instead; not sure what other exceptions can arise from `HexBytes`)

nucypher/policy/conditions/lingo.py
```diff
@@ -330,6 +330,18 @@ def __init__(self, operand: Condition):
     "float": lambda a: float(a),
     "int": lambda a: int(a),
     "str": lambda a: str(a),
+    # JSON conversion
+    "fromJson": lambda a: json.loads(a if isinstance(a, str) else a.decode()),
```
derekpierre: `json.loads` already takes care of typing for us:
```
def loads(s, *, cls=None, object_hook=None, parse_float=None,
        parse_int=None, parse_constant=None, object_pairs_hook=None, **kw):
    """Deserialize ``s`` (a ``str``, ``bytes`` or ``bytearray`` instance
    containing a JSON document) to a Python object.
```

```suggestion
    "fromJson": lambda a: json.loads(a),
```

tests/integration/conditions/test_ecdsa_condition_integration.py
```diff
@@ -129,30 +136,257 @@ def test_ecdsa_in_compound_condition():
     assert and_lingo.eval(**context) is True
 
 
-def test_discord_ed25519_with_ecdsa_condition():
+def test_discord_json_from_body_string():
```
derekpierre, line 147: There is an existing test that leverages ECDSA and JsonCondtion already - see https://github.com/nucypher/nucypher/blob/v7.7.x/tests/unit/conditions/test_sequential_condition.py#L407. Feel free to adjust that test or write a new one for your needs.


Side note: I know this isn't all your code, but any reason most (if not all) of these tests in "integration" can't be moved to the ecdsa "unit" test. I guess the compound condition test is an integration test, but you can move that to the "unit" tests for compound condition i.e. `test_compound_condition.py`. (cc @KPrasch )

nucypher/policy/conditions/lingo.py
```diff
@@ -330,6 +330,18 @@ def __init__(self, operand: Condition):
...
+    "toJson": lambda a: json.dumps(a),
+    # hex conversion
+    "fromHex": lambda a: bytes(HexBytes(a)),
+    "toHex": lambda a: (
+        HexBytes(a).hex()
+        if isinstance(a, (bytes, bytearray))
+        else HexBytes(a.encode() if isinstance(a, str) else str(a).encode()).hex()
+    ),
+    # hashing
+    "keccak": lambda a: HexBytes(keccak(a.encode() if isinstance(a, str) else a)).hex(),
```
derekpierre: Why automatically convert the hash to hex? The user has the option to pair it with the use of "toHex" operation if they want hex.

```suggestion
    "keccak": lambda a: keccak(a.encode() if isinstance(a, str) else a),
```

tests/unit/conditions/test_variable_operation.py
```diff
@@ -62,6 +62,28 @@
...
+        None,
+        "",
+        "0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470",
+    ),
+    (
+        "keccak",
+        None,
+        "test",
+        "0x9c22ff5f21f0b81b113e63f7db6da94fedef11b2119b4088b89664fb9a3cb658",
+    ),
```
derekpierre, line 88: Let's also add a case for a number and for bytes. Example (but not limited to):
- 24
- b"testing"

tests/unit/conditions/test_variable_operation.py
```diff
@@ -107,9 +129,9 @@ def test_type_errors_in_evaluation(operation):
         op = VariableOperation(operation=operation)
     else:
         op = VariableOperation(operation=operation, value=value)
-    # Skip type error test for bool and str casting operations because
-    # they can handle any input without raising TypeError
-    if operation in ["bool", "str"]:
+    # Skip type error test for operations that can handle any input without raising TypeError
+    # or that raise different exceptions (e.g., JSONDecodeError for fromJson)
+    if operation in ["bool", "str", "toJson", "fromJson", "toHex", "fromHex", "keccak"]:
```
derekpierre: Made some recommendations on changing the logic for these new operations logic. Consequently,
- `toJson` (i.e. `json.loads()`) fails if bytes are passed to it
- `fromJson` (i.e. json.dumps()`) fails if bytes are passed to it
- `toHex` fails if an array like that in this test or a float is provided to it
- `fromHex` fails if an array like that in this test or a float is provided to it 
- `keccak` fails if an array like that in this test a float is provided to it.

So you shouldn't need to skip these - try playing around with the above.

nucypher/policy/conditions/lingo.py
```diff
@@ -330,6 +330,18 @@ def __init__(self, operand: Condition):
...
 }
 
 MAX_VARIABLE_OPERATIONS = 5
```
derekpierre, line 406: Given the complexity of the processing you are now using, do you think a max of 5 is still sufficient? I'm purely curious since 5 was just random, but it can be modified.

derekpierre (dismissed):
Nice work. Recommending some adjustments to the logic for the new operations.

Side note: Do these operations fully solve our problem outlined in https://github.com/nucypher/sprints/issues/271? We've tested this on the test Discord server? Just want to make sure we are being comprehensive here before merging.

manumonti (dismissed):
I would have approved the PR if I hadn't read Derek's concerns on  `toHex`. So I'm going to wait until this is solved.

Looks great to me!

theref:
> I would have approved the PR if I hadn't read Derek's concerns on `toHex`. So I'm going to wait until this is solved.
> 
> Looks great to me!

I've already addressed that issue with a new `_to_hex` function

nucypher/policy/conditions/lingo.py
theref: i can't remember how it's applied. But i think we'll probably do more than 5 total variable operations in the production CL condition

tests/integration/conditions/test_ecdsa_condition_integration.py
```diff
@@ -129,30 +144,257 @@ def test_ecdsa_in_compound_condition():
...
-    # Construct the message to be signed
-    message = timestamp.encode("utf-8") + body.encode("utf-8")
+    # This is the raw JSON string as it would come from Discord
+    body = DISCORD_BODY
 
-    # Create ECDSACondition using the Discord public key and signature
+    # Create ECDSA condition for signature verification with message template
     ecdsa_condition = ECDSACondition(
-        message=":message",
+        message=":timestamp:discordPayload",  # Constructs message from timestamp + payload
```
KPrasch: I accidentally left this in... I’ll refactor to remove it entirely (unless we decide this type of expression is useful?).

nucypher/policy/conditions/lingo.py
```diff
@@ -330,6 +352,18 @@ def __init__(self, operand: Condition):
...
+    "toJson": lambda a: json.dumps(a),
+    # hex conversion
+    "fromHex": lambda a: bytes(HexBytes(a)),
+    "toHex": _to_hex,
+    # hashing
+    "keccak": lambda a: keccak(
+        a.encode()
+        if isinstance(a, str)
+        else (a if isinstance(a, (bytes, bytearray)) else str(a).encode())
+    ),
```
derekpierre: Let's use `keccak()` as is without casting to string since it already handles the relevant types for you. The only reason we need to convert a string to bytes is because if you are passing string to `keccak` you are supposed to use the `text=` kwarg; instead we convert it to bytes which `keccak` would have done internally anyway.

```suggestion
    "keccak": lambda a: keccak(a.encode() if isinstance(a, str) else a)
```

tests/unit/conditions/test_variable_operation.py
```diff
@@ -107,9 +143,17 @@ def test_type_errors_in_evaluation(operation):
...
     else:
         op = VariableOperation(operation=operation, value=value)
-    # Skip type error test for bool and str casting operations because
-    # they can handle any input without raising TypeError
-    if operation in ["bool", "str"]:
+    # Skip type error test for operations that can handle any input without raising TypeError.
+    # These operations are designed to accept any input type and will not raise TypeError.
+    # - bool, str, toJson: explicitly designed to handle any type
+    # - keccak: converts any type to string representation before hashing
+    if operation in ["bool", "str", "toJson", "keccak"]:
```
derekpierre: `toJson` and  `keccak` doesn't actually handle any value...  - see comment here - https://github.com/nucypher/nucypher/pull/3664#discussion_r2472940190.

derekpierre:
Side note: @theref, if you're wondering "what next", we'll need to add these new operators to `taco-web` as well.

tests/unit/conditions/test_variable_operation.py
derekpierre: Fixed via #3667

manumonti (approved):
LGTM! 💫

Muhammad-Altabba (commented):
I am not very much into python and nucypher, but the PR looks good to me.

tests/unit/conditions/test_variable_operation.py
```diff
@@ -107,11 +143,17 @@ def test_type_errors_in_evaluation(operation):
...
-    # they can handle any input without raising TypeError
+    # Skip type error test for operations that can handle any input without raising TypeError.
+    # - bool, str: explicitly designed to handle any type
     if operation in ["bool", "str"]:
         return
 
+    # Skip type error test for operations that may raise exceptions other than TypeError,
+    # such as JSONDecodeError for 'fromJson' or ValueError for 'fromHex'.
+    # Or operations that have their own specific TypeError tests.
+    if operation in ["fromJson", "fromHex", "toHex", "toJson", "keccak"]:
```
derekpierre: @theref - I no longer see the changes from https://github.com/nucypher/nucypher/pull/3667 - was this lost as part of the rebase?

derekpierre (dismissed):

tests/unit/conditions/test_variable_operation.py
derekpierre: ^ Happy to approve once this is addressed/investigated.

derekpierre: FIxed rebase

KPrasch (approved):
Tested on Devnet ✔️

PR #3668, opened 2025-11-05
feat(conditions): add :nullAddress protected context variable

theref:
## Summary

Adds a new protected context variable `:nullAddress` that resolves to the Ethereum null address (0x0000000000000000000000000000000000000000) without requiring any context data.

## Motivation

This allows conditions to reference the null address directly, which is useful for:
- Checking token burns (balanceOf null address)
- Ownership validation against the zero address
- Any condition that needs to reference the standard null/zero address

## Changes

- Add `NULL_ADDRESS_CONTEXT = ":nullAddress"` constant
- Implement `_resolve_null_address()` function that returns the checksummed null address
- Register `:nullAddress` in `_DIRECTIVES` dictionary as a protected variable
- Add comprehensive unit test including usage in `ContractCondition`

## Usage Example

```python
condition = ContractCondition(
    contract_address="0x...",
    method="balanceOf",
    standard_contract_type="ERC20",
    chain=TESTERCHAIN_CHAIN_ID,
    return_value_test=ReturnValueTest(">=", 0),
    parameters=[":nullAddress"],  # Resolves to 0x0000...0000
)
```

## Testing

- All existing context tests pass (28 tests)
- New test verifies `:nullAddress` works correctly in all contexts:
  - Direct resolution
  - In lists with other context variables
  - In ReturnValueTest
  - In ContractCondition parameters

### DIFF
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -5,6 +5,7 @@ from typing import Any, Dict, List, Optional, Union
 from eth_typing import ChecksumAddress
 from eth_utils import to_checksum_address
 
+from nucypher.blockchain.eth.constants import NULL_ADDRESS
 from nucypher.policy.conditions.auth.evm import EvmAuth
 from nucypher.policy.conditions.exceptions import (
     ContextVariableVerificationFailed,
@@ -103,9 +104,10 @@ def _resolve_null_address(
     Returns the null address (0x0000000000000000000000000000000000000000).
 
     This is a protected context variable that doesn't require any input data.
-    """
-    from nucypher.blockchain.eth.constants import NULL_ADDRESS
 
+    Note: The `providers` and `**context` parameters are unused but maintained
+    for interface consistency with other resolver functions in `_DIRECTIVES`.
+    """
     return to_checksum_address(NULL_ADDRESS)
 
 
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -18,6 +18,7 @@ from nucypher.policy.conditions.utils import (
 
 USER_ADDRESS_CONTEXT = ":userAddress"
 USER_ADDRESS_EIP4361_EXTERNAL_CONTEXT = ":userAddressExternalEIP4361"
+NULL_ADDRESS_CONTEXT = ":nullAddress"
 
 CONTEXT_PREFIX = ":"
 CONTEXT_REGEX = re.compile(":[a-zA-Z_][a-zA-Z0-9_]*")
@@ -94,6 +95,20 @@ def _resolve_user_address(
     return expected_address
 
 
+def _resolve_null_address(
+    providers: Optional[ConditionProviderManager] = None,
+    **context,
+) -> ChecksumAddress:
+    """
+    Returns the null address (0x0000000000000000000000000000000000000000).
+
+    This is a protected context variable that doesn't require any input data.
+    """
+    from nucypher.blockchain.eth.constants import NULL_ADDRESS
+
+    return to_checksum_address(NULL_ADDRESS)
+
+
 _DIRECTIVES = {
     USER_ADDRESS_CONTEXT: partial(
         _resolve_user_address, user_address_context_variable=USER_ADDRESS_CONTEXT
@@ -102,6 +117,7 @@ _DIRECTIVES = {
         _resolve_user_address,
         user_address_context_variable=USER_ADDRESS_EIP4361_EXTERNAL_CONTEXT,
     ),
+    NULL_ADDRESS_CONTEXT: _resolve_null_address,
 }
 
 
--- a/nucypher/policy/conditions/context.py
+++ b/nucypher/policy/conditions/context.py
@@ -108,7 +108,7 @@ def _resolve_null_address(
     Note: The `providers` and `**context` parameters are unused but maintained
     for interface consistency with other resolver functions in `_DIRECTIVES`.
     """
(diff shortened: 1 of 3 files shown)


derekpierre:
> This allows conditions to reference the null address directly, which is useful for:
> Checking token burns (balanceOf null address)
> Ownership validation against the zero address
> Any condition that needs to reference the standard null/zero address

Trying to understand the motivation a bit better - how often do we need to use the null address?

> Checking token burns (balanceOf null address)
> Ownership validation against the zero address

What's the use case for these - could you elaborate a bit more? 

For token burns, does `balanceOf null address` assume only one isolated call to burn as opposed to multiple and done at different times? Would the wallet the tokens was sent from to the null address be more pertinent, as opposed to checking the balance in the null address itself?

theref:
I'm just trying to get the pieces in place for easy to read, high order conditions. Checking that an address != null address is very common, and will likely come up at some point (eg. dead simple guardrail to add to smarttag 2.0)

Burns is just an idea, if you wanted to to use taco signing to build  a defi product/bridge, you're likely to want to check that a tx has happened where the receiver is the null address.

derekpierre:
> I'm just trying to get the pieces in place for easy to read, high order conditions. Checking that an address != null address is very common, and will likely come up at some point (eg. dead simple guardrail to add to smarttag 2.0)
> 
> Burns is just an idea, if you wanted to to use taco signing to build a defi product/bridge, you're likely to want to check that a tx has happened where the receiver is the null address.

You’ll also need to update `taco-web` to add this context variable as an “automatically injected value” (see `context.ts` here: https://github.com/nucypher/taco-web/blob/signing-epic/packages/taco/src/conditions/context/context.ts#L63
). That does two things:
1. Conditions that reference it won’t raise an error in `taco-web` if the condition context doesn’t supply a value when calling decrypt/sign.
2. It prevents users from being able to manually set that variable in the context.

Additionally, you'll need to document this reserved context variable etc.

---

To clarify, I don’t have an objection to using the null-address context variable; it’s essentially a constant that replaces `0x00…00`.

For the broader “higher order” aspect, I view that more as something that would evolve on the `taco-web` side during condition creation. For example, take your burn example (even though the exact semantics are still fuzzy), there could be a `VerifyNullAddressBurnTx` helper that takes parameters (eg. perhaps token address, user address, ...?) and encodes _“check that some tx was a burn of a token by the user address”_ into one or more underlying Conditions/ContractCondition/logical combinations and hides the details of the specific foundational conditions used to craft the entire condition. That helper might internally use the null-address context variable or simply use `0x00…00` directly. From the user perspective they wouldn't even need to specify the "null address" (context variable or explicitly).

We would design wrappers such as these in `taco-web` that internally generate the appropriate lower-level condition(s).

For display purposes, there’s only so much we can do, since we still need to rely on the foundational conditions that the nodes use.

tests/unit/conditions/test_context.py
```diff
@@ -313,3 +314,45 @@ def test_user_address_context_variable_verification(
...
         get_context_value(context_variable_name, **invalid_signature_context)
+
+
+def test_null_address_context_variable():
+    """Test that :nullAddress resolves to the null address without requiring context data."""
+    from eth_utils import to_checksum_address
+
+    from nucypher.blockchain.eth.constants import NULL_ADDRESS
+    from nucypher.policy.conditions.evm import ContractCondition
+    from tests.constants import TESTERCHAIN_CHAIN_ID
```
cygnusv: Can we move this to the initial import block?

nucypher/policy/conditions/context.py
```diff
@@ -94,6 +95,20 @@ def _resolve_user_address(
...
+def _resolve_null_address(
+    providers: Optional[ConditionProviderManager] = None,
+    **context,
+) -> ChecksumAddress:
+    """
+    Returns the null address (0x0000000000000000000000000000000000000000).
+
+    This is a protected context variable that doesn't require any input data.
+    """
+    from nucypher.blockchain.eth.constants import NULL_ADDRESS
```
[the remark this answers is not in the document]
cygnusv: I think this import can be moved to the top of the file since `nucypher/blockchain/eth/constants.py` has no imports, so there shouldn't be any circular import issues.

cygnusv (approved):
LGTM! just some minor stylistic comments. While I agree with the general idea of this PR, I also agree with @derekpierre that we should strive for adding higher-level functionality in `taco-web`. Reasoning about complex conditions using condition lingo directly is becoming very difficult, and we should discuss where's the best layer to add these new abstractions; `taco-web` feels like a better layer to introduce complex, use-case specific conditions, but I think there's room for discussion here (of course, this is outside the scope of this function).

nucypher/policy/conditions/context.py
```diff
@@ -94,6 +96,21 @@ def _resolve_user_address(
...
+) -> ChecksumAddress:
+    """
+    Returns the null address (0x0000000000000000000000000000000000000000).
+
+    This is a protected context variable that doesn't require any input data.
+
+    Note: The `providers` and `**context` parameters are unused but maintained
+    for interface consistency with other resolver functions in `_DIRECTIVES`.
+    """
+    return to_checksum_address(NULL_ADDRESS)
```
derekpierre: You can gain some efficiency by just doing the following since the null address is already checksummed (nothing to checksum 😄 ):

```suggestion
    return ChecksumAddress(NULL_ADDRESS)
```

derekpierre (approved):
🎸  - just one suggestion.

theref:
merge?

PR #3673, opened 2025-11-27
[WIP] Lingo: Plain Variable Assignments

KPrasch:
**Type of PR:**
Feature

**Required reviews:** 
2

### DIFF
--- a/nucypher/policy/conditions/base.py
+++ b/nucypher/policy/conditions/base.py
@@ -54,7 +54,12 @@ class _Serializable:
         return instance
 
     def _validate(self, **kwargs):
-        errors = self.Schema().validate(data=self.to_dict())
+        schema = self.Schema()
+        # When validating a directly-constructed object (not deserializing from user input),
+        # allow optional fields like returnValueTest to be None. This is indicated by the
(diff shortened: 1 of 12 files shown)


derekpierre:
Interesting. Still processing but some things that jump out at me: 

1. Taking a step back, philosophically, what is a "Condition" without a `returnValueTest`? It feels like this is no longer a "Condition" but something different i.e. the resulting value is not compared against some other value 

2. Building on 1) some time ago we did this separation of a "Condition" into an `ExecutionCall` and a `returnValueTest` (on the Python side), so it feels like `ExecutionCall` is an existing abstraction that can be leveraged for execution without a `returnValueTest`. See background information in the description of #3500 - we just didn't take it that far since it was out of scope at the time.

3. Building on 2) there was a prior musing about `SequentialConditions` possibly being made up of a combination of "Condition Variables" based on "Conditions" and "Execution Variables" based on "Execution Calls" - see #3555 for an earlier exploration of that idea that you and I chatted about.

4. Any validation logic we have in python needs to be repeatable in `taco-web`. So I would recommend having PRs for both to ensure that whatever logic we employ is repeatable in both codebases.

PR #3677, opened 2025-12-10
Check ETH funding on startup

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [x] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Much like we check for POL balance for submitting DKG ritual txs, we need to check for ETH balance for submitting Signing ritual txs.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/blockchain/eth/actors.py
+++ b/nucypher/blockchain/eth/actors.py
@@ -42,6 +42,7 @@ from nucypher.blockchain.eth.agents import (
     TACoApplicationAgent,
     TACoChildApplicationAgent,
 )
+from nucypher.blockchain.eth.clients import EthereumClient
 from nucypher.blockchain.eth.constants import (
     NULL_ADDRESS,
     PUBLIC_CHAINS,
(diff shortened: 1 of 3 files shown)


KPrasch:
I'm surprised this check does not already exist, but this essentially expands the existing funding check to both ETH and POL, correct?

cygnusv:
> I'm surprised this check does not already exist, but this essentially expands the existing funding check to both ETH and POL, correct?

To be fair, we dropped the need to have ETH funds when we moved the protocol to Polygon, so this is a new requirement. We have to think carefully how we can introduce this to the general node operators population.

derekpierre:
> I'm surprised this check does not already exist, but this essentially expands the existing funding check to both ETH and POL, correct?

Correct. The `SigningCoordinator` is on Ethereum so the node operator wallets need some ETH.

In the past, I think the confirm activity functionality required it (which burned through a bunch of ETH 😬 ) but as @cygnusv  pointed out, nodes haven't needed ETH since we got rid of that and shifted to the `Coordinator` contract with all transactions executed on Polygon - submitting ferveo key, dkg ritual txs. 

> We have to think carefully how we can introduce this to the general node operators population.

Totally - this is the discussion I wanted to have.

It is indeed a new requirement, and something we'll need to announce to nodes before issuing a new `nucypher` release. We don't want instances of watchtower to automatically upgrade nodes to v7.7.0 whose operators aren't funded with ETH or else the nodes will not start. Something like announce to nodes the new signing functionality and that some ETH is needed (how much is TBD), and give them a period of time to fund the operator before releasing v7.7.0 - more thought needed here on the best course of action.

For now, this check will be utilized for our internal `mainnet` testing. 

(cc @arjunhassard )

theref (approved):

vzotova (approved):

nucypher/blockchain/eth/actors.py
```diff
@@ -1536,42 +1537,54 @@ def get_staking_provider_address(self) -> ChecksumAddress:
...
-                else:
-                    emitter.message(
-                        f"! Operator {self.operator_address} is not funded with MATIC",
-                        color="yellow",
-                    )
+            if not funded_pol:
+                funded_pol = self._is_funded(taco_child_client, "POL", emitter)
+
+            if not funded_eth:
+                funded_eth = self._is_funded(taco_app_client, "ETH", emitter)
```
cygnusv, line 1587: 👨‍🍳 😘

cygnusv (approved):

PR #3683, opened 2026-01-13
feat(conditions): add exponent operator and expand condition limits

theref:
## Summary

Enhances condition capabilities by adding an exponent operator and increasing condition limits to support more complex condition chains.

## Changes

- **Add `toTokenBaseUnits` operator  for variable operations** - enables use cases like ERC-20 token decimal conversion (e.g., `10 ** 18` for wei conversion)
- **Increase `SequentialCondition.MAX_NUM_CONDITIONS`** from 10 to 20
- **Increase `MAX_MULTI_CONDITION_NESTED_LEVEL`** from 2 to 4

## Testing

- Added unit tests for the exponent operator including edge cases (zero exponent, negative exponent, float base)


tests/unit/conditions/test_variable_operation.py
```diff
@@ -453,3 +456,77 @@ def test_tojson_type_errors():
...
+    # Cascading: get decimals, compute multiplier, then multiply amount
+    # Simulates: amount * (10 ** decimals)
+    initial = 10  # base for exponent
+    operations = [
+        VariableOperation(operation="**=", value=6),  # 10 ** 6 = 1000000
+        VariableOperation(
+            operation="*=", value=100
+        ),  # 1000000 * 100 = 100000000 (100 USDC in smallest unit)
+    ]
+    result = VariableOperation.evaluate_operations(operations, initial)
```
derekpierre: If we are trying to mimic what will be needed in the condition involving USDC or any other non-18-decimal ERC20, I think we may need to tweak this test a little. Variable operations are only done on results of conditions/calls, i.e. the "initial value" is obtained from a condition/call.

So then the question is what condition/call produces `10` for the initial value? I'm not sure if there is such a case unless I'm missing something.

So in the USDC case seems like it would be `6` as the result of the `decimals()` call to the ERC20 contract. So given an initial value of `6`, what operations do we do on it to get the final token amount value we want? 

OR 

the initial value could be the amount, `100` in this case, obtained from parsing the discord payload then what operations do we do on the `100` to get the correct final token amount? 

While useful for other cases, I'm unsure if `**=`,  solves this particular problem in either of the above situations. Let me know if I've misunderstood.

Something to consider could be to have an operation like `to_token_base_units(decimals)` (needs better name 😅 ), so something like:

```python
VariableOperation(operation="to_token_base_units", value=6)
```
which is basically like `ethToWei` but instead takes a decimals parameter.

In the real condition, the operation may be something like
```python
VariableOperation(operation="to_token_base_units", value=":tokenDecimals")
```
where the initial value would be the amount obtained from the discord payload and `:tokenDecimals` is the condition variable stored from calling `decimals()` on the relevant ERC20 contract.

Thoughts?

theref: yeah this is definitely a limitation of our conditions. The calculation needs to be `:amount * 10 ^ :decimals`, but you're right that we don't have a way to do that. I don't like the idea of adding `to_token_base_units`, I'd rather fix this issue of constants and calculations.

derekpierre: > I don't like the idea of adding to_token_base_units, I'd rather fix this issue of constants and calculations.
You mean 

I'm not sure you can avoid a new operation, but these are very easy to add.

If you are looking for a lower-level and less token specific type of operation, you can add an operation like "*10^=" (or some equivalent) where it multiplies the initial value by some power of 10. It's the same as what `toTokenBaseUnits` would do, just named more generally i.e.
```python
VariableOperation(operation="*10^=", value=":decimals")
```
where initial value is the amount.

You'll need to be cognizant that the multiplication can produce floats, and you'll probably need to cast to int before using the result for token amounts i.e.
```python
operations=[
    VariableOperation(operation="*10^=", value=":decimals"),
    VariableOperation(operation="int")  # cast float to int
]
```

derekpierre: I guess your exponent operation uses `**=`, so following that format would technically yield `*10**=` , but that may/may not be a bit ugly/confusing 😅 

If you are moving forward with adding an operator for powers of 10, then we should agree on a cohesive format for both operations:
- `**=`, `*10**=`
- `**=`, `*=10**`
- `^=`,  `*10^=`
- something explicit like `pow=`, `*10pow=`
OR 
... 

The main issue is being clear and consistent.

theref:
## Alternative Solution: `initial_value` Parameter for ConditionVariable

The core issue isn't about token decimals specifically—it's that **we can't use literal constants in condition chains**. To calculate `:amount * 10 ** :decimals`, we need the constant `10`, but currently must pass it as `:base` in the context.

### Proposed Approach
Add an optional `initial_value` parameter to `ConditionVariable` that allows embedding constants directly:

```python
SequentialCondition(
    condition_variables=[
        ConditionVariable(
            var_name="multiplier",
            initial_value=10,  # Constant embedded in condition definition
            operations=[
                VariableOperation(operation="**=", value=":decimals")
            ]
        ),
        ConditionVariable(
            var_name="decimals",
            condition=erc20_decimals_call
        ),
        # :multiplier = 10 ** :decimals
    ]
)
```

### Implementation
- Make `condition` field optional in ConditionVariable schema
- Add `initial_value` field (mutually exclusive with `condition`)
- In `SequentialCondition.verify()`: if `initial_value` is set, skip condition evaluation and use the constant
- Validation: exactly one of `condition` or `initial_value` must be provided
- Require `operations` when `initial_value` is used (otherwise it's just a useless constant)

### Pros
- **Generic solution**: Solves the constant problem broadly, not just for powers of 10
- **No new operators**: Works with existing `**=`, `*=`, etc.
- **Flexible**: Can use any constant type (int, float, string, list, dict)
- **Clear intent**: `initial_value=10` is explicit and readable
- **Composable**: Can chain multiple constant-based operations

### Cons
- **Breaking change potential**: Must carefully handle serialization/deserialization for backwards compatibility
- **Adds complexity**: Now two ways to initialize a ConditionVariable
- **Validation overhead**: Need to ensure XOR logic between `condition` and `initial_value`

### Edge Cases to Consider
1. **Serialization**: BigInt handling for large constants
2. **Type coercion**: What if `initial_value=":someVar"` (string that looks like context var)? Probably should NOT resolve it
3. **Schema validation**: Ensure JSON schema properly reflects the mutual exclusivity
4. **Error messages**: Clear errors when both or neither are provided
5. **Backwards compat**: Existing conditions without `initial_value` continue working unchanged

Thoughts on which direction?

tests/unit/conditions/test_variable_operation.py
derekpierre: You could even create a "*pow=" with a base and exponent:
```python
VariableOperation(operation="*pow=", value=[base, exponent]),
```
So 
```python
VariableOperation(operation="*pow=", value=[10, ":decimals"]),
```

derekpierre:
> To calculate :amount * 10 ** :decimals, we need the constant 10, but currently must pass it as :base in the context.

Not sure what you mean by ":base" in the context.

> The core issue isn't about token decimals specifically—it's that we can't use literal constants in condition chains. 

For this case, the core issue is the ordering of how the calc needs to be done, not the use of a literal constant. We can easily use 10 as a constant in the lingo/operations directly, it's that the initial value used to start with for the calc doesn't allow this specific calc to be done given the existing available operators. I think a simpler solution is just adding a new operator (see latest comments above); we can make it generic and not specific to token decimals.

As outlined in the "Cons" and "Edge Cases to Consider", changing the schema is a much more complex process, and isn't really necessary to resolve this issue; unless I'm missing more to the story.

theref:
`:base` would be one way of inserting the value of 10. but because it's then from context it's open to maniplution (by the discord bot i this case)

>  the core issue is the ordering of how the calc needs to be done,

yep exactly :) do we not want to just fix that? rather than adding a random operator that won't be used for anything else

derekpierre:
> :base would be one way of inserting the value of 10. but because it's then from context it's open to manipulation (by the discord bot i this case)

👍  - Thanks for the clarification.

> yep exactly :) do we not want to just fix that? rather than adding a random operator that won't be used for anything else

But is something generic like `*pow=` any more random than `**=`? We can make the operator generic enough that it can use any base (not just 10), and any exponent, and therefore not just specific to token arithmetic.

The crux of my concern is that adding a new operator is like 30 lines of code (excluding tests) and about an hour of work. In general, making lingo closer to Turing complete is basically just adding operators. 

Whereas updating schema etc. is a significant endeavour and unclear how much it will be used for other cases; so same as your concern with the operator but with significant less investment of time ... if we encounter more data points, we can revisit. The larger investment of time vs equivalent functionality via a new operation in this case is the main worry.

tests/unit/conditions/test_variable_operation.py
```diff
@@ -13,6 +13,9 @@
     ("*=", 2, 3, 6),
     ("/=", 2, 6, 3.0),
     ("%=", 2, 5, 1),
+    ("pow", [3, 2], None, 9),  # 3 ** 2 = 9
+    ("pow", [2, 3], None, 8),  # 2 ** 3 = 8
+    ("pow", [10, 18], None, 1000000000000000000),  # 10 ** 18 for token decimals
```
derekpierre: Operations are always performed on an `initial value`, so `initial value` should never be `None`.

We may be conflating two things. 

Originally, you coded an operator that could do `<initial_value> ** <exponent>` eg. 2 ** 3 = 8. We discovered that wasn't sufficient because an `<initial_value>` of 10 wasn't available to do `10 ** 6`. 

Therefore we want to be able to something like `<initial_value> * 10 ** 6` for USDC. To generalize for calcs other than USDC, ERC20 tokens etc. we want `<base>` and `<exponent>` to be customizable, so we want an operation that does `<initial_value> * (<base> ** <exponent>)`.

For VariableOperations `(a, b)`, `a` is **always** the initial value,

So, it feels like there are two operations being discussed:
1. "pow" that does `<initial_value> ** <value>`

```python
VariableOperation(operation="pow", value=<value>)
```

this is just the `operator.pow(a,b)`, so the entry for `_OPERATOR_FUNCTIONS` is just:
```python
"pow": pyoperator.pow,
```

2. "*pow=" (or some equivalent name) that does `<initial_value> * (<base> ** <exponent>)`
```python
VariableOperation(operation="*pow=", value=[<base>, <exponent>])
```

this is a customized lamda function entry for `_OPERATOR_FUNCTIONS`, so:
```python
"*pow=": lambda a, b: pyoperator.mul(a, pyoperator.pow(*b))
```
This is what will allow you to do the calc for USDC i.e. `:amount (obtained from discord payload) * (10**6)`

Perhaps we only want 2), but that is different than just `pow`.

Let me know if I've misunderstood/misclassified what is going on here. Also, happy to do a voice chat to resolve.

derekpierre (changes requested):

tests/unit/conditions/test_variable_operation.py
theref: no you're totally right, let me fix that. i think `*pow=` is good, even though it's weird

tests/unit/conditions/test_variable_operation.py
```diff
@@ -453,3 +457,86 @@ def test_tojson_type_errors():
...
+    operations = [
+        VariableOperation(
+            operation="*pow=", value=[10, 6]
+        ),  # 250.5 * (10 ** 6) = 250500000
+    ]
+    result = VariableOperation.evaluate_operations(operations, initial)
+    assert result == 250500000
+
+
+def test_pow_operator_edge_cases():
```
derekpierre: Nice tests.

derekpierre (dismissed):
🎸 - looks good!

We'll need to make the corresponding updates to `taco-web`.

tests/unit/conditions/test_variable_operation.py
```diff
@@ -13,6 +13,10 @@
     ("*=", 2, 3, 6),
     ("/=", 2, 6, 3.0),
     ("%=", 2, 5, 1),
+    ("*pow=", [3, 2], 1, 9),  # 1 * (3 ** 2) = 9
+    ("*pow=", [2, 3], 1, 8),  # 1 * (2 ** 3) = 8
+    ("*pow=", [10, 18], 1, 1000000000000000000),  # 1 * (10 ** 18) for wei conversion
+    ("*pow=", [10, 6], 100, 100000000),  # 100 * (10 ** 6) = 100000000 for USDC decimals
```
cygnusv:
```suggestion
    ("*pow=", [10, 18], 1, 1_000_000_000_000_000_000),  # 1 * (10 ** 18) for wei conversion
    ("*pow=", [10, 6], 100, 100_000_000),  # 100 * (10 ** 6) = 100000000 for USDC decimals
```
Just for extra readability

tests/unit/conditions/test_variable_operation.py
```diff
@@ -453,3 +457,86 @@ def test_tojson_type_errors():
...
+
+    # Another example: 250.5 tokens with 6 decimals
+    initial = 250.5
+    operations = [
+        VariableOperation(
+            operation="*pow=", value=[10, 6]
+        ),  # 250.5 * (10 ** 6) = 250500000
+    ]
+    result = VariableOperation.evaluate_operations(operations, initial)
+    assert result == 250500000
```
cygnusv, line 486: I have a negative test here:
https://github.com/cygnusv/nucypher/commit/fcb4345e87db6349358e19479d28a269eb55452f

cygnusv (changes requested):
I have a negative test, but I don't have any constructive ideas yet on how to fix it. I thought these issues were not going to appear since I raised a similar comment a while ago (<https://github.com/nucypher/nucypher/pull/3658#issuecomment-3373172957>), which I think was resolved with a generic solution.

nucypher/policy/conditions/utils.py
```diff
@@ -41,6 +41,40 @@ def _wei_to_eth(value) -> Union[int, decimal.Decimal]:
...
+        with localcontext() as ctx:
+            ctx.prec = 999
+            # Convert to Decimal using string representation to preserve precision
+            if isinstance(value, float):
+                d_value = Decimal(str(value))
+            elif isinstance(value, Decimal):
+                d_value = value
+            else:
+                d_value = Decimal(value)
+            result = d_value * (Decimal(10) ** decimals)
```
cygnusv, line 70: Note there's a `to/from_wei_decimals` in `eth_utils.currency` (https://eth-utils.readthedocs.io/en/stable/utilities.html#to-wei-decimals-value-decimals-integer), in case we want to have a unified function here. Not a strong comment anyway, just a potential improvement.

cygnusv (approved):

newsfragments/3683.feature.rst
```diff
@@ -0,0 +1 @@
+Added ``toTokenBaseUnits`` operator for condition variable operations (takes ``decimals`` as value to convert token amounts to base units, e.g., ``1.5`` with ``decimals=18`` becomes ``1500000000000000000``), useful for ERC-20 token comparisons, and increased condition limits (``SequentialCondition.MAX_NUM_CONDITIONS`` to 20, ``MAX_MULTI_CONDITION_NESTED_LEVEL`` to 4).
```
derekpierre: Slight reword and used a different example than decimals=18 since we already have `ethToWei` that does that.

```suggestion
Added ``toTokenBaseUnits`` operator for variable operations, which converts token amounts to base units (e.g., ``1.5`` with ``decimals=6`` becomes ``1500000``), useful for ERC-20 token comparisons. Also increased condition limits (``SequentialCondition.MAX_NUM_CONDITIONS`` to 20 and ``MAX_MULTI_CONDITION_NESTED_LEVEL`` to 4).
```

nucypher/policy/conditions/utils.py
derekpierre: I feel a bit more strongly about this than @cygnusv . Let's use `to_wei_decimals(...)` instead since you may not even need to define this function and instead just use a lambda function that calls `to_wei_decimals(...)` directly.

nucypher/policy/conditions/lingo.py
```diff
@@ -377,6 +378,7 @@ def _to_hex(value):
     "%=": pyoperator.mod,
     "index": lambda a, b: a[b],
     "round": lambda a, b: round(a, b),
+    "toTokenBaseUnits": _to_token_base_units,
```
derekpierre, line 381: Thinking more about this naming, and the available funciton `to_wei_decimals`, should we call this operation `toWeiDecimals` instead? It would better link `ethToWei` (default decimals is 18) and this operations (decimals must be specified).

Alternatively, you could modify `ethToWei` to take an optional argument, to cover both cases. No argument is 18 decimals, otherwise use argument for decimals. Might be leaning to this instead.

wdyt? (cc @cygnusv )

theref: but wei is very eth specific, it doesn't make sense in the context of tBTC or USDC

theref: oh wait I understand, it's just to match the underlying function. Ok, let's do that

derekpierre: Since `eth_utils` is exposing a function named `to_wei_decimals(...)`,`wei` is more of the smallest unit and not specifically 18 decimals.

Perhaps a good counterpoint to my comment would be someone in the future potentially wanting to convert `BTC` to `sats` in which case `toTokenBaseUnits` would be better.

👍  - You've convinced me to keep it.

(While we are not changing the operation name, the backend calc can still use the `to_wei_decimals(...)` function.)

theref: `to_wei_decimals` was added in eth-utils v5.3. and we're currently on 2.3.2 - that's a big jump and it's breaking other dependencies

derekpierre: Surprised how old our version is.

That's annoying; might be due to, https://github.com/nucypher/nucypher/blob/v7.7.x/pyproject.toml#L49, and is something we can look more into some other time.

Not much we can do, so we can continue using your `_to_token_base_units()` function.

nucypher/policy/conditions/utils.py
derekpierre: Can't use `to_wei_decimals` since our eth-utils version is too old and doesn't have it. See https://github.com/nucypher/nucypher/pull/3683#discussion_r2705293181

nucypher/policy/conditions/lingo.py
theref: eth-ape was one of them yeah, also eip712

derekpierre (approved):
Nice work 🎸 .

Just one minor comment/suggestion for the newsfragment.

PR #3685, opened 2026-01-14
feat(conditions): add subIndices for ABI parameter nested structure navigation

theref:
## Summary

Adds `subIndices` to `AbiParameterValidation` for navigating nested ABI structures (arrays, tuples, and combinations), enabling validation of batch transactions and multicall operations.

### Implementation
- Add `subIndices` / `sub_indices` field to `AbiParameterValidation` schema as a list of sequential indices
- The ABI type determines interpretation at each step: array types (`[]`) use array indexing, tuple types (`(...)`) use tuple field indexing
- Implement runtime value extraction via `get_value()` with bounds checking
- Add schema-time structural validation using `resolve_abi_type_with_indices()` against ABI type definitions
- Extract common `_split_comma_separated_types()` helper to deduplicate tuple parsing logic
- Support arbitrarily nested structures: arrays of tuples, tuples of arrays, etc.

### Tests
- Basic array indexing for simple array types (`address[]`, `uint256[]`)
- Arrays of tuples for batch operations (`(address,uint256,bytes)[]`)
- Tuples of arrays for the reverse nesting pattern (`(address[],uint256[],bytes)`)
- Error cases: non-indexable type validation, out-of-range index checks at both schema and runtime
- Integration with existing nested ABI validation (`nested_abi_validation`)
- Unit tests for `parse_tuple_fields()` and `resolve_abi_type_with_indices()` helpers

## Use Cases

Enables validation of:
- Batch transfers: `batchTransfer(address[], uint256[])`
- Multicall operations: `executeBatch((address,uint256,bytes)[])`
- Tuple-of-array patterns: `multiTransfer((address[],uint256[],bytes))`
- Any function with nested array/tuple parameters

## Example

```json
{
  "allowedAbiCalls": {
    "executeBatch((address,uint256,bytes)[])": [
      {
        "parameterIndex": 0,
        "subIndices": [0, 1],
        "returnValueTest": {
          "comparator": "==",
          "value": ":expectedAmount"
        }
      }
    ]
  }
}
```

This validates the amount field (tuple index 1) in the first batch item (array index 0) using `subIndices: [0, 1]`.


derekpierre:
I think the existing "index" operation in `VariableOperation` may already solve the issue here:

```python
"executeBatch((address,uint256,bytes)[])": [
                    AbiParameterValidation(
                        parameter_index=0,
                        return_value_test=ReturnValueTest(
                            comparator="==",
                            value=":expectedAmount",
                            operations=[
                                VariableOperation(operation="index", value=0),
                                VariableOperation(operation="index", value=1),
                            ]
                        ),
                    ),
                ]
 ```
 
 Try it out for yourself and let me know.

**Comment by theref** [comment]
I'm still running into issues with the batch transactions. We need to validate batch transactions where the `calldata` is `execute(bytes32,bytes)`. The second parameter `(bytes)` contains ABI-encoded `(address,uint256,bytes)[]` (an array of Execution tuples).

Currently, Taco's `signing-abi-attribute` condition can:
- Match the function signature `execute(bytes32,bytes)`
- Access parameter 1 as raw bytes

But we cannot apply index operations to extract values from the encoded tuple array because the bytes aren't decoded as `(address,uint256,bytes)[]`.

I tried using `execute(bytes32,(address,uint256,bytes)[])` as the signature, but this has a different function selector than `execute(bytes32,bytes)`, so it doesn't match the actual `calldata`.

We need a  way to tell Taco: "parameter 1 is bytes, but decode it as `(address,uint256,bytes)[]` before applying operations.

File: nucypher/policy/conditions/signing/base.py [hunk 1]
```diff
@@ -277,8 +298,27 @@ def validate_allowed_abi_calls(
...
+                        # Note: We cannot validate the array index bounds at schema validation time
+                        # since array length is only known at runtime
+
                     if parameter_value_check.index_within_tuple is not None:
                         tuple_args = arg_types[parameter_value_check.parameter_index]
+
+                        # If we have index_within_array, the arg_type is an array of tuples (e.g., "(address,uint256,bytes)[]")
+                        # We need to strip the [] to get the tuple type
+                        if parameter_value_check.index_within_array is not None:
+                            if tuple_args.endswith("[]"):
```
derekpierre: We need to do something similar for the `nested_abi_validation` below, see L343.

You'll need to strip out the trailing "[]" if `index_wthin_array` is defined. Something like:
```
                        if parameter_value_check.index_within_array is not None:
                            arg_type_to_check = arg_type_to_check.rstrip("[]")
```

Your tests have covered checking one level of abi validation, but not for the `nested_abi_validation`.

It's the case where you have something like the following (diff from modified test `test_signing_object_abi_attribute_condition_more_than_2_levels_nested_calldata` that fails without the above change):
```diff
--- a/tests/unit/conditions/test_signing_condition.py
+++ b/tests/unit/conditions/test_signing_condition.py
@@ -915,8 +915,9 @@ def test_signing_object_abi_attribute_condition_more_than_2_levels_nested_callda
 
     # Level 2: eg. proxy.execute(token_address, 0, transfer_data)
     proxy_execute_data = encode_function_call(
-        "execute((address,uint256,bytes))",
-        [(get_random_checksum_address(), 0, transfer_data)],
+        "execute((address,uint256,bytes)[])", [
+            [(get_random_checksum_address(), 0, transfer_data)]
+        ],
     )
 
     # Level 1: smartAccount.execute(proxy_address, 0, proxy_execute_data)
@@ -938,10 +939,11 @@ def test_signing_object_abi_attribute_condition_more_than_2_levels_nested_callda
                         2,
                         nested_abi_validation=AbiCallValidation(
                             {
-                                "execute((address,uint256,bytes))": [
+                                "execute((address,uint256,bytes)[])": [
                                     # only allow transfer call
                                     AbiParameterValidation(
                                         parameter_index=0,
+                                        index_within_array=0,
                                         index_within_tuple=2,
                                         nested_abi_validation=AbiCallValidation(
                                             {"transfer(address,uint256)": []}
```

tests/unit/conditions/test_signing_condition.py
```diff
@@ -436,6 +436,226 @@ def test_invalid_signing_object_abi_attribute_condition():
...
+    condition = SigningObjectAbiAttributeCondition(
+        attribute_name="call_data",
+        abi_validation=AbiCallValidation(
+            {
+                "executeBatch((address,uint256,bytes)[])": [
+                    AbiParameterValidation(
+                        parameter_index=0,
+                        index_within_array=0,
+                        index_within_tuple=1,
+                        return_value_test=ReturnValueTest("==", 1000000),
```
derekpierre, line 604: Currently we purposely establish `index_within_array` before `index_within_tuple` so the above test makes sense. What about the case when it's the flip i.e. array within a tuple; could this be confusing?

For example, `"executeBatch((address[],uint256[],bytes[]))"`:
Something like the following would currently fail validation.
```python
                "executeBatch((address[],uint256[],bytes[]))": [
                    AbiParameterValidation(
                        parameter_index=0,
                        index_within_tuple=1,
                        index_within_array=0,
                        return_value_test=ReturnValueTest("==", 1000000),
                    )
                ]
```

That can be handled by operations of course, but just on the surface, given that we have both `index_within_(tuple|array)` the example above is fair expectation.

You'll also have this problem for trying to do nested validation of an entry from the `bytes[]` array entry at index 2 in the tuple above.

Perhaps we need something like a single parameter with a list of indices that cover both tuples and arrays instead of differentiating between tuples and arrays eg. 
```
   parameter_index=0,
   sub_indices=[1, 0],
   return_value_test=ReturnValueTest("==", 1000000)
```
So, look at parameter index at 0, then take value at sub-index 1 of that, then take value at sub-index 0 of that.

We need to think a bit more on this.

theref: I like the idea of `sub_indices` - it means we don't need either `index_within_(tuple|array)` - although one will stay for backwards compatability

derekpierre: We don't really need backwards compatibility since `ABIParmaterValidation` etc. was never officially released; it is only part of the pending `v7.7.x` branch.

The only user of the already existing `index_within_tuple` is Collab.Land and it is being used for a time-scoped tip tournament with an end date. If we needed to, although I don't expect we'll need to, we could update their signing condition accordingly.

derekpierre: Perhaps you could explore the `sub_indices` strategy (or some alternative) a bit more in depth? We could see where it takes us, and if there are any separate issues it raises.

theref: yep will do. And with regards to CL, this work is to enable fee collection so it will require new conditions anyway

theref:
rebased so ready for review again

nucypher/policy/conditions/signing/base.py
```diff
@@ -162,8 +163,15 @@ class AbiParameterValidation(_Serializable):
...
-            validate=Range(min=0), allow_none=True, required=False
+        # Sequential indices for navigating nested structures (arrays and tuples)
+        # The ABI type determines interpretation at each step:
+        # - If current type ends with "[]" -> array index
+        # - If current type is "(...)'" -> tuple field index
+        sub_indices = fields.List(
+            fields.Integer(validate=Range(min=0)),
+            load_default=None,
+            allow_none=True,
+            required=False,
```
derekpierre, line 172: - `load_default` is already None by default
- `allow_none=True` means that None is a valid value during serialization/deserialization i.e. the field will look like the following in the JSON
```
sub_indices: null
```
instead of just not including the field when the value isn't set.

Both of these should be removed.

```suggestion
            fields.Integer(validate=Range(min=0)),
            required=False,
```

derekpierre:
This is a breaking change, so any existing condition that uses `index_within_tuple` will fail after this code is deployed to nodes.

@theref let's double check what existing conditions on `lynx` / `mainnet` (CL only) will be affected.

nucypher/policy/conditions/types.py
```diff
@@ -199,14 +199,14 @@ class SigningObjectAttributeCondition(_BaseSigningObjectAttributeCondition):
 
 # AbiParameterValidation represents:
 # {
-#     "parameterIndex": str
-#     "indexWithinTuple": int
+#     "parameterIndex": int
+#     "subIndices": [int]  # Sequential indices for navigating nested structures
```
derekpierre:
```suggestion
#     "subIndices": List[int]  # Sequential indices for navigating nested structures
```

nucypher/utilities/abi.py
```diff
@@ -76,6 +76,89 @@ def encode_human_readable_call(human_signature: str, args: list) -> bytes:
...
+                depth += 1
+            elif char == ")":
+                depth -= 1
+            current.append(char)
+
+    if current:
+        fields.append("".join(current).strip())
+
+    if depth != 0:
+        raise ValueError(f"Mismatched parentheses in tuple type: {tuple_type}")
```
derekpierre: This seems to be the same as L26-45. Perhaps this can be made into a common function and used in both locations?

tests/unit/conditions/test_signing_condition.py
```diff
@@ -436,6 +436,314 @@ def test_invalid_signing_object_abi_attribute_condition():
...
+                        )
+                    ]
+                }
+            ),
+        )
+
+
+def test_abi_parameter_validation_array_indexing(condition_provider_manager):
+    """Test sub_indices for simple array types."""
+    from nucypher_core import UserOperation
```
derekpierre: Might as well put this import at the top of the test module instead of repeating it per test.

vzotova (approved):

tests/unit/conditions/test_signing_condition.py
```diff
@@ -436,6 +436,314 @@ def test_invalid_signing_object_abi_attribute_condition():
...
+
+    # Create test data with an array of addresses
+    recipient1 = "0x" + "11" * 20
+    recipient2 = "0x" + "22" * 20
```
derekpierre: This is a fixture that you can use (in this test and others) that will save you from needing to generate your own addresses, `get_random_checksum_address` - see https://github.com/nucypher/nucypher/blob/v7.7.x/tests/unit/test_node_latency_collector.py#L14 for an example.

Same for `sender=` below.

tests/unit/test_abi_utils.py
```diff
@@ -152,3 +154,81 @@ def test_encode_decode_human_readable_abi(human_signature, expected_method_name,
...
+
+def test_parse_tuple_fields_invalid():
+    with pytest.raises(ValueError, match="Not a tuple type"):
+        parse_tuple_fields("address")
+
+    with pytest.raises(ValueError, match="Not a tuple type"):
+        parse_tuple_fields("uint256[]")
+
+
+@pytest.mark.parametrize(
```
derekpierre, line 184: Nice tests!

tests/unit/conditions/test_signing_condition.py
```diff
@@ -436,6 +436,314 @@ def test_invalid_signing_object_abi_attribute_condition():
...
+                        AbiParameterValidation(
+                            parameter_index=0,
+                            sub_indices=[5],  # ERROR: tuple only has 3 fields
+                            return_value_test=ReturnValueTest("==", ":userAddress"),
+                        )
+                    ]
+                }
+            ),
+        )
+
```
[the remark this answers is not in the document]
derekpierre: ^ Agree here. I know we match the text which is great, but it is also good to use the more specific exception as well.

nucypher/policy/conditions/signing/base.py
```diff
@@ -162,8 +163,15 @@ class AbiParameterValidation(_Serializable):
...
         )
```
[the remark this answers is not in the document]
derekpierre: ^ Given the change of direction for the PR, it'll be good to update the PR title and description.

tests/unit/conditions/test_signing_condition.py
```diff
@@ -335,16 +336,16 @@ def test_invalid_signing_object_abi_attribute_condition():
             ),
         )
 
-    # invalid tuple arg
-    with pytest.raises(ValueError, match="is not a tuple"):
+    # invalid sub_indices on non-indexable type
+    with pytest.raises(ValueError, match="not indexable"):
```
derekpierre, line 340: Given this is on condition creation this should be `InvalidCondition` (most specific exception).


```suggestion
    with pytest.raises(InvalidCondition, match="not indexable"):
```

tests/unit/conditions/test_signing_condition.py
```diff
@@ -353,17 +354,15 @@ def test_invalid_signing_object_abi_attribute_condition():
         )
 
     # tuple index out of range
-    with pytest.raises(
-        ValueError, match="Tuple value index '3' for parameter is out of range"
-    ):
+    with pytest.raises(ValueError, match="out of range"):
```
derekpierre, line 357: Same comment as above

```suggestion
    with pytest.raises(InvalidCondition, match="out of range"):
```

tests/unit/conditions/test_signing_condition.py
```diff
@@ -436,6 +437,318 @@ def test_invalid_signing_object_abi_attribute_condition():
         )
 
 
+def test_abi_parameter_validation_sub_indices_errors():
+    """Test error handling for invalid sub_indices."""
+    # Using sub_indices on non-indexable type should fail at schema validation
+    with pytest.raises(ValueError, match="not indexable"):
```
derekpierre, line 443:
```suggestion
    with pytest.raises(InvalidCondition, match="not indexable"):
```

tests/unit/conditions/test_signing_condition.py
```diff
@@ -436,6 +437,318 @@ def test_invalid_signing_object_abi_attribute_condition():
...
+                            sub_indices=[0],  # ERROR: address is not indexable
+                            return_value_test=ReturnValueTest("==", ":userAddress"),
+                        )
+                    ]
+                }
+            ),
+        )
+
+    # Using sub_indices with out-of-range tuple index should fail
+    with pytest.raises(ValueError, match="out of range"):
```
derekpierre, line 460:
```suggestion
    with pytest.raises(InvalidCondition, match="out of range"):
```

nucypher/policy/conditions/signing/base.py
```diff
@@ -277,37 +296,29 @@ def validate_allowed_abi_calls(
...
+                    if parameter_value_check.sub_indices:
+                        try:
+                            final_type = resolve_abi_type_with_indices(
+                                arg_type, parameter_value_check.sub_indices
+                            )
+                        except ValueError as e:
                             raise ValidationError(
-                                f"Tuple value index '{parameter_value_check.index_within_tuple}' for parameter is out of range for "
-                                f"the ABI decoded tuple '{tuple_args}'. "
+                                f"Invalid sub_indices for parameter {parameter_value_check.parameter_index} "
```
derekpierre: These messages get passed back to `porter`/`taco-web` which doesn't use snake case. We can just make the exception message more generic:


```suggestion
                                f"Invalid sub indices for parameter {parameter_value_check.parameter_index} "
```

nucypher/policy/conditions/signing/base.py
```diff
@@ -196,29 +202,42 @@ def make(self, data, **kwargs):
...
+        At runtime, both arrays and tuples are represented as Python tuples/lists,
+        so we simply index into them without type distinction.
+        """
+        value = args[self.parameter_index]
+
+        if self.sub_indices:
+            for i, idx in enumerate(self.sub_indices):
+                if not isinstance(value, (list, tuple)):
+                    raise ValueError(
+                        f"Cannot index into {type(value).__name__} at sub_indices position {i}; "
```
derekpierre: Same comment about `sub_indices` and snake case. 
```suggestion
                        f"Cannot index into {type(value).__name__} at sub indices position {i}; "
```

nucypher/policy/conditions/signing/base.py
```diff
@@ -196,29 +202,42 @@ def make(self, data, **kwargs):
...
+                        f"expected list or tuple"
+                    )
+                if idx >= len(value):
+                    raise ValueError(
+                        f"Index {idx} at sub_indices position {i} is out of range "
```
derekpierre:
```suggestion
                        f"Index {idx} at sub indices position {i} is out of range "
```

nucypher/policy/conditions/signing/base.py
```diff
@@ -277,37 +296,29 @@ def validate_allowed_abi_calls(
...
-                            tuple_args = arg_type_to_check.strip("(").strip(")")
-                            arg_type_to_check = tuple_args.split(",")[
-                                parameter_value_check.index_within_tuple
-                            ]
-                        if arg_type_to_check != "bytes":
+                        if final_type != "bytes":
                             raise ValidationError(
-                                f"Nested ABI validation is only supported for bytes type, but found '{arg_type_to_check}'."
+                                f"Nested ABI validation is only supported for bytes type, "
+                                f"but sub_indices resolve to '{final_type}'."
```
derekpierre:
```suggestion
                                f"but sub indices resolve to '{final_type}'."
```

nucypher/policy/conditions/signing/base.py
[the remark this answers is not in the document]
derekpierre: ^ Let's update the title and description of the PR in case in the future we need to search for it.

newsfragments/3685.feature.rst
```diff
@@ -0,0 +1 @@
+Add ``subIndices`` parameter to ABI parameter validation for signing conditions, enabling validation of nested array and tuple elements in batch transactions and multicall operations. This unified approach replaces the need for separate array/tuple index fields and supports arbitrary nesting patterns.
```
derekpierre:
```suggestion
Add ``sub_indices`` parameter to ABI parameter validation for signing conditions, enabling validation of nested array and tuple elements in batch transactions and multicall operations. This unified approach replaces the need for separate array/tuple index fields and supports arbitrary nesting patterns.
```

nucypher/policy/conditions/signing/base.py
```diff
@@ -317,7 +328,10 @@ def make(self, data, **kwargs):
     def __init__(self, allowed_abi_calls: Dict[str, List[AbiParameterValidation]]):
         self.allowed_abi_calls = allowed_abi_calls
 
-        self._validate()
+        try:
+            self._validate()
+        except ValueError as e:
+            raise InvalidCondition(f"{e}")
```
derekpierre: How come we need this? `ABICallValidation` is not a condition so the ValueError should be raised. Whatever condition is using it for validation should be the one to catch and raise the `InvalidCondition`.

derekpierre (dismissed):

tests/unit/conditions/test_signing_condition.py
```diff
@@ -337,7 +337,7 @@ def test_invalid_signing_object_abi_attribute_condition():
         )
 
     # invalid sub_indices on non-indexable type
-    with pytest.raises(InvalidCondition, match="not indexable"):
+    with pytest.raises(ValueError, match="not indexable"):
```
derekpierre, line 340: Oh I see why you needed `ValueError`, it's not being raised from the condition, it's being raised from the object creation on L343.

Apologies for missing that originally 🤦 .

derekpierre (approved):
🎸  - we just need to be strategic about when we merge this PR. Before merging, let's check whether any existing cohorts (`mainnet` or `lynx`) would break if we merge and nodes get updated.

PR #3690, opened 2026-01-26
Scan for DKG/Signing events less often

derekpierre:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [x] Other

**Required reviews:** 
- [ ] 1
- [x] 2
- [ ] 3

**What this does:**
Now that we are scanning for both DKG and Signing ritual events, this could exhaust RPC free tier usage credits. The biggest consumption of credits is for event scanning, so we should do it less often.

There's a balance between not exhausting all free tier credits and ensuring nodes remain responsive to ritual events. That being said, given current usage patterns, nodes don't need to be THAT responsive - rituals tend to occur way before usage.

Let's start with cutting the calls in half by doubling the interval. We can adjust in a separate PR once we collect some data. It's been tricky always having the base this PR over updated changes and then update lynx nodes. Instead let's have this change be the base that we potentially modify later on if needed.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/blockchain/eth/trackers/events.py
+++ b/nucypher/blockchain/eth/trackers/events.py
@@ -45,7 +45,7 @@ class EventActuator(EventScanner):
 class EventScannerTask(SimpleTask):
     """Task that runs the event scanner in a looping call."""
 
-    INTERVAL = 120  # seconds
+    INTERVAL = 180  # 3 mins in seconds
 
     def __init__(self, scanner: Callable):
         self.scanner = scanner
--- a/nucypher/blockchain/eth/trackers/events.py
+++ b/nucypher/blockchain/eth/trackers/events.py
@@ -45,7 +45,7 @@ class EventActuator(EventScanner):
 class EventScannerTask(SimpleTask):
     """Task that runs the event scanner in a looping call."""
 
-    INTERVAL = 180  # 3 mins in seconds
+    INTERVAL = 240  # 4 mins in seconds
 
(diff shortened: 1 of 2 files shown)


nucypher/blockchain/eth/trackers/events.py
```diff
@@ -45,7 +45,7 @@ def process_event(
 class EventScannerTask(SimpleTask):
     """Task that runs the event scanner in a looping call."""
 
-    INTERVAL = 120  # seconds
+    INTERVAL = 180  # 3 mins in seconds
```
KPrasch: What are the factors you're considering here, is there an upper bound that will lead to poor service delivery performance?

derekpierre: The goal is to strike a balance between polling frequently enough to ensure DKG and signing rituals complete in a reasonable timeframe, while staying within commonly available free-tier usage limits.

Remaining within free-tier limits maintains the ease and low cost requirements of running a node.

At the same time, adequate responsiveness prevents rituals from take a long time. While the current workflow initiates rituals well ahead of actual usage, and so longer completion times are tolerable, we still want to avoid excessive delays, since this dynamic can change in the future and also slows down testing.

theref (approved):

vzotova (approved):

PR #3693, opened 2026-01-29
Allow additional chain endpoints beyond mandatory

theref:
Previously, operators could only configure RPC endpoints for exactly the mandatory chains (ETH and Polygon). This prevented configuring custom endpoints for additional chains like Base, forcing reliance on potentially rate-limited public fallback endpoints.

Change validation from exact equality to subset check - mandatory chains must still be configured, but additional chains are now allowed.

### DIFF
--- a/newsfragments/3693.feature.rst
+++ b/newsfragments/3693.feature.rst
@@ -1,0 +1,1 @@
+Allow operators to configure custom RPC endpoints for additional chains beyond mandatory ETH and Polygon chains.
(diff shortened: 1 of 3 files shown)


derekpierre (approved):
🎸 . If there is a quick test to be added that would be great. But if not, speed of deployment wins.

KPrasch (approved):

cygnusv (approved):

PR #3695, opened 2026-01-30
feat: Add caching layer to reduce RPC traffic during signing requests

theref:
## Problem

During heavy signing request use, Ursula nodes generate excessive RPC traffic to blockchain endpoints, causing timeouts for users. Each signing request triggers multiple RPC calls for cohort data.

## Solution

Implements a two-layer caching strategy:

### Layer 1: Cohort Data Cache (`Operator`)
- Caches `SigningCohort` objects with 60s TTL (configurable via `NUCYPHER_COHORT_CACHE_TTL`)
- Cache TTL is capped to cohort's remaining lifetime to prevent serving expired cohorts
- When cache is warm, signing requests make 0 RPC calls for cohort data
- On cache miss, fetches fresh data from chain and caches it

**Why this is safe:** Cohort data only changes via on-chain transactions (`SigningCohortDeployed`). These events trigger immediate cache invalidation (Layer 2), so stale data is never served after an update. The TTL is a fallback, not the primary invalidation mechanism.

**Note:** Conditions are intentionally NOT cached. They are always fetched directly from the contract via `get_signing_cohort_conditions()` to ensure we never sign against stale conditions.

### Layer 2: Event-Based Cache Invalidation (`SigningRitualTracker`)
- Listens for `SigningCohortDeployed` events
- Invalidates cohort cache when on-chain cohort data changes
- TTL provides fallback if events are missed

**Why this is safe:** The tracker already scans for cohort events periodically. Adding cache invalidation to the event handler ensures cached data is invalidated the moment we learn of on-chain changes.

## Configuration

| Variable | Default | Description |
|----------|---------|-------------|
| `NUCYPHER_COHORT_CACHE_TTL` | 60 | Seconds to cache cohort data |


nucypher/blockchain/eth/agents.py
```diff
@@ -997,6 +998,20 @@ def get_ritual_public_key(self, ritual_id: int) -> Optional[DkgPublicKey]:
 class SigningCoordinatorAgent(EthereumContractAgent):
     contract_name: str = "SigningCoordinator"
 
+    # Environment variable for cohort cache TTL (default 60 seconds)
+    _COHORT_CACHE_TTL_ENV_VAR = "NUCYPHER_COHORT_CACHE_TTL"
```
derekpierre: It's helpful to keep these env vars in a common location. See https://github.com/nucypher/nucypher/blob/main/nucypher/config/constants.py#L11

nucypher/blockchain/eth/agents.py
```diff
@@ -997,6 +998,20 @@ def get_ritual_public_key(self, ritual_id: int) -> Optional[DkgPublicKey]:
...
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Initialize cohort cache with configurable TTL
+        cache_ttl = int(
+            os.environ.get(
+                self._COHORT_CACHE_TTL_ENV_VAR, self._DEFAULT_COHORT_CACHE_TTL
+            )
+        )
+        self._cohort_cache = TTLCache(ttl=cache_ttl)
```
derekpierre: You could save the extra constant.

```suggestion
    _COHORT_CACHE_TTL = int(
        os.environ.get(self._COHORT_CACHE_TTL_ENV_VAR, default=60)
    )

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self._cohort_cache = TTLCache(ttl=self._COHORT_CACHE_TTL)
```

nucypher/blockchain/eth/agents.py
```diff
@@ -1052,6 +1105,13 @@ def get_signing_cohort(
             )
         return signing_cohort
 
+    def invalidate_cohort_cache(self, cohort_id: int) -> None:
+        """Remove a cohort from the cache, forcing next access to fetch from chain."""
+        self._cohort_cache.remove(cohort_id)
+        self._cohort_cache.remove(f"is_cohort_active:{cohort_id}")
+        # Note: is_signer entries are not invalidated here as they use
+        # provider-specific keys. They will expire naturally via TTL.
```
derekpierre: More generally, the agent/something needs to periodically call `purge_expired()` on the TTL cache or else you will have a memory leak over time i.e. this kind of lingering entry (provider address key) or data about short lived DKGs e.g. heartbeat rounds will remain stored in memory until purge is called.

nucypher/policy/conditions/utils.py
```diff
@@ -120,10 +122,20 @@ def _convert_any_decimals_to_floats(
 
 
 class ConditionProviderManager:
+    # Environment variable for block cache TTL (default 2 seconds)
+    _BLOCK_CACHE_TTL_ENV_VAR = "NUCYPHER_CONDITION_BLOCK_CACHE_TTL"
+    _DEFAULT_BLOCK_CACHE_TTL = 2
```
derekpierre: Same comment as in agents.py.

nucypher/policy/conditions/utils.py
```diff
@@ -120,10 +122,20 @@ def _convert_any_decimals_to_floats(
...
+
     def __init__(self, providers: Dict[int, List[HTTPProvider]]):
         self.providers = providers
         self.logger = Logger(__name__)
 
+        # Initialize block cache with configurable TTL
+        cache_ttl = int(
+            os.environ.get(self._BLOCK_CACHE_TTL_ENV_VAR, self._DEFAULT_BLOCK_CACHE_TTL)
+        )
+        self._block_cache = TTLCache(ttl=cache_ttl)
```
derekpierre: This cache also needs `purge_expired()` to be called every so often to prevent a memory leak.

derekpierre (changes requested):
The TTL Cache is useful but needs to be purged every so often or else it leads to memory leaks. Entries are not proactively removed unless either:
- request for entry is made and the entry is expired
OR
- `purge_expired` is explicitly called

nucypher/blockchain/eth/agents.py
```diff
@@ -1006,14 +1016,24 @@ def get_signing_cohort_status(self, cohort_id: int) -> int:
         result = self.contract.functions.getSigningCohortState(cohort_id).call()
         return result
 
-    @contract_api(CONTRACT_CALL)
```
derekpierre, line 1009: Any reason this decorator and other usages got removed?

nucypher/blockchain/eth/agents.py
```diff
@@ -1024,11 +1044,40 @@ def get_signer(
...
+        if cached is not None:
+            return cached
+
+        # Cache miss - fetch from chain
+        signing_cohort = self._fetch_signing_cohort_from_chain(cohort_id)
+
+        # Only cache if cohort is active and has conditions configured
+        # Check that at least one chain has non-empty condition bytes
+        cohort_status = self.get_signing_cohort_status(cohort_id)
+        has_conditions = any(signing_cohort.conditions.values())
```
derekpierre: - Minor optimization; perhaps check conditions first and then if `has_conditions` then check status. Conditions are already pulled down as part of the `fetch_cohort` whereas, `get_signing_cohort_status()` is a contract call.
- Any reason we call `get_signing_cohort_status()` instead of calling `is_cohort_active()` which already has additional caching?

nucypher/blockchain/eth/trackers/signing.py
```diff
@@ -146,3 +151,33 @@ def _get_latest_participation_state_values(
...
+
+        # Delegate to parent handler for normal event processing
+        return super()._handle_event(event, get_block_when)
+
+    def scan(self):
+        """
+        Scan for events and purge expired cache entries to prevent memory leaks.
+        """
+        # Purge expired entries from the cohort cache on each scan cycle
+        self.signing_coordinator_agent.purge_expired_cache_entries()
```
derekpierre: These are a bit of an awkward responsibility for the tracker itself. Going to think on this a little bit more.

nucypher/blockchain/eth/trackers/signing.py
```diff
@@ -146,3 +151,33 @@ def _get_latest_participation_state_values(
...
+        Handle an event, invalidating cohort cache if needed before
+        delegating to parent handler.
+        """
+        # Invalidate cohort cache for events that modify cohort data
+        if event.event in self._CACHE_INVALIDATION_EVENTS:
+            cohort_id = event.args.cohortId
+            self.signing_coordinator_agent.invalidate_cohort_cache(cohort_id)
+            self.log.debug(
+                f"Invalidated cohort cache for cohort {cohort_id} due to {event.event} event"
+            )
```
derekpierre: You could instead have a "action" that is called on the operator when either `SigningCohortDeployed` or  `SigningCohortConditionsSet` occurs, and then that does this functionality.

events match to actions see L37.

wdyt?

nucypher/blockchain/eth/trackers/signing.py
derekpierre: At least for now, it feels like the TTL cache should really be on the `Operator` instead of the `Agent`, and then the `Operator` will be able to more easily manage the cache.

Any thoughts there?

nucypher/policy/conditions/time.py
```diff
@@ -46,10 +46,9 @@ def __init__(
     ):
         super().__init__(chain=chain, method=method, parameters=parameters)
 
-    def _execute(self, w3: Web3, resolved_parameters: List[Any]) -> Any:
-        """Execute onchain read and return result."""
-        # TODO may need to rethink as part of #3051 (multicall work).
-        latest_block = w3.eth.get_block("latest")
+    def execute(self, providers: ConditionProviderManager, **context) -> Any:
+        """Execute time condition using cached latest block."""
+        latest_block = providers.get_latest_block(self.chain)
```
derekpierre: Given the convo on Discord and the granularity of how often this value changes, should we leave this optimization out for now? We can investigate more of it as a follow-up.

derekpierre (dismissed):

nucypher/blockchain/eth/trackers/signing.py
theref: yeah i'm happy with that, I wasn't sure tbh

nucypher/blockchain/eth/agents.py
theref: eugh, just refactoring

nucypher/blockchain/eth/agents.py
theref: yes to both 🚀

nucypher/blockchain/eth/trackers/signing.py
```diff
@@ -35,6 +39,7 @@ def __init__(
         events = [
             contract.events.InitiateSigningCohort,
             contract.events.SigningCohortDeployed,
+            contract.events.SigningCohortConditionsSet,
```
derekpierre: David and I had a discussion about it and since we would be always getting conditions directly from the contract, I don't believe we need cache invalidation for the condition set event. That was something that I had initially but then removed once the condition was obtained from the contract and never cached - see https://github.com/nucypher/nucypher/pull/3695/commits/5caa79ebca2595421707929ba85544016a19caa6

derekpierre: The main issue with caching the condition itself is that any staleness is unacceptable since an outdated condition that allows a transaction to be signed would be catastrophic. Although events are polled frequently (every 4 minutes in #3690), conditions can change at any time, which makes cache invalidation inherently unreliable for this purpose.

Other properties of the cohort may also change, but they do so even less frequently and with far lower risk. For example, if a signer is no longer part of the cohort but still produces a signature, the multisig will already recognize that the signer is invalid and reject the signature. In the worst case, the transaction fails temporarily which is much more preferable to a transaction being incorrectly accepted.

nucypher/blockchain/eth/trackers/signing.py
```diff
@@ -23,6 +24,9 @@ def __init__(
             super().__init__(participating)
             self.already_posted_signature = already_posted_signature
 
+    # Events that should trigger cohort cache invalidation
+    _CACHE_INVALIDATION_EVENTS = {"SigningCohortDeployed", "SigningCohortConditionsSet"}
```
derekpierre: What's the need for checking `SigningCohortDeployed`? We also only cache cohorts that are already active.

nucypher/utilities/cache.py
```diff
@@ -109,6 +109,16 @@ def pop(self, key, default=None):
...
 
+    def add_with_ttl(self, key, value, ttl):
+        """
+        Add the provided key entry to be the provided value with a custom TTL.
+        """
+        if key is None or value is None:
+            raise ValueError(f"Invalid key-value pair ({key}, {value})")
+
+        with self.__cache_lock:
+            self.__cache[key] = self.TTLEntry(value=value, ttl=ttl)
```
cygnusv, line 118: Can we redefine now `__setitem__` method above to use this method internally? Just to avoid repeated code.

nucypher/blockchain/eth/trackers/signing.py
derekpierre: Let me know if you disagree - https://github.com/nucypher/nucypher/pull/3695/commits/8fdfc26b84065012970cd55a4cc59a9ce6747947

nucypher/utilities/cache.py
```diff
@@ -109,6 +109,16 @@ def pop(self, key, default=None):
 
             return value
 
+    def add_with_ttl(self, key, value, ttl):
+        """
+        Add the provided key entry to be the provided value with a custom TTL.
+        """
+        if key is None or value is None:
+            raise ValueError(f"Invalid key-value pair ({key}, {value})")
```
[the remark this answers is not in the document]
derekpierre: ✔️

nucypher/blockchain/eth/actors.py
```diff
@@ -1455,34 +1460,48 @@ def _produce_decryption_share_for_request(
...
+
+        signing_cohort = self.signing_coordinator_agent.get_signing_cohort(cohort_id)
+        # safety measure that cohort doesn't expire before caching TTL
+        time_remaining = signing_cohort.end_timestamp - maya.now().epoch
+        custom_ttl = min(time_remaining, self._signing_cohort_cache.ttl)
+        self._signing_cohort_cache.add_with_ttl(cohort_id, signing_cohort, custom_ttl)
+        return signing_cohort
+
+    def clear_signing_cohort_cache(self, cohort_id: int, **kwargs) -> None:
+        self._signing_cohort_cache.remove(cohort_id)
```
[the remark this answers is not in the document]
derekpierre: Removed function for now since we don't need it.

nucypher/blockchain/eth/actors.py
```diff
@@ -1455,34 +1460,46 @@ def _produce_decryption_share_for_request(
...
+            return cached_cohort
+
+        if not self.signing_coordinator_agent.is_cohort_active(cohort_id):
+            raise self.UnauthorizedRequest(
+                f"Cohort #{cohort_id} is not active",
+            )
+
+        signing_cohort = self.signing_coordinator_agent.get_signing_cohort(cohort_id)
+        # safety measure that cohort doesn't expire before caching TTL
+        time_remaining = signing_cohort.end_timestamp - maya.now().epoch
```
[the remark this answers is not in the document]
derekpierre: ✔️  - Highly unlikely but added as a safety check.

derekpierre (approved):
🎸 - nice start with caching.

I contributed some code to this PR, so my approval is implicit.

nucypher/blockchain/eth/actors.py
```diff
@@ -1455,34 +1460,52 @@ def _produce_decryption_share_for_request(
...
-            raise self.UnauthorizedRequest(
-                f"Cohort #{signing_request.cohort_id} is not active",
-            )
+        signing_cohort = self._get_signing_cohort(signing_request.cohort_id)
 
-        if not self.signing_coordinator_agent.is_signer(
-            cohort_id=signing_request.cohort_id,
-            provider_address=self.staking_provider_address,
+        if not any(
+            s.provider == self.staking_provider_address for s in signing_cohort.signers
```
cygnusv, line 1498: Perhaps more idiomatic:

`if not self.staking_provider_address in (s.provider for s in signing_cohort.signers):`

Although current approach will short-circuit, so feel free to discard my comment if you prefer.

cygnusv (approved):

nucypher/blockchain/eth/actors.py
derekpierre: The current short circuit was the goal, as a more efficient way of only iterating through the list once.

nucypher/utilities/cache.py
theref: done

PR #3700, opened 2026-02-02
Remove unused end_block_timestamp from scan_chunk

theref:
The `end_block_timestamp` was fetched via `get_block` RPC call on every chunk scan but never used by callers. This removes the unnecessary RPC call, reducing idle network traffic to providers like Infura.

**Type of PR:**
- Bugfix

**Required reviews:** 
- 2


derekpierre (approved):
🎸  - nice find!

vzotova (approved):

KPrasch (approved):

PR #3701, opened 2026-02-03
Expose Docker build metadata as envvars to Ursula container

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Have you ever wanted to programatically get the version that's running on a node? I don't mean the `nucypher --version`, but the actual commit used. Look no more, this PR is for you (I'm looking at you @theref).

This PR adds the last git commit and known git reference name as part of docker build arguments, which are in turn exposed as environment variables to the container. This allows for Ursula to consume them, and for instance, render them in the status page.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


theref (approved):

.github/workflows/docker.yml
```diff
@@ -44,6 +44,9 @@ jobs:
         file: ./deploy/docker/Dockerfile
         push: true
         tags: nucypher/nucypher:lynx
+        build-args: |
+          GIT_COMMIT=${{ github.sha }}
+          GIT_REF_NAME=${{ github.ref_name }}
```
derekpierre, line 49: Is there much added value in including `ref_name`?

(https://docs.github.com/en/actions/reference/workflows-and-actions/variables)
GITHUB_REF_NAME | The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, feature-branch-1.For pull requests that were not merged, the format is <pr_number>/merge.
-- | --

This CI job only runs when:
1. A PR is merged into a `v*.*.*` EPIC branch eg. `v7.7.x`
OR
2. A `v*.*.*` tag is created from `main` for a release eg. `v7.6.1`

In a practical sense, the `ref_name` will typically be one of two values: the EPIC branch OR the release tag, and therefore not that informative since the existing version number will already separate the two situations i.e. the version number in the EPIC branch is already ahead of the latest stable release.

Basically:
- Version already distinguishes track (EPIC vs release) because the EPIC's version is ahead by definition
- SHA distinguishes the exact build within the EPIC track.

Let me know if I'm missing some other advantage.

.github/workflows/docker.yml
derekpierre, line 1: Should we remove the logic on L34-37 if we are going to expose the github sha separately in status?

.github/workflows/docker.yml
cygnusv: Yeah, good point. Actually, I was interested in this: "_For pull requests that were not merged, the format is <pr_number>/merge._". But if this job only runs for merged PRs or tags on main, then I agree the refname is redundant. I'll update the PR to remove it.

derekpierre: If our build process was different i.e. building from open PRs, the `<pr_number>/merge` would be interesting. However, that approach would result in the `lynx` image being repeatedly overwritten by different concurrent open PR builds, leading to inconsistent behaviour and functionality. 

Using the EPIC branch as the common build source gives us a stable, shared reference point for `lynx` docker images.

cygnusv:
Closing, let's continue in #3705

PR #3702, opened 2026-02-04
fix: obfuscate RPC endpoint URLs in logs to prevent API key exposure

theref:
## Summary
Obfuscates sensitive API keys in RPC endpoint URLs when logging to prevent accidental exposure of credentials.

## Changes
- Add `obfuscate_rpc_url()` function that masks path segments 16+ characters (likely API keys)
- Update all 11 log statements in `utils.py` to use obfuscation
- Add parametrized tests for URL obfuscation

## Example
**Before:** `RPC endpoint https://base-sepolia.infura.io/v3/aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa is healthy`

**After:** `RPC endpoint https://base-sepolia.infura.io/v3/aaa*** is healthy`

## Testing
Added parametrized tests covering:
- Infura-style URLs with API keys
- Alchemy-style URLs with API keys  
- URLs without API keys (unchanged)
- Short path segments (preserved)

### DIFF
--- a/nucypher/blockchain/eth/utils.py
+++ b/nucypher/blockchain/eth/utils.py
@@ -2,6 +2,7 @@ import time
 from decimal import Decimal
 from functools import cache
 from typing import Any, Dict, List, Optional, Union
+from urllib.parse import urlparse, urlunparse
 
 import requests
 from eth_typing import ChecksumAddress
@@ -24,8 +25,6 @@ def obfuscate_rpc_url(url: str) -> str:
     Example: https://mainnet.infura.io/v3/abc123 -> https://mainnet.infura.io/v3/abc***
     """
     try:
-        from urllib.parse import urlparse, urlunparse
-
         parsed = urlparse(url)
         if parsed.path:
             # Split path into segments and obfuscate segments that look like API keys
--- a/nucypher/blockchain/eth/utils.py
+++ b/nucypher/blockchain/eth/utils.py
@@ -17,7 +17,35 @@ from nucypher.utilities.logging import Logger
 LOGGER = Logger("utility")
 
 
-def prettify_eth_amount(amount, original_denomination: str = 'wei') -> str:
+def obfuscate_rpc_url(url: str) -> str:
+    """
+    Obfuscates sensitive parts of an RPC URL for safe logging.
+    Replaces API keys and other path segments after the host with asterisks.
+    Example: https://mainnet.infura.io/v3/abc123 -> https://mainnet.infura.io/v3/abc***
+    """
+    try:
+        from urllib.parse import urlparse, urlunparse
+
+        parsed = urlparse(url)
+        if parsed.path:
+            # Split path into segments and obfuscate segments that look like API keys
+            segments = parsed.path.split("/")
+            obfuscated_segments = []
+            for segment in segments:
+                # Obfuscate segments that are 16+ chars (likely API keys/secrets)
+                if len(segment) >= 16:
+                    obfuscated_segments.append(segment[:3] + "***")
+                else:
+                    obfuscated_segments.append(segment)
+            obfuscated_path = "/".join(obfuscated_segments)
+            parsed = parsed._replace(path=obfuscated_path)
+        return urlunparse(parsed)
+    except Exception:
+        # If parsing fails, return a safe placeholder
+        return "<RPC endpoint>"
+
+
+def prettify_eth_amount(amount, original_denomination: str = "wei") -> str:
     """
     Converts any ether `amount` in `original_denomination` and finds a suitable representation based on its length.
     The options in consideration are representing the amount in wei, gwei or ETH.
@@ -29,29 +57,33 @@ def prettify_eth_amount(amount, original_denomination: str = 'wei') -> str:
         # First obtain canonical representation in wei. Works for int, float, Decimal and str amounts
         amount_in_wei = Web3.to_wei(Decimal(amount), original_denomination)
 
-        common_denominations = ('wei', 'gwei', 'ether')
+        common_denominations = ("wei", "gwei", "ether")
 
         options = [str(Web3.from_wei(amount_in_wei, d)) for d in common_denominations]
 
         best_option = min(zip(map(len, options), options, common_denominations))
         _length, pretty_amount, denomination = best_option
 
-        if denomination == 'ether':
-            denomination = 'ETH'
+        if denomination == "ether":
+            denomination = "ETH"
         pretty_amount += " " + denomination
 
-    except Exception:  # Worst case scenario, we just print the str representation of amount
+    except (
+        Exception
+    ):  # Worst case scenario, we just print the str representation of amount
         pretty_amount = str(amount)
 
     return pretty_amount
 
 
-def get_transaction_name(contract_function: Union[ContractFunction, ContractConstructor]) -> str:
+def get_transaction_name(
+    contract_function: Union[ContractFunction, ContractConstructor],
+) -> str:
     deployment = isinstance(contract_function, ContractConstructor)
     try:
         transaction_name = contract_function.fn_name.upper()
     except AttributeError:
-        transaction_name = 'DEPLOY' if deployment else 'UNKNOWN'
+        transaction_name = "DEPLOY" if deployment else "UNKNOWN"
     return transaction_name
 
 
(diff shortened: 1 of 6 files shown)


nucypher/blockchain/eth/utils.py
```diff
@@ -17,7 +17,35 @@
...
+
+        parsed = urlparse(url)
+        if parsed.path:
+            # Split path into segments and obfuscate segments that look like API keys
+            segments = parsed.path.split("/")
+            obfuscated_segments = []
+            for segment in segments:
+                # Obfuscate segments that are 16+ chars (likely API keys/secrets)
+                if len(segment) >= 16:
+                    obfuscated_segments.append(segment[:3] + "***")
```
cygnusv, line 46: Can't help seeing a cat face there 🐱

nucypher/blockchain/eth/utils.py
```diff
@@ -181,27 +215,35 @@ def _get_json_rpc_call_result(endpoint: str, query: dict) -> Optional[Any]:
...
-        LOGGER.debug(f"RPC endpoint {endpoint} is unhealthy: network error")
+        LOGGER.debug(
+            f"RPC endpoint {obfuscate_rpc_url(endpoint)} is unhealthy: network error"
+        )
         return None
 
     if response.status_code != 200:
         LOGGER.debug(
-            f"RPC endpoint {endpoint} is unhealthy: {response.status_code} | {response.text}"
+            f"RPC endpoint {obfuscate_rpc_url(endpoint)} is unhealthy: {response.status_code} | {response.text}"
```
cygnusv: As part of the review, I noticed that one of my log lines for this type of error (specifically, `https://endpoints.omniatech.io/v1/op/sepolia/public`) prints a huge blob of data returned in the `response.text`. What do you think if we trim the `response.text` to a reasonable max size?

image

nucypher/blockchain/eth/utils.py
```diff
@@ -181,27 +215,35 @@ def _get_json_rpc_call_result(endpoint: str, query: dict) -> Optional[Any]:
...
         if "result" not in data:
-            LOGGER.debug(f"RPC endpoint {endpoint} is unhealthy: no response data")
+            LOGGER.debug(
+                f"RPC endpoint {obfuscate_rpc_url(endpoint)} is unhealthy: no response data"
+            )
             return None
     except requests.exceptions.JSONDecodeError:
-        LOGGER.debug(f"RPC endpoint {endpoint} is unhealthy: {response.text}")
+        LOGGER.debug(
+            f"RPC endpoint {obfuscate_rpc_url(endpoint)} is unhealthy: {response.text}"
```
cygnusv: Potentially same comment as above

cygnusv (approved):

tests/unit/test_rpc_healthcheck.py
```diff
@@ -118,3 +120,27 @@ def test_get_healthy_default_rpc_endpoints(mocker):
...
+            "https://eth-mainnet.alchemyapi.io/v2/abc***",
+        ),
+        # URLs without API keys should remain unchanged
+        ("https://rpc.ankr.com/eth", "https://rpc.ankr.com/eth"),
+        ("https://cloudflare-eth.com", "https://cloudflare-eth.com"),
+        # Short path segments preserved
+        ("https://example.com/v3/short", "https://example.com/v3/short"),
+    ],
+)
+def test_obfuscate_rpc_url(url, expected):
```
derekpierre: Nice test! I checked with drpc endpoints and those get properly obfuscated as well.

wdyt about putting this test in a separate test file (since there doesn't already seem to be one for `blockchain.eth.utils` eg. `test_blockchain_utils.py`? Just because this is a standalone utility function which is used by rpc healthcheck instead of "part" of the rpc healthcheck.

nucypher/blockchain/eth/utils.py
```diff
@@ -17,7 +17,35 @@
...
 
-def prettify_eth_amount(amount, original_denomination: str = 'wei') -> str:
+def obfuscate_rpc_url(url: str) -> str:
+    """
+    Obfuscates sensitive parts of an RPC URL for safe logging.
+    Replaces API keys and other path segments after the host with asterisks.
+    Example: https://mainnet.infura.io/v3/abc123 -> https://mainnet.infura.io/v3/abc***
+    """
+    try:
+        from urllib.parse import urlparse, urlunparse
```
derekpierre: Any reason this import isn't at the top of the file instead of in the function itself?

nucypher/blockchain/eth/utils.py
```diff
@@ -17,7 +17,35 @@
 LOGGER = Logger("utility")
 
 
-def prettify_eth_amount(amount, original_denomination: str = 'wei') -> str:
+def obfuscate_rpc_url(url: str) -> str:
```
derekpierre, line 31: Some other spots that could use this utility 😅 :
- https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/interfaces.py#L315
- https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/interfaces.py#L322
- https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/interfaces.py#L345
- https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/interfaces.py#L349
- https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/interfaces.py#L259

- (Maybe some duplicated log messages as well, so you can determine whether the extra logging is still needed based on logging already done by health check)
  - https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/actors.py#L364
  - https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/actors.py#L371
  - https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/actors.py#L376
  - https://github.com/nucypher/nucypher/blob/v7.7.x/nucypher/blockchain/eth/actors.py#L388 
  
  
Not comprehensive, but you can search around for "endpoint".

tests/unit/test_rpc_healthcheck.py
```diff
@@ -90,6 +90,7 @@ def test_get_default_rpc_endpoints(mocker):
 
     # Mock a failed response
     mock_get.return_value.status_code = 500
+    mock_get.return_value.text = "Internal Server Error"
```
derekpierre, line 93: Curious: Was this specifically needed for a reason; was a test was failing?  I guess you actually needed a string now.

tests/unit/test_blockchain_utils.py
```diff
@@ -25,3 +28,20 @@
...
+        # Exactly 200 chars unchanged
+        ("x" * 200, "x" * 200),
+        # Long text truncated with ellipsis
+        ("x" * 250, "x" * 200 + "..."),
+        # Empty string unchanged
+        ("", ""),
+    ],
+)
+def test_truncate_response_text(text, expected):
+    assert _truncate_response_text(text) == expected
```
derekpierre, line 49: 🚀

tests/unit/test_blockchain_utils.py
```diff
@@ -0,0 +1,47 @@
...
+        (
+            "https://eth-mainnet.alchemyapi.io/v2/abcdef1234567890abcdef1234567890",
+            "https://eth-mainnet.alchemyapi.io/v2/abc***",
+        ),
+        # URLs without API keys should remain unchanged
+        ("https://rpc.ankr.com/eth", "https://rpc.ankr.com/eth"),
+        ("https://cloudflare-eth.com", "https://cloudflare-eth.com"),
+        # Short path segments preserved
+        ("https://example.com/v3/short", "https://example.com/v3/short"),
+    ],
```
derekpierre, line 29: Nitpick based on code coverage
```suggestion
        ("https://example.com/v3/short", "https://example.com/v3/short"),
        # Non-urls are handled
        ("not a url", "<RPC endpoint>"),
    ],
```

tests/unit/test_rpc_healthcheck.py
theref: Yes - the new `_truncate_response_text()` function calls `len()` on the response text, which fails on an unmocked Mock object.

tests/unit/test_blockchain_utils.py
theref: Added a test for the exception path, but used `123` instead - `urlparse` doesn't throw on invalid strings like `'not a url'`, it just parses what it can. Non-string types trigger the exception.

derekpierre (approved):
🎸

PR #3703, opened 2026-02-04
feat: add create2 operation for local address computation

### LINKED ISSUE #3698: Potential use of local computation to calculate equivalent `Create2.computeAddress` smart contract function

theref:
Closes #3698 

## Summary

Add a new `create2` operation to the `VariableOperation` system that computes Ethereum CREATE2 addresses locally, eliminating RPC calls to factory contracts.

## Motivation

Currently, computing counterfactual AA (Account Abstraction) addresses requires an RPC call to a factory contract's `computeAddress` method. This:
- Consumes RPC endpoint quota (Infura, etc.)
- Adds latency to condition evaluation
- Requires chain connectivity for a purely deterministic computation

The CREATE2 address formula is deterministic and can be computed locally.

## How CREATE2 Works

CREATE2 computes deterministic addresses using:
```
address = keccak256(0xff ++ deployer_address ++ salt ++ bytecode_hash)[12:]
```

## Usage

```json
{
  "operation": "create2",
  "value": {
    "deployerAddress": "0x69Aa2f9fe1572F1B640E1bbc512f5c3a734fc77c",
    "bytecodeHash": "0x210ffc0da7f274285c4d6116aaef8420ecb9054faced33862197d6b951cb32f5"
  }
}
```

Both `deployerAddress` and `bytecodeHash` support context variables.

## Example: Full Pipeline

```json
{
  "varName": "validateSender",
  "condition": {
    "conditionType": "context-variable",
    "contextVariable": ":senderId",
    "returnValueTest": {
      "comparator": "==",
      "value": ":userOpSender"
    }
  },
  "operations": [
    { "operation": "str" },
    { "operation": "+=", "value": "|Discord|Collab.Land" },
    { "operation": "keccak" },
    { "operation": "create2", "value": { "deployerAddress": "0x69Aa...", "bytecodeHash": "0x210f..." } }
  ]
}
```

## Testing

- Added comprehensive unit tests for the new operation
- All 207 tests in `test_variable_operation.py` pass
- Tests cover: valid computation, invalid inputs (salt/deployer/bytecode length), missing fields, cascading operations, and context variable resolution

### DIFF
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -10,7 +10,7 @@ from hashlib import md5
 from inspect import signature
 from typing import Any, List, Optional, Tuple, Type, Union
 
-from eth_utils import is_hexstr, keccak
+from eth_utils import is_hexstr, keccak, to_checksum_address
 from hexbytes import HexBytes
 from marshmallow import (
     Schema,
@@ -367,6 +367,58 @@ def _to_hex(value):
         raise TypeError(f"Invalid value for hex conversion: {e}")
 
 
+def _compute_create2_address(salt: bytes, value: dict) -> str:
+    """
+    Compute CREATE2 address locally.
+
+    Formula: keccak256(0xff ++ deployer ++ salt ++ bytecode_hash)[12:]
+
+    :param salt: The salt value (must be 32 bytes)
+    :param value: Dict containing 'deployerAddress' and 'bytecodeHash'
+    :return: Checksummed Ethereum address
+    :raises TypeError: If inputs are invalid
+    """
+    try:
+        deployer_address = value["deployerAddress"]
+        bytecode_hash = value["bytecodeHash"]
+    except (KeyError, TypeError) as e:
+        raise TypeError(
+            f"create2 operation requires 'deployerAddress' and 'bytecodeHash' in value: {e}"
+        )
+
+    # Validate and convert deployer address
+    try:
+        deployer_bytes = HexBytes(deployer_address)
+    except Exception as e:
+        raise TypeError(f"Invalid deployerAddress: {e}")
+    if len(deployer_bytes) != 20:
+        raise TypeError(f"deployerAddress must be 20 bytes, got {len(deployer_bytes)}")
+
+    # Validate and convert salt
+    try:
+        salt_bytes = HexBytes(salt)
+    except Exception as e:
+        raise TypeError(f"Invalid salt: {e}")
+    if len(salt_bytes) != 32:
+        raise TypeError(f"salt must be 32 bytes, got {len(salt_bytes)}")
+
+    # Validate and convert bytecode hash
+    try:
+        bytecode_hash_bytes = HexBytes(bytecode_hash)
+    except Exception as e:
+        raise TypeError(f"Invalid bytecodeHash: {e}")
+    if len(bytecode_hash_bytes) != 32:
+        raise TypeError(
+            f"bytecodeHash must be 32 bytes, got {len(bytecode_hash_bytes)}"
+        )
+
+    # CREATE2: keccak256(0xff ++ deployer ++ salt ++ bytecode_hash)[12:]
+    pre_image = b"\xff" + deployer_bytes + salt_bytes + bytecode_hash_bytes
+    address_bytes = keccak(pre_image)[12:]
+
+    return to_checksum_address(address_bytes)
+
+
 # should raise TypeError for invalid inputs
 _OPERATOR_FUNCTIONS = {
     # We can add all kinds of operators over time, this is just a base start - given that
@@ -405,6 +457,8 @@ _OPERATOR_FUNCTIONS = {
     "toHex": _to_hex,
     # hashing
     "keccak": lambda a: keccak(a.encode() if isinstance(a, str) else a),
+    # address computation
+    "create2": _compute_create2_address,
 }
 
 MAX_VARIABLE_OPERATIONS = 5
@@ -1133,7 +1187,7 @@ class ConditionLingo(_Serializable):
             raise InvalidConditionLingo(f"Invalid condition grammar: {e}")
 
     @classmethod
-    def from_json(cls, data: str) -> 'ConditionLingo':
+    def from_json(cls, data: str) -> "ConditionLingo":
         try:
             return super().from_json(data)
         except ValidationError as e:
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -383,7 +383,7 @@ def _compute_create2_address(salt: bytes, value: dict) -> str:
         bytecode_hash = value["bytecodeHash"]
     except (KeyError, TypeError) as e:
         raise TypeError(
-            f"create2 operation requires 'deployerAddress' and 'bytecodeHash' in value: {e}"
+            f"create2 operation requires dictionary with 'deployerAddress' and 'bytecodeHash' values: {e}"
         )
 
     # Validate and convert deployer address
--- a/tests/acceptance/conditions/test_create2.py
+++ b/tests/acceptance/conditions/test_create2.py
@@ -1,0 +1,99 @@
+import os
+
+import pytest
+
+from nucypher.policy.conditions.lingo import VariableOperation
+
+
+@pytest.fixture
+def create2_factory(project, deployer_account):
+    """Deploy the Create2Factory contract for testing."""
+    return deployer_account.deploy(project.Create2Factory)
+
+
+def test_create2_matches_contract_computation(create2_factory):
+    """
+    Test that our local create2 address computation matches the
+    OpenZeppelin Create2.computeAddress implementation on-chain.
+    """
+    # Use random test values
+    salt = os.urandom(32)
+    bytecode_hash = os.urandom(32)
+    deployer_address = create2_factory.address
+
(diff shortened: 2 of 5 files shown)


nucypher/policy/conditions/lingo.py
```diff
@@ -367,6 +367,58 @@ def _to_hex(value):
         raise TypeError(f"Invalid value for hex conversion: {e}")
 
 
+def _compute_create2_address(salt: bytes, value: dict) -> str:
```
cygnusv, line 370: I assume that the choice of a dictionary comes from the limitations on how lingo operators work (particular about how parameters are fed). Can you just use `value` as a single input, which includes `salt`? There are other unary operators in lingo, so from that perspective, this is just another one.

tests/unit/conditions/test_variable_operation.py
```diff
@@ -564,3 +576,219 @@ def test_to_token_base_units_type_errors():
...
+        ":factoryAddress": "0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f",
+        ":initCodeHash": "0x96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f",
+    }
+
+    operations = [
+        VariableOperation(
+            operation="create2",
+            value={
+                "deployerAddress": ":factoryAddress",
+                "bytecodeHash": ":initCodeHash",
```
cygnusv, line 745: 🚀

tests/unit/conditions/test_variable_operation.py
```diff
@@ -564,3 +576,219 @@ def test_to_token_base_units_type_errors():
...
+
+    Values from discord-taco-web validate-aa-derivation.ts script.
+    """
+    # Real Discord user ID
+    discord_id = 405651072460259339
+
+    # SimpleFactory on Base Sepolia
+    deployer = "0x69Aa2f9fe1572F1B640E1bbc512f5c3a734fc77c"
+    # Bytecode hash for MetaMask Delegation Toolkit MultiSig
+    bytecode_hash = "0x210ffc0da7f274285c4d6116aaef8420ecb9054faced33862197d6b951cb32f5"
```
cygnusv, line 777: 💯

nucypher/policy/conditions/lingo.py
```diff
@@ -367,6 +367,58 @@ def _to_hex(value):
...
+    :param value: Dict containing 'deployerAddress' and 'bytecodeHash'
+    :return: Checksummed Ethereum address
+    :raises TypeError: If inputs are invalid
+    """
+    try:
+        deployer_address = value["deployerAddress"]
+        bytecode_hash = value["bytecodeHash"]
+    except (KeyError, TypeError) as e:
+        raise TypeError(
+            f"create2 operation requires 'deployerAddress' and 'bytecodeHash' in value: {e}"
```
derekpierre:
```suggestion
            f"create2 operation requires dictionary with 'deployerAddress' and 'bytecodeHash' values: {e}"
```

tests/unit/conditions/test_variable_operation.py
```diff
@@ -564,3 +576,219 @@ def test_to_token_base_units_type_errors():
...
+                "deployerAddress": deployer,
+                "bytecodeHash": bytecode_hash,
+            },
+        ),
+    ]
+
+    result = VariableOperation.evaluate_operations(operations, discord_id)
+
+    # Expected AA address for this Discord ID
+    assert result == "0x420AcFa51fdB2821dFb407e212A882144807737C"
```
derekpierre, line 797: Nice unit tests.

In addition to unit tests we should add an acceptance test that compares the result of the `create2` operation with an actual `Create2` calc on a smart contract. That way we can be sure our calculation always matches the smart contract result, instead of relying on a couple know outcomes.

It's similar to what we did with the test `EntryPoint` contract - https://github.com/nucypher/nucypher/commit/be89d895627034a8739839a4437535302284d51a#diff-9b69ba5544d2611df5c0a65c8117c01896a7f7239cd1308b79281c03a750d2df. 

You can add a really simple `Create2Factory` contract to the `acceptance/contracts` folder that calls `Create2.computeAddress` eg.
```solidity
// SPDX-License-Identifier: AGPL-3.0-or-later

pragma solidity ^0.8.0;

import "@openzeppelin/contracts/utils/Create2.sol";

contract Create2Factory {
    function computeAddress(bytes32 salt, bytes32 bytecodeHash, address deployer) internal pure returns (address addr) {
        return Create2.computeAddress(salt, bytecodeHash, deployer);
    }
}
```

Then you can add test(s) that use any random bytes for input values (instead of specific ones) and check that the local operation and the contract function always return the same address.

nucypher/policy/conditions/lingo.py
derekpierre: Interested to hear @theref 's thoughts here as well. 

My assumption is clarity in the JSON for parameter values provided to the function as opposed to requiring a specific order of unnamed parameters in a list...(?)

~~Technically variable operation functions can take any number of parameters, but so far it has been only one (unary) or two, in some cases the 2nd parameter being a list of values (eg. "avg")~~

nucypher/policy/conditions/lingo.py
theref: yeah that seems better

tests/unit/conditions/test_variable_operation.py
theref: this is my one, taken straight from the CL bot

nucypher/policy/conditions/lingo.py
theref: we definitely could do that, but I was just trying to match how other operators worked. Basically i wanted conditions to look like this:
```python
    operations = [
        VariableOperation(operation="str"),  # "405651072460259339"
        VariableOperation(
            operation="+=", value="|Discord|Collab.Land"
        ),  # "405651072460259339|Discord|Collab.Land"
        VariableOperation(operation="keccak"),  # salt as bytes32
        VariableOperation(
            operation="create2",
            value={
                "deployerAddress": deployer,
                "bytecodeHash": bytecode_hash,
            },
        ),
    ]
```
What happens if we put salt inside `value`?

derekpierre: Oh right. It's the fact that a `VariableOperation` is always performed "on something" i.e. there is an input value from a calc or prior operation (the variable) that is then operated on; in this case that's the "salt" calculation done before it.

If the salt was included in the `VariableOperation `definition itself then this wouldn't be an operation on something, it would be a self-contained operation which is more like a condition/calc rather than a variable operation.

tests/acceptance/conditions/test_create2.py
```diff
@@ -0,0 +1,99 @@
+import os
+
+import pytest
+
+from nucypher.policy.conditions.lingo import VariableOperation
+
+
+@pytest.fixture
```
derekpierre: This contract doesn't store state or anything so the fixture can be module scoped and deployed once instead of default function scope and deployed for each test.

```suggestion
@pytest.fixture(scope="module")
```

tests/acceptance/conditions/test_create2.py
```diff
@@ -0,0 +1,99 @@
...
+def test_create2_matches_contract_with_different_deployers(create2_factory, accounts):
+    """
+    Test create2 computation with various deployer addresses.
+    """
+    salt = os.urandom(32)
+    bytecode_hash = os.urandom(32)
+
+    # Test with multiple different deployer addresses
+    for account in accounts[:5]:
+        deployer_address = account.address
```
derekpierre: The following  will give you better randomness for the deployer address than the "accounts" fixture which produces the same addresses every time since we use a test seed phrase to initialize it

```suggestion
def test_create2_matches_contract_with_different_deployers(create2_factory, get_random_checksum_address):
    """
    Test create2 computation with various deployer addresses.
    """
    salt = os.urandom(32)
    bytecode_hash = os.urandom(32)

    # Test with multiple different deployer addresses
    for _ in range(5):
        deployer_address = get_random_checksum_address()
```

tests/acceptance/conditions/test_create2.py
```diff
@@ -0,0 +1,99 @@
...
+from nucypher.policy.conditions.lingo import VariableOperation
+
+
+@pytest.fixture
+def create2_factory(project, deployer_account):
+    """Deploy the Create2Factory contract for testing."""
+    return deployer_account.deploy(project.Create2Factory)
+
+
+def test_create2_matches_contract_computation(create2_factory):
```
derekpierre: Feels like this test, `test_create2_matches_contract_computation`, is not needed since the `test_create2_matches_contract_multiple_random_inputs` does the same thing but in a loop...?

tests/acceptance/conditions/test_create2.py
```diff
@@ -0,0 +1,99 @@
...
+                },
+            ),
+        ]
+        local_result = VariableOperation.evaluate_operations(operations, salt)
+
+        contract_result = create2_factory.computeAddress(
+            salt, bytecode_hash, deployer_address
+        )
+
+        assert local_result == contract_result
```
derekpierre, line 71: 🔥

derekpierre (approved):
🎸 .

We'll need to do a `taco-web` addition to support this variable operation. Let's make sure there aren't any hurdles there before merging this PR.

cygnusv (approved):

PR #3705, opened 2026-02-09
Build metadata in version string in Docker builds

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
This PR fixes the process of adding the last git commit as part of the version (specifically, as [build metadata](https://semver.org/#spec-item-10) following SemVer spec) during the docker specific workflow on CI. Specifically, `bumpversion` config is tweaked to accept SHA commits instead of only accepting numbers.

The original approach in #3701 doesn't really solve anything, because similarly to this PR, it only works when docker runs from CI. Next steps (for a subsequent PR) is to tweak the CI process to allow for custom docker builds that don't depend on tags or on merging on the current target release branch.

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**

### DIFF
--- a/nucypher/characters/banners.py
+++ b/nucypher/characters/banners.py
@@ -73,7 +73,5 @@ Yb, `88        88                                   IP'`Yb
      Y8b,____,d88,,dP     Y8,,8'_   8) ,d8b,  ,d8b,,d8b,_ ,d8,   ,d8b,
       "Y888888P"Y88P      `Y8P' "YY8P8P8P'"Y88P"`Y88P'"Y88P"Y8888P"`Y8
 
-
-the Untrusted Re-Encryption Proxy.
 {}
 '''
--- a/nucypher/cli/commands/ursula.py
+++ b/nucypher/cli/commands/ursula.py
@@ -2,6 +2,7 @@ from pathlib import Path
 
 import click
 
+import nucypher
 from nucypher.cli.actions.auth import (
     collect_mnemonic,
     get_client_password,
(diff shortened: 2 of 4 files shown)


nucypher/characters/banners.py
```diff
@@ -73,7 +73,5 @@
      Y8b,____,d88,,dP     Y8,,8'_   8) ,d8b,  ,d8b,,d8b,_ ,d8,   ,d8b,
       "Y888888P"Y88P      `Y8P' "YY8P8P8P'"Y88P"`Y88P'"Y88P"Y8888P"`Y8
 
-
-the Untrusted Re-Encryption Proxy.
```
derekpierre, line 77: 😮

nucypher/cli/commands/ursula.py
```diff
@@ -630,12 +631,18 @@ def run(
 ):
     """Run an "Ursula" node."""
 
-    emitter = setup_emitter(general_config)
+    emitter = setup_emitter(
+        general_config, character_options.config_options.operator_address
```
derekpierre: How come `operator_address` is passed as the `banner` parameter?

cygnusv: This is what other `nucypher ursula` subcommands do. If you ask me, I'd get rid of printing addresses in the banner, and even of the whole banner thing, but at the very least, just wanted to keep it consistent.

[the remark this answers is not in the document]
derekpierre: ^ This is related to my earlier question about the `banner` parameter.

derekpierre (approved):
🎸

KPrasch (approved):

Issue #3582, opened 2025-03-17
Handling Exceptions in Logical Conditions: Opt-in Exception Failure Suppression


derekpierre:
Currently, logical conditions invoke verify(...) on their component conditions. For instance, an IfThenElseCondition operates as follows:
```python
values = []

# if
if_result, if_value = self.if_condition.verify(*args, **kwargs)
if if_result:
    then_result, then_value = self.then_condition.verify(*args, **kwargs)
    values.append(then_value)
    return then_result, values

# else
if isinstance(self.else_condition, AccessControlCondition):
    # actual condition
    else_result, else_value = self.else_condition.verify(*args, **kwargs)
else:
    # boolean value
    else_result, else_value = self.else_condition, self.else_condition

values.append(else_value)
return else_result, values
```

If any component condition's `verify()` call raises an exception, the entire `IfThenElseCondition` also raises an exception, halting execution immediately. This also applies to other logical conditions eg. (the `or` `CompoundCondition`)

Suppressing exceptions entirely isn’t always feasible, as certain individual conditions—such as `JWTCondition`—may need to raise exceptions to indicate failure properly.

However, there may be scenarios where users of logical conditions prefer to catch exceptions and treat the condition as failed rather than propagating the exception. This would allow execution to continue within the logical structure. Such behaviour would need to be explicitly enabled by the user (i.e., opt-in), ensuring that exceptions are only suppressed when explicitly permitted.

We can consider introducing this as an optional feature in the future. Since it would be opt-in, it would maintain backward compatibility with older versions of the condition that don't have the flag set.

PR #3706, opened 2026-02-11
fix(conditions): Catch ConditionEvaluationFailed in OR/AT_LEAST branches

theref:
When a JSON path query finds no matches, `ConditionEvaluationFailed` is raised. In compound OR conditions this killed the entire evaluation, even when other branches would return True.

For OR and AT_LEAST operators, catch `ConditionEvaluationFailed` per operand and treat it as False so remaining branches can still satisfy the condition. AND and NOT still propagate the exception since they require all operands to be evaluable.

### DIFF
--- a/newsfragments/3706.bugfix.rst
+++ b/newsfragments/3706.bugfix.rst
@@ -1,0 +1,4 @@
+Handle ``ConditionEvaluationFailed`` in compound OR and AT_LEAST conditions.
+Previously, if one operand raised (e.g. a JSON path with no matches), the entire
+compound condition failed even when other branches would satisfy it. Failing
+operands are now treated as ``False`` for OR and AT_LEAST operators.
--- a/nucypher/policy/conditions/lingo.py
+++ b/nucypher/policy/conditions/lingo.py
@@ -279,8 +279,15 @@ class CompoundCondition(MultiCondition):
         values = []
         num_passed_conditions = 0
         overall_result = True if self.operator == self.AND_OPERATOR else False
+        _tolerant = self.operator in (self.OR_OPERATOR, self.AT_LEAST_OPERATOR)
         for condition in self.operands:
-            current_result, current_value = condition.verify(*args, **kwargs)
+            try:
+                current_result, current_value = condition.verify(*args, **kwargs)
+            except ConditionEvaluationFailed as e:
+                if not _tolerant:
+                    raise
+                current_result = False
+                current_value = str(e)
             values.append(current_value)
             if current_result:
                 num_passed_conditions += 1
--- a/tests/unit/conditions/test_compound_condition.py
+++ b/tests/unit/conditions/test_compound_condition.py
@@ -5,6 +5,7 @@ import pytest
 
 from nucypher.policy.conditions.base import Condition
 from nucypher.policy.conditions.exceptions import (
+    ConditionEvaluationFailed,
     InvalidCondition,
     InvalidConditionLingo,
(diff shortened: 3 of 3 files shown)


theref:
@derekpierre @cygnusv @KPrasch what are your thoughts on this issue? I'm in favour of having conditions branches that error, return False. If the other branch returns True, the condition should evaluate to True.

I can understand the worry about not surfacing errors though. Are there any other issues I'm not seeing?

cygnusv:
> @derekpierre @cygnusv @KPrasch what are your thoughts on this issue? I'm in favour of having conditions branches that error, return False. If the other branch returns True, the condition should evaluate to True.
> 
> I can understand the worry about not surfacing errors though. Are there any other issues I'm not seeing?

There's been some previous discussion on the matter. See https://github.com/nucypher/nucypher/pull/3570#discussion_r1907323547 and the follow-up issue https://github.com/nucypher/nucypher/issues/3582. Also see this commit for a WIP PR  https://github.com/nucypher/nucypher/pull/3670/changes/3d775d0e2ce3784b84661f906dbdfe0eac0619e9, where I muse about the same thing.

derekpierre:
> @derekpierre @cygnusv @KPrasch what are your thoughts on this issue? I'm in favour of having conditions branches that error, return False. If the other branch returns True, the condition should evaluate to True.
> 
> I can understand the worry about not surfacing errors though. Are there any other issues I'm not seeing?

That's the biggest worry from my end.

My take, to add some colour to my description in #3582, is that your concept of "tolerance" should be controlled by the condition creator, not inferred / used as a default. In other words, it should be an explicit, opt-in boolean on applicable logical combinations of conditions (i.e., `MultiCondition` classes such as `if-then-else` or `compound`).

While this flag could technically be used with `sequential` or `compound(and)`, it wouldn’t change their intrinsic short-circuit behaviour (since false already stops evaluation). However, it would prevent exceptions from propagating if those conditions are nested within other higher-level conditions.

This would be an optional boolean field added to the `MultiCondition` schema, defaulting to false for backward compatibility. Condition creators could explicitly enable it when they want exceptions to be treated as non-fatal during evaluation.

For example, an `if-then-else condition` might look like:
```json
{
        "conditionType": "if-then-else",
        "ifCondition": CONDITION,
        "thenCondition": CONDITION,
        "elseCondition": CONDITION | true | false,
}
```
(current behaviour i.e. any exception propagated)

OR 
```json
{
        "conditionType": "if-then-else",
        "treatExceptionAsFalse": True,     <-- better field name needed :)
        "ifCondition": CONDITION,
        "thenCondition": CONDITION,
        "elseCondition": CONDITION | true | false,
}
```
(modified behaviour: any exception raised from `if`, `then` or `else` condition treated as if that condition returned False)

where "treatExceptionAsFalse" is optional and defaults to False if not defined (backwards-compatible) but can be overridden.

The flag would also allow the condition creator to specify selective tolerance by picking and choosing which sub-condition exceptions are tolerated and which are not, which was one of the concerns with the failed JWT condition since it is a cryptographic operation.

There is also the possibility to have one overall boolean flag for the entire lingo, but that seems too broad and inflexible.

Open to other thoughts/approaches/additions of course.

cygnusv:
A similar approach to `treatExceptionAsFalse` was explored in #3670, specifically https://github.com/nucypher/nucypher/pull/3670/changes/7814e9ebba1f4acf212d093fe656e7edefbcf4c8

PR #3723, opened 2026-03-05
feat(conditions): Add nestedAbiDecode for selectorless ABI data

theref:
## Problem

`nestedAbiValidation` decodes bytes by calling `decode_human_readable_call`, which always strips a 4-byte function selector from the front. This means it can only validate bytes that contain function-call-encoded data (selector + args).

Some use cases produce ABI-encoded bytes with **no selector** — just raw `eth_abi.encode(types, values)` output. There's currently no way to validate these.

## Solution

Add `nestedAbiDecode` as a new option on `AbiParameterValidation`, alongside the existing `nestedAbiValidation`. The difference:

| | `nestedAbiValidation` | `nestedAbiDecode` |
|---|---|---|
| **Input** | Function-call bytes (selector + args) | Raw ABI-encoded bytes (no selector) |
| **Config key** | `allowedAbiCalls` with function signatures | `type` with a Solidity type string |
| **Decodes via** | `decode_human_readable_call` | `decode_abi_encoded` (new) |
| **Inner validations** | Keyed by function signature | Flat `validations` list |

After decoding, both work the same way — you use `parameterIndex`, `subIndices`, `returnValueTest`, and further nesting as usual.

## What changed

**`AbiParameterValidation`** (the existing class) now accepts exactly one of three options:
- `returnValueTest` — compare a value directly (unchanged)
- `nestedAbiValidation` — decode as a function call and validate (unchanged)
- `nestedAbiDecode` — **new** — decode as raw ABI and validate

**`AbiDecodeValidation`** (new class) has two fields:
- `type`: a Solidity type string, e.g. `(address,uint256,bytes)[]`
- `validations`: list of `AbiParameterValidation` (same structure used everywhere else)

**`decode_abi_encoded`** (new utility in `abi.py`) — thin wrapper around `eth_abi.decode` that takes a type string and raw bytes. No selector stripping.

**`AbiCallValidation`** — updated the bytes-type guard to also check `nestedAbiDecode` (same as existing `nestedAbiValidation` check).

**TypedDict** — added `AbiDecodeValidationDict`, updated `AbiParameterValidation`.

## JSON example

```json
{
  "parameterIndex": 1,
  "nestedAbiDecode": {
    "type": "(address,uint256,bytes)[]",
    "validations": [
      {
        "parameterIndex": 0,
        "subIndices": [0, 0],
        "returnValueTest": { "comparator": "==", "value": "0x036C..." }
      },
      {
        "parameterIndex": 0,
        "subIndices": [0, 2],
        "nestedAbiValidation": {
          "allowedAbiCalls": {
            "transfer(address,uint256)": [
              { "parameterIndex": 0, "returnValueTest": { "comparator": "==", "value": ":recipientAA" } }
            ]
          }
        }
      }
    ]
  }
}
```

Note: `nestedAbiDecode` validations can themselves use `nestedAbiValidation` (or another `nestedAbiDecode`) for further nesting.

## Tests

- **Integration**: Build batch calldata (outer function call wrapping raw-encoded inner data), validate the full chain including nested function calls inside the decoded data
- **Serialization**: Round-trip to dict and back
- **Rejection**: Wrong inner value → `allowed=False`
- **Schema**: `nestedAbiDecode` on a non-`bytes` parameter type is rejected

### DIFF
--- a/newsfragments/3723.feature.rst
+++ b/newsfragments/3723.feature.rst
@@ -1,0 +1,1 @@
+Add ``nestedAbiDecode`` for validating selectorless ABI-encoded data in signing conditions, enabling ERC-7579 batch execution support.
--- a/nucypher/policy/conditions/signing/base.py
+++ b/nucypher/policy/conditions/signing/base.py
@@ -28,6 +28,7 @@ from nucypher.policy.conditions.utils import (
     is_camel_case,
 )
 from nucypher.utilities.abi import (
+    decode_abi_encoded,
     decode_human_readable_call,
     extract_arg_types,
     is_valid_human_readable_signature,
@@ -182,6 +183,11 @@ class AbiParameterValidation(_Serializable):
             lambda: AbiCallValidation.Schema(), allow_none=True, required=False
         )
 
+        # ...or a nested ABI decode (no selector) to decode and evaluate
+        nested_abi_decode = fields.Nested(
+            lambda: AbiDecodeValidation.Schema(), allow_none=True, required=False
+        )
+
         # maintain field declaration ordering
         class Meta:
             ordered = True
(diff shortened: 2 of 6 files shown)


derekpierre:
Is it possible to subsume the `AbiDecodeValidation` logic into the existing `AbiCallValidation` object instead of creating a new one? You can rename `AbiCallValidation` to something more general like `AbiValidation` or ...

Seems like the only difference is whether the 4-byte selector is present or not in the signature, and you can subsequently key by the abi string. Everything else seems analogous, unless I've missed something.

Making the existing class more flexible would reduce any repetition of common logic.

PR #3726, opened 2026-03-19
Adapt nucypher to post-migration changes

cygnusv:
**Type of PR:**
- [ ] Bugfix
- [ ] Feature
- [ ] Documentation
- [ ] Other

**Required reviews:** 
- [ ] 1
- [ ] 2
- [X] 3

**What this does:**
Depends on:
- https://github.com/nucypher/nucypher-contracts/pull/466
- https://github.com/nucypher/nucypher-contracts/pull/470

**Issues fixed/closed:**
> - Fixes #...

**Why it's needed:**

**Notes for reviewers:**


nucypher/blockchain/eth/agents.py
```diff
@@ -1188,9 +1157,6 @@ def get_agent(
 
     @staticmethod
     def _contract_name_to_agent_name(name: str) -> str:
-        if name == NUCYPHER_TOKEN_CONTRACT_NAME:
```
derekpierre, line 1191: Love this removal. I never liked the special case we had to add here.

tests/integration/blockchain/test_currency.py
derekpierre, line 1: Do we have a separate test for the base `ERC20` class or was this really the only currency test we had? If the latter, then perhaps we keep the test but change from using `NU` to instead using `T` or some made up subclass of `ERC20`?

derekpierre:
Surgical stuff here 🎸

tests/integration/blockchain/test_currency.py
cygnusv: Good point. I restored the test, using T isntead of NU. Also relocated it to tests/unit, since it's not an integration test.

vzotova (approved):

derekpierre (approved):
🎸 - nice work!

We'll merge the `nucypher-contracts` PR first, then reset the `ape-config.yaml` in this PR before merging.

derekpierre:
Just realized, we should also copy over the updated `lynx.json` contract registry from https://github.com/nucypher/nucypher-contracts/pull/470 in this PR or a follow-up.

